PR #240, opened 2020-05-20
Improves Dutch translations

jessehouwing:
While I'm generally not a fan of translated software, the Dutch translations could use quite a bit of love.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winget-cli/240)

### DIFF
--- a/Localization/Resources/nl-NL/winget.resw
+++ b/Localization/Resources/nl-NL/winget.resw
@@ -223,7 +223,7 @@
     <value>Argumenten die worden doorgegeven aan het installatieprogramma overschrijven</value>
   </data>
   <data name="PendingWorkError" xml:space="preserve">
-    <value>Opes, dit zijn we vergeten...</value>
+    <value>Oeps, dit zijn we vergeten...</value>
   </data>
   <data name="PlainArgumentDescription" xml:space="preserve">
     <value>Weergave van de voortgang in de standaardkleur</value>
--- a/Localization/Resources/nl-NL/winget.resw
+++ b/Localization/Resources/nl-NL/winget.resw
@@ -147,7 +147,7 @@
     <value>Niet meer dan het opgegeven aantal resultaten weergeven</value>
   </data>
   <data name="ExactArgumentDescription" xml:space="preserve">
-    <value>App vinden met een exacte overeenkomst</value>
+    <value>App vinden op basis van een exacte overeenkomst</value>
   </data>
   <data name="ExtraPositionalError" xml:space="preserve">
     <value>Er is een positioneel argument gevonden, terwijl er geen werd verwacht</value>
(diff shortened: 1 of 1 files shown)


roycornelissen:
There’s a typo in there... “Opes” >> “Oeps”

JohnMcPMS:
@jessehouwing ,

We will have to investigate how to integrate your suggestions into our localization pipeline.  Since we are using the internal localizations services of Microsoft, just accepting your change would be overwritten the next time we take a drop.  I'll follow up when I know more.

Thanks for your interest.

jessehouwing:
Some of the current translations make very little sense. The rest is an improvement and mostly aligns the sentence structure of sentences that convey a similar message.

thlac:
@JohnMcPMS I was actually thinking of a PR of my own for Danish, so I'm sorry for "hijacking" this, but any way it's possible to take up the general issues with using those internal localization services? For example in this project I see you translate the word "Preview" into a Danish word that is more akin to the english word "Sample".

I've also seen this become a trend in Windows, just to give a few examples, there's translation to danish like, "Browse Directory"(as in a folder on the file system) into "Browse Active Directory" or stuff like a TPM (Trusted Platform Module) into TMP (Trusted Module Platform?).

jessehouwing:
The main issue I saw in the Dutch translation:

 * Some words are the wrong synonym. 
    * Preview => 'Voorbeeld' Example, 
 * Different words for the same concept: app: 'programma', 'app', 'toepassing', 'hulpmiddel'
 * Some of the command descriptions start with "hiermee wordt" => "This will do...".  Others with "Does X".

It looks like multiple people have worked on the translations independently, or they haven't seen the translations in action.

Some terms are super hard to translate, because the English word is also commonly used in Dutch. "Preview", "Tool". Some words don't have a concept yet in the language. I don't think most people will know what a 'verbonden vlagalias" (adjoined flag alias) is. But then again, I hardly understand the English term, let alone the very literal translation. It's why I left that untranslated.

thlac:
Yeah I saw most of the same in the danish translation, personally I don't think "Preview" should be translated at all, we have a word for like "forpremiere" or "snigpremiere" but really it's kind of like how a word like e-mail or app just translates very poorly to the given language.

The 2 other points I think is an issue with the actual language usage throughout winget as I see the same inconsistency on the English version. Especially "Apps" vs. "Applications" stuck out to me as it's in the help text for the command.

JohnMcPMS:
@thlac , no I don't think you can directly do that.  The only option that I'm aware of is either using Feedback Hub, which will create a bug that can be routed internally, or something like a PR.  The nice thing about the PR here is we can point the internal team to it, where they can see the string's symbolic name and I assume it makes the process easier.

Any inconsistencies in the English are probably my fault ☹️ Feel free to submit a PR for those as well, as I can actually review them myself.

@jessehouwing , I have submitted this PR to the localization team.  I cannot guarantee that the strings will come out exactly as you have translated them though.  I'm going to leave the PR open, but the update will come in another PR that I will link with this one.

thlac:
@JohnMcPMS Sure, I did do that, and I also think the localization bugs I mentioned are fixed already, mostly just meant it as general feedback, I really love what Microsoft has done with stuff like docs, I just hope over time you can do something similar for translations :)

If nobody else have beat me to it, I'll see about making a pull request regarding the inconsistent language when I get up tomorrow.

JohnMcPMS:
Closing this since I've merged #474 .  If there are any more changes needed, opening a new PR will be the best way to get them in.

Thanks for your help 😄

PR #283, opened 2020-05-22
Update resource file app(s) -> application(s)

thlac:
A pull request about the language use we talked about in #240 

I have some notes about the changes, we can discuss:
1. I am not sure if we should use the word "app" or "application", there are already requests in the pipeline to support stuff like fonts, my opinion is that it _should_ be packages.
2. The above not withstanding, I think it should be "application" over "app", my two main reasons being that "application" translates better to other languages, and "app" I think infers something I would get from the Store exclusively.
3. I've stayed true to the original text, but I think SourceArgumentDescription should probably be "applications", the source of your repository can still result in multiple results. (I think ExactArgumentDescription should stay singular as that parameter will result in one exact result), but maybe both of them shouldn't talk about applications at all, but results, which would be in line with the rest of parameters.

If we change consensus, I can update the pull request accordingly.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/283)

### DIFF
--- a/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
+++ b/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
@@ -151,7 +151,7 @@
     <value>Show no more than specified number of results</value>
   </data>
   <data name="ExactArgumentDescription" xml:space="preserve">
-    <value>Find app using exact match</value>
+    <value>Find application using exact match</value>
   </data>
   <data name="ExtraPositionalError" xml:space="preserve">
     <value>Found a positional argument when none was expected</value>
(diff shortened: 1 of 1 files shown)


JohnMcPMS:
@denelon , @KevinLaMS 

I had originally used application (as is quite evident in the type names throughout) as the generic term for what we are managing.  Should we switch the documentation to use the term package?  I can guess your responses, but this way you will be on record 😄

denelon:
I have had plenty of discussion with others on the topic of "package" vs. "App" vs. "Application" etc.

I would lean in the direction of the generic "package" as the largest set of things the client can install.

PRIMETSS:
Yes, agree, 'Package' is so more generic and a bit more future proof. It;s a 'Package Manager', 'Packages of items' should be what it handles. The term App  or even Application muddies the waters

JohnMcPMS:
The change to use the term packages everywhere was released as part of https://github.com/microsoft/winget-cli/releases/tag/v0.1.42101-preview 

Closing out this PR, thanks for the feedback.

PR #294, opened 2020-05-22
fix files without carriage return on the last line

flug:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/294)


JohnMcPMS:
Sorry, I just don't think we need to update 243 files just to add newlines.

PR #295, opened 2020-05-22
Remove phrase

phrocker:
Just a little cleanup!

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/295)

### DIFF
--- a/src/YamlCppLib/yaml-cpp/src/scanscalar.cpp
+++ b/src/YamlCppLib/yaml-cpp/src/scanscalar.cpp
@@ -73,7 +73,7 @@ std::string ScanScalar(Stream& INPUT, ScanScalarParams& params) {
         continue;
       }
 
-      // otherwise, just add the damn character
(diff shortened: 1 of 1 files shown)


JohnMcPMS:
While I agree with the sentiment, this is a subtree of the [yaml-cpp](https://github.com/jbeder/yaml-cpp) project.  I would rather not update the file directly so as to not have the management overhead of potential merge conflicts.

phrocker:
I see that now. I'll change the source of the pr and send it there. Thanks!

JohnMcPMS:
Policheck is mad at us for this now, and I see that your PR to yaml-cpp didn't result in success.  So I guess we have to take the overhead and make the change.  Let's see how well this PR works when necro'd.

PR #296, opened 2020-05-22
Fix #137: Add SuccessExitCode to ManifestInstaller

### LINKED ISSUE #137: Support non-zero exit codes as success

Some applications return non-zero codes that are not errors. There should be a way for the client to know the difference between these codes and actual errors.

A key could be added to the manifest syntax to indicate which non-zero codes should be treated as success. The client could then treat these non-zero return codes as success.

chausner:
This PR suggests a fix for #137 by adding an optional `SuccessExitCode` attribute to the Installers entries in the manifest. If present, it overrides the default value of zero.

For now, this only allows a single value to be defined as success status code. I am not sure if there are cases where we would want multiple exit codes to count as success.

Right now, SuccessExitCode is not printed by the "show" command but we could add it if desired.

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
@@ -138,6 +138,7 @@ namespace AppInstaller::CLI::Workflow
     {
         context.Reporter.Info() << "Installing ..." << std::endl;
 
+        const auto& installer = context.Get<Execution::Data::Installer>();
         const std::string& installerArgs = context.Get<Execution::Data::InstallerArgs>();
 
         auto installResult = context.Reporter.ExecuteWithProgress(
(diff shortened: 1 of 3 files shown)


JohnMcPMS:
We have a manifest versioning strategy that isn't really mature in the codebase yet, but we want to ensure we build that muscle/work out the kinks.  So any new additions to the manifest would need increment the manifest version and be interpreted appropriately.

In addition, I think we probably do want to support multiple exit codes for success.  While the uses are probably low, it would be better to support it from the beginning than to have a versioning problem later on.  But this kind of thing is best worked out in a spec (which I think we need to document the process better).  See https://github.com/microsoft/winget-cli/projects/1

chausner:
Thank you for your feedback. I'd be happy to make changes once the spec is finalized and the details have been clarified.

benferse:
@JohnMcPMS The canonical example that comes to mind are the "pseudo-success" codes that are returned for installs that require a reboot to complete:

- ERROR_SUCCESS_REBOOT_REQUIRED
- ERROR_SUCCESS_REBOOT_INITIATED
- ERROR_INSTALL_SUSPEND

bhavik-sp:
Change docs

yao-msft:
Thanks for the contribution. This pr is included as part of #778

Issue #268, opened 2020-05-21
search output doesn't align properly with foreign characters in the Name


smabuk:
# Description
If the Name of an application has foreign characters in it then the other columns do line up correctly.

I assume that multi-byte characters are causing the the display length to be incorrectly calculated.

# Steps to reproduce

```none
winget search lo
```

# Expected behavior

The start of all of the columns should line up vertically.

# Actual behavior

The row with LÖVE is incorrect.

```none
Name                    Id                             Version        Matched
------------------------------------------------------------------------------------------------
OpenVPN                 OpenVPNTechnologies.OpenVPN    2.4.9
Azure Storage Explorer  Microsoft.AzureStorageExplorer 1.13.1
LastPass                LogMeIn.LastPass               4.43.0.7
Logitech Harmony Remote Logitech.Harmony               1.0.1.308
CozyDrive               CozyCloud.CozyDrive            3.20.0
Winlogbeat              Elastic.Winlogbeat             7.7.0
Slack                   SlackTechnologies.Slack        4.5.1
LÖVE                   Love2d.Love2d                  11.3
Maxthon 5 Cloud Browser Maxthon.Maxthon5               5.3.8.2000
ConfigMgrToolkit        Microsoft.ConfigMgrToolkit     5.00.7958.1151 Tag: Deployment Monitoring
Evernote                evernote.evernote              6.24.2.8919    Tag: cloud
Dropbox                 Dropbox.Dropbox                94.4.384       Tag: cloud
Postman                 Postman.Postman                7.24.0         Tag: development
Gpg4win                 gnupg.Gpg4win                  3.1.11         Tag: cryptology
```

# Environment

```none
[winget --info]
Windows Package Manager v0.1.41331 Preview
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19041.264
Package: Microsoft.DesktopAppInstaller v1.0.41331.0
```

PR #313, opened 2020-05-23
Fix misalignment of table output for multi-byte chars (#268)

chausner:
This PR suggests a fix for #268. The problem is that `std::string::length()` is used in `TableOutput` to determine the length of table cell values. Since `std::string` is not UTF8-aware, it returns the number of bytes, not characters. The two are different if the string contains special characters.

This fix adds a `Utility::UTF8Length` method to obtain the number of characters of a UTF8 string, and uses it in `TableOutput` to compute the display length of cell values.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/313)

### DIFF
--- a/src/AppInstallerCommonCore/AppInstallerStrings.cpp
+++ b/src/AppInstallerCommonCore/AppInstallerStrings.cpp
@@ -58,6 +58,19 @@ namespace AppInstaller::Utility
         return result;
     }
 
+    size_t UTF8Length(std::string_view input)
+    {
+        if (input.empty())
+        {
+            return 0;
+        }
+
+        int utf16CharCount = MultiByteToWideChar(CP_UTF8, 0, input.data(), wil::safe_cast<int>(input.length()), nullptr, 0);
+        THROW_LAST_ERROR_IF(utf16CharCount == 0);
+
+        return wil::safe_cast<size_t>(utf16CharCount);
+    }
+
     std::string Normalize(std::string_view input, NORM_FORM form)
     {
         if (input.empty())
--- a/src/AppInstallerCommonCore/AppInstallerStrings.cpp
+++ b/src/AppInstallerCommonCore/AppInstallerStrings.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 #include "pch.h"
 #include "Public/AppInstallerStrings.h"
+#include "icu.h"
 
 namespace AppInstaller::Utility
 {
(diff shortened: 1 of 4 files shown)


nikeee:
If I understand this code correctly, it converts the UTF-8 string into an UTF-16 string and counts the number of wchar_t.
If I'm not mistaking, this would suffer from the same issue as the current implementation, pro problem just occurs on more exotic chars like Chinese or Emojis.
Counting the number of grapheme clusters would be the way to go, I think.
See: https://mortoray.com/2016/04/28/what-is-the-length-of-a-string-a-tricky-question/

chausner:
> If I'm not mistaking, this would suffer from the same issue as the current implementation, the problem just occurs on more exotic chars like Chinese or Emojis.

I was wondering about this, too, and you are probably right.

In any case, the way UTF-8 is currently handled in the code base (in plain `std::string`) does not seem ideal because it is very likely that similar mistakes will slip into the code in the future again. C++20 has `std::u8string` but I guess the project cannot use it yet.

JohnMcPMS:
Unfortunately u8string won't help either, as @nikeee points out what we really want to count are grapheme clusters.  I was aware of this when I wrote it, but I:
1. Mistakenly thought that it would simply result in a wider column (not thinking that different strings would have different disparities in encoded vs cluster count).
2. Didn't find a simple Windows API to handle it.
3. Ok, I was aware that there were issues, but not to the depths of terminology / semantics around Unicode and characters...

I think that the right thing to do is to implement "character" breakpoint iteration counting from ICU.  Since we use the version of ICU that ships with Windows, that means using the C API as described http://userguide.icu-project.org/boundaryanalysis

In addition to UTF8Length, the TableOutput as written would also need "UTF8Substring" for truncating strings properly.  Also, since we are on the subject, is the ellipses (...) a language independent construct for indicating truncation?

I do appreciate this fix for my haste, but I don't want to sweep it under the rug by fixing only some of the problem.

chausner:
> I think that the right thing to do is to implement "character" breakpoint iteration counting from ICU. Since we use the version of ICU that ships with Windows, that means using the C API as described http://userguide.icu-project.org/boundaryanalysis
> 
> In addition to UTF8Length, the TableOutput as written would also need "UTF8Substring" for truncating strings properly.

I changed the implementation to use the code you linked. I also added a `UTF8Substring` method.

Independent from that, in bb92fd9, I changed the three dots into a single ellipsis character. This wastes a little less space and allows to show two more characters of the abbreviated string. But feel free to revert this if you preferred the three dots.

> Also, since we are on the subject, is the ellipses (...) a language independent construct for indicating truncation?

According to [Wikipedia](https://en.wikipedia.org/wiki/Ellipsis), ellipsis seems common in other languages, as well. But some languages use variations of it, e.g. in Chinese they use six instead of three dots.

JohnMcPMS:
Looks great @chausner 

I would probably move the ICU setup boilerplate into a separate type just so its only written once.  And you can use wil::unique_any_t to define a smart pointer without having to use scope_exits.

Yes, Wikipedia does seem to suggest that it is used in other languages, although it also points to varying ways in which it is represented.  I think that your move to use the single character ellipsis may actually prove to be the right thing with regards to this in any case, both because it shows more characters and because it may just render as something more appropriate in the user's chosen font.

JohnMcPMS:
I made the changes that I was suggesting on your branch.  Let me know how you feel about them.

/azp run

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -12,6 +14,55 @@ namespace AppInstaller::Utility
...
+                if (U_FAILURE(err))
+                {
+                    AICLI_LOG(Core, Error, << "ubrk_setUText returned " << err);
+                    THROW_HR(E_UNEXPECTED);
+                }
+
+                int32_t i = ubrk_first(m_brk.get());
+                if (i != 0)
+                {
+                    AICLI_LOG(Core, Error, << "ubrk_first returned 0");
```
chausner, line 50: The log message should read "returned non-zero".

chausner:
> I made the changes that I was suggesting on your branch. Let me know how you feel about them.

Just the one typo in the log message, otherwise looks good to me. 👍

JohnMcPMS:
Resolves #268

PR #341, opened 2020-05-27
Automatically supply default silent args when possible

wjk:
This is a rough first draft. Silent args for NSIS and InnoSetup were copied from Chocolatey scripts that call installers written using those technologies. Silent args for Burn were copied from the help text of a Burn-based installer for one of my projects; I understand all Burn-based bootstrappers use the same syntax. Feedback is welcome. Thanks!

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/341)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
@@ -65,7 +65,20 @@ namespace AppInstaller::CLI::Workflow
             const std::map<ManifestInstaller::InstallerSwitchType, Utility::NormalizedString>& installerSwitches = context.Get<Execution::Data::Installer>()->Switches;
 
             // Construct install experience arg.
-            if (context.Args.Contains(Execution::Args::Type::Silent) && installerSwitches.find(ManifestInstaller::InstallerSwitchType::Silent) != installerSwitches.end())
+            auto installerType = context.Get<Execution::Data::Installer>()->InstallerType;
+            if (context.Args.Contains(Execution::Args::Type::Silent) && installerType == ManifestInstaller::InstallerTypeEnum::Nullsoft)
+            {
+                installerArgs += "/S";
+            }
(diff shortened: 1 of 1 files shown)


src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -65,7 +65,20 @@ namespace AppInstaller::CLI::Workflow
             const std::map<ManifestInstaller::InstallerSwitchType, Utility::NormalizedString>& installerSwitches = context.Get<Execution::Data::Installer>()->Switches;
 
             // Construct install experience arg.
-            if (context.Args.Contains(Execution::Args::Type::Silent) && installerSwitches.find(ManifestInstaller::InstallerSwitchType::Silent) != installerSwitches.end())
+            auto installerType = context.Get<Execution::Data::Installer>()->InstallerType;
+            if (context.Args.Contains(Execution::Args::Type::Silent) && installerType == ManifestInstaller::InstallerTypeEnum::Nullsoft)
+            {
+                installerArgs += "/S";
```
yao-msft, line 71: >installerArgs += "/S"; [](start = 16, length = 22)

Thanks for your input.

In current implementation, the default args for known installer types are stored in
ManifestInstaller::GetDefaultKnownSwitches as a map in [https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerRepositoryCore/Manifest/ManifestInstaller.cpp](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerRepositoryCore/Manifest/ManifestInstaller.cpp)

The default args are auto populated during manifest parsing if the original manifest does not declare the args.

Please update there if you feel the default args in our map needs improvement.

Thanks.

wjk:
@yao-msft I didn’t know that. The current switches in ManifestInstaller.cpp look good to me. Closing this PR. Thanks!

Issue #292, opened 2020-05-22
winget should install an app if there is an exact match (eg 'git')


mikemaccana:
# Brief description of your issue

```
C:\Program Files $ winget install Git
Multiple apps found matching input criteria. Please refine the input.
Name                                    Id                                       Version      Matched
---------------------------------------------------------------------------------------------------------
Git                                     Git.Git                                  2.26.2
Logitech Harmony Remote                 Logitech.Harmony                         1.0.1.308
GitHub Desktop                          GitHub.GitHubDesktop                     2.5.0
GitHub CLI                              GitHub.cli                               0.8.0
Atom                                    GitHub.Atom                              1.45.0
AdobeDigitalEditions                    Adobe.AdobeDigitalEditions               4.5.11
MicrosoftGitCredentialManagerforWindows Microsoft.GitCredentialManagerforWindows 1.20.0
Git Large File Storage                  GitHub.GitLFS                            2.11.0
Git Extensions                          GitExtensionsTeam.GitExtensions          3.3.1
MaxTo                                   DigitalCreations.MaxTo                   2.0.1
Gitter IM                               Gitlab.Gitter.IM                         4.1.0
GitKraken                               Axosoft.GitKraken                        7.0.0
TortoiseGit                             TortoiseGit.TortoiseGit                  2.10.0.2
AppInstallerFileBuilder                 Microsoft.AppInstallerFileBuilder        1.2020.211.0 Tag: GitHub
Sublime Merge                           SublimeHQ.SublimeMerge                   1119         Tag: git
```

# Steps to reproduce

`winget install Git`

# Expected behavior

git installs, as the name is an exact match (case insensitive).

# Actual behavior

The error above.

# Environment

```none
Windows Package Manager v0.1.41331 Preview
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19628.1
Package: Microsoft.DesktopAppInstaller v1.0.41331.0

Links:
  Privacy Statement: https://aka.ms/winget-privacy
  License agreement: https://aka.ms/winget-license
  3rd Party Notices: https://aka.ms/winget-3rdPartyNotice
  Homepage:          https://aka.ms/winget
```

CalvinAllen:
I'm having this exact issue with "Powershell" and/or "Microsoft.Powershell", and I have no idea how to actually get the one I want.

kipters:
[...]

PR #358, opened 2020-05-29
Create #292 - winget should install an app if there is an exact match.md

denelon:


doc/specs/#292 - winget should install an app if there is an exact match.md
```diff
@@ -0,0 +1,75 @@
...
+issue id: 292
+---
+
+# Package matching bug fix
+
+## Abstract
+
+The winget.exe client attempts to be generous with the `search` command, but is a bit too generous with `install`. The *id*
+should be the unique key to identifying a package. It should also be case insensitve from the perspective of command execution,
+but it should be case sensitive in terms of the displayed value.
```
megamorf, line 14: Could you try rephrasing `it should be case sensitive in terms of the displayed value.` - do you mean that it's expected to display the original casing used in the manifest instead of a (lowercase, camelCase, PascalCase) formatted string?

doc/specs/#292 - winget should install an app if there is an exact match.md
```diff
@@ -0,0 +1,75 @@
...
+
+Users may not realize another package exists with a longer name, and may install a program other than what they intended.
+If "git.git" and "git.gity" both existed and the user were to press enter before typing the final "y", the git.git package would
+be installed rather than what they intended.
+
+## Future considerations
+
+There are changes to how results should displayed to reduce the liklihood of a user mistakenly assuming the "Name" is a key value
+for a package. The client commands should be case insensitive, but the display should still be case sensitive to support ease of
+reading. Long names can be easier to understand when presnted in camel case, pascal case, or with a branded letter casing.
```
megamorf, line 71: typo: presented

doc/specs/#292 - winget should install an app if there is an exact match.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Potential Issues
+
+Users may not realize another package exists with a longer name, and may install a program other than what they intended.
+If "git.git" and "git.gity" both existed and the user were to press enter before typing the final "y", the git.git package would
+be installed rather than what they intended.
+
+## Future considerations
+
+There are changes to how results should displayed to reduce the liklihood of a user mistakenly assuming the "Name" is a key value
```
megamorf, line 69: typo: likelihood

doc/specs/#292 - winget should install an app if there is an exact match.md
```diff
@@ -0,0 +1,75 @@
...
+last updated: <2020-05-28>
+issue id: 292
+---
+
+# Package matching bug fix
+
+## Abstract
+
+The winget.exe client attempts to be generous with the `search` command, but is a bit too generous with `install`. The *id*
+should be the unique key to identifying a package. It should also be case insensitve from the perspective of command execution,
```
zachcarp, line 13: Version should play a roll in identifying a package. We can default to a specific version where this is concerned, but we should keep it in mind as we move forward.

doc/specs/#292 - winget should install an app if there is an exact match.md
denelon: More sleep reduces mistakes... Maybe an IDE with a spell check wouldn't hurt.

doc/specs/#292 - winget should install an app if there is an exact match.md
denelon: I will call version out.

megamorf:
Looks good to me 👍

Issue #117, opened 2020-05-11
Install Microsoft Store Apps


denelon:
From MVP Summit feedback:

Will there be a way to automate installation of Store apps? I have a script I run to install my programs after Windows reinstall with Chocolatey, but I have to install Microsoft Store apps manually afterwards, which is a bit annoying.

maxkatz6:
How should it work with MS Store Policies?

**10.1.5**
Your app may promote or distribute software only through the Microsoft Store.

https://docs.microsoft.com/en-us/windows/uwp/publish/store-policies#101-distinct-function--value-accurate-representation

By the way, I would love to see store apps in winget too.

Jasius:
> How should it work with MS Store Policies?
> 
> **10.1.5**
> Your app may promote or distribute software only through the Microsoft Store.
> 
> https://docs.microsoft.com/en-us/windows/uwp/publish/store-policies#101-distinct-function--value-accurate-representation
> 
> By the way, I would love to see store apps in winget too.

I think you're misunderstanding what it means. It means that lets say a DLC for a game would be purchasable/downloadable through Windows store instead of leading you to third party service.

yaichenbaum:
@maxkatz6 I don't think it's an issue since Microsoft can make exceptions for it's own software, for example if you try to get Microsoft Office through the store it will redirect you to download the installer from the Microsoft website.

blogcraft:
> @maxkatz6 I don't think it's an issue since Microsoft can make exceptions for it's own software, for example if you try to get Microsoft Office through the store it will redirect you to download the installer from the Microsoft website.

I hate when that happens (if I used the store I expecto the store download the app and updates it when neccesary)

ThomasAunvik:
I feel like it should have already been an option to install an app from microsoft store via cli. Since that would be just sending install arguments to the executable.
But im guessing there isn't and that's why we now have wget.
Cant seem to get anything out of `WinStore.App.exe` executable either (missing dll errors).

RobCannon:
The Microsoft Store never really had that option.  There are some ways to sideload apps from a command line.  If the Store had that command line option and it embraced other application types (like command line apps and Win32 apps), we probably wouldn't have needed this project.

yaichenbaum:
I seem to remember that the Microsoft Store does have a way to install apps through a cli, it just isn't available for use outside of Microsoft.

aetos382:
[...]

PR #367, opened 2020-05-29
Draft of Store Support

KevinLaMS:
Per issue 117, we need to add store support.  This is a draft of the requirements to help drive the dev spec.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/367)


doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
+---
+author: <kevin> <larkin> <kevinlams/<kevinla@microsoft.com>
+created on: <2020-05-28>
+last updated: <2020-05-28>
+issue id: <github issue id>
```
yao-msft, line 5: Missing issue id

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+The Windows Store is a curated list of applications.  The store apps include tools and the best of class applications and must be available to Package Manager users.
+
+## Solution Design
+
+Windows Package Manager will consume a manifest file like it does with all other formats.  The Windows Package Manager will then use a store API to initiate downloads and install the apps.
+
+
+
+**Store Apps**
+The Store has multiple app types.  Those that are free for download and free, but with purchase options, as well as paid apps.
```
yao-msft, line 27: Let me try the "free for download" apps and see how the API handles the case when user does not own the app yet

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+***Free Apps***  
+Apps that are free, will be downloaded directly when requested.
+
+***Pay apps [NOT SUPPORTED]***
```
yao-msft, line 32: Paid apps?

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+Apps that require in app purchase, can be allowed, as long as the app can correctly handle activation and treat the in-app purchase out of band from the Windows Package Manager.
+
+## Manifest Requirements 
+To support store apps, the manifest file will need to be updated to support the new format.  Most fields remain the same, but additional changes are required to support the store.
+
+### YAML Spec updates
+<b>URL</b> the URL used for the store is not the same as the URL users and the developers expect.  Therefore this field is unnecessary.  
+<b>REQUIRED: NEW: StoreID</b> the Store ID will be used to identify the app. example: 9nblggh4nns1   
+<b>InstallerType</b> A new format, the WindowsStore type will be provided
+<b> InstallerType - Default </b> this needs to be defined, but essentially the author can specify if the user does not select anything... which URL to default to.
```
yao-msft, line 46: Can you explain a bit more what this means?
I thought we already have enough info on which route to proceed.
InstallerType: WindowsStore -> Use StoreId
InstallerType: msi,exe, etc -> Use Url

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+<b>Architecture</b> The architecture will support the architectures supported by the store.  neutral will most likely be the default.
+<b>-prelease</b> In order to support flighing, the client will need to support only showing apps with out the prerelease flag AND enabling the user override on the command line.
+
+### Flighting
+The store provides the ability to flight apps to partners.  But you must be logged in with your MSA to get the update.  By default the Store App support will be for providing the user with the GA version.  
+
+#### MSA Credentials
+In order to support flighting, the Windows Package Manager would need to pass credentials to the Windows Store.  This is not in scope for this release, therefore only GA apps are enabled.  
+
+Developers wishing to flight store apps will need to create a different app and BigID in order to make it available for flighting.  Developers will use the -preview flag in the version to restrict access to those that want prerelease builds.
```
yao-msft, line 61: Nit: BigID should be StoreId to be consistent :)

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+PS C:\> winget search appinstaller
+Name                    Id                                Version
+----------------------------------------------------------------------
+AppInstallerFileBuilder Microsoft.AppInstallerFileBuilder 1.2020.211.0
+```
+And Install
+### Store App installation experience
+Installing the store app will provide the MSIX installation experience.  The Windows Store App will stream install.
+
+![](images\117-Download.png)
```
yao-msft, line 79: Somehow the image does not resolve in rich view

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+
+## Manifests
+Developers will be able to submit their app by providing meta data that includes the store ID or URL to the store app.
+
+### Duplicate app entries
+With the addition of the store, we will have an increase in duplicate entries.  If we have 2 manifests with the same application, we should try to eliminate the duplicate entries by default.  Otherwise it will make it difficult to limit the selection to the one file.
```
yao-msft, line 86: > we should try to eliminate the duplicate entries by default
Seems like we need to prioritize the support for multiple installer in one manifest. And enforce it in manual review/ reject the manifest in index creation?

doc/specs/#117-Windows-Store-Support.md
yao-msft: I see. After reading later parts, default means default InstallerType to use.

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+
+<b>  Example: <appname>  </b>  
+```
+...  
+InstallerType: MSI  
+...  
+InstallerType: MSIX  
+```
+When the user specifies install by default the client will choose to install the version of the app that is flagged as default in the manifest.  If no default is specified, then the app will default to the store as the default app.
```
yao-msft, line 102:
```suggestion
When the user specifies install by default the client will choose to install the InstallerType of the app that is flagged as default in the manifest.  If no default is specified, then the app will default to the store as the default app.
```

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
+---
+author: <kevin> <larkin> <kevinlams/<kevinla@microsoft.com>
+created on: <2020-05-28>
+last updated: <2020-05-28>
+issue id: <github issue id>
+---
+# ===== DRAFT ======
```
denelon, line 7: I think you can use a "DRAFT" PR rather than committing a document with draft in the title. If you merge this, you will have to come back and change it to remove this line.

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Manifest Requirements 
+To support store apps, the manifest file will need to be updated to support the new format.  Most fields remain the same, but additional changes are required to support the store.
+
+### YAML Spec updates
+<b>URL</b> the URL used for the store is not the same as the URL users and the developers expect.  Therefore this field is unnecessary.  
+<b>REQUIRED: NEW: StoreID</b> the Store ID will be used to identify the app. example: 9nblggh4nns1   
+<b>InstallerType</b> A new format, the WindowsStore type will be provided
+<b> InstallerType - Default </b> this needs to be defined, but essentially the author can specify if the user does not select anything... which URL to default to.
+<b>Homepage</b> The Homepage will be the URL with the bigid https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1
```
denelon, line 47: I think we need to explain about the Microsoft catalog ID

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+These apps will fail to activate and therefore should not pass our approval policy.
+
+***In app purchase***
+Apps that require in app purchase, can be allowed, as long as the app can correctly handle activation and treat the in-app purchase out of band from the Windows Package Manager.
+
+## Manifest Requirements 
+To support store apps, the manifest file will need to be updated to support the new format.  Most fields remain the same, but additional changes are required to support the store.
+
+### YAML Spec updates
+<b>URL</b> the URL used for the store is not the same as the URL users and the developers expect.  Therefore this field is unnecessary.
```
denelon, line 43: I think we should retain the URL field and have it point to the store application https:// address

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+### YAML Spec updates
+<b>URL</b> the URL used for the store is not the same as the URL users and the developers expect.  Therefore this field is unnecessary.  
+<b>REQUIRED: NEW: StoreID</b> the Store ID will be used to identify the app. example: 9nblggh4nns1   
+<b>InstallerType</b> A new format, the WindowsStore type will be provided
+<b> InstallerType - Default </b> this needs to be defined, but essentially the author can specify if the user does not select anything... which URL to default to.
+<b>Homepage</b> The Homepage will be the URL with the bigid https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1  
+<B>NEW: TransactionType</b> TransactionType will be used to indicate what type of 
+<b>NEW: Privacy Policy</b>  
+<b>NEW: Terms of transaction</b>
```
denelon, line 50: How is this different from a EULA? I'd like to try and get as much value in the manifest for each attribute. I'd also like to make sure optional attributes or those that only apply in certain situations to be located in a meaningful place.

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+### YAML Spec updates
+<b>URL</b> the URL used for the store is not the same as the URL users and the developers expect.  Therefore this field is unnecessary.  
+<b>REQUIRED: NEW: StoreID</b> the Store ID will be used to identify the app. example: 9nblggh4nns1   
+<b>InstallerType</b> A new format, the WindowsStore type will be provided
+<b> InstallerType - Default </b> this needs to be defined, but essentially the author can specify if the user does not select anything... which URL to default to.
+<b>Homepage</b> The Homepage will be the URL with the bigid https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1  
+<B>NEW: TransactionType</b> TransactionType will be used to indicate what type of 
+<b>NEW: Privacy Policy</b>  
+<b>NEW: Terms of transaction</b>  
+<b>License terms</b>
```
denelon, line 51: Would LicenseURL not work?

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Future considerations
+
+The wizard to ease metadata transfer.
+
+## Resources
+
+[comment]: # Be sure to add links to references, resources, footnotes, etc.
+
+## History
```
denelon, line 142: Git keeps a history of all revisions to a document, this probably isn't necessary

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+<b>REQUIRED: NEW: StoreID</b> the Store ID will be used to identify the app. example: 9nblggh4nns1   
+<b>InstallerType</b> A new format, the WindowsStore type will be provided
+<b> InstallerType - Default </b> this needs to be defined, but essentially the author can specify if the user does not select anything... which URL to default to.
+<b>Homepage</b> The Homepage will be the URL with the bigid https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1  
+<B>NEW: TransactionType</b> TransactionType will be used to indicate what type of 
+<b>NEW: Privacy Policy</b>  
+<b>NEW: Terms of transaction</b>  
+<b>License terms</b>  
+<b>Architecture</b> The architecture will support the architectures supported by the store.  neutral will most likely be the default.
+<b>-prelease</b> In order to support flighing, the client will need to support only showing apps with out the prerelease flag AND enabling the user override on the command line.
```
zachcarp, line 53: flighing -> flighting

doc/specs/#117-Windows-Store-Support.md
zachcarp, line 47: bigid casing is inconsistent throughout the document.

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+***In app purchase***
+Apps that require in app purchase, can be allowed, as long as the app can correctly handle activation and treat the in-app purchase out of band from the Windows Package Manager.
+
+## Manifest Requirements 
+To support store apps, the manifest file will need to be updated to support the new format.  Most fields remain the same, but additional changes are required to support the store.
+
+### YAML Spec updates
+<b>URL</b> the URL used for the store is not the same as the URL users and the developers expect.  Therefore this field is unnecessary.  
+<b>REQUIRED: NEW: StoreID</b> the Store ID will be used to identify the app. example: 9nblggh4nns1
```
zachcarp, line 44: Is there a distinction between StoreId and BigId? If so, can you expand on it?

yao-msft: It's the same, so I'd suggest to use StoreId across the doc to be consistent.

JohnMcPMS:
The issue for installing things comes in two parts:
1. Processing a set of metadata (aka manifest) to result in an installed application
2. Discovery of said metadata

This spec seems to cover installation, but mostly glosses over discovery with the assumption that it would be the same as non-Store (aka through winget-pkgs).  I think that we do need an installation spec much like this one that covers the manifest side of things.  But I believe that discovery should be handled as a separate source, which will allow for the entire catalog to be available (at least, to the limits that we impose).

doc/specs/#117-Windows-Store-Support.md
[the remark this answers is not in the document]
riverar: Nits: Remove "in order", "flighting" is spelled incorrectly.

riverar:
Great start.

Would like to see coverage/consideration of:
* Applications with durables and/or optional packages
* Applications with flight audience groups (can be handled via channel?)
* Applications that are labeled as having IAP but don't really (Store bug with durables/optional packages)
* Dependencies (e.g. framework packages)
* Discovery, especially around localized product display
* Windows SKUs without Store support
  - Can we have two variants (but same version) of our app in winget to alleviate issues that arise due to SKU limitations?

doc/specs/#117-Windows-Store-Support.md
zooba, line 102: > then the app will default to the store as the default app.

Good. This is the best way to handle concerns about flaky remote servers causing download failures.

lordcheeto:
> The issue for installing things comes in two parts:
> 
> 1. Processing a set of metadata (aka manifest) to result in an installed application
> 2. Discovery of said metadata
> 
> This spec seems to cover installation, but mostly glosses over discovery with the assumption that it would be the same as non-Store (aka through winget-pkgs). I think that we do need an installation spec much like this one that covers the manifest side of things. But I believe that discovery should be handled as a separate source, which will allow for the entire catalog to be available (at least, to the limits that we impose).

That would result in more collisions when searching for a program that has been distributed through both a standalone installer and the store. I don't think the behavior for multiple sources has been defined, so I'm not sure what that would look like. 

They wouldn't have the benefit of being linked by their shared presence in a single versioned manifest, and it would be impossible to determine with confidence that two hits are for the same bits, so any collision would have to prompt the user to make a decision. You can't rely on the `Publisher.Name.Version` fully qualified name to resolve collisions, because those fields in the store are not necessarily consistent with the manifests as defined in `winget-pkgs`. For example, Windows Terminal has the Publisher field set to `Microsoft` in `winget-pkgs`, but is `Microsoft Corporation` in the Windows Store.

> Because we have manifests that are based on publisher and appname, there should not be a case where we have multiple manifests with different installers.  If we do, there is something wrong with the metadata.  

With a curated central repository, that's manageable.

---

General question: is support for the [Microsoft Store for Business and Education
](https://docs.microsoft.com/en-us/microsoft-store/) on the radar, and would this spec be compatible without breaking changes later on down the line?

lordcheeto:
> Indeed, we have already received feedback that we should value the curation over the full list of Store offerings. This would also enable direct linkage between things offered in multiple ways, as you state. I just want to ensure the user choice of Store vs external. It may end up being a setting to prefer one or the other.

Agreed. Having it as a separate source, if linked in some way, would allow users to just remove the source. Robust `InstallerType` filtering (configurable with flags or user configurable settings) would allow for that, as well as filtering of other installer fields like language or scope.

doc/specs/#117-Windows-Store-Support.md
KevinLaMS: Are you suggesting it would be required?  I think the URL you are suggesting would typically be the HomePage.

doc/specs/#117-Windows-Store-Support.md
riverar: A EULA requires acceptance, a privacy policy does not. You cannot combine the two.

riverar:
In addition to [earlier comments](https://github.com/microsoft/winget-cli/pull/367#issuecomment-636381714), would like to also throw out there that downloading from the Store should _not_ require initial configuration of credentials or deal with any authentication. I typically click through quite a few dialogs in the Store client to get our supported download-without-login experience. I'd expect winget to just work.

doc/specs/#117-Windows-Store-Support.md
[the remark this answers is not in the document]
KevinLaMS: Agree this needs to be the same as the model we settle on for all packages.

Karl-WE:
@KevinLaMS is this still a thing being worked on?

riverar:
Uh guys. @KevinLaMS @JohnMcPMS @denelon

denelon:
Sorry all. We've had quite a bit of internal discussion going on about store support. The latest release has some code behind a feature toggle with a POC that we've been working on. Unfortunately, we haven't gotten as far along as we had hoped.

There is other activity going on right now to look at how we can build a set of manifests for some store apps that are free, and rated e for everyone. We're thinking about building the manifests for those and putting them into a different source for the package manager.

The challenges associated with commerce and other requirements we have with store Apps have been a bit more involved than originally anticipated. We may have to split #117 into smaller pieces in order to get functionality out, and to make the right subset of store apps available as we can.

Karl-WE:
Good news @denelon my previous mentioned general issues that store app might not update automatically seemed to be fixed in Insider Dev but also outside. Needs more careful observation if is entirely fixed.

Issue #120, opened 2020-05-11
Upgrade an installed App


denelon:
As a user I want to be able to use winget to install the most current version of a currently installed program so I don't have to go find it.

**Experimental Feature**
`winget upgrade` - displays installed packages with an available upgrade.
`winget upgrade -?` - displays help for the upgrade command.
`winget upgrade --all` - upgrades all packages with an available upgrade.
`winget upgrade <package>` - upgrades the specified package.

Note: upgrade depends on list also being enabled.

`winget features` - displays all experimental features and their status
`winget settings` - launches settings.json to configure settings like experimental features

Edited: experimental status

jantari:
I would strongly suggest we stick to the verbs familiar from `apt`:

    winget update #Refreshes the repositories
    winget upgrade #Actually updates apps

One of the very confusing thing about package managers are the different syntaxes they use. I don't 
speak for everyone, but I genuinely find `pacman`, `yum` and `dnf` confusing. `zypper` is ok.

`apt` is popular and - imo - easy to understand.

Rican7:
> I would strongly suggest we stick to the verbs familiar from `apt`:
> 
> ```
> winget update #Refreshes the repositories
> winget upgrade #Actually updates apps
> ```

Yea, agreed. This is what Homebrew (`brew`) uses too, so even more prior-art here.

khobbits:
I think we can probably get away from the requirement to have a command to to 'refresh the repositories'.  It feels like a little backwards in time using apt, compared to yum.  Surely the tool can work out if the repositories require updating automatically?

If we need to operate in a cached/offline mode that should require a parameter rather than be the default.

As to which verb, update/upgrade? Why not allow the use of both?

despotak:
I agree with @khobbits that a "refresh the repos" commands is a backward step. I really hope that `winget` does not need to manually refresh the repositories before an update.
On the other hand, I see @jantari's and @Rican7's PoV. Probably then the solution is somewhere in the middle. Disregard `winget update` and use `winget upgrade` as the default command.

Another solution might be that `update` is an alias for `upgrade` (or via versa).

seanfisher:
> Yea, agreed. This is what Homebrew (brew) uses too, so even more prior-art here.

[...]

PR #397, opened 2020-06-04
Draft of List, Upgrade and Uninstall.

KevinLaMS:
This includes issue 120, 121 and 119.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/397)


doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,144 @@
...
+
+
+**Store Apps**
+The Store has multiple app types.  Those that are free for download and free, but with purchase options, as well as paid apps.
+
+***Free Apps***  
+Apps that are free, will be downloaded directly when requested.
+
+***Pay apps [NOT SUPPORTED]***
+Apps that are for purchase, will not be allowed to be submitted.  If an app from the store requires commerce, to run, we will not allow them in the Windows Package Manager repository.
```
megamorf, line 33: What about paid apps that have been purchased already?

megamorf:
@KevinLaMS Some general feedback.

* Based on the contents on these drafts you haven't worked with Markdown that much so here are some potential issues that can be solved fairly quickly:
  * When using inline or fenced code formatting you don't need to escape characters (e.g. angle brackets) with a backslash `` `winget list [[-q] <query>] [<options>]` `` 
  * Instead of using HTML tags for formatting try to stick with the markdown versions, e.g. `**bold string**` instead of `<b>bold string</b>`
  * All path references in markdown use a regular slash as path separator
  * lists, headings and tables should be surrounded by new lines

See: https://www.markdownguide.org/basic-syntax/

I like where this is going and will add more feedback later. 👍

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+| **--name**   |      Filter results by name. |
+| **--moniker**   |  Filter results by application moniker. |
+| **--tag** |     Filter results by tag
+| **--command**  | Filter results by command
+| **-s,--source** |   Find the application using the specified [source](source.md). |
+| **-e,--exact**     | Find the application using exact match. |
+
+
+*Upgrade app when no update is available*  
+If the user attempts to update an app that there is no known update for, the package manager will issue an error.
```
martinmine, line 140: Should `winget` exit with a non-successful status code if this is the case?

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+
+To obtain this data, the Windows Package Manager will query the 
+**Apps and Features**  installed apps.  This appears to be a superset of 
+**Add Remove Programs** (ARP).
+
+
+### Installed with Windows Package Manager
+If the app is installed with the Windows Package Manager, we have the ID to map the installed version to the repository version.  If the version installed on the PC is lower than the the version in the repository, the list command will show the available version in the "available update."
+
+### Installed out of band from the Windows Package Manaager but in repo
```
fmeyertoens, line 77:
```suggestion
### Installed out of band from the Windows Package Manager but in repo
```

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+If the app is installed out of band from the Package Manager, the app will be discovered if it registers correctly with the **Apps and Features**.  The Windows Package Manager will map the installed app back to the repo.  If the version installed is less than the version available in the repo, the **list** command will show the available version in the "available update."
+
+If the installed version is greater than or equal to the version of the app available in the repo, then the "available update" will be left blank.
+
+### Installed out of band from the Windows Package Manaager but not in repo
```
fmeyertoens, line 82:
```suggestion
### Installed out of band from the Windows Package Manager but not in repo
```

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+| **-e,--exact**     | Find the application using exact match. |
+
+
+*Upgrade app when no update is available*  
+If the user attempts to update an app that there is no known update for, the package manager will issue an error.
+"There is no update available from the community repository.  If you need to reinstall or repair the app, try ```winget repair <appname>```.
+
+
+
+*Depnedencies*
```
fmeyertoens, line 145:
```suggestion
*Dependencies*
```

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+
+
+
+
+
+
+
+[comment]: # Outline the design of the solution. Feel free to include ASCII-art diagrams, etc.
+
+## UI/UX Design
```
martinmine, line 194: A suggestion for what could be discussed here: The impact on users that have not used a package manager, will they naturally use `update` instead of `upgrade` (or sometimes have issues with distinguishing these two)?  And on the other hand, if `update` was an alias for `upgrade`, how can similar functionality of `upgrade` be used with the alias and vice versa? (For example: https://github.com/microsoft/winget-cli/issues/120#issuecomment-631945047) It would be really nice to see what is the plan for dealing with this kind of complexity (`update` vs. `upgrade`) from a user standpoint.

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+The **upgrade** command is designed to update one or more applications.  The **upgrade** command when executed with no parameters or with --all will update all packages that have updates available.
+
+
+## Usage
+
+`winget upgrade [[-q] \<query>] [\<options>]`
+
+![show command]()
+
+**Upgrade** when executed by itself will upgrade all apps ready for updates by the Windows Package Manager.
```
martinmine, line 115: Would the user be prompted to confirm the package update? If yes, should there be flag to override this when being used in automation tasks, similar to how other package managers does this?

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+| Argument  | Description |
+|--------------|-------------|
+| **-q,--query** |  The query used to search for an application. |
+| **-?, --help** |  Gets additional help on this command. |
+
+## Options
+
+The following options are available.
+
+| Option  | Description |
```
martinmine, line 127: What about specifying a manifest file that is located on the user's machine (or another repository)? Would this be part of another spec?

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+For MSI it is the product code.  Typically a GUID that is typically found in the uninstall registry location and includes the brackets.
+For example: ```{5740BD44-B58D-321A-AFC0-6D3D4556DD6C}```
+
+```HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{3740BD44-B58D-321A-AFC0-6D3D4556DD6C}]```
+    
+```HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\{3740BD44-B58D-321A-AFC0-6D3D4556DD6C}]```
+
+For inno, wix, nullsoft, and exe, the SystemAppId should be a string that is located in either of the Uninstall keys above. 
+
+### Upgrade
```
martinmine, line 104: It would be nice to hear how syncing with the remote repository will be handled, will the user have to do this manually (i.e. using an `update`), or will it be done automatically if the repository hasn't been synced after a said time?

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,230 @@
...
+
+See abstract.  :)
+
+## Solution Design
+
+**List**
+**Update \ Upgrade**
+**Uninstall**
+
+### List
```
martinmine, line 25: Handling available updates with this would be similar to how `dotnet` handles packages updates, nice 👍  https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-list-package

doc/specs/#120,121,119-Update-Uninstall-List.md
KevinLaMS: Martin, I am curious.  I **think** that the syncing of the remote repository is handled via the source:
C:\Users\kevinla>winget source update
Updating all sources...
Updating source: winget...
  ██████████████████████████████  100%
Done.

Do you think this addresses the Update concern?

doc/specs/#120,121,119-Update-Uninstall-List.md
JohnMcPMS: I prefer "update'.  Upgrade has other semantics (at least to me), and update fits more cleanly into what is happening.

doc/specs/#120,121,119-Update-Uninstall-List.md
JohnMcPMS: Yes, I believe so.  If the command requested cannot be executed, the process should return a failure exit code.

doc/specs/#120,121,119-Update-Uninstall-List.md
martinmine: It is true you can handle this using the `winget source` command. I just think the spec should outline how the syncing of the remote repo is handled in context of running an update. Is a user required to run an `winget source update` before running `winget upgrade` or will this be implicit from running `winget upgrade`? (i.e. upgrade will invoke source update)

doc/specs/#120,121,119-Update-Uninstall-List.md
martinmine: @JohnMcPMS If you decide to go with `update`, then I would suggest to make `upgrade` alias of `update` to make the life easier for users familiar with other package managers that use `upgrade` for updating packages.

doc/specs/#117-Windows-Store-Support.md
yao-msft: If the entitlement is already granted, I think we can continue with installation.

doc/specs/#120,121,119-Update-Uninstall-List.md
rmarquis: This seems counter intuitive to me, and it's also not what Linux package managers do as far as I know (using `pacman` here). If the purpose of the `upgrade` command is to check if newer version of package(s) are available and get them, then it didn't fail if the last available package(s) are installed already. The output should be different (f.e. "there is nothing to do") but the command would be considered a success anyway.

JohnMcPMS: I guess it comes down to what the purpose is deemed to be semantically.  I am not opposed to the purpose being "Make sure I have the latest version of this", which then I agree with you that no error is correct.  I was reading it more as "Get a newer version of this", which while I agree is a subtle difference, it is still the difference between a command that cannot be executed, versus one that can.

I suppose there is the question of what is done in the --all case.  In that case, should there be an error if no updates are available for anything?  That feels wrong to me, as I don't think anyone would ever say "Get a newer version of *everything*".  So for consistency, I believe that the semantics should be Ensure, not Get.

So I agree that there should not be an error exit code, and the verbage here should be "issue a warning" not an error.

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,229 @@
...
+
+**Open issue** if the application is not side by side, it will force an uninstall of the previous version to upgrade.  we will need to assume the installer wil handle.
+
+### Uninstall
+
+The **uninstall** command is designed to uninstall an application.   
+
+## Usage
+
+```winget uninstall [[-q] <query>] [<options>]```
```
yao-msft, line 156: Does uninstall handle apps installed outside of WinGet?
If there're multiple matches with the query. Should we return a warning message saying "please refine the input ..." like what install command does today?

doc/specs/#120,121,119-Update-Uninstall-List.md
megamorf: Sounds like a good idea to stick to that already established pattern where by default a confirmation is required to proceed that can be skipped by passing an additional parameter.

doc/specs/#120,121,119-Update-Uninstall-List.md
megamorf: In Linux package managers update the package cache for all enabled repositories. If a package was installed from repository "foo" then the search for available updates for that specific application will be narrowed down to repository "foo". If you'd actually want to install a package that is available in multiple repositories without explicitly providing the correct repository name, the package will be installed from the repository with the highest priority.  This priority system is also how the upcoming version 3 of PowerShellGet has been designed to define repository ordering.

lordcheeto: As of now, whenever a source is opened, it updates the source if it's been more than 5 minutes. `winget source update` is just the explicit command to update.

```
// TODO: Enable some amount of user control over this.
constexpr static auto s_DefaultAutoUpdateTime = 5min;
```

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,229 @@
...
+**List** when executed by itself will show all apps installed by the Windows Package Manager.
+
+## Arguments
+
+The following arguments are available.
+
+| Argument  | Description |
+|--------------|-------------|
+| **-q,--query** |  The query used to search for an application. |
+| **-?, --help** |  Gets additional help on this command. |
```
dev-nicolaos, line 46: Not a huge deal, but I'd personally prefer the `-h` alias as it is far more common and more clearly maps to `--help` than `-?`

GauthierPLM: Isn't `-?` the standard on Windows while `-h` is more a Unix thing?

doc/specs/#120,121,119-Update-Uninstall-List.md
RobGThai: Brew user here and every time I want to update an app, I have to look up help to see if I need to `update` or `upgrade`. One is making sure Brew is up to date, the other is for updating the app. If you trigger update without supplying app identifier then Brew will attempt to update every installed packages. Which I'm not a fan of. 

Personally, I'm against making alias between `update` and `upgrade`. Having just one version of them at the start should reduce confusion on which one to use for newcomers. Instead of doing this, I'd suggest WinGet to only offer just one.

- `winget update` - displaying help for update. 
- `winget update <package_id>` - updating provided package.
- `winget update winget` - updating WinGet itself.
- `winget update -a|--all` - updating every installed packages.

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,324 @@
...
+*If we can calculate the community version, to compare against the installed version, that is of course the preference.
+Proposal for those applications we cannot tell if there is an update. Compare hashes. If hashes match, then assume there is no update available. If hashes do not match, then display Get Latest.
+
+For Windows Store, the client will need to query the store to see if an update is available.
+
+##### Freshness of the database
+
+For `list` to display accurate information, the database of content must be up to date. The default experience for a downloaded catalog will be to refresh every 5 minutes. The default experience for RESTFUL apis will be to query the server at `list` execution.
+
+### Update
```
quangkieu, line 129: I think winget should use keywork like apt of Debian.

- Update: for refresh the catalog
- Upgrade: for actual installation operation of new version

This way we could remove the refresh frequency as winget user does not expect winget run in background as a CLI tool usually run and exit

doc/specs/#120,121,119-Update-Uninstall-List.md
quangkieu: I think winget should use keywork like apt of Debian.

- Update: for refresh the catalog
- Upgrade: for actual installation operation of new version

This way we could remove the refresh frequency as winget user does not expect winget run in background as a CLI tool usually run and exit

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Additional Options Considered
+#### Store as a Source
+We evaluated whether we should set the Windows Store as a source, rather than update the manifests.  After evaluation, we felt the metadata provide in the Store Catalog that we would use for Search was lacking data that we would want in order to provide a great user experience.  
+
+## Phased Approach ##
+Supporting the Windows store apps adds a bunch of complexity.  Among other things, developers when submitting their applications to the store agreed to certain terms that did not include the Windows Package Manager.  There we will take a phased approach to enabling store support.
+
+### Phase 1 ###
+Phase 1 will be kick the tires.  The Windows Package Manager will only install apps that meet the following critieria:
```
jsoref, line 30:
```suggestion
Phase 1 will be kick the tires.  The Windows Package Manager will only install apps that meet the following criteria:
```

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,237 @@
...
+last updated: <2020-06-04>
+issue id: <#117>
+---
+# Package Manager Windows Store Support
+
+## Abstract
+This spec identifies requirements and implementation for supporting Windows Store apps in the Windows Package Manager.
+
+## Glossary
+* **StoreId**:  The StoreID is the 12 character number that is used to reference the application in the store.  This is a Windows Store ID, often referred to as BigId when interacting with the store team.  Example: 9nblggh4nns1.
```
jsoref, line 13:
```suggestion
* **StoreId**:  The StoreID is the 12 character number that is used to reference the application in the store.  This is a Windows Store ID, often referred to as **BigId** when interacting with the store team.  Example: `9nblggh4nns1`.
```

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,237 @@
...
+
+
+## Manifest Requirements 
+To support store apps, the manifest file will need to be updated to support the new format.  Most fields remain the same, but additional changes are required to support the store.
+
+### YAML Spec updates
+**URL** the URL used for the store is not the same as the URL users and the developers expect.  Therefore this field is unnecessary.  
+
+**REQUIRED: NEW: StoreID** the Store ID will be used to identify the app.  
+Example: 9nblggh4nns1
```
jsoref, line 71:
```suggestion
Example: `9nblggh4nns1`   
```

doc/specs/#117-Windows-Store-Support.md
```diff
@@ -0,0 +1,237 @@
...
+**REQUIRED: NEW: StoreID** the Store ID will be used to identify the app.  
+Example: 9nblggh4nns1   
+
+**InstallerType:** A new format, the WindowsStore type will be provided  
+
+**InstallerTypeDefault** The InstallerTypeDefault field will allow the manifest author to specify their preferred app format.  For example: if a manifest includes 3 InstallerTypes, MSI, EXE and WindowsStore.  Setting the InstallerTypeDefault to one of the existing types will cause winget to try and install that InstallerType.
+
+> [!NOTE] The users preference can override the manifest settings.  
+
+**Homepage:** The Homepage will be the URL with the StoreId https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1
```
jsoref, line 79: Is this an example?

doc/ManifestSpecv0.2.md
```diff
@@ -0,0 +1,315 @@
...
+
+  - Arch: x86
+    # Url is a required field. This provides the path to the installer.
+    # The Url must begin with https:// and followed by a hostname. 
+    # Restrictions: [min: 10, max:2000]
+    Url: https://contosa.net/publiccontainer/contosainstaller.exe
+
+    # Sha256 is a required field. The value is the hash of the installer and used to verify the executable
+    # Restrictions: [valid sha256 hash]
+    Sha256: "69D84CA8899800A5575CE31798293CD4FEBAB1D734A07C2E51E56A28E0DF8C82"
```
jsoref, line 210: For this, you'll want to add an entry into `patterns.txt`,

probably something like:
```
(?:Signature|)Sha256: "[0-9A-f]{64}"
```

jsoref (changes requested):
@KevinLaMS: can you please pull master and then rebase this (start with a temporary branch -- see below)? The spell checker will be upset if this merges as is.

Most of the items it found just need to be added to `expect.txt`.
Ideally you'd only add words in this manner and you'd use `patterns.txt` for random strings (I've highlighted some random strings here). But if there's no pattern that works usefully (the random BigId that appears in multiple contexts is a good example), then using `expect.txt` is a good choice.

Note: you can test on another branch and only force push to this branch once the spell checker is happy.

doc/specs/#120,121,119-Update-Uninstall-List.md
```diff
@@ -0,0 +1,324 @@
...
+| `--name` | Filter results by name. |
+| `--moniker` | Filter results by application moniker. |
+| `--tag` | Filter results by tag
+| `--command` | Filter results by command
+| `-s,--source` | Find the application using the specified [source](source.md). |
+| `-e,--exact` | Find the application using exact match. |
+
+#### Uninstall Command
+
+Determining the uninstall string will be based on the type of installer. For EXEs and MSIs the command will be shellexec. If the app requires elevation, then the user will be prompted.
```
lechacon, line 232: I assume shellexec means using ShellExecute instead of CreateProcess? Is there a reason for that? ShellExecute requires a file path and arguments so the uninstall string would need to be parsed to split it, whereas CreateProcess can take the command line string directly which seems like a better fit.

PR #400, opened 2020-06-04
Fix build errors if there are spaces in an ancestor folder

MarcusP-P:
If there are spaces in a folder anywhere between the root directory and the project directory, builds will fail when copying files.

This patch puts quotes around the parameters for the copy commands in the project files.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/400)

### DIFF
--- a/src/AppInstallerCLIPackage/AppInstallerCLIPackage.wapproj
+++ b/src/AppInstallerCLIPackage/AppInstallerCLIPackage.wapproj
@@ -143,8 +143,8 @@
   <Import Project="$(WapProjPath)\Microsoft.DesktopBridge.targets" />
   <PropertyGroup Label="Configuration">
     <PostBuildEvent>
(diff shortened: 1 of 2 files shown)


MarcusP-P:
It looks like tests also have issues with spaces in paths. Do you want me to add that to this PR, or file another PR?

JohnMcPMS:
> It looks like tests also have issues with spaces in paths. Do you want me to add that to this PR, or file another PR?

This one please.  Thanks for helping us fix this 😄

MarcusP-P:
No matter what I did, I wasn't able to get the E2E tests working (even when I didn't have a space in the path), so unless I'm missing something obvious, there isn't much to gain from making them work with a space. The unit tests all work with a space in the path.

JohnMcPMS:
Ok, no problem.  Thanks for trying.

Issue #364, opened 2020-05-29
Feature Toggle


denelon:
# Description of the new feature/enhancement

We need a feature toggle mechanism to allow new features to be built and released from the mainline at any time.

megamorf:
The PowerShell team ships experimental features with their preview and stable builds. They are disabled by default and stored in a json file called `powershell.config.json`

```
PS /Users/megamorf> Get-ExperimentalFeature

Name                                Enabled Source                              Description
----                                ------- ------                              -----------
PSCommandNotFoundSuggestion           False PSEngine                            Recommend potential commands based on fuzzy search on a CommandNotFoundException
PSImplicitRemotingBatching            False PSEngine                            Batch implicit remoting proxy commands to improve performance
PSNullConditionalOperators            False PSEngine                            Support the null conditional member access operators in PowerShell language
PSUnixFileStat                        False PSEngine                            Provide unix permission information for files and directories
Microsoft.PowerShell.Utility.PSMan…   False /usr/local/microsoft/powershell/7/… Enables -BreakAll parameter on Debug-Runspace and Debug-Job cmdlets to allow users to decide if they want PowerShell to break immediately in the current location when they attach a debugger.
PSDesiredStateConfiguration.Invoke…   False /usr/local/microsoft/powershell/7/… Enables the Invoke-DscResource cmdlet and related features.
```

You might get some inspiration from their RFC:
https://github.com/PowerShell/PowerShell-RFC/blob/master/5-Final/RFC0029-Support-Experimental-Features.md

denelon:
https://github.com/microsoft/winget-cli/releases/tag/v0.1.42101-preview

PR #414, opened 2020-06-10
Feature Toggle Spec

JohnMcPMS:
Spec for #364 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/414)

### DIFF
--- a/doc/specs/#364 - Feature Toggle.md
+++ b/doc/specs/#364 - Feature Toggle.md
@@ -1,0 +1,73 @@
+---
+author: John McPherson @JohnMcPMS
+created on: 2020-06-04
+last updated: 2020-06-04
+issue id: 364
+---
+
+# Feature Toggle
+
+For [#364](https://github.com/microsoft/winget-cli/issues/364)
+
+## Abstract
+
+As features are implemented within winget, they may cause disruption to users oblivious to the fact that they are in progress. In order to allow work to be done in master, and distributed to early adopters for their feedback, this spec suggests that settings should be used to control "experimental" features. We use this term [as others have](https://github.com/PowerShell/PowerShell-RFC/blob/master/5-Final/RFC0029-Support-Experimental-Features.md) to mean work in progress that is not yet ready for relase to the general audience.
+
+## Inspiration
+
+We realize that we need to be able to:
+1. deliver work in progress to users so that we can receive feedback
+2. work in master without creating multiple, divergent branches
+3. not disrupt users who do not wish to investigate unfinished features
+
+Thank you to @megamorf for the link to the PowerShell Experimental Features RFC.
+
+## Solution Design
+
+A new settings block will be added to the settings file for experimental features to be toggled. By default, all experimental features will be disabled.
+
+A hidden command will be added to the root:
+```
+winget features
+```
+When run, this command will give the status of each feature (enabled/disabled), as well as a link to the Issue or Spec.
+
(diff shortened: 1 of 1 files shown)


doc/specs/#364 - Feature Toggle.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+
+### Accessibility
+
+This should have no direct impact on accessibility.
+
+### Security
+
+There should be no security impact directly, although we must remember that any medium IL process will be able to enable any feature my writing to the settings file. But any experimental feature should be created with the intention of becoming released, and so should have it's own security consideration.
```
denelon, line 58: "feature my writing" should be "feature by writing" I believe.

doc/specs/#364 - Feature Toggle.md
```diff
@@ -0,0 +1,80 @@
...
+
+Feature  Status   Link
+--------------------------------------------------------------------
+Example1 Enabled  https://github.com/microsoft/winget-cli/issues/364
+Example2 Disabled https://github.com/microsoft/winget-cli/issues/396
+```
+When run, this command will give the status of each feature (enabled/disabled), as well as a link to the Issue or Spec.
+
+Internally, a single flag enum will be used to reference experimental features. All command line parsing objects will have a field added for the feature(s) to which they belong. Any internal behavioral changes will be made based on a check using this enum. In this way, transitioning a feature either to release or remove can be done by finding by a single identifier.
+
```
denelon, line 43: We may want to consider a "global" revert back to stable like `winget features --stable` or `winget features --enable` and the complimentary `winget features --disable`. I could go either way with this.

doc/specs/#364 - Feature Toggle.md
```diff
@@ -0,0 +1,80 @@
...
+### Performance, Power, and Efficiency
+
+## Potential Issues
+
+No known issues.
+
+## Future considerations
+
+This feature enables all larger features in the future to have a phased rollout.
+
```
denelon, line 77: What are your thoughts on removing feature toggles? (At GA for the feature, or after a while, or never)

doc/specs/#364 - Feature Toggle.md
msftrubengu: What does stable means? It seems to me that stable means no experimental features, which means disable. 

What does enable means? Is it turn all experimental features?

riverar: There's also `stable` ambiguity when you start rolling out A/B testing.

doc/specs/#364 - Feature Toggle.md
riverar: If you're shooting for velocity-like behavior (as in Windows feature control), I'd say they're a dev-only feature and you own the lifecycle. Anything meant for a user to control should be introduced/managed via settings, not these toggles.

JohnMcPMS: Agree with @riverar.  New settings can be introduced along with the feature as it is developed.  They should have no impact when it is disabled.

A feature goes GA by removing the feature enum value in code, and all related references.  The value in settings will become ignored, and we can keep a list in a comment to prevent future reuse of settings names.

doctordns:
Feature flags or toggles are a great idea. Why not look at how Powershell does this and leverage their code. So Get-WgExperimentalFeature, Set-WgExperimentalFeature?

JohnMcPMS:
> Feature flags or toggles are a great idea. Why not look at how Powershell does this and leverage their code. So Get-WgExperimentalFeature, Set-WgExperimentalFeature?

I think that is a good thing to include in future PowerShell work, but we also want to support non-PowerShell paths.

Issue #396, opened 2020-06-03
Settings command


msftrubengu:
# Description of the new feature/enhancement

Add a settings file and associated management command(s).

Prerequisite for #364

PR #427, opened 2020-06-15
Enable differentiation of settings streams

JohnMcPMS:
## Motivation
As groundwork for #396 , we need a way to differentiate settings streams.  This change adds an enum that allows us to do just that.

## Change
The GetPathTo* functions were all collapsed into a single one that uses and enum to select the desired path.  This enables new paths to be added more easily, which should lead to each one being semantically meaningful, rather than potentially muddied.

The settings code was moved to its own files, and a new enum added to enable differentiation of semantics.  The existing paths are called 'Standard', while the new path is called 'UserFile'.  These new streams are intended to be user editable files, and so always use files instead of containers.

The settings code was also coalesced under an interface internally, making the call site much cleaner, and the code properly divided.  This also allows for a more dynamic choice in where settings are stored, as the one function directs all settings requests based on package/not and type.

## Validation
Existing regression tests and a new one targeting the UserFile type.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/427)

### DIFF
--- a/src/AppInstallerCommonCore/Runtime.cpp
+++ b/src/AppInstallerCommonCore/Runtime.cpp
@@ -73,41 +73,6 @@ namespace AppInstaller::Runtime
         static std::filesystem::path s_Settings_TestHook_ForcedContainerPrepend;
 #endif
 
-        void ValidateSettingNamePath(std::filesystem::path& name)
-        {
-            THROW_HR_IF(E_INVALIDARG, !name.has_relative_path());
-            THROW_HR_IF(E_INVALIDARG, name.has_root_path());
-            THROW_HR_IF(E_INVALIDARG, !name.has_filename());
-        }
-
-        // Gets the container within LocalSettings for the given path.
-        auto GetLocalSettingsContainerForPath(const std::filesystem::path& name)
-        {
-            auto result = winrt::Windows::Storage::ApplicationData::Current().LocalSettings();
-
-            std::filesystem::path pathToUse;
-
-#ifndef AICLI_DISABLE_TEST_HOOKS
-            if (!s_Settings_TestHook_ForcedContainerPrepend.empty())
-            {
-                pathToUse = s_Settings_TestHook_ForcedContainerPrepend;
-                pathToUse /= name;
-            }
-            else
-#endif
-            {
-                pathToUse = name;
-            }
-
-            for (const auto& part : pathToUse.parent_path())
-            {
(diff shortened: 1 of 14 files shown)


src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -262,135 +201,86 @@ namespace AppInstaller::Runtime
...
-            return result;
+                break;
+            case PathName::DefaultLogLocation:
+                // To enable UIF collection through Feedback hub, we must put our logs here.
+                result.assign(appStorage.LocalFolder().Path().c_str());
+                result /= WINGET_DEFAULT_LOG_DIRECTORY;
+                break;
+            case PathName::StandardSettings:
+#ifndef AICLI_DISABLE_TEST_HOOKS
+                result = s_Settings_TestHook_ForcedContainerPrepend;
```
yao-msft, line 236: >result = s_Settings_TestHook_ForcedContainerPrepend; [](start = 16, length = 52)

Do we need the empty check if (!s_Settings_TestHook_ForcedContainerPrepend.empty()) here as above?

src/AppInstallerCommonCore/Settings.cpp
```diff
@@ -0,0 +1,193 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "Public/winget/Settings.h"
+#include "Public/AppInstallerRuntime.h"
+#include "Public/AppInstallerStrings.h"
+
+namespace AppInstaller::Settings
```
yao-msft, line 8: >AppInstaller [](start = 10, length = 12)

Just to confirm that we're still using AppInstaller as root namespace for new code and we'll rename once for all in the future?

src/AppInstallerCommonCore/Runtime.cpp
JohnMcPMS: No.  We probably don't need the other one, but I feel safer assigning an empty one than appending.

src/AppInstallerCommonCore/Settings.cpp
JohnMcPMS: Yes, the namespace will litter the code with both if we do it piecemeal.  Placing new header files in a directory won't.

PR #436, opened 2020-06-17
Settings command

msftrubengu:
This changes introduce the settings command and two settings that can be configured.

When the setting command is executed the first time a settings file with empty values will be created. It will also open in the default text editor that is configured for json extensions. When there is no file type association for json files, older versions of Windows will fail to open. In this scenario notepad will be used as the text editor. Newer version of windows will show the "choose app" dialog where the user can select the text editor of their preference.

**progressBar**
This settings allows to modify the default color of the progress bar. Currently it supports three values, but it can be expanded to include more colors.

**autoUpdateIntervalInMinutes**
This is the value in minutes in which the source will auto update. The default value is 5 minutes. A user can set 0 to disable auto updates. 

See [this spec](https://github.com/microsoft/winget-cli/blob/master/doc/specs/%23396%20-%20Settings%20command.md) for more information.

Validated manually and created several unittest for new functionality.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/436)


doc/Settings.md
```diff
@@ -0,0 +1,46 @@
...
+
+```
+    "source": {
+        "autoUpdateIntervalInMinutes": 3
+    },
+``` 
+
+### autoUpdateIntervalInMinutes
+
+Positive integer that represents the interval in minutes on how often to update the WinGet source automatically. For no updates use 0.
```
JohnMcPMS, line 23: >Positive integer that represents the interval in minutes on how often to update the WinGet source automatically [](start = 0, length = 111)

Positive integer that represents the interval in minutes of how often to automatically check for updates to a WinGet source. The check for updates only happens when a source is used, and if no update is available the interval will be reset.

src/AppInstallerCLICore/Core.cpp
```diff
@@ -78,6 +79,13 @@ namespace AppInstaller::CLI
         // The root command is our fallback in the event of very bad or very little input
         std::unique_ptr<Command> command = std::make_unique<RootCommand>();
 
+        // Load the settings file and warn if necessary.
+        auto warnings = Settings::UserSettings::Instance().GetWarnings();
```
JohnMcPMS, line 83: >Settings::UserSettings::Instance() [](start = 24, length = 34)

I would suggest a standalone function that returns this, such as:

Settings::User()

Or similar.  I really hate cluttered call sites.

JohnMcPMS: Also, this could just be directly in the for init.

---
In reply to: [442408327](https://github.com/Microsoft/pkg/pull/436#discussion_r442408327) [](ancestors = 442408327)

src/AppInstallerCLICore/ExecutionArgs.h
```diff
@@ -54,7 +54,7 @@ namespace AppInstaller::CLI::Execution
             Force,      // Generic flag to enable a command to skip some check
             ListVersions, // Used in Show command to list all available versions of an app
             NoVT, // Disable VirtualTerminal outputs
-            PlainStyle, // Makes progress display as plain
+            RetroStyle, // Makes progress display as plain
```
JohnMcPMS, line 57: >plain [](start = 53, length = 5)

Your vendetta against plain is not complete.

src/AppInstallerCLICore/ExecutionProgress.h
```diff
@@ -13,17 +14,9 @@
...
 {
-    // The visual style of the progress bar.
-    enum class VisualStyle
-    {
-        NoVT,
-        Plain,
-        Accent,
-        Rainbow,
-    };
+    using namespace Settings;
```
JohnMcPMS, line 19: >using namespace Settings; [](start = 4, length = 25)

Not in headers, unless it is a string literals namespace.

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    Resource::LocString SettingsCommand::LongDescription() const
+    {
+        return { Resource::String::SettingsCommandLongDescription };
+    }
+
+    std::string SettingsCommand::HelpLink() const
+    {
+        return "https://aka.ms/winget-settings";
```
JohnMcPMS, line 33: >https://aka.ms/winget-settings [](start = 16, length = 30)

Did someone create this?

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -0,0 +1,61 @@
...
+        if (userSettingType == UserSettingsType::Default)
+        {
+            // Create settings file if it doesn't exist.
+            UserSettings::Instance().CreateFileIfNeeded();
+        }
+        else if (userSettingType == UserSettingsType::Standard)
+        {
+            // Settings file was loaded correctly, create backup.
+            UserSettings::Instance().CreateBackup();
+        }
```
JohnMcPMS, line 49: This whole block feels like it should be in:

UserSettings::PrepareToShellExecuteFile()

or someting. I don't think the command should be the one responsible for understanding this process.

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+        std::string filePathUTF8Str = UserSettings::Instance().SettingsFilePath().u8string();
+
+        // Kudos to the terminal team for this work around.
```
JohnMcPMS, line 53: >// Kudos to the terminal team for this work around. [](start = 8, length = 51)

Does one have to do something special to invoke the "choose which app to use" dialog?  I expected that would just be automatic.

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -268,4 +268,9 @@ namespace AppInstaller::Utility
 
         return result;
     }
+
+    std::string ToString(const std::string_view key)
```
JohnMcPMS, line 272: >ToString [](start = 16, length = 8)

this is 2 chars less than std::string, but a bit more confusing as one doesn't know exactly what is going to happen without reading it.

Please remove and just use std::string()

src/AppInstallerCommonCore/Public/winget/JsonUtil.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include <json.h>
+
+#include <optional>
+#include <string>
+
+namespace AppInstaller::Utility
```
JohnMcPMS, line 9: Why does this file need to be public?  Probably fine for it to be for potential future use, but it shouldn't be need right now.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "AppInstallerStrings.h"
+#include "winget/settings/Source.h"
+#include "winget/settings/Visual.h"
+#include <json.h>
```
JohnMcPMS, line 7: A consumer of this header should not need to include JSON.  That should all be implementation in the cpp file.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+        // Loaded settings.json.backup
+        Backup,
+    };
+
+    // Representation of the parsed settings file.
+    struct UserSettings
+    {
+        static UserSettings& Instance()
+        {
+            // If we go multi-threaded secure this.
```
JohnMcPMS, line 31: >// If we go multi-threaded secure this. [](start = 12, length = 39)

It already is secured by what STL (the person) likes to call magic statics.  Basically auto-thread safe init of function local statics.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+        std::filesystem::path SettingsFilePath();
+        std::filesystem::path SettingsBackupFilePath();
+
+        // Settings
+        inline const Source& GetSource() const { return *m_source; }
+        inline const Visual& GetVisual() const { return *m_visual; }
+
+    private:
+        UserSettings();
+        ~UserSettings() {};
```
JohnMcPMS, line 55: > {}; [](start = 23, length = 4)

= default;

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+    {
+        static UserSettings& Instance()
+        {
+            // If we go multi-threaded secure this.
+            static UserSettings userSettings;
+            return userSettings;
+        }
+
+        UserSettings(const UserSettings&) = delete;
+        UserSettings& operator=(const UserSettings&) = delete;
```
JohnMcPMS, line 37: Also delete the moves, the only one should be Instance.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+        // Loaded settings.json
+        Standard,
+        // Loaded settings.json.backup
+        Backup,
+    };
+
+    // Representation of the parsed settings file.
+    struct UserSettings
+    {
+        static UserSettings& Instance()
```
JohnMcPMS, line 29: >UserSettings& [](start = 15, length = 13)

const&?  We don't really want anyone changing anything, do we?

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+            // If we go multi-threaded secure this.
+            static UserSettings userSettings;
+            return userSettings;
+        }
+
+        UserSettings(const UserSettings&) = delete;
+        UserSettings& operator=(const UserSettings&) = delete;
+
+        UserSettingsType GetType() const { return m_type; }
+        std::vector<std::string> GetWarnings() const { return m_warnings; }
```
JohnMcPMS, line 40: >std::vector<std::string> [](start = 8, length = 24)

const&

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+
+        // Settings
+        inline const Source& GetSource() const { return *m_source; }
+        inline const Visual& GetVisual() const { return *m_visual; }
+
+    private:
+        UserSettings();
+        ~UserSettings() {};
+
+        std::optional<Json::Value> ParseFile(const std::string_view& fileName);
```
JohnMcPMS, line 57: >std::optional<Json::Value> ParseFile(const std::string_view& fileName); [](start = 8, length = 71)

Move to unnamed namespace in cpp.
Use std::filesystem::path.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+            return userSettings;
+        }
+
+        UserSettings(const UserSettings&) = delete;
+        UserSettings& operator=(const UserSettings&) = delete;
+
+        UserSettingsType GetType() const { return m_type; }
+        std::vector<std::string> GetWarnings() const { return m_warnings; }
+
+        void Reload();
```
JohnMcPMS, line 42: >Reload [](start = 13, length = 6)

When would this be used?

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+        void CreateFileIfNeeded();
+        void CreateBackup();
```
JohnMcPMS, line 44: These feel like they should be static.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+
+        std::filesystem::path SettingsFilePath();
+        std::filesystem::path SettingsBackupFilePath();
```
JohnMcPMS, line 47: Also static.

src/AppInstallerCommonCore/UserSettings/UserSettings.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+    using namespace std::string_view_literals;
+    using namespace Runtime;
+    using namespace Utility;
+
+    static constexpr std::string_view s_SettingFileName = "settings.json"sv;
+    static constexpr std::string_view s_SettingBackupFileName = "settings.json.backup"sv;
+
+    static constexpr std::string_view s_SettingEmpty =
+        R"""({
```
JohnMcPMS, line 24: >"" [](start = 10, length = 2)

" is an interesting choice for a raw string delimiter.

src/AppInstallerCommonCore/UserSettings/UserSettings.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "pch.h"
+#include <AppInstallerRuntime.h>
+#include "SettingValidation.h"
+#include "winget/Settings.h"
+#include "winget/UserSettings.h"
+
+#include "winget/settings/Source.h"
+#include "winget/settings/Visual.h"
+
+#include <iostream>
```
JohnMcPMS, line 12: If not in the pch, add it there.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,66 @@
...
+
+        // Settings
+        inline const Source& GetSource() const { return *m_source; }
+        inline const Visual& GetVisual() const { return *m_visual; }
```
JohnMcPMS, line 51: I'm not a fan of having to create classes and functions for every setting.  I would prefer a data driven model, thus requiring probably two files to be touched.  Will look at the code for parsing and see if that still seems like a reasonable position.

src/AppInstallerCommonCore/UserSettings/UserSettings.cpp
```diff
@@ -0,0 +1,132 @@
...
+        auto stream = GetSettingStream(Type::UserFile, fileName);
+        if (stream)
+        {
+            Json::Value root;
+            Json::CharReaderBuilder builder;
+            const std::unique_ptr<Json::CharReader> reader(builder.newCharReader());
+
+            std::stringstream settingsContent;
+            settingsContent << stream->rdbuf();
+            std::string settingsContentStr = settingsContent.str();
```
JohnMcPMS, line 91: There is a utility function to read the entire stream already (ReadEntireStream).  I don't know if its implementation is better than this, but you should just use that function and replace the body if warranted.

src/AppInstallerCommonCore/UserSettings/UserSettings.cpp
```diff
@@ -0,0 +1,132 @@
...
+            SetSetting(Type::UserFile, s_SettingFileName, s_SettingEmpty);
+        }
+    }
+
+    void UserSettings::CreateBackup()
+    {
+        THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), !std::filesystem::exists(SettingsFilePath()));
+
+        auto from = SettingsFilePath();
+        auto to = GetPathTo(PathName::UserFileSettings) / s_SettingBackupFileName;
```
JohnMcPMS, line 129: >GetPathTo(PathName::UserFileSettings) / s_SettingBackupFileName [](start = 18, length = 63)

SettingsBackupFilePath?

src/AppInstallerCommonCore/UserSettings/SettingValidation.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace AppInstaller::Settings
+{
+    namespace SettingsWarnings
+    {
+        const char* const Field = " Field: ";
+        const char* const Value = " Value: ";
+        const char* const InvalidFieldValue = "Invalid field value.";
+        const char* const InvalidFieldFormat = "Invalid field format.";
+        const char* const LoadedBackupSettings = "Loaded settings from backup file.";
```
JohnMcPMS, line 16: constexpr std::string_view

src/AppInstallerCommonCore/Public/winget/UserSettings.h
JohnMcPMS: So you kind of already do this, just not to the level that would make it easier to add new things.

There are many ways that it could be done, but for type safety I think it would need to be similar to the Execution::Context.

enum for settings
templated struct with specializations per enum
templated get and parse functions

To add a new setting:

add new enum value
add new specialization
add parse call in main Parse method

Templated struct contains:

json "path"
json data type
internal data type
default value
parse function to turn json type into internal type

Parsed data gets stored in a map similar to Execution::Context.  Get<enum>() looks there first, or returns struct<enum>::default if not present.

---
In reply to: [442441037](https://github.com/Microsoft/pkg/pull/436#discussion_r442441037) [](ancestors = 442441037)

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
msftrubengu: I did, is not pointing to docs right now, but will update it once settings.md is on master

---
In reply to: [442413601](https://github.com/microsoft/winget-cli/pull/436#discussion_r442413601) [](ancestors = 442413601)

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
msftrubengu: It really depends on the version of Windows. Previous version will fail and that's where notepad will be open. Newer versions will show the dialog.

---
In reply to: [442416924](https://github.com/microsoft/winget-cli/pull/436#discussion_r442416924) [](ancestors = 442416924)

src/AppInstallerCommonCore/Public/winget/UserSettings.h
msftrubengu: not needed anymore

---
In reply to: [442426732](https://github.com/microsoft/winget-cli/pull/436#discussion_r442426732) [](ancestors = 442426732)

src/AppInstallerCLICore/Core.cpp
```diff
@@ -78,6 +79,13 @@ namespace AppInstaller::CLI
...
+        for (const auto& warning : warnings)
+        {
+            context.Reporter.Warn() << warning << std::endl;
```
msftrubengu, line 85: >context.Reporter.Warn() << warning << std::endl; [](start = 12, length = 48)

Print only a warning message and only print the syntax error when winget settings

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    std::string SettingsCommand::HelpLink() const
+    {
+        return "https://aka.ms/winget-settings";
+    }
+
+    void SettingsCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        // Show warnings only when the setting command is executed.
```
JohnMcPMS, line 38: >// Show warnings only when the setting command is executed. [](start = 8, length = 59)

You might want to have some text beforehand that will set up the warnings that are about to come out, rather than just showing them.

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        User().PrepareToShellExecuteFile();
+
+        std::string filePathUTF8Str = UserSettings::SettingsFilePath().u8string();
+
+        // Some versions of windows will fail if no file extension association exists, other will pop up the dialog
+        // to make the user pick their default.
+        // Kudos to the terminal team for this work around.
+        HINSTANCE res = ShellExecuteA(nullptr, nullptr, filePathUTF8Str.c_str(), nullptr, nullptr, SW_SHOW);
```
JohnMcPMS, line 51: >ShellExecuteA [](start = 24, length = 13)

I don't think we have done anything to set our process to UTF-8 defaults, enabling UTF-8 *A apis.  So this should probably be use the W function.  Or potentially figure out how we do set ourself as UTF-8 aware (although I suspect that older OS versions would still need to use W).

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -299,12 +296,24 @@
...
     <value>Find and show basic info of apps</value>
   </data>
+  <data name="SettingsCommandLongDescription" xml:space="preserve">
+    <value>Open settings in the default json text editor. If no editor is configured, opens settings in notepad. For available settings see https://aka.ms/winget-settings</value>
+  </data>
+  <data name="SettingsCommandShortDescription" xml:space="preserve">
+    <value>Open settings</value>
+  </data>
+  <data name="SettingsWarnings" xml:space="preserve">
+    <value>Unexpected error while loading settings. Please verify your settings by running settings command </value>
```
JohnMcPMS, line 315: >by running settings command  [](start = 80, length = 28)

by running the settings command.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,143 @@
...
+        };
+
+        template <>
+        struct SettingMapping<Setting::ProgressBarVisualStyle>
+        {
+            using json_t = std::string;
+            using value_t = VisualStyle;
+
+            static constexpr value_t DefaultValue = VisualStyle::Accent;
+            static constexpr std::string_view Path = ".visual.progressBar";
```
JohnMcPMS, line 63: >static constexpr std::string_view Path = ".visual.progressBar" [](start = 12, length = 62)

Do you actually get constexpr from this constructor?  It claims to, but I suppose I'm wary of that.  That's why I always use the string literal tag (sv).  But maybe it isn't necessary.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,143 @@
...
+
+            static std::optional<value_t> Validate(json_t value);
```
JohnMcPMS, line 65: >json_t [](start = 51, length = 6)

const json_t& as a general rule.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -0,0 +1,143 @@
...
+        };
+
+        template <>
+        struct SettingMapping<Setting::AutoUpdateTimeInMinutes>
+        {
+            using json_t = uint32_t;
+            using value_t = uint32_t;
```
JohnMcPMS, line 72: >uint32_t [](start = 28, length = 8)

std::chrono::minutes?

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -0,0 +1,229 @@
...
+                settingsRoot = settingsBackupJson.value();
+            }
+        }
+
+        if (!settingsRoot.isNull())
+        {
+            Validate<
+                Setting::AutoUpdateTimeInMinutes,
+                Setting::ProgressBarVisualStyle
+                >(settingsRoot, m_settings, m_warnings);
```
JohnMcPMS, line 200: Should probably just make Validate take a single template param and have multiple lines.

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
msftrubengu: Changed. We are using ShellExecuteExA in ShellExecuteInstallerHandler.cpp, is that expected?

---
In reply to: [443074390](https://github.com/microsoft/winget-cli/pull/436#discussion_r443074390) [](ancestors = 443074390)

src/AppInstallerCommonCore/Public/winget/UserSettings.h
msftrubengu: I guess. Added just in case

---
In reply to: [443076363](https://github.com/microsoft/winget-cli/pull/436#discussion_r443076363) [](ancestors = 443076363)

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -0,0 +1,64 @@
...
+            for (const auto& warning : User().GetWarnings())
+            {
+                context.Reporter.Warn() << warning << std::endl;
+            }
+        }
+
+        User().PrepareToShellExecuteFile();
+
+        
+        auto filePathUTF16 = ConvertToUTF16(UserSettings::SettingsFilePath().u8string());
```
JohnMcPMS, line 52: >ConvertToUTF16(UserSettings::SettingsFilePath().u8string()); [](start = 29, length = 60)

UserSettings::SettingsFilePath().c_str() [I think... the point is there is a method that just gives back the wide version]

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
JohnMcPMS: That has its own issue, and yes, it should be changed (but not in the PR).

---
In reply to: [443093477](https://github.com/Microsoft/pkg/pull/436#discussion_r443093477) [](ancestors = 443093477,443074390)

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -0,0 +1,235 @@
...
+        template <size_t... S>
+        void ValidateAll(
+            Json::Value& root,
+            std::map<Setting, details::SettingVariant>& settings,
+            std::vector<std::string>& warnings,
+            std::index_sequence<S...>)
+        {
+            // Use folding to call each setting validate function.
+            // Do not change this expression without understanding the implications to the bind order.
+            // See: https://en.cppreference.com/w/cpp/language/fold for more details.
```
JohnMcPMS, line 138: These lines are not relevant; order should not matter here and it certainly doesn't have anything to do with binding (that is SQLite related, no C++).

PR #455, opened 2020-06-23
Secure settings and sources

### LINKED ISSUE #454: Source management should require administrator privileges

To prevent changes to sources from swapping out the target package to install, the source management functionality should require administrator privileges.  This will prevent an unelevated attacker from tricking the user into elevating an installer.

JohnMcPMS:
Resolves #454

## Motivation
To prevent an unelevated attacker from changing sources, and thus potentially tricking the user into elevating an installer from an untrusted location, the source configuration should require administrator privileges to change.

## Change
This change adds a secure setting type, and moves the source configuration to that type.  This also has the effect of leaving the old source configuration behind.  As multiple sources are not yet well supported, this is deemed an acceptable data loss.

The secure settings are secured by storing a hash of the stream on write.  This hash is stored in a location that requires administrator privileges to write, while the primary setting stream is stored in the app data.  If the app is uninstalled then reinstalled, or reset from settings, this stream will be lost and the sources reset.  The secure data will be left behind, but not referenced when no stream is present.  If a stream is present, its hash must also be present or an exception is thrown.  For sources, an attempt to tamper with the configuration would result in an unusable winget, until 'winget source reset --force' was executed, fully resetting the source data.

To support not writing to the source configuration, the sources are now layered.  This means that all user sources are read, then any default sources are added to the list.  To allow a default source to be removed, a tombstone concept is added that will allow the 'higher' layers to override the lower ones.

The update metadata is also separated out from the identity data, allowing it to be written without elevation.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/455)


msftrubengu:
>     }

You can get rid of this and use GetPathTo for the backup

---
Refers to: src/AppInstallerCLITests/UserSettings.cpp:30 in ed8b83f. [](commit_id = ed8b83f3047db8ba95af5be9db874dc85b99389d, deletion_comment = False)

yao-msft:
>         const char* const GetMessageForAppInstallerHR(HRESULT hr)

We may need to localize these in the future.

---
Refers to: src/AppInstallerCommonCore/Errors.cpp:13 in ed8b83f. [](commit_id = ed8b83f3047db8ba95af5be9db874dc85b99389d, deletion_comment = False)

JohnMcPMS:
>         const char* const GetMessageForAppInstallerHR(HRESULT hr)

Agreed, I filed an issue for the fact that localization is not 100% complete.  These are only shown if an exception is caught in main with these HRs, so for this change specifically, it has no impact.

---
In reply to: [648466612](https://github.com/Microsoft/pkg/pull/455#issuecomment-648466612) [](ancestors = 648466612)

---
Refers to: src/AppInstallerCommonCore/Errors.cpp:13 in ed8b83f. [](commit_id = ed8b83f3047db8ba95af5be9db874dc85b99389d, deletion_comment = False)

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -123,27 +129,145 @@ namespace AppInstaller::Repository
...
+                    Settings::Streams::UserSources,
+                    s_SourcesYaml_Sources,
+                    [&](SourceDetailsInternal& details, const std::string& settingValue, const YAML::Node& source)
+                    {
+                        std::string_view name = Settings::Streams::UserSources.Path;
+                        if (!TryReadScalar(name, settingValue, source, s_SourcesYaml_Source_Name, details.Name)) { return false; }
+                        if (!TryReadScalar(name, settingValue, source, s_SourcesYaml_Source_Type, details.Type)) { return false; }
+                        if (!TryReadScalar(name, settingValue, source, s_SourcesYaml_Source_Arg, details.Arg)) { return false; }
+                        if (!TryReadScalar(name, settingValue, source, s_SourcesYaml_Source_Data, details.Data)) { return false; }
+                        int32_t isTombstoneNumber{};
```
msftrubengu, line 205: >int32_t isTombstoneNumber{}; [](start = 24, length = 28)

why is it not a bool in yaml? is it easier this way?

JohnMcPMS: Because I didn't realize that there was a bool type in YAML...

JohnMcPMS: Changed it, and it is far easier to just use bool.

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -216,8 +208,17 @@ namespace AppInstaller::Repository::Microsoft
 
             void RemoveInternal(const SourceDetails& details, IProgressCallback& callback) override
             {
-                AICLI_LOG(Repo, Info, << "Removing package: " << GetPackageFullNameFromDetails(details));
-                Deployment::RemovePackage(GetPackageFullNameFromDetails(details), callback);
+                auto fullName = Msix::GetPackageFullNameFromFamilyName(GetPackageFamilyNameFromDetails(details));
+
+                if (!fullName)
+                {
+                    AICLI_LOG(Repo, Info, << "No full name found for family name: " << GetPackageFamilyNameFromDetails(details));
```
yao-msft, line 215: >No full name found for family name [](start = 46, length = 34)

Could it be the case that multiple found?

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -250,21 +243,38 @@ namespace AppInstaller::Runtime
                 wchar_t tempPath[MAX_PATH + 1];
                 DWORD tempChars = GetTempPathW(ARRAYSIZE(tempPath), tempPath);
                 result.assign(std::wstring_view{ tempPath, static_cast<size_t>(tempChars) });
+
+                result /= "WinGet";
```
msftrubengu, line 247: >"WinGet" [](start = 26, length = 8)

AICLI_DEFAULT_TEMP_DIRECTORY?

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
JohnMcPMS: Yes, it is possible, but that case is supposed to be a bug.  

---
In reply to: [444598040](https://github.com/Microsoft/pkg/pull/455#discussion_r444598040) [](ancestors = 444598040)

PR #460, opened 2020-06-25
Feature toggle

msftrubengu:
# Change

Implementation for #364 

Experimental features enable allow features to be distributed to early adopters for their feedback without the consequences of affecting already released functionalities.

## Developers
The first in the workflow of a developer when implementing a new feature is to add a new Feature in ExperimentalFeature.h and follow the instruction in the comment. For commands or arguments, the disable protection is already in place. If the feature requires additional checks use `ExperimentalFeatures::isEnabled(<your feature>)`. 

When a feature is ready to be release, the enum value must be removed and all the reference to it. By always using the feature enum value we guarantee a cleaner removal.

To provide guidance, winget includes two test experimental features. 

## Users
To enable an experimental feature, go to the settings file and enable it under the experimentalFeatures property. Settings.md must have information about all the existing experimental feature (there are currently none).

To see all the available features, a user can execute the hidden `winget features` and see the status, as well as a link to the spec for the feature.

# Validation
Implement new unittest and test locally the experimental test features

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/460)


src/AppInstallerCLICore/Argument.h
```diff
@@ -31,7 +32,7 @@ namespace AppInstaller::CLI
     };
 
     // Controls the visibility of the field.
-    enum class Visibility
+    enum class VisibilityArg
```
JohnMcPMS, line 35: >VisibilityArg [](start = 15, length = 13)

I would suggest moving it under Argument rather than just smooshing names together.

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -91,4 +94,14 @@ namespace AppInstaller::CLI
         args.push_back(ForType(Args::Type::RetroStyle));
         args.push_back(ForType(Args::Type::VerboseLogs));
     }
+
+    void Argument::OutputEnableMessage(Execution::Context& context) const
```
JohnMcPMS, line 98: >OutputEnableMessage [](start = 19, length = 19)

This should not be here, at least as a direct output function.  If I'm guessing at its use, then it should generated a string to be part of the CommandException generated by parsing.

src/AppInstallerCLICore/Command.cpp
```diff
@@ -162,8 +164,11 @@ namespace AppInstaller::CLI
 
             for (const auto& command : commands)
             {
-                size_t fillChars = (maxCommandNameLength - command->Name().length()) + 2;
-                infoOut << "  "_liv << Execution::HelpCommandEmphasis << command->Name() << Utility::LocIndString{ std::string(fillChars, ' ') } << command->ShortDescription() << std::endl;
+                if ((command->Visibility() == VisibilityCmd::Show) && User().isEnabled(command->Feature()))
```
JohnMcPMS, line 167: >if ((command->Visibility() == VisibilityCmd::Show) && User().isEnabled(command->Feature())) [](start = 16, length = 91)

Might be better to encode the feature check into Command::Visibility

src/AppInstallerCLICore/Command.cpp
```diff
@@ -199,16 +204,21 @@ namespace AppInstaller::CLI
 
             if (hasArguments)
             {
-                infoOut << Resource::String::AvailableArguments << std::endl;
-
+                bool atLeastOne = false;
```
JohnMcPMS, line 207: >bool atLeastOne = false; [](start = 16, length = 24)

Shouldn't be needed if the above code takes Visibility into account properly (as it does not right now).

src/AppInstallerCLICore/Command.cpp
```diff
@@ -489,4 +512,14 @@ namespace AppInstaller::CLI
...
         THROW_HR(E_NOTIMPL);
     }
+
+    void Command::OutputEnableMessage(Execution::Context& context) const
+    {
+        THROW_HR_IF(E_UNEXPECTED, m_feature == ExperimentalFeature::None);
+
+        auto featureInfo = UserSettings::GetFeatureInfo(m_feature);
+        context.Reporter.Warn() << Resource::String::CommandDisabled << " " << m_name << std::endl;
+        context.Reporter.Warn() << Resource::String::ExperimentalFeatureName << " " << std::get<0>(featureInfo) << std::endl;
```
JohnMcPMS, line 522: >context.Reporter.Warn() [](start = 8, length = 23)

No need to get a new output object, just use:

Warn() <<
    stuff << endl <<
    more stuff << endl;

Also, it probably is not a great idea to try and inject multiple strings into the middle of localized text like this.  Much better to have a single localized string and value after it.

src/AppInstallerCLICore/Commands/FeaturesCommand.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+#include "Command.h"
+#include <winget/UserSettings.h>
+
+namespace AppInstaller::CLI
+{
+    struct FeaturesCommand final : public Command
+    {
+        // This command is used as an example on how experimental features can be used.
+        // To enable this command set ExperimentalCmd = true in the settings file.
```
JohnMcPMS, line 12: Copy pasta

src/AppInstallerCLICore/Commands/FeaturesCommand.cpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    Resource::LocString FeaturesCommand::LongDescription() const
+    {
+        return { Resource::String::FeaturesCommandLongDescription };
+    }
+
+    std::string FeaturesCommand::HelpLink() const
+    {
+        return "https://aka.ms/winget-seattings";
```
JohnMcPMS, line 25: >https://aka.ms/winget-seattings [](start = 16, length = 31)

Incorrect copy pasta

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -156,9 +162,34 @@
...
+  </data>
+  <data name="FeaturesCommandLongDescription" xml:space="preserve">
+    <value>Shows the status of experimental features. Experimental features can be turn on via winget settings.</value>
+  </data>
+  <data name="FeaturesCommandShortDescription" xml:space="preserve">
+    <value>Shows the status of experimental features.</value>
+  </data>
+  <data name="FeaturesMessage" xml:space="preserve">
+    <value>The following experimental features are in progress.
+They can be configured through the settings file (winget settings).</value>
```
JohnMcPMS, line 191: >winget settings [](start = 50, length = 15)

I think putting the ' ' quotes around it makes it look more like a command to run.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -156,9 +162,34 @@
...
+    <value>Experimental feature:</value>
+  </data>
   <data name="ExtraPositionalError" xml:space="preserve">
     <value>Found a positional argument when none was expected</value>
   </data>
+  <data name="FeatureDisabledMessage" xml:space="preserve">
+    <value>If you want to try this feature, please go to winget settings and enable it. See</value>
+  </data>
+  <data name="FeaturesCommandLongDescription" xml:space="preserve">
+    <value>Shows the status of experimental features. Experimental features can be turn on via winget settings.</value>
```
JohnMcPMS, line 184: >winget settings [](start = 95, length = 15)

You will have to put a "lock" on this, see this elsewhere in the file:
    <comment>{Locked="VirtualTerminal"}</comment>

src/AppInstallerCLICore/Commands/FeaturesCommand.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+        if (!featuresInfo.empty())
+        {
+            table.Complete();
+        }
+        else
+        {
+            // Better work hard to get some out there!
+            context.Reporter.Info() << Resource::String::NoExperimentalFeaturesMessage << std::endl;
+        }
```
JohnMcPMS, line 47: Feels like everything should be inside this if/else

src/AppInstallerCLICore/Commands/ExperimentalCommand.cpp
```diff
@@ -0,0 +1,86 @@
...
+                        ZNk#QNQQdx-           !N@#RgQy,-                             
+                       _Zy ,h03o= `           -!G@BNQ@#85T.                          
+                        `      `                :*vxL`)L! '                          
+)"sv;
+                context.Reporter.Info() << s_ninjaCat << std::endl;
+            }
+            else
+            {
+                // Here we warn the user that the argument exists and is not enabled.
+                Argument::ForType(Execution::Args::Type::ExperimentalArg).OutputEnableMessage(context);
```
JohnMcPMS, line 81: This type of thing should be done either at parsing or by Command::Execute.

src/AppInstallerCommonCore/JsonUtil.cpp
```diff
@@ -32,5 +32,18 @@ namespace AppInstaller::Utility
         return value;
     }
 
+    template<>
+    std::optional<bool> GetValue(const Json::Value& node)
+    {
+        std::optional<bool> value = std::nullopt;
```
JohnMcPMS, line 38: > = std::nullopt [](start = 33, length = 15)

This is the default constructor's behavior already.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -106,6 +157,8 @@ namespace AppInstaller::Settings
         }
 
         static std::filesystem::path SettingsFilePath();
+        static std::tuple<std::string, std::string, std::string> GetFeatureInfo(ExperimentalFeature feature);
```
JohnMcPMS, line 160: >std::tuple<std::string, std::string, std::string> [](start = 15, length = 49)

Don't use a tuple, take the time to create a struct with 3 names so that we can actually read the code.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -106,6 +157,8 @@ namespace AppInstaller::Settings
...
+        static std::vector<std::tuple<std::string, std::string, std::string>> GetFeaturesInfo();
```
JohnMcPMS, line 161: >static std::vector<std::tuple<std::string, std::string, std::string>> GetFeaturesInfo(); [](start = 8, length = 88)

I don't really agree with tacking the features directly into the UserSettings.  It is definitely a layer on top of them, and so it should be a separate type.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -78,12 +116,25 @@ namespace AppInstaller::Settings
...
+        {
+            using json_t = ExperimentalFeature;
+            using value_t = ExperimentalFeature;
+
+            static constexpr value_t DefaultValue = ExperimentalFeature::None;
+            static constexpr std::string_view Path = ".experimentalFeatures.%"sv;
+
+            // Not needed
+            // static std::optional<value_t> Validate(const json_t& value);
+        };
```
JohnMcPMS, line 136: Special case sense is tingling...

src/AppInstallerCLICore/Commands/FeaturesCommand.cpp
msftrubengu: This was intentional. Are we saying we will have a special md file/microsoft docs page for this hidden command? I was thinking settings.md will have in the experimentalFeatures entry something like "to know which experimental features are available run winget features"

---
In reply to: [446445866](https://github.com/microsoft/winget-cli/pull/460#discussion_r446445866) [](ancestors = 446445866)

msftrubengu: I guess the link can point to <github url>#experimentalFeatures to be clear

---
In reply to: [447183011](https://github.com/microsoft/winget-cli/pull/460#discussion_r447183011) [](ancestors = 447183011,446445866)

src/AppInstallerCLITests/UserSettings.cpp
```diff
@@ -271,3 +271,44 @@ TEST_CASE("SettingAutoUpdateIntervalInMinutes", "[settings]")
         REQUIRE(userSettingTest.GetWarnings().size() == 1);
     }
 }
+
+// Test one experimental feature in usersettingstest context because there's no good way to test ExperimentalFeature
```
JohnMcPMS, line 275: >ExperimentalFeature [](start = 97, length = 19)

I think that you could have tests, but you would either need your Reload function back, or a test hook for features that allows you to inject your own settings object to be the one it uses.

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    std::vector<ExperimentalFeature> ExperimentalFeature::GetAllFeatures()
+    {
+        return
+        {
+            // Commented out on purpose these are just to provide example on how experimental features work.
+            // GetFeature(Feature::ExperimentalCmd),
+            // GetFeature(Feature::ExperimentalArg)
```
JohnMcPMS, line 46: You can do this using the make_integer_sequence and folding, even if you want to exclude some you can use a Max that isn't at the end (these test ones are after).

src/AppInstallerCLITests/UserSettings.cpp
msftrubengu: I kind of gave up after seeing all the amounts of changes i was doing plus me not wanting to mess with the constness of UserSettings::Instance

---
In reply to: [447942903](https://github.com/microsoft/winget-cli/pull/460#discussion_r447942903) [](ancestors = 447942903)

ranm-msft:
>     }

add , at the end (like in line 18)

---
Refers to: doc/Settings.md:37 in 2c407ae. [](commit_id = 2c407ae50e306ffe1ea36e0e52c67b2dbd685ff3, deletion_comment = False)

doc/Settings.md
```diff
@@ -43,4 +43,17 @@ Color of the progress bar that WinGet displays when not specified by arguments.
...
+
+In order to allow work to be done in master, and distributed to early adopters for their feedback, settings have the ability to control "experimental" features.
+
+Currently there are no experimental features, just sample ones. Once they get implemented this file will be updated.
+
+```
+   "experimentalFeatures": {
+       "experimentalCmd": true,
+       "experimentalArg": false,
+   }
```
ranm-msft, line 58: >} [](start = 3, length = 1)

add coma

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -383,6 +412,9 @@
   <data name="TagArgumentDescription" xml:space="preserve">
     <value>Filter results by tag</value>
   </data>
+  <data name="ThankYou" xml:space="preserve">
+    <value>Thank you from using winget</value>
```
ranm-msft, line 416: >from [](start = 21, length = 4)

for

Issue #315, opened 2020-05-24
The Chinese path cannot be used when using the hash command.File does not exist.


masx200:
# Brief description of your issue

The Chinese path cannot be used when using the hash command.File does not exist.

# Steps to reproduce

```
ls D:\迅雷下载\winrar-x64-590.exe
```

```
    目录: D:\迅雷下载


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        2020/5/24     18:23        3226248 winrar-x64-590.exe

```

```
 winget hash D:\迅雷下载\winrar-x64-590.exe

```
# Expected behavior

```
 winget hash C:\Users\ma\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\AC\INetCache\KXK2CX9E\winrar-x64-590[1].exe
```
```
File Hash: af32c365c56b8a107af335fcb471298335a708d7ff803025e2b17b207c3d83ab
```
# Actual behavior

```
File does not exist: D:\杩呴浄涓嬭浇\winrar-x64-590.exe
```

# Environment

```none
[winget --info]
Windows Package Manager version
Windows: Windows.Desktop version
Package: Microsoft.DesktopAppInstaller version

Any other software?
Windows Package Manager v0.1.41331 预览
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.18363.836
Package: Microsoft.DesktopAppInstaller v1.0.41331.0

Links:
  Privacy Statement: https://aka.ms/winget-privacy
  License agreement: https://aka.ms/winget-license
  3rd Party Notices: https://aka.ms/winget-3rdPartyNotice
  Homepage:          https://aka.ms/winget
```

yao-msft:
Related to #181

yao-msft:
The fix should have been included with release [v0.2.2521](https://github.com/microsoft/winget-cli/releases/tag/v.0.2.2521-preview). Below is a working snapshot.

![Capture2](https://user-images.githubusercontent.com/50888816/94089120-22a85d80-fdc7-11ea-9cb2-1f141f034022.JPG)

Please feel free to reopen the issue if the fix does not work as expected.

Issue #181, opened 2020-05-19
Unable to install packages when user name contains non-ASCII characters


Callidior:
# Brief description of your issue

My windows user name is "Björn", and hence contains a non-ASCII character. When I try to install any package using `winget`, I receive an error message similar to the following one:

> "C:\Users\BjÃ¶rn\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\TempState\WinGet\gimp.gimp.2.10.18.exe" konnte nicht gefunden werden. Stellen Sie sicher, dass Sie den Namen richtig eingegeben haben und wiederholen Sie den Vorgang.

The specified path indeed does not exist. Replacing "BjÃ¶rn" with my correct username, "Björn", though, leads to the correct file downloaded by `winget`.

# Steps to reproduce

1. Log on with a Windows username containing non-ASCII characters.
2. Open `cmd`.
3. Run: `winget install gimp.gimp -e`

# Expected behavior

`winget` should download the installer and execute it.

# Actual behavior

`winget` downloads the installer, but fails to execute it, because it uses an invalid path for running the downloaded file. This leads to the following error message in a message box window:

> "C:\Users\BjÃ¶rn\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\TempState\WinGet\gimp.gimp.2.10.18.exe" konnte nicht gefunden werden. Stellen Sie sicher, dass Sie den Namen richtig eingegeben haben und wiederholen Sie den Vorgang.

On the command line, I get this message:

> Installer failed with exit code: 1223

# Environment

```none
[winget --info]
Windows Package Manager v0.1.41331 Vorschau
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.18363.836
Package: Microsoft.DesktopAppInstaller v1.0.41331.0
```

tomgron:
Got this too : 
![image](https://user-images.githubusercontent.com/1743381/83458351-0d4fa200-a46b-11ea-92dd-80b250043d1d.png)

KristofKekesi:
me too (my username is Kristóf)

Ed1123:
Any workaround?

yao-msft:
We plan to fix this in v0.2 milestone. Thanks.

yao-msft:
[...]

PR #466, opened 2020-06-27
Fix unicode path issues

yao-msft:
For #181 #315 

## Cause
There're multiple places a std::filesystem::path is created with narrow string input. It's ok most of the time since we control the expected path(i.e. index.msix, etc). But if the file path involves user input, this will be a problem.

## Change
Examine the code and determine places that a path involves user input. Call Utility::ConvertToUtf16 before creating std::filesystem::path.

## Validation
Updated E2E tests to contain extended character scenario. Install e2e tests will be updated when it's fully enabled.
Manually tested install scenario by creating a user account with extended characters and try some installation.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/466)


src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -16,16 +16,15 @@ namespace AppInstaller::CLI::Workflow
         {
             AICLI_LOG(CLI, Info, << "Starting installer. Path: " << filePath);
 
-            SHELLEXECUTEINFOA execInfo = { 0 };
-            execInfo.cbSize = sizeof(SHELLEXECUTEINFO);
+            SHELLEXECUTEINFOW execInfo = { 0 };
+            execInfo.cbSize = sizeof(SHELLEXECUTEINFOW);
```
JohnMcPMS, line 20: >SHELLEXECUTEINFOW [](start = 37, length = 17)

Just use sizeof(execInfo), then you don't have to chase its type.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -16,16 +16,15 @@ namespace AppInstaller::CLI::Workflow
...
-            SHELLEXECUTEINFOA execInfo = { 0 };
-            execInfo.cbSize = sizeof(SHELLEXECUTEINFO);
+            SHELLEXECUTEINFOW execInfo = { 0 };
+            execInfo.cbSize = sizeof(SHELLEXECUTEINFOW);
             execInfo.fMask = SEE_MASK_NOCLOSEPROCESS;
-            std::string filePathUTF8Str = filePath.u8string();
-            execInfo.lpFile = filePathUTF8Str.c_str();
-            execInfo.lpParameters = args.c_str();
+            execInfo.lpFile = filePath.c_str();
+            execInfo.lpParameters = Utility::ConvertToUTF16(args).c_str();
```
JohnMcPMS, line 23: >Utility::ConvertToUTF16(args) [](start = 36, length = 29)

This is a temporary whose life ends at the semicolon.  You need to keep it local to be able to pass it in this way.

JohnMcPMS:
This seems like a lot of band-aids that won't cover new cases well.  Is there a more fundamental change that we can make to prevent future recurrence?  If we had char8_t, that would solve the majority of the problems, but we don't with C++17.  We could turn it on, either with /Zc:char8_t or /std:c++latest, but they both have their issues.  However, it is the best bet to prevent all future issues.

But changing *every* std::string to std::u8string (or at least every one that might possibly be UTF-8), and the knock on effects that would have, could be a massive change.

The only thought for an intermediate solution would be to create a new filesystem type that treats std::string as std::u8string, much like I did for NormalizedString.  But even that seems like it would be easy to miss one of the functions that we don't use or don't notice.

Can you think of any other ways to prevent potential future regression in this area?

yao-msft:
Discussed offline of a couple options. All require fair amount of work over a bug fix. Will file an issue to track the larger work.

---
In reply to: [650522963](https://github.com/microsoft/winget-cli/pull/466#issuecomment-650522963) [](ancestors = 650522963)

yao-msft:
> Discussed offline of a couple options. All require fair amount of work over a bug fix. Will file an issue to track the larger work.
> 
> In reply to: 650522963 [](ancestors = 650522963)

#473 filed

PR #482, opened 2020-07-07
Block elevated process from overriding hash mismatch

JohnMcPMS:
## Change
To prevent the user from running a package whose hash does not match in an elevated context, this change explicitly blocks the scenario.  In addition, the inline prompt to proceed when the hash does not match is replaced with usage of the --force option.

While executing on this change, I also took on a fair amount of additional localization work that has yet to be completed.  This included transitioning many existing strings to resources, and removal of the prompt functionality altogether.  This is partly due to localizing it being a headache, and partly because it prevents easy scripting.

## Validation
Existing regression tests.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/482)


src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -18,21 +19,22 @@ namespace AppInstaller::CLI
...
+            Argument::ForType(Args::Type::Channel),
+            Argument::ForType(Args::Type::Source),
+            Argument::ForType(Args::Type::Exact),
+            Argument::ForType(Args::Type::Interactive),
+            Argument::ForType(Args::Type::Silent),
+            Argument::ForType(Args::Type::Language),
+            Argument::ForType(Args::Type::Log),
+            Argument::ForType(Args::Type::Override),
+            Argument::ForType(Args::Type::InstallLocation),
+            Argument{ "force", Argument::NoAlias, Args::Type::Force, Resource::String::InstallForceArgumentDescription, ArgumentType::Flag },
```
yao-msft, line 37: >Argument [](start = 12, length = 8)

Curious why is this one not added to ForType?

JohnMcPMS: It uses different help text for each case.  Honestly, now that we have the resource strings that whole function could be done away with.

---
In reply to: [451283891](https://github.com/Microsoft/pkg/pull/482#discussion_r451283891) [](ancestors = 451283891)

PR #493, opened 2020-07-13
Implement store app installation from manifest

yao-msft:
This is the first part of #117

## Change
- Refactor manifest code to separate parsing logic from object model
- Added ProductId and MSStore InstallerType and related validation
- Added Store app installation code from a manifest

## Validation
- Existing tests passed
- Added new tests to Manifest and Workflow
- Manually tested on a VM and make sure the UI behaves as expected

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/493)


src/AppInstallerCLICore/ExecutionProgress.cpp
```diff
@@ -192,7 +192,6 @@ namespace AppInstaller::CLI::Execution
         for (size_t i = 0; !m_canceled; ++i)
         {
             constexpr size_t repetitionCount = 20;
-            ApplyStyle(i % repetitionCount, repetitionCount, true);
```
JohnMcPMS, line 195: >ApplyStyle(i % repetitionCount, repetitionCount, true); [](start = 12, length = 55)

Why has this been removed?  Seems unrelated.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -139,6 +143,12 @@ namespace AppInstaller::CLI::Workflow
 
     void VerifyInstallerHash(Execution::Context& context)
     {
+        const auto& installer = context.Get<Execution::Data::Installer>().value();
+        if (installer.InstallerType == ManifestInstaller::InstallerTypeEnum::MSStore)
+        {
+            return;
+        }
```
JohnMcPMS, line 150: I would rather change how this is decided to be run than to start building early outs into every function.  Probably better to move calling it out of the base flow and into DownloadInstaller where it is needed.

JohnMcPMS:
>     void ShowInstallationDisclaimer(Execution::Context& context)

This might be different for Store.  Should investigate more.

---
Refers to: src/AppInstallerCLICore/Workflows/InstallFlow.cpp:41 in ae95b7f. [](commit_id = ae95b7faa30743b1dea118f5ab752baf3408f404, deletion_comment = False)

src/AppInstallerCLICore/ExecutionProgress.cpp
yao-msft: When I run locally to verify the UI behavior, the spinner applies the theme color, but on exit, it's not reverting to default. And this method is called NoVT, so I thought we left it here by accident?

---
In reply to: [453968675](https://github.com/microsoft/winget-cli/pull/493#discussion_r453968675) [](ancestors = 453968675)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +261,128 @@ namespace AppInstaller::CLI::Workflow
             std::filesystem::remove(path);
         }
     }
+
+    void MSStoreInstall(Execution::Context& context)
+    {
+        // Feature toggle check
+        if (!Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::ExperimentalMSStore))
+        {
+            context.Reporter.Error() << Resource::String::FeatureDisabledMessage << std::endl;
```
JohnMcPMS, line 270: >context.Reporter.Error() << Resource::String::FeatureDisabledMessage << std::endl; [](start = 12, length = 82)

This is typically coupled with the string that indicates which feature needs to be enabled.  I would suggest creating a Workflow object that can handle this generically.  So when invoking this from ExecuteInstaller:

context <<
    EnsureFeatureEnabled(Settings::ExperimentalFeature::Feature::ExperimentalMSStore) <<
    MSStoreInstall;

That way it can be reused easily.

src/AppInstallerCLICore/ExecutionProgress.cpp
JohnMcPMS: No style is applied in the NoVT case.  If you are seeing that happening, it is due to a different bug.

---
In reply to: [453998322](https://github.com/Microsoft/pkg/pull/493#discussion_r453998322) [](ancestors = 453998322,453968675)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +261,128 @@ namespace AppInstaller::CLI::Workflow
...
+        }
+
+        auto productId = Utility::ConvertToUTF16(context.Get<Execution::Data::Installer>()->ProductId);
+
+        constexpr std::wstring_view s_StoreClientName = L"Microsoft.WindowsStore"sv;
+        constexpr std::wstring_view s_StoreClientPublisher = L"CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US"sv;
+
+        // Policy check
+        AppInstallManager installManager;
+        if (installManager.IsStoreBlockedByPolicyAsync(s_StoreClientName, s_StoreClientPublisher).get())
```
JohnMcPMS, line 282: >IsStoreBlockedByPolicyAsync [](start = 27, length = 27)

Given the name of this function, I would expect that it already knows the Store's identity and that I wouldn't need to provide it.

yao-msft: Yeah, unfortunately this is how the api is supposed to be called. Maybe they are thinking about multiple stores when they design this API

---
In reply to: [454000116](https://github.com/microsoft/winget-cli/pull/493#discussion_r454000116) [](ancestors = 454000116)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +261,128 @@ namespace AppInstaller::CLI::Workflow
...
+            productId,              // ProductId
+            winrt::hstring(),       // CatalogId
+            winrt::hstring(),       // FlightId
+            L"WinGetCli",           // ClientId
+            false,                  // repair
+            false,
+            winrt::hstring(),
+            nullptr).get();
+
+        auto installItem = installItems.GetAt(0);
```
JohnMcPMS, line 338: >auto installItem = installItems.GetAt(0); [](start = 8, length = 41)

Do we not need to handle all of the items?  A comment is probably required here.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +261,128 @@ namespace AppInstaller::CLI::Workflow
...
+        while (installItem.GetCurrentStatus().PercentComplete() == 0)
+        {
+            context.Reporter.ShowIndefiniteProgress(true);
+            CheckInstallStatus();
+            Sleep(500);
+        }
+
+        context.Reporter.ShowIndefiniteProgress(false);
+
+        context.Reporter.ExecuteWithProgress(std::bind(
```
JohnMcPMS, line 366: >ExecuteWithProgress [](start = 25, length = 19)

ExecuteWithProgress handles the "show indefinite until progress" if you don't send any progress.  I think you can roll the above while into the lower loop to achieve the same (don't call OnProgress until PercentComplete is > 0.

yao-msft: That's convenient. I'll clean it up

---
In reply to: [454018290](https://github.com/microsoft/winget-cli/pull/493#discussion_r454018290) [](ancestors = 454018290)

JohnMcPMS, line 366: >std::bind [](start = 45, length = 9)

You don't need to bind; the whole point of that is to create a callable with IProgressCallback& progress as its only parameter.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +261,128 @@ namespace AppInstaller::CLI::Workflow
...
+                while (installItem.GetCurrentStatus().PercentComplete() < 100)
+                {
+                    progress.OnProgress((uint64_t)installItem.GetCurrentStatus().PercentComplete(), 100, ProgressType::Percent);
+                    CheckInstallStatus();
+
+                    if (progress.IsCancelled())
+                    {
+                        installItem.Cancel();
+                    }
+                    Sleep(500);
```
JohnMcPMS, line 377: >Sleep(500); [](start = 20, length = 11)

I would suggest sleeping more like 100ms, otherwise cancellation appears to be sluggish.

src/AppInstallerRepositoryCore/Manifest/YamlManifestParser.h
```diff
@@ -0,0 +1,64 @@
...
+#include "ManifestValidation.h"
+#include "Manifest.h"
+
+namespace AppInstaller::Manifest
+{
+    static const uint64_t MaxSupportedMajorVersion = 1;
+    static const ManifestVer PreviewManifestVersion = ManifestVer("0.1.0", false);
+    static const ManifestVer PreviewManifestVersionV2 = ManifestVer("0.2.0", false);
+
+    struct YamlManifestParser
```
JohnMcPMS, line 13: >YamlManifestParser [](start = 11, length = 18)

Can we drop Manifest from so many names since we are in the Manifest namespace?

Manifest::YamlParser::Create
Manifest::YamlParser::CreateFromPath

You don't need to have names repeated if you don't using namespace the previous names out.

src/AppInstallerRepositoryCore/Manifest/YamlManifestParser.h
```diff
@@ -0,0 +1,64 @@
...
+
+    private:
+        // These pointers are referenced in the processing functions in manifest field info table.
+        YAML::Node* m_p_installersNode;
+        YAML::Node* m_p_switchesNode;
+        YAML::Node* m_p_localizationsNode;
+        AppInstaller::Manifest::Manifest* m_p_manifest;
+        AppInstaller::Manifest::ManifestInstaller* m_p_installer;
+        std::map<ManifestInstaller::InstallerSwitchType, Utility::NormalizedString>* m_p_switches;
+        AppInstaller::Manifest::ManifestLocalization* m_p_localization;
```
JohnMcPMS, line 30: Initialize all with nullptr.

src/AppInstallerRepositoryCore/Manifest/YamlManifestParser.h
yao-msft: Sure

---
In reply to: [454024340](https://github.com/microsoft/winget-cli/pull/493#discussion_r454024340) [](ancestors = 454024340)

src/AppInstallerRepositoryCore/Manifest/YamlManifestParser.cpp
```diff
@@ -0,0 +1,425 @@
...
+            return result;
+        }
+    }
+
+    void YamlManifestParser::PrepareManifestFieldInfos(const ManifestVer& manifestVer)
+    {
+        RootFieldInfos =
+        {
+            { "ManifestVersion", PreviewManifestVersion, [this](const YAML::Node&) { /* ManifestVersion already processed */ }, false,
+            // Regex here is to prevent leading 0s in the version, this also keeps consistent with other versions in the manifest
```
JohnMcPMS, line 45: >Regex [](start = 15, length = 5)

Moving the regex seems to be too much.  If we are separating the parser, it should be to have it parse the syntax of the format, and then we should add these semantics onto that.

Parser > Raw Manifest > Validation

Either that or the parser gets an interface to expose selection of individual values out via some syntax we define, and the top level parser gets passed a syntactic parser to provide that.

Basically, if we are separating the code out, it should be written to enable any syntax.

src/AppInstallerRepositoryCore/Manifest/ManifestValidation.cpp
```diff
@@ -6,126 +6,125 @@
...
+            if (installer.InstallerType == ManifestInstaller::InstallerTypeEnum::Unknown)
+            {
+                resultErrors.emplace_back(ManifestError::InvalidFieldValue, "InstallerType");
+            }
+
+            if (installer.InstallerType == ManifestInstaller::InstallerTypeEnum::MSStore)
+            {
+                // MSStore type is not supported in community repo
+                resultErrors.emplace_back(
+                    ManifestError::FieldNotSupported, "InstallerType",
```
JohnMcPMS, line 88: >ManifestError::FieldNotSupported, "InstallerType", [](start = 20, length = 50)

Probably need ValueNotSupported so that it can be more clear that it isn't InstallerType, but it's value that is bad.

src/AppInstallerRepositoryCore/Manifest/ManifestValidation.cpp
```diff
@@ -6,126 +6,125 @@
...
+                if (installer.Url.empty())
+                {
+                    resultErrors.emplace_back(ManifestError::RequiredFieldMissing, "Url");
+                }
+                if (installer.Sha256.empty())
                 {
-                    errors.emplace_back(ManifestError::FieldUnknown, key, "", keyValuePair.first.Mark().line, keyValuePair.first.Mark().column, ValidationError::Level::Warning);
+                    resultErrors.emplace_back(ManifestError::RequiredFieldMissing, "Sha256");
+                }
+                // ProductId should be used
```
yao-msft, line 107: >should be [](start = 29, length = 9)

should not be used

src/AppInstallerRepositoryCore/Manifest/YamlManifestParser.cpp
yao-msft: Chatted offline and leave it for now.

---
In reply to: [454031238](https://github.com/microsoft/winget-cli/pull/493#discussion_r454031238) [](ancestors = 454031238)

src/AppInstallerRepositoryCore/Manifest/YamlManifestParser.h
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the MIT License.
+#pragma once
+#include "ManifestValidation.h"
+#include "Manifest.h"
+
+namespace AppInstaller::Manifest
+{
+    static const uint64_t MaxSupportedMajorVersion = 1;
+    static const ManifestVer PreviewManifestVersion = ManifestVer("0.1.0", false);
+    static const ManifestVer PreviewManifestVersionV2 = ManifestVer("0.2.0", false);
```
yao-msft, line 11: >"0.2.0" [](start = 68, length = 7)

Per discussion, this should be 0.2.0-msstore

yao-msft:
>     void ShowInstallationDisclaimer(Execution::Context& context)

Added placeholder message

---
In reply to: [657845039](https://github.com/microsoft/winget-cli/pull/493#issuecomment-657845039) [](ancestors = 657845039)

---
Refers to: src/AppInstallerCLICore/Workflows/InstallFlow.cpp:41 in ae95b7f. [](commit_id = ae95b7faa30743b1dea118f5ab752baf3408f404, deletion_comment = False)

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -236,6 +240,10 @@ They can be configured through the settings file 'winget settings'.</value>
   <data name="InstallationDisclaimer2" xml:space="preserve">
     <value>Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.</value>
   </data>
+  <data name="InstallationDisclaimerMSStore" xml:space="preserve">
+    <value>This package is provided through Microsoft Store. Winget may need to acquire the package from Microsoft Store on behalf of the current user.</value>
```
JohnMcPMS, line 244: >Winget [](start = 61, length = 6)

Should probably not be capitalized, as the proper noun is not.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -173,7 +173,11 @@
     <value>Found a positional argument when none was expected</value>
   </data>
   <data name="FeatureDisabledMessage" xml:space="preserve">
-    <value>This command is a work in progress, and may be changed dramatically or removed altogether in the future. To enable it, edit your settings ('winget settings') to include the experimental feature</value>
+    <value>This feature is a work in progress, and may be changed dramatically or removed altogether in the future. To enable it, edit your settings ('winget settings') to include the experimental feature</value>
+    <comment>{Locked="winget settings"}</comment>
+  </data>
+  <data name="FeatureDisabledMessageWithFeatureName" xml:space="preserve">
```
JohnMcPMS, line 179: >FeatureDisabledMessageWithFeatureName [](start = 14, length = 37)

The other value is already used this way, just as the exception message that gets output as:

<message>: <parameter>

Don't make a second one.

src/AppInstallerCLICore/Workflows/WorkflowBase.h
```diff
@@ -127,6 +127,12 @@ namespace AppInstaller::CLI::Workflow
     // Inputs: None
     // Outputs: None
     void EnsureRunningAsAdmin(Execution::Context& context);
+
+    // Ensures that the feature is enabled.
+    // Required Args: None
+    // Inputs: Feature
+    // Outputs: None
+    void EnsureFeatureEnabled(Execution::Context& context);
```
JohnMcPMS, line 135: >EnsureFeatureEnabled [](start = 9, length = 20)

Don't read the feature out of a context value, construct an object (see VerifyFile above).

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -236,6 +240,10 @@ They can be configured through the settings file 'winget settings'.</value>
   <data name="InstallationDisclaimer2" xml:space="preserve">
     <value>Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.</value>
   </data>
+  <data name="InstallationDisclaimerMSStore" xml:space="preserve">
```
JohnMcPMS, line 243: >InstallationDisclaimerMSStore [](start = 14, length = 29)

If this is a placeholder, did you send mail to ask for an official message?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -192,6 +205,10 @@ namespace AppInstaller::CLI::Workflow
         case ManifestInstaller::InstallerTypeEnum::Msix:
             context << MsixInstall;
             break;
+        case ManifestInstaller::InstallerTypeEnum::MSStore:
+            context.Add<Execution::Data::Feature>(Settings::ExperimentalFeature::Feature::ExperimentalMSStore);
+            context << EnsureFeatureEnabled << MSStoreInstall;
```
JohnMcPMS, line 210: >EnsureFeatureEnabled [](start = 23, length = 20)

So this becomes:

```
context <<
    EnsureFeatureEnabled(Settings::ExperimentalFeature::Feature::ExperimentalMSStore) <<
    MSStoreInstall;
```

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: From my reading of the docs, if the target is a bundle this will have every package.  I think we do need to watch all of them.

---
In reply to: [454004580](https://github.com/Microsoft/pkg/pull/493#discussion_r454004580) [](ancestors = 454004580)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +265,116 @@ namespace AppInstaller::CLI::Workflow
...
+        auto CheckInstallStatus = [&]()
+        {
+            HRESULT errorCode = installItem.GetCurrentStatus().ErrorCode();
+            if (!SUCCEEDED(errorCode))
+            {
+                // Cancel the installation request on failure. Otherwise it'll leave an error
+                // request in Store client installation queue.
+                installItem.Cancel();
+
+                context.Reporter.Info() << Resource::String::MSStoreInstallFailed << ' ' << errorCode << std::endl;
```
JohnMcPMS, line 346: >errorCode [](start = 92, length = 9)

I don't think HRESULT automatically outputs nicely.  I would suggest adding an operator<<(HRESULT) to do so.

out << "0x" << std::hex << std::fill('0') << std::width(8) << hr;

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +265,116 @@ namespace AppInstaller::CLI::Workflow
...
+                AICLI_LOG(CLI, Error, << "MSStore install failed. ProductId: " << Utility::ConvertToUTF8(productId) << " HResult: " << errorCode);
+                AICLI_TERMINATE_CONTEXT(errorCode);
+            }
+        };
+
+        context.Reporter.ExecuteWithProgress(
+            [&](IProgressCallback& progress)
+            {
+                while (installItem.GetCurrentStatus().PercentComplete() < 100)
```
JohnMcPMS, line 355: >installItem.GetCurrentStatus().PercentComplete() [](start = 23, length = 48)

Since you need this 3 times each iteration and it might require a cross process COM call to get, maybe just save it in a local each time through the loop?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -321,6 +322,19 @@ namespace AppInstaller::CLI::Workflow
...
+
+    void EnsureFeatureEnabled(Execution::Context& context)
+    {
+        auto feature = context.Get<Execution::Data::Feature>();
+
+        if (!Settings::ExperimentalFeature::IsEnabled(feature))
+        {
+            context.Reporter.Error() << Resource::String::FeatureDisabledMessageWithFeatureName << ' ' <<
+                Settings::ExperimentalFeature::GetFeature(feature).Name() << std::endl;
+            AICLI_LOG(CLI, Error, << Settings::ExperimentalFeature::GetFeature(feature).Name() << "feature is disabled. Execution cancelled.");
```
JohnMcPMS, line 334: >feature [](start = 99, length = 7)

space in front of feature

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +265,116 @@ namespace AppInstaller::CLI::Workflow
...
+            winrt::hstring(),       // CatalogId
+            winrt::hstring(),       // FlightId
+            L"WinGetCli",           // ClientId
+            false,                  // repair
+            false,
+            winrt::hstring(),
+            nullptr).get();
+
+        // We install one package at a time so the AppInstallItem list returned should contain only one entry.
+        auto installItem = installItems.GetAt(0);
```
JohnMcPMS, line 335: >auto installItem = installItems.GetAt(0); [](start = 8, length = 41)

We should probably log a bunch of info about these requests so that we can easily associate them with store/deployment logs.  Store request ids, PFNs, anything else that will make it easier to link the two.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
yao-msft: I asked Kevin to give me one. We'll update later

---
In reply to: [455231560](https://github.com/microsoft/winget-cli/pull/493#discussion_r455231560) [](ancestors = 455231560)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
yao-msft: Looking at the properties of AppInstallItem, only ProductId and PFN are related to our diagnosis. Logged these 2.

---
In reply to: [455243165](https://github.com/microsoft/winget-cli/pull/493#discussion_r455243165) [](ancestors = 455243165)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
yao-msft: I created a #define so that not every long will be printed as HResult

---
In reply to: [455237189](https://github.com/microsoft/winget-cli/pull/493#discussion_r455237189) [](ancestors = 455237189)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -321,6 +322,18 @@ namespace AppInstaller::CLI::Workflow
             AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_COMMAND_REQUIRES_ADMIN);
         }
     }
+
+    void EnsureFeatureEnabled::operator()(Execution::Context& context) const
+    {
+        if (!Settings::ExperimentalFeature::IsEnabled(m_feature))
+        {
+            context.Reporter.Error() << Resource::String::FeatureDisabledMessage << ' ' <<
+                Resource::String::FeatureDisabledFeatureName << ' ' <<
```
JohnMcPMS, line 331: Why not just: ": ".  That is what is happening in the case that the whole command or an argument is not available.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +266,124 @@ namespace AppInstaller::CLI::Workflow
...
+                uint64_t currentProgress = 0;
+
+                while (currentProgress < overallProgressMax)
+                {
+                    currentProgress = 0;
+
+                    for (auto const& installItem : installItems)
+                    {
+                        const auto& status = installItem.GetCurrentStatus();
+                        currentProgress += (uint64_t)(status.PercentComplete());
```
JohnMcPMS, line 357: >(uint64_t) [](start = 43, length = 10)

static_cast<uint64_t>

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -248,4 +266,124 @@ namespace AppInstaller::CLI::Workflow
...
+
+                        HRESULT errorCode = status.ErrorCode();
+                        if (!SUCCEEDED(errorCode))
+                        {
+                            context.Reporter.Info() << Resource::String::MSStoreInstallFailed << ' ' << WINGET_OSTREAM_FORMAT_HRESULT(errorCode) << std::endl;
+                            AICLI_LOG(CLI, Error, << "MSStore install failed. ProductId: " << Utility::ConvertToUTF8(productId) << " HResult: " << errorCode);
```
JohnMcPMS, line 363: >errorCode [](start = 147, length = 9)

WINGET_OSTREAM_FORMAT_HRESULT

Whoerr (commented):
exit

Whoerr:
closed

PR #503, opened 2020-07-17
Added homepage command using the search command as a base

skyhoshi:
## Summary of the Pull Request
Adds the "homepage" command that when referencing one manifest / app Id it will open the homepage that is within the manifest in the systems default browser.

* If there is no homepage within the manifest it will alert the user to that effect.
* If more than one manifest is found it will list the found manifests and not open a homepage.

## PR Checklist

* [X] Closes #501 
* [X] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [X] Tests added/passed : No documentation on where tests were required/needed/should be preformed. All existing tests pass.
* [X] Requires documentation to be updated : adding new page to https://github.com/skyhoshi/windows-uwp/tree/docs/hub/package-manager/winget
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx  
NOTE, In Response to above: I haven't had a conversation with anyone about the work. and I understand if this is rejected but I do ask that someone explains clearly why. this is one of my first edits on a project where the language is one that i consider myself weak in. and would appreciate this learning opportunity from the community. thanks!

## Detailed Description of the Pull Request / Additional comments
Added a new command that mimics the search function, added locallizable descriptions and added new telemetry call to monitor use of the homepage command and output.
the telemetry action would be to respond to bad actors attempting to include sites and webpages not intended to describe or provide information on the application included within the repository.
The command does a fuzzy lookup on the ID, Name, Monkier and tag fields to provide a list of possible matches, the output includes hints on how to refine or pick a certain manifest from the list. 

## Validation Steps Performed

Demonstration of a list of manifests returned when no query specified

```powershell
.\AppInstallerCLI.exe homepage
```

Results

```txt
Multiple packages found matching input criteria. Please refine the input.
Name                                  Id                                  Version
------------------------------------------------------------------------------------------
Robo3T                                3T.Robo3T                           1.3.1
~~~~~~~~~~~~~~~~~~~ Removed large portion of list for brevity ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Zerotier                              zerotier.zerotier                   1.4.6.0
More than one result was found please refine your query.
```

Demonstration of a list of manifests return from a query

```powershell
.\AppInstallerCLI.exe homepage Microsoft
```

Results

```txt
Multiple packages found matching input criteria. Please refine the input.
Name                      Id                               Version        Match
[...]

PR #507, opened 2020-07-22
Show cmd extension openurls of manifest

### LINKED ISSUE #501: Open Manifest's Homepage in default browser via a command
I was expecting to have a command that would open the homepage of the manifest without having to use copy paste within the command window itself. 
## Current Process I have to employ
exploring winget's : List of available manifests via 
```cmd
winget show 
```
The above command lists all the manifests. 
looking through the manifests some may interest me and I run 
```cmd
winget show <app.mainfest>
```
This outputs the data of the manifest in a human readable form.

Its at this point i have to Select the URL of the homepage and copy paste it into my browser. 
I was expecting to find an option to open the Homepage URL via the command line. 

I was epxecting to be able to type something like
```cmd
winget homepage <app.manifest>
```
Expected Functions of this command.
1. This command would [if homepage or other URL, is found in the manifest] take the url and open the url in the user's/systems default browser. 
1. If the homepage is not present in the manifest the output would notify the user the URL was not found in the manifest. 

Example: Scoop home <appName> would open the app's homepage listed in the manifest allowing for exploration. 

# Additional Consideration
I would also consider making at least one informatic URL (be it license[licenseURL], privacy policy[not found in manifest], or homepage url[Homepage]) required in the minimum required.

skyhoshi:
# Summary of the Pull Request
Adds subcommand's "--homepage" and "--license" to the show command that lookup, validates (has http at start) and will open the url in the default brower based on the HomepageUrl and LicenseUrl respectively.
 
# References
This PR replaces #503. 
Fixes #501 

# PR Checklist
* [X] Closes #501 : REPLACES PR: #503
* [X] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [X] Tests added/passed
* * NOTE: All existing tests pass, no new tests were added. ([:red_square:][:beginner:] -> help wanted | Needed [:white_flag:])
* [X] Requires documentation to be updated
* * NOTE: Documentation has been updated in a seperate branch of a seperate project that can be found at [package-manager-show Add Homepage License](https://github.com/skyhoshi/windows-uwp/blob/package-manager-add-show-homepagelicense/hub/package-manager/winget/show.md)
* [X] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #501

# Detailed Description of the Pull Request / Additional comments
So from my previous PR #503, I learned alot and upon discovering that workflows were being extracted into single seperate files of the same namespace I did just that, pulled the methods and logic out into two seperate workflow files. 

1. OpenUrlWorkflow.(h|cpp)
1. GetRequestedUrlFromManifest.(h|cpp)

## Open Url Workflow
This file and it's methods simply look for the URL's in question from the context and opens them as a shell execute command. this method of call is equivelent to opening the run box and typing in a URL and clicking ok. it opens the url in your system's default browswer.

## Get Requested Url From Manifest
This file and it's methods take the context's current search result and attempts to identify if only one manifest was retrieved. if not, it says so and outputs the search results allowing the user to refine or pick specificly.
if it only finds one manifest, it then extracts the url from the manifest starting and adds it as a data value back into the context for use in the Open Url Workflow.

# Validation Steps Performed

### Command Performed

```powershell
wingetdev show git
```

### Expectation

Display's the Mainfest information in human readable format

### Result (Includes Command Output)

```powershell
PS C:\> wingetdev show git
Found Git [Git.Git]
Version: 2.27.0
Publisher: Git
Description: Git version control system.
Homepage: https://git-scm.com/
License: GNU General Public License, version 2
License Url: https://github.com/git-for-windows/git/blob/master/COPYING
Installer:
  Type: Inno
  Download Url: https://github.com/git-for-windows/git/releases/download/v2.27.0.windows.1/Git-2.27.0-64-bit.exe
  SHA256: 5974db8c52b32f5e575ee021e8b47948892ce0e760095eef98c31e3bbd5167b6
```
### Command Performed

```powershell
wingetdev show git --homepage
```

### Expectation

1. Display's the Mainfest for the app information in human readable format
1. If the homepage exists (which it does for git) it will automatically open the homepage in the system's default browser.

### Result (Includes Command Output)

```powershell
PS C:\> wingetdev show git --homepage
Found Git [Git.Git]
Version: 2.27.0
Publisher: Git
Description: Git version control system.
Homepage: https://git-scm.com/
License: GNU General Public License, version 2
License Url: https://github.com/git-for-windows/git/blob/master/COPYING
Installer:
  Type: Inno
  Download Url: https://github.com/git-for-windows/git/releases/download/v2.27.0.windows.1/Git-2.27.0-64-bit.exe
  SHA256: 5974db8c52b32f5e575ee021e8b47948892ce0e760095eef98c31e3bbd5167b6
Opening Homepage Url: https://git-scm.com/
<Opens Homepage Default Browser with URL>
```

### Command Performed

```powershell
wingetdev show git --homepage --License
```

### Expectation

1. Display's the Mainfest for the app information in human readable format
1. If the homepage exists (which it does for git) it will automatically open the homepage in the system's default browser.
1. If the License exists (which it does for git) it will automatically open the license page in the system's default brower.

### Result (Includes Command Output)

```powershell
PS C:\> wingetdev show git --homepage --License
Found Git [Git.Git]
Version: 2.27.0
Publisher: Git
Description: Git version control system.
Homepage: https://git-scm.com/
License: GNU General Public License, version 2
License Url: https://github.com/git-for-windows/git/blob/master/COPYING
Installer:
  Type: Inno
  Download Url: https://github.com/git-for-windows/git/releases/download/v2.27.0.windows.1/Git-2.27.0-64-bit.exe
  SHA256: 5974db8c52b32f5e575ee021e8b47948892ce0e760095eef98c31e3bbd5167b6
Opening Homepage Url: https://git-scm.com/
Opening License Url: https://github.com/git-for-windows/git/blob/master/COPYING
<Opens Homepage Default Browser with URL>
<Opens License Default Browser with URL>
```

### Command Performed

```powershell
wingetdev show Playstation.PSnow --homepage --License
```

### Expectation

1. Display's the Mainfest for the app information in human readable format
1. No Homepage is found. States it did not find a homepage url for this manifest
1. If the License exists (which it does for Playstation.PsNow) it will automatically open the license page in the system's default brower.

### Result (Includes Command Output)

```powershell
PS C:\> wingetdev show Playstation.PSnow --homepage --License
Found PlayStation Now [PlayStation.PSNow]
Version: 11.1.2
Publisher: Sony Interactive Entertainment
Description: Get instant access to a huge on demand game collection – the only place to play PlayStation exclusive titles on Windows PC.
License: Copyright (c) Sony Interactive Entertainment
License Url: https://www.playstation.com/legal/important-information-about-the-ps-now-service/
Installer:
  Type: Exe
  Download Url: https://download-psnow.playstation.com/downloads/psnow/pc/PlayStationNow-11.1.2.exe
  SHA256: 660eaa555fd3376174c1ebccceb5008099c1da19a4ddfd1076436eefa67fedfe
No Homepage found within manifest
Opening License Url: https://www.playstation.com/legal/important-information-about-the-ps-now-service/

<Opens License Default Browser with URL> (License is 404 for some reason)
```

### Command Performed

```powershell
wingetdev show Playstation.PSnow --homepage
```

### Expectation

1. Display's the Mainfest for the app information in human readable format
1. No Homepage is found. States it did not find a homepage url for this manifest

### Result (Includes Command Output)

```powershell
Found PlayStation Now [PlayStation.PSNow]
Version: 11.1.2
Publisher: Sony Interactive Entertainment
Description: Get instant access to a huge on demand game collection – the only place to play PlayStation exclusive titles on Windows PC.
License: Copyright (c) Sony Interactive Entertainment
License Url: https://www.playstation.com/legal/important-information-about-the-ps-now-service/
Installer:
  Type: Exe
  Download Url: https://download-psnow.playstation.com/downloads/psnow/pc/PlayStationNow-11.1.2.exe
  SHA256: 660eaa555fd3376174c1ebccceb5008099c1da19a4ddfd1076436eefa67fedfe
No Homepage found within manifest
```

### Command Performed

```powershell
wingetdev show Wings3D.Wings3D --license
```

### Expectation

1. Display's the Mainfest for the app information in human readable format
1. No Homepage is found. States it did not find a homepage url for this manifest

### Result (Includes Command Output)

```powershell
PS C:\> wingetdev show Wings3D.Wings3D --license
Found Wings3D [Wings3D.Wings3D]
Version: 2.2.5
Publisher: Wings3D Team
Description: Wings 3D is an advanced subdivision modeler that is both powerful and easy to use.
Homepage: http://www.wings3d.com
License: BSD
Installer:
  Type: Nullsoft
  Download Url: https://sourceforge.net/projects/wings/files/wings/2.2.5/wings-x64-2.2.5.exe/download
  SHA256: d14f8b07c07d2bf6a6766a540be7a82b7e5b367c5ba88da54a5728c7f173a1c8
No License url found within manifest
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/507)


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -8,96 +8,100 @@
 
 namespace AppInstaller::CLI
 {
-    using namespace AppInstaller::CLI::Execution;
-    using namespace Settings;
+	using namespace AppInstaller::CLI::Execution;
```
JohnMcPMS, line 11: Your editor has changed spaces to tabs.  We use 4 spaces rather than a tab.  This is the case throughout your change (although auto-format hit this file really hard).

src/AppInstallerCLICore/Commands/ShowCommand.cpp
```diff
@@ -62,10 +66,22 @@ namespace AppInstaller::CLI
...
+                    Workflow::ShowManifestInfo <<
+                    Workflow::GetRequestedUrlFromManifest <<
+                    Workflow::OpenUrlInDefaultBrowser;
+            }
+            else {
+                context <<
+                    Workflow::GetManifest <<
+                    Workflow::SelectInstaller <<
+                    Workflow::ShowManifestInfo;
+            }
```
JohnMcPMS, line 84: Given that these two args add additional work, I would suggest a layout like:
```
context << existing_work;
if (has_new_args)
  context << do_more_work;

File: src/AppInstallerCLICore/Workflows/OpenUrlWorkflow.h [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+
+// ShellExecuteInstallerHandler handles installers run through ShellExecute.
+// Exe, Wix, Nullsoft, Msi and Inno should be handled by this installer handler.
+namespace AppInstaller::CLI::Workflow
+{
+    // Install is done through invoking ShellExecute on downloaded installer.
+    // Required Args: None
+    // Inputs: Manifest?, InstallerPath, InstallerArgs
+    // Outputs: None
+    void OpenUrlInDefaultBrowser(Execution::Context& context);
```
JohnMcPMS, line 16: >void OpenUrlInDefaultBrowser(Execution::Context& context); [](start = 4, length = 58)

The workflow organization is not super crisp.  I would have put this all in ShowFlow, as it is specific to show right now.

src/AppInstallerCLICore/Workflows/OpenUrlWorkflow.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+#include <AppInstallerProgress.h>
+#include "ExecutionContext.h"
+
+#include <filesystem>
+#include <optional>
+
+// ShellExecuteInstallerHandler handles installers run through ShellExecute.
+// Exe, Wix, Nullsoft, Msi and Inno should be handled by this installer handler.
```
JohnMcPMS, line 9: Comments need to be updated from copy.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -317,6 +317,15 @@ They can be configured through the settings file 'winget settings'.</value>
   <data name="ManifestValidationSuccess" xml:space="preserve">
     <value>Manifest validation succeeded.</value>
   </data>
+  <data name="ManifestHomepage" xml:space="preserve">
```
JohnMcPMS, line 320: >ManifestHomepage [](start = 14, length = 16)

The name should match the existing style: "HomepageArgumentDescription"

For value, I would suggest something more succinct:
Opens the homepage URL (if in manifest)

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -317,6 +317,15 @@ They can be configured through the settings file 'winget settings'.</value>
   <data name="ManifestValidationSuccess" xml:space="preserve">
     <value>Manifest validation succeeded.</value>
   </data>
+  <data name="ManifestHomepage" xml:space="preserve">
+    <value>If the Homepage Url exists it will be opened in your default browser</value>
+  </data>
+  <data name="ManifestLicense" xml:space="preserve">
+    <value>If the License Url exists it will be opened in your default browser</value>
+  </data>
+  <data name="ManifestPrivacy" xml:space="preserve">
```
JohnMcPMS, line 326: >ManifestPrivacy [](start = 14, length = 15)

Can hold off on this until we have it in the manifest.

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "ManifestComparator.h"
+#include "ShellExecuteInstallerHandler.h"
+
+using namespace AppInstaller::CLI;
+using namespace AppInstaller::Utility;
+using namespace AppInstaller::Manifest;
+
+namespace AppInstaller::CLI::Workflow
+{
+	void GetRequestedUrlFromManifest(Execution::Context& context)
```
JohnMcPMS, line 13: >void GetRequestedUrlFromManifest(Execution::Context& context) [](start = 1, length = 61)

[Get ready for my manifesto on the workflow and context object.]

The goal of the workflow design was to prevent the need to have these kinds of nested if statements.  Each workflow step can be seen as a transform (or transition in the state machine) that takes the previous state and produces a new one.  Any transform also has the option of failing the workflow, and thus future steps need not worry about the presence or correctness of the data (assuming that the developer properly constructed the flow).

So in the case of this function, we need to extract a string from the manifest.  Assuming that we have put this after the appropriate prior transforms (since show requires them, you have), we can be assured that there is already a single manifest stored in the context and can extract it directly.  So you should be able to start at line 24 (```const auto& manifest = context.Get<Execution::Data::Manifest>();```) and remove the outer scopes.

As for the context object, the data it stores is intended to be how information is passed from one transform to another.  This can be either because it is expensive to acquire, or to allow divergent paths to converge.  I don't see that this use case needs either; you can simply extract the appropriate string, then use it later in the same function.  If we have need to launch arbitrary URLs later, we can refactor the code back out.  In either case, a single Data entry should be used (ex. Data::LaunchURL) so that it can be generically applied.

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
```diff
@@ -0,0 +1,71 @@
...
+			else
+			{
+				if (context.Contains(Execution::Data::Manifest)) {
+					const auto& manifest = context.Get<Execution::Data::Manifest>();
+					Workflow::ManifestComparator manifestComparator(context.Args);
+					const auto selectedLocalization = manifestComparator.GetPreferredLocalization(manifest);
+					if (context.Args.Contains(Execution::Args::Type::Homepage)) {
+						if (selectedLocalization.Homepage.empty())
+						{
+							context.Reporter.Info() << "No Homepage url found within manifest" << std::endl;
```
JohnMcPMS, line 30: >"No Homepage url found within manifest" [](start = 34, length = 39)

Any string that is output to the Reporter should be a resource to allow for localization.  I haven't gotten quite far enough down the path to make that a compilation error, but I don't want to add any debt :)

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
```diff
@@ -0,0 +1,71 @@
...
+						}
+						else {
+							if (!selectedLocalization.Homepage.empty()) {
+								if (selectedLocalization.Homepage._Starts_with("http")) {
```
JohnMcPMS, line 34: >selectedLocalization.Homepage._Starts_with("http") [](start = 12, length = 50)

See Utility::IsUrlRemote, although this brings up a great point about security around this feature.  I do think we need to do a review to ensure that we have the appropriate checks to ensure that this is in fact a valid URL.  We do currently ensure that SmartScreen is OK with it before it can get into winget-pkgs, but the client should at least do as much as we can to ensure that it isn't some form of injection attack.

skyhoshi: i agree, having the security check here would add additional safeguard when private/local/other Repositories. 
one question though, does [Utility::IsUrlRemote] provide security or just do this check? 
do i need to provide this check? (keep in mind i haven't looked into this as of yet. thought it would be easier to ask)

JohnMcPMS: Utility::IsUrlRemote just does the 'starts with' but for "http://" or "https://" (and not using a non-std method).

I've started a discussion internally on the security implications here.  I think the most likely result will be in hardening the parsing logic to reject the strings before they reach this part of the code.  But when we have such a check we might put it here as well just to be sure.

In the meantime, I believe that there is a good chance that the feature will need to go behind a toggle to ensure that we don't create a security issue.  If you follow the example of ```EnsureFeatureEnabled(Settings::ExperimentalFeature::Feature::ExperimentalMSStore)``` you can add one fairly easily.

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
skyhoshi: Just got to this point, I've removed the outer scopes and as i read and understand you correctly your saying extract and make available one url. then in a later workflow use it. 
what I've done with this part of the code is allow you to call 
```
winget show git --homepage --license
```
which would launch BOTH url's 
I could convert this "Data::LaunchUrls" as a string array of some sort. and process that. 
OR for the matter of simplification and expectation remove the duality in the command. 
which would be advisable? 
1.  Data::LaunchUrls
2. Launch only one URL at a time per command.
I think the proper way is # 2 but I like the thought of having all the urls specified and available open as requested in one command.

JohnMcPMS: I wouldn't even really bother with putting it into the context, just extract and use it right there.  But if you did, then yes, collapse it to one data member (either a single string or a vector of them).

I don't mind whether it is one or multiple honestly; the user is in control of requesting it, and they won't really interfere with each other.

src/AppInstallerCLICore/Argument.cpp
skyhoshi: I'm going to leave this comment here, but for the other items, as I check them off I'm not going to add additional comments. I'm not sure your practice but I usually address each comment with an acknowledgement of the comment and the change. but since there's > 10 items I'm going to forego them and leave this one comment separate the src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp comment of the " Get URL's from Manifest" comment as it may not be clear what I've done to resolve it.  (and the fact that at this time, I haven't tested the changes 🔍 ) :)

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
skyhoshi: Breaking with my no comment from above :) I did add the Experimental flag to this feature as you requested. it's working on and off but my concern is that i may have missed something because the Feature and settings names were/are different. I attempted to use the same var names and the like but it appears one differs from the other. I've run a search and attempted to resolve this. it works but I would like a second pair of eyes on this new edit. Thanks

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
skyhoshi: I created a vector of string, and changed the Open method to just go over this list and execute. 
the concept is that additional methods can add to this list in the future and can open any number of urls.
Thanks for the hint of the vector :) 
the addition of the url's to the vector handles all the logic for parsing which ones to add so i removed the logic to parse which ones to show.

src/AppInstallerCLICore/Core.cpp
```diff
@@ -66,12 +66,12 @@ namespace AppInstaller::CLI
...
-                {
-                    strstr << " '" << arg << '\'';
-                }
-                return strstr.str();
+            std::stringstream strstr;
+            for (const auto& arg : utf8Args)
+            {
+                strstr << " '" << arg << '\'';
+            }
+            return strstr.str();
```
JohnMcPMS, line 74: Not sure why/how, but this change isn't needed.

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
```diff
@@ -0,0 +1,56 @@
...
+using namespace AppInstaller::Manifest;
+
+namespace AppInstaller::CLI::Workflow
+{
+    void GetRequestedUrlFromManifest(Execution::Context& context)
+    {
+        const auto& manifest = context.Get<Execution::Data::Manifest>();
+        Workflow::ManifestComparator manifestComparator(context.Args);
+        const auto selectedLocalization = manifestComparator.GetPreferredLocalization(manifest);
+        if (context.Args.Contains(Execution::Args::Type::Homepage)) {
```
JohnMcPMS, line 19: > { [](start = 67, length = 2)

More style things, we use braces on new lines except in rare one line cases:

if (condition)
{
}

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
```diff
@@ -0,0 +1,56 @@
...
+            if (selectedLocalization.Homepage.empty())
+            {
+                context.Reporter.Info() << Resource::String::UrlNotFound << std::endl;
+            }
+            else {
+                if (!selectedLocalization.Homepage.empty()) {
```
JohnMcPMS, line 25: Don't need to test again.

src/AppInstallerCLICore/Workflows/GetRequestedUrlFromManifest.cpp
```diff
@@ -0,0 +1,56 @@
...
+        if (context.Args.Contains(Execution::Args::Type::License)) {
+            if (selectedLocalization.LicenseUrl.empty())
+            {
+                context.Reporter.Info() << Resource::String::UrlNotFound << std::endl;
+            }
+            else {
+                if (!selectedLocalization.LicenseUrl.empty()) {
+                    if (Utility::IsUrlRemote(selectedLocalization.LicenseUrl)) {
+                        context.Reporter.Info() << Resource::String::ShowLicenseOpeningUrl << ": " << selectedLocalization.LicenseUrl << std::endl;
+                        context.Add<Execution::Data::ShowUrls>({ selectedLocalization.LicenseUrl.c_str() });
```
JohnMcPMS, line 46: >context.Add<Execution::Data::ShowUrls>({ selectedLocalization.LicenseUrl.c_str() }); [](start = 24, length = 84)

This second time will overwrite the first vector.  Would probably be easiest to always insert an empty vector at the beginning, then always add here:

At top:
context.Add<Execution::Data::ShowUrls>({});

When adding:
context.Get<Execution::Data::ShowUrls>().emplace_back(selectedLocalization.LicenseUrl);

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -350,6 +358,16 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
+    void OpenUrlInDefaultBrowser(Execution::Context& context)
+    {
+        if (context.Contains(Execution::Data::ShowUrls)) {
+            std::vector<std::string>& vectorOfUrls = context.Get<Execution::Data::ShowUrls>();
+            for (auto url : vectorOfUrls)
```
JohnMcPMS, line 365: >auto [](start = 17, length = 4)

const auto& (or auto const& if you want to get ahead of the curve of us old timers and our west const).

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -36,6 +38,8 @@ namespace AppInstaller::Settings
             return ExperimentalFeature{ "Argument Sample", "experimentalArg", "https://aka.ms/winget-settings", Feature::ExperimentalArg };
         case Feature::ExperimentalMSStore:
             return ExperimentalFeature{ "Microsoft Store Support", "experimentalMSStore", "https://aka.ms/winget-settings", Feature::ExperimentalMSStore };
+        case Feature::BrowseOrOpenRemoteUrl:
+            return ExperimentalFeature{ "Open Url", "experimentalBrowseOrOpenRemoteUrl", "https://aka.ms/winget-settings", Feature::BrowseOrOpenRemoteUrl };
```
JohnMcPMS, line 42: >experimentalBrowseOrOpenRemoteUrl [](start = 53, length = 33)

Can we just start naming these without experimental* at the front? My fault for not catching the last one. So just "browseOrOpenRemoteUrl".

Also this should have changes to /doc/Settings.md to describe it.

Tagging @yao-msft as I forgot to mention this in his PR for experimentalMSStore.

jsoref:
At some point you'll need to rebase onto master at which point the spell checker will ask you to add a couple of items to the `expect.txt` file

JohnMcPMS:
Closing due to Terminal's support of hyperlinks and the age of this PR.

PR #515, opened 2020-07-28
Update winget.resw

moolwalk:
Translation of InstallationDisclaimer1 was wrong and it is fixed.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/515)


JohnMcPMS:
Filed as 27986301 internally for update.  Thank you for the feedback.

JohnMcPMS:
The fix should be in #560 , thanks for your contribution 😄

Issue #331, opened 2020-05-25
Add hidden winget complete command to enable tab-completion


rkeithhill:
# Description of the new feature/enhancement

The `dotnet` cli tool from the .NET Core SDK has a hidden [complete command](https://github.com/dotnet/sdk/blob/b1223209644d900702287faea8e9b71f95ec49f8/src/Cli/dotnet/commands/dotnet-complete/CompleteCommand.cs) that works like this:

```
PS> dotnet complete --position 2 "re"
remove
restore
store
```

This can be hooked into environments like PowerShell to enable command/parameter completion for dotnet.exe e.g.:

```powershell
Register-ArgumentCompleter -Native -CommandName dotnet -ScriptBlock {
    param($wordToComplete, $commandAst, $cursorPosition)

    dotnet complete --position $cursorPosition $commandAst | ForEach-Object {
        [System.Management.Automation.CompletionResult]::new($_, $_, 'ParameterValue', $_)
    }
}
```

# Proposed technical implementation details (optional)

I would like to see winget implement a hidden `complete` command that would provide completions for top-level commands e.g.:  install, show, source, search, hash, validate.  And then for each command, it should provide parameter completions e.g. `winget install --id Git.Git --s` should return `--source` and `--silent`.

megamorf:
Relates to #146.

rkeithhill:
Closing issue as this has been released, see [the doc](https://github.com/microsoft/winget-cli/blob/master/doc/Completion.md) for more details.

Issue #146, opened 2020-05-15
Autocomplete [tab] winget commands in PowerShell


denelon:
I'd like to be able to use autocomplete for winget commands in PowerShell so I don't have to type them in their entirety.

# Proposed technical implementation details (optional)

rkeithhill:
You can do this yourself with PowerShell's `Register-ArgumentCompleter` command like so:
```powershell
$wingetCompleter = {
    param($wordToComplete, $commandAst, $cursorPosition)

    $tokens = $commandAst.Extent.Text.Trim() -split '\s+'
    $completions = switch ($tokens[1]) {
        'install' { "-m","--manifest","--name","--moniker","--version","--soruce","--exact","--interactive","--silent",
                    "--log","--override","--location","-v","-s","-e","-i","-h","-o","-"; break }
        'show'    { "--query","--manifest","--id","--name","--moniker","--version","--source","--exact","--versions",
                    "-q","-m","-v","-s","-e"; break }

        default { "install","show","source","search","hash","validation","--help","--info","-v","--version" }
    }

    $completions | Where {$_ -like "${wordToComplete}*"} | Sort-Object | ForEach-Object {
        [System.Management.Automation.CompletionResult]::new($_, $_, 'ParameterValue', $_)
    }
}

Register-ArgumentCompleter -Native -CommandName winget -ScriptBlock $wingetCompleter
```
This isn't fully fleshed out but could be done so fairly easily.  That said, rather than folks external to the is project having to provide the completions and then update those completions when new commands are added, it would be better to have winter return completions via a hidden `complete` command. This is what `dotnet` does which makes the arg completed very simple:
```powershell
Register-ArgumentCompleter -Native -CommandName dotnet -ScriptBlock {
    param($wordToComplete, $commandAst, $cursorPosition)

    dotnet complete --position $cursorPosition $commandAst | ForEach-Object {
        [System.Management.Automation.CompletionResult]::new($_, $_, 'ParameterValue', $_)
    }
}
```

JohnMcPMS:
Closing issue as this has been released, see [the doc](https://github.com/microsoft/winget-cli/blob/master/doc/Completion.md) for more details.

PR #521, opened 2020-07-30
Enable command line completion with complete command

JohnMcPMS:
This year as part of Microsoft's Hackathon, I decided I wanted to explore how to implement a `complete` command following the pattern of the `dotnet` CLI's complete command, as referenced in #331 and #146.

## Change
The goal of the change was to add the complete command, but it ended up having many side effects and becoming quite large. It may have also fixed a few bugs along the way. After I cover the complete command portion, I will detail the other side effects and bug fixes changes.

The MD file that I added to describe tab completion and the command can be seen [here](https://github.com/JohnMcPMS/winget-cli/blob/complete/doc/Completion.md). The basic premise is that we take in the current command line, cursor position, and "word" to be completed.

From this we construct a `CompletionData` object, which breaks the arguments into our standard command line processing structures. These are segregated into before and after the word to complete. We then create a new `Context` object, attach the `CompletionData` and use the same parsing methods as standard inputs to determine which command is being targeted. Finally, we pass the `Context` to the new `Command::Complete` method.

The `Complete` method breaks down the remaining `CompletionData` to generate three different potential sets of options:
1. A sub command :: The cursor is located just after a command and there are sub commands available.
2. An argument specifier :: The cursor is not positioned after an argument specifier that expects a value, and there are arguments available.
3. An argument value :: The cursor is positioned after an argument specifier that expects a value, or a positional argument is expected.

If an argument value is a potential option, a virtual function on the command is called to allow it to provide context sensitive results. It receives the `Context` created originally, which has had it's `Args` filled out during processing. Using this `Context`, existing `Workflow` functionality and paradigms can be leveraged to construct potential completions.

## Examples

> These examples assume that the tab completion in your shell works similar to PowerShell; repeated presses of tab (`⇥`) will result in cycling through the possible values.

Input | Result | Reason
--- | --- | ---
`winget ⇥` | `winget install` | `install` is the first command below the root
`winget sh⇥` | `winget show` | `show` is the first command that starts with `sh`
`winget source l⇥` | `winget source list` | `list` is the first sub-command of source that starts with `l`
`winget -⇥` | `winget --version` | `--version` is the first argument defined for the root
`winget install power⇥` | `winget install "Power Toys"` | `"Power Toys"` is the first package whose Id, Name, or Moniker starts with `power`
`winget install "Power Toys" --version ⇥` | `winget install "Power Toys" --version 0.19.2` | `0.19.2` is the highest version of Power Toys at the time of writing

## Other Changes
In addition to the `complete` command, these other changes are included:
- Refactored the `OutputStream` out of `Reporter`, which should result in more consistent `--no-vt` behavior.
- Added the concept of channels to `Reporter`, which enables us to select which type of output is allowed.
- Refactored the argument parsing to enable re-use for completion.
- Refactored `Reporter` to make it more of an instance; this meant making `ConsoleModeRestore` a singleton as it should have been.
- Added log cleanup functionality; currently deletes any log files older than 7 days when winget runs.
- Fixed `UTF8Substring` when count causes the substring to extend past the end of the input.
- Added an RAII type to disable telemetry while it exists, and use it while running `Command::Complete` to prevent any errors in common code from showing up as noise.
- Added a `StartsWith` match type to the search interface.

## Validation
Tests are added for `CompletionData` and `Command::Complete` handling of said data. Tests are also added for `StartsWith` searches and string comparisons.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/521)


doc/Completion.md
```diff
@@ -0,0 +1,58 @@
+﻿# WinGet Command Line Tab Completion
+
+WinGet offers a `complete` command that can be leveraged by your shell to provide context sensitive tab completion. It allows for completion of command names, argument names, and argument values, dependent on the current command line state.
+
+> Note, this feature has not yet been released. Hold tight, it is coming soon!
```
denelon, line 5: Are you going to be able to remember this line when it's released?

JohnMcPMS: That's the plan, but if not I'm sure someone will poke us.

src/AppInstallerCLICore/Commands/CompleteCommand.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "pch.h"
+#include "CompleteCommand.h"
+#include "RootCommand.h"
+#include "Resources.h"
+
+namespace AppInstaller::CLI
+{
+    using namespace std::string_view_literals;
+
+    using namespace Execution;
```
msftrubengu, line 12: nit

src/AppInstallerCLICore/Commands/CompleteCommand.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+#include "Command.h"
+
+namespace AppInstaller::CLI
+{
+    // Command to enable tab completion scenarios, including allowing them to
+    // be context sensitive in their data output.
+    struct CompleteCommand final : public Command
+    {
+        CompleteCommand(std::string_view parent) : Command("complete", parent, Visibility::Hidden) {}
```
msftrubengu, line 12: >CompleteCommand(std::string_view parent) : Command("complete", parent, Visibility::Hidden) {} [](start = 8, length = 93)

So we are not going to release it as an experimental feature?

JohnMcPMS: No, since you have to opt in by setting up your shell to call into this, and it is otherwise out of the way, it isn't really necessary.

---
In reply to: [463771454](https://github.com/Microsoft/pkg/pull/521#discussion_r463771454) [](ancestors = 463771454)

src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -20,40 +22,19 @@ namespace AppInstaller::CLI::Execution
...
-        // Strings explicitly declared as localization independent.
-        WINGET_CREATE_ISAPPROVEDFOROUTPUT_SPECIALIZATION(Utility::LocIndView);
-        WINGET_CREATE_ISAPPROVEDFOROUTPUT_SPECIALIZATION(Utility::LocIndString);
-        // Normalized strings come from user data and should therefore already by localized
-        // by how they are chosen (or there is no localized version).
-        WINGET_CREATE_ISAPPROVEDFOROUTPUT_SPECIALIZATION(Utility::NormalizedString);
-    }
-
     // Reporter should be the central place to show workflow status to user.
     // Todo: need to implement actual console output to show progress bar, etc
```
yao-msft, line 26: >// Todo: need to implement actual console output to show progress bar, etc [](start = 4, length = 74)

nit: this could be removed

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -127,29 +155,27 @@ namespace AppInstaller::Utility
...
+            throw std::out_of_range("UTF8Substring: offset past end of input");
+        }
+
+        size_t utf8Offset = itr.CurrentOffset();
         size_t utf8Count = 0;
-        size_t graphemeClusterOffset = 0;
-        int32_t i = 0;
 
-        while (i != UBRK_DONE)
+        // Count past end, convert to npos to get all of string
```
yao-msft, line 167: >Count past end [](start = 11, length = 14)

curious why do we not treat this as a bad input?

JohnMcPMS: For consistency with std::string::substr

---
In reply to: [463899310](https://github.com/Microsoft/pkg/pull/521#discussion_r463899310) [](ancestors = 463899310)

Wolf1098:
Maybe look at making it work with 

```
# Shows navigable menu for all module's and their options when hitting Tab

Set-PSReadlineKeyHandler -Key Tab -Function MenuComplete
```

PR #527, opened 2020-08-06
Search across multiple sources and add msstore source

yao-msft:
This is part 2 of #117

## Change
- Added support to search across multiple sources
- Added store source as one of the default sources if experimental feature is enabled
- Fixed a minor bug where removing a source will always restore default sources

## Validation
- Existing tests passed
- Added new tests to test multiple source search
- Manually tested the test msstore source and community source work together

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/527)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -11,29 +11,28 @@ namespace AppInstaller::CLI::Workflow
     bool InstallerComparator::operator() (const ManifestInstaller& installer1, const ManifestInstaller& installer2)
     {
         // Todo: Compare only architecture for now. Need more work and spec.
-        if (Utility::IsApplicableArchitecture(installer1.Arch) < Utility::IsApplicableArchitecture(installer2.Arch))
+        if (Utility::IsApplicableArchitecture(installer1.Arch) > Utility::IsApplicableArchitecture(installer2.Arch))
         {
-            return false;
+            return true;
         }
 
-        return true;
+        return false;
     }
 
     bool LocalizationComparator::operator() (const ManifestLocalization& loc1, const ManifestLocalization& loc2)
     {
-        UNREFERENCED_PARAMETER(loc2);
-
         // Todo: Compare simple language for now. Need more work and spec.
         std::string userPreferredLocale = std::locale("").name();
 
-        auto found = userPreferredLocale.find(loc1.Language);
+        auto foundLoc1 = userPreferredLocale.find(loc1.Language);
+        auto foundLoc2 = userPreferredLocale.find(loc2.Language);
 
-        if (found != std::string::npos)
+        if (foundLoc1 != std::string::npos && foundLoc2 == std::string::npos)
         {
-            return false;
+            return true;
         }
 
-        return true;
+        return false;
     }
 
     std::optional<Manifest::ManifestInstaller> ManifestComparator::GetPreferredInstaller(const Manifest::Manifest& manifest)
--- a/src/AppInstallerRepositoryCore/AggregatedSource.cpp
+++ b/src/AppInstallerRepositoryCore/AggregatedSource.cpp
@@ -1,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "AggregatedSource.h"
+
+namespace AppInstaller::Repository
+{
+    AggregatedSource::AggregatedSource()
+    {
+        m_details.Name = "AggregatedSource";
+        m_details.IsAggregated = true;
+    }
+
+    const SourceDetails& AppInstaller::Repository::AggregatedSource::GetDetails() const
+    {
+        return m_details;
+    }
+
+    void AggregatedSource::AddSource(std::shared_ptr<ISource> source)
+    {
+        m_sources.emplace_back(std::move(source));
+    }
+
+    SearchResult AggregatedSource::Search(const SearchRequest& request)
+    {
+        SearchResult result;
+
+        for (auto& source : m_sources)
+        {
+            auto oneSourceResult = source->Search(request);
+
+            for (auto& r : oneSourceResult.Matches)
+            {
+                r.SourceName = source->GetDetails().Name;
+                result.Matches.emplace_back(std::move(r));
+            }
+        }
+
+        SortResultMatches(result.Matches);
+
+        if (request.MaximumResults > 0 && result.Matches.size() > request.MaximumResults)
+        {
+            result.Truncated = true;
+            result.Matches.erase(result.Matches.begin() + request.MaximumResults, result.Matches.end());
+        }
+
+        return result;
+    }
+
+    void AggregatedSource::SortResultMatches(std::vector<ResultMatch>& matches)
+    {
+        // TODO: for now just simply prefer winget source.
+        struct ResultMatchComparator
+        {
+            bool operator() (
+                const ResultMatch& match1,
+                const ResultMatch& match2)
+            {
+                if (Utility::CaseInsensitiveEquals(match1.SourceName, s_Source_WingetCommunityDefault_Name) &&
+                    Utility::CaseInsensitiveEquals(match2.SourceName, s_Source_WingetCommunityDefault_Name))
+                {
+                    return true;
+                }
+
+                return false;
+            }
+        };
+
+        std::stable_sort(matches.begin(), matches.end(), ResultMatchComparator());
+    }
+}
+
--- a/src/AppInstallerRepositoryCore/AggregatedSource.cpp
+++ b/src/AppInstallerRepositoryCore/AggregatedSource.cpp
@@ -49,17 +49,37 @@ namespace AppInstaller::Repository
 
     void AggregatedSource::SortResultMatches(std::vector<ResultMatch>& matches)
     {
-        // TODO: for now just simply prefer winget source.
         struct ResultMatchComparator
         {
+            // The comparator compares the ResultMatch by MatchType first, then Field in a predefined order.
             bool operator() (
                 const ResultMatch& match1,
                 const ResultMatch& match2)
             {
-                if (Utility::CaseInsensitiveEquals(match1.SourceName, s_Source_WingetCommunityDefault_Name) &&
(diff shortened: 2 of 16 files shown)


src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -11,29 +11,28 @@ namespace AppInstaller::CLI::Workflow
...
         // Todo: Compare only architecture for now. Need more work and spec.
-        if (Utility::IsApplicableArchitecture(installer1.Arch) < Utility::IsApplicableArchitecture(installer2.Arch))
+        if (Utility::IsApplicableArchitecture(installer1.Arch) > Utility::IsApplicableArchitecture(installer2.Arch))
         {
-            return false;
+            return true;
         }
 
-        return true;
+        return false;
```
JohnMcPMS, line 19: >false [](start = 15, length = 5)

Why did you need to invert this?

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
```diff
@@ -14,6 +14,17 @@
...
+    using namespace std::string_view_literals;
+
+    constexpr std::string_view s_Source_WingetCommunityDefault_Name = "winget"sv;
+    constexpr std::string_view s_Source_WingetCommunityDefault_Arg = "https://winget.azureedge.net/cache"sv;
+    constexpr std::string_view s_Source_WingetCommunityDefault_Data = "Microsoft.Winget.Source_8wekyb3d8bbwe"sv;
+
+    constexpr std::string_view s_Source_WingetMSStoreDefault_Name = "msstore"sv;
+    // TODO: update to use prod endpoint
+    constexpr std::string_view s_Source_WingetMSStoreDefault_Arg = "https://winget-test.azureedge.net/msstore"sv;
+    constexpr std::string_view s_Source_WingetMSStoreDefault_Data = "Microsoft.Winget.MSStore.Source_8wekyb3d8bbwe"sv;
```
JohnMcPMS, line 26: This should not need to be in the header.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -190,6 +186,16 @@ namespace AppInstaller::Repository
                 details.Data = s_Source_WingetCommunityDefault_Data;
                 result.emplace_back(std::move(details));
             }
+
+            if (Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::ExperimentalMSStore))
```
JohnMcPMS, line 190: >if [](start = 12, length = 2)

Put the block inside the above one so it is more clear that they are part of the same case.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -512,12 +518,36 @@ namespace AppInstaller::Repository
...
+                for (auto& source : currentSources)
+                {
+                    AICLI_LOG(Repo, Info, << "Adding to aggregated source: " << source.Name);
+
+                    if (ShouldUpdateBeforeOpen(source))
+                    {
+                        UpdateSourceFromDetails(source, progress);
+                        sourceUpdated = true;
+                    }
+                    aggregatedSource->AddSource(std::move(CreateSourceFromDetails(source, progress)));
```
JohnMcPMS, line 541: >std::move [](start = 48, length = 9)

std::move isn't needed on an rvalue

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -512,12 +518,36 @@ namespace AppInstaller::Repository
...
+                auto aggregatedSource = std::make_shared<AggregatedSource>();
+
+                bool sourceUpdated = false;
+                for (auto& source : currentSources)
+                {
+                    AICLI_LOG(Repo, Info, << "Adding to aggregated source: " << source.Name);
+
+                    if (ShouldUpdateBeforeOpen(source))
+                    {
+                        UpdateSourceFromDetails(source, progress);
```
JohnMcPMS, line 538: >progress [](start = 56, length = 8)

At some point we are going to have to add a context callback to indicate "we are doing this action now".  This loop could potentially have the progress bar fill up 6+ times with just two sources.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
yao-msft: I want to make equal case return false.

---
In reply to: [467208009](https://github.com/microsoft/winget-cli/pull/527#discussion_r467208009) [](ancestors = 467208009)

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
yao-msft: The only reason I move them in the header is s_Source_WingetCommunityDefault_Name is used in AggregatedSource.cpp

Do you want me to move them all back to cpp and create a copy in AggregatedSource.cpp Or keep the *_Name constants here and move *_Arg and *_Data back to cpp? 

---
In reply to: [467209878](https://github.com/microsoft/winget-cli/pull/527#discussion_r467209878) [](ancestors = 467209878)

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
```diff
@@ -111,6 +111,9 @@ namespace AppInstaller::Repository
         // The highest order field on which the application matched the search.
         ApplicationMatchFilter MatchCriteria;
 
+        // The name of the source where the result is from. Used in aggregated source scenario.
+        std::string SourceName = {};
```
JohnMcPMS, line 115: >std::string SourceName = {}; [](start = 8, length = 28)

A lot of string duplication with this pattern.  The only solution I see to that is to have this or the IApplication just store a shared_ptr to the source itself.  Not necessarily in need of a change, but maybe some thought.

yao-msft: IApplication has a weak pointer to ISource already, then I thought the SearchResult should better be used without having to hold the ISource unless it needs to fetch manifest later.
I also considered making it string_view and store the list of source names somewhere, but it complicates the interface so this ends up as what I can think of here.


---
In reply to: [467338451](https://github.com/microsoft/winget-cli/pull/527#discussion_r467338451) [](ancestors = 467338451)

src/AppInstallerRepositoryCore/AggregatedSource.cpp
```diff
@@ -0,0 +1,72 @@
...
+            result.Truncated = true;
+            result.Matches.erase(result.Matches.begin() + request.MaximumResults, result.Matches.end());
+        }
+
+        return result;
+    }
+
+    void AggregatedSource::SortResultMatches(std::vector<ResultMatch>& matches)
+    {
+        // TODO: for now just simply prefer winget source.
```
JohnMcPMS, line 52: If the highest sort priority is source, then no sorting is required.  The sources should be left in their defined order, and so the results will already be good.

The current source already sorts the returned results, first by MatchType, then by Field.  The only other sort option that I have thought of beyond source first is to merge the results in this same sort priority: { MatchType, Field, Source }.  This should allow the results to better represent the search request.

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
JohnMcPMS: Move them back to the cpp and don't make a copy elsewhere because they shouldn't be needed.

---
In reply to: [467338302](https://github.com/Microsoft/pkg/pull/527#discussion_r467338302) [](ancestors = 467338302,467209878)

src/AppInstallerRepositoryCore/RepositorySource.cpp
yao-msft: Added todo

---
In reply to: [467334631](https://github.com/microsoft/winget-cli/pull/527#discussion_r467334631) [](ancestors = 467334631)

src/AppInstallerRepositoryCore/AggregatedSource.cpp
```diff
@@ -0,0 +1,92 @@
...
+            // The comparator compares the ResultMatch by MatchType first, then Field in a predefined order.
+            bool operator() (
+                const ResultMatch& match1,
+                const ResultMatch& match2)
+            {
+                std::vector<MatchType> MatchTypeOrder =
+                { MatchType::Exact, MatchType::CaseInsensitive, MatchType::StartsWith, MatchType::Fuzzy,
+                  MatchType::Substring, MatchType::FuzzySubstring, MatchType::Wildcard };
+
+                auto matchTypeOrder1 = std::find(MatchTypeOrder.begin(), MatchTypeOrder.end(), match1.MatchCriteria.Type);
```
JohnMcPMS, line 63: >auto matchTypeOrder1 = std::find(MatchTypeOrder.begin(), MatchTypeOrder.end(), match1.MatchCriteria.Type); [](start = 16, length = 106)

This is a very expensive comparator.  I would rather see a comment placed on these enums to state that they need to be in order of preference, then simply compare the integral values of the enums.

src/AppInstallerRepositoryCore/AggregatedSource.cpp
```diff
@@ -0,0 +1,76 @@
...
+        struct ResultMatchComparator
+        {
+            // The comparator compares the ResultMatch by MatchType first, then Field in a predefined order.
+            bool operator() (
+                const ResultMatch& match1,
+                const ResultMatch& match2)
+            {
+                if (match1.MatchCriteria.Type != match2.MatchCriteria.Type)
+                {
+                    return (static_cast<int>(match2.MatchCriteria.Type) - static_cast<int>(match1.MatchCriteria.Type)) > 0;
```
JohnMcPMS, line 61: >return (static_cast<int>(match2.MatchCriteria.Type) - static_cast<int>(match1.MatchCriteria.Type)) > 0; [](start = 20, length = 103)

```
return ToIntegral(match2.MatchCriteria.Type) > ToIntegral(match1.MatchCriteria.Type);
```
There is no need to do math on them, just direct compare.  The language utilities header has ToIntegral to convert the enum to its actual integral type.

PR #535, opened 2020-08-11
Replace yaml-cpp with libyaml

### LINKED ISSUE #312: winget should return a non-zero exit code if manifest validation fails
Currently, `winget validate` returns with exit code 0 even if the manifest validation fails. To make it easier to perform validation as part of a script (e.g. https://github.com/microsoft/winget-pkgs/pull/827), winget should allow to determine via its exit code whether validation succeeded or not.

JohnMcPMS:
Also fixes #312 , see below for details.

## Change
Due to certain bugs in yaml-cpp, it has not had significant fuzzing applied.  Even if those bugs were fixed, it is unknown how many more might exist behind it.  In order to avoid chasing those bugs, we are switching to a YAML parser that does have significant fuzz testing coverage, [libyaml](https://github.com/yaml/libyaml).

In addition to the use of libyaml, a C++ wrapper is written to create a near identical surface area for use by the existing manifest validation.  It also includes encoding detection to enable support for ANSI (assumed Windows-1252, which wasn't supported by yaml-cpp), UTF-8 (with and without BOM), UTF-16 LE/BE (both with and without BOM).

Finally, the `validate` command now returns a non-zero value when there are issues.  If there is a failure during validation, the value returned is `0x8A150029`.  If there are only warnings, the value returned is `0x8A150028`.

## Validation
In addition to the regression testing, new tests are added for the `validate` command and its return values.  New tests are also added for the various different text encoding types.

I also used scripts to run `validate` and `show` against every manifest in the winget-pkgs repo.  They all validate successfully, and the only difference in show presentation that I found was a single file that contained an ANSI © character which was being translated to unknown in the older parser, but is working properly with the new code.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/535)


src/YamlCppLib/readme.md
```diff
@@ -1,17 +1,17 @@
...
+    git subtree add --prefix=src/YamlCppLib/libyaml https://github.com/yaml/libyaml.git release/0.2.5 --squash
 
-The YamlCppLib.vcxproj and YamlCppLib.vcxproj.filters are created to make yaml-cpp code compiled as part of the AppInstallerClient solution.
+The YamlCppLib.vcxproj and YamlCppLib.vcxproj.filters are created to make libyaml code compiled as part of the WinGet solution.
 
 #### Steps used to create the VS project files.
 
-1. VS Create project from existing code wizard to pint to yaml-cpp directory. Use static library template.
+1. VS Create project from existing code wizard to point to libyaml directory. Use static library template.
 2. Removed code not needed by AppInstallerClient project. Basically keeping only files under src and include.
```
yao-msft, line 13: >AppInstallerClient [](start = 30, length = 18)

nit: forgot to update to WinGet here and below

yao-msft:
>                     // Syntax error, yaml-cpp error is stored in FailureInfo

libyaml?

---
Refers to: src/AppInstallerRepositoryCore/Manifest/ManifestValidation.h:89 in 7afa8a5. [](commit_id = 7afa8a5bd78a99e7750a74c60432a307b7cbd761, deletion_comment = False)

src/AppInstallerCLIE2ETests/AppInstallerCLIE2ETests.csproj
```diff
@@ -14,16 +14,16 @@
...
-  </ItemGroup>
-
   <ItemGroup>
     <Content Include="TestData\**">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="TestData\Manifests\TestWarningManifest.yaml" />
```
yao-msft, line 24: > <None Remove="TestData\Manifests\TestWarningManifest.yaml" /> [](start = 3, length = 62)

looks like not needed

src/AppInstallerCLIE2ETests/TestData/Manifests/TestWarningManifest.yaml
```diff
@@ -0,0 +1,12 @@
+Id: TestInvalid.Manifest
+Name: TestInvalidManifest
```
yao-msft, line 2: >Invalid [](start = 10, length = 7)

maybe Warning here

src/AppInstallerCommonCore/Public/winget/Yaml.h
```diff
@@ -0,0 +1,253 @@
...
+        // Adds a child node to the mapping.
+        template <typename... Args>
+        Node& AddMappingNode(Node&& key, Args&&... args)
+        {
+            Require(Type::Mapping);
+            return m_mapping->emplace(std::move(key), Node(std::forward<Args>(args)...))->second;
+        }
+
+        bool IsDefined() const { return m_type != Type::Invalid; }
+        bool IsNull() const { return m_type == Type::Invalid || m_type == Type::None || (m_type == Type::Scalar && m_scalar.empty()); }
```
yao-msft, line 96: >m_type == Type::Invalid [](start = 37, length = 23)

should we allow Invalid as null?

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -41,6 +41,17 @@
 #define APPINSTALLER_CLI_ERROR_EXPERIMENTAL_FEATURE_DISABLED    ((HRESULT)0x8A15001D)
 #define APPINSTALLER_CLI_ERROR_MSSTORE_INSTALL_FAILED           ((HRESULT)0x8A15001E)
 #define APPINSTALLER_CLI_ERROR_COMPLETE_INPUT_BAD               ((HRESULT)0x8A15001F)
+#define APPINSTALLER_CLI_ERROR_YAML_INIT_FAILED                 ((HRESULT)0x8A150020)
+#define APPINSTALLER_CLI_ERROR_INVALID_MAPPING_KEY              ((HRESULT)0x8A150021)
+#define APPINSTALLER_CLI_ERROR_DUPLICATE_MAPPING_KEY            ((HRESULT)0x8A150022)
```
yao-msft, line 46: maybe *_YAML_* for these 2 as well?

src/AppInstallerRepositoryCore/Manifest/YamlParser.cpp
```diff
@@ -197,13 +196,19 @@ namespace AppInstaller::Manifest
 
     std::vector<ValidationError> YamlParser::ParseManifest(const YAML::Node& rootNode, Manifest& manifest, bool fullValidation)
     {
+        // Detects empty files with a better error.
+        if (!rootNode.IsMap())
+        {
+            THROW_EXCEPTION_MSG(ManifestException(APPINSTALLER_CLI_ERROR_INVALID_MANIFEST), "The manifest does not contain a valid root.");
```
yao-msft, line 202: >APPINSTALLER_CLI_ERROR_INVALID_MANIFEST [](start = 50, length = 39)

There's an existing error code for manifest failure APPINSTALLER_CLI_ERROR_MANIFEST_FAILED, should we merge the 2?

src/AppInstallerCommonCore/pch.h
```diff
@@ -12,7 +12,8 @@
 
 #include "TraceLogging.h"
 
-#include <yaml-cpp/yaml.h>
+#define YAML_DECLARE_STATIC
```
yao-msft, line 15: >YAML_DECLARE_STATIC [](start = 8, length = 19)

Is this intended here?

src/AppInstallerCommonCore/YamlWrapper.cpp
```diff
@@ -0,0 +1,504 @@
...
+            return { mark.line + 1, mark.column + 1 };
+        }
+    }
+
+    Document::Document(bool init) :
+        m_token(init), m_document{}
+    {
+        if (init)
+        {
+            if (!yaml_document_initialize(&m_document, NULL, NULL, NULL, 1, 1))
```
yao-msft, line 85: >NULL, NULL, NULL, 1, 1 [](start = 55, length = 22)

maybe a comment on what these parameters mean?

src/AppInstallerCommonCore/Yaml.cpp
```diff
@@ -0,0 +1,478 @@
...
+        m_what = out.str();
+    }
+
+    Exception::Exception(Type type, const char* problem, size_t offset, int value) :
+        wil::ResultException(APPINSTALLER_CLI_ERROR_LIBYAML_ERROR)
+    {
+        std::ostringstream out;
+        OutputExceptionHeader(out, type);
+
+        out << (problem ? problem : "Unexplained error");
```
yao-msft, line 79: >"Unexplained error" [](start = 36, length = 19)

nit; I feel "Unknown error" is less confusing for me

src/AppInstallerCommonCore/Yaml.cpp
```diff
@@ -0,0 +1,478 @@
...
+        Require(Type::Mapping);
+        return m_mapping.value();
+    }
+
+    void Node::Require(Type type) const
+    {
+        THROW_HR_IF(APPINSTALLER_CLI_ERROR_YAML_INVALID_OPERATION, m_type != type);
+    }
+
+    void Node::Require(Type type1, Type type2) const
```
yao-msft, line 236: >Require [](start = 15, length = 7)

I think this is never used or not even possible?

src/AppInstallerCommonCore/Yaml.cpp
```diff
@@ -0,0 +1,478 @@
...
+    }
+
+    int64_t Node::as_dispatch(int64_t*) const
+    {
+        return std::stoll(m_scalar);
+    }
+
+    bool Node::as_dispatch(bool*) const
+    {
+        if (m_scalar == "true")
```
yao-msft, line 253: >m_scalar [](start = 12, length = 8)

I think yaml spec recommends case insensitive

src/AppInstallerCommonCore/Yaml.cpp
```diff
@@ -0,0 +1,478 @@
...
+    Emitter::Emitter() :
+        m_document(std::make_unique<Wrapper::Document>(true))
+    {
+        SetAllowedInputs<InputType::BeginMap, InputType::BeginSeq>();
+    }
+
+    Emitter::Emitter(Emitter&&) = default;
+    Emitter& Emitter::operator=(Emitter&&) = default;
+
+    Emitter::~Emitter() = default;
```
yao-msft, line 318: curious why not put these in the header?

src/AppInstallerCommonCore/Public/winget/Yaml.h
JohnMcPMS: This was to match the behavior of yaml-cpp with respect to what we generate from libyaml.

---
In reply to: [468973641](https://github.com/Microsoft/pkg/pull/535#discussion_r468973641) [](ancestors = 468973641)

src/AppInstallerRepositoryCore/Manifest/YamlParser.cpp
JohnMcPMS: That is the more generic parsing error.  It is rarely bad to have a more specific error.

---
In reply to: [468974896](https://github.com/Microsoft/pkg/pull/535#discussion_r468974896) [](ancestors = 468974896)

src/AppInstallerCommonCore/pch.h
JohnMcPMS: Yes.  I could put it in the project, but I don't see the value.  Here it is tied to the header it actually is related to.

---
In reply to: [468976197](https://github.com/Microsoft/pkg/pull/535#discussion_r468976197) [](ancestors = 468976197)

src/AppInstallerCommonCore/Yaml.cpp
yao-msft: After a second thought, "unexplained" seems better fit

---
In reply to: [468993477](https://github.com/microsoft/winget-cli/pull/535#discussion_r468993477) [](ancestors = 468993477)

src/AppInstallerCommonCore/Yaml.cpp
JohnMcPMS: It was used previously, but I had to change the code.  It is possible, it is simply !(type1 || type2).

---
In reply to: [468995615](https://github.com/Microsoft/pkg/pull/535#discussion_r468995615) [](ancestors = 468995615)

src/AppInstallerCommonCore/Yaml.cpp
JohnMcPMS: When I looked at it I didn't see it called out.  I can do case insensitive though.

---
In reply to: [468996962](https://github.com/Microsoft/pkg/pull/535#discussion_r468996962) [](ancestors = 468996962)

src/AppInstallerCommonCore/Yaml.cpp
JohnMcPMS: I'm using pImpl idioms here to remove the need for the libyaml header to be included in all of the consumers of our Yaml.h.  In order to do that, the std::unique_ptr<Emitter>'s destructor (and anything that requires the complete type) must only be referenced within the cpp.  So these functions are actually "implemented" here, rather than allowing them to be written out in every other TU and be folded by the compiler (or however they actually implement inline header functions).

---
In reply to: [468997684](https://github.com/Microsoft/pkg/pull/535#discussion_r468997684) [](ancestors = 468997684)

PR #562, opened 2020-09-11
Fix alignment of table output with extended characters

yao-msft:
## Change
We assumed 1 unit of terminal output per character in table output code while some extended characters take 2 units of space. This change uses icu's u_getIntPropertyValue to try to get character's space and align accordingly.

## Validation
Added unit tests.
Also validated manually that table aligns correctly with current msstore source.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/562)

### DIFF
--- a/src/AppInstallerCLICore/TableOutput.h
+++ b/src/AppInstallerCLICore/TableOutput.h
@@ -42,7 +42,7 @@ namespace AppInstaller::CLI::Execution
             for (size_t i = 0; i < FieldCount; ++i)
             {
                 m_columns[i].Name = std::move(header[i]);
-                m_columns[i].MinLength = Utility::UTF8TerminalLength(m_columns[i].Name.get());
+                m_columns[i].MinLength = Utility::UTF8ColumnWidth(m_columns[i].Name.get());
                 m_columns[i].MaxLength = 0;
             }
         }
@@ -93,7 +93,7 @@ namespace AppInstaller::CLI::Execution
             {
                 for (size_t i = 0; i < FieldCount; ++i)
                 {
-                    m_columns[i].MaxLength = std::max(m_columns[i].MaxLength, Utility::UTF8TerminalLength(line[i]));
+                    m_columns[i].MaxLength = std::max(m_columns[i].MaxLength, Utility::UTF8ColumnWidth(line[i]));
                 }
             }
 
@@ -189,20 +189,18 @@ namespace AppInstaller::CLI::Execution
 
                 if (col.MaxLength)
                 {
-                    size_t valueLength = Utility::UTF8TerminalLength(line[i]);
+                    size_t valueLength = Utility::UTF8ColumnWidth(line[i]);
 
                     if (valueLength > col.MaxLength)
                     {
-                        std::string_view trimmedOutput = Utility::UTF8TrimRightToTerminalLength(line[i], col.MaxLength - 1);
-                        out << trimmedOutput;
+                        size_t actualWidth;
+                        out << Utility::UTF8TrimRightToColumnWidth(line[i], col.MaxLength - 1, actualWidth) << "\xE2\x80\xA6"; // UTF8 encoding of ellipsis (…) character
 
-                        // Since some characters take 2 unit space, the trimmed string length might be 1 less than the expected length.
-                        if (Utility::UTF8TerminalLength(trimmedOutput) != col.MaxLength - 1)
+                        // Some characters take 2 unit space, the trimmed string length might be 1 less than the expected length.
+                        if (actualWidth != col.MaxLength - 1)
                         {
                             out << ' ';
                         }
-
-                        out << "\xE2\x80\xA6"; // UTF8 encoding of ellipsis (…) character
 
                         if (col.SpaceAfter)
                         {
--- a/src/AppInstallerCLICore/TableOutput.h
+++ b/src/AppInstallerCLICore/TableOutput.h
@@ -42,7 +42,7 @@ namespace AppInstaller::CLI::Execution
             for (size_t i = 0; i < FieldCount; ++i)
             {
                 m_columns[i].Name = std::move(header[i]);
-                m_columns[i].MinLength = Utility::UTF8Length(m_columns[i].Name.get());
+                m_columns[i].MinLength = Utility::UTF8TerminalLength(m_columns[i].Name.get());
                 m_columns[i].MaxLength = 0;
             }
         }
@@ -93,7 +93,7 @@ namespace AppInstaller::CLI::Execution
             {
                 for (size_t i = 0; i < FieldCount; ++i)
                 {
-                    m_columns[i].MaxLength = std::max(m_columns[i].MaxLength, Utility::UTF8Length(line[i]));
+                    m_columns[i].MaxLength = std::max(m_columns[i].MaxLength, Utility::UTF8TerminalLength(line[i]));
                 }
             }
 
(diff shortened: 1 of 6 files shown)


src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -137,6 +137,12 @@ namespace AppInstaller::Utility
         return result;
     }
 
+    std::u32string ConvertToUTF32(std::string_view input)
+    {
+        std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> convU32;
```
JohnMcPMS, line 142: >std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> [](start = 8, length = 59)

This is deprecated in C++17.  You will either need to find another way to convert to UTF-32, or use the break iterator rather than converting the entire string.

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -151,6 +157,21 @@ namespace AppInstaller::Utility
         return numGraphemeClusters;
     }
 
+    size_t UTF8TerminalLength(std::string_view input)
```
JohnMcPMS, line 160: >std::string_view input [](start = 30, length = 22)

I think you might have to take in a NormalizedString to guarantee that this will work properly.  If the individual codepoints don't map to a grapheme cluster 1:1, I'm not sure what the width will report.  It is possible that there is a 0 width return value, but you don't have that represented here.

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -177,6 +198,29 @@ namespace AppInstaller::Utility
         return input.substr(utf8Offset, utf8Count);
     }
 
+    std::string_view UTF8TrimRightToTerminalLength(std::string_view input, size_t length)
+    {
+        THROW_HR_IF(E_INVALIDARG, length < 0);
```
JohnMcPMS, line 203: >length < 0 [](start = 34, length = 10)

size_t is unsigned.

src/AppInstallerCommonCore/AppInstallerStrings.cpp
JohnMcPMS, line 160: >UTF8TerminalLength [](start = 11, length = 18)

I would suggest calling it "ColumnWidth" rather than "TerminalLength".

src/AppInstallerCommonCore/AppInstallerCommonCore.vcxproj
```diff
@@ -153,7 +153,7 @@
       <PrecompiledHeaderOutputFile>$(IntDir)pch.pch</PrecompiledHeaderOutputFile>
       <PreprocessorDefinitions>_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <WarningLevel>Level4</WarningLevel>
-      <AdditionalOptions>%(AdditionalOptions) /permissive- /D _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING</AdditionalOptions>
+      <AdditionalOptions>%(AdditionalOptions) /permissive- /D _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING /D _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING</AdditionalOptions>
```
JohnMcPMS, line 156: >_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING [](start = 120, length = 49)

And apparently you were aware of the deprecation...

src/AppInstallerCLICore/TableOutput.h
```diff
@@ -189,11 +189,19 @@ namespace AppInstaller::CLI::Execution
...
                     {
-                        out << Utility::UTF8Substring(line[i], 0, col.MaxLength - 1);
+                        std::string_view trimmedOutput = Utility::UTF8TrimRightToTerminalLength(line[i], col.MaxLength - 1);
+                        out << trimmedOutput;
+
+                        // Since some characters take 2 unit space, the trimmed string length might be 1 less than the expected length.
+                        if (Utility::UTF8TerminalLength(trimmedOutput) != col.MaxLength - 1)
+                        {
+                            out << ' ';
+                        }
```
JohnMcPMS, line 203: This should happen after the ellipsis, and it would be great if UTF8TrimRightToTerminalLength could tell us the actual trimmed column width rather than recalculating it (out param?).

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerCommonCore/AppInstallerStrings.cpp
yao-msft: I changed to take NormalizedString but the '...' character which takes 1 column width becomes 3 dots '.''.''.' after normalization.
So I kept string_view here and instead using NormalizedString within the function using NormalizationC (instead of the default NormalizationKC)

---
In reply to: [486834016](https://github.com/microsoft/winget-cli/pull/562#discussion_r486834016) [](ancestors = 486834016)

src/AppInstallerCommonCore/AppInstallerCommonCore.vcxproj
yao-msft: Yeah, this is the cleanest I can find to convert to utf32, anyway, I used break iterator instead.

---
In reply to: [486835987](https://github.com/microsoft/winget-cli/pull/562#discussion_r486835987) [](ancestors = 486835987)

src/AppInstallerCommonCore/pch.h
```diff
@@ -17,6 +17,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <codecvt>
```
JohnMcPMS, line 20: Remove deprecated header

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -79,6 +79,14 @@ namespace AppInstaller::Utility
                 return m_currentBrk;
             }
 
+            UChar32 CodePointAt(int32_t offset)
```
JohnMcPMS, line 82: >CodePointAt [](start = 20, length = 11)

Just from an API design standpoint, I would think it better for this to be `UChar32 CurrentCodePoint() const` and to have a free function `UChar32 CodePointAt(std::string_view, int32_t)` [you don't actually need this second one, I'm just pointing out how I would expect it to be designed]

src/AppInstallerCommonCore/AppInstallerStrings.cpp
JohnMcPMS: Can you put tests in for the ligature fi as mentioned on [https://docs.microsoft.com/en-us/windows/win32/api/winnls/ne-winnls-norm_form](https://docs.microsoft.com/en-us/windows/win32/api/winnls/ne-winnls-norm_form)

I think that by using KC first, we have already destructed it and C won't put it back together.  Overall C should work here I think, but you might have to abandon it based on another comment.

---
In reply to: [487358887](https://github.com/microsoft/winget-cli/pull/562#discussion_r487358887) [](ancestors = 487358887,486834016)

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -151,6 +157,28 @@ namespace AppInstaller::Utility
...
+    {
+        ICUBreakIterator itr{ input, UBRK_CHARACTER };
+
+        size_t columnWidth = 0;
+        UChar32 currentCP = 0;
+        int32_t nextBrk = 0;
+
+        currentCP = itr.CurrentCodePoint();
+        nextBrk = itr.Next();
+        while (nextBrk != UBRK_DONE && currentCP != U_SENTINEL)
```
JohnMcPMS, line 170: >nextBrk [](start = 15, length = 7)

Could drop nextBrk and just use itr.Next() here

JohnMcPMS (approved):
:shipit:

src/AppInstallerCommonCore/AppInstallerStrings.cpp
yao-msft: I blame myself copy pasting code ...

---
In reply to: [488238193](https://github.com/microsoft/winget-cli/pull/562#discussion_r488238193) [](ancestors = 488238193)

PR #564, opened 2020-09-11
Create DB 1.1 schema for system reference strings

JohnMcPMS:
## Change
The primary motivation of this change was to add new fields for use in the list and upgrade commands; system reference strings that can be used to associate a winget package with software currently installed on a machine.  This required creation of schema 1.1, as well as some refactoring to make inheritance and behavior changes easier between versions.

The two types of strings added are PackageFamilyName for MSIX packages, and ProductCode for packages that register through ARP (Add/Remove Programs).  These are added with the option of having multiple of each in the event that a package's installers have unique values.  The values are stored with their casing folded to allow for ordinal comparisons, a performance requirement due to future work on list and upgrade.

Additional noteworthy changes:
1. Changed from implicit indices to explicit ones in 1.1, and drop more of them during preparation for packaging. This should result in a ~35% reduction in page count in the database, with no impact to performance.
2. Added a dev debug tool in SQLiteWrapper.cpp; set WINGET_SQLITE_EXPLAIN_QUERY_PLAN_ENABLED to 1 to have the query plan for every SQLite prepared statement output to logging.

## Validation
The existing tests have been updated to use GENERATE to run the tests not only against all versions, but also with backcompat scenarios.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/564)

### DIFF
--- a/src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
+++ b/src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
@@ -66,6 +66,12 @@ namespace AppInstaller::Manifest
         // Store Product Id
         string_t ProductId;
 
+        // Package family name for MSIX type only.
+        string_t PackageFamilyName;
+
+        // Product code for non-MSIX only.
+        string_t ProductCode;
+
         // If present, has more precedence than root
         InstallerTypeEnum InstallerType;
 
--- a/src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
+++ b/src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
@@ -66,10 +66,10 @@ namespace AppInstaller::Manifest
         // Store Product Id
         string_t ProductId;
 
-        // Package family name for MSIX type only.
+        // Package family name for MSIX packaged installers.
         string_t PackageFamilyName;
 
-        // Product code for non-MSIX only.
+        // Product code for ARP (Add/Remove Programs) installers.
         string_t ProductCode;
 
         // If present, has more precedence than root
--- a/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
+++ b/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
@@ -51,9 +51,11 @@ namespace AppInstaller::Repository::Microsoft
         // Gets the schema version of the index.
         Schema::Version GetVersion() const { return m_version; }
 
+#ifndef AICLI_DISABLE_TEST_HOOKS
         // Changes the version of the interface being used to operate on the database.
         // Should only be used for testing.
         void ForceVersion(const Schema::Version& version);
+#endif
 
         // Gets the last write time for the index.
         std::chrono::system_clock::time_point GetLastWriteTime();
--- a/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
+++ b/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
@@ -51,6 +51,10 @@ namespace AppInstaller::Repository::Microsoft
         // Gets the schema version of the index.
         Schema::Version GetVersion() const { return m_version; }
 
+        // Changes the version of the interface being used to operate on the database.
+        // Should only be used for testing.
+        void ForceVersion(const Schema::Version& version);
+
         // Gets the last write time for the index.
         std::chrono::system_clock::time_point GetLastWriteTime();
 
--- a/src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/SearchResultsTable_1_0.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/SearchResultsTable_1_0.cpp
@@ -38,27 +38,6 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_0
             return (match != MatchType::Exact);
         }
 
-        int BuildSearchStatement(SQLite::Builder::StatementBuilder& builder, ApplicationMatchField field, MatchType match)
-        {
-            bool useLike = MatchUsesLike(match);
-
-            switch (field)
-            {
-            case ApplicationMatchField::Id:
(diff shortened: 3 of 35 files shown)


src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/Interface.h
```diff
@@ -0,0 +1,26 @@
...
+#include "Microsoft/Schema/ISQLiteIndex.h"
+#include "Microsoft/Schema/1_0/Interface.h"
+
+
+namespace AppInstaller::Repository::Microsoft::Schema::V1_1
+{
+    // Interface to this schema version exposed through ISQLiteIndex.
+    struct Interface : public V1_0::Interface
+    {
+        // Version 1.0
```
msftrubengu, line 13: >1.0 [](start = 19, length = 3)

1.1?

msftrubengu (approved):
:shipit:

src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/Interface.h
JohnMcPMS: No, this 1.0 refers to the interface version where the functions come from.  These are 1.0 functions.  If 1.1 added new accessor functions, they would be in a 1.1 section.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/ProductCodeTable.h
```diff
@@ -0,0 +1,22 @@
...
+        using namespace std::string_view_literals;
+
+        struct ProductCodeTableInfo
+        {
+            inline static constexpr std::string_view TableName() { return "productcodes"sv; }
+            inline static constexpr std::string_view ValueName() { return "productcode"sv; }
+        };
+    }
+
+    // The table for PackageFamilyName.
```
yao-msft, line 20: >PackageFamilyName [](start = 21, length = 17)

Nit: wrong description

src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/Interface_1_1.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    ISQLiteIndex::SearchResult Interface::Search(SQLite::Connection& connection, const SearchRequest& request)
+    {
+        // Update any system reference strings to be folded
+        SearchRequest foldedRequest = request;
+
+        auto foldIfNeeded = [](ApplicationMatchFilter& filter)
+        {
+            if ((filter.Field == ApplicationMatchField::PackageFamilyName || filter.Field == ApplicationMatchField::ProductCode) &&
+                filter.Type == MatchType::Exact)
```
yao-msft, line 184: >filter.Type == MatchType::Exact [](start = 16, length = 31)

Should we just fold regardless of Matchtype for these 2 fields?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/SearchResultsTable_1_0.cpp
```diff
@@ -157,12 +136,18 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_0
         From().BeginParenthetical();
 
         // Add the field specific portion
-        int bindIndex = BuildSearchStatement(builder, field, match);
+        std::optional<int> bindIndex = BuildSearchStatement(builder, field, match);
+
+        if (!bindIndex)
+        {
+            AICLI_LOG(Repo, Verbose, << "ApplicationMatchField not supported in this version: " << ApplicationMatchFieldToString(field));
```
yao-msft, line 143: >"ApplicationMatchField not supported in this version: " [](start = 40, length = 55)

Do we want to surface this back to user in some form?

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
```diff
@@ -51,6 +51,10 @@ namespace AppInstaller::Repository::Microsoft
         // Gets the schema version of the index.
         Schema::Version GetVersion() const { return m_version; }
 
+        // Changes the version of the interface being used to operate on the database.
+        // Should only be used for testing.
+        void ForceVersion(const Schema::Version& version);
```
yao-msft, line 56: >void ForceVersion(const Schema::Version& version); [](start = 8, length = 50)

should we put it inside existing #def DISABLE***TEST_HOOK since it's test only?

src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
```diff
@@ -66,6 +66,12 @@ namespace AppInstaller::Manifest
         // Store Product Id
         string_t ProductId;
 
+        // Package family name for MSIX type only.
```
yao-msft, line 69: >MSIX  [](start = 35, length = 5)

or Store type?

yao-msft (approved):
:shipit:

src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/Interface_1_1.cpp
JohnMcPMS: The other MatchTypes use like for comparison, and so get folded as part of that.  While it wouldn't hurt anything, it isn't necessary.

---
In reply to: [488313354](https://github.com/microsoft/winget-cli/pull/564#discussion_r488313354) [](ancestors = 488313354)

src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/SearchResultsTable_1_0.cpp
JohnMcPMS: No, this is how a 1.0 database will avoid attempting to search for the new fields.  It is an intended short circuit, but is logged for informational purposes.

---
In reply to: [488314712](https://github.com/microsoft/winget-cli/pull/564#discussion_r488314712) [](ancestors = 488314712)

denelon:
Related to #1243

PR #570, opened 2020-09-16
Put milestones in the right order

luzeagithub:
There was a milestone missing (the /milestone/22 one) and because of that, the links were redirecting to the wrong milestone.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/570)


denelon (approved):
Thanks for catching my mistake 👍

PR #574, opened 2020-09-17
Add system reference values to manifest parsing

JohnMcPMS:
## Change
Adds the PackageFamilyName and ProductCode values to parsing by the manifest.  These can be at the root level, where they will be inherited.  They can also be provided per installer, but they must match the InstallerType or it will be an error.

Also makes major changes to the way the ManifestVersion is handled.  For now, the version is largely ignored (except it must be major version 0 if provided).  A concept of extensions is added, which are defined as the hyphen separated values after the primary version.  These can also be versioned, although the code does not exist to check it.  The syntax is:
```
ManifestVersion := Version (- Extension)
Extension := Version (- Extension)
```
where the Version portion of Extension is treated as `Name.Version`.  For instance, `1.0.0-extension` and `1.0.0-extension.2`.

Additionally:
- Fixes an issue with the version sorting where additional version characters would sort higher than those without (ex. 1.0 will now be greater than 1.0-alpha)

## Validation
Tests are added for PackageFamilyName and ProductCode, as well as the manifest version parsing.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/574)

### DIFF
--- a/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
+++ b/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
@@ -409,6 +409,7 @@
     <CopyFileToFolders Include="TestData\Manifest-Encoding-UTF8.yaml">
       <DeploymentContent>true</DeploymentContent>
     </CopyFileToFolders>
+    <None Include="TestData\Manifest-Good-SystemReferenceComplex.yaml" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\AppInstallerCLICore\AppInstallerCLICore.vcxproj">
--- a/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
+++ b/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
@@ -409,7 +409,18 @@
     <CopyFileToFolders Include="TestData\Manifest-Encoding-UTF8.yaml">
       <DeploymentContent>true</DeploymentContent>
     </CopyFileToFolders>
-    <None Include="TestData\Manifest-Good-SystemReferenceComplex.yaml" />
+    <CopyFileToFolders Include="TestData\Manifest-Bad-PackageFamilyNameOnMSI.yaml">
+      <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent>
+      <FileType>Document</FileType>
+    </CopyFileToFolders>
+    <CopyFileToFolders Include="TestData\Manifest-Bad-ProductCodeOnMSIX.yaml">
+      <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent>
+      <FileType>Document</FileType>
+    </CopyFileToFolders>
+    <CopyFileToFolders Include="TestData\Manifest-Good-SystemReferenceComplex.yaml">
+      <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent>
+      <FileType>Document</FileType>
+    </CopyFileToFolders>
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\AppInstallerCLICore\AppInstallerCLICore.vcxproj">
--- a/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
+++ b/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
@@ -410,15 +410,15 @@
       <DeploymentContent>true</DeploymentContent>
     </CopyFileToFolders>
     <CopyFileToFolders Include="TestData\Manifest-Bad-PackageFamilyNameOnMSI.yaml">
-      <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent>
+      <DeploymentContent>true</DeploymentContent>
       <FileType>Document</FileType>
     </CopyFileToFolders>
     <CopyFileToFolders Include="TestData\Manifest-Bad-ProductCodeOnMSIX.yaml">
-      <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent>
+      <DeploymentContent>true</DeploymentContent>
       <FileType>Document</FileType>
     </CopyFileToFolders>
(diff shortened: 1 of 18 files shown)


src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
```diff
@@ -409,6 +409,18 @@
     <CopyFileToFolders Include="TestData\Manifest-Encoding-UTF8.yaml">
       <DeploymentContent>true</DeploymentContent>
     </CopyFileToFolders>
+    <CopyFileToFolders Include="TestData\Manifest-Bad-PackageFamilyNameOnMSI.yaml">
+      <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent>
```
yao-msft, line 413: ><DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent> [](start = 6, length = 99)

These should apply to all. And better keep them consistent with all above

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -226,7 +226,9 @@ TEST_CASE("ReadBadManifests", "[ManifestValidation]")
         { "Manifest-Bad-VersionInvalid.yaml", "Invalid field value. Field: Version" },
         { "Manifest-Bad-VersionMissing.yaml", "Required field missing. Field: Version" },
         { "Manifest-Bad-InvalidManifestVersionValue.yaml", "Invalid field value. Field: ManifestVersion" },
-        { "InstallFlowTest_MSStore.yaml", "Field value is not supported. Field: InstallerType Value: MSStore" },
```
yao-msft, line 229: >Field value is not supported. Field: InstallerType Value: MSStore [](start = 43, length = 65)

Is it intentional to remove this error message? I think it's good and informational to manifest authors

yao-msft:
> TEST_CASE("ReadGoodManifestAndVerifyContents", "[ManifestValidation]")

This test was supposed to a complete content verification on a complete manifest. Since you updated Manifest_Good.yaml, maybe check the 2 new values here too

---
Refers to: src/AppInstallerCLITests/YamlManifest.cpp:31 in 97ba58f. [](commit_id = 97ba58f41035aa056a65802af9cc1f792d985c1e, deletion_comment = False)

src/AppInstallerCommonCore/Manifest/Manifest.cpp
```diff
@@ -6,20 +6,53 @@
...
+                if (!m_parts[i].Other.empty())
+                {
+                    validationSuccess = false;
+                    break;
+                }
+            }
+
+            for (const Version& ext : m_extensions)
+            {
+                if (ext.GetParts().empty() || ext.GetParts()[0].Integer != 0)
```
yao-msft, line 55: >[0].Integer != 0 [](start = 60, length = 16)

Are we going to do anything about the remaining parts in extensions?

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
```diff
@@ -260,6 +253,28 @@ namespace AppInstaller::Manifest
...
+            // Validate system reference strings if they are set at the installer level
+            if (fullValidation && !installer.PackageFamilyName.empty() && !ManifestInstaller::DoesInstallerTypeUsePackageFamilyName(installer.InstallerType))
+            {
+                resultErrors.emplace_back(ManifestError::InstallerTypeDoesNotSupportPackageFamilyName, "InstallerType", ManifestInstaller::InstallerTypeToString(installer.InstallerType));
+            }
+
+            if (fullValidation && !installer.ProductCode.empty() && !ManifestInstaller::DoesInstallerTypeUseProductCode(installer.InstallerType))
+            {
+                resultErrors.emplace_back(ManifestError::InstallerTypeDoesNotSupportProductCode, "InstallerType", ManifestInstaller::InstallerTypeToString(installer.InstallerType));
+            }
```
yao-msft, line 265: In my previous change to separate manifest population vs manifest semantic validation, these checks fit in the ValidateManifest() in manifestvalidation.cpp

yao-msft:
Since we are not checking minor versions any more, do you expect us to do strict minor version after 1.0? Or minor version is no longer used as all?

JohnMcPMS:
I don't see the value in being strict about minor version.  It seems to me to only serve to increase complexity for users adding manifests without adding value.  If I'm missing something, I am happy to become more learned, but I currently don't see any point now or in the future.

---
In reply to: [694644474](https://github.com/microsoft/winget-cli/pull/574#issuecomment-694644474) [](ancestors = 694644474)

src/AppInstallerCLITests/YamlManifest.cpp
JohnMcPMS: I believe the problem comes from the failure in ManifestVersion completely blocking all other validation.  I don't think we can proceed if we don't like the version.  But I can drop the fullValidation from ManifestVer and allow it to get back to the original error here.

---
In reply to: [490678974](https://github.com/microsoft/winget-cli/pull/574#discussion_r490678974) [](ancestors = 490678974)

src/AppInstallerCommonCore/Manifest/Manifest.cpp
JohnMcPMS: What do you mean "do anything"?  Validate?  I don't see any value to doing so.  Again, happy to be shown a reason and change it.

---
In reply to: [490692976](https://github.com/microsoft/winget-cli/pull/574#discussion_r490692976) [](ancestors = 490692976)

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
JohnMcPMS: Agreed, *but* this check can only be performed in this way here.  But I can go the stricter route and move them.

---
In reply to: [490695526](https://github.com/microsoft/winget-cli/pull/574#discussion_r490695526) [](ancestors = 490695526)

JohnMcPMS: Actually, I was wrong, and simply moving the validation parts will have the same net effect.

---
In reply to: [491164478](https://github.com/microsoft/winget-cli/pull/574#discussion_r491164478) [](ancestors = 491164478,490695526)

yao-msft (approved):
:shipit:

PR #575, opened 2020-09-20
Refactor source and search functionality

JohnMcPMS:
## Motivation
This change is being made to support future work for `list`, `update`, and `uninstall`.  It should have 0 functional impact at this time.  It also renames the existing classes to be more in line with the goals, changing from the work Application to Package.  Finally, more const'ness is imposed.

This change was made separate from any new use to create less churn when the actual functional changes are included.

## Change
Starting with `ISource`, a unique identifier is added to enable tying installs to them regardless of the name given to the source by the user.  `Search` is not changed at face value, but the result struct now returns a collection of `IPackage`s rather than the previous `IApplication`.  `IPackage` is a collection of `IPackageVersion`s in different states.  Currently that is [0, 1] possible installed versions, and [0, N] available versions.

`IPackageVersion` contains functionality to get individual properties of the specific version, as well as the manifest.

All other changes fall out of the above.

## Validation
Existing regression tests cover all affected code.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/575)

### DIFF
--- a/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
@@ -37,11 +37,11 @@ namespace AppInstaller::Repository::Microsoft
                 SourceReference(source), m_manifestId(manifestId) {}
 
             // Inherited via IPackageVersion
-            Utility::LocIndString GetProperty(PackageProperty property) const override
+            Utility::LocIndString GetProperty(PackageVersionProperty property) const override
             {
                 switch (property)
                 {
-                case PackageProperty::SourceId:
+                case PackageVersionProperty::SourceId:
                     return LocIndString{ GetSource()->GetIdentifier() };
                 default:
                     // Values coming from the index will always be localized/independent.
@@ -52,7 +52,7 @@ namespace AppInstaller::Repository::Microsoft
             std::optional<Manifest::Manifest> GetManifest() const override
             {
                 std::shared_ptr<const SQLiteIndexSource> source = GetSource();
-                std::optional<std::string> relativePathOpt = source->GetIndex().GetPropertyByManifestId(m_manifestId, PackageProperty::RelativePath);
+                std::optional<std::string> relativePathOpt = source->GetIndex().GetPropertyByManifestId(m_manifestId, PackageVersionProperty::RelativePath);
                 if (!relativePathOpt)
                 {
                     return {};
--- a/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
@@ -12,36 +12,52 @@ namespace AppInstaller::Repository::Microsoft
 
     namespace
     {
-        // The IApplication impl for SQLiteIndexSource.
-        struct Application : public IApplication
+        // The base for the package objects.
+        struct SourceReference
         {
-            Application(std::shared_ptr<SQLiteIndexSource>& source, SQLiteIndex::IdType id) :
-                m_id(id), m_source(source) {}
+            SourceReference(const std::shared_ptr<const SQLiteIndexSource>& source) :
+                m_source(source) {}
 
-            // Inherited via IApplication
-            LocIndString GetId() override
+        protected:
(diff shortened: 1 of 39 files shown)


src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -528,7 +528,7 @@ namespace AppInstaller::Repository
             else
             {
                 AICLI_LOG(Repo, Info, << "Default source requested, multiple sources available, creating aggregated source.");
-                auto aggregatedSource = std::make_shared<AggregatedSource>();
+                auto aggregatedSource = std::make_shared<AggregatedSource>("*DefaultSource");
```
yao-msft, line 531: >DefaultSource [](start = 77, length = 13)

"*DefaultAggregatedSource" to better denote it's a multiple default source?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/Interface_1_0.cpp
```diff
@@ -58,7 +58,7 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_0
         }
 
         // Gets a manifest id by the given key values.
-        std::optional<SQLite::rowid_t> GetManifestIdByKey(SQLite::Connection& connection, SQLite::rowid_t id, std::string_view version = "", std::string_view channel = "")
+        std::optional<SQLite::rowid_t> StaticGetManifestIdByKey(const SQLite::Connection& connection, SQLite::rowid_t id, std::string_view version = "", std::string_view channel = "")
```
yao-msft, line 61: >StaticGetManifestIdByKey [](start = 39, length = 24)

Curious why this refactoring? I didn't see it's used in other places other than the original GetManifestIdByKey

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
```diff
@@ -68,26 +87,70 @@ namespace AppInstaller::Repository::Microsoft
...
+        };
+
+        // The IPackage impl for SQLiteIndexSource.
+        struct Package : public SourceReference, public IPackage
+        {
+            Package(const std::shared_ptr<const SQLiteIndexSource>& source, SQLiteIndex::IdType idId) :
+                SourceReference(source), m_idId(idId) {}
+
+            // Inherited via IPackage
+            std::shared_ptr<IPackageVersion> GetInstalledVersion() const override
```
yao-msft, line 100: >GetInstalledVersion() [](start = 45, length = 21)

Question outside of this PR, will there be GetInstalledInstallerType() or something similar under this interface? Or it'll be from other interfaces?

yao-msft (approved):
:shipit:

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: I don't think that is necessary.

---
In reply to: [492435549](https://github.com/microsoft/winget-cli/pull/575#discussion_r492435549) [](ancestors = 492435549)

src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/Interface_1_0.cpp
JohnMcPMS: The name change? Or why is it in it's own function? I changed the name rather than figure out how to reference it from the member function that now has the same name.  As to why it wasn't just in the other function, I don't know.  But I see no reason to move it at this time.

---
In reply to: [492439957](https://github.com/microsoft/winget-cli/pull/575#discussion_r492439957) [](ancestors = 492439957)

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
JohnMcPMS: Yes, my intention is to record such data when we install, or to reproduce it when we read from the system.  The list of properties will be dynamic, so adding new ones should be easy as well.

---
In reply to: [492441969](https://github.com/microsoft/winget-cli/pull/575#discussion_r492441969) [](ancestors = 492441969)

PR #577, opened 2020-09-23
Interface changes for list/update/uninstall

JohnMcPMS:
## Change
The expected interface definition for use by `list`, `update`, and `uninstall` commands.  This will allow others to make progress implementing these commands, but note that they are not themselves yet implemented.

The expected use is:
```
auto source = OpenPredefinedSource(PredefinedSource::Installed, progress);
auto results = source->Search(...);
if (results.Matches.empty()) { ... } // Found none
else if (results.Matches.size() == 1) { ... } // For targeting a single application
else
{
  for (const auto& match : results.Matches) // For targeting all
  {
    if (match.IsUpdateAvailable())
    { ... } // An update is available
  }
}
```

## Validation
No functional changes should be observed.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/577)

### DIFF
--- a/src/AppInstallerCLITests/WorkFlow.cpp
+++ b/src/AppInstallerCLITests/WorkFlow.cpp
@@ -102,6 +102,11 @@ namespace
                 }
             }
 
+            bool IsUpdateAvailable() const override
+            {
+                return false;
+            }
+
             Manifest m_manifest;
         };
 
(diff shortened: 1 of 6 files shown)


yao-msft (approved):
:shipit:

yao-msft:
Thanks for unblocking!

PR #583, opened 2020-09-29
Fix debug build failure when rebuilding solution

yao-msft:
## Change
AppInstallerTestMsiInstaller project requires manual setting before building. It should not be built during solution build. The release configuration has correct setting to skip the project. This change skips building the project in debug setting.

## Validation
No change to product code.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/583)


JohnMcPMS (approved):
:shipit:

PR #588, opened 2020-10-01
Implement update flow

yao-msft:
## Change
- Added update flow from a local manifest
- Integrate update flow with new repo source interfaces
- Added update all

## Validation
- Added new tests
- Manually tried with local manifest scenario

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/588)


src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -52,9 +52,20 @@ namespace AppInstaller::CLI::Execution
         LogPath,
         InstallerArgs,
         CompletionData,
+        CommandType,
+        InstalledPackageVersion,
         Max
     };
 
+    enum class CommandType
```
JohnMcPMS, line 60: >CommandType [](start = 15, length = 11)

This should be defined elsewhere, it is not directly relevant to this file.

Second, this strongly suggests that decisions are going to be made based on CommandType.  It is a bad idea to start down that path, as it leads to confusion and bugs.  It is far better to create explicit, semantic decision points.  This will lead to more scoped use and a clearer picture of the implications.  I'm not sure what that looks like for your change yet.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -168,6 +191,9 @@ namespace AppInstaller::CLI::Execution
         // Set the context to the terminated state.
         void Terminate(HRESULT hr, std::string_view file = {}, size_t line = {});
 
+        // Resumes the context
+        void Resume() { m_terminationHR = S_OK; m_isTerminated = false; }
```
JohnMcPMS, line 195: >Resume [](start = 13, length = 6)

I think there is a better way to handle this with the existing tools.  For instance, completion creates a `Clone()` of the `Context` and disables telemetry while it is running.  Disabling telemetry might be too extreme, but we do want a way to indicate that a single update has failed, not the entire run.

So I would suggest cloning the `Context`, adding a new member that indicates this is a subset of the operation, and proceeding with each update as a singular operation on the clone.  Then we can pass the "subset" flag to the telemetry call on failure and have an easy way to separate them in the stream.

src/AppInstallerCLICore/Commands/UpdateCommand.h
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "Command.h"
+
+namespace AppInstaller::CLI
+{
+    struct UpdateCommand final : public Command
+    {
+        UpdateCommand(std::string_view parent) : Command("update", parent) {}
```
JohnMcPMS, line 10: >UpdateCommand(std::string_view parent) : Command("update", parent) {} [](start = 8, length = 69)

Experimental?  Definitely want this behind a feature for now.

src/AppInstallerCLICore/Commands/UpdateCommand.cpp
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return { Resource::String::UpdateCommandShortDescription };
+    }
+
+    Resource::LocString UpdateCommand::LongDescription() const
+    {
+        return { Resource::String::UpdateCommandLongDescription };
+    }
+
+    void UpdateCommand::Complete(Execution::Context& context, Execution::Args::Type valueType) const
```
JohnMcPMS, line 50: >Complete [](start = 24, length = 8)

This won't be correct for completion; it should look much more like `list` completion than this (which I know isn't in).  Maybe just put in a todo for now?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -200,6 +204,10 @@ namespace AppInstaller::CLI::Workflow
         case ManifestInstaller::InstallerTypeEnum::Msi:
         case ManifestInstaller::InstallerTypeEnum::Nullsoft:
         case ManifestInstaller::InstallerTypeEnum::Wix:
+            if (isUpdate && installer.UpdateBehavior == ManifestInstaller::UpdateBehaviorEnum::UninstallPrevious)
+            {
+                // TODO: hook up with uninstall when uninstall is implemented
```
JohnMcPMS, line 209: >// TODO: hook up with uninstall when uninstall is implemented [](start = 16, length = 61)

For now, this should probably terminate the context to prevent the install from happening.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -10,6 +11,28 @@ namespace AppInstaller::CLI::Workflow
 {
     bool InstallerComparator::operator() (const ManifestInstaller& installer1, const ManifestInstaller& installer2)
     {
+        // Applicable architecture should always come before inapplicable architecture
+        if (Utility::IsApplicableArchitecture(installer1.Arch) != -1 && Utility::IsApplicableArchitecture(installer2.Arch) == -1)
```
JohnMcPMS, line 15: >-1 [](start = 66, length = 2)

Can we have a constant defined for this?

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -10,6 +11,28 @@ namespace AppInstaller::CLI::Workflow
...
+        {
+            return true;
+        }
+
+        // If there's installation metadata, pick the preferred one or compatible one
+        if (m_installationMetadata.has_value())
+        {
+            auto installedType = Manifest::ManifestInstaller::ConvertToInstallerTypeEnum(
+                m_installationMetadata.value().find(s_InstallationMetadata_Key_InstallerType)->second);
```
JohnMcPMS, line 24: >find(s_InstallationMetadata_Key_InstallerType) [](start = 47, length = 46)

Assumes value will be in map; should not.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -97,22 +100,29 @@ namespace AppInstaller::CLI::Workflow
...
                 installerArgs += ' ' + installerSwitches.at(ManifestInstaller::InstallerSwitchType::Log);
             }
 
-            // Construct custom arg.
-            if (installerSwitches.find(ManifestInstaller::InstallerSwitchType::Custom) != installerSwitches.end())
+            if (isUpdate && installerSwitches.find(ManifestInstaller::InstallerSwitchType::Update) != installerSwitches.end())
+            {
+                installerArgs += ' ' + installerSwitches.at(ManifestInstaller::InstallerSwitchType::Update);
+            }
+            else
```
JohnMcPMS, line 113: >else [](start = 12, length = 4)

Why are the custom args excluded from update? Do we need to have an entire "update" switches block that can override the base ones?

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -97,22 +100,29 @@ namespace AppInstaller::CLI::Workflow
...
+            {
+                // Construct custom arg.
+                if (installerSwitches.find(ManifestInstaller::InstallerSwitchType::Custom) != installerSwitches.end())
+                {
+                    installerArgs += ' ' + installerSwitches.at(ManifestInstaller::InstallerSwitchType::Custom);
+                }
+            }
+
+            // Construct install location arg if necessary.
+            if (!isUpdate && context.Args.Contains(Execution::Args::Type::InstallLocation) && installerSwitches.find(ManifestInstaller::InstallerSwitchType::InstallLocation) != installerSwitches.end())
```
JohnMcPMS, line 123: >InstallLocation [](start = 74, length = 15)

Will the "remove first" update flow just be uninstall + install?  Because I could see that potentially needing the install location here.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -107,7 +107,7 @@ namespace AppInstaller::CLI::Workflow
         }
         catch (...)
         {
-            context.Reporter.Error() << "Failed to open the source; try removing and re-adding it" << std::endl;
+            context.Reporter.Error() << Resource::String::OpenSourceFailed << std::endl;
```
JohnMcPMS, line 110: >OpenSourceFailed [](start = 58, length = 16)

Oh no, we both did this :)

src/AppInstallerCLICore/Workflows/ManifestComparator.h
```diff
@@ -26,16 +32,19 @@ namespace AppInstaller::CLI::Workflow
     };
 
     // Class in charge of comparing manifest entries
-    class ManifestComparator
+    struct ManifestComparator
     {
-    public:
-        ManifestComparator(const Execution::Args&) {}
+        ManifestComparator(const Execution::Args&, std::optional<std::map<std::string, std::string>> installationMetadata) :
```
JohnMcPMS, line 37: >optional [](start = 56, length = 8)

Do you need an optional here, and empty map should work just fine as well, right?  I don't see how "map present, but key missing" is an different from "map missing".

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -426,6 +472,24 @@ namespace AppInstaller::CLI::Workflow
             AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_EXPERIMENTAL_FEATURE_DISABLED);
         }
     }
+
+    void SearchSourceUsingManifest(Execution::Context& context)
```
JohnMcPMS, line 476: >SearchSourceUsingManifest [](start = 9, length = 25)

I'm not sure that this is going to work out terribly well as implemented.

I would start with a similar search pattern as we correlate with elsewhere; the system reference strings.  Then fall back to the Id / Name combo.  And I would suggest Id to be CaseInsensitive as the match type.

If you want, you could wait to see how I'm doing it in list.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -426,6 +472,24 @@ namespace AppInstaller::CLI::Workflow
...
+
+        SearchRequest searchRequest;
+        searchRequest.Inclusions.emplace_back(PackageMatchFilter(PackageMatchField::Id, MatchType::Exact, manifest.Id));
+        // In case there're same Ids from different sources, filter the result using package name
+        searchRequest.Filters.emplace_back(PackageMatchFilter(PackageMatchField::Name, MatchType::Exact, manifest.Name));
+
+        context.Add<Execution::Data::SearchResult>(context.Get<Execution::Data::Source>()->Search(searchRequest));
+    }
+
+    void GetInstalledPackage(Execution::Context& context)
```
JohnMcPMS, line 488: >GetInstalledPackage [](start = 9, length = 19)

This appears to be getting the installed PackageVersion.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -655,4 +674,29 @@ They can be configured through the settings file 'winget settings'.</value>
...
+  </data>
+  <data name="OpenSourceFailedNoMatch" xml:space="preserve">
+    <value>No sources match the given value:</value>
+  </data>
+  <data name="OpenSourceFailedNoMatchHelp" xml:space="preserve">
+    <value>The configured sources are:</value>
+  </data>
+  <data name="OpenSourceFailedNoSourceDefined" xml:space="preserve">
+    <value>No sources defined; add one with 'source add' or reset to defaults with 'source reset'</value>
+    <comment>'source add' and 'source reset' are winget commands. It would be beneficial to not translate them</comment>
```
JohnMcPMS, line 691: ><comment>'source add' and 'source reset' are winget commands. It would be beneficial to not translate them</comment> [](start = 4, length = 116)

Use <comment>{Locked="VirtualTerminal"}</comment>

src/AppInstallerCommonCore/Manifest/ManifestInstaller.cpp
```diff
@@ -50,6 +50,23 @@ namespace AppInstaller::Manifest
         return result;
     }
 
+    ManifestInstaller::UpdateBehaviorEnum ManifestInstaller::ConvertToUpdateBehaviorEnum(const std::string& in)
+    {
+        std::string inStrLower = Utility::ToLower(in);
+        UpdateBehaviorEnum result = UpdateBehaviorEnum::Unknown;
+
+        if (inStrLower == "install")
```
JohnMcPMS, line 58: >inStrLower == "install" [](start = 12, length = 23)

Utility::CaseInsensitiveEquals?

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
```diff
@@ -421,7 +425,8 @@ namespace AppInstaller::Manifest
                 {ManifestInstaller::InstallerSwitchType::Silent, ManifestInstaller::string_t("/quiet")},
                 {ManifestInstaller::InstallerSwitchType::SilentWithProgress, ManifestInstaller::string_t("/passive")},
                 {ManifestInstaller::InstallerSwitchType::Log, ManifestInstaller::string_t("/log \"" + std::string(ARG_TOKEN_LOGPATH) + "\"")},
-                {ManifestInstaller::InstallerSwitchType::InstallLocation, ManifestInstaller::string_t("TARGETDIR=\"" + std::string(ARG_TOKEN_INSTALLPATH) + "\"")}
+                {ManifestInstaller::InstallerSwitchType::InstallLocation, ManifestInstaller::string_t("TARGETDIR=\"" + std::string(ARG_TOKEN_INSTALLPATH) + "\"")},
+                {ManifestInstaller::InstallerSwitchType::Update, ManifestInstaller::string_t("REINSTALL=ALL REINSTALLMODE=vamus")}
```
JohnMcPMS, line 429: >vamus [](start = 122, length = 5)

Are these the correct flags to use? a means that every file will be rewritten, so it makes me wonder...

src/AppInstallerCLICore/Commands/UpdateCommand.cpp
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    std::string UpdateCommand::HelpLink() const
+    {
+        // TODO: point to correct location
+        return "https://aka.ms/winget-command-update";
+    }
+
+    void UpdateCommand::ExecuteInternal(Execution::Context& context) const
```
JohnMcPMS, line 83: >ExecuteInternal [](start = 24, length = 15)

There should be three distinct scenarios for update per the spec:
1. `update` :: Basically the same as `list` but filtered to only packages with updates available.
2. `update foo` :: Searches for package `foo`, proceed to update if found.
3. `update --all` :: Equivalent to using the output from 1 to run 2 for each package.

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+                    context <<
+                        SelectLatestApplicableUpdate(*(match.Package)) <<
+                        ShowInstallationDisclaimer <<
+                        DownloadInstaller <<
+                        ExecuteInstaller <<
+                        RemoveInstaller;
+                });
+
+            tryOneUpdate.join();
```
JohnMcPMS, line 171: >tryOneUpdate.join(); [](start = 12, length = 20)

Why are you running this on a separate thread just to immediately wait for it?

yao-msft: AICLI_Terminate() has return; which will exit the current method immediately.

---
In reply to: [499935455](https://github.com/microsoft/winget-cli/pull/588#discussion_r499935455) [](ancestors = 499935455)

yao-msft: I can put these in a separate method 

---
In reply to: [499936643](https://github.com/microsoft/winget-cli/pull/588#discussion_r499936643) [](ancestors = 499936643,499935455)

JohnMcPMS: Even with a forced return, you can make it a lambda or named function.

---
In reply to: [499938571](https://github.com/microsoft/winget-cli/pull/588#discussion_r499938571) [](ancestors = 499938571,499936643,499935455)

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerCLICore/Commands/UpdateCommand.cpp
JohnMcPMS: And in all cases I think it would be very beneficial to have the common steps of the workflow (source set up, search) performed once, and only diverge once needed.  As you have it here, every path does its own setup, and the nested workflows make it very hard to follow.

---
In reply to: [499918627](https://github.com/microsoft/winget-cli/pull/588#discussion_r499918627) [](ancestors = 499918627)

src/AppInstallerCLICore/Commands/UpdateCommand.h
yao-msft, line 10: >update [](start = 58, length = 6)

upgrade

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
yao-msft: Chatted offline, we can start with this for most compatible as default

---
In reply to: [499912738](https://github.com/microsoft/winget-cli/pull/588#discussion_r499912738) [](ancestors = 499912738)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
yao-msft: :)

---
In reply to: [499904348](https://github.com/microsoft/winget-cli/pull/588#discussion_r499904348) [](ancestors = 499904348)

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -52,9 +52,18 @@ namespace AppInstaller::CLI::Execution
         LogPath,
         InstallerArgs,
         CompletionData,
+        ContextFlag,
+        InstalledPackageVersion,
         Max
     };
 
+    // bit masks used as Context flags
+    namespace ContextFlag
```
JohnMcPMS, line 61: >namespace ContextFlag [](start = 4, length = 21)

enum class

JohnMcPMS: And then define some operators.

---
In reply to: [503506092](https://github.com/microsoft/winget-cli/pull/588#discussion_r503506092) [](ancestors = 503506092)

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -139,7 +160,10 @@ namespace AppInstaller::CLI::Execution
     // arguments via Execution::Args.
     struct Context
     {
-        Context(std::ostream& out, std::istream& in) : Reporter(out, in) {}
+        Context(std::ostream& out, std::istream& in) : Reporter(out, in) { m_data[Data::ContextFlag] = ContextFlag::None; }
```
JohnMcPMS, line 163: >m_data[Data::ContextFlag] = ContextFlag::None; [](start = 75, length = 46)

Use Add

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -139,7 +160,10 @@ namespace AppInstaller::CLI::Execution
     // arguments via Execution::Args.
     struct Context
     {
-        Context(std::ostream& out, std::istream& in) : Reporter(out, in) {}
+        Context(std::ostream& out, std::istream& in) : Reporter(out, in) { m_data[Data::ContextFlag] = ContextFlag::None; }
+
+        // Clone the reporter for this constructor.
+        Context(Execution::Reporter& reporter) : Reporter(reporter, Execution::Reporter::clone_t{}) {}
```
JohnMcPMS, line 166: Why are you making this public?  It wasn't to force the use of the Clone method.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -150,7 +174,7 @@ namespace AppInstaller::CLI::Execution
         Args Args;
 
         // Creates a copy of this context as it was at construction.
-        Context Clone();
+        virtual std::unique_ptr<Context> Clone();
```
JohnMcPMS, line 177: >virtual std::unique_ptr<Context> Clone(); [](start = 8, length = 41)

Why does this need to be a unique_ptr?  I agree that I didn't put in all the copy and move functions, but that would be the better choice.

src/AppInstallerCLICore/Commands/UpdateCommand.h
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "Command.h"
+
+namespace AppInstaller::CLI
+{
+    struct UpdateCommand final : public Command
```
JohnMcPMS, line 8: >UpdateCommand [](start = 11, length = 13)

UpgradeCommand and relevant file name changes.

src/AppInstallerCLICore/ExecutionContext.h
JohnMcPMS: Actually, I think having GetOrAdd would be better

---
In reply to: [503506625](https://github.com/microsoft/winget-cli/pull/588#discussion_r503506625) [](ancestors = 503506625)

src/AppInstallerCLICore/Commands/UpdateCommand.cpp
```diff
@@ -0,0 +1,134 @@
...
+                    // This step also populates Manifest and Installer in context data
+                    context <<
+                        SelectLatestApplicableUpdate(*(context.Get<Execution::Data::SearchResult>().Matches.at(0).Package));
+                }
+
+                context <<
+                    ShowInstallationDisclaimer <<
+                    DownloadInstaller <<
+                    ExecuteInstaller <<
+                    RemoveInstaller;
```
JohnMcPMS, line 131: Extra indentation.

src/AppInstallerCLICore/Commands/UpdateCommand.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    void UpdateCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        context.Get<Execution::Data::ContextFlag>() |= Execution::ContextFlag::InstallerExecutionUseUpdate;
+
+        context <<
+            OpenSource <<
+            GetCompositeSourceFromInstalledAndAvailable;
+
+        if (context.Args.Empty())
```
JohnMcPMS, line 69: >if (context.Args.Empty()) [](start = 8, length = 25)

I think this might need an args validation function.

JohnMcPMS: Not Empty, but the command.

---
In reply to: [503511290](https://github.com/microsoft/winget-cli/pull/588#discussion_r503511290) [](ancestors = 503511290)

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
JohnMcPMS: Can the constant not have s_?  A public name should be cleaner.

---
In reply to: [499894685](https://github.com/microsoft/winget-cli/pull/588#discussion_r499894685) [](ancestors = 499894685)

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -15,7 +15,7 @@ namespace AppInstaller::CLI::Workflow
         const auto& manifest = context.Get<Execution::Data::Manifest>();
         const auto& installer = context.Get<Execution::Data::Installer>();
 
-        ManifestComparator manifestComparator(context.Args);
+        ManifestComparator manifestComparator(context.Args, {});
```
JohnMcPMS, line 18: >{} [](start = 60, length = 2)

Consider a default value of = {} in the function definition.

src/AppInstallerCLICore/ExecutionContext.h
JohnMcPMS: Should have seen the virtual, got it now.

---
In reply to: [503507999](https://github.com/microsoft/winget-cli/pull/588#discussion_r503507999) [](ancestors = 503507999)

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -36,6 +38,8 @@ namespace AppInstaller::Settings
             return ExperimentalFeature{ "Argument Sample", "experimentalArg", "https://aka.ms/winget-settings", Feature::ExperimentalArg };
         case Feature::ExperimentalMSStore:
             return ExperimentalFeature{ "Microsoft Store Support", "experimentalMSStore", "https://aka.ms/winget-settings", Feature::ExperimentalMSStore };
+        case Feature::ExperimentalUpgrade:
+            return ExperimentalFeature{ "Upgrade Command", "experimentalUpgrade", "https://aka.ms/winget-settings", Feature::ExperimentalUpgrade };
```
JohnMcPMS, line 42: >experimentalUpgrade [](start = 60, length = 19)

Just "upgrade"

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
JohnMcPMS: Would be better handled by upgrade simply removing the flag so that every switch could benefit at the same time.

---
In reply to: [499903564](https://github.com/microsoft/winget-cli/pull/588#discussion_r499903564) [](ancestors = 499903564)

src/AppInstallerCLICore/ExecutionContext.h
yao-msft: std::make_unique in the clone90 needs to access it 

---
In reply to: [503507085](https://github.com/microsoft/winget-cli/pull/588#discussion_r503507085) [](ancestors = 503507085)

src/AppInstallerCLICore/Command.cpp
```diff
@@ -720,6 +722,25 @@ namespace AppInstaller::CLI
...
+             execArgs.Contains(Execution::Args::Type::Name) ||
+             execArgs.Contains(Execution::Args::Type::Moniker) ||
+             execArgs.Contains(Execution::Args::Type::Version) ||
+             execArgs.Contains(Execution::Args::Type::Channel) ||
+             execArgs.Contains(Execution::Args::Type::Source) ||
+             execArgs.Contains(Execution::Args::Type::Exact) ||
+             execArgs.Contains(Execution::Args::Type::All)))
+        {
+            throw CommandException(Resource::String::BothManifestAndSearchQueryProvided, "");
+        }
```
JohnMcPMS, line 743: I don't really think this is the best place to put these validations, there may come a time in the future where we some command needs this to be false and then we have to move this to every command.  But I think it's ok to leave it for now; the better solution is parameter groups like PowerShell.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -52,9 +52,19 @@ namespace AppInstaller::CLI::Execution
...
         CompletionData,
+        InstalledPackageVersion,
         Max
     };
 
+    // bit masks used as Context flags
+    enum class ContextFlag : int
+    {
+        None = 0x0,
+        InstallerExecutionUseUpdate,
```
JohnMcPMS, line 63: >InstallerExecutionUseUpdate [](start = 8, length = 27)

As a flags enum, we will have to specify the values after this one, so we should still have it be 0x1.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -192,6 +193,8 @@ namespace AppInstaller::CLI::Workflow
     {
         const auto& installer = context.Get<Execution::Data::Installer>().value();
 
+        bool isUpdate = (context.GetFlags() & Execution::ContextFlag::InstallerExecutionUseUpdate) != Execution::ContextFlag::None;
```
JohnMcPMS, line 196: >(context.GetFlags() & Execution::ContextFlag::InstallerExecutionUseUpdate) != Execution::ContextFlag::None; [](start = 24, length = 107)

WI_IsFlagSet

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -200,6 +203,12 @@ namespace AppInstaller::CLI::Workflow
         case ManifestInstaller::InstallerTypeEnum::Msi:
         case ManifestInstaller::InstallerTypeEnum::Nullsoft:
         case ManifestInstaller::InstallerTypeEnum::Wix:
+            if (isUpdate && installer.UpdateBehavior == ManifestInstaller::UpdateBehaviorEnum::UninstallPrevious)
+            {
+                // TODO: hook up with uninstall when uninstall is implemented
+                context.GetFlags() &= ~Execution::ContextFlag::InstallerExecutionUseUpdate;
```
JohnMcPMS, line 209: >context.GetFlags() &= ~Execution::ContextFlag::InstallerExecutionUseUpdate; [](start = 16, length = 75)

WI_ClearFlag (and generally throughout use the WI macros)

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
```diff
@@ -0,0 +1,135 @@
...
+        bool updateAllHasFailure = false;
+        for (const auto& match : matches)
+        {
+            // We want to do best effort to update all applicable updates regardless on previous update failure
+            auto updateContextPtr = context.Clone();
+            Execution::Context& updateContext = *updateContextPtr;
+            updateContext.Reporter.Info() << std::endl;
+
+            updateContext.Add<Execution::Data::InstalledPackageVersion>(match.Package->GetInstalledVersion());
+            updateContext.GetFlags() = context.GetFlags();
```
JohnMcPMS, line 114: This seems like something that should always happen, and thus be part of cloning.

src/AppInstallerCLICore/ExecutionContext.h
yao-msft: I guess I forgot to add back when modifying..

---
In reply to: [504133540](https://github.com/microsoft/winget-cli/pull/588#discussion_r504133540) [](ancestors = 504133540)

src/AppInstallerCLICore/Command.cpp
yao-msft: Moved to individual commands

---
In reply to: [504132887](https://github.com/microsoft/winget-cli/pull/588#discussion_r504132887) [](ancestors = 504132887)

JohnMcPMS (approved):
:shipit:

PR #598, opened 2020-10-08
List command initial implementation behind feature toggle

JohnMcPMS:
## Change
This phase introduces the `list` command and much of the associative plumbing.  It currently only handles enumerating MSIX packages from the system, but it should be able to support additional enumeration simply by adding more to the index.

The goal of the `list` command is to enumerate packages installed on the machine, as well as to indicate which packages have an update available from a configured source.  This is achieved by creating predefined sources for installed packages, and a `CompositeSource` class to do the work of correlating the installed packages with those from external sources.

The `CompositeSource` performs much as the `AggregateSource` did before if it is simply used with external sources.  When an installed source is provided though, `Search` begins correlating between the installed and external sources of available packages.  The process is as follows:
1. `Search` against the installed source using the input request
2. For each package found:
  a. `Search` for available packages using the system reference strings (currently PackageFamilyName or ProductCode)
  b. The first source that returns a match is the upgrade target
  c. Place this package in the results, along with the associated available package
3. `Search` against each available source using the input request
4. For each package found:
  a. Determine if there is an existing matching package in the results (via system reference string) and update the match criteria if it is better
  b. If there is no existing match, `Search` the installed source using system reference strings
  c. If an installed package is found this way, place it into the results along with the available package
5. Sort and possibly truncate the results

This allows us to correlate the external metadata with packages that are installed locally, as long as the system reference strings are correct.  This means that installed packages can be targeted by their `Id` from the `winget-pkgs` source, even if that value is not the same one used by the local identifying data.

Also:
- Enables searches by system reference string (ex. `winget search Microsoft.WindowsTerminal_8wekyb3d8bbwe`)
- Makes `CaseInsensitiveEquals` better (uses `FoldCase` via ICU)
- Adds the schema to the default settings file
- Creates a special case for Latest and Unknown versions

## Validation
Adds tests for:
- `CompositeSource` all up
- The predefined installed source
- `SQLiteIndex` multi-properties
- Latest and Unknown special versions
- and refactors out the various test `ISource` implementations in to a single one that can provide the majority of the base functionality to all

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/598)


src/AppInstallerCLICore/TableOutput.h
```diff
@@ -30,6 +30,7 @@ namespace AppInstaller::CLI::Execution
     }
 
     // Enables output data in a table format.
+    // TODO: Improve for use with sparse data.
```
yao-msft, line 33: >TODO: Improve for use with sparse data. [](start = 7, length = 39)

what kind of improvement are you considering?

JohnMcPMS: Probably a bit more control over minimum/maximum/collapse state for columns.  For instance, list could easily have only 1 update to show but far to the end and the code here would not show it.  I suppose that could be fixed by simply sorting the updates to the beginning of the output though.

---
In reply to: [506790702](https://github.com/microsoft/winget-cli/pull/598#discussion_r506790702) [](ancestors = 506790702)

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
```diff
@@ -104,6 +115,9 @@ namespace AppInstaller::Repository
         // Gets a property of this package version.
         virtual Utility::LocIndString GetProperty(PackageVersionProperty property) const = 0;
 
+        // Gets a property of this package version.
```
yao-msft, line 118: >property [](start = 18, length = 8)

multi value?

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
```diff
@@ -153,14 +206,48 @@ namespace AppInstaller::Repository::Microsoft
...
+            {
+                return false;
+            }
         };
     }
 
-    SQLiteIndexSource::SQLiteIndexSource(const SourceDetails& details, std::string identifier, SQLiteIndex&& index, Synchronization::CrossProcessReaderWriteLock&& lock) :
-        m_details(details), m_identifier(std::move(identifier)), m_lock(std::move(lock)), m_index(std::move(index))
+    SQLiteIndexSource::SQLiteIndexSource(const SourceDetails& details, std::string identifier, SQLiteIndex&& index, Synchronization::CrossProcessReaderWriteLock&& lock, bool isInstalledSource) :
+        m_details(details), m_identifier(std::move(identifier)), m_lock(std::move(lock)), m_isInstalled(isInstalledSource), m_index(std::move(index))
```
yao-msft, line 250: You might have answered this before, should this be just in the header?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -198,6 +222,11 @@ namespace AppInstaller::CLI::Workflow
         if (args.Contains(Execution::Args::Type::Query))
         {
             std::string_view query = args.GetArg(Execution::Args::Type::Query);
+
+            // Regardless of match type, always use an exact match for the system reference strings.
+            searchRequest.Inclusions.emplace_back(PackageMatchFilter(PackageMatchField::PackageFamilyName, MatchType::Exact, query));
```
yao-msft, line 227: >PackageMatchField::PackageFamilyName [](start = 69, length = 36)

I can get rid of the Moniker hack instore source now!

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -42,6 +48,26 @@ namespace AppInstaller::Utility
 
     bool Version::operator<(const Version& other) const
```
yao-msft, line 49: >< [](start = 26, length = 1)

should we sort Unknown < Latest? Looks like currently Latest < Unknown?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -304,6 +308,13 @@ They can be configured through the settings file 'winget settings'.</value>
     <value>Links</value>
     <comment>Links to different webpages</comment>
   </data>
+  <data name="ListCommandLongDescription" xml:space="preserve">
+    <value>The list command displays the packages installed on the system, as well as whether an update is available. Additional options can be provided to filter the output, much like the search command.</value>
+    <comment>{Locked="list"}{Locked="search"}</comment>
```
yao-msft, line 313: >{Locked="list"}{Locked="search"} [](start = 13, length = 32)

The sample I saw is {Locked="list","search"}, not sure if it matters

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
```diff
@@ -95,17 +110,57 @@ namespace AppInstaller::Repository::Microsoft
...
                 SourceReference(source), m_idId(idId) {}
 
+            Utility::LocIndString GetProperty(PackageProperty property) const
+            {
+                std::shared_ptr<IPackageVersion> truth = GetLatestVersionInternal();
+
+                switch (property)
+                {
+                case PackageProperty::Id:
+                    return truth->GetProperty(PackageVersionProperty::Id);
```
yao-msft, line 126: >truth [](start = 27, length = 5)

check null since GetLatestVersionInternal() may return nullptr?

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
```diff
@@ -124,21 +179,19 @@ namespace AppInstaller::Repository::Microsoft
...
-                // Although we could potentially increase efficiency here, this should be fine.
-                std::vector<PackageVersionKey> versions = GetAvailableVersionKeys();
-
-                if (!versions.empty())
-                {
-                    return GetAvailableVersion(versions[0]);
-                }
-
-                return {};
+                return GetLatestVersionInternal();
```
yao-msft, line 182: >GetLatestVersionInternal() [](start = 23, length = 26)

Maybe PackageBase::GetLatestVersionInternal() to be explicit and match style in GetProperty above

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.h
```diff
@@ -0,0 +1,43 @@
...
+
+    // A source of installed packages on the local system.
+    // Arg  ::  A value indicating how the list is to be filtered.
+    // Data ::  Not used.
+    struct PredefinedInstalledSourceFactory
+    {
+        // Get the type string for this source.
+        static constexpr std::string_view Type()
+        {
+            using namespace std::string_view_literals;
```
yao-msft, line 21: >using namespace std::string_view_literals; [](start = 12, length = 42)

nit: already had the using above

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -0,0 +1,175 @@
...
+        {
+            return Filter::ARP_System;
+        }
+        else if (filter == FilterToString(Filter::ARP_User))
+        {
+            return Filter::ARP_User;
+        }
+        else if (filter == FilterToString(Filter::MSIX))
+        {
+            return Filter::ARP_User;
```
yao-msft, line 163: >ARP_User [](start = 27, length = 8)

MSIX

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        // Populates the index with the entries from MSIX.
+        void PopulateIndexFromMSIX(SQLiteIndex& index)
+        {
+            using namespace winrt::Windows::ApplicationModel;
+            using namespace winrt::Windows::Management::Deployment;
+
+            PackageManager packageManager;
+            auto packages = packageManager.FindPackagesForUserWithPackageTypes({}, PackageTypes::Main);
```
yao-msft, line 30: >PackageTypes::Main [](start = 83, length = 18)

Do we need a TODO for other types?

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+            // Reuse the same manifest object, as we will be setting the same values every time.
+            Manifest::Manifest manifest;
+            // Add one installer for storing the package family name.
+            manifest.Installers.emplace_back();
+            // Every package will have the same tags currently.
+            manifest.Tags = { "msix" };
```
yao-msft, line 37: >manifest.Tags = { "msix" }; [](start = 12, length = 27)

What's this tag used for?

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -0,0 +1,175 @@
...
+                if (signatureKind == PackageSignatureKind::System)
+                {
+                    continue;
+                }
+
+                auto packageId = package.Id();
+                Utility::NormalizedString familyName = Utility::ConvertToUTF8(packageId.FamilyName());
+
+                manifest.Id = familyName;
+                manifest.Name = Utility::ConvertToUTF8(package.DisplayName());
```
yao-msft, line 58: >Utility::ConvertToUTF8(package.DisplayName()) [](start = 32, length = 45)

This might be interesting for different locales

yao-msft: The doc says: This property is only supported for UWP app. If you call this property on an instance of a package that was created by any mechanism other than Package.Current, this property returns an empty string ("").
Should we still use this?

---
In reply to: [507499499](https://github.com/microsoft/winget-cli/pull/598#discussion_r507499499) [](ancestors = 507499499)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+                if (manifest.Name.empty())
+                {
+                    manifest.Name = Utility::ConvertToUTF8(packageId.Name());
+                }
+
+                std::ostringstream strstr;
```
yao-msft, line 65: >strstr [](start = 35, length = 6)

strstream?

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -578,36 +584,44 @@ namespace AppInstaller::Repository
...
     }
 
-    std::shared_ptr<ISource> CreateCompositeSource(std::shared_ptr<ISource>& source1, std::shared_ptr<ISource>&)
+    std::shared_ptr<ISource> CreateCompositeSource(const std::shared_ptr<ISource>& installedSource, const std::shared_ptr<ISource>& availableSource)
     {
-        // TODO: needs implementation
-        return source1;
+        std::shared_ptr<CompositeSource> result = std::dynamic_pointer_cast<CompositeSource>(availableSource);
+
+        if (!result)
```
yao-msft, line 616: >!result [](start = 12, length = 7)

In what case will this happen?

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -0,0 +1,175 @@
...
+        }
+        else if (filter == FilterToString(Filter::ARP_User))
+        {
+            return Filter::ARP_User;
+        }
+        else if (filter == FilterToString(Filter::MSIX))
+        {
+            return Filter::ARP_User;
+        }
+        else
```
yao-msft, line 165: >else [](start = 8, length = 4)

Do we need unknown to be safe or we don't quite care?

src/AppInstallerCLITests/TestSource.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+        return AvailableVersions[0];
+    }
+
+    std::shared_ptr<IPackageVersion> TestPackage::GetAvailableVersion(const PackageVersionKey& versionKey) const
+    {
+        for (const auto& version : AvailableVersions)
+        {
+            if ((versionKey.Version.empty() || versionKey.Version == version->GetProperty(PackageVersionProperty::Version).get()) && 
+                (versionKey.Channel.empty() ||  versionKey.Channel == version->GetProperty(PackageVersionProperty::Channel).get()))
```
yao-msft, line 154: >  [](start = 47, length = 1)

nit: extra space

src/AppInstallerCLITests/PredefinedInstalledSource.cpp
```diff
@@ -0,0 +1,46 @@
...
+    details.Type = Factory::Type();
+    details.Arg = Factory::FilterToString(filter);
+
+    TestProgress progress;
+
+    auto factory = Factory::Create();
+    return factory->Create(details, progress);
+}
+
+TEST_CASE("PredefinedInstalledSource_Create", "[installed][list]")
```
yao-msft, line 32: >PredefinedInstalledSource_Create [](start = 11, length = 32)

Maybe redundant to make this a separate test since PredefinedInstalledSource_Search calls it anyway

src/AppInstallerCLITests/CompositeSource.cpp
```diff
@@ -0,0 +1,563 @@
...
+using namespace AppInstaller::Utility;
+
+constexpr std::string_view s_Everything_Query = "everything"sv;
+
+// A test source that has two modes:
+//  1. A request that IsForEverything returns the stored result. This models the
+//      incoming search request to a CompositeSource.
+//  2. A request that is not for everything invokes TestSource::SearchFunction to
+//      enable verification of expectations.
+struct ComponentTestSource : public TestSource
```
yao-msft, line 22: >ComponentTestSource [](start = 7, length = 19)

Typo CompositeTestSource?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -0,0 +1,578 @@
...
+            }
+
+        private:
+            std::shared_ptr<IPackage> m_installedPackage;
+            Utility::LocIndString m_installedChannel;
+            std::shared_ptr<IPackage> m_availablePackage;
+        };
+
+        // A sentinel package with an unknown version.
+        struct UnknownAvailablePackage : public IPackage
```
yao-msft, line 121: >UnknownAvailablePackage [](start = 15, length = 23)

Why is this needed? Shall we just return empty if no available package matched?

yao-msft: Got it, it's for ambiguous match

---
In reply to: [507624727](https://github.com/microsoft/winget-cli/pull/598#discussion_r507624727) [](ancestors = 507624727)

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+                SearchResult installedCrossRef = m_installedSource->Search(systemReferenceSearch);
+
+                for (auto&& crossRef : installedCrossRef.Matches)
+                {
+                    // Ensure that we don't pick up the same package from two available sources by recording it in the map.
+                    auto installedVersion = crossRef.Package->GetInstalledVersion();
+                    auto installedPackageData = result.ReserveInstalledPackageSlot(installedVersion.get());
+
+                    result.Matches.emplace_back(std::make_shared<CompositePackage>(std::move(crossRef.Package), std::move(match.Package)), match.MatchCriteria);
```
yao-msft, line 530: >result.Matches.emplace_back [](start = 20, length = 27)

I think we should check installedPackageData before emplace?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -0,0 +1,578 @@
...
+                    {
+                        data.SystemReferenceStrings.emplace_back(srs);
+                        m_systemReferenceMap.emplace(std::move(srs), data.MatchIndex);
+                    }
+                }
+            }
+
+            bool HandleSystemReferenceStringTypeForCheckForExistingResultFromAvailablePackageMatch(
+                const ResultMatch& match,
+                IPackageVersion* installedVersion,
```
yao-msft, line 327: >installedVersion [](start = 33, length = 16)

availableVersion

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -276,8 +313,60 @@ namespace AppInstaller::CLI::Workflow
...
             });
+
+        for (const auto& match : searchResult.Matches)
+        {
+            auto installedVersion = match.Package->GetInstalledVersion();
+
+            if (installedVersion)
+            {
+                auto latestVersion = match.Package->GetLatestAvailableVersion();
+                bool updateAvailable = match.Package->IsUpdateAvailable();
```
yao-msft, line 348: >updateAvailable = match.Package->IsUpdateAvailable(); [](start = 21, length = 53)

This updatability only considers version, are we considering "applicable updates" in the future?

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
JohnMcPMS: I prefer non-trivial functions, including constructors, to be defined in the cpp.

---
In reply to: [507279645](https://github.com/microsoft/winget-cli/pull/598#discussion_r507279645) [](ancestors = 507279645)

src/AppInstallerCommonCore/Versions.cpp
JohnMcPMS: Added a test to ensure Unknown < Latest, currently passing.

---
In reply to: [507387816](https://github.com/microsoft/winget-cli/pull/598#discussion_r507387816) [](ancestors = 507387816)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
JohnMcPMS: I don't think so.

---
In reply to: [507556942](https://github.com/microsoft/winget-cli/pull/598#discussion_r507556942) [](ancestors = 507556942)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
JohnMcPMS: What other types would you expect to enumerate?
    enum class PackageTypes : uint32_t
    {
        None = 0,
        Main = 0x1,
        Framework = 0x2,
        Resource = 0x4,
        Bundle = 0x8,
        Xap = 0x10,
        Optional = 0x20,
    };

---
In reply to: [507496922](https://github.com/microsoft/winget-cli/pull/598#discussion_r507496922) [](ancestors = 507496922)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
JohnMcPMS: My expectation was that it would allow `list msix`, but there could be a better way to allow that in the future.

---
In reply to: [507498234](https://github.com/microsoft/winget-cli/pull/598#discussion_r507498234) [](ancestors = 507498234)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
JohnMcPMS: I can tell you that it does return a string in this case (the expected one), so MSDN is wrong here.

---
In reply to: [507500957](https://github.com/microsoft/winget-cli/pull/598#discussion_r507500957) [](ancestors = 507500957,507499499)

JohnMcPMS: And this does in fact (on latest internal selfhost of Windows) give back a localized name:
```
PS C:\git\pkg> wingetdev list store
Name                        Id                                            Version
-------------------------------------------------------------------------------------------
WinGet MSStore Source       Microsoft.Winget.MSStore.Source_8wekyb3d8bbwe 2020.819.2444.404
Microsoft Store 환경 호스트 Microsoft.StorePurchaseApp_8wekyb3d8bbwe      12010.1001.2.0
Microsoft Store             Microsoft.WindowsStore_8wekyb3d8bbwe          12010.1001.2.0
```

---
In reply to: [508136853](https://github.com/microsoft/winget-cli/pull/598#discussion_r508136853) [](ancestors = 508136853,507500957,507499499)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
JohnMcPMS: strstr forever!

Then you can do `strstr.str()` to get a str from your str str.

---
In reply to: [507501616](https://github.com/microsoft/winget-cli/pull/598#discussion_r507501616) [](ancestors = 507501616)

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: If there are multiple sources configured and no specific one is selected on the command line, then the incoming source should be Composite already.  If there is only one source configured, or a single one is selected via `--source`, then it won't be Composite.

---
In reply to: [507556019](https://github.com/microsoft/winget-cli/pull/598#discussion_r507556019) [](ancestors = 507556019)

src/AppInstallerCLITests/PredefinedInstalledSource.cpp
JohnMcPMS: If this is failing, both tests will fail.  If this test passes, but the other does not, we know the issue is in Search (or a non-throwing failure to populate data).

Throwing everything in one test does not provide as much diagnostic data on failures as having multiple tests, even though they may repeat some of the same actions.

---
In reply to: [507576177](https://github.com/microsoft/winget-cli/pull/598#discussion_r507576177) [](ancestors = 507576177)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
JohnMcPMS: I don't think that `list` should.  While upgrade may not automatically "change lanes", I'm sure that we will provide the ability to force it to do so.  If we don't even show that a newer version is available, then how would the user even know that they were falling behind?

---
In reply to: [507683891](https://github.com/microsoft/winget-cli/pull/598#discussion_r507683891) [](ancestors = 507683891)

src/AppInstallerRepositoryCore/CompositeSource.cpp
JohnMcPMS: Indeed; the goal is for upgrade to provide more information about the options, and once one is selected and stored in the winget installation index, to provide a better match behavior.

---
In reply to: [507650355](https://github.com/microsoft/winget-cli/pull/598#discussion_r507650355) [](ancestors = 507650355,507624727)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
yao-msft: haha ok 

---
In reply to: [508138280](https://github.com/microsoft/winget-cli/pull/598#discussion_r508138280) [](ancestors = 508138280,507501616)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
yao-msft: I guess we might enable optional once we have dependencies working?

---
In reply to: [508136073](https://github.com/microsoft/winget-cli/pull/598#discussion_r508136073) [](ancestors = 508136073,507496922)

yao-msft (approved):
:shipit:

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
JohnMcPMS: Added a TODO to consider optional packages

---
In reply to: [508228016](https://github.com/microsoft/winget-cli/pull/598#discussion_r508228016) [](ancestors = 508228016,508136073,507496922)

PR #604, opened 2020-10-09
Create integer primary keys in tables that need to be referenced

JohnMcPMS:
## Motivation
The previous change to create stand-alone indexes rather than primary keys, and subsequently drop more of them when packaging, exposed an issue with our database schema.  SQLite's implicit rowid is not guaranteed to be stable, especially across a `VACUUM`.  As the index was built around referencing other data by it's rowid, this lead to corrupted databases being created when manifests were removed or updated.

## Change
This change causes all relevant tables (every table except the map tables) to have an integer primary key column, which will prevent the rowid from changing, as it is now a data member.

## Validation
A new test is added to ensure that rowid's do not change after a manifest removal and `VACUUM`. It failed prior to the fix being in place, and succeeds afterward.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/604)

### DIFF
--- a/src/AppInstallerCLITests/SQLiteIndex.cpp
+++ b/src/AppInstallerCLITests/SQLiteIndex.cpp
@@ -376,6 +376,68 @@ TEST_CASE("SQLiteIndex_RemoveManifest", "[sqliteindex][V1_0]")
     REQUIRE(Schema::V1_0::CommandsTable::IsEmpty(connection));
 }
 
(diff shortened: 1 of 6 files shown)


msftrubengu (approved):
:shipit:

PR #609, opened 2020-10-13
Create consistency check functionality

JohnMcPMS:
## Change
To ensure at least a base level of consistency within the index, these functions ensure that all referenced rows actually exist.  Failures are logged out if requested.  More extensive checks can be added as needed, although some form of cycle detection in the pathparts is the only thing that I can foresee without additional data being added.  The function is also exported via WinGetUtil.

## Validation
A call is made in the recently added test for rowid migration, as well as adding a test that directly removes a row to ensure failure from the function.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/609)


msftrubengu (approved):
:shipit:

PR #617, opened 2020-10-20
Telemetry reporting improvements

yao-msft:
## Change
- Move the reporting of IsManifestLocal upfront to filter out some noises
- Added execution stage info when failures or exceptions occur.
- Add sub execution telemetry support (for upgrade --all)

## Validation
This should not change the client behavior. Also manually validated basic install and search.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/617)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionContext.h
+++ b/src/AppInstallerCLICore/ExecutionContext.h
@@ -33,7 +33,7 @@
 namespace AppInstaller::CLI::Workflow
 {
     struct WorkflowTask;
-    enum class ExecutionStage;
+    enum class ExecutionStage : uint32_t;
 }
 
 namespace AppInstaller::CLI::Execution
@@ -201,6 +201,12 @@ namespace AppInstaller::CLI::Execution
         void Add(typename details::DataMapping<D>::value_t&& v)
         {
             m_data[D].emplace<details::DataIndex(D)>(std::forward<typename details::DataMapping<D>::value_t>(v));
+        }
+        template <Data D>
+        void Add(const typename details::DataMapping<D>::value_t& v)
+        {
+            auto copy = v;
+            m_data[D].emplace<details::DataIndex(D)>(std::forward<typename details::DataMapping<D>::value_t>(std::move(copy)));
         }
 
         // Return a value indicating whether the given data type is stored in the context.
--- a/src/AppInstallerCLICore/ExecutionContext.h
+++ b/src/AppInstallerCLICore/ExecutionContext.h
@@ -33,6 +33,7 @@
 namespace AppInstaller::CLI::Workflow
 {
     struct WorkflowTask;
+    enum class ExecutionStage;
 }
 
 namespace AppInstaller::CLI::Execution
@@ -53,6 +54,7 @@ namespace AppInstaller::CLI::Execution
         InstallerArgs,
         CompletionData,
         InstalledPackageVersion,
+        ExecutionStage,
         Max
     };
 
@@ -137,6 +139,12 @@ namespace AppInstaller::CLI::Execution
         struct DataMapping<Data::InstalledPackageVersion>
         {
             using value_t = std::shared_ptr<Repository::IPackageVersion>;
+        };
+
+        template <>
+        struct DataMapping<Data::ExecutionStage>
+        {
+            using value_t = Workflow::ExecutionStage;
         };
 
         // Used to deduce the DataVariant type; making a variant that includes std::monostate and all DataMapping types.
--- a/src/AppInstallerCLICore/ExecutionContext.h
+++ b/src/AppInstallerCLICore/ExecutionContext.h
@@ -205,8 +205,7 @@ namespace AppInstaller::CLI::Execution
         template <Data D>
         void Add(const typename details::DataMapping<D>::value_t& v)
         {
-            auto copy = v;
-            m_data[D].emplace<details::DataIndex(D)>(std::forward<typename details::DataMapping<D>::value_t>(std::move(copy)));
+            m_data[D].emplace<details::DataIndex(D)>(v);
         }
 
         // Return a value indicating whether the given data type is stored in the context.
--- a/src/AppInstallerCLICore/Workflows/WorkflowBase.h
+++ b/src/AppInstallerCLICore/Workflows/WorkflowBase.h
@@ -19,13 +19,13 @@ namespace AppInstaller::CLI::Workflow
     static const char* s_InstallationMetadata_Key_InstallerType = "InstallerType";
 
     // Values are ordered in a typical workflow stages
-    enum class ExecutionStage
+    enum class ExecutionStage : uint32_t
     {
-        None,
-        Discovery,
-        Download,
-        Execution,
-        PostExecution,
+        ParseArgs = 1000,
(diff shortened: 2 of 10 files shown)


src/AppInstallerCLICore/Workflows/WorkflowBase.h
```diff
@@ -18,6 +18,16 @@ namespace AppInstaller::CLI::Workflow
 {
     static const char* s_InstallationMetadata_Key_InstallerType = "InstallerType";
 
+    // Values are ordered in a typical workflow stages
+    enum class ExecutionStage
+    {
+        None,
```
JohnMcPMS, line 24: >None [](start = 8, length = 4)

The first phase should probably be something about command line parsing.

src/AppInstallerCLICore/Workflows/WorkflowBase.h
```diff
@@ -197,6 +207,24 @@ namespace AppInstaller::CLI::Workflow
...
+    // Inputs: CompletionData, Source
+    // Outputs: None
+    struct ReportExecutionStage : public WorkflowTask
+    {
+        ReportExecutionStage(ExecutionStage stage, bool allowBackward = false) : WorkflowTask("ReportExecutionStage"), m_stage(stage), m_allowBackward(allowBackward) {}
+
+        void operator()(Execution::Context& context) const override;
+
+    private:
+        mutable ExecutionStage m_stage;
```
JohnMcPMS, line 222: >mutable [](start = 8, length = 7)

This doesn't need mutable; you aren't writing to it in a const context. (And the move you are doing is completely unnecessary)

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -35,6 +35,10 @@ namespace AppInstaller::Logging
         // Used to disable telemetry on the fly.
         std::atomic_bool s_isTelemetryEnabled{ true };
 
+        std::string s_executionStage;
```
JohnMcPMS, line 38: >std::string s_executionStage; [](start = 8, length = 29)

I think that while logging a string will be more flexible for adding new stages, it will make processing that data far harder.  Instead, we should just log an integer, and the current stages can be ParseArgs = 1000, Discovery = 2000, etc.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -140,6 +147,7 @@ namespace AppInstaller::Logging
                 "CommandFound",
                 GetActivityId(),
                 nullptr,
+                TraceLoggingGuid(s_subExecutionId, "SubExecutionId"),
```
JohnMcPMS, line 150: Not needed

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -156,6 +164,7 @@ namespace AppInstaller::Logging
                 "CommandSuccess",
                 GetActivityId(),
                 nullptr,
+                TraceLoggingGuid(s_subExecutionId, "SubExecutionId"),
```
JohnMcPMS, line 167: Not needed

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -35,6 +35,10 @@ namespace AppInstaller::Logging
         // Used to disable telemetry on the fly.
         std::atomic_bool s_isTelemetryEnabled{ true };
 
+        std::string s_executionStage;
+
+        std::atomic<GUID> s_subExecutionId{ GUID_NULL };
```
JohnMcPMS, line 40: >std::atomic<GUID> s_subExecutionId{ GUID_NULL }; [](start = 8, length = 48)

Could we instead just use an integer?  These don't need to be unique across sessions, so for the same ActivityId, 0 can be the root (like GUID_NULL), and 1+ can be a subexecution.  That way we also get an order that might be useful.

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerCLICore/Workflows/WorkflowBase.h
```diff
@@ -197,6 +207,24 @@ namespace AppInstaller::CLI::Workflow
     // Inputs: SearchResult
     // Outputs: InstalledPackageVersion
     void GetInstalledPackageVersion(Execution::Context& context);
+
+    // Searches the source for the specific field as a completion.
```
yao-msft, line 211: >/ Searches the source fo [](start = 5, length = 24)

update comments

src/AppInstallerCLICore/Workflows/WorkflowBase.h
yao-msft: Yea, but the compiler is not happy if I don't use std::move() since the context.Add() only takes rvalue.

---
In reply to: [508324457](https://github.com/microsoft/winget-cli/pull/617#discussion_r508324457) [](ancestors = 508324457)

JohnMcPMS: Then fix Add to have an lvalue version.

---
In reply to: [508751833](https://github.com/microsoft/winget-cli/pull/617#discussion_r508751833) [](ancestors = 508751833,508324457)

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -405,4 +440,22 @@ namespace AppInstaller::Logging
...
+    void SetExecutionStage(uint32_t stage)
+    {
+        s_executionStage = stage;
+    }
+
+    std::atomic_uint32_t SubExecutionTelemetryScope::m_sessionId{ s_RootExecutionId };
+
+    SubExecutionTelemetryScope::SubExecutionTelemetryScope()
+    {
+        THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), s_subExecutionId.exchange(++m_sessionId) != s_RootExecutionId,
```
JohnMcPMS, line 453: >exchange [](start = 82, length = 8)

Use compare_exchange to not actually change the value if it is not root.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -194,6 +202,12 @@ namespace AppInstaller::CLI::Execution
         {
             m_data[D].emplace<details::DataIndex(D)>(std::forward<typename details::DataMapping<D>::value_t>(v));
         }
+        template <Data D>
+        void Add(const typename details::DataMapping<D>::value_t& v)
+        {
+            auto copy = v;
+            m_data[D].emplace<details::DataIndex(D)>(std::forward<typename details::DataMapping<D>::value_t>(std::move(copy)));
```
JohnMcPMS, line 209: >std::move(copy) [](start = 109, length = 15)

You shouldn't need to move a copy; why can't you just emplace v and have it do the copy?

JohnMcPMS (approved):
:shipit:

PR #623, opened 2020-10-22
Move the default location for installer logs

JohnMcPMS:
## Change
Moves the default log location for installer logs to be the same as the location for `winget` logs, which on a default install is:
```
%LOCALAPPDATA%\Packages\WinGetDevCLI_8wekyb3d8bbwe\LocalState\DiagOutputDir
```
This will allow reports through [Feedback Hub](https://aka.ms/winget-feedback) to gather those logs to better help us diagnose issues with specific installers.

Additionally, `winget --info` will now also show this log location.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/623)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -219,6 +219,7 @@ jobs:
                          /p:AppxBundle=Never 
                          /p:UapAppxPackageBuildMode=SideLoadOnly 
                          /p:AppxPackageSigningEnabled=false'
+    condition: succeededOrFailed()
 
   - task: PowerShell@2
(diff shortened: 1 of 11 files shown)


ranm-msft (approved):
can we add tests to validate location?

JohnMcPMS:
> can we add tests to validate location?

There are tests that validate the location, and I even forgot to update them and they failed.  So they are doing their job 😄

PR #625, opened 2020-10-22
Improve Motw related experience

yao-msft:
## Change
- Replace Motw writing logic to use IZoneIdentifier instead of previous direct writing
- For installers from trusted winget sources, set zone to trusted if installer hash matches. Otherwise, let Attachment Execution Service to determine security zone.

## Validation
Added test. And manual verification by setting break point in download step.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/625)


src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -353,6 +353,33 @@ namespace AppInstaller::Runtime
         return wil::test_token_membership(nullptr, SECURITY_NT_AUTHORITY, SECURITY_BUILTIN_DOMAIN_RID, DOMAIN_ALIAS_RID_ADMINS);
     }
 
+    bool IsNTFS(const std::filesystem::path& filePath)
+    {
+        wil::unique_hfile fileHandle{ CreateFileW(
+            filePath.c_str(), /*lpFileName*/
+            GENERIC_READ, /*dwDesiredAccess*/
```
JohnMcPMS, line 360: >GENERIC_READ [](start = 12, length = 12)

I think we could ask for even less access than this to improve usability outside of this specific scenario.

src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -171,51 +171,64 @@ namespace AppInstaller::Utility
...
+            return;
+        }
+
+        // Attachment execution service needs STA to succeed, so we'll create a new thread and CoInitialize with STA.
+        auto updateMotw = [&]() -> HRESULT
+        {
+            Microsoft::WRL::ComPtr<IAttachmentExecute> attachmentExecute;
+            RETURN_IF_FAILED(CoCreateInstance(CLSID_AttachmentServices, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&attachmentExecute)));
+            RETURN_IF_FAILED(attachmentExecute->SetLocalPath(filePath.c_str()));
+            RETURN_IF_FAILED(attachmentExecute->SetSource(Utility::ConvertToUTF16(source).c_str()));
```
JohnMcPMS, line 211: >SetSource [](start = 48, length = 9)

Does this end up saving the source string into the alternate stream?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -189,6 +191,30 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
+    void UpdateInstallerFileMotwIfApplicable(Execution::Context& context)
+    {
+        const std::string TrustedSourceIds[] = { "Microsoft.Winget.Source_8wekyb3d8bbwe" };
```
JohnMcPMS, line 196: >const std::string TrustedSourceIds[] = { "Microsoft.Winget.Source_8wekyb3d8bbwe" }; [](start = 8, length = 83)

SourceDetails should be added to so that we can set trust levels on arbitrary sources.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -47,6 +47,7 @@ namespace AppInstaller::CLI::Execution
         SearchResult,
         SourceList,
         Manifest,
+        SourceId,
```
JohnMcPMS, line 50: >SourceId [](start = 8, length = 8)

This value can easily be retrieved from Source; why do we need to copy it into the context?

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -57,6 +57,7 @@
 #define APPINSTALLER_CLI_ERROR_INVALID_MANIFEST                 ((HRESULT)0x8A15002A)
 #define APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE            ((HRESULT)0x8A15002B)
 #define APPINSTALLER_CLI_ERROR_UPDATE_ALL_HAS_FAILURE           ((HRESULT)0x8A15002C)
+#define APPINSTALLER_CLI_ERROR_REMOVE_INSTALLER_FAILED          ((HRESULT)0x8A15002D)
```
JohnMcPMS, line 60: >APPINSTALLER_CLI_ERROR_REMOVE_INSTALLER_FAILED [](start = 8, length = 46)

There is a switch statement with error strings that is in the cpp for this file.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -273,7 +299,22 @@ namespace AppInstaller::CLI::Workflow
...
+
+            try
+            {
+                // best effort
+                std::filesystem::remove(path);
+            }
+            catch (const std::exception& e)
+            {
+                AICLI_LOG(CLI, Warning, << "Failed to remove installer file after execution. Reason: " << e.what());
+                AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_REMOVE_INSTALLER_FAILED);
```
JohnMcPMS, line 311: >APPINSTALLER_CLI_ERROR_REMOVE_INSTALLER_FAILED [](start = 40, length = 46)

I'm not sure that this should be an error, at most it seems like it should result in a warning output to the Reporter.

src/AppInstallerCLICore/ExecutionContext.h
JohnMcPMS: I see better now, but I think that the better solution might still be to have the IPackageVersion in the context, and create an IPackageVersion::GetSource method.

---
In reply to: [510503004](https://github.com/microsoft/winget-cli/pull/625#discussion_r510503004) [](ancestors = 510503004)

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerCommonCore/Runtime.cpp
yao-msft: [https://docs.microsoft.com/en-us/windows/win32/secauthz/generic-access-rights](https://docs.microsoft.com/en-us/windows/win32/secauthz/generic-access-rights)

Constant Generic meaning
GENERIC_ALL All possible access rights
GENERIC_EXECUTE Execute access
GENERIC_READ Read access
GENERIC_WRITE Write access

I think GENERIC_READ is the least access

---
In reply to: [510499353](https://github.com/microsoft/winget-cli/pull/625#discussion_r510499353) [](ancestors = 510499353)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
yao-msft: I can rename the error code to be APPINSTALLER_CLI_WARNING... and output a warning message. Definitely I don't want E_ACCESSDENIED in the reports to be mixed with others

---
In reply to: [510504068](https://github.com/microsoft/winget-cli/pull/625#discussion_r510504068) [](ancestors = 510504068)

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
yao-msft: yea, I can add to that for completeness. But code paths using AICLI_TERMINATE does not invoke that GetFriendlyNameForErrorCode() method. Code paths using AICLI_TERMINATE should handle output message individually.

---
In reply to: [510503586](https://github.com/microsoft/winget-cli/pull/625#discussion_r510503586) [](ancestors = 510503586)

src/AppInstallerCommonCore/Downloader.cpp
yao-msft: Sent to internal discussion

---
In reply to: [510501417](https://github.com/microsoft/winget-cli/pull/625#discussion_r510501417) [](ancestors = 510501417)

src/AppInstallerCommonCore/Runtime.cpp
JohnMcPMS: Of the GENERIC ones, sure, but FILE_READ_ATTRIBUTES is certainly less than GENERIC_READ, as it is one part of it.

[https://docs.microsoft.com/en-us/windows/win32/fileio/file-security-and-access-rights](https://docs.microsoft.com/en-us/windows/win32/fileio/file-security-and-access-rights)

---
In reply to: [510523433](https://github.com/microsoft/winget-cli/pull/625#discussion_r510523433) [](ancestors = 510523433,510499353)

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
JohnMcPMS: I agree, it was simply for completeness.

---
In reply to: [510541016](https://github.com/microsoft/winget-cli/pull/625#discussion_r510541016) [](ancestors = 510541016,510503586)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: I am suggesting that we simply move on without an error of any kind.  We could log to telemetry if we want, but I don't think a failure to clean up fully warrants an error.  Full OS upgrade (not the little CBS ones) will clean files out, and if it does become a problem we can put in some cleanup like I did for logs.

---
In reply to: [510525154](https://github.com/microsoft/winget-cli/pull/625#discussion_r510525154) [](ancestors = 510525154,510504068)

yao-msft: Removed the error code and ignored this failure, we can add back reporting if needed in the future

---
In reply to: [511105679](https://github.com/microsoft/winget-cli/pull/625#discussion_r511105679) [](ancestors = 511105679,510525154,510504068)

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
```diff
@@ -121,6 +123,9 @@ namespace AppInstaller::Repository
         // Gets the manifest of this package version.
         virtual Manifest::Manifest GetManifest() const = 0;
 
+        // Gets the source where this package version is from.
```
JohnMcPMS, line 126: >// Gets the source where this package version is from. [](start = 8, length = 54)

This comment is actually what I was thinking of, getting the *source*:

virtual std::shared_ptr<ISource> GetSource() const = 0;

JohnMcPMS (approved):
:shipit:

PR #626, opened 2020-10-22
Implement manifest metadata in index

JohnMcPMS:
## Change
In order to store arbitrary amounts of data about installed packages, we need a more flexible table.  It isn't needed for available packages (yet), and so is only created on first write.  The data that is stored is { manifest rowid, metadata enum, value string }.  This part is very straightforward.

In order for the installed package enumeration for `list` et al. to be able to populate this, I realized it needed to be an enum stored in the repository code (sorry @yao-msft).  So part of this change is moving everything over to that.  In addition, I fixed some issues while I was around the code changing to an enum.

Finally, I hooked up the current MSIX enumeration to the only metadata item currently in use, and the SQLiteIndex source to read the metadata rather than just return an empty result every time.

## Validation
A test is added to set and read metadata for a manifest.  Existing regression tests cover other cases.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/626)


src/AppInstallerCLIE2ETests/TestCommon.cs
```diff
@@ -156,9 +156,32 @@ public static RunCommandResult RunAICLICommandViaInvokeCommandInDesktopPackage(s
...
+                bool success = false;
+
+                try
+                {
+                    result.ExitCode = File.Exists(exitCodeFile) ? int.Parse(File.ReadAllText(exitCodeFile).Trim()) : unchecked((int)0x80004005);
+                    result.StdOut = File.Exists(stdOutFile) ? File.ReadAllText(stdOutFile) : "";
+                    result.StdErr = File.Exists(stdErrFile) ? File.ReadAllText(stdErrFile) : "";
+                    success = true;
+                }
+                catch (System.Exception e)
```
yao-msft, line 171: >System [](start = 23, length = 6)

nit: already using System

src/AppInstallerCommonCore/Architecture.cpp
```diff
@@ -46,7 +83,6 @@ namespace AppInstaller::Utility
         switch (systemInfo.wProcessorArchitecture)
         {
         case PROCESSOR_ARCHITECTURE_AMD64:
-        case PROCESSOR_ARCHITECTURE_IA64:
```
yao-msft, line 49: >PROCESSOR_ARCHITECTURE_IA64 [](start = 13, length = 27)

Is this removal intended?

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -9,86 +9,125 @@ using namespace AppInstaller::Manifest;
...
 namespace AppInstaller::CLI::Workflow
 {
-    bool InstallerComparator::operator() (const ManifestInstaller& installer1, const ManifestInstaller& installer2)
+    namespace
     {
-        // Applicable architecture should always come before inapplicable architecture
-        if (Utility::IsApplicableArchitecture(installer1.Arch) != Utility::InapplicableArchitecture &&
-            Utility::IsApplicableArchitecture(installer2.Arch) == Utility::InapplicableArchitecture)
+        // Determine if the installer is applicable.
+        bool IsInstallerApplicable(const Manifest::ManifestInstaller& installer, Manifest::ManifestInstaller::InstallerTypeEnum currentType)
```
yao-msft, line 15: >currentType [](start = 128, length = 11)

installedType is more intuitive

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -9,86 +9,125 @@ using namespace AppInstaller::Manifest;
...
 
-        // If there's installation metadata, pick the preferred one or compatible one
-        auto installerTypeItr = m_installationMetadata.find(s_InstallationMetadata_Key_InstallerType);
-        if (installerTypeItr != m_installationMetadata.end())
+        // This is used in sorting the list of available installers to get the best match.
+        // Determines if installer1 is a better match than installer2.
+        bool IsInstallerBetterMatch(
+            const Manifest::ManifestInstaller& installer1,
+            const Manifest::ManifestInstaller& installer2,
+            Manifest::ManifestInstaller::InstallerTypeEnum installerType)
```
yao-msft, line 36: >installerType [](start = 59, length = 13)

installedType

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -9,86 +9,125 @@ using namespace AppInstaller::Manifest;
...
         AICLI_LOG(CLI, Info, << "Starting installer selection.");
 
-        // Sorting the list of available installers according to rules defined in InstallerComparator.
-        auto installers = manifest.Installers;
-        std::sort(installers.begin(), installers.end(), m_installerComparator);
-
-        // If the first one's architecture is inapplicable, then no installer is applicable.
-        if (Utility::IsApplicableArchitecture(installers[0].Arch) == Utility::InapplicableArchitecture)
+        // Get the currently installed package's type (if present)
+        Manifest::ManifestInstaller::InstallerTypeEnum installerType = Manifest::ManifestInstaller::InstallerTypeEnum::Unknown;
```
yao-msft, line 99: >installerType [](start = 55, length = 13)

installedType

yao-msft (approved):
:shipit:

src/AppInstallerCommonCore/Architecture.cpp
```diff
@@ -8,30 +8,67 @@
...
+    namespace
+    {
+        // Gets the applicable architectures for the current machine.
+        std::vector<Architecture> CreateApplicableArchitecturesVector()
+        {
+            std::vector<Architecture> applicableArchs;
+
+            switch (GetSystemArchitecture())
+            {
+            case Architecture::Arm64:
```
JohnMcPMS, line 20: Use IsWow64GuestMachineSupported to truly determine if WOW architectures are enabled.

src/AppInstallerCommonCore/Architecture.cpp
JohnMcPMS: Yes, IA64 is Itanium, which is *not* AMD64.  If anything it would be its own value, but I don't think we need to support it unless there is large customer demand.

---
In reply to: [511303758](https://github.com/microsoft/winget-cli/pull/626#discussion_r511303758) [](ancestors = 511303758)

PR #627, opened 2020-10-29
Update Smart Screen experience for msix

yao-msft:
## Change
- Use AddPackageAsync to bypass smart screen when adding or updating trusted source
- Use AddPackageAsync to bypass smart screen when deploying msix from trusted source

## Validation
Modified unit test to also verify IsInstallerTrusted flag

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/627)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionContext.h
+++ b/src/AppInstallerCLICore/ExecutionContext.h
@@ -230,7 +230,7 @@ namespace AppInstaller::CLI::Execution
         }
 
         // Gets context flags
-        ContextFlag GetFlags()
+        ContextFlag GetFlags() const
         {
             return m_flags;
         }
--- a/src/AppInstallerCLICore/ExecutionContext.h
+++ b/src/AppInstallerCLICore/ExecutionContext.h
@@ -64,6 +64,8 @@ namespace AppInstaller::CLI::Execution
     {
         None = 0x0,
         InstallerExecutionUseUpdate = 0x1,
+        InstallerHashMatched = 0x2,
+        InstallerTrusted = 0x4,
     };
 
     DEFINE_ENUM_FLAG_OPERATORS(ContextFlag);
--- a/src/AppInstallerCLICore/ExecutionContext.h
+++ b/src/AppInstallerCLICore/ExecutionContext.h
@@ -229,10 +229,22 @@ namespace AppInstaller::CLI::Execution
             return std::get<details::DataIndex(D)>(itr->second);
         }
 
-        // Gets context flags; which can be modified in place.
-        ContextFlag& GetFlags()
+        // Gets context flags
+        ContextFlag GetFlags()
         {
             return m_flags;
+        }
+
+        // Set context flags
+        void SetFlags(ContextFlag flags)
+        {
+            WI_SetAllFlags(m_flags, flags);
+        }
+
+        // Clear context flags
+        void ClearFlags(ContextFlag flags)
+        {
+            WI_ClearAllFlags(m_flags, flags);
         }
 
 #ifndef AICLI_DISABLE_TEST_HOOKS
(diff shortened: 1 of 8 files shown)


src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -68,17 +68,17 @@ namespace AppInstaller::CLI::Workflow
         case ManifestInstaller::InstallerTypeEnum::Msi:
         case ManifestInstaller::InstallerTypeEnum::Nullsoft:
         case ManifestInstaller::InstallerTypeEnum::Wix:
-            context << DownloadInstallerFile << VerifyInstallerHash;
+            context << DownloadInstallerFile << VerifyInstallerHash << UpdateInstallerFileMotwIfApplicable;
```
JohnMcPMS, line 71: >UpdateInstallerFileMotwIfApplicable [](start = 71, length = 35)

Why not move VerifyInstallerHash down, and make MSStore return instead of break?  I don't think even more duplication is really the right answer.

Also, does this mean that Store installs were broken?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -189,28 +187,30 @@ namespace AppInstaller::CLI::Workflow
         {
             AICLI_LOG(CLI, Info, << "Installer hash verified");
             context.Reporter.Info() << Resource::String::InstallerHashVerified << std::endl;
+
+            WI_SetFlag(context.GetFlags(), Execution::ContextFlag::InstallerHashMatched);
```
JohnMcPMS, line 191: >WI_SetFlag [](start = 12, length = 10)

If this is suddenly going to become this prevalent, can we actually have GetFlags return by value, and have a SetFlag method?  I'm just imagining a desire to update said function with logging or some other behavior and that is so much harder when the assignments happen out here.

src/AppInstallerCommonCore/Public/AppInstallerDeployment.h
```diff
@@ -7,6 +7,13 @@
 
 namespace AppInstaller::Deployment
 {
+    // Calls winrt::Windows::Management::Deployment::PackageManager::AddPackageAsync
+    // which does not trigger smart screen scan
+    void AddPackage(
```
JohnMcPMS, line 12: >AddPackage [](start = 9, length = 10)

It would be better to have a single function in our code that takes in an additional bool for whether SmartScreen should be called.  Having every call site need to have a `if (trust) call one; else call other;` is overly verbose, and potentially leads to bugs.

Think about the next time someone adds an AddPackage type call, and whether they will remember to call both depending on some trust level.  If the function demands the parameter, they will stop and think about it and make a choice.  If comes down to which function they call, they may not be aware of the choice at all.

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
yao-msft: Store install is not broken. The first line in UpdateInstallerFileMotwIfApplicable checks if there's a downloaded file, so it's technically skipped in store scenario.

I moved up just to be more explicit that each step is applicable to which type. I think moving down and let individual type (i.e. msstore) opt-out by using return is more likely to be overlooked in the future;

---
In reply to: [514492065](https://github.com/microsoft/winget-cli/pull/627#discussion_r514492065) [](ancestors = 514492065)

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -227,12 +229,24 @@ namespace AppInstaller::CLI::Execution
             return std::get<details::DataIndex(D)>(itr->second);
         }
 
-        // Gets context flags; which can be modified in place.
-        ContextFlag& GetFlags()
+        // Gets context flags
+        ContextFlag GetFlags()
```
JohnMcPMS, line 233: >ContextFlag GetFlags() [](start = 8, length = 22)

Can now be const

JohnMcPMS (approved):
:shipit:

PR #629, opened 2020-10-29
Update spec for PackageFamilyName and ProductCode

denelon:


JohnMcPMS:
`PackageFamilyName` and `ProductCode` can both exist in the root as the default value for installers whose `InstallerType` is appropriate.

JohnMcPMS (approved):

PR #634, opened 2020-11-04
Improve upgrade output experience

yao-msft:
## Change
- Previously, upgrade flow assumes at least one available version is there for each search match from composite installed source. It's not the case. So in this change, all places calling GetLatestAvailableVersion() are examined and made sure there's proper null check.
- Fixed output spamming in upgrade --all case
- Fixed an issue where any user source operation will restore tombstoned default source
- Always log E_ABORT context termination to reduce noise

## Validation
- Updated tests
- Manual validation and fix

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/634)

### DIFF
--- a/src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
@@ -14,6 +14,15 @@ using namespace AppInstaller::CLI::Workflow;
 
 namespace AppInstaller::CLI
 {
+    namespace
+    {
+        bool ShouldListUpgrade(Context& context)
+        {
+            return context.Args.Empty() ||
+                (context.Args.GetArgsCount() == 1 && context.Args.Contains(Execution::Args::Type::Source));
+        }
+    }
+
     std::vector<Argument> UpgradeCommand::GetArguments() const
     {
         return {
@@ -120,13 +129,12 @@ namespace AppInstaller::CLI
             OpenSource <<
             OpenCompositeSource(Repository::PredefinedSource::Installed);
 
-        if (context.Args.Empty() ||
-            (context.Args.GetArgsCount() == 1 && context.Args.Contains(Execution::Args::Type::Source)))
+        if (ShouldListUpgrade(context))
         {
             // Upgrade with no args list packages with updates available
             context <<
                 Workflow::SearchSourceForMany <<
-                Workflow::EnsureMatchesFromSearchResult <<
+                Workflow::EnsureMatchesFromSearchResult(true) <<
                 Workflow::ReportListResult(true);
         }
         else if (context.Args.Contains(Execution::Args::Type::All))
@@ -134,7 +142,7 @@ namespace AppInstaller::CLI
             // --all switch updates all packages found
             context <<
                 SearchSourceForMany <<
-                EnsureMatchesFromSearchResult <<
+                EnsureMatchesFromSearchResult(true) <<
                 UpdateAllApplicable;
         }
         else if (context.Args.Contains(Execution::Args::Type::Manifest))
@@ -144,7 +152,7 @@ namespace AppInstaller::CLI
                 GetManifestFromArg <<
                 ReportManifestIdentity <<
                 SearchSourceUsingManifest <<
-                EnsureOneMatchFromSearchResult <<
+                EnsureOneMatchFromSearchResult(true) <<
                 GetInstalledPackageVersion <<
                 EnsureUpdateVersionApplicable <<
                 EnsureMinOSVersion <<
@@ -163,20 +171,15 @@ namespace AppInstaller::CLI
             // The remaining case: search for single installed package to update
             context <<
                 SearchSourceForSingle <<
-                EnsureOneMatchFromSearchResult <<
-                ReportSearchResultIdentity <<
+                EnsureOneMatchFromSearchResult(true) <<
+                ReportPackageIdentity <<
                 GetInstalledPackageVersion;
-
-            if (context.IsTerminated())
-            {
-                return;
-            }
 
             if (context.Args.Contains(Execution::Args::Type::Version))
             {
                 // If version specified, use the version and verify applicability
                 context <<
-                    GetManifestFromSearchResult <<
+                    GetManifestFromPackage <<
                     EnsureUpdateVersionApplicable <<
                     EnsureMinOSVersion <<
                     SelectInstaller <<
@@ -186,13 +189,7 @@ namespace AppInstaller::CLI
             {
                 // iterate through available versions to find latest applicable update
                 // This step also populates Manifest and Installer in context data
-                context <<
-                    SelectLatestApplicableUpdate(*(context.Get<Execution::Data::SearchResult>().Matches.at(0).Package));
-
-                if (context.GetTerminationHR() == APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE)
-                {
-                    context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl;
-                }
+                context << SelectLatestApplicableUpdate(true);
             }
 
             context <<
--- a/src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
@@ -120,7 +120,8 @@ namespace AppInstaller::CLI
             OpenSource <<
             OpenCompositeSource(Repository::PredefinedSource::Installed);
 
-        if (context.Args.Empty())
+        if (context.Args.Empty() ||
+            (context.Args.GetArgsCount() == 1 && context.Args.Contains(Execution::Args::Type::Source)))
         {
             // Upgrade with no args list packages with updates available
             context <<
@@ -166,6 +167,11 @@ namespace AppInstaller::CLI
                 ReportSearchResultIdentity <<
                 GetInstalledPackageVersion;
 
+            if (context.IsTerminated())
+            {
+                return;
+            }
+
             if (context.Args.Contains(Execution::Args::Type::Version))
             {
                 // If version specified, use the version and verify applicability
@@ -182,6 +188,11 @@ namespace AppInstaller::CLI
                 // This step also populates Manifest and Installer in context data
                 context <<
                     SelectLatestApplicableUpdate(*(context.Get<Execution::Data::SearchResult>().Matches.at(0).Package));
+
+                if (context.GetTerminationHR() == APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE)
+                {
+                    context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl;
+                }
             }
 
             context <<
--- a/src/AppInstallerCLICore/ExecutionContext.cpp
+++ b/src/AppInstallerCLICore/ExecutionContext.cpp
@@ -116,14 +116,12 @@ namespace AppInstaller::CLI::Execution
             // Unless we want to spin a separate thread for all work, we have to just exit here.
             if (m_CtrlSignalCount >= 2)
             {
+                Logging::Telemetry().LogCommandTermination(hr, file, line);
                 std::exit(hr);
             }
         }
 
-        if (!file.empty() && line)
-        {
-            Logging::Telemetry().LogCommandTermination(hr, file, line);
-        }
+        Logging::Telemetry().LogCommandTermination(hr, file, line);
 
         m_isTerminated = true;
         m_terminationHR = hr;
--- a/src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
@@ -122,7 +122,9 @@ namespace AppInstaller::CLI::Workflow
 
             updateContext.Reporter.Info() << std::endl;
 
-            if (updateContext.GetTerminationHR() != S_OK)
+            // msstore update might still terminate with APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE
+            if (updateContext.GetTerminationHR() != S_OK &&
+                updateContext.GetTerminationHR() != APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE)
             {
                 updateAllHasFailure = true;
             }
--- a/src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
@@ -66,7 +66,6 @@ namespace AppInstaller::CLI::Workflow
 
         if (!updateFound)
         {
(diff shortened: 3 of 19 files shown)


src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -116,14 +116,12 @@ namespace AppInstaller::CLI::Execution
             // Unless we want to spin a separate thread for all work, we have to just exit here.
             if (m_CtrlSignalCount >= 2)
             {
+                Logging::Telemetry().LogCommandTermination(hr, file, line);
                 std::exit(hr);
             }
         }
 
-        if (!file.empty() && line)
-        {
```
JohnMcPMS, line 124: Why are these removed?  I can't remember why I added them, but I don't think it was for no reason (it actually might have been to prevent the first CTRL-C from being logged).

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -120,7 +120,8 @@ namespace AppInstaller::CLI
             OpenSource <<
             OpenCompositeSource(Repository::PredefinedSource::Installed);
 
-        if (context.Args.Empty())
+        if (context.Args.Empty() ||
+            (context.Args.GetArgsCount() == 1 && context.Args.Contains(Execution::Args::Type::Source)))
```
JohnMcPMS, line 124: This should be a function with a semanticly relevant name to help future proof it when new arguments are added.  For instance `ShouldListUpgrades`.

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -166,6 +167,11 @@ namespace AppInstaller::CLI
                 ReportSearchResultIdentity <<
                 GetInstalledPackageVersion;
 
+            if (context.IsTerminated())
+            {
+                return;
+            }
```
JohnMcPMS, line 173: This should never be needed, as the rest of the output operators will simply be skipped.

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -182,6 +188,11 @@ namespace AppInstaller::CLI
                 // This step also populates Manifest and Installer in context data
                 context <<
                     SelectLatestApplicableUpdate(*(context.Get<Execution::Data::SearchResult>().Matches.at(0).Package));
+
+                if (context.GetTerminationHR() == APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE)
+                {
+                    context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl;
+                }
```
JohnMcPMS, line 195: This should happen when it is determined, not here.

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
```diff
@@ -66,7 +66,6 @@ namespace AppInstaller::CLI::Workflow
 
         if (!updateFound)
         {
-            context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl;
```
JohnMcPMS, line 69: >context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl; [](start = 12, length = 78)

Take in a bool to determine whether to output or not.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -174,6 +174,14 @@ namespace AppInstaller::CLI::Workflow
...
         // Overwrite the source with the composite.
         context.Add<Execution::Data::Source>(std::move(compositeSource));
+
+        if (m_predefinedSource == PredefinedSource::ARP_System ||
+            m_predefinedSource == PredefinedSource::ARP_User ||
+            m_predefinedSource == PredefinedSource::MSIX ||
+            m_predefinedSource == PredefinedSource::Installed)
+        {
+            context.SetFlags(Execution::ContextFlag::CompositeInstalledSource);
+        }
```
JohnMcPMS, line 184: Why not `ISource::IsComposite`?

JohnMcPMS:
>     void EnsureMatchesFromSearchResult(Execution::Context& context)

I think it would be better to make this a workflow functor and take in the error message to use.

---
Refers to: src/AppInstallerCLICore/Workflows/WorkflowBase.cpp:393 in 20b560b. [](commit_id = 20b560b3f1ab1b537a614160d6d97a3315151b5d, deletion_comment = False)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -371,6 +379,11 @@ namespace AppInstaller::CLI::Workflow
 
         table.Complete();
 
+        if (table.IsEmpty())
+        {
+            context.Reporter.Info() << Resource::String::NoInstalledPackageFound << std::endl;
+        }
```
JohnMcPMS, line 385: Are there cases of this not outputting a message?  It currently does so when nothing is found.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -410,21 +423,22 @@ namespace AppInstaller::CLI::Workflow
             {
                 Logging::Telemetry().LogMultiAppMatch();
 
-                if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerExecutionUseUpdate))
+                if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::CompositeInstalledSource))
                 {
                     context.Reporter.Warn() << Resource::String::MultipleInstalledPackagesFound << std::endl;
+                    context << ReportListResult(false);
```
JohnMcPMS, line 429: >context << ReportListResult(false); [](start = 20, length = 35)

I don't think we want this.  See `ReportListResult` and `IPackage::GetProperty` and create another method to report these with only Name and Id; I don't think version is very important here, and both cases are used to simply give the options of things that were found.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -49,6 +49,12 @@ namespace AppInstaller::Repository
             return std::find_if(sources.begin(), sources.end(), [&name](const SourceDetailsInternal& sd) { return Utility::ICUCaseInsensitiveEquals(sd.Name, name); });
         }
 
+        // Finds a source from the given vector by its name. Returned source cannot be modified.
+        auto FindSourceByNameConst(const std::vector<SourceDetailsInternal>& sources, std::string_view name)
```
JohnMcPMS, line 53: >FindSourceByNameConst [](start = 13, length = 21)

There shouldn't be a need to make a new function; just make the one above take a const vector.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -307,6 +313,22 @@ namespace AppInstaller::Repository
                 }
             }
 
+            // Copy tombstoned default sources info
+            for (const auto& userSource : GetSourcesByOrigin(SourceOrigin::User))
```
JohnMcPMS, line 317: I think it would be better to fix this by adding a flag to request that tombstones be preserved in the Get operation.  Then modifications can work off of that list, thus preserving the tombstones without a need to copy like this.

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerCLICore/ExecutionContext.cpp
yao-msft: yes. The initial intention is to not log ctrl_c. But we want ctrl_c to be logged explicitly to reduce noise

---
In reply to: [518297562](https://github.com/microsoft/winget-cli/pull/634#discussion_r518297562) [](ancestors = 518297562)

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
yao-msft: The below:
 context << SelectLatestApplicableUpdate(*(context.Get<Execution::Data::SearchResult>().Matches.at(0).Package));

will construct the workflowtask first which throws invalid vector position 

Do you want to check size of context.Get<Execution::Data::SearchResult>().Matches instead below?

---
In reply to: [518308965](https://github.com/microsoft/winget-cli/pull/634#discussion_r518308965) [](ancestors = 518308965)

src/AppInstallerRepositoryCore/RepositorySource.cpp
yao-msft: I tried initially to add const to input vector. Then I found existing code used FindSourceByName() and then modifys the source details pointed by returned itr. So I created this const version one.

---
In reply to: [518347549](https://github.com/microsoft/winget-cli/pull/634#discussion_r518347549) [](ancestors = 518347549)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
yao-msft: searching across multiple sources like "winget install foo" when there are more than 2 sources configured will make ISource::IsComposite() to be true, but we want installed composite source for our case

---
In reply to: [518336942](https://github.com/microsoft/winget-cli/pull/634#discussion_r518336942) [](ancestors = 518336942)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
yao-msft: We output "no package found" during EnsureMatchesFound workflow task. But in this method, when filter to update only is true, we further filter the search result, it's possible after filtering the table is empty. and the table outputs 2 empty lines. So I made the change to the table and output no package found for this case

---
In reply to: [518338774](https://github.com/microsoft/winget-cli/pull/634#discussion_r518338774) [](ancestors = 518338774)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
yao-msft: Removed the flag after discussion

---
In reply to: [518454148](https://github.com/microsoft/winget-cli/pull/634#discussion_r518454148) [](ancestors = 518454148,518336942)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -419,22 +453,25 @@ namespace AppInstaller::CLI::Workflow
...
+
                 AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_MULTIPLE_APPLICATIONS_FOUND);
             }
 
-            auto latestVersion = searchResult.Matches.at(0).Package->GetLatestAvailableVersion();
-            Logging::Telemetry().LogAppFound(latestVersion->GetProperty(PackageVersionProperty::Name), latestVersion->GetProperty(PackageVersionProperty::Id));
+            auto package = searchResult.Matches.at(0).Package;
+            Logging::Telemetry().LogAppFound(package->GetProperty(PackageProperty::Name), package->GetProperty(PackageProperty::Id));
+
+            context.Add<Execution::Data::Package>(std::move(package));
```
JohnMcPMS, line 464: >std::move(package) [](start = 50, length = 18)

I don't know the rules around auto clearly enough to say whether this will be a reference or a copy.  I don't think we want to dismantle the SearchResult; that might just lead to bugs.  Can we change the `auto package` to be a clear copy of the Package shared_ptr so that this move is also clearly not taking it from the SearchResult?

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -43,8 +43,8 @@ namespace AppInstaller::Repository
             bool IsTombstone = false;
         };
 
-        // Finds a source from the given vector by its name.
-        auto FindSourceByName(std::vector<SourceDetailsInternal>& sources, std::string_view name)
+        // Finds a source from the given vector by its name. Returned source cannot be modified.
+        auto FindSourceByNameConst(const std::vector<SourceDetailsInternal>& sources, std::string_view name)
```
JohnMcPMS, line 47: >FindSourceByNameConst [](start = 13, length = 21)

Does not appear to be used (and if it was the comment would be to move it to SourceListInternal).

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -446,6 +396,158 @@ namespace AppInstaller::Repository
...
+        struct SourceListInternal
+        {
+            SourceListInternal();
+
+            // Get a list of current sources references which can be used to update the contents in place.
+            // e.g. update the LastTimeUpdated value of sources.
+            std::vector<std::reference_wrapper<SourceDetailsInternal>> GetCurrentSourceRefs();
+
+            // Current source means source that's not in tombstone
+            bool ContainsCurrentSource(std::string_view name);
```
JohnMcPMS, line 411: >ContainsCurrentSource [](start = 17, length = 21)

const function

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -446,6 +396,158 @@ namespace AppInstaller::Repository
...
+
+        SourceListInternal::SourceListInternal()
+        {
+            for (SourceOrigin origin : { SourceOrigin::User, SourceOrigin::Default })
+            {
+                auto forOrigin = GetSourcesByOrigin(origin);
+
+                for (auto&& source : forOrigin)
+                {
+                    if (!ContainsSource(source.Name))
```
JohnMcPMS, line 440: >ContainsSource [](start = 25, length = 14)

It would be more efficient here if `GetSource` could return an invalid value.  That way you don't need to search twice (applies to `LastUpdateTime` below even more so).

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -446,6 +396,158 @@ namespace AppInstaller::Repository
...
+        }
+
+        SourceDetailsInternal& SourceListInternal::GetSource(std::string_view name)
+        {
+            auto itr = FindSource(name, true);
+            THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), itr == m_sourceList.end());
+            return *itr;
+        }
+
+        void SourceListInternal::AddSource(const SourceDetailsInternal& details)
```
JohnMcPMS, line 517: >AddSource [](start = 33, length = 9)

This function will need to remove any existing tombstone before adding the details, or they would be blocked behind said tombstone.

src/AppInstallerRepositoryCore/RepositorySource.cpp
yao-msft: oops, forgot to clean up this one.

---
In reply to: [520185301](https://github.com/microsoft/winget-cli/pull/634#discussion_r520185301) [](ancestors = 520185301)

JohnMcPMS (approved):
:shipit:

PR #635, opened 2020-11-05
Swedish spelling fix

NBerlin:
Small changes to the Swedish spelling.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/635)

### DIFF
--- a/Localization/Resources/sv-SE/winget.resw
+++ b/Localization/Resources/sv-SE/winget.resw
@@ -628,7 +628,7 @@ De kan konfigureras via inställningsfilen "winget settings".</value>
     <value>Okänt kommando</value>
   </data>
   <data name="Usage" xml:space="preserve">
-    <value>Användning</value>
+    <value>användning</value>
     <comment>The way to use the software</comment>
   </data>
   <data name="ValidateCommandLongDescription" xml:space="preserve">
--- a/Localization/Resources/sv-SE/winget.resw
+++ b/Localization/Resources/sv-SE/winget.resw
@@ -138,7 +138,7 @@
     <comment>Nested commands that can be run in context of the selected command</comment>
   </data>
   <data name="ChannelArgumentDescription" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


Localization/Resources/sv-SE/winget.resw
```diff
@@ -628,7 +628,7 @@ De kan konfigureras via inställnings filen "winget settings".</value>
     <value>Okänt kommando</value>
   </data>
   <data name="Usage" xml:space="preserve">
-    <value>användning</value>
+    <value>Användning</value>
```
JohnMcPMS, line 631: Should this be capitalized?  You can see when it is used with:
```
> winget -?
Windows Package Manager v0.2.2941 Preview

The winget command line utility enables installing applications and other packages from the command line.

usage: winget [<command>] [<options>] <- !!Beginning of this line!!
...
```

I don't know if Swedish would expect this to be capitalized in this case (my grandmother never taught me any, as she was only 2 when she came to the US :smile:).

NBerlin: You're absolutely correct. I think I just presumed that it was a miss and didn't consider the context. I pushed up a fix for it now. 

Fun that you have Swedish heritage, have you ever visited?

JohnMcPMS:
Filed 30195245 internally to fix.

Localization/Resources/sv-SE/winget.resw
JohnMcPMS: No, though I've often thought it would be an interesting place to go (in the summer).  My mother and aunt both went and found the likely location where my great grandparents lived, although records are sparse and I believe the only thing they had to confirm was older locals claiming that someone of that family name living there long ago.

JohnMcPMS:
Fixed in #707

PR #645, opened 2020-11-14
Spell check

### LINKED ISSUE #559: Add SpellChecker to GH Action
Adding the Spellchecker will help with docs and code with misspellings. You can find it here: https://github.com/jsoref/spelling and is available in the GH Action Marketplace. Windows Terminal is already using it for example: https://github.com/microsoft/terminal/pull/4799

jsoref:
closes #559

This enables https://github.com/marketplace/actions/check-spelling

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/645)


Localization/Resources/af-ZA/winget.resw
```diff
@@ -534,7 +534,7 @@ Dit kan deur die instellinglêer ‘winget settings’ gekonfigureer word.</valu
   </data>
   <data name="SourceListField" xml:space="preserve">
     <value>Veld</value>
-    <comment>The name of a peice of information about a source.</comment>
+    <comment>The name of a piece of information about a source.</comment>
```
jsoref, line 537: It's possible I should drop the localization comment changes (I never know what the policy is)

src/AppInstallerCLITests/Completion.cpp
```diff
@@ -283,7 +283,7 @@ TEST_CASE("CommandComplete_PartialCommandMatch", "[complete]")
     CompletionTestContext ctc{ "cart", "winget cart", "11" };
 
     CompletionTestCommand command;
-    command.SubCommandNames = { "car", "cart", "cartesion", "carpet" };
+    command.SubCommandNames = { "car", "cart", "cartesian", "carpet" };
```
jsoref, line 286: This change is probably the biggest

src/AppInstallerCommonCore/DateTime.cpp
```diff
@@ -6,7 +6,7 @@
 namespace AppInstaller::Utility
 {
     // If moved to C++20, this can be replaced with standard library implementations.
-    void OutputTimepoint(std::ostream& stream, const std::chrono::system_clock::time_point& time)
+    void OutputTimePoint(std::ostream& stream, const std::chrono::system_clock::time_point& time)
```
jsoref, line 9: Given that the underlying type is `time_point`, it made sense to me to use this capitalization

src/AppInstallerCommonCore/Errors.cpp
```diff
@@ -104,7 +104,7 @@ namespace AppInstaller
             case APPINSTALLER_CLI_ERROR_UPDATE_ALL_HAS_FAILURE:
                 return "winget upgrade --all completed with failures";
             default:
-                return "Uknown Error Code";
+                return "Unknown Error Code";
```
jsoref, line 107: This could be a cli api change

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -486,7 +486,7 @@ namespace AppInstaller::Repository
                         // We did not find an exact match on Id in the results
                         if (!availablePackage)
                         {
-                            AICLI_LOG(Repo, Warning, << "  Appropriate available package could not be determined, setting availablility state to unknown");
+                            AICLI_LOG(Repo, Warning, << "  Appropriate available package could not be determined, setting availability state to unknown");
```
jsoref, line 489: This may be a cli api change

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.h
```diff
@@ -59,7 +59,7 @@ namespace AppInstaller::Repository::Microsoft
         // The priority is:
         //  DisplayVersion
         //  Version
-        //  MajorVerison, MinorVersion
+        //  MajorVersion, MinorVersion
```
jsoref, line 62: Google Sheets suggested `Verizon`

jsoref (commented):
It's possible to split the spelling fixes from the spell-check action, but it's important to land the spell-checker w/o letting there be too much drift between its PR base and master -- it's bad to land it if there have been typos introduced since its base (as all new PRs from its landing will have confusing failures).

For reference, this PR gets a ✅ : https://github.com/jsoref/winget-cli/commit/4ad354357b3cf5ba5a4100b0ba2bb89f7cebc695

Localization/Resources/af-ZA/winget.resw
JohnMcPMS: Certainly the localization comments can be updated; they will just be overwritten by future changes if needed.

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.h
JohnMcPMS: Major Verizon

A lot of these changes are my extremely common typos that I failed to correct.  I'm sure that some sort of ML (or just analysis of qwerty) could probably figure out that one of my fingers jumps the gun on the others occasionally.

.github/actions/spelling/excludes.txt
```diff
@@ -0,0 +1,17 @@
...
+\.vcxproj$
+\.vcxproj\.filters$
+\.vdproj$
+\.wapproj$
+^cgmanifest\.json$
+^Localization/
+^NOTICE$
+^src/AppInstallerCLICore/Commands/ExperimentalCommand\.cpp$
+^src/catch2/
+^src/YamlCppLib/
```
JohnMcPMS, line 16: Needs to exclude jsoncpplib as well:
```
^src/JsonCppLib/
```

.github/workflows/spelling.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  build:
+    name: Spell checking
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2.0.0
+      with:
+        fetch-depth: 5
+    - uses: check-spelling/check-spelling@0.0.16-alpha
```
JohnMcPMS, line 20: >check-spelling/check-spelling@0.0.16-alpha [](start = 12, length = 42)

I'm not familiar with GitHub Actions.

1. Will we need to update this version manually to keep up with new releases?
2. Do we need to do something to manage the secrets.GITHUB_TOKEN?  Or is that automatically done by the Actions infrastructure?
3. Generally, are there any other actions we need to take to manage this (beyond excludes, patterns, etc. for new changes)?

Thanks for your work here.

jsoref: 1. In theory @dependabot can tell you about updates, but for the near term, I'll probably make the PR myself.
1. The `GITHUB_TOKEN` is automatic https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret
1. Beyond that and trying to avoid merging things with spelling errors (if the main branch is unhappy, anyone who branches from that point will be unhappy / confused), no.
   * There is the option to add an [advice.txt](https://github.com/microsoft/terminal/blob/main/.github/actions/spell-check/advice.txt) file which is included in the message the bot leaves, but that's totally tunable and that's really something you'll work through as a project.
   * Technically GitHub has a [budget](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions) for actions, it's theoretically possible for someone to run into this limit, I suspect that Microsoft has a reasonably large budget and you won't notice (certainly microsoft/terminal hasn't complained) -- if someone publicly runs into this, I'd love to hear about it. (Generally other actions are more expensive than this one -- which runs at about 1 minute and is thus billed as 1-2 units.) Most likely if a project like yours hits something like this, it means you have a bunch of files that should be ignored that you aren't ignoring (one project I'm looking at takes ~20 minutes to scan). -- If this becomes a problem for some other project in the future, I'll probably write some code so that it tries to scan [only changed files](https://github.com/check-spelling/check-spelling/wiki/Feature:-Only-changes) instead of the whole repository.

JohnMcPMS: Thanks for the info, I agree that only checking changed files seems like a good optimization opportunity.

JohnMcPMS (approved):

PR #652, opened 2020-11-24
Adding export to WinGetUtil.dll to expose Version comparing

palenshus:
Adding new export, WinGetCompareVersions, which compares two versions for ordering purposes. This allows other tools and services to be guaranteed the same version sorting when working with the OWC that WinGet uses.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/652)

### DIFF
--- a/src/WinGetUtil/Exports.cpp
+++ b/src/WinGetUtil/Exports.cpp
@@ -239,7 +239,7 @@ extern "C"
         Version vA{ ConvertToUTF8(versionA) };
         Version vB{ ConvertToUTF8(versionB) };
 
-        *comparisonResult = vA < vB ? -1 : vA == vB ? 0 : 1;
+        *comparisonResult = vA < vB ? -1 : (vA == vB ? 0 : 1);
 
         return S_OK;
     }
--- a/src/WinGetUtil/Exports.cpp
+++ b/src/WinGetUtil/Exports.cpp
@@ -227,4 +227,21 @@ extern "C"
         return S_OK;
     }
     CATCH_RETURN()
+
+    WINGET_UTIL_API WinGetCompareVersions(
+        WINGET_STRING versionA,
+        WINGET_STRING versionB,
+        INT* comparisonResult) try
+    {
+        THROW_HR_IF(E_INVALIDARG, !versionA);
+        THROW_HR_IF(E_INVALIDARG, !versionB);
+
+        Version vA{ ConvertToUTF8(versionA) };
+        Version vB{ ConvertToUTF8(versionB) };
+
+        *comparisonResult = vA < vB ? -1 : vA == vB ? 0 : 1;
+
+        return S_OK;
+    }
+    CATCH_RETURN()
 }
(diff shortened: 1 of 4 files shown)


palenshus:
Did some perf testing and comparisons against just passing in the entire array of versions to be sorted in from .NET and marshaling the sorted array back out. Surprisingly, it's actually faster to do it this way. Took about 12s to sort 10k versions using this method, and 15s to pass in and sort the entire array. Since this code is also cleaner and simpler, seems like a win-win.

src/WinGetUtil/WinGetUtil.vcxproj
```diff
@@ -144,7 +144,7 @@
       <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">Source.def</ModuleDefinitionFile>
       <AdditionalDependencies Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <AdditionalDependencies Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;%(AdditionalDependencies)</AdditionalDependencies>
-      <AdditionalDependencies Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">version.lib;wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
yao-msft, line 147: >version.lib [](start = 86, length = 11)

This should be applicable to all configurations

yao-msft: eh, wait, I think we don't need this?

---
In reply to: [529897938](https://github.com/microsoft/winget-cli/pull/652#discussion_r529897938) [](ancestors = 529897938)

src/WinGetUtil/Exports.cpp
```diff
@@ -227,4 +227,21 @@ extern "C"
...
+        WINGET_STRING versionB,
+        INT* comparisonResult) try
+    {
+        THROW_HR_IF(E_INVALIDARG, !versionA);
+        THROW_HR_IF(E_INVALIDARG, !versionB);
+
+        Version vA{ ConvertToUTF8(versionA) };
+        Version vB{ ConvertToUTF8(versionB) };
+
+        *comparisonResult = vA < vB ? -1 : vA == vB ? 0 : 1;
```
yao-msft, line 242: >vA == vB ? 0 : 1 [](start = 43, length = 16)

nit: maybe surround with () to make it better readable?
vA < vB ? -1 : (vA == vB ? 0 : 1)

palenshus: Sure no prob

src/WinGetUtil/WinGetUtil.vcxproj
palenshus: Looks like you're correct

yao-msft (approved):
:shipit:

Issue #121, opened 2020-05-11
Uninstall an installed App


denelon:
As a user I want to be able to use winget to uninstall a package so I don't have to go through the Add/Remove software UX.

**Experimental Feature**
`winget uninstall <package>` - uninstalls the specified package.
`winget uninstall -?` - displays help for the uninstall command.

Note: uninstall depends on list also being enabled.

`winget features` - displays all experimental features and their status
`winget settings` - launches settings.json to configure settings like experimental features

Note:
There are limitations about how programs get installed and whether or not that binary supports uninstall. This may also need to be supported in the manifest with new key(s) and or syntax.

Edited: experimental status

phiresky:
This is never going to happen since packages just consist of random executable files that can do whatever they want. There's no package building step or anything.

jedieaston:
Chocolatey (and scoop in certain scenarios) already does it through running the uninstaller that is registered in the control panel, I thought. winget could do a similar thing, although the flags for a silent uninstall would have to be in the manifest somewhere.

michaelvolz:
IMHO this is basic functionality which is paramount for the success of winget.

Smart123s:
> This is never going to happen since packages just consist of running random executable files that can do whatever they want. There's no package building step or anything.

There could be references to the programs' "uninstaller" executables as well.
For example, it could launch `C:\Program Files (x86)\Steam\uninstall.exe` to uninstall Steam.

Karl-WE:
The uninstaller reference comes with every installation. There are RARE cases of applications that cannot be uninstalled this way but it is doubtful they will be ever accessible / installable via winget.

Other than that you can clearly uninstall all apps using a WMI call and I have even borrowed a method someone else that is even able to crawl the registry for uninstalling applications by a name or wildcards. 

Without blaming anyone, I certainly support an /u --uninstall switch, which can also be combined with /h --silent

Karl-WE:
in terms of naming i would not use Winget remove, imho as it is "install" it should be named "uninstall" for ease of use. [https://github.com/microsoft/winget-cli/issues/192](url)

masx200:
i support it

phiresky:
For clarification about what I meant with my (admittedly flippant) comment above, see these three threads:

* https://github.com/microsoft/winget-cli/issues/223
* https://news.ycombinator.com/item?id=23238051
[...]

PR #659, opened 2020-12-02
Implement uninstall flow

lechacon:
This is in progress and does not work completely yet. I'm opening it as a draft to see if there are any major issues.

Changes:
* Added an UninstallCommand as an experimental feature.
* Added workflow tests for uninstall (not passing yet).
* Added uninstall WorkflowTasks for uninstalling MSIX/MSStore apps with the package name, and everything else with the uninstall string. This uses already existing information form manifests/ARP.

Still missing:
* Logging, reporting, telemetry
* E2E tests
* More manual validation to work out details/edge cases

Related: #121 #397 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/659)


src/AppInstallerCLITests/TestData/InstallFlowTest_MSStore.yaml
```diff
@@ -9,4 +9,5 @@ Installers:
       Url: https://ThisIsNotUsed
       InstallerType: MSStore
       ProductId: 9WZDNCRFJ364
+      PackageFamilyName: Microsoft.SkypeApp_kzf8qxf38zg5c
```
jsoref, line 12: You'll probably want to add a `pattern` entry for this.
```
PackageFamilyName: \S+
```

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
```diff
@@ -0,0 +1,135 @@
...
+    {
+        return { Resource::String::UninstallCommandShortDescription };
+    }
+
+    Resource::LocString UninstallCommand::LongDescription() const
+    {
+        return { Resource::String::UninstallCommandLongDescription };
+    }
+
+    void UninstallCommand::Complete(Execution::Context& context, Execution::Args::Type valueType) const
```
JohnMcPMS, line 46: >Complete [](start = 27, length = 8)

I expect that this is a copy of the upgrade command Complete function.  Is that correct?

lechacon: Yes, I copied it from there but then I removed some arguments that are not needed, like InstallLocation.

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -44,6 +46,8 @@ namespace AppInstaller::Settings
             return ExperimentalFeature{ "List Command", "list", "https://aka.ms/winget-settings", Feature::ExperimentalList };
         case Feature::ExperimentalUpgrade:
             return ExperimentalFeature{ "Upgrade Command", "upgrade", "https://aka.ms/winget-settings", Feature::ExperimentalUpgrade };
+        case Feature::ExperimentalUninstall:
```
JohnMcPMS, line 49: >ExperimentalUninstall [](start = 22, length = 21)

Will need to update the settings page referenced here and the schema file.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+                GetExitCodeProcess(process.get(), &exitCode);
+                return exitCode;
+            }
+        }
+    }
+
+    void GetUninstallInfo(Execution::Context& context)
+    {
+        auto installedPackageVersion = context.Get<Execution::Data::InstalledPackageVersion>();
+        const std::string insalledTypeString = installedPackageVersion->GetMetadata()[PackageVersionMetadata::InstalledType];
```
JohnMcPMS: >insalledTypeString [](start = 26, length = 18)

typo: installed

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+        case ManifestInstaller::InstallerTypeEnum::Burn:
+        case ManifestInstaller::InstallerTypeEnum::Inno:
+        case ManifestInstaller::InstallerTypeEnum::Msi:
+        case ManifestInstaller::InstallerTypeEnum::Nullsoft:
+        case ManifestInstaller::InstallerTypeEnum::Wix:
+        {
+            IPackageVersion::Metadata packageMetadata = installedPackageVersion->GetMetadata();
+            IPackageVersion::Metadata::iterator itr;
+
+            if (context.Args.Contains(Execution::Args::Type::Interactive))
```
JohnMcPMS, line 86: For this part, I think it is actually better to work with whatever we have available.  That is to say, the context.Args indicates which to prefer, but if only one is present we should use it regardless of the requested interactivity level.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+                if (itr == packageMetadata.end())
+                {
+                    itr = packageMetadata.find(PackageVersionMetadata::SilentUninstallCommand);
+                }
+            }
+
+            if (itr == packageMetadata.end())
+            {
+                AICLI_TERMINATE_CONTEXT(E_ABORT);
+            }
```
JohnMcPMS, line 106: There needs to be some `context.Reporter` output here to indicate that no uninstall string was available.  Also, you cannot use E_ABORT, as that is the HR used for user cancellation.  It is probably best to create a new HR in our list.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+            std::bind(InvokeCreateProcess,
+                context.Get<Execution::Data::UninstallString>(),
+                std::placeholders::_1));
+
+        if (!uninstallResult)
+        {
+            // TODO: log & report
+            AICLI_TERMINATE_CONTEXT(E_ABORT);
+        }
+        else if (uninstallResult.value() != 0)
```
JohnMcPMS, line 162: >(uninstallResult.value() != 0) [](start = 16, length = 30)

This may need to be smarter; at a minimum I would expect a lot of "reboot required" errors coming out.  I think that for at least MSI we should probably figure out what can come back from msiexec.exe and be able to handle those values.

src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -940,6 +986,97 @@ TEST_CASE("UpdateFlow_UpdateAllApplicable", "[UpdateFlow][workflow]")
...
+    uninstall.Execute(context);
+    INFO(uninstallOutput.str());
+
+    // Verify Uninstaller is called with the package full name.
+    REQUIRE(std::filesystem::exists(uninstallResultPath.GetPath()));
+    std::ifstream uninstallResultFile(uninstallResultPath.GetPath());
+    REQUIRE(uninstallResultFile.is_open());
+    std::string uninstallResultStr;
+    std::getline(uninstallResultFile, uninstallResultStr);
+    REQUIRE(uninstallResultStr.find("20477fca-282d-49fb-b03e-371dca074f0f_8wekyb3d8bbwe") != std::string::npos);
```
jsoref, line 1034: This is why the spell checker doesn't like that thing.
You can write a pattern for something, or even just use:
`"20477fca-282d-49fb-b03e-371dca074f0f_8wekyb3d8bbwe"`

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+            {
+                auto packageFullName = Msix::GetPackageFullNameFromFamilyName(packageFamilyName);
+                if (packageFullName.has_value())
+                {
+                    AICLI_LOG(CLI, Info, << "Removing MSIX package: " << packageFullName.value());
+                    context.Reporter.ExecuteWithProgress(std::bind(Deployment::RemovePackage, packageFullName.value(), std::placeholders::_1));
+                }
+            }
+        }
+        catch (const wil::ResultException& re)
```
JohnMcPMS, line 191: >catch (const wil::ResultException& re) [](start = 8, length = 38)

There is no need to catch here, the root level will handle it.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+            }
+            context.Add<Execution::Data::UninstallString>(itr->second);
+
+            break;
+        }
+        case ManifestInstaller::InstallerTypeEnum::Msix:
+        case ManifestInstaller::InstallerTypeEnum::MSStore:
+        {
+            context.Add<Execution::Data::PackageFamilyNames>(installedPackageVersion->GetMultiProperty(PackageVersionMultiProperty::PackageFamilyName));
+            break;
```
JohnMcPMS, line 115: This should check for the case where no values are present and fail similarly to no uninstall string.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+    namespace
+    {
+        std::optional<DWORD> InvokeCreateProcess(const std::string& command, IProgressCallback& progress)
+        {
+            std::wstring commandUtf16 = Utility::ConvertToUTF16(command);
+
+            // Parse the command string as application and command line for CreateProcess
+            PWSTR app, commandLine;
+            THROW_IF_FAILED(SHEvaluateSystemCommandTemplate(commandUtf16.c_str(), &app, &commandLine, NULL));
```
JohnMcPMS, line 25: >SHEvaluateSystemCommandTemplate [](start = 28, length = 31)

I thought you explicitly wanted to use CreateProcess because it didn't require this.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+namespace AppInstaller::CLI::Workflow
+{
+    namespace
+    {
+        std::optional<DWORD> InvokeCreateProcess(const std::string& command, IProgressCallback& progress)
+        {
+            std::wstring commandUtf16 = Utility::ConvertToUTF16(command);
+
+            // Parse the command string as application and command line for CreateProcess
+            PWSTR app, commandLine;
```
JohnMcPMS, line 24: >PWSTR app, commandLine; [](start = 12, length = 23)

Initialize to nullptr.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+            THROW_IF_FAILED(SHEvaluateSystemCommandTemplate(commandUtf16.c_str(), &app, &commandLine, NULL));
+
+            STARTUPINFOW startupInfo = { sizeof(startupInfo) };
+            PROCESS_INFORMATION processInfo;
```
JohnMcPMS, line 28: >processInfo [](start = 32, length = 11)

init: = {}

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,201 @@
...
+            BOOL success = CreateProcessW(app, commandLine, NULL, NULL, FALSE, 0, NULL, NULL, &startupInfo, &processInfo);
+
+            CoTaskMemFree(app);
```
JohnMcPMS, line 31: >CoTaskMemFree [](start = 12, length = 13)

If these require free'ing, you should use the wil RAII classes.

src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -940,6 +986,97 @@ TEST_CASE("UpdateFlow_UpdateAllApplicable", "[UpdateFlow][workflow]")
...
+    uninstall.Execute(context);
+    INFO(uninstallOutput.str());
+
+    // Verify Uninstaller is called with the package full name
+    REQUIRE(std::filesystem::exists(uninstallResultPath.GetPath()));
+    std::ifstream uninstallResultFile(uninstallResultPath.GetPath());
+    REQUIRE(uninstallResultFile.is_open());
+    std::string uninstallResultStr;
+    std::getline(uninstallResultFile, uninstallResultStr);
+    REQUIRE(uninstallResultStr.find("microsoft.skypeapp_kzf8qxf38zg5c") != std::string::npos);
```
jsoref, line 1057: This is where the other pair is coming from.

I think using a `pattern` makes sense:
```
uninstallResultStr\.find\("[^"]+"\)
```

jsoref (commented):
I think that `INET` should be resolved by 4f7ce38037c0392ef856c3c407f216c9a357f55f, so you'll probably want to rebase (or merge?) to avoid doubling this.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
lechacon: Thanks. I didn't know about that.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
lechacon: That was originally because I thought it would be harder to parse the string before I knew of this function. I ended using CreateProcess anyway because that's what the "Programs & Features" UI does.

src/AppInstallerCLITests/WorkFlow.cpp
lechacon: Thanks. That seems better than listing the specific strings.

yao-msft:
> 

closing ``` to end the code block?

---
Refers to: doc/Settings.md:99 in 1181019. [](commit_id = 1181019282a551bc48a36b57fecb37d71a6eb55f, deletion_comment = False)

.github/actions/spelling/patterns.txt
```diff
@@ -15,3 +15,6 @@ data:[a-zA-Z=;,/0-9+-]+
 \b[0-9a-f]{40}\b
 \b([A-Za-z])\1{3,}\b
 El proyecto .* diferentes
+# Package family names and product codes in tests
+PackageFamilyName: \S+
+uninstallResultStr\.find\("[^"]+"\)
```
yao-msft, line 20: I personally think a regex pattern for any pfn would be better and future proof, instead of every place we'll be referencing a pfn

jsoref: Sure. I just don't have a good idea how to define them.

jsoref: Sure. I just don't know offhand how to match them. Hopefully it shouldn't be too hard.

jsoref:
@lechacon: you can visit https://github.com/lechacon/winget-cli/actions to enable the spell checker -- if you do that, it'll try to highlight the specific line it doesn't like when it finds a new misspelling.

.github/actions/spelling/patterns.txt
```diff
@@ -15,3 +15,5 @@ data:[a-zA-Z=;,/0-9+-]+
 \b[0-9a-f]{40}\b
 \b([A-Za-z])\1{3,}\b
 El proyecto .* diferentes
+# Package family names
+[-.A-Za-z0-9]+_[a-z0-9]{13}
```
jsoref, line 19: This matches:
https://github.com/lechacon/winget-cli/blob/bb9016b37277bd43d4c7de6fc1d07cd5d5f85a5f/src/AppInstallerCLITests/Settings.cpp#L87
So instead of seeing:
```
    StreamDefinition name{ Type::Secure, "secure_nonexistentsetting" };
```
The spell checker saw:
```
    StreamDefinition name{ Type::Secure, "tting" };
```

jsoref: This fixes the specific complaint:
```suggestion
\b[-.A-Za-z0-9]+_[a-z0-9]{13}\b
```
(The `\b` at the beginning isn't important, the one at the end is.)

I can't promise the spell checker won't complain about something else. (I'm out of time for today.)

lechacon: Thank you! I couldn't have figured that out myself.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,182 @@
...
+        case ManifestInstaller::InstallerTypeEnum::Nullsoft:
+        case ManifestInstaller::InstallerTypeEnum::Wix:
+            context << Workflow::ExecuteUninstallString;
+            break;
+        case ManifestInstaller::InstallerTypeEnum::Msix:
+            context << Workflow::MsixUninstall;
+            break;
+        case ManifestInstaller::InstallerTypeEnum::MSStore:
+            context <<
+                Workflow::EnsureFeatureEnabled(Settings::ExperimentalFeature::Feature::ExperimentalMSStore) <<
```
yao-msft, line 131: >Workflow::EnsureFeatureEnabled(Settings::ExperimentalFeature::Feature::ExperimentalMSStore) << [](start = 16, length = 94)

I'm not sure if we need this for uninstallation.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+    void MsixUninstall(Execution::Context& context)
+    {
+        const auto& packageFamilyNames = context.Get<Execution::Data::PackageFamilyNames>(); 
+        context.Reporter.Info() << Resource::String::UninstallFlowStartingPackageUninstall << std::endl;
+
+        for (const auto& packageFamilyName : packageFamilyNames)
+        {
+            auto packageFullName = Msix::GetPackageFullNameFromFamilyName(packageFamilyName);
+            if (packageFullName.has_value())
```
yao-msft, line 173: >if (packageFullName.has_value()) [](start = 12, length = 32)

I understand it's less likely since this is from installed source. but do we want to log and show something to the user for "package not found"?

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,182 @@
...
+        case ManifestInstaller::InstallerTypeEnum::Msi:
+        case ManifestInstaller::InstallerTypeEnum::Nullsoft:
+        case ManifestInstaller::InstallerTypeEnum::Wix:
+        {
+            IPackageVersion::Metadata packageMetadata = installedPackageVersion->GetMetadata();
+            IPackageVersion::Metadata::iterator itr = packageMetadata.find(PackageVersionMetadata::StandardUninstallCommand);
+
+            if (itr == packageMetadata.end() || context.Args.Contains(Execution::Args::Type::Silent))
+            {
+                itr = packageMetadata.find(PackageVersionMetadata::SilentUninstallCommand);
```
yao-msft, line 83: >itr = packageMetadata.find(PackageVersionMetadata::SilentUninstallCommand); [](start = 16, length = 75)

I think if --silent is passed in and there exists StndardUninstallString, this will lead to ERROR_NO_UNINSTALL_INFO_FOUND? Which is not what John wanted above?

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,182 @@
...
+                GetExitCodeProcess(processInfo.hProcess, &exitCode);
+                return exitCode;
+            }
+        }
+    }
+
+    void GetUninstallInfo(Execution::Context& context)
+    {
+        auto installedPackageVersion = context.Get<Execution::Data::InstalledPackageVersion>();
+        const std::string installedTypeString = installedPackageVersion->GetMetadata()[PackageVersionMetadata::InstalledType];
```
yao-msft, line 68: >[PackageVersionMetadata::InstalledType] [](start = 86, length = 39)

I think John mentioned that this key may not always be present, or this is not the case any more?

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,182 @@
...
+            THROW_IF_FAILED(SHEvaluateSystemCommandTemplate(commandUtf16.c_str(), &app, &commandLine, NULL));
+
+            STARTUPINFOW startupInfo = { sizeof(startupInfo) };
+            wil::unique_process_information processInfo = {};
+            BOOL success = CreateProcessW(app.get(), commandLine.get(), NULL, NULL, FALSE, 0, NULL, NULL, &startupInfo, &processInfo);
+
+            if (!success)
+            {
+                return GetLastError();
+            }
```
yao-msft, line 35: THROW_LAST_ERROR_IF() since it's not the actual process exit code?

yao-msft:
By the way, please hook up with the UpdateFlow for UninstallPrevious behavior if possible when this pr becomes official

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
yao-msft: Chatted offline, we found "Programs & Features" uses CreateProcess and Settings(which is modern) uses ShellExecute.

I think for code consistency and extra ShellExecute benefits like ensuring policy, etc, It might be better to use ShellExecute here if not much refactoring effort needed?

---
In reply to: [535774167](https://github.com/microsoft/winget-cli/pull/659#discussion_r535774167) [](ancestors = 535774167)

lechacon: I think it makes sense to change it to ShellExecute. I'll do that.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
JohnMcPMS: We should at least AICLI_LOG that case, but I don't think it would be bad to output a warning as well.

---
In reply to: [536447224](https://github.com/Microsoft/winget-cli/pull/659#discussion_r536447224) [](ancestors = 536447224)

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
JohnMcPMS: Agreed, this should get an another iterator and only assign to `itr` if it is valid.

---
In reply to: [536449750](https://github.com/Microsoft/winget-cli/pull/659#discussion_r536449750) [](ancestors = 536449750)

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
JohnMcPMS: No, that one should always be present.  It may not be super accurate, but it should at least get the ARP/MSIX split that is needed here.

---
In reply to: [536451002](https://github.com/Microsoft/winget-cli/pull/659#discussion_r536451002) [](ancestors = 536451002)

mimi89999:
1. Many package managers offer 2 commands for removing packages: `uninstall` that only uninstalls the package and `purge` that also removes package data. How should silent `uninstall` work in winget? I think that it should only remove the package, but that should be documented so that the correct switches are added in the repo and admins uninstalling a package don't get a surprise.
2. Package installs are silent by default. Why for uninstalls a CLI switch is required? This doesn't seem very consistent.
3. What about `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{ProductCode}\UninstallString`, `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{ProductCode}\QuietUninstallString`, etc. Why aren't they used?

JohnMcPMS:
> 1. Many package managers offer 2 commands for removing packages: `uninstall` that only uninstalls the package and `purge` that also removes package data. How should silent `uninstall` work in winget? I think that it should only remove the package, but that should be documented so that the correct switches are added in the repo and admins uninstalling a package don't get a surprise.
> 2. Package installs are silent by default. Why for uninstalls a CLI switch is required? This doesn't seem very consistent.
> 3. What about `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{ProductCode}\UninstallString`, `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{ProductCode}\QuietUninstallString`, etc. Why aren't they used?

1. As our goal is to manage the 20+ year history of existing Windows "packages", across as many different technologies as possible, we don't have the level of control required to offer the uninstall/purge as you have described.  We could do something similar for MSIX, as underneath it is a more traditional package manager.  The same could be true for future types like ZIP.  But for EXE/MSI installers, the uninstall is under the control of the ISV.

2. I can agree that it makes sense to default to using the silent uninstall, however it is actually not common to offer a value there.  We should certainly document that we will do our best, but uninstall may not be very scriptable.

3. These are the values being used, they are just being collected in other parts of the code in a previous PR.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
lechacon: I changed that, but in InvokeShellExecute since this part wasn't needed anymore.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
lechacon: I'm leaving this as a TODO for now as I'm not entirely sure what's the right thing to do with other exit codes. I'm trying to find an application that returns something non-zero to see what it does and how it is handled when uninstalling with Porgrams and Features or the Settings app. I'll probably add a telemetry point for errors that will help here.

src/AppInstallerTestExeInstaller/main.cpp
```diff
@@ -14,7 +14,7 @@ using namespace std::filesystem;
 std::string_view registrySubkey = "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\";
 std::string_view defaultProductID = "{A499DD5E-8DC5-4AD2-911A-BCD0263295E9}";
 
-std::wstring GenerateUninstaller(std::ostream& out, const path& installDirectory) {
+std::wstring GenerateUninstaller(std::ostream& out, const path& installDirectory, const std::wstring& productID) {
```
yao-msft, line 17: >{ [](start = 113, length = 1)

nit: move { below for consistent code style

src/AppInstallerTestExeInstaller/main.cpp
```diff
@@ -23,11 +23,21 @@ std::wstring GenerateUninstaller(std::ostream& out, const path& installDirectory
     path uninstallerOutputTextFilePath = installDirectory;
     uninstallerOutputTextFilePath /= "TestExeUninstalled.txt";
 
-    // TODO: Needs to re-invoke the installer and remove the Uninstall key that it added
+    path registryKey = "HKEY_CURRENT_USER" / path{ registrySubkey };
```
yao-msft, line 26: >path registryKey [](start = 4, length = 16)

This is not a typical std::filesystem::path, I'd suggest constructing regkey as string

yao-msft:
>             productCodeStream << argv[i];

I know it's not your fault but probably a previous unaddressed comment. This should just be std::string productCode; (not std::wstring) declaration in line 131 and direct assignment here.

In line 131,
std::string produceCode;

And here:
productCode = argv[i];

---
Refers to: src/AppInstallerTestExeInstaller/main.cpp:151 in bc73961. [](commit_id = bc739617f1a375564fa6f43a731b43004b48a8fc, deletion_comment = False)

src/AppInstallerTestExeInstaller/main.cpp
yao-msft, line 17: >std::wstring [](start = 0, length = 12)

Not your fault either, but it's suggested to return path here and WriteToUninstallRegistry() below(line 46) should take in std::filesystem::Path as input

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -734,4 +734,25 @@ They can be configured through the settings file 'winget settings'.</value>
...
+    <value>Uninstalls the given package</value>
+  </data>
+  <data name="UninstallFlowStartingPackageUninstall" xml:space="preserve">
+    <value>Starting package uninstall...</value>
+  </data>
+  <data name="UninstallFlowUninstallSuccess" xml:space="preserve">
+    <value>Successfully uninstalled</value>
+  </data>
+  <data name="NoUninstallInfoFound" xml:space="preserve">
+    <value>The Windows Package Manager cannot locate the uninstall command for this application. Please reach out to the application publisher for support.</value>
```
yao-msft, line 750: >The Windows Package Manager [](start = 11, length = 27)

nit: winget to be used in output message.
Also application should be replaced with package since we did a project wide renaming effort to use "package" to be consistent

src/AppInstallerCLIE2ETests/BaseCommand.cs
```diff
@@ -12,7 +12,9 @@ namespace AppInstallerCLIE2ETests
 
     public class BaseCommand
     {
-        public readonly string SettingsJsonFilePath = @"Packages\WinGetDevCLI_8wekyb3d8bbwe\LocalState\settings.json";
+        public string SettingsJsonFilePath => TestCommon.PackagedContext ?
```
yao-msft, line 15: >TestCommon.PackagedContext [](start = 46, length = 26)

nice, thanks!

src/AppInstallerCLIE2ETests/UninstallCommand.cs
```diff
@@ -8,23 +8,101 @@ namespace AppInstallerCLIE2ETests
...
+            // Uninstall an Exe
+            var installDir = TestCommon.GetRandomTestDir();
+            TestCommon.RunAICLICommand("install", $"{Constants.ExeInstallerPackageId} --silent -l {installDir}");
+            var result = TestCommon.RunAICLICommand("uninstall", Constants.ExeInstallerPackageId);
+            Assert.AreEqual(Constants.ErrorCode.S_OK, result.ExitCode);
+            Assert.True(result.StdOut.Contains("Successfully uninstalled"));
+            Assert.True(VerifyTestExeUninstalled(installDir));
+        }
+
+        [Test]
```
yao-msft, line 41: >[Test] [](start = 8, length = 6)

Should we comment out the tag instead of checking for ignoring since msi project is never supported in azp?

src/AppInstallerCLIE2ETests/UninstallCommand.cs
```diff
@@ -8,23 +8,101 @@ namespace AppInstallerCLIE2ETests
 
     public class UninstallCommand : BaseCommand
     {
-        private const string UninstallTestExeInstalledFile = @"TestExeUninstalled.txt";
+        private const string CustomProductCode = "{f08fc03c-0b7e-4fca-9b3c-3a384d18a9f3}";
+        private const string UninstallTestExeUninstalledFile = "TestExeUninstalled.txt";
+        private const string UninstallTestMsiInstalledFile = "AppInstallerTestExeInstaller.exe";
```
yao-msft, line 13: >UninstallTestMsiInstalledFile = "AppInstallerTestExeInstaller.exe"; [](start = 29, length = 67)

this line does not look correct

yao-msft: ah, I see now. Please add a comment on why it's this because on first sight, it's confusing

---
In reply to: [543043894](https://github.com/microsoft/winget-cli/pull/659#discussion_r543043894) [](ancestors = 543043894)

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -0,0 +1,126 @@
...
+            context.Add<Execution::Data::UninstallString>(uninstallCommandItr->second);
+            break;
+        }
+        case ManifestInstaller::InstallerTypeEnum::Msi:
+        case ManifestInstaller::InstallerTypeEnum::Wix:
+        {
+            // Uninstall strings for MSI don't include UI level (/q) needed to avoid interactivity,
+            // so we handle them differently.
+            auto productCodes = installedPackageVersion->GetMultiProperty(PackageVersionMultiProperty::ProductCode);
+            if (productCodes.empty())
```
yao-msft, line 55: >if (productCodes.empty()) [](start = 12, length = 25)

I believe in current state we'll have to fall back to uninstallstring here.
The installedPackageVersion->GetMultiProperty() is a method for querying index. This will only work iff the installed package is correlated back to an entry in index and the manifest in the index contains productcode declarations.

Or if the only reason we do this is /q not in the uninstall string, can we add /q by default unless it's /interactive mode?

yao-msft: Also do you know if it's common that ISV sets custom args? if we construct own own from product code, I'm afraid custom args will be lost, if there exists one

---
In reply to: [543052441](https://github.com/microsoft/winget-cli/pull/659#discussion_r543052441) [](ancestors = 543052441)

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -247,4 +267,82 @@ namespace AppInstaller::CLI::Workflow
...
+            context.Reporter.Warn() << Resource::String::UninstallAbandoned << std::endl;
+            AICLI_TERMINATE_CONTEXT(E_ABORT);
+        }
+        else if (uninstallResult.value() != 0)
+        {
+            const auto installedPackageVersion = context.Get<Execution::Data::InstalledPackageVersion>();
+            Logging::Telemetry().LogUninstallerFailure(
+                installedPackageVersion->GetProperty(PackageVersionProperty::Id),
+                installedPackageVersion->GetProperty(PackageVersionProperty::Version),
+                "UninstallString",
```
yao-msft, line 298: >"UninstallString" [](start = 16, length = 17)

I think this is for InstallerType in telemetry backend? Or is it intentional you create a new type for this telemetry event? Same below for line 338

src/AppInstallerCLIE2ETests/UninstallCommand.cs
lechacon: I doubt anybody would ever run this test if the tag is commented out. If we keep it but ignore it, even if it never runs in azp it would at least run if somebody runs it locally and sets up the MSI project. The downside is that the pass rate for the tests in azp goes down from 100% with this marked as "Other".

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
lechacon: > The installedPackageVersion->GetMultiProperty() is a method for querying index. This will only work iff the installed package is correlated back to an entry in index

I believe this should always come from the Installed source which always sets the product code on `ARPHelper::PopulateIndexFromKey`. If we have the uninstall string then we necessarily also have the product code from the ARP key name.

> Or if the only reason we do this is /q not in the uninstall string, can we add /q by default unless it's /interactive mode?

I considered it, but to add /q we need to either be completely sure it is not possible to already have a /q flag here, be sure that the flag we add at the end will always override any flag that may already be there, or parse the string to see if it has one to remove it. That all seems too complicated and I'm pretty sure that constructing the command ourselves works fine because that's what the Settings app, and Programs and Features do for uninstalling.

> Also do you know if it's common that ISV sets custom args? if we construct own own from product code, I'm afraid custom args will be lost, if there exists one

Yes, it would be lost but I couldn't find anything about how to set custom args in an MSI, so I'm not even sure if it is possible without going around what the MSI provides. I know it's not proof, but anecdotically on my machines no entry for MSI had any custom args.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
lechacon: I copied the telemetry event from the installer failure one. I kept the type string so we can distinguish when a failure comes from an Exe and when it is from and Msi as they will likely have different errors.

src/AppInstallerCLIE2ETests/UninstallCommand.cs
yao-msft: ok, if we decide to go with this "ignore" route, please update the InstallMSI() in InstallCommand test to also do the ignoring and uncomment the tag. Thanks.

---
In reply to: [543545150](https://github.com/microsoft/winget-cli/pull/659#discussion_r543545150) [](ancestors = 543545150)

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -734,4 +734,26 @@ They can be configured through the settings file 'winget settings'.</value>
...
+    <value>Uninstalls the given package</value>
+  </data>
+  <data name="UninstallFlowStartingPackageUninstall" xml:space="preserve">
+    <value>Starting package uninstall...</value>
+  </data>
+  <data name="UninstallFlowUninstallSuccess" xml:space="preserve">
+    <value>Successfully uninstalled</value>
+  </data>
+  <data name="NoUninstallInfoFound" xml:space="preserve">
+    <value>winget cannot locate the uninstall command for this application. Please reach out to the package publisher for support.</value>
```
yao-msft, line 750: >application [](start = 63, length = 11)

forgot "package" here :)

yao-msft (approved):
:shipit:

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
yao-msft: Ah, I overlooked the index.addmanifest line in that method. This makes sense.

---
In reply to: [543567716](https://github.com/microsoft/winget-cli/pull/659#discussion_r543567716) [](ancestors = 543567716)

PR #667, opened 2020-12-16
Ensure progress widgets respect UseVT

### LINKED ISSUE #665: Text formatting is not reset at exit

winget may emit coloured text, for instance when installing a package:
```
src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
36:            context.Reporter.Info() << Resource::String::ReportIdentityFound << ' ' << Execution::NameEmphasis << name << " [" << Execution::IdEmphasis << id << ']' << std::endl;
```
However, even if the destructor of `Reporter` should reset the text formatting
```
    Reporter::~Reporter()
    {
        // The goal of this is to return output to its previous state.
        // For now, we assume this means "default".
        GetBasicOutputStream() << TextFormat::Default;
    }
```
I’ve reach a case where the text format isn't reset and spills into the next commands.
It may also spill between logs of winget-cli. I believe the error lies in the Reporter class that doesn't reset the text format properly a the end of the stream.

# Steps to reproduce

Install a package with winget.

# Expected behavior

The text format should be reset. I have to reset it by an explicit ANSI code.

```
winget install --silent git & echo [0m
```

# Actual behavior

The reset ANSI code is not emitted.

# Environment

I use v0.2.3162-preview.

dra27:
Otherwise the Spinner and Progress widgets merrily set the foreground colour but `TextFormat::Default` is never sent because the Reporter actually checks the VT setting!

Fixes #665

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/667)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionProgress.cpp
+++ b/src/AppInstallerCLICore/ExecutionProgress.cpp
@@ -133,11 +133,13 @@ namespace AppInstaller::CLI::Execution
     {
         void ProgressVisualizerBase::ApplyStyle(size_t i, size_t max, bool enabled)
         {
+            if (!UseVT())
+            {
+                // Either no style set or VT disabled
+                return;
+            }
             switch (m_style)
             {
-            case VisualStyle::NoVT:
-                // No VT means no style set
-                break;
             case VisualStyle::Retro:
                 if (enabled)
                 {


src/AppInstallerCLICore/ExecutionProgress.cpp
```diff
@@ -133,6 +133,10 @@ namespace AppInstaller::CLI::Execution
     {
         void ProgressVisualizerBase::ApplyStyle(size_t i, size_t max, bool enabled)
         {
+            if (!UseVT() || m_style == VisualStyle::NoVT)
```
yao-msft, line 136: Can you remove the "m_style == VisualStyle::NoVT" part since it's already checked in UseVT() method? Thanks.

dra27: Done - I had intended to remove the `case` below as well, as it's now unreachable

yao-msft: Thanks. It's ok to leave the NoVT case for completeness but I'm also fine if you removed it.

yao-msft (approved):

PR #694, opened 2021-01-09
Force output paths to use u8string with filesystem::path

JohnMcPMS:
## Issue
When paths with non-ASCII characters are used, logging them is resulting in an exception due to not being able to convert them to ASCII (the default behavior when forced to convert to a narrow char).

## Change
To better handle all situations of output of a path, both the logging and reporting infrastructure have been enlightened to force the use of `u8String()` when they see one.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/694)

### DIFF
--- a/src/AppInstallerCommonCore/Public/AppInstallerLogging.h
+++ b/src/AppInstallerCommonCore/Public/AppInstallerLogging.h
@@ -17,7 +17,7 @@
         auto& _aicli_log_log = AppInstaller::Logging::Log(); \
         if (_aicli_log_log.IsEnabled(_aicli_log_channel, _aicli_log_level)) \
         { \
-            std::stringstream _aicli_log_strstr; \
+            AppInstaller::Logging::LoggingStream _aicli_log_strstr; \
             _aicli_log_strstr _outstream_; \
             _aicli_log_log.Write(_aicli_log_channel, _aicli_log_level, _aicli_log_strstr.str()); \
         } \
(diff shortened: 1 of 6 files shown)


src/AppInstallerCommonCore/Public/AppInstallerLogging.h
```diff
@@ -140,6 +140,37 @@ namespace AppInstaller::Logging
 
     // Calls the various stream format functions to produce an 8 character hexadecimal output.
     std::ostream& SetHRFormat(std::ostream& out);
+
+    // This type allows us to override the default behavior of output operators for logging.
+    struct LoggingStream
+    {
+        // Force use of the UTF-8 string from a file path.
+        // This should not be necessary when we move to C++20 and convert to using u8string.
+        friend AppInstaller::Logging::LoggingStream& operator<<(AppInstaller::Logging::LoggingStream& out, std::filesystem::path& path)
```
yao-msft, line 149: >std::filesystem::path& path [](start = 107, length = 27)

Is this override with non const variant used?

yao-msft (approved):
:shipit:

msftrubengu (approved):

src/AppInstallerCommonCore/Public/AppInstallerLogging.h
JohnMcPMS: Yes; when doing overload resolution, only the exact match will be considered a better candidate than the template.  So a non-const local will not match the const& function.  Unfortunately I'm not aware of a better way to force the compiler to "do everything as you would have except in this one case".  I suppose it might be possible to add some dummy parameters to the template to make it a worse candidate function.

Issue #220, opened 2020-05-19
Export/Import apps list


brunovieira97:
# Description of the new feature/enhancement

Allow the user to export a list of all currently installed apps from WinGet.

This list can then be imported (with the specific versions too) on WinGet and its apps will be batch installed. Useful for a "dotfiles-based" computer setup, where I just have a winget-apps.json and can set my machine up with a single command.

**Poposed commands:**
 * `winget export <app-list>`
   * Here the user can specify an app list for exporting, or use `--all` or `*` for exporting the entire list.
 * `winget import <file-path>`

---
Edited by denelon:

**Stable Feature**
`winget import packages.json` - installs packages referenced in packages.json (requires sources to be configured).
`winget import -?` - displays help for the import command.

**Experimental Feature**
`winget export packages.json` - generates export file for packages in Add / Remove Programs matching installed sources.
`winget export -?` - displays help for the export command.

Note: export depends on list also being enabled.

`winget features` - displays all experimental features and their status
`winget settings` - launches settings.json to configure settings like experimental features

Edited: experimental status

gibfahn:
For an example in other package managers see Homebrew's [brew bundle](https://github.com/Homebrew/homebrew-bundle).

Nice thing about it is it's easy to add support for other tools like [Casks](https://github.com/Homebrew/homebrew-cask), [mas](https://github.com/mas-cli/mas), and [whalebrew](https://github.com/whalebrew/whalebrew), all of which are (or started as) third-party extensions to brew.

denelon:
This looks like a duplicate of #155 and #156

denelon:
The schema for the import file is located at https://aka.ms/winget-packages.schema.1.0.json

denelon:
I'll be publishing a build with "import" enabled next week if testing goes well. I'm publishing a sample file here as a starting point.

```JSON
{
	"$schema" : "https://aka.ms/winget-packages.schema.1.0.json",
	"CreationDate" : "2021- 3-18 13:14:44.599",
	"Sources" : 
	[
		{
			"Packages" : 
			[
				{
					"Id" : "Microsoft.Whiteboard"
				}
			],
			"SourceDetails" : 
			{
				"Argument" : "https://winget.azureedge.net/msstore",
				"Identifier" : "Microsoft.Winget.MSStore.Source_8wekyb3d8bbwe",
				"Name" : "msstore",
				"Type" : "Microsoft.PreIndexed.Package"
			}
		},
		{
[...]

Issue #156, opened 2020-05-15
Support a collection of packages


denelon:
As a developer I want to be able to specify a set of packages for the Package Manager to install.

Ideally this would allow a single file to be able to drive a more complex scenario that supports installing multiple packages. The shapshot ask would likely output it's results in this form. The collection could reference multiple manifests in one or more repositories. Initially the scope might need to be reduced to a single repository. If that's the case, this Issue should be split.

masx200:
I support it!

denelon:
This is not the same thing as #219

mimi89999:
Is this what other package managers call virtual packages? A package that does not have files or an installer in itself, but only has dependencies.

denelon:
Duplicate of #220

mimi89999:
Unless I misunderstood what this issue is about, it's not a duplicate of the other issue.

denelon:
The import/export feature leverages a JSON file representing a set of packages. You can either perform an export to build the file, or craft the file by hand. Once you have this file, you can "import" the file and all of the valid packages will be installed.

denelon:
Here is an example of the file:

```JSON
{
	"$schema" : "https://aka.ms/winget-packages.schema.1.0.json",
	"CreationDate" : "2021- 3- 2 14:40:17.181",
	"Sources" : 
	[
		{
			"Packages" : 
			[
				{
					"Id" : "Microsoft.Whiteboard"
				},
				{
					"Id" : "SixgillLLC.hyperlabel"
				}
			],
			"SourceDetails" : 
			{
				"Argument" : "https://winget.azureedge.net/msstore",
				"Identifier" : "Microsoft.Winget.MSStore.Source_8wekyb3d8bbwe",
				"Name" : "msstore",
				"Type" : "Microsoft.PreIndexed.Package"
			}
		},
		{
			"Packages" : 
			[
				{
					"Id" : "Microsoft.WindowsTerminalPreview"
				},
				{
					"Id" : "Microsoft.WindowsTerminal"
				},
				{
					"Id" : "Microsoft.dotnetRuntime"
				}
			],
			"SourceDetails" : 
			{
				"Argument" : "https://winget.azureedge.net/cache",
				"Identifier" : "Microsoft.Winget.Source_8wekyb3d8bbwe",
				"Name" : "winget",
				"Type" : "Microsoft.PreIndexed.Package"
			}
		}
	],
	"WinGetVersion" : "0.2.10572"
}
```

Issue #155, opened 2020-05-15
Installed Packages Snapshot


denelon:
As a developer I want to take a snapshot of the packages installed on my machine so I can sync that setup on another machine.

Effectively the initial ask was for something that would allow one to define a desired state based on what is currently installed and then configure another machine to match that state

Note:
This is a much more complicated scenario to fully resolve.
One of the concerns is that the manifest file could have been altered from the time one user installed it until the second one did

The challenge is making sure the subsequent users could be informed if something was different from what was available or originally installed on the first machine.

Bjego:
I'm really looking forward to this feature. Hopefully it will support the following changes:

- install (new components)
- update (existing components)
- uninstall (no longer needed components)

denelon:
Duplicate of #220

PR #699, opened 2021-01-13
Add Export/Import commands

lechacon:
This implements a prototype of the import and export commands. Opening as draft to gather feedback on schema and command options.
Overview of changes:
* Added a schema for a list of packages in JSON format. This lists packages (id, version, channel) and the sources they come from (name, arg, type, id).
* Export command with syntax `winget export [--output] file.json [common query filters]`. This creates a list of the matching packages. Packages that are not available from any source are still listed so that they can be imported if they are added later to a source.
* Import command with syntax `winget import [--input] file.json [--exactVersions]`. This installs all the packages from a list. Each package is installed from the source listed from it, unless there is no source specified in which case it is taken from any available. Installs latest version, unless the `--exactVersions` flag is used.

Still missing tests & resource strings.

Related: #155, #156, #220

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/winget-cli/pull/699)


src/AppInstallerCLICore/Commands/ImportCommand.cpp
```diff
@@ -0,0 +1,55 @@
...
+            // TODO
+            throw CommandException(Resource::String::VerifyFileFailedNotExist, execArgs.GetArg(Execution::Args::Type::ImportFile));
+        }
+    }
+
+    void ImportCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        context <<
+            Workflow::ReadImportFile <<
+            Workflow::SearchPackagesForImport <<
```
JohnMcPMS, line 52: I would imagine this set of tasks in the flow:
```
Read file
Add source(s) as needed
For each package in import file
    Search for appropriate version to install given inputs
    If viable package version found, proceed to install
    Else, output warning/error
```
Notably, we would *not* attempt to search for everything before installing the first package.

There are certainly complications with adding the sources and the requirement to be admin to do so, but I don't think that will be a blocker.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+        // Look for the packages needed from each source independently.
+        // If a package is available from multiple sources, this ensures we will get it from the right one.
+        for (auto& requiredSource : context.Get<Execution::Data::PackageCollection>().RequestsFromSources)
+        {
+            if (!requiredSource.Details.Name.empty())
+            {
+                // For packages that come from a specific source, find the matching available source.
+                // Match required sources with the available sources by their arguments, as they may have been added with a different name.
+                // TODO: Can we do this with identifiers without opening the source?
```
JohnMcPMS, line 131: Identifiers should be used, and they should not require the source to be opened to find them.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -0,0 +1,204 @@
...
+                {
+                    // TODO: Add option for ignoring/installing missing sources?
+                    AICLI_LOG(CLI, Error, << "Missing required source: " << requiredSource.Details.Name);
+                    AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_SOURCE_NAME_DOES_NOT_EXIST);
+                }
+            }
+
+            context << OpenNamedSource(requiredSource.Details.Name);
+            auto source = context.Get<Execution::Data::Source>();
+            aggregatedSource->AddAvailableSource(source);
```
JohnMcPMS, line 156: This pattern does not force the source to actually be used for a given package, which is what we want.  The goal should be that if a package should come from a specific source, then that source should be used exclusively.  Otherwise, the standard unnamed source (all combined) should be used.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -0,0 +1,204 @@
...
+            {
+                // Search for all the packages in the source
+                Logging::SubExecutionTelemetryScope subExecution;
+
+                // We want to do best effort to install all packages regardless of previous failures
+                auto searchContextPtr = context.Clone();
+                Execution::Context& searchContext = *searchContextPtr;
+                searchContext.Add<Execution::Data::Source>(source);
+
+                // TODO: Case insensitive?
```
JohnMcPMS, line 167: Finding the appropriate package and version to install is going to be far more complex than this.  We may not have the Id handy; we may just have the name and publisher and will require the ability to find something based on that.

But I think that having a review on the spec and schema will help solidify the requirements.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -368,4 +368,39 @@ namespace AppInstaller::CLI::Workflow
...
+        {
+            Logging::SubExecutionTelemetryScope subExecution;
+
+            // We want to do best effort to install all packages regardless of previous failures
+            auto installContextPtr = context.Clone();
+            Execution::Context& installContext = *installContextPtr;
+
+            // Extract the data needed for installing
+            installContext.Add<Execution::Data::PackageVersion>(package);
+            installContext.Add<Execution::Data::Manifest>(package->GetManifest());
```
JohnMcPMS, line 399: Is there not already a workflow task that does this step?

JohnMcPMS (changes requested):
We need to ensure that everyone is on the same page about the direction here before more work is done.  Some of my comments may not apply afterward, but I have given feedback based on how I expect export/import to work.

doc/packages.schema.json
```diff
@@ -0,0 +1,102 @@
...
+            "type": "array",
+            "required": true,
+            "minItems": 1,
+
+            "items": {
+              "description": "A package to be installed from this source",
+              "type": "object",
+              "additionalProperties": true,
+              "properties": {
+                "Id": {
```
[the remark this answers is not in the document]
lechacon: I didn't know there could be packages without a fixed Id. When does that happen? My thinking was that we either have the Id from the manifest, or it was an unmatched package we found in ARP so we have the product code or the package family name.

src/AppInstallerCLICore/Commands/ExportCommand.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace AppInstaller::CLI
+{
+    using namespace std::string_view_literals;
+
+    std::vector<Argument> ExportCommand::GetArguments() const
+    {
+        return {
+            Argument::ForType(Execution::Args::Type::OutputFile),
+            Argument::ForType(Execution::Args::Type::Query),
```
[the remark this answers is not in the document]
lechacon: Maybe not a single package, but in my mind it made sense for someone to want to say "export everything available in my private source" or "export everything I have from this ISV/publisher" or "export everything I have related to Python", not necessariliy everything in the machine. Thinking that, I copied every query parameter from list. But yes, I think it wasn't something actually asked for.

src/AppInstallerCLICore/Commands/ImportCommand.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    std::string ImportCommand::HelpLink() const
+    {
+        // TODO: point to correct location
+        return "https://aka.ms/winget-command-import";
+    }
+
+    void ImportCommand::ValidateArgumentsInternal(Execution::Args& execArgs) const
+    {
+        if (!std::filesystem::exists(execArgs.GetArg(Execution::Args::Type::ImportFile)))
```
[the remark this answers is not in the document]
lechacon: Thanks, didn't know that. Will change it.

src/AppInstallerCLICore/PackageCollection.cpp
```diff
@@ -0,0 +1,188 @@
...
+            }
+
+            THROW_HR_IF(E_NOT_VALID_STATE, node[propertyName].type() != T);
+            return node[propertyName];
+        }
+
+        // Sets a property of a JSON object to a string if it is not empty
+        void SetJsonProperty(Json::Value& node, const std::string& propertyName, const std::string& value)
+        {
+            if (!value.empty())
```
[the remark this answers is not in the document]
lechacon: I think I just assumed it wasn't good to add empty values, same for the unknown source. I'll check if there is anything standard for empty strings.

src/AppInstallerCLICore/PackageCollection.cpp
```diff
@@ -0,0 +1,188 @@
...
+            else if (node.isMember(propertyName))
+            {
+                node.removeMember(propertyName);
+            }
+        }
+
+        // Reads the description of a package from a Package node in the JSON.
+        PackageRequest ParsePackageNode(const Json::Value& packageNode)
+        {
+            std::string id = packageNode[PACKAGE_ID_PROPERTY].asString();
```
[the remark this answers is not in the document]
lechacon: I haven't done it, but it should be checked by validating the schema. But since you say we may not have an Id, I'll probably have to check here anyway.

src/AppInstallerCLICore/PackageCollection.h
```diff
@@ -0,0 +1,83 @@
...
+
+        // Requests from each individual source.
+        std::vector<PackageRequestsFromSource> RequestsFromSources;
+    };
+
+    namespace PackagesJson
+    {
+        // Strings used in the Packages JSON file.
+        // Most will be used to access a JSON value, so they need to be std::string
+        const std::string SCHEMA_PROPERTY = "$schema";
```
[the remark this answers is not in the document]
lechacon: Not by the callers, but I thought it could be useful for some unit tests.

src/AppInstallerCLICore/Commands/ImportCommand.cpp
lechacon: I disagree here. I would expect it to search everything first and tell me if there is anything missing so I can start the import, leave and come back to a full install. If I was installing a bunch of stuff, I would be quite annoyed if I came back to see that it didn't install half of what I wanted because it didn't find it without telling me upfront.

doc/packages.schema.json
```diff
@@ -0,0 +1,102 @@
...
+
+              "Type": {
+                "description": "Type of the source",
+                "type": "string",
+                "required": true
+              }
+            }
+          },
+
+          "Packages": {
```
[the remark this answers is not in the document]
lechacon: I did this based on a sample @denelon gave me of what he thought a file could look like. He may have a better reasoning for doing it this way (or maybe not, not sure).

My personal take on it is that it's likely that many people may want to view, edit or even write these files by hand, and I'd imagine most of the time they would be installing many files from a few sources or even just one. In that case, maybe it would be a slightly experience to not have dozens of lines just saying `source: "winget"`. Of course, I may be completely wrong here in how this is going to be used and what the expierence would be like.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
lechacon: > This pattern does not force the source to actually be used for a given package

What do you mean?

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
lechacon: Do you know if this is something we already have?
From what I found, the SQLIndexSource takes its identifier from the SourceDetails' Data (package family name) but we don't have anything in just the SourceDetails that gives the identifier directly, and that's all we have before opening the source. I didn't dig deep enough to find where the SourceDetails are first taken from to add the identifier there, or to know if it would be safe to always take the Data as the identifier.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
JohnMcPMS: You are finding source 1, then adding it to the aggregate, then searching.  All of these packages will come from the correct source.

Then you find source 2, add it to the aggregate, and search.  These packages could be coming from source 1 or 2.

You should simply open the source and use it for the packages in question.  No need to use the aggregate.

lechacon: When I do the search, I'm not doing it in `context` which has the aggregate, but in `searchContext` which only has the individual source, so it should use only that one source unless I'm missing something.

I'm using the aggregate because some step of the installation (can't remember which atm) references the package's source so we need to keep it alive somehow (as the package has only a weak reference). Maybe keeping it as the root context's Source was not the best idea, more so since the composite source wasn't intended to be public. I can change it to keep around the sources as a list in the context.

But this all depends on me wanting to do all the searches before installing. If we change it to searching right before each install as you say, the aggregate can go away.

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -67,6 +59,20 @@ namespace AppInstaller::CLI
             return Argument{ "type", 't', Args::Type::SourceType, Resource::String::SourceTypeArgumentDescription, ArgumentType::Positional };
         case Args::Type::ValidateManifest:
             return Argument{ "manifest", NoAlias, Args::Type::ValidateManifest, Resource::String::ValidateManifestArgumentDescription, ArgumentType::Positional, true };
+        case Args::Type::HashFile:
+            return Argument{ "file", 'f', Args::Type::HashFile, Resource::String::FileArgumentDescription, ArgumentType::Positional, true };
+        case Args::Type::Msix:
+            return Argument{ "msix", 'm', Args::Type::Msix, Resource::String::MsixArgumentDescription, ArgumentType::Flag };
+        case Args::Type::OutputFile:
```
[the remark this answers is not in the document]
lechacon: Done.

src/AppInstallerCLICore/Commands/ExportCommand.cpp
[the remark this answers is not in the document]
lechacon: Removed all the querying arguments save for source, as per offline talk with @denelon

src/AppInstallerCLICore/PackageCollection.cpp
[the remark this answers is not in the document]
lechacon: I removed this altogether as I wasn't even using it.

src/AppInstallerCLICore/PackageCollection.h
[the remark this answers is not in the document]
lechacon: Moved to the cpp. I thought about using it for tests, but saw other similar places that don't do it (e.g. the sources yaml).

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -0,0 +1,204 @@
...
+                installedVersion->GetProperty(PackageVersionProperty::Version).get(),
+                installedVersion->GetProperty(PackageVersionProperty::Channel).get(),
+            };
+            return package.GetAvailableVersion(installedVersionKey);
+        }
+
+        // Selects which version of an installed package to list when exporting.
+        std::shared_ptr<IPackageVersion> SelectPackageVersionToExport(const IPackage& package)
+        {
+            // See if the installed version is available from some source
```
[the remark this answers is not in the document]
lechacon: Removed the mapping and changed to always use the installed version, but kept the check for available versions so we can show a warning if the installed version is no longer available in the source.

src/AppInstallerCLICore/Commands/ImportCommand.cpp
lechacon: Per offline talk, we will not add the sources automatically and we will be validating the file before installing everything.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
lechacon: Added the Identifier to the SourceDetails and persisted it in the sources yaml file. When opening, the identifier is still calculated for the SQLIndexSource from the data instead of read from the file, so that it is available even if the file does not have it. For any future source type we can really read the value from the file. When adding a source, the identifier is also calculated the same way without needing to open the source.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
lechacon: There is a GetManifest task but it does more work than is needed here (e.g. checking the arguments and using them to select the version). Here we want to seed the sub context with info we already have so that it can start doing its work.

src/AppInstallerCLICore/PackageCollection.cpp
```diff
@@ -0,0 +1,187 @@
...
+            {
+                AddSourceNode(root, source);
+            }
+
+            return root;
+        }
+
+        std::optional<PackageCollection> ParseJson(const Json::Value& root)
+        {
+            // TODO: Embed schema in binaries & validate file
```
lechacon, line 166: I had some trouble getting this to work. @yao-msft is also working on something with JSON schemas and seems to have this figured out, so I'll wait for him to check in his change and then hook this up with his code.

JohnMcPMS:
> #### New misspellings found, please review:
> * wcout
> * wofstream
> * wostream

The presence of these suggests misuse; neither our production code nor our tests should be using (w)cout directly.  [Although I do frequently do some local debugging of tests with them.]

.github/actions/spelling/allow.txt
```diff
@@ -450,6 +450,7 @@ VOS
 vso
 wapproj
 wchar
+wcout
```
jsoref, line 453: https://github.com/microsoft/winget-cli/pull/699#issuecomment-772702860

lechacon: This is not in the product code or tests directly, but in a standalone .exe used as a test installer. It was mixing narrow and wide chars and that was getting complicated, but it didn't seem worth it to add conversion between them there, so I moved it all to wide. It should be fine as it's independent of everything else.

jsoref: The only thing I'd note is that by using it here instead of, e.g. just telling the bot to ignore the file entirely, the bot will never warn about the next violation.

Think of it like page-fault protection.

Obviously reviewers are responsible for the content, so in theory some reviewer should catch future incursions.

(Just feedback from the peanut gallery.)

src/AppInstallerCLITests/TestSource.h
```diff
@@ -17,7 +17,8 @@ namespace TestCommon
         using LocIndString = AppInstaller::Utility::LocIndString;
         using MetadataMap = AppInstaller::Repository::IPackageVersion::Metadata;
 
-        TestPackageVersion(const Manifest& manifest, MetadataMap installationMetadata = {});
+        TestPackageVersion(const Manifest& manifest, std::weak_ptr<const ISource> source = {});
```
JohnMcPMS, line 20: Was there a lifetime bug?

src/AppInstallerRepositoryCore/CompositeSource.h
```diff
@@ -41,7 +41,7 @@ namespace AppInstaller::Repository
         void AddAvailableSource(std::shared_ptr<ISource> source);
 
         // Sets the installed source to be composited.
-        void SetInstalledSource(std::shared_ptr<ISource> source);
+        void SetInstalledSource(std::shared_ptr<ISource> source, bool keepInstalledOnly = true);
```
JohnMcPMS, line 44: This seems unnecessary, as one can just filter out packages with no available versions.  Is that what you are using it for?

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -146,6 +147,8 @@ namespace AppInstaller::Repository::Microsoft
 
                 SQLiteIndex index = SQLiteIndex::Open(indexLocation.u8string(), SQLiteIndex::OpenDisposition::Immutable);
 
+                // We didn't use to store the source identifier, so we compute it here in case it's
```
JohnMcPMS, line 150: Strange comment with no code change.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/CompositeSource.h
lechacon: This argument is to keep the existing behavior for list and upgrade (return only installed; filter out available not-installed packages) while extending it to allow me to get all the packages whether installed or not. For import we need all those packages here so we can do all the cases of "this is already installed, nothing to do", "this is already installed but can be updated" and "this is not yet installed".

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
lechacon: Since I added the identifier to the source details, it may look like it doesn't make sense to get the identifier here as `GetPackageFamilyNameFromDetails(details)` instead of as `details.Identifier`. But we need to do it because we didn't have the identifier in the sources file before, so it may be missing from `details`. I added the comment to clear it up for the future.

src/AppInstallerCLICore/PackageCollection.cpp
```diff
@@ -0,0 +1,187 @@
...
+            Json::Value root = CreateRoot(packages.ClientVersion);
+            for (const auto& source : packages.Sources)
+            {
+                AddSourceNode(root, source);
+            }
+
+            return root;
+        }
+
+        std::optional<PackageCollection> ParseJson(const Json::Value& root)
```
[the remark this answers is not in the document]
lechacon: Good catch. It was from when I tried adding schema validation (in that case it returned empty), but then I removed it. I need to check what @yao-msft did for validation errors to update this.

src/AppInstallerCLICore/Commands/ImportCommand.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace AppInstaller::CLI
+{
+    using namespace std::string_view_literals;
+
+    std::vector<Argument> ImportCommand::GetArguments() const
+    {
+        return {
+            Argument{ "import-file", 'i', Execution::Args::Type::ImportFile, Resource::String::ImportFileArgumentDescription, ArgumentType::Positional, true },
+            Argument{ "exact-versions", Argument::NoAlias, Execution::Args::Type::ExactVersions, Resource::String::ExactVersionsArgumentDescription, ArgumentType::Flag },
```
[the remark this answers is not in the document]
lechacon: The file can already drive it as you say if the flag is not present. I had the flag because that wasn't the case before. I'll remove it.

src/AppInstallerCLITests/TestSource.h
lechacon: There wasn't one. I changed it from shared to weak because of another change I made. I made the test composite source pass a pointer to `this` to the packages it returns and there was some issue with it being shared, but I can't remember exactly what.

schemas/JSON/packages/packages.schema.1.0.json
```diff
@@ -0,0 +1,97 @@
...
+
+  "title": "winget Packages List Schema",
+  "description": "Describes a list of packages for batch installs",
+
+  "type": "object",
+  "required": [ "WinGetVersion", "Sources" ],
+  "additionalProperties": true,
+
+  "properties": {
+    "WinGetVersion": {
```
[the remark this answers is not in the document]
lechacon: The json should already have a `$schema` property, and that URI is versioned.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -3,84 +3,72 @@
...
+                << "Installed package is available. Package Id [" << availablePackageVersion->GetProperty(PackageVersionProperty::Id) << "], Source [" << sourceIdentifier << "]");
+
+            // Find the exported source for this package
+            auto sourceItr = std::find_if(exportedSources.begin(), exportedSources.end(), [&](const PackageCollection::Source& s) { return s.Details.Identifier == sourceIdentifier; });
+            if (sourceItr == exportedSources.end())
             {
-                requestsFromSource.emplace_back(Utility::LocIndString{ sourceIdentifier }, packageVersion->GetSource()->GetDetails());
-                sourceItr = std::prev(requestsFromSource.end());
+                exportedSources.emplace_back(availablePackageVersion->GetSource()->GetDetails());
+                sourceItr = std::prev(exportedSources.end());
```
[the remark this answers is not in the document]
lechacon: `back()` returns the last element, but I wanted an iterator to it to match the result of `std::find_if()`. I couldn't think of a better way to do "find or add".

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -94,111 +82,159 @@ namespace AppInstaller::CLI::Workflow
...
+                        foundAll = false;
+                        continue;
+                    }
                 }
+
+                packagesToInstall.push_back(std::move(searchContext.Get<Execution::Data::PackageVersion>()));
             }
         }
 
+        if (!foundAll)
```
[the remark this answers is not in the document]
lechacon: Will add a `--force` flag for this.

src/AppInstallerRepositoryCore/CompositeSource.h
JohnMcPMS: Ok, I think the name is confusing then.  I thought it meant:
```
true == results would include all installed packages
false == results would only include packages that are installed and correlated to an available package
```

I think it should be something more like `includeAllAvailablePackagesInSearchResults` or use an enum instead to drive the intended behavior:
```
// inside CompositeSource
enum class SearchBehavior
{
   AllInstalled,
   AllPackages,
   // More in the future maybe
};
```

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -139,16 +178,14 @@ namespace AppInstaller::CLI::Workflow
                 AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_SOURCE_NAME_DOES_NOT_EXIST);
             }
 
-            context << Workflow::OpenNamedSource(requiredSource.Details.Name);
+            context <<
```
JohnMcPMS, line 181: What I meant was:
```
context << Workflow::OpenNamedSourceForSources(requiredSource.Details.Name);
```
That way the source is never on the Sources data element and won't accidentally be used for that.

JohnMcPMS (changes requested):
Just want to clarify how version should be handled by default, as I believe that wanting a specific version is a bit of an edge case (at least wanting a specific version of every package).  So I think that export should not include the version, but allow for it with a flag.  And import should probably just have flag more like `--ignoreVersions` that will ignore every version rather than allow fallback to latest in some cases.

quangkieu:
@JohnMcPMS how about by default is >= export version. User could add the top limit

JohnMcPMS (approved):

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -108,11 +108,18 @@ namespace AppInstaller::CLI::Workflow
             }
 
             // Take the Id from the available package because that is the one used in the source,
-            // but take the exported version from the installed package.
-            sourceItr->Packages.emplace_back(
-                availablePackageVersion->GetProperty(PackageVersionProperty::Id),
-                version.get(),
-                channel.get());
+            // but take the exported version from the installed package if needed.
+            if (context.Args.Contains(Execution::Args::Type::IncludeVersions))
```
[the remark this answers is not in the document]
lechacon: Updated to do that.

PR #724, opened 2021-01-28
Name and publisher normalization for cross reference

JohnMcPMS:
## Change
This change adds code for creating normalized strings from package names and publishers.  It is not actually used anywhere yet, but it is intended for use in correlating locally installed packages with remote ones, as well correlating one version to the next.  I also plan to use the values to create "nicer" Ids for locally installed packages that are not found in a remote source.

In order to support all of Unicode, the code does not use STL's regex, but instead an ICU regex wrapper is implemented to enable global support.  This wrapper's interface is purpose built for this use case, so it may need to be adapted as needed for other uses.

The expressions and their use were adapted from code that is used to do similar work for app compatibility on Windows.  The goal is to remove versions and extraneous strings from names and publishers to create values unique to the package.  The code also attempts to capture the architecture and locale if it is present in the name.

A large set of real world examples can be found in the test data files.  These contain real name `InputNames.txt` and publisher `InputPublishers.txt` strings from packages found in winget-pkgs, as well as a few others added to smoke test proper functionality outside of English and other European character sets.  The result file `NormalizationInitialIds.txt` contains the strings that result from normalizing both name and publisher and appending them as `Publisher + . + Name`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/724)


.github/actions/spelling/patterns.txt
```diff
@@ -17,3 +17,5 @@ data:[a-zA-Z=;,/0-9+-]+
 El proyecto .* diferentes
 # Package family names
 \b[-.A-Za-z0-9]+_[a-z0-9]{13}\b
+# Locales for name normalization
+\b\p{Lu}{2,3}(-(CANS|CYRL|LATN|MONG))?-\p{Lu}{2}(?![A-Z])(?:-VALENCIA)?\b
```
jsoref, line 21: I'd generally suggest doing:
```suggestion
\b\p{Lu}{2,3}(?:-(?:CANS|CYRL|LATN|MONG))?-\p{Lu}{2}(?![A-Z])(?:-VALENCIA)?\b
```
Basically, if you aren't capturing you should tell perl you aren't.

Otherwise, if you have a rule like:
```
\b([A-Za-z])\1{3,}\b
```
(which is earlier in this file), you may be unhappy.

JohnMcPMS: The entire file is run as one big regex?  I would assume that the capture reference would be per-line.  I don't mind changing it, I just didn't expect that the captures really mattered for this use case.

jsoref: Yeah, it's run as one big regex.

It's a lazy optimization that only trips things up occasionally.

I need to look into whether it's worth treating each line as a unique regex / how to make that work.

src/AppInstallerCommonCore/NameNormalization.cpp
```diff
@@ -0,0 +1,441 @@
...
+                L"GA-IE", L"GD-DB", L"GL-ES", L"GSW-FR", L"GU-IN", L"HE-IL", L"HI-IN", L"HR-BA", L"HR-HR", L"HSB-DE", L"HU-HU",
+                L"HY-AM", L"ID-ID", L"IG-NG", L"II-CN", L"IS-IS", L"IT-CH", L"IT-IT", L"JA-JP", L"KA-GE", L"KK-KZ", L"KL-GL",
+                L"KM-KH", L"KN-IN", L"KOK-IN", L"KO-KR", L"KY-KG", L"LB-LU", L"LO-LA", L"LT-LT", L"LV-LV", L"MI-NZ", L"MK-MK",
+                L"ML-IN", L"MN-MN", L"MOH-CA", L"MR-IN", L"MS-BN", L"MS-MY", L"MT-MT", L"NB-NO", L"NE-NP", L"NL-BE", L"NL-NL",
+                L"NN-NO", L"NSO-ZA", L"OC-FR", L"OR-IN", L"PA-IN", L"PL-PL", L"PRS-AF", L"PS-AF", L"PT-BR", L"PT-PT", L"QUT-GT",
+                L"QUZ-BO", L"QUZ-EC", L"QUZ-PE", L"RM-CH", L"RO-RO", L"RU-RU", L"RW-RW", L"SAH-RU", L"SA-IN", L"SE-FI", L"SE-NO",
+                L"SE-SE", L"SI-LK", L"SK-SK", L"SL-SI", L"SMA-NO", L"SMA-SE", L"SMJ-NO", L"SMJ-SE", L"SMN-FI", L"SMS-FI", L"SQ-AL",
+                L"SV-FI", L"SV-SE", L"SW-KE", L"SYR-SY", L"TA-IN", L"TE-IN", L"TH-TH", L"TK-TM", L"TN-ZA", L"TR-TR", L"TT-RU",
+                L"UG-CN", L"UK-UA", L"UR-PK", L"VI-VN", L"WO-SN", L"XH-ZA", L"YO-NG", L"ZH-CN", L"ZH-HK", L"ZH-MO", L"ZH-SG",
```
yao-msft, line 326: >L"ZH-CN" [](start = 86, length = 8)

The comment said the values must be in sorted order but looks like no?

src/AppInstallerCommonCore/NameNormalization.cpp
```diff
@@ -0,0 +1,441 @@
...
+        // arbitrary new processor architecture from names would hopefully only affect existing packages
+        // that were not matching properly. Fixing a bug that was causing bad strings to be produced would
+        // be impactful, and thus should likely result in a new iteration.
+        class NormalizationInitial : public details::INameNormalizer
+        {
+            static std::wstring PrepareForValidation(std::string_view value)
+            {
+                std::wstring result = Utility::Normalize(ConvertToUTF16(value));
+                Trim(result);
+                size_t atPos = result.find(L"@@", 3);
```
yao-msft, line 37: >@@ [](start = 45, length = 2)

what's special with this?

yao-msft (approved):
:shipit:

src/AppInstallerCommonCore/NameNormalization.cpp
JohnMcPMS: They also need to be folded, which is a harder thing to expect people to do.  So I fold and sort them into another container in the constructor.  I can update the comment to better reflect this.

src/AppInstallerCommonCore/NameNormalization.cpp
JohnMcPMS: I have no idea; the starting point for this code was something doing similar work, and I left the majority of it alone.

PR #740, opened 2021-02-02
Implement V1 manifest  and schema validation

yao-msft:
## Change
- Implemented schema validation using Valijson
- Added manifest schemas for preview and v1 manifests
- Changed Manifest object model to better reflect v1 manifests
- Changed ManifestLocalization to be std::map based for easier future locale processing
- Implemented merge logic for multiple file manifests

## Validation
- Existing tests and added new tests
- Validated all 3000+ manifests currently in winget-pkgs passed manifest validation

Note: After this change, we might need to revisit and relax restrictions in some of v1 manifest schemas as I found around 300+ manifests in current winget-pkgs will fail in new restrictions imposed by V1 manifest schemas

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/740)


src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -18,18 +18,6 @@ namespace AppInstaller::CLI::Workflow
...
 
-    void EnsureMinOSVersion(Execution::Context& context)
-    {
-        const auto& manifest = context.Get<Execution::Data::Manifest>();
-
-        if (!manifest.MinOSVersion.empty() &&
-            !Runtime::IsCurrentOSVersionGreaterThanOrEqual(Version(manifest.MinOSVersion)))
-        {
-            context.Reporter.Error() << Resource::String::InstallationRequiresHigherWindows << ' ' << manifest.MinOSVersion << std::endl;
-            AICLI_TERMINATE_CONTEXT(HRESULT_FROM_WIN32(ERROR_OLD_WIN_VERSION));
```
JohnMcPMS, line 29: I understand the motivation to remove this, but it does mean that we will not message the reason for failure to the user (or ourselves for the failure code).  Maybe it is upcoming in my reading of the PR, but we need a way to message the failure state of no applicable installers.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -33,8 +40,16 @@ namespace AppInstaller::CLI::Workflow
...
-            Manifest::ManifestInstaller::InstallerTypeEnum installedType)
+            Manifest::InstallerTypeEnum installedType)
         {
+            bool isOSVersionSatisfied1 = installer1.MinOSVersion.empty() || Runtime::IsCurrentOSVersionGreaterThanOrEqual(Utility::Version(installer1.MinOSVersion));
+            bool isOSVersionSatisfied2 = installer2.MinOSVersion.empty() || Runtime::IsCurrentOSVersionGreaterThanOrEqual(Utility::Version(installer2.MinOSVersion));
+
+            if (isOSVersionSatisfied1 && !isOSVersionSatisfied2)
+            {
+                return true;
+            }
```
JohnMcPMS, line 51: Let's re-write the loop to check `IsInstallerApplicable` first before ever calling `IsInstallerBetterMatch`.  That way we don't need to duplicate these kinds of boolean checks.  It ends up making this function just the arch > arch comparison.  Not sure why I didn't think of it when I rewrote this a while back.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
yao-msft: Yes, the MinOSVersion check is moved to installer selection logic since in the new world each installer has its own MinOSVersion. It will be part of the No_Applicable_Installer error

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -77,8 +80,34 @@ namespace AppInstaller::CLI::Workflow
     {
         const auto& manifest = context.Get<Execution::Data::Manifest>();
 
+        // Channel is moved to installer level, get all channels from all installers
```
JohnMcPMS, line 83: >// Channel is moved to installer level, get all channels from all installers [](start = 8, length = 76)

Channel is not an installer level item.  It is a part of the version.  If the schema has this change, I obviously missed it in review.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -524,16 +525,28 @@ namespace AppInstaller::CLI::Workflow
...
+        }
+    }
+
     void GetManifestFromArg(Execution::Context& context)
     {
         Logging::Telemetry().LogIsManifestLocal(true);
 
         context <<
-            VerifyFile(Execution::Args::Type::Manifest) <<
+            VerifyPath(Execution::Args::Type::Manifest) <<
```
JohnMcPMS, line 544: >VerifyPath [](start = 12, length = 10)

This one does not seem like it should have changed, `winget install -m a_directory` will now work?

src/AppInstallerCLIE2ETests/HashCommand.cs
```diff
@@ -39,7 +39,7 @@ public void HashFileNotFound()
         {
             var result = TestCommon.RunAICLICommand("hash", TestCommon.GetTestDataFile("DoesNot.Exist"));
             Assert.AreEqual(Constants.ErrorCode.ERROR_FILE_NOT_FOUND, result.ExitCode);
-            Assert.True(result.StdOut.Contains("File does not exist"));
+            Assert.True(result.StdOut.Contains("Path does not exist"));
```
JohnMcPMS, line 42: >Path [](start = 48, length = 4)

This one should not have changed.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
yao-msft: It will work if the directory contains only one manifest(either split manifest files or a singleton inside). The input will be checked in ValidateInput method in manifest parsing

---
In reply to: [571315313](https://github.com/microsoft/winget-cli/pull/740#discussion_r571315313) [](ancestors = 571315313)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: But I think we need to do a better job of messaging the reason that there are not applicable installers.  Some of that burden could be placed on search to filter them out before we even attempt to install, but I prefer to show everything and have the user be informed that we don't have their flavor of installer rather than we don't even know what they are talking about.

It might be more complex, and thus not piled on to this PR, but I think we should attempt to give some reason as to why there are no applicable installers.  Imagine:
```
for each installer
  reason[i] = GetAllReasonsWhyNotApplicable(installer)
if (AndAllReasons() != 0)
  ReportReasons(AndAllReasons())
```
So if all installers require a newer OS, we say that.  If they are all ARM, we can say that they are for a different processor architecture.  If one required a newer OS and one is ARM, we say both somehow.

---
In reply to: [571311595](https://github.com/microsoft/winget-cli/pull/740#discussion_r571311595) [](ancestors = 571311595)

src/AppInstallerCLITests/TestData/ManifestV1-MultiFile-DefaultLocale.yaml
```diff
@@ -0,0 +1,23 @@
+PackageIdentifier: microsoft.msixsdk
```
JohnMcPMS, line 1: These split manifests should be in a directory for organization and testing.  Was it complicating the copy on build?

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -0,0 +1,207 @@
...
+            else
+            {
+                THROW_HR(E_UNEXPECTED);
+            }
+
+            return result;
+        }
+    }
+
+    std::string LoadResourceAsString(PCWSTR resourceModuleName, PCWSTR resourceName, PCWSTR resourceType)
```
JohnMcPMS, line 89: >PCWSTR [](start = 37, length = 6)

std::wstring_view

JohnMcPMS: Although I suppose that doesn't require null termination, so it is suspect.  We should probably add a utility type that can handle this for us.

---
In reply to: [571321711](https://github.com/microsoft/winget-cli/pull/740#discussion_r571321711) [](ancestors = 571321711)

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -0,0 +1,207 @@
...
+    {
+        HMODULE resourceModule = GetModuleHandle(resourceModuleName);
```
JohnMcPMS, line 91: >GetModuleHandle [](start = 33, length = 15)

If you are going to allow arbitrary modules here, you should use the proper, thread safe GetModuleHandleEx and a wil::unique_hmodule to hold it.

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
```diff
@@ -3,448 +3,486 @@
...
+        // Input validations:
+        // - Determine manifest version
+        // - Check multi file manifest input integrity
+        //   - All manifests use same PackageIdentifier, PackageVersion, ManifestVersion
+        //   - All required types exist and exist only once. i.e. version, installer, defaultLocale
+        //   - No duplicate locales across manifests
+        //   - DefaultLocale matches in version manifest and defaultLocale manifest
+        // - Validate manifest type correctness
+        //   - Allowed file type in multi file manifest: version, installer, defaultLocale, locale
+        //   - Allowed file type in multi file manifest: preview manifest, merged and singleton
```
yao-msft, line 23: >multi [](start = 36, length = 5)

single

src/AppInstallerCommonCore/Public/winget/ManifestLocalization.h
```diff
@@ -3,20 +3,169 @@
...
+        Tags,
+        Max
+    };
+
+    namespace details
+    {
+        template <Localization L>
+        struct LocalizationMapping
+        {
+            // value_t type specifies the type of this data
```
JohnMcPMS, line 36: >// value_t type specifies the type of this data [](start = 12, length = 47)

Since almost everything is going to be string_t, you could define it here in the unspecialized definition.  Then one only needs to specify the very few exceptions to the rule (1 right now).

src/AppInstallerCommonCore/Public/winget/ManifestYamlParser.h
```diff
@@ -3,61 +3,51 @@
...
-            bool fullValidation);
-
-        void PrepareManifestFieldInfos(const ManifestVer& manifestVer);
+        ManifestTypeEnum ManifestType;
     };
+
+    // fullValidation: Bool to set if manifest creation should perform extra validation that client does not need.
+    //                 e.g. Channel should be null. Client code does not need this check to work properly.
+    // throwOnWarning: Bool to indicate if an exception should be thrown with only warnings detected in the manifest.
+    // resourceDll:    Binary where schemas are embedded, or nullptr if they are embedded in the binary that created the process
```
JohnMcPMS, line 26: >// resourceDll:    Binary where schemas are embedded, or nullptr if they are embedded in the binary that created the process [](start = 4, length = 124)

This seems like it is only marginally useful for some negative test case.  Isn't it possible to determine the module that the current code is executing in?

src/WinGetUtil/WinGetUtil.h
```diff
@@ -76,9 +76,11 @@ extern "C"
     // Validates a given manifest. Returns a bool for validation result and
     // a string representing validation errors if validation failed.
     WINGET_UTIL_API WinGetValidateManifest(
```
JohnMcPMS, line 78: >WinGetValidateManifest [](start = 20, length = 22)

This will need to be a new function and the old one will need to continue to work with it's existing signature.

src/WinGetUtil/Exports.cpp
```diff
@@ -173,16 +173,18 @@ extern "C"
     CATCH_RETURN()
 
     WINGET_UTIL_API WinGetValidateManifest(
-        WINGET_STRING manifestPath,
+        WINGET_STRING inputPath,
         BOOL* succeeded,
-        WINGET_STRING_OUT* message) try
+        WINGET_STRING_OUT* message,
+        WINGET_STRING mergedManifestPath,
+        BOOL isPartialManifest) try
```
JohnMcPMS, line 180: >isPartialManifest [](start = 13, length = 17)

We need to know the expected type, or return it, or the service code will need to parse the manifest enough to answer the question of "Does this file name pattern match the type in the manifest?".

Rather than a BOOL here, we should take in an enum with values for single file, split, sub-component of a split.  It could even be an in-out where we also let them give us "don't know" and we give them back the answer.

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerCommonCore/Public/winget/ManifestYamlParser.h
yao-msft: It's mainly for WinGetUtil.dll, I cannot find a simple way to detect the code is being called by wingetutil(or you prefer a bool like isFromWinGetUtil)?

---
In reply to: [571327922](https://github.com/microsoft/winget-cli/pull/740#discussion_r571327922) [](ancestors = 571327922)

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
yao-msft: I'll add a todo for better message regarding installer selection

---
In reply to: [571318031](https://github.com/microsoft/winget-cli/pull/740#discussion_r571318031) [](ancestors = 571318031,571311595)

src/AppInstallerCommonCore/Public/winget/ManifestYamlParser.h
JohnMcPMS: Either use GetModuleHandleEx or [https://devblogs.microsoft.com/oldnewthing/20041025-00](https://devblogs.microsoft.com/oldnewthing/20041025-00)

---
In reply to: [571330356](https://github.com/microsoft/winget-cli/pull/740#discussion_r571330356) [](ancestors = 571330356,571327922)

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
yao-msft: per offline chat, we decided to remove the resourceModuleName parameter, leave the other 2 PCWSTR as they are because that's what MAKEINTRESOURCE produces and later consumed by resource related apis

---
In reply to: [571322386](https://github.com/microsoft/winget-cli/pull/740#discussion_r571322386) [](ancestors = 571322386,571321711)

JohnMcPMS (approved):
:shipit:

PR #741, opened 2021-02-03
Azure pipeline to generate and publish Microsoft.WindowsPackageManager.Utils nuget package

palenshus:
- Adding nuspec for the package
- Updating Update-BinVer.ps1 to allow passing in the major/minor version instead of getting from release tag
- Creating new build pipeline which takes a major/minor version as a parameter, updates the version in source and builds and signs, generates a nuget package and signs that, then publishes it

I haven't tested the actual push command yet, as I'd like to get PR approval before doing so

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/741)


WinGetUtil.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <id>Microsoft.WindowsPackageManager.Utils</id>
+    <version>$version$</version>
+    <title></title>
+    <authors>Microsoft</authors>
+    <owners></owners>
+    <projectUrl>https://github.com/microsoft/winget-cli</projectUrl>
+    <license type="expression">MIT</license>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>The utility binary for use with the WinGet CLI.</description>
+    <releaseNotes>Summary of changes made in this release of the package.</releaseNotes>
```
lechacon, line 13: Remove?

azure-pipelines.nuget.yml
```diff
@@ -0,0 +1,141 @@
...
+      - task: NuGetCommand@2
+        displayName: Restore Packages
+        inputs:
+          restoreSolution: "$(solution)"
+
+      - task: DotNetCoreCLI@2
+        displayName: DotNet Restore
+        inputs:
+          command: "restore"
+          projects: "**/*.csproj"
```
lechacon, line 36: Restoring the solution doesn't restore individual projects?

azure-pipelines.nuget.yml
```diff
@@ -0,0 +1,141 @@
...
+        displayName: Build Solution x64
+        inputs:
+          platform: "x64"
+          solution: "$(solution)"
+          configuration: "$(buildConfiguration)"
+
+      - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
+        displayName: "ESRP CodeSigning - Package contents"
+        inputs:
+          ConnectedServiceName: "Peet ESRP CodeSigning"
```
lechacon, line 63: Maybe rename the connection to remove the internal team name?

jsoref:
For `ative`, you probably want to add a pattern.

Possibilities include:
`\\native`
`target="[^"]*"`

azure-pipelines.nuget.yml
palenshus: It's a good question, I copied this from https://github.com/microsoft/winget-cli/blob/master/azure-pipelines.yml. I'll try to find out why it's in there

palenshus: We believe it was just an oversight, trying without...

palenshus: Tried only restoring the .sln, that failed. Trying with only csproj

palenshus: Turns out both are necessary. "Dotnet restore" only restores csproj's, not the vcxproj's. The "nuget restore sln" restores both types, but some projects fail at build time with the following:

```
##[error]C:\Program Files\dotnet\sdk\5.0.102\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(241,5): Error NETSDK1004: Assets file 'D:\a\1\s\src\LocalhostWebServer\obj\project.assets.json' not found. Run a NuGet package restore to generate this file.

##[error]C:\Program Files\dotnet\sdk\5.0.102\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(241,5): Error NETSDK1004: Assets file 'D:\a\1\s\src\IndexCreationTool\obj\project.assets.json' not found. Run a NuGet package restore to generate this file.
```

azure-pipelines.yml
```diff
@@ -1,314 +1,315 @@
...
   paths:
     include:
-    - azure-pipelines.yml
-    - src/*
+      - azure-pipelines.yml
+      - src/*
 
 pool:
-  vmImage: 'windows-latest'
+  vmImage: "windows-latest"
```
lechacon, line 16: Is there any difference in this file beyond whitespace and quotes?

azure-pipelines.nuget.yml
lechacon: Can you add a comment that both are needed?

lechacon (approved):

azure-pipelines.yml
palenshus: Oops, not anymore (I'd tried de-duping the "restore" steps, but to no avail) I'll revert this file

azure-pipelines.nuget.yml
palenshus: Sure, good idea

azure-pipelines.nuget.yml
```diff
@@ -0,0 +1,143 @@
+trigger: none
+
+parameters:
+  - name: version
+    displayName: Version to stamp on binaries and nuget package
```
JohnMcPMS, line 5: Include an example of the format expected by this parameter.  I don't know if `displayName` is the right field for that, but if there is not another one to use then it would be fine.

azure-pipelines.nuget.yml
```diff
@@ -0,0 +1,143 @@
...
+          filePath: 'src\binver\Update-BinVer.ps1'
+          arguments: '-TargetFile binver\binver\version.h -BuildVersion $(BuildVer) -MajorMinorOverride ${{ parameters.version }}'
+          workingDirectory: "src"
+
+      - task: VSBuild@1
+        displayName: Build Solution x86
+        inputs:
+          platform: "x86"
+          solution: "$(solution)"
+          configuration: "$(buildConfiguration)"
```
JohnMcPMS, line 53: Is there not a way to build only the WinGetUtil (and associated dependencies)?

JohnMcPMS (approved):
Only a minor change requested.

azure-pipelines.nuget.yml
palenshus: I did look for that, since building the whole solution takes so long! I couldn't find anything though. I did try just building the csproj, but that failed of course, and I didn't find any parameters on the various msbuild tasks that supported specifying only a certain project and dependencies. One could hardcode a chain of build commands to build all the individual dependencies and then wingetutil but that didn't feel right.

azure-pipelines.nuget.yml
palenshus: There doesn't appear to be a better field for this, I'll put it in the displayName. Only other fields are "default" and "values"

jsoref:
I'm pretty sure right now that the bot is still upset with you.

I'm going to make a PR this weekend to make that more visible (upgrading the bot to the latest release).

azure-pipelines.nuget.yml
lechacon: I believe you can specify that by setting the target to build as an argument to MSBuild. Like `msbuildArguments: "/target:WingetUtil"`

.github/actions/spelling/patterns.txt
```diff
@@ -17,3 +17,10 @@ data:[a-zA-Z=;,/0-9+-]+
 El proyecto .* diferentes
 # Package family names
 \b[-.A-Za-z0-9]+_[a-z0-9]{13}\b
+# URLs
+(?:(?:(?:ht|f)tp(?:s?))\://)?(?:www.|[a-zA-Z].)[a-zA-Z0-9\-\.]+\.(?:com|edu|gov|mil|net|org|biz|info|name|museum|us|ca|uk)(?:\:[0-9]+)*(?:/(?:$|[a-zA-Z0-9\.\,\;\?\'\\\+&%\$#\=~_\-]+))*
```
jsoref, line 21: fwiw, often `http://` is a bug of `https://`.
And people regularly misspell domains and pieces of urls.

You're of course welcome to ignore both.

I regularly find typos in domains and paths, and regularly find places that are using `http:` when they really should use `https:`.

.github/actions/spelling/patterns.txt
```diff
@@ -17,3 +17,10 @@ data:[a-zA-Z=;,/0-9+-]+
...
+# All tokens bounded by a slash, as these will be switches or parts of a file path for example
+(\\|/)\S+
+\S+(\\|/)
```
jsoref, line 24: Filenames are also regularly misspelled in my experience. You can of course ignore everything if you don't want the checking.

For microsoft/terminal, they have 3 files to specifically allow a small set of `\r`/`n`/`\t` names:
https://github.com/microsoft/terminal/tree/main/.github/actions/spelling/patterns

jsoref (commented):
Obviously it's up to the project, but I wouldn't do this.

.github/actions/spelling/patterns.txt
palenshus: Fair enough. If I want to allow this specific url (http://rfc3161.gtm.corp.microsoft.com/TSS/HttpTspServer) should I just add it to allow.txt, or create a pattern for it?

palenshus:
@jsoref is there a way I can run the spelling checker tool locally so I don't have to keep bombarding the PR with iterations while I troubleshoot this?

.github/actions/spelling/patterns.txt
palenshus: Removed catch-all slash rules, and added a few specific patterns. Adding more pattern files like Terminal repo has would be a good idea I think, but I'll let the repo owners tackle that :)

.github/actions/spelling/patterns.txt
palenshus: For now, I've created a pattern for it

jsoref:
@palenshus: if you enable actions in your fork, a push to any branch (not just one associated with a PR) will trigger the action.

jsoref:
@palenshus: there's an optional `advice` file that one could add which could help people learn to do that.

(At some point I'll probably collate some of this advice into the defaults... I'm still collecting feedback.)

.github/actions/spelling/patterns.txt
jsoref: allow only tolerates individual "words", so you could only add:
`rfc`
`gtm`
`corp`
`TSS`
`Http`
`Tsp`
`Server` (this should already be in the dictionary)

So, yes, if you have a specific sequence of tokens you have to use a pattern. -- Sorry it's a bit awkward. Again, still collecting feedback.

jsoref: Fwiw, here's the advice text from another project:

> * You can probably just add items into `.github/actions/spell-check/expect.txt`.
> * If you need to use a specific token in one place and it shouldn't generally be used, you can
> add an item in `.github/actions/spell-check/patterns.txt`.

PR #749, opened 2021-02-09
Adding PDBs for WingetUtil to nuget package

palenshus:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/749)


JohnMcPMS (approved):
One assumes that these files are already present where they are referenced and we are just picking them up.

palenshus:
> One assumes that these files are already present where they are referenced and we are just picking them up.

Yep, that's correct. The other pipeline actually publishes them to artifacts already

PR #758, opened 2021-02-17
Add Hyperlink function to ViritualTerminal TextFormat namespace

### LINKED ISSUE #558: Add OSC8 support for generating hyperlinks.

Update https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCLICore/VTSupport.cpp to support hyperlinks that will become available to the console and Windows Terminal with https://github.com/microsoft/terminal/pull/7251

Helps with https://github.com/microsoft/winget-cli/issues/501.

marstr:
Adding VirtualTerminal TextFormat helper for making text a clickable link.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/758)

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -140,6 +140,7 @@ hstring
 html
 http
 https
+Hyperlink
 IApplication
 IAppx
 IAsync
--- a/src/AppInstallerCLICore/VTSupport.cpp
+++ b/src/AppInstallerCLICore/VTSupport.cpp
@@ -149,6 +149,13 @@ namespace AppInstaller::CLI::VirtualTerminal
         {
 
         }
+
+        ConstructedSequence Hyperlink(const std::string& ref, const std::string& text)
+        {
+            std::ostringstream result;
+            result << AICLI_VT_OSC "8;;" << ref << AICLI_VT_ESCAPE << "\\" << text << AICLI_VT_OSC << "8;;" << AICLI_VT_ESCAPE << "\\";
+            return ConstructedSequence{ result.str() };
+        }
     }
 
     namespace TextModification
--- a/src/AppInstallerCLICore/VTSupport.cpp
+++ b/src/AppInstallerCLICore/VTSupport.cpp
@@ -150,7 +150,7 @@ namespace AppInstaller::CLI::VirtualTerminal
 
         }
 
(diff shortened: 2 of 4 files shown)


JohnMcPMS:
I removed the link to #558 as I believe the spirit of that issue is that we should also use the VT support when appropriate.

Let us know if you want tackle that in this PR as well so that we don't merge it before you are ready.  If you don't, I'm perfectly happy to accept the assist and use it when we get the chance.

marstr:
Thanks, @JohnMcPMS. I'd be happy to break this into two PRs. This one to add the helper, and then if I get some time this weekend or over lunch or something I can look into plumbing the rest through. 

I should point out, WT seems to automatically make links that are emitted clickable. For instance, project page URLs in the manifest are already navigable. We could make that text explicitly a link, but this helper will be most helpful when text that is not intrinsically a link should navigate somewhere. Maybe well-known license names could point to their text? Any other scenarios that you had in mind?

JohnMcPMS:
I hadn't really paid attention to Terminal making links automatically.  I think that means that we do actually have all of our current uses covered (at least in Terminal), although we should probably do the right thing and output the VT sequence.  We could use it to compress some of our output, such as having `show` only output one line for License that shows the text and has the LicenseUrl link.  That assumes that the terminal we are running in does support these OSC codes, which may not be true.  And it would probably mean moving some knowledge of whether VT is enabled up to the business logic layer, which isn't great either.

JohnMcPMS (approved):

WSLUser:
> that the terminal we are running in does support these OSC codes, which may not be true

I should point out Mintty and ConEmu, the 2 other major terminals on Windows support the OSC 8 sequence and I believe a few others that ported to Windows also support it. Fundamentally we'll need to output the OSC 8 VT sequence in the case of using conhost (which doesn't automatically display hyperlinks unlike Terminal) or a terminal running on top of conhost that doesn't support the VT sequence natively. I think the general idea is that while there are Windows specific APIs for the console for compatibility purposes, new CLI applications and terminals should support using the VT sequences provided by the console or implement the support themselves (if a terminal). It's also possible that one day winget gets run under a .net5 cli on Linux or Mac to install .net applications more easily. Supporting the necessary VT sequences would become more critical at that point.

marstr:
We should probably update #558to include this conversation, and the new understanding of the requirements.

PR #765, opened 2021-02-22
Rest source addition

ashpatil-msft:
- Ability to add a Rest based source by turning on the Rest source experimental feature in settings
- Ability to search for a package from rest source. Currently the search implementation doesn't include filters and inclusions and those will be added in subsequent PRs. Searching for a package will fetch all the results in the Rest source currently.

Todos:
- Complete implementing other methods that have Todo's in the PR.
- Use Search request's filters and inclusions to the Rest API search call.
- Add Json parser to support show command and others.
- Add Tests

Tests:
- Manually tested adding Rest based source and searching for packages returns a set of entries from the Rest source.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/765)


src/AppInstallerRepositoryCore/Rest/HttpClientHelper.h
```diff
@@ -0,0 +1,34 @@
...
+		json::value HandleGet();
+
+	protected:
+		pplx::task<web::http::http_response> MakeRequest(web::http::http_request req);
+
+	private:
+		http_client m_client;
+		utility::string_t m_url;
+	};
+}
```
jsoref, line 34: if you could add missing newlines to these files so you don't have this indicator, that'd let people who use normal text editors later not be blamed for changing the last line in your files :-).

src/WinGetUtil/packages.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="cpprestsdk.v142" version="2.10.15" targetFramework="native" />
   <package id="Microsoft.Windows.CppWinRT" version="2.0.200729.8" targetFramework="native" />
   <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.200519.2" targetFramework="native" />
 </packages>
```
jsoref, line 6: The same thing applies to a number (but not all) of the files here.

Tangentially related, the spelling bot is similarly upset about the patterns file below (I can't comment on that file because it isn't part of the PR), although for some reason (I need to investigate) it can't seem to properly count to the last line of the file...

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+            {
+                UNREFERENCED_PARAMETER(property);
+                std::vector<Utility::LocIndString> result;
+                return result;
+            }
+
+            // TODO
+            Manifest::Manifest GetManifest() const override
+            {
+                AICLI_LOG(Repo, Info, << "Downloading manifest");
```
JohnMcPMS, line 72: Verbose

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+                // TODO: Create Manifest object from result using JSON parser.
+                THROW_HR(ERROR_CALL_NOT_IMPLEMENTED);
+            }
+
+            std::shared_ptr<const ISource> GetSource() const override
+            {
+                return GetReferenceSource();
+            }
+
+            // TODO
```
JohnMcPMS, line 85: Metadata is currently only used for storing extra information about installed packages, so you likely don't need to do any more than this until we actually have some reason.

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+                std::shared_ptr<PackageVersion> packageVersion;
+                if (!versionKey.Version.empty() && !versionKey.Channel.empty())
+                {
+                    for (const auto& versionInfo : m_package.versions)
+                    {
+                        if (CaseInsensitiveEquals(versionInfo.GetVersion().ToString(), versionKey.Version)
+                            && CaseInsensitiveEquals(versionInfo.GetChannel().ToString(), versionKey.Channel))
+                        {
+                            packageVersion = std::make_shared<PackageVersion>(source, m_package.packageInfo, versionInfo);
```
JohnMcPMS, line 179: break afterward

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+                {
+                    return GetLatestAvailableVersion();
+                }
+                else if (versionKey.Version.empty())
+                {
+                    for (const auto& versionInfo : m_package.versions)
+                    {
+                        if (CaseInsensitiveEquals(versionInfo.GetChannel().ToString(), versionKey.Channel))
+                        {
+                            packageVersion = std::make_shared<PackageVersion>(source, m_package.packageInfo, versionInfo);
```
JohnMcPMS, line 193: If the versions are already sorted, this will work.  If not, this just gets the first version for a channel in the list.

Also, break after

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+                std::vector<Utility::LocIndString> result;
+                return result;
+            }
+
+            // TODO
+            Manifest::Manifest GetManifest() const override
+            {
+                AICLI_LOG(Repo, Info, << "Downloading manifest");
+                std::string manifest = GetReferenceSource()->GetRestClient().GetManifestByVersion(
+                    m_packageInfo.packageIdentifier, m_versionInfo.GetVersion().ToString()).value();
```
JohnMcPMS, line 74: Needs to pass the channel along as well.

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+                        }
+                    }
+                }
+                else if (versionKey.Channel.empty())
+                {
+                    for (const auto& versionInfo : m_package.versions)
+                    {
+                        if (CaseInsensitiveEquals(versionInfo.GetVersion().ToString(), versionKey.Version))
+                        {
+                            packageVersion = std::make_shared<PackageVersion>(source, m_package.packageInfo, versionInfo);
```
JohnMcPMS, line 203: Same as above

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+        std::shared_ptr<const RestSource> sharedThis = shared_from_this();
+        for (auto& result : results.Matches)
+        {
+            std::unique_ptr<IPackage> package = std::make_unique<AvailablePackage>(sharedThis, result);
+            PackageMatchFilter packageFilter({}, {}, {});
+
+            // Sort the versions
+            std::sort(result.versions.begin(), result.versions.end(),
+                [](const VersionAndChannel& a, const VersionAndChannel& b)
```
JohnMcPMS, line 252: This is the default < sort and shouldn't need to be specified.

JohnMcPMS: Wait, I see that you are only sorting based on version.  But why?

JohnMcPMS: Also, this sort is either useless due to a copy inside AvailablePackage or indicative of a lifetime bug there because it wasn't copied.

I would suggest just moving the sort inside of the object that is going to keep the copy (looks like PackageBase).

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+	namespace
+	{
+		// The base class for data that comes from a rest based source.
+		struct RestSourceFactoryBase : public ISourceFactory
+		{
+			std::shared_ptr<ISource> Create(const SourceDetails& details, IProgressCallback& progress) override final
+			{
+				THROW_HR_IF(E_INVALIDARG, details.Type != RestSourceFactory::Type());
+				return CreateInternal(details, progress);
```
JohnMcPMS, line 22: No need to keep the pattern since there isn't actually a split based on the runtime here.  Move CreateInternal insides here.

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.cpp
```diff
@@ -0,0 +1,80 @@
...
+					details.Type = RestSourceFactory::Type();
+					AICLI_LOG(Repo, Info, << "Initializing source type: " << details.Name << " => " << details.Type);
+				}
+				else
+				{
+					THROW_HR_IF(E_INVALIDARG, details.Type != RestSourceFactory::Type());
+				}
+
+				// Check if URL is remote and secure
+				THROW_HR_IF(APPINSTALLER_CLI_ERROR_SOURCE_NOT_SECURE, Utility::IsUrlRemote(details.Arg) && !Utility::IsUrlSecure(details.Arg));
```
JohnMcPMS, line 43: Should probably require "remote" and secure in this instance.  The other source allows for local files to be served for the index and manifests.  Unless you plan to support that here, it should just be two lines, with the first requiring remote and the second secure (different errors).

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.cpp
```diff
@@ -0,0 +1,80 @@
...
+				AICLI_LOG(Repo, Info, << "Initializing source from: " << details.Name << " => " << details.Arg);
```
JohnMcPMS, line 44: This log line doesn't make sense when copied.

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.h
```diff
@@ -0,0 +1,27 @@
...
+    // A source where the information is stored on a REST based server.
+    // In addition, the manifest information is also available on the server.
+    // Arg  ::  Expected to be a API which supports querying functionality.
+    struct RestSourceFactory
+    {
+        // Get the type string for this source.
+        static constexpr std::string_view Type()
+        {
+            using namespace std::string_view_literals;
+            return "ThirdParty.RestSource"sv;
```
JohnMcPMS, line 21: I don't think we should call it "ThirdParty", as Microsoft will certainly implement and use these sources.

I would prefer `Microsoft.Rest` as the API definition is ours.

src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "pch.h"
+#include "Microsoft/Schema/Version.h"
+#include <AppInstallerVersions.h>
+#include <vector>
+
+using namespace AppInstaller::Utility;
```
JohnMcPMS, line 9: No using namespace in headers

src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h
```diff
@@ -0,0 +1,50 @@
...
+		{
+			std::vector<Package> Matches;
+			bool Truncated = false;
+		};
+
+		// Performs a search based on the given criteria.
+		virtual SearchResult Search(const std::string& restApiUri, const SearchRequest& request) const = 0;
+
+		// Gets the manifest for given version
+		virtual std::optional<std::string> GetManifestByVersion(const std::string& restApiUri, const std::string& packageId, const std::string& version) const = 0;
```
JohnMcPMS, line 48: This should return a Manifest so that the details about how to parse the result stays internal to the versioned interface.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,143 @@
...
+using namespace web;
+using namespace web::json;
+using namespace web::http;
+using namespace web::http::client;
+using namespace AppInstaller::Repository::Rest::Schema;
+
+namespace AppInstaller::Repository::Rest::Schema::V1_0
+{
+    // Endpoint constants
+    const std::string ManifestSearchPostEndpoint = "/api/manifestSearch?";
```
JohnMcPMS, line 19: Use `constexpr std::string_view` for holding string literals.  (or `constexpr std::wstring_view` for wide literals below)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,143 @@
...
+        utility::string_t GetSearchEndpoint(const std::string& restApiUri)
+        {
+            std::string fullSearchAPI = GetRestAPIBaseUri(restApiUri) + ManifestSearchPostEndpoint;
+            utility::string_t searchAPI = utility::conversions::to_string_t(fullSearchAPI);
+            return searchAPI;
+        }
+
+        utility::string_t GetManifestByVersionEndpoint(const std::string& restApiUri, const std::string& packageId, const std::string& version)
+        {
+            // TODO: Replace with manifest version endpoint
```
JohnMcPMS, line 70: TODONE?

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+        utility::string_t GetManifestByVersionEndpoint(const std::string& restApiUri, const std::string& packageId, const std::string& version)
+        {
+            // TODO: Replace with manifest version endpoint 
+            std::string versionEndpoint = GetRestAPIBaseUri(restApiUri) + "/api/packages/" + packageId + "/versions/" + version;
+            utility::string_t versionApi = utility::conversions::to_string_t(versionEndpoint);
+            return versionApi;
+        }
+
+        std::string GetStringFromJsonStringValue(const json::value& value)
```
JohnMcPMS, line 76: Return a `std::optional<std::string>` to preserve ability to model a missing value.

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,262 @@
...
+    SearchResult RestSource::Search(const SearchRequest& request) const
+    {
+        // Note: Basic search functionality to fetch everything.
+        RestClient::SearchResult results = m_restClient.Search(request);
+        SearchResult searchResult;
+
+        std::shared_ptr<const RestSource> sharedThis = shared_from_this();
+        for (auto& result : results.Matches)
+        {
+            std::unique_ptr<IPackage> package = std::make_unique<AvailablePackage>(sharedThis, result);
```
JohnMcPMS, line 247: Should `std::move` the result into the package since it will no longer be needed.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,143 @@
...
+            {
+                continue;
+            }
+
+            std::vector<VersionAndChannel> versionList;
+            for (auto& versionItem : versionValue.as_array())
+            {
+                std::string version = GetStringFromJsonStringValue(versionItem.at(Version));
+                std::string channel = GetStringFromJsonStringValue(versionItem.at(Channel));
+                versionList.emplace_back(VersionAndChannel(version, channel));
```
JohnMcPMS, line 118: std::move

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            PackageInfo packageInfo = PackageInfo(packageId, packageName, publisher);
```
JohnMcPMS, line 121: std::move

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    std::optional<std::string> Interface::GetManifestByVersion(const std::string& restApiUri, const std::string& packageId, const std::string& version) const
+    {
+        HttpClientHelper clientHelper(GetManifestByVersionEndpoint(restApiUri, packageId, version));
+        json::value jsonObject = clientHelper.HandleGet();
+
+        if (jsonObject.is_null())
+        {
+            return std::string();
```
JohnMcPMS, line 136: In this case you would want to `return {};` to return an empty optional, and probably want to preserve that when you change the function to return `std::optional<Manifest>`.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "pch.h"
+#include "cpprest/http_client.h"
+#include "cpprest/json.h"
```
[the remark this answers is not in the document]
yao-msft: I would love to keep only one json implementation but Valijson does not have adaptors for cpprestsdk's implementation of json. So we have to keep jsoncpp lib for now.

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
ashpatil-msft: you are right, fixed it to  sort by version and channel

---
In reply to: [583295822](https://github.com/microsoft/winget-cli/pull/765#discussion_r583295822) [](ancestors = 583295822)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
ashpatil-msft: changed all to use string_view

---
In reply to: [583302491](https://github.com/microsoft/winget-cli/pull/765#discussion_r583302491) [](ancestors = 583302491)

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.h
```diff
@@ -0,0 +1,34 @@
...
+using namespace web::http;
+using namespace web::http::client;
+
+namespace AppInstaller::Repository::Rest
+{
+	struct HttpClientHelper
+	{
+		HttpClientHelper(utility::string_t url);
+
+		pplx::task<web::http::http_response> Post(const json::value& body);
```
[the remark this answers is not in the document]
ashpatil-msft: the http_client::request method called in MakeRequest is non-const, and since all the methods call MakeRequest, we cant make these methods const. Please correct me if wrong.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,153 @@
...
+    constexpr std::string_view Data = "data"sv;
+
+    // Search body constants
+    constexpr std::string_view FetchAllManifests = "fetchAllManifests"sv;
+
+    // Search response constants
+    constexpr std::string_view PackageIdentifier = "PackageIdentifier"sv;
+    constexpr std::string_view PackageName = "PackageName"sv;
+    constexpr std::string_view Publisher = "Publisher"sv;
+    constexpr std::string_view PackageFamilName = "PackageFamilyName"sv;
```
jsoref, line 28:
```suggestion
    constexpr std::string_view PackageFamilyName = "PackageFamilyName"sv;
```

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        auto& dataArray = jsonObject.at(GetJsonKeyNameString(Data)).as_array();
+
+        for (auto& manifestItem : dataArray)
+        {
+            std::optional<std::string> packageId = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageIdentifier)));
+            std::optional<std::string> packageName = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageName)));
+            std::optional<std::string> publisher = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(Publisher)));
+            std::optional<std::string> packageFamilyName = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageFamilName)));
```
jsoref, line 111:
```suggestion
            std::optional<std::string> packageFamilyName = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageFamilyName)));
```

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -6,32 +6,27 @@
 #include "cpprest/json.h"
 #include "HttpClientHelper.h"
 
-using namespace web;
```
JohnMcPMS: You can leave using namespace in the cpp files, it is the headers where it can lead to problems (well, it can lead to problems in the cpp but they will be contained to the cpp).

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.cpp
```diff
@@ -18,11 +18,15 @@ namespace AppInstaller::Repository::Rest
 		{
 			std::shared_ptr<ISource> Create(const SourceDetails& details, IProgressCallback& progress) override final
 			{
+				UNREFERENCED_PARAMETER(progress);
```
JohnMcPMS, line 21: If you are not going to use it ever, just remove the name from the function signature.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -60,22 +62,27 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
 
         utility::string_t GetSearchEndpoint(const std::string& restApiUri)
         {
-            std::string fullSearchAPI = GetRestAPIBaseUri(restApiUri) + ManifestSearchPostEndpoint;
+            std::string fullSearchAPI = GetRestAPIBaseUri(restApiUri) + ManifestSearchPostEndpoint.data();
```
JohnMcPMS, line 65: Using `.data()` and not passing to an external C API usually means there is a more efficient way to leverage a `string_view`.  For instance, you can use `append` instead if there is no `operator+` overload to do what you want.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -60,22 +62,27 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
...
             return searchAPI;
         }
 
-        utility::string_t GetManifestByVersionEndpoint(const std::string& restApiUri, const std::string& packageId, const std::string& version)
+        utility::string_t GetManifestByVersionEndpoint(
+            const std::string& restApiUri, const std::string& packageId, const std::string& version, const std::string& channel)
         {
-            // TODO: Replace with manifest version endpoint 
-            std::string versionEndpoint = GetRestAPIBaseUri(restApiUri) + "/api/packages/" + packageId + "/versions/" + version;
+            std::string versionEndpoint = GetRestAPIBaseUri(restApiUri) + ManifestByVersionAndChannelGetEndpoint.data() + packageId + "?version=" + version + "&channel=" + channel;
```
JohnMcPMS, line 73: I think you followed my model too closely with the SQLite interface.  I separated the lifetime of the connection from the interface used to access it for arguable reasons.  But you have no such concern here.  I think the only concern you have is that `Create` on the factory might want to determine the interface version to use (and at the same time will ensure that the endpoint exists), but there is no (complex) lifetime to need to manage.

I would suggest making the interface less static, taking in some data on construction and storing what it needs to operate with rather than requiring every call into to pass in the data repeatedly.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -86,58 +93,61 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
...
+            std::optional<std::string> packageId = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageIdentifier)));
+            std::optional<std::string> packageName = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageName)));
+            std::optional<std::string> publisher = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(Publisher)));
+            std::optional<std::string> packageFamilyName = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageFamilyName)));
+            std::optional<std::string> productCode = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(ProductCode)));
+            web::json::value versionValue = manifestItem.at(GetJsonKeyNameString(Versions));
+
+            if (!packageId.has_value() || !packageName.has_value() || !publisher.has_value() || versionValue.is_null() || versionValue.as_array().size() == 0)
             {
                 continue;
```
JohnMcPMS, line 117: It seems like at least a verbose log line is in order to indicate an error in the incoming data.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -86,58 +93,61 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
...
+            std::vector<AppInstaller::Utility::VersionAndChannel> versionList;
             for (auto& versionItem : versionValue.as_array())
             {
-                std::string version = GetStringFromJsonStringValue(versionItem.at(Version));
-                std::string channel = GetStringFromJsonStringValue(versionItem.at(Channel));
-                versionList.emplace_back(VersionAndChannel(version, channel));
+                std::optional<std::string> version = GetStringFromJsonStringValue(versionItem.at(GetJsonKeyNameString(Version)));
+                std::optional<std::string> channel = GetStringFromJsonStringValue(versionItem.at(GetJsonKeyNameString(Channel)));
+
+                versionList.emplace_back(AppInstaller::Utility::VersionAndChannel(std::move(version.value()), std::move(channel.value())));
```
JohnMcPMS, line 126: Check that the values exist before attempting to get them here.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT License.
+#pragma once
+#include "pch.h"
+#include "cpprest/http_client.h"
+#include "cpprest/json.h"
+#include "HttpClientHelper.h"
+
+namespace AppInstaller::Repository::Rest
+{
+	HttpClientHelper::HttpClientHelper(utility::string_t url) : m_client(url), m_url(url) {}
```
JohnMcPMS, line 11: >	 [](start = 0, length = 1)

Another reason to use CF, easier to see tabs.  Please convert to 4 spaces.

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
ashpatil-msft: Yeah, I misread that comment initially, by the time I read it again, I had removed the namespaces here. I will use them in cpp  files in the future.

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
ashpatil-msft: I have modified my settings to use spaces, but still see some tabs. Have replaced them, let me know if there are more. Btw, Do you have a plugin for codeflow that detects tabs? 

---
In reply to: [586819753](https://github.com/microsoft/winget-cli/pull/765#discussion_r586819753) [](ancestors = 586819753)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
ashpatil-msft: Gotcha, that makes sense. Modified to have Rest API uri to avoid sending it repeatedly. 

---
In reply to: [586811608](https://github.com/microsoft/winget-cli/pull/765#discussion_r586811608) [](ancestors = 586811608)

src/AppInstallerCommonCore/Errors.cpp
```diff
@@ -67,6 +67,8 @@ namespace AppInstaller
                 return "Command requires administrator privileges to run";
             case APPINSTALLER_CLI_ERROR_SOURCE_NOT_SECURE:
                 return "The source location is not secure";
+            case APPINSTALLER_CLI_ERROR_SOURCE_NOT_REMOTE:
+                return "The source location is not remote";
```
yao-msft, line 71: Nit: the ordering of these follows order of declarations in AppInstallerErrors.h
Also by the way, can you help add  APPINSTALLER_CLI_ERROR_JSON_INVALID_FILE  which we somehow failed previously?

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -56,6 +58,8 @@ namespace AppInstaller::Settings
             return ExperimentalFeature{ "Import Command", "import", "https://aka.ms/winget-settings", Feature::ExperimentalImport };
         case Feature::ExperimentalExport:
             return ExperimentalFeature{ "Export Command", "export", "https://aka.ms/winget-settings", Feature::ExperimentalExport };
+        case Feature::ExperimentalRestSource:
+            return ExperimentalFeature{ "Rest Source Support", "restSource", "https://aka.ms/winget-settings", Feature::ExperimentalRestSource };
```
yao-msft, line 62: >https://aka.ms/winget-settings [](start = 78, length = 30)

Please remember to actually update the settings.md

yao-msft: and the settings.schema.json

---
In reply to: [594659756](https://github.com/microsoft/winget-cli/pull/765#discussion_r594659756) [](ancestors = 594659756)

src/AppInstallerCommonCore/Errors.cpp
ashpatil-msft: Add the error in order and also added APPINSTALLER_CLI_ERROR_JSON_INVALID_FILE.

src/AppInstallerCommonCore/ExperimentalFeature.cpp
ashpatil-msft: Added to both settings.md and settings.schema.json.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -327,6 +341,11 @@ namespace AppInstaller::Repository
             {
                 return Microsoft::PredefinedInstalledSourceFactory::Create();
             }
+            else if (Utility::CaseInsensitiveEquals(Rest::RestSourceFactory::Type(), type)
```
yao-msft, line 344: >Utility::CaseInsensitiveEquals(Rest::RestSourceFactory::Type(), type) [](start = 21, length = 69)

From the comment above "Should always come from code, so no need for case insensitivity"
Shall we just use '==' for source type comparisons? Here and line 230

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.cpp
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
```
yao-msft, line 3: not needed

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
```
yao-msft, line 3: not needed

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "HttpClientHelper.h"
+
+namespace AppInstaller::Repository::Rest
+{
+    HttpClientHelper::HttpClientHelper(utility::string_t url) : m_client(url), m_url(url) {}
+
+    pplx::task<web::http::http_response> HttpClientHelper::Post(const web::json::value& body)
+    {
+        AICLI_LOG(Repo, Verbose, << "Sending http POST request to: " << utility::conversions::to_utf8string(m_url));
+        web::http::http_request request(web::http::methods::POST);
```
yao-msft, line 16: >( [](start = 39, length = 1)

nit: we use {} for constructors across this project

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,67 @@
...
+                }
+            }).wait();
+
+        return result;
+    }
+
+    pplx::task<web::http::http_response> HttpClientHelper::Get()
+    {
+        AICLI_LOG(Repo, Verbose, << "Sending http GET request to: " << utility::conversions::to_utf8string(m_url));
+        web::http::http_request request(web::http::methods::GET);
```
yao-msft, line 41: >( [](start = 39, length = 1)

{} for constructors

src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h
```diff
@@ -0,0 +1,48 @@
...
+    // The common interface used to interact with RestAPI responses.
+    struct IRestClient
+    {
+    virtual ~IRestClient() = default;
+
+    struct PackageInfo
+    {
+        std::string packageIdentifier;
+        std::string packageName;
+        std::string publisher;
```
yao-msft, line 20: Pascal case for public accessible members. i.e. PackageIdentifier

src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h
```diff
@@ -0,0 +1,48 @@
...
+
+        PackageInfo(std::string packageIdentifier, std::string packageName, std::string publisher) 
+        : packageIdentifier(packageIdentifier), packageName(packageName), publisher(publisher) {}
+    };
+
+    // Minimal information retrieved for any search request.
+    struct Package
+    {
+        PackageInfo packageInfo;
+        std::vector<AppInstaller::Utility::VersionAndChannel> versions;
```
yao-msft, line 30: variable name PascalCase

doc/Settings.md
```diff
@@ -107,3 +107,13 @@ While work is in progress for import, the command is hidden behind a feature tog
        "import": true
    },
 ```
+
+### restSource
+
+While work is in progress for rest source support, the command is hidden behind a feature toggle. One can enable it as below:
```
JohnMcPMS, line 113: >command [](start = 55, length = 7)

feature is hidden behind a toggle.

Might also want to point out that enabling this will not change how anything works, just allow for testing of the REST source.

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: That comment is because those are specifically internal sources.  A REST source may be configured externally, so yes, we should do case insensitive.

---
In reply to: [594688379](https://github.com/microsoft/winget-cli/pull/765#discussion_r594688379) [](ancestors = 594688379)

src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h
```diff
@@ -0,0 +1,48 @@
...
+
+        Package(PackageInfo packageInfo, std::vector<AppInstaller::Utility::VersionAndChannel> versions)
+        : packageInfo(packageInfo), versions(versions) {}
+    };
+
+    struct SearchResult
```
yao-msft, line 36: >SearchResult [](start = 11, length = 12)

Is there a reason we cannot use SearchResult definition in Repository::SearchResult?

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.h
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+#include "pch.h"
+#include "cpprest/http_client.h"
+#include "cpprest/json.h"
+
+namespace AppInstaller::Repository::Rest
+{
+    struct HttpClientHelper
+    {
+        HttpClientHelper(utility::string_t url);
```
yao-msft, line 12: >utility::string_t url [](start = 25, length = 21)

const &

src/AppInstallerRepositoryCore/Rest/RestClient.cpp
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+#include "pch.h"
+#include "RestClient.h"
+#include "Rest/Schema/1_0/Interface.h"
+
+using namespace AppInstaller::Repository::Rest::Schema;
+
+namespace AppInstaller::Repository::Rest
+{
+    RestClient::RestClient(std::string restApi)
```
yao-msft, line 11: >std::string restApi [](start = 27, length = 19)

I think const & here
I followed through the constructors code, I did not see a big benefit for using copy here then std::move below

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.h
```diff
@@ -0,0 +1,28 @@
...
+#include "Rest/Schema/IRestClient.h"
+#include "cpprest/json.h"
+#include <vector>
+
+namespace AppInstaller::Repository::Rest::Schema::V1_0
+{
+    // Interface to this schema version exposed through IRestClient.
+    struct Interface : public IRestClient
+    {
+        Interface(std::string restApi);
```
yao-msft, line 13: >std::string [](start = 18, length = 11)

const &

src/AppInstallerRepositoryCore/RepositorySource.cpp
ashpatil-msft: So, we need case insensitivity check because the Rest source type doesnt come from code. I also realized that I missed adding case-insensitivity check in the rest factory. I will add it there as well.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,202 @@
...
+            // Add Channel Query param
+            if (!version.empty())
+            {
+                versionEndpoint.append("&channel=").append(channel);
+            }
+
+            return utility::conversions::to_string_t(versionEndpoint);
+        }
+
+        bool IsStringWhitespace(const std::string& value)
```
yao-msft, line 89: >bool IsStringWhitespace(const std::string& value) [](start = 8, length = 49)

I think we have this utility in AppInstallerStrings.hpp. If not, it's a good idea to move the utility there

src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h
ashpatil-msft: Yes, the sqlite index returns different type of SearchResult (with row ids and such). So we have one specific for Rest source result format.

---
In reply to: [594707007](https://github.com/microsoft/winget-cli/pull/765#discussion_r594707007) [](ancestors = 594707007)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
ashpatil-msft: My bad, I forgot to delete this. I am already using Utiltity::IsStringEmptyOrWhitespace 

---
In reply to: [594712782](https://github.com/microsoft/winget-cli/pull/765#discussion_r594712782) [](ancestors = 594712782)

yao-msft: Ah, I saw you already  used Utility::IsEmptyOrWhitespace(result) below. Then we can delete this 

---
In reply to: [594712782](https://github.com/microsoft/winget-cli/pull/765#discussion_r594712782) [](ancestors = 594712782)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,202 @@
...
+            if (Utility::IsEmptyOrWhitespace(result))
+            {
+                return {};
+            }
+
+            return result;
+        }
+    }
+
+    Interface::Interface(std::string restApi)
```
yao-msft, line 120: >std::string [](start = 25, length = 11)

const & no need for move below

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,202 @@
...
+         m_searchEndpoint = GetSearchEndpoint(m_restApiUri);
+    }
+
+    IRestClient::SearchResult Interface::Search(const SearchRequest& request) const
+    {
+        UNREFERENCED_PARAMETER(request);
+        SearchResult result;
+
+        // TODO: Handle continuation token.
+        HttpClientHelper clientHelper(m_searchEndpoint);
```
yao-msft, line 132: >( [](start = 37, length = 1)

{}

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,202 @@
...
+            }
+
+            return result;
+        }
+    }
+
+    Interface::Interface(std::string restApi)
+    {
+         m_restApiUri = GetRestAPIBaseUri(std::move(restApi));
+         m_searchEndpoint = GetSearchEndpoint(m_restApiUri);
```
yao-msft, line 123: >m_searchEndpoint [](start = 9, length = 16)

Do we want to save a copy for GetManifestByVersionEndpoint too?

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    IRestClient::SearchResult Interface::Search(const SearchRequest& request) const
+    {
+        UNREFERENCED_PARAMETER(request);
```
yao-msft, line 128: > UNREFERENCED_PARAMETER(request); [](start = 7, length = 33)

not needed, used below

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,202 @@
...
+        if (jsonObject.is_null())
+        {
+            return result;
+        }
+
+        auto& dataArray = jsonObject.at(GetJsonKeyNameString(Data)).as_array();
+
+        for (auto& manifestItem : dataArray)
+        {
+            std::optional<std::string> packageId = GetStringFromJsonStringValue(manifestItem.at(GetJsonKeyNameString(PackageIdentifier)));
```
yao-msft, line 145: >packageId [](start = 39, length = 9)

Make sure to revist these fields as some of them are definitely required.
You can add a todo for now since there'll be more error checking, manifest integrity checkto be performed later and they could potentially be in another file.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,202 @@
...
+        web::json::value jsonObject = clientHelper.HandleGet();
+
+        if (jsonObject.is_null())
+        {
+            return {};
+        }
+
+        // Parse json and return Manifest
+        auto& manifestObject = jsonObject.at(GetJsonKeyNameString(Data));
+        UNREFERENCED_PARAMETER(manifestObject);
```
yao-msft, line 199: nit: you can just (void)jsonObject.at(GetJsonKeyNameString(Data)); So you don't need the UNREFERENCED_PARAMETER below

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
yao-msft: I saw you handled bad cases by skipping them, I think that's probably enough.

---
In reply to: [594722790](https://github.com/microsoft/winget-cli/pull/765#discussion_r594722790) [](ancestors = 594722790)

ashpatil-msft: Yes, I will be adding a json parser that will handle the checks in my next PR.

---
In reply to: [594726878](https://github.com/microsoft/winget-cli/pull/765#discussion_r594726878) [](ancestors = 594726878,594722790)

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+                // TODO: Change identifier if required.
+                return std::make_shared<RestSource>(details, details.Arg, std::move(restClient));
+            }
+
+            void Add(SourceDetails& details, IProgressCallback&) override final
+            {
+                if (details.Type.empty())
+                {
+                    // With more than one source implementation, we will probably need to probe first
```
yao-msft, line 33: >// With more than one source implementation, we will probably need to probe first [](start = 20, length = 81)

nit: is this comment copy pasted? Does not seem very relevant to what the code is doing.

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,272 @@
...
+                std::vector<Utility::LocIndString> result;
+                return result;
+            }
+
+            // TODO
+            Manifest::Manifest GetManifest() const override
+            {
+                AICLI_LOG(Repo, Verbose, << "Downloading manifest");
+                Manifest::Manifest manifest = GetReferenceSource()->GetRestClient().GetManifestByVersion(
+                    m_packageInfo.packageIdentifier, m_versionInfo.GetVersion().ToString(), m_versionInfo.GetChannel().ToString()).value();
```
yao-msft, line 74: >.value() [](start = 130, length = 8)

std::optional check it's set before calling Value()

ashpatil-msft: you are right, I will just receive it as std::optional. I will implementing and using this in my next PR.

---
In reply to: [594733374](https://github.com/microsoft/winget-cli/pull/765#discussion_r594733374) [](ancestors = 594733374)

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,272 @@
...
+                            && CaseInsensitiveEquals(versionInfo.GetChannel().ToString(), versionKey.Channel))
+                        {
+                            packageVersion = std::make_shared<PackageVersion>(source, m_package.packageInfo, versionInfo);
+                            break;
+                        }
+                    }
+                }
+                else if (versionKey.Version.empty() && versionKey.Channel.empty())
+                {
+                    return GetLatestAvailableVersion();
```
yao-msft, line 188: >return GetLatestAvailableVersion(); [](start = 20, length = 35)

nit packageVersion = GetLatestAvailableVersion(); for code consistency

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -0,0 +1,272 @@
...
+    {
+        // Note: Basic search functionality to fetch everything.
+        RestClient::SearchResult results = m_restClient.Search(request);
+        SearchResult searchResult;
+
+        std::shared_ptr<const RestSource> sharedThis = shared_from_this();
+        for (auto& result : results.Matches)
+        {
+            std::unique_ptr<IPackage> package = std::make_unique<AvailablePackage>(sharedThis, std::move(result));
+            PackageMatchFilter packageFilter({}, {}, {});
```
yao-msft, line 265: >packageFilter({}, {}, {}); [](start = 31, length = 26)

Are we going to fill this in the future?

ashpatil-msft: I dont think so..

---
In reply to: [594736746](https://github.com/microsoft/winget-cli/pull/765#discussion_r594736746) [](ancestors = 594736746)

yao-msft: It's used in some of the sorting logic for getting the order of matches and display to user more relevant matches. We might want to add to todo for improve item in the future


---
In reply to: [594738139](https://github.com/microsoft/winget-cli/pull/765#discussion_r594738139) [](ancestors = 594738139,594736746)

yao-msft (approved):
:shipit:

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+            web::json::value json_body;
+            json_body[GetJsonKeyNameString(FetchAllManifests)] = web::json::value::string(L"true");
+
+            return json_body;
+        }
+
+        std::string GetRestAPIBaseUri(std::string restApiUri)
+        {
+            if (restApiUri.back() == '/')
```
JohnMcPMS, line 54: >back [](start = 27, length = 4)

Calling back on an empty string is UB (undefined behavior).

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,187 @@
...
+            versionEndpoint.append(ManifestByVersionAndChannelGetEndpoint).append(packageId);
+
+            // Add Version Query param
+            if (!version.empty())
+            {
+                versionEndpoint.append("?version=").append(version);
+            }
+
+            // Add Channel Query param
+            if (!version.empty())
```
JohnMcPMS, line 81: >version [](start = 17, length = 7)

`channel`.  Also, your query string will be bad if version is empty and channel is not (start with a `&` rather than a `?`).

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        std::optional<std::string> GetStringFromJsonStringValue(const web::json::value& value)
+        {
+            if (value.is_null() || !value.is_string())
+            {
+                return {};
+            }
+
+            std::string result = Utility::Trim(utility::conversions::to_utf8string(value.as_string()));
```
JohnMcPMS, line 96: >Trim [](start = 42, length = 4)

Why are you trimming the value?  Shouldn't it just return whatever is defined in the JSON?  Also, why are you treating and empty value as not present.  That doesn't allow for intentionally empty values.

I would suggest a method that does the basic json::value to string converion, then another one that applies these additional semantics with a name that implies it is doing so.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
JohnMcPMS: Actually given how `GetRestAPIBaseUri` is written, taking in a value and moving it to that function was correct.  I think it is likely not to matter who makes the copy, so any of the patterns could work in practice.  But the higher you can move the forced copy up the stack, the more efficient you can be usually.

---
In reply to: [594717708](https://github.com/microsoft/winget-cli/pull/765#discussion_r594717708) [](ancestors = 594717708)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
ashpatil-msft: your are right, I will correct it

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,187 @@
...
+         m_restApiUri = GetRestAPIBaseUri(restApi);
+         m_searchEndpoint = GetSearchEndpoint(m_restApiUri);
+    }
+
+    IRestClient::SearchResult Interface::Search(const SearchRequest& request) const
+    {
+        SearchResult result;
+
+        // TODO: Handle continuation token.
+        HttpClientHelper clientHelper{ m_searchEndpoint };
```
JohnMcPMS, line 118: >HttpClientHelper [](start = 8, length = 16)

Are these really so cheap that we should make one for every search?

src/AppInstallerRepositoryCore/Rest/RestClient.cpp
JohnMcPMS: The benefit of moving it higher in the stack is that the caller might have a value that they can move in.  If you can't take that value, that is one more string copy that is needed.

---
In reply to: [594708888](https://github.com/microsoft/winget-cli/pull/765#discussion_r594708888) [](ancestors = 594708888)

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    web::json::value HttpClientHelper::HandlePost(const web::json::value& body)
+    {
+        web::json::value result;
+        HttpClientHelper::Post(body).then([&result](const web::http::http_response& response)
+            {
+                AICLI_LOG(Repo, Verbose, << "Response status: " << response.status_code());
+
+                if (response.status_code() == web::http::status_codes::OK)
```
JohnMcPMS, line 28: >if (response.status_code() == web::http::status_codes::OK) [](start = 16, length = 58)

I think we should try to capture the HTTP status and throw it outside of the wait if not OK.

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    web::json::value HttpClientHelper::HandlePost(const web::json::value& body)
+    {
+        web::json::value result;
+        web::http::http_response httpResponse;
+        HttpClientHelper::Post(body).then([&httpResponse](const web::http::http_response& response)
+            {
+                AICLI_LOG(Repo, Verbose, << "Response status: " << response.status_code());
+                httpResponse = std::move(response);
```
JohnMcPMS, line 28: >httpResponse = std::move(response); [](start = 16, length = 35)

Shouldn't move this as it is not an rvalue (&&).  I would suggest just capturing the status_code.

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,77 @@
...
+            }).wait();
+
+        if (httpResponse.status_code() == web::http::status_codes::OK)
+        {
+            result = httpResponse.extract_json().get();
+        }
+        else
+        {
+            THROW_HR_MSG(E_UNEXPECTED, "Encountered an error while trying to reach server: %s", httpResponse.status_code());
```
JohnMcPMS, line 37: >E_UNEXPECTED [](start = 25, length = 12)

Use the HTTP_FACILITY to make the proper HRESULT.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -0,0 +1,219 @@
...
+            if (jsonObject.is_null())
+            {
+                return result;
+            }
+
+            auto& dataArray = jsonObject.at(GetJsonKeyNameString(Data)).as_array();
+
+            for (auto& manifestItem : dataArray)
+            {
+                try
```
JohnMcPMS, line 142: >try [](start = 16, length = 3)

Exceptions are not for control flow.  We should try to avoid them when we can write logic to better handle the situation.  Certainly we do not want to throw in the underlying function call just to catch it here.

JohnMcPMS (approved):
:shipit:

src/AppInstallerRepositoryCore/Rest/HttpClientHelper.cpp
```diff
@@ -0,0 +1,77 @@
...
+                httpResponse = response;
+            }).wait();
+
+        if (httpResponse.status_code() == web::http::status_codes::OK)
+        {
+            result = httpResponse.extract_json().get();
+        }
+        else
+        {
+            THROW_HR_IF(MAKE_HRESULT(SEVERITY_ERROR, FACILITY_HTTP, httpResponse.status_code()), httpResponse.status_code() != web::http::status_codes::OK);
```
yao-msft, line 37: >httpResponse.status_code() != web::http::status_codes::OK [](start = 97, length = 57)

looks like this is  always true since it's in the 'else'

AngelaOkpa (approved):

PR #767, opened 2021-02-24
Add normalized name and publisher to index and correlation telemetry

JohnMcPMS:
## Change
This change adds the normalized name and publisher to the index, as well as leveraging this to collect telemetry on how well this normalization strategy is doing to correlate a manifest with the ARP entry on the machine.

The normalized name and publisher tables are added to the index as 1:N entries due to the introduction of full localization of name and publisher into the new manifest schema.  These tables don't have any link between them, which would allow a combination of name and publisher from any localization to work.  This tradeoff of precision seems worthwhile over writing more complex code to handle two values at a time, and it also allows the potential for one of the fields to change while the other remains constant and consume less space overall.

Searching is done by sending in the observed values for name and publisher, and the normalization is done by the source before comparing with its internal store.  In this way any type of normalization can occur internal to the source without the workflow logic needing to be aware.

In order to collect telemetry, we first snapshot the set of packages in ARP before executing the install.  Afterward, we compare the set with the snapshot to determine which ones are net new.  We also search the source using the manifest used to install to determine if there are any matches.  Finally, we intersect these two sets of packages.

The telemetry contains information about the package that was being installed `{source, id, version, channel}`, information about the number of packages in each state above `{changed, matching, intersection}`, and information from the ARP entry (if any) that we have decided is the one that represents the main package `{name, version, publisher, language}`.  We will be able to leverage this telemetry to improve both our normalization code and the manifests in winget-pkgs to improve our ability to correlate installed packages with remote ones.

For instance, installing Notepad++ show the following:
```
[CLI ] During package install, 1 changes to ARP were observed, 0 matches were found for the package, and 0 packages were in both
[CLI ] The entry determined to be associated with the package is 'Notepad++ (64-bit x64)', with publisher 'Notepad++ Team'
```
Notepad++ does not correlate with the manifest because its publisher does not match (note the lack of “Team” in the manifest):
```
PS> wingetdev show Notepad++.Notepad++
Found Notepad++ [Notepad++.Notepad++]
Version: 7.9.3
Publisher: Notepad++
```

## Validation
New tests are added for the index changes and to verify behavior is consistent with expectations with the ARP reporting.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/767)


src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -214,6 +215,13 @@ namespace AppInstaller::CLI::Execution
             using value_t = std::vector<std::shared_ptr<Repository::ISource>>;
         };
 
+        template <>
+        struct DataMapping<Data::ARPSnapshot>
+        {
+            // Contains the { Id, Version, Channel }
```
yao-msft, line 221: >{ Id, Version, Channel } [](start = 28, length = 24)

Do we want a simple struct for this?

JohnMcPMS: I would have, but it needs to be a concrete type at this point and I was being lazy.  If we are going to do that I think the best choice would be to make a separate header just for defining these Execution::Data types to not crowd this header.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -8,6 +8,9 @@
 #include "MSStoreInstallerHandler.h"
 #include "WorkflowBase.h"
 
+#include <winget/NameNormalization.h>
+#include <winget/Manifest.h>
```
yao-msft, line 12: These should be in pch.h?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -365,9 +387,12 @@ namespace AppInstaller::CLI::Workflow
             Workflow::ShowInstallationDisclaimer <<
             Workflow::ReportExecutionStage(ExecutionStage::Download) <<
             Workflow::DownloadInstaller <<
+            Workflow::ReportExecutionStage(ExecutionStage::PreExecution) <<
+            Workflow::SnapshotARPEntries <<
```
yao-msft, line 391: Do we want to report these for upgrade scenarios?

src/AppInstallerCLITests/CompositeSource.cpp
```diff
@@ -561,3 +557,18 @@ TEST_CASE("CompositeSource_MultipleAvailableSources_ReverseMatchBoth", "[Composi
     REQUIRE(result.Matches[0].Package->GetInstalledVersion());
     REQUIRE(result.Matches[0].Package->GetAvailableVersionKeys().size() == 1);
 }
+
+TEST_CASE("CompositeSource_IsSame", "[CompositeSource]")
```
yao-msft, line 561: >"CompositeSource_IsSame" [](start = 10, length = 24)

Curious what this test actually do, are we concerned that calling search twice will have different results?

src/AppInstallerCommonCore/Public/winget/ManifestLocalization.h
```diff
@@ -70,7 +70,7 @@ namespace AppInstaller::Manifest
         }
 
         // Return a value indicating whether the given localization type exists.
-        bool Contains(Localization l) { return (m_data.find(l) != m_data.end()); }
+        bool Contains(Localization l) const { return (m_data.find(l) != m_data.end()); }
```
yao-msft, line 73: >const  [](start = 38, length = 6)

I copied this from ExecustionContext, we should update there as well

src/AppInstallerCLITests/TestSource.cpp
```diff
@@ -181,6 +183,36 @@ namespace TestCommon
         return false;
     }
 
+    bool TestPackage::IsSame(const IPackage* other) const
+    {
+        const TestPackage* otherAvailable = dynamic_cast<const TestPackage*>(other);
+
+        if (otherAvailable)
```
yao-msft, line 190: >if (otherAvailable) [](start = 8, length = 19)

nit: maybe return false early if null so we have a clearer flow(avoiding return true between return falses)

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
```diff
@@ -42,23 +42,43 @@ namespace AppInstaller::Repository
...
         ProductCode,
+        NormalizedNameAndPublisher,
     };
 
     // A single match to be performed during a search.
     struct RequestMatch
     {
         MatchType Type;
         Utility::NormalizedString Value;
+        std::optional<Utility::NormalizedString> Second;
```
yao-msft, line 53: >Second [](start = 49, length = 6)

nit: Second is not intuitive to me(as another match value) on first sight

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
```diff
@@ -122,6 +142,10 @@ namespace AppInstaller::Repository
         StandardUninstallCommand,
         // An uninstall command that should be non-interactive
         SilentUninstallCommand,
+        // The publisher of the package
+        Publisher,
+        // The language of the package
+        Language,
```
yao-msft, line 148: >Language [](start = 8, length = 8)

Locale to be consistent with other places in the project

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
```diff
@@ -42,23 +42,43 @@ namespace AppInstaller::Repository
...
 
-        PackageMatchFilter(PackageMatchField f, MatchType t, std::string_view v) : RequestMatch(t, v), Field(f) {}
+        PackageMatchFilter(PackageMatchField f, MatchType t) : RequestMatch(t), Field(f) { CheckValuesCount(); }
+        PackageMatchFilter(PackageMatchField f, MatchType t, Utility::NormalizedString& v) : RequestMatch(t, v), Field(f) { CheckValuesCount(); }
+        PackageMatchFilter(PackageMatchField f, MatchType t, const Utility::NormalizedString& v) : RequestMatch(t, v), Field(f) { CheckValuesCount(); }
+        PackageMatchFilter(PackageMatchField f, MatchType t, Utility::NormalizedString&& v) : RequestMatch(t, std::move(v)), Field(f) { CheckValuesCount(); }
+        PackageMatchFilter(PackageMatchField f, MatchType t, std::string_view v1, std::string_view v2) : RequestMatch(t, v1, v2), Field(f) { CheckValuesCount(); }
+
+    protected:
+        void CheckValuesCount()
```
yao-msft, line 74: >CheckValuesCount [](start = 13, length = 16)

EnsureRequiredValues?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_2/Interface.h
```diff
@@ -0,0 +1,34 @@
...
+
+        // Version 1.0
+        Schema::Version GetVersion() const override;
+        void CreateTables(SQLite::Connection& connection) override;
+        SQLite::rowid_t AddManifest(SQLite::Connection& connection, const Manifest::Manifest& manifest, const std::filesystem::path& relativePath) override;
+        std::pair<bool, SQLite::rowid_t> UpdateManifest(SQLite::Connection& connection, const Manifest::Manifest& manifest, const std::filesystem::path& relativePath) override;
+        SQLite::rowid_t RemoveManifest(SQLite::Connection& connection, const Manifest::Manifest& manifest, const std::filesystem::path& relativePath) override;
+        bool CheckConsistency(const SQLite::Connection& connection, bool log) const override;
+
+        // Version 1.1
```
yao-msft, line 23: >1.1 [](start = 19, length = 3)

1.2?

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
jsoref: Additional?

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
```diff
@@ -288,6 +302,14 @@ namespace AppInstaller::Repository::Microsoft
             //       is from it's ARP location, despite it very clearly being a specific architecture. And note that user
             //       scope does not have separate ARP locations, so every architecture would appear as native.
 
+            // Publisher is needed for certain scenarios but we don't store it from the manifest
```
yao-msft, line 305: >don't store it [](start = 64, length = 14)

do need to store?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -186,6 +218,18 @@ namespace AppInstaller::Repository
                 // Lie here so that list and upgrade will carry on to be able to output the diagnostic information.
                 return true;
             }
+
+            bool IsSame(const IPackage* other) const override
```
yao-msft, line 222: >IsSame [](start = 17, length = 6)

Should UnknownPackage != UnknownPackage? Or there could only be one UnknownPackage?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -399,4 +424,213 @@ namespace AppInstaller::CLI::Workflow
...
+                }
+            }
+
+            // Also attempt to find the entry based on the manifest data
+            const auto& manifest = context.Get<Execution::Data::Manifest>();
+
+            SearchRequest nameAndPublisherRequest;
+
+            // The default localization must contain the name or we cannot do this lookup
+            if (manifest.DefaultLocalization.Contains(Localization::PackageName))
```
yao-msft, line 495: >manifest.DefaultLocalization.Contains(Localization::PackageName) [](start = 16, length = 64)

PackageName and Publisher is required in default locale in schema

yao-msft (approved):
:shipit:

jsoref:
Fwiw, I'd suggest adding `\bE2EPackaged\b` on its own line to `patterns.txt` instead of adding the suggested item to `expect.txt`.

src/AppInstallerCLICore/ExecutionContext.h
JohnMcPMS: Also, if I were to do so I would need to implement several operators that tuple does for me.  I just don't see the value in that to get named fields.

However, I did do the work to make a new header and move the data part there so that a future item could be added.  I also moved the variant code to language utilities.  Eventually the uses of it that you added for manifests should be moved over to the common code.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: Without actual measurement, I have come to the conclusion that we should only put external code in the pch by default.  Internal code ends up causing a lot of build churn when something changes.  So for something that is used in almost every file, that should go in the pch.  But for something like NameNormalization, it should stay in the specific cpp files that use it.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: I have added it to upgrade.  In doing so, I coalesced all of the disparate install flows into a single task that should hopefully be the only one used to do actual download and installation from now on.

src/AppInstallerCLITests/CompositeSource.cpp
JohnMcPMS: This test is to exercise the IsSame function of the CompositeSource packages.

src/AppInstallerCLITests/TestSource.cpp
JohnMcPMS: Update this and other cases to be much more terse.

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySearch.h
JohnMcPMS: Changed to Additional

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
JohnMcPMS: We only keep name in the index for searching.  Publisher isn't really needed there, but it needs to flow from ARP, so it goes in the metadata.

src/AppInstallerRepositoryCore/CompositeSource.cpp
JohnMcPMS: The goal is that two of the same searches should return results with IsSame being 1:1 correlated.  A null != null kind of behavior would make that fail.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: But the client code does not assume that manifests are 100% schema compliant.

PR #777, opened 2021-03-04
Add setting to allow control over telemetry event writing

### LINKED ISSUE #279: Opt-Out of Telemetry

I'd like to be able to opt-out of telemetry for the Windows Package Manager client.

A command to control (disable / enable) telemetry in the client.
The --info output would display the state of the telemetry setting.

If the OS settings are disabled, the client would not report telemetry regardless of the setting.
The setting would persist if the client was updated/upgraded

JohnMcPMS:
Fixes #279 

## Change
Adds a setting to allow control over telemetry event writing.
```json
"telemetry":
{
    "disable": true
}
```
When disabled, the ETW events will not be written at all.
## Validation
Manually ran ETW event log listener with setting in both states; noted lack of events when disabled.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/777)


doc/Settings.md
```diff
@@ -45,6 +45,22 @@ Color of the progress bar that WinGet displays when not specified by arguments.
 - retro
 - rainbow
 
+## Telemetry
+
+The `telemetry` settings control whether winget writes ETW events that may be sent to Microsoft on a default installation of Windows.
+
+Details on telemetry can be found [here](/README.md#datatelemetry), and our primary privacy statement can be found [here](/privacy.md).
```
jsoref, line 52: Please don't linkify `here`. See [F84: Failure of Success Criterion 2.4.9 due to using a non-specific link such as "click here" or "more" without a mechanism to change the link text to specific text.
](https://www.w3.org/TR/WCAG20-TECHS/F84.html)
```suggestion
See [details on telemetry](../README.md#datatelemetry), and our [primary privacy statement](/privacy.md).
```

(And can you add the missing newline character to `.github/actions/spelling/patterns.txt` while you're updating this PR?)

yao-msft:
>   "allOf": [ 

Just found out it's unnecessary use of allOf and all below ""additionalItems": true" are incorrect. "additionalItems" were for array types. We should revisit this when we do schema validation on settings file

---
Refers to: schemas/JSON/settings/settings.schema.0.2.json:85 in 717b4b6. [](commit_id = 717b4b699681210dc6d45b7cc3e3ca4367fcfddc, deletion_comment = False)

yao-msft:
>   "allOf": [ 

Looks like the schema itself was not quite correct..

---
In reply to: [790236688](https://github.com/microsoft/winget-cli/pull/777#issuecomment-790236688) [](ancestors = 790236688)

---
Refers to: schemas/JSON/settings/settings.schema.0.2.json:85 in 717b4b6. [](commit_id = 717b4b699681210dc6d45b7cc3e3ca4367fcfddc, deletion_comment = False)

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -10,11 +10,11 @@
 #include <variant>
 #include <vector>
 
+using namespace std::chrono_literals;
+using namespace std::string_view_literals;
+
 namespace AppInstaller::Settings
 {
```
yao-msft, line 17: why we move usings outside for this one specifically?

yao-msft (approved):
:shipit:

doc/Settings.md
JohnMcPMS: Thanks for pointing this out, and yes, I will add the newline.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
JohnMcPMS: Putting the using statements inside the namespace imports them into that namespace.  So `AppInstaller::Settings` would now have definitions for `sv`, `ms`, etc.

PR #778, opened 2021-03-04
Support custom installer success exit codes

yao-msft:
Also added tests for custom installer success codes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/778)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
@@ -193,6 +193,7 @@ namespace AppInstaller::CLI::Workflow
         context.Reporter.Info() << Resource::String::InstallFlowStartingPackageInstall << std::endl;
 
         const std::string& installerArgs = context.Get<Execution::Data::InstallerArgs>();
+        const auto& successCodes = context.Get<Execution::Data::Installer>()->InstallerSuccessCodes;
 
         auto installResult = context.Reporter.ExecuteWithProgress(
             std::bind(InvokeShellExecute,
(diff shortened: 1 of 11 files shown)


src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -205,7 +206,7 @@ namespace AppInstaller::CLI::Workflow
             context.Reporter.Warn() << "Installation abandoned" << std::endl;
             AICLI_TERMINATE_CONTEXT(E_ABORT);
         }
-        else if (installResult.value() != 0)
+        else if (installResult.value() != 0 && (std::find(successCodes.begin(), successCodes.end(), static_cast<int>(installResult.value())) == successCodes.end()))
```
JohnMcPMS, line 209: >installResult.value() != 0 [](start = 17, length = 26)

Should 0 be included by default if codes are provided?  What if 0 is a failure for some installer?  I can understand the confusion that this could create, so it might be better to have a separate flag in the manifest for making 0 a failure value.

JohnMcPMS, line 209: >int [](start = 112, length = 3)

Should we just store the values in the manifest as DWORD then, since we will always compare against this value type?

JohnMcPMS, line 209: >successCodes [](start = 58, length = 12)

I think we are making a problem for ourselves later if we simply add this support without a healthy set of defaults in place.  If we take this as is, this is the problem I imagine:
1. Installer always return some value stating "reboot required for full functionality"
2. Manifest author puts this value in success codes
3. We add defaults later that make this value and others success for installation
4. Installer starts returning one of the other values in some cases (maybe reboot required for any functionality)
5. We fail the install because it isn't one of the values in the list
Then do we need to start treating the values in the manifest as additional ones, beyond 0 and our defaults?  How could they indicate to us that our defaults are incorrect; that one of them means a failure?

If the manifest called these `AdditionalSuccessCodes`, we would probably be fine to treat 0 and any values we wanted (on a per-InstallerType basis) as success.  But if it calls them `InstallerSuccessCodes`, it seems like this is the list, full stop.

JohnMcPMS (approved):
:shipit:

yao-msft:
Thanks to @chausner for initiating the first version of the fix at #296

PR #779, opened 2021-03-04
Enable JSON schema validation for import

lechacon:
Enabling validation of JSON files for import according to the schema.

* Extracted common logic from existing manifest JSON validation.
* Added new shared items project for the schema resource.
* Added validation when reading the import file. The caller already handled the error case and needed no changes.
* Enabled existing tests for this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/779)


src/Schemas/Schemas.vcxitems
```diff
@@ -15,10 +15,11 @@
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="$(MSBuildThisFileDirectory)ManifestSchema.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)PackagesSchema.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)resource.h" />
   </ItemGroup>
   <ItemGroup>
-    <ResourceCompile Include="$(MSBuildThisFileDirectory)ManifestSchema.rc" />
+    <ResourceCompile Include="$(MSBuildThisFileDirectory)Schemas.rc" />
```
yao-msft, line 22: >Schemas.rc [](start = 57, length = 10)

Last time I chatted with John, we think we'll need a new vcxitems for import/export and other schemas that's specifically for AppInstallerCli.exe
We should leave the manifest one untouched because this is also used in wingetutil

lechacon: Restored ManifestSchema and added a separate vcxitems in latest commit.

src/AppInstallerCLI.sln
```diff
@@ -68,6 +66,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Valijson", "Valijson\Valijs
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ManifestSchema", "ManifestSchema\ManifestSchema.vcxitems", "{7D05F64D-CE5A-42AA-A2C1-E91458F061CF}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "AppInstallerCLISchemas", "AppInstallerCLISchemas\AppInstallerCLISchemas.vcxitems", "{952B513F-8A00-4D74-9271-925AFB3C6252}"
```
JohnMcPMS, line 69: >AppInstallerCLISchemas [](start = 79, length = 22)

We are trying to move away from AppInstallerCLI by at least not adding more instances of it.  Use WinGet instead.

src/AppInstallerCLICore/PackageCollection.cpp
```diff
@@ -157,8 +160,19 @@ namespace AppInstaller::CLI
...
-            // TODO: Embed schema in binaries & validate file. This will return nullopt on failure.
+            // Validate the JSON against the schema.
+            Json::Value schemaJson = JsonSchema::LoadResourceAsSchemaDoc(MAKEINTRESOURCE(IDX_PACKAGES_SCHEMA_V1), MAKEINTRESOURCE(PACKAGESSCHEMA_RESOURCE_TYPE));
+            valijson::Schema schema;
+            JsonSchema::PopulateSchema(schemaJson, schema);
+
+            valijson::ValidationResults results;
+            if (!JsonSchema::Validate(schema, root, results))
+            {
+                AICLI_LOG(CLI, Error, << JsonSchema::GetErrorStringFromResults(results));
```
JohnMcPMS, line 171: >GetErrorStringFromResults [](start = 53, length = 25)

We need to get this value to the user so that they can immediately see the issue.

lechacon: Would it be okay to show that even if it is not localized?

JohnMcPMS: I was going to add in my comment "And put one more thing on the list that needs localization."  But yes, I think even if everything else was localized, I would think it better to show a descriptive error in English (I assume) than simply an error code.  At least then the user can throw it at an automated translation and get some kind of understanding as to what is wrong (maybe).  And we can have our own localized preamble before the unlocalized part.

src/AppInstallerCLICore/PackageCollection.cpp
```diff
@@ -157,8 +160,19 @@ namespace AppInstaller::CLI
 
         std::optional<PackageCollection> TryParseJson(const Json::Value& root)
         {
-            // TODO: Embed schema in binaries & validate file. This will return nullopt on failure.
+            // Validate the JSON against the schema.
+            Json::Value schemaJson = JsonSchema::LoadResourceAsSchemaDoc(MAKEINTRESOURCE(IDX_PACKAGES_SCHEMA_V1), MAKEINTRESOURCE(PACKAGESSCHEMA_RESOURCE_TYPE));
```
JohnMcPMS, line 164: >LoadResourceAsSchemaDoc [](start = 49, length = 23)

How will we determine which schema to load in the future when there are more than one version?  That can probably be a problem for future developer, but let's consider if we are just putting off the work or if we are making that work harder.

For instance, will a file with no "schema" json item validate as good?  Do we want to support that and thus default to V1 schema if no value present?  Or would we prefer to require the "schema" to be present and that it match the value that we expect, regardless of version.

lechacon: In the future we could load the schema depending on the $schema property, and have some fall back if it is missing or not recognized (e.g. use latest schema).

This does not validate that the file has a $schema property or that it has a specific value. We may be able to require it in the schema if we wanted to.

If we create newer versions this function will accept them as long as they are backwards compatible. All the nodes accept additional unrecognized properties to allow room for extending without causing validation failures.

src/AppInstallerCLI.sln
lechacon: Renamed it to WinGetSchemas

src/AppInstallerCLICore/PackageCollection.cpp
JohnMcPMS: I think we should require the $schema property and that it contains the expected value, even if we don't actually use it to select the schema to use yet.  This seems like something that should be done by the code rather than in the schema file itself.

src/WinGetSchemas/resource.h
```diff
@@ -0,0 +1,14 @@
+//{{NO_DEPENDENCIES}}
+// Microsoft Visual C++ generated include file.
+// Used by Schemas.rc
```
yao-msft, line 3: >Schemas.rc [](start = 11, length = 10)

nit: WinGetSchemas.rc

JohnMcPMS (approved):
:shipit:

PR #801, opened 2021-03-17
Prototype draft work for refactoring winget to enable COM API exposure for winget install command

sachintaMSFT:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/801)

### DIFF
--- a/src/AppInstallerCLICore/Core.cpp
+++ b/src/AppInstallerCLICore/Core.cpp
@@ -1,9 +1,11 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 #include "pch.h"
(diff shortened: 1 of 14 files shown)


src/AppInstallerCLICore/Core.cpp
```diff
@@ -1,9 +1,11 @@
-﻿// Copyright (c) Microsoft Corporation.
```
JohnMcPMS, line 1: What is going on here, has your editor injected a BOM?

sachintaMSFT: No. I had authentication issues submitting the PR from Visual Studio, yesterday. So I tried updating each file on github web. Looks like I mistakenly updated this file from the diff rather than from the file itself. Sorry about that.

PR #809, opened 2021-03-18
Add ADMX definition for Group Policy

lechacon:
Adding definition file for a group policy. See #154 

Still pending:
* Setting up localization for the strings.
* Figuring out a representation for the lists of allowed and additional sources. Currently it is only a list of name/value pairs but we can't represent all three of name, arg & type.

Code to respect the policy will come as a separate PR

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/809)

Issue #154, opened 2020-05-15
Support for Group Policy Management


denelon:
As an IT professionalI want to be able to control the winget configuration for machines I manage policies for.

denelon:
Here is a list of policies we're considering:

**Disable Windows Package Manager**
This policy will prevent users from being able to use the Windows Package Manager. 

**Disable Settings Command**
This policy will prevent users from being able to change their settings.

**Disable Experimental Features**
This policy will prevent users from being able to enable experimental features in the Windows Package Manager.

**Disable Local Manifest Files**
This policy will prevent users from being able to install packages with local manifest files.

**Set Auto Update Interval In Minutes**
This policy will apply the auto update interval for package-based sources. Note the default setting is five minutes.

**Set Progress Bar**
This policy will apply a visual style to the progress bar. Note the options are: accent, rainbow, plain, and retro.

**Exclude Default Sources**
This policy will remove the default sources included with the Windows Package Manager. Enterprises may add these sources with the Include Enterprise Sources policy.

**Include Enterprise Sources**
This policy will add sources provided by the enterprise IT administrator. Keep in mind this policy can be used to add the default sources removed by the Exclude Default Sources policy.

**Disable Source Configuration Options**
This policy will prevent users from being able to add or remove sources other than those specified by the Exclude Default Sources and Include Enterprise Sources policies. Users will still be able to receive updates to any existing sources in their configurations.

sebawern:
I‘d rather (or also?) like to be able to configure this using Intune (even if that means using CSPs).

denelon:
We will have CSP as well.

jedieaston:
These look great! 

Two questions:

1. For "Enable Windows Package Manager Settings Command", does this block the user from editing the file in %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\settings.json by changing file system permissions, or does it just block the ` winget settings ` command?

2. For "Enable Windows Package Manager Allowed Sources", is there a way (currently or planned) for a user to get a list of the allowed sources?

denelon:
@jedieaston, thanks for taking a look. We're planning on "retaining" the settings such that if the policies are "removed", the user will default back to their settings file. It's just going to block the "winget -settings" command. The intent is just to reduce the confusion when there are settings conflicting with policies.

[...]

PR #810, opened 2021-03-18
Implement Group Policy

lechacon:
Implementing group policy. Interface is roughly based on ExperimentalFeatures and UserSettings.

This is still a draft. Still pending:
* E2E testing
* Better reporting and logging
* Use final registry keys/values from ADMX definition
* Implement policies for additional/allowed sources
* Showing policies enabled on winget --info

Related: #809 #154 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/810)


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -18,7 +18,7 @@ namespace AppInstaller::CLI
         case Args::Type::Query:
             return Argument{ "query", 'q', Args::Type::Query, Resource::String::QueryArgumentDescription, ArgumentType::Positional};
         case Args::Type::Manifest:
-            return Argument{ "manifest", 'm', Args::Type::Manifest, Resource::String::ManifestArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
+            return Argument{ "manifest", 'm', Args::Type::Manifest, Resource::String::ManifestArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help, Settings::TogglePolicy::LocalManifestFiles };
```
JohnMcPMS, line 21: >Settings::TogglePolicy::LocalManifestFiles [](start = 167, length = 42)

There you go making a case for these to be back in this shared location again.

If in the process of this work you put this on an argument that is constructed in a command, please move all instances here.  Any arguments with a feature flag should also be in here.

Basically, if the price of missing on is a policy is not properly enforced, then it should go through this method.  Please enforece that by making the constructor that takes the feature/policy private to force this path.  Hopefully that will make someone who is copying an existing argument do the right thing.

src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -80,6 +80,7 @@ namespace AppInstaller::CLI
 
     void InstallCommand::ValidateArgumentsInternal(Execution::Args& execArgs) const
     {
+        Command::ValidateArgumentsInternal(execArgs);
```
JohnMcPMS, line 83: >Command::ValidateArgumentsInternal(execArgs); [](start = 8, length = 45)

It seems far to easy for this to happen; we should move the existing contents of Command::ValidateArgumentsInternal into Command::ValidateArguments so that it will not be overridden on accident.

src/AppInstallerCLICore/Commands/RootCommand.cpp
```diff
@@ -66,6 +66,19 @@ namespace AppInstaller::CLI
         return "https://aka.ms/winget-command-help";
     }
 
+    void RootCommand::Execute(Execution::Context& context) const
```
JohnMcPMS, line 69: >void RootCommand::Execute(Execution::Context& context) const [](start = 4, length = 60)

We explicitly want to allow running `winget` when the whole thing is disabled?  Is that going to apply to other commands as well?

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -3,36 +3,51 @@
...
+    bool ExperimentalFeature::IsEnabled(Feature feature, const UserSettings& userSettings)
+    {
+        if (feature == Feature::None)
         {
-        case Feature::None:
             return true;
+        }
+
+        // TODO: How do we report that this was due to policy higher up?
+        if (!GroupPolicies().IsEnabledOrNotConfigured(TogglePolicy::ExperimentalFeatures))
```
JohnMcPMS, line 24: >if (!GroupPolicies().IsEnabledOrNotConfigured(TogglePolicy::ExperimentalFeatures)) [](start = 8, length = 82)

Can we have an `IsDisabled` method as well?  It will read much more clearly instead of "not is enabled or not configured".

src/AppInstallerCommonCore/Public/winget/ExperimentalFeature.h
```diff
@@ -48,6 +49,9 @@ namespace AppInstaller::Settings
         ExperimentalFeature& operator=(ExperimentalFeature&&) = default;
 
         static bool IsEnabled(Feature feature);
+        // For unit testing
```
JohnMcPMS, line 52: >// For unit testing [](start = 8, length = 19)

We have a macro that we wrap around these, if for no other reason than to track them.

```
#ifndef AICLI_DISABLE_TEST_HOOKS
```

src/AppInstallerCommonCore/Public/winget/GroupPolicy.h
```diff
@@ -0,0 +1,136 @@
...
+// Licensed under the MIT License.
+#pragma once
+
+#include "AppInstallerLanguageUtilities.h"
+#include "winget/Registry.h"
+#include <string_view>
+
+namespace AppInstaller::Settings
+{
+    using namespace std::string_view_literals;
```
JohnMcPMS, line 11: >using namespace std::string_view_literals; [](start = 4, length = 42)

Outside of the namespace.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -81,10 +83,21 @@
...
         }
 
+#define SETTINGMAPPING_SPECIALIZATION(_setting_, _json_, _value_, _default_, _path_) \
+        SETTINGMAPPING_SPECIALIZATION_EXTEND(_setting_, _json_, _value_, _default_, _path_, )
+
+#define SETTINGMAPPING_SPECIALIZATION_POLICY(_setting_, _json_, _value_, _default_, _path_, _valuePolicy_) \
+        SETTINGMAPPING_SPECIALIZATION_EXTEND(_setting_, _json_, _value_, _default_, _path_, \
+            static constexpr ValuePolicy Policy = _valuePolicy_; \
+            using policy_t = decltype(std::declval<GroupPolicy>().GetValue<Policy>())::value_type; \
+            static_assert(std::is_same<json_t, policy_t>::value); \
```
JohnMcPMS, line 96: :heart:

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -84,12 +84,56 @@ namespace AppInstaller::Settings
...
             Json::Value& root,
             std::map<Setting, details::SettingVariant>& settings,
             std::vector<std::string>& warnings)
         {
+            // Settings set by Group Policy override anything else. See if there is one.
+            auto policyValue = GetValueFromPolicy<S>();
+            if (policyValue.has_value())
+            {
+                // If the value is valid, use it.
+                // Otherwise, fall back to the settings.
```
JohnMcPMS, line 118: I think we actually have to have "if the policy is valid, use it.  otherwise fall back to the default".  If an admin has configured a value, I don't think we can use the user value ever.

JohnMcPMS:
>         // Settings can be loaded from settings.json or settings.json.backup files.

If settings is disabled by policy, we should just quit before ever even loading a file.

---
Refers to: src/AppInstallerCommonCore/UserSettings.cpp:252 in b96cf8a. [](commit_id = b96cf8a646f754d8c24206e26d462cbae9c8ba98, deletion_comment = False)

src/AppInstallerCLICore/Commands/RootCommand.cpp
lechacon: When disabling `winget` we want to at least be able to run `winget --info` and show the policies. I don't have a strong argument for `--help` but I guess it doesn't hurt either?
We also allow `winget features` even if they are all disabled.

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+        return itr->second;
+    }
+
+    bool GroupPolicy::IsEnabledOrNotConfigured(TogglePolicy policy) const
+    {
+        return GetState(policy) != PolicyState::Disabled;
+    }
+
+    static std::unique_ptr<GroupPolicy> s_groupPolicy;
```
JohnMcPMS, line 168: >static std::unique_ptr<GroupPolicy> s_groupPolicy; [](start = 4, length = 50)

It looks like you are doing it this way to enable testing scenarios.  It isn't thread safe this way though.

I would suggest a test hook be implemented this way:
```
GP& InstanceInternal(std::optional<GroupPolicy*> overridePolicy = {})
{
static GroupPolicy s_groupPolicy = make...;
static GroupPolicy* s_override = nullptr;

if (overridePolicy)
{
s_override = oP.value()
}

return (s_override ? *s_override : s_gP);
}

Instance()
{
return InstanceInternal();
}

TEST_HOOK_OVERRIDE_GP(GP* gp)
{
  InstanceInternal(gp);
}
```

Can also make the internal function the basic version when test hooks are disabled.

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -0,0 +1,186 @@
...
+        auto itr = m_toggles.find(policy);
+        if (itr == m_toggles.end())
+        {
+            return PolicyState::NotConfigured;
+        }
+
+        return itr->second;
+    }
+
+    bool GroupPolicy::IsEnabledOrNotConfigured(TogglePolicy policy) const
```
JohnMcPMS, line 163: >IsEnabledOrNotConfigured [](start = 22, length = 24)

Just wait until one of them is default disabled...

src/AppInstallerCommonCore/ExperimentalFeature.cpp
lechacon: I can add it if you think it's a good idea.

I originally had a single `IsAllowed` but then I wanted to be more explicit because some policies need a distinction between not-configured and enabled (e.g. default source), but in others it means the same thing (e.g. settings). We currently don't have policies where not-configured and disabled mean the same but I don't discard that possibility in the future.

src/AppInstallerCLICore/Commands/RootCommand.cpp
JohnMcPMS: I'm not a huge fan of having --info spew all of the policies, but I may not have much say.

---
In reply to: [598016873](https://github.com/microsoft/winget-cli/pull/810#discussion_r598016873) [](ancestors = 598016873)

src/AppInstallerCommonCore/ExperimentalFeature.cpp
JohnMcPMS: I agree that the semantics are an issue, mostly because they could change in the future.  "IsEnabled" or "IsDisabled" implies some amount of "handle the not set case by knowing the default".  However, we need that decision to be centralized.  We don't want the caller deciding how to handle the unset state; it could lead to bugs.

Given that we need that control to be centralized, I think that `IsEnabled` is the best choice of action, and we would have to encode the default state into a type if it is ever disabled.

---
In reply to: [598018838](https://github.com/microsoft/winget-cli/pull/810#discussion_r598018838) [](ancestors = 598018838)

lechacon: Okay. I had a function like that and a member like `TogglePolicyInternal::m_defaultIsEnabled` to guide it; I'll go back to that. And for the few places where we care about whether it was by default or explicit I'll use the enum I had for the state.

src/AppInstallerCLICore/Commands/RootCommand.cpp
denelon: The policy display was an explicit request. If a user has a configuration and a policy blocks it, the user needs to know. When or if a policy is removed, the settings should again apply.

src/AppInstallerCLITests/UserSettings.cpp
```diff
@@ -217,11 +201,32 @@ TEST_CASE("SettingProgressBar", "[settings]")
         REQUIRE(userSettingTest.Get<Setting::ProgressBarVisualStyle>() == VisualStyle::Accent);
         REQUIRE(userSettingTest.GetWarnings().size() == 1);
     }
+    SECTION("Overriden by Group Policy")
```
jsoref, line 204:
```suggestion
    SECTION("Overridden by Group Policy")
```

src/AppInstallerCLITests/UserSettings.cpp
```diff
@@ -270,12 +275,32 @@ TEST_CASE("SettingAutoUpdateIntervalInMinutes", "[settings]")
         REQUIRE(userSettingTest.Get<Setting::AutoUpdateTimeInMinutes>() == cinq);
         REQUIRE(userSettingTest.GetWarnings().size() == 1);
     }
+    SECTION("Overriden by Group Policy")
```
jsoref, line 278:
```suggestion
    SECTION("Overridden by Group Policy")
```

src/AppInstallerCommonCore/Public/winget/GroupPolicy.h
```diff
@@ -0,0 +1,136 @@
...
+            // value_t - type of the policy
+            // ReadAndValidate() - Function that reads the value and does semantic validation.
+
+            // For simple values:
+            //  ValueName - Name of the registry value
+            //  ValueType - Type of the registry value
+            //  reg_value_t - Type returned by the registry when reading the value
+        };
+
+#define POLICYMAPPING_SPECIALIZATION(_policy_, _type_) \
```
jsoref, line 62: Do you need to write this as one word instead of:
```suggestion
#define POLICY_MAPPING_SPECIALIZATION(_policy_, _type_) \
```

?

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -81,10 +83,21 @@ namespace AppInstaller::Settings
...
+            _extension_ \
         }
 
+#define SETTINGMAPPING_SPECIALIZATION(_setting_, _json_, _value_, _default_, _path_) \
+        SETTINGMAPPING_SPECIALIZATION_EXTEND(_setting_, _json_, _value_, _default_, _path_, )
+
+#define SETTINGMAPPING_SPECIALIZATION_POLICY(_setting_, _json_, _value_, _default_, _path_, _valuePolicy_) \
+        SETTINGMAPPING_SPECIALIZATION_EXTEND(_setting_, _json_, _value_, _default_, _path_, \
+            static constexpr ValuePolicy Policy = _valuePolicy_; \
+            using policy_t = decltype(std::declval<GroupPolicy>().GetValue<Policy>())::value_type; \
```
jsoref, line 95: `declval` should be added to https://github.com/microsoft/terminal/blob/main/.github/actions/spelling/dictionary/apis.txt

src/AppInstallerCommonCore/Public/winget/GroupPolicy.h
lechacon: I used one word to mirror the existing SETTINGMAPPING_SPECIALIZATION, but I do like it more as two words. I will change it to two.

lechacon:
This is still missing E2E tests and the policies for controlling sources. I'll submit that as a separate PR as this one already got too big.

src/AppInstallerCLICore/Argument.cpp
lechacon: I made the constructors private. I didn't find anything using policies or features that wasn't already there.

src/AppInstallerCLICore/Command.h
```diff
@@ -16,21 +17,22 @@
...
+        CommandException(Resource::LocString message, std::string_view param) : m_message(std::move(message)), m_param(Utility::LocIndString{ param }) {}
 
         const Resource::LocString& Message() const { return m_message; }
-        const Utility::LocIndString Param() const { return m_param; }
+        const auto Param() const { return m_param; }
 
     private:
         Resource::LocString m_message;
-        Utility::LocIndString m_param;
+        std::variant<Resource::LocString, Utility::LocIndString> m_param;
```
JohnMcPMS, line 35: >std::variant<Resource::LocString, Utility::LocIndString> m_param; [](start = 8, length = 65)

I was just touching this and I didn't think to do it then, but now it comes to mind that LocString should just derive from LocIndString.

JohnMcPMS: Why did this become a shared_ptr?

---
In reply to: [602594039](https://github.com/microsoft/winget-cli/pull/810#discussion_r602594039) [](ancestors = 602594039)

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -36,7 +36,35 @@ namespace AppInstaller::CLI
             context.Reporter.Warn() << Resource::String::SettingLoadFailure << std::endl;
             for (const auto& warning : User().GetWarnings())
             {
-                context.Reporter.Warn() << warning << std::endl;
+                auto warn = context.Reporter.Warn();
```
JohnMcPMS, line 39: >auto [](start = 16, length = 4)

const auto&

src/AppInstallerCLICore/Command.h
lechacon: I wanted it to be able to hold a LocString or a LocIndString so I made it a pointer. I tried with a unique_ptr but `ParseArgumentsStateMachine::ThrowIfError()` needed to copy it to throw so I made it shared. I need to check if I can move it instead.

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
lechacon: This can't be auto& because Warn() does not return a reference.

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+namespace AppInstaller::Settings
+{
+    namespace
+    {
+        GroupPolicy& InstanceInternal(std::optional<GroupPolicy*> overridePolicy = {})
+        {
+            // TODO: Use final location
+            constexpr std::string_view PoliciesKeyPath = "test\\policy"; // "SOFTWARE\\Policies\\Microsoft\\Windows\\WinGet";
+            static GroupPolicy s_groupPolicy{ Registry::Key::OpenIfExists(HKEY_CURRENT_USER /* HKEY_LOCAL_MACHINE */, PoliciesKeyPath) };
```
JohnMcPMS, line 17: I don't like checking this in with test values here; we should disable it until we have the real location.  I assume that constructing a default `GroupPolicy` would do that.

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -0,0 +1,222 @@
...
+            if (!value.has_value())
+            {
+                return std::nullopt;
+            }
+
+            if (value->GetType() != T)
+            {
+                AICLI_LOG(Core, Warning, << "Value for policy '" << valueName << "' does not have expected type");
+                return std::nullopt;
+            }
```
JohnMcPMS, line 47: This is restrictive; some of the types are somewhat interchangeable (string vs expand string).  Given that we do not control the code that writes these, we should be more flexible.  I realize that might require catching an exception and continuing, but another option would be to create a TryGetValue method that doesn't throw on a failure.

src/AppInstallerCLICore/Command.h
lechacon: Changed to use a plain LocIndString (not a pointer).

src/AppInstallerCommonCore/GroupPolicy.cpp
lechacon: Added a Value::TryGetValue() function

src/AppInstallerCommonCore/Public/winget/Registry.h
```diff
@@ -87,14 +87,27 @@ namespace AppInstaller::Registry
         template <Type T>
         typename details::ValueTypeSpecifics<static_cast<DWORD>(T)>::value_t GetValue() const
         {
-            EnsureType(T);
+            THROW_HR_IF(E_INVALIDARG, !HasCompatibleType(T));
```
JohnMcPMS, line 90: >THROW_HR_IF(E_INVALIDARG, !HasCompatibleType(T)); [](start = 12, length = 49)

nit (for now): Better to implement by calling TryGetValue and throwing when it returns nullopt to prevent future changes from needing to be made to both code paths.

JohnMcPMS (approved):

PR #816, opened 2021-03-22
Create ManifestSpecv1.0.md

denelon:
Manifest Schema v1.0

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/816)

### DIFF
--- a/doc/ManifestSpecv1.0.md
+++ b/doc/ManifestSpecv1.0.md
@@ -47,13 +47,12 @@ Path: manifests / m / Microsoft / WindowsTerminal / 1.6.10571.0 / WindowsTermina
 ```YAML
 PackageIdentifier: "Microsoft.WindowsTerminal"
 PackageVersion: "1.6.10571.0"
-PackageLocale: "en-US"
-Publisher: "Microsoft"
-PackageName: "Windows Terminal"
+Locale: "en-US"
 License: "MIT"
 ShortDescription: "The new Windows Terminal, a tabbed command line experience for Windows."
 Installers: 
  - Architecture: "x64"
+   InstallerType: "msix"
    InstallerUrl: "https://github.com/microsoft/terminal/releases/download/v1.6.10571.0/Microsoft.WindowsTerminal_1.6.10571.0_8wekyb3d8bbwe.msixbundle"
    InstallerSha256: 092aa89b1881e058d31b1a8d88f31bb298b5810afbba25c5cb341cfa4904d843
    SignatureSha256: e53f48473621390c8243ada6345826af7c713cf1f4bbbf0d030599d1e4c175ee
--- a/doc/ManifestSpecv1.0.md
+++ b/doc/ManifestSpecv1.0.md
@@ -9,8 +9,6 @@ YAML Specification
    3) Minimal singleton YAML file example
    4) Multiple file YAML example
    5) Best Practices
-History
-
 
 # YAML Specification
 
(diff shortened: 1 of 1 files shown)


doc/ManifestSpecv1.0.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+Path: manifests / m / Microsoft / WindowsTerminal / 1.6.10571.0 / Microsoft.WindowsTerminal.locale.fr-FR.yaml
+
+```YAML
+PackageIdentifier: "Microsoft.WindowsTerminal"
+PackageVersion: "1.6.10571.0"
+PackageLocale: "fr-FR"
+Publisher: "Microsoft"
+ShortDescription: "Le nouveau terminal Windows, une expérience de ligne de commande à onglets pour Windows."
```
jsoref, line 121: Easiest way to ignore this is to add something like:
```
^ShortDescription: "Le nouveau.*"$
```
to `patterns.txt` (preferably w/ a comment referencing this line).

doc/ManifestSpecv1.0.md
```diff
@@ -0,0 +1,168 @@
...
+
+```YAML
+PackageIdentifier: "Microsoft.WindowsTerminal"
+PackageVersion: "1.6.10571.0"
+PackageLocale: "en-US"
+Publisher: "Microsoft"
+PackageName: "Windows Terminal"
+License: "MIT"
+ShortDescription: "The new Windows Terminal, a tabbed command line experience for Windows."
+Installers:
```
jedieaston, line 55: Is InstallerType no longer required a required attribute as it is currently?

doc/ManifestSpecv1.0.md
```diff
@@ -0,0 +1,168 @@
...
+PackageLocale: "en-US"
+Publisher: "Microsoft"
+PackageName: "Windows Terminal"
+License: "MIT"
+ShortDescription: "The new Windows Terminal, a tabbed command line experience for Windows."
+Installers: 
+ - Architecture: "x64"
+   InstallerUrl: "https://github.com/microsoft/terminal/releases/download/v1.6.10571.0/Microsoft.WindowsTerminal_1.6.10571.0_8wekyb3d8bbwe.msixbundle"
+   InstallerSha256: 092aa89b1881e058d31b1a8d88f31bb298b5810afbba25c5cb341cfa4904d843
+   SignatureSha256: e53f48473621390c8243ada6345826af7c713cf1f4bbbf0d030599d1e4c175ee
```
jedieaston, line 59: Is InstallerType no longer required?

denelon: Great catch! I've been working on this one so much I missed it.

quangkieu:
I wonder if we could have field for windows shim like wt.exe of windows terminal, or wsl.exe, unbuntu.exe of wsl distro?

jedieaston:
> 
> 
> I wonder if we could have field for windows shim like wt.exe of windows terminal, or wsl.exe, unbuntu.exe of wsl distro?

That should be going in the command field, I think.

denelon:
@quangkieu, yes the "commands" key is intended to identify the command one would execute to run the program/package. It's currently just an informational field, but that is the intent.

quangkieu:
I wonder how about winget generate one, for example the unpacked zip install, or git.exe so we only need folder windowsapp for those shim but the app could install, unpack in correct folder. We also have a flag if user want to disable the generated shim?

denelon:
Our plan for supporting .zip and standalone .exe as a package is a bit different. We're looking at having a path where they will be placed on a drive, and an entry in the PATH environment variable. We will also add an entry in Add/Remove Programs so they can be removed and upgraded as other packages are. This Add/Remove Programs integration would require a more recent version of Windows, but the package manager would still be able to handle the scenarios for uninstall and upgrade.

quangkieu:
My only concern is the amount of entries on PATH variable. And if other microsoft team, product already has shim feature, I would want to suggest to improve and promote it to be official solution instead just patch work in those products only

KevinLaMS (approved):
Looks great.

PR #818, opened 2021-03-25
Add a new COMContext object in AppInstaller Core to enable COM Interface to consume AppInstaller

sachintaMSFT:
Add a new COMContext object in AppInstaller Core to enable COM Interface to issue any commands to AppInstaller. The COM Context object will also enable COM Interface to see the progress, receive errors and cancel execution of commands, where applicable.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/818)

### DIFF
--- a/src/AppInstallerCLICore/COMContext.cpp
+++ b/src/AppInstallerCLICore/COMContext.cpp
@@ -8,38 +8,28 @@ namespace AppInstaller
 
     NullStream::NullStream()
     {
-        null_Out.reset(new std::ostream(&nullStreamBuf));
-        null_In.reset(new std::istream(&nullStreamBuf));
+        m_nullOut.reset(new std::ostream(&m_nullStreamBuf));
+        m_nullIn.reset(new std::istream(&m_nullStreamBuf));
     }
 
     void COMContext::BeginProgress()
     {
-        m_comProgressCallback(ReportType::BeginProgress, m_current, m_maximum, m_type, m_executionPhase);
+        m_comProgressCallback(ReportType::BeginProgress, 0, 0, ProgressType::None, m_executionStage);
     };
 
-    void COMContext::OnProgress(uint64_t current, uint64_t maximum, ProgressType type)
+    void COMContext::OnProgress(uint64_t current, uint64_t maximum, ProgressType progressType)
     {
-        SetProgress(current, maximum, type);
-        m_comProgressCallback(ReportType::Progressing, m_current, m_maximum, m_type, m_executionPhase);
+        m_comProgressCallback(ReportType::Progressing, current, maximum, progressType, m_executionStage);
     }
 
-    void COMContext::EndProgress(bool hideProgressWhenDone)
+    void COMContext::EndProgress(bool)
     {
-        UNREFERENCED_PARAMETER(hideProgressWhenDone);
-        m_comProgressCallback(ReportType::EndProgress, m_current, m_maximum, m_type, m_executionPhase);
+        m_comProgressCallback(ReportType::EndProgress, 0, 0, ProgressType::None, m_executionStage);
     };
 
-    void COMContext::SetExecutionStage(ExecutionStage executionPhase)
+    void COMContext::SetExecutionStage(CLI::Workflow::ExecutionStage executionStage, bool)
     {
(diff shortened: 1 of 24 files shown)


sachintaMSFT:
> #### Misspellings found, please review:
> * streambuf
> 
> To accept these changes, run the following commands from this repository on this branch

will add streambuf word to .github/actions/spelling/expect.txt in alphabetical sort order along in the commit for incorporation of any review comments received.

src/AppInstallerCLICore/Public/AppInstallerCLICoreExecute.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include "ExecutionContext.h"
+#include "Command.h"
+
+namespace AppInstaller::CLI
+{
+    int Execute(Execution::Context& context, std::unique_ptr<Command>& command);
```
JohnMcPMS, line 7: Assuming that the plan is to implement the COM interface classes in another static lib (I support that plan), then many other headers will "need" to be placed in public.  I think it makes sense for Command and Context to be made public, and then COMContext will actually be moved over to that lib.

This function should then be in Command.h, and the "make this command" function should be too.

For now, COMContext can live where you have it in this PR, but at least move this function to Command.h.

src/AppInstallerCLICore/Core.cpp
```diff
@@ -115,6 +117,17 @@ namespace AppInstaller::CLI
...
+        return Execute(context, command);
+    }
+    // End of the line exceptions that are not ever expected.
+    // Telemetry cannot be reliable beyond this point, so don't let these happen.
+    catch (...)
+    {
+        return APPINSTALLER_CLI_ERROR_INTERNAL_ERROR;
+    }
+
+    int Execute(Execution::Context& context, std::unique_ptr<Command>& command)
```
JohnMcPMS, line 129: RE: my other comment, move this to Command.cpp.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -101,6 +101,10 @@ namespace AppInstaller::CLI::Execution
             WI_ClearAllFlags(m_flags, flags);
         }
 
+        void SetExecutionStage(Workflow::ExecutionStage stage, bool allowBackward);
+        virtual void SetExecutionStage(Workflow::ExecutionStage stage);
```
JohnMcPMS, line 105: No need to separate these functions, but if you are going to do so, the virtual function should be protected.

src/AppInstallerCLITests/TestCommon.cpp
```diff
@@ -139,6 +139,15 @@ namespace TestCommon
         }
     }
 
+    void TestProgress::BeginProgress()
+    {
+    }
+
+    void TestProgress::EndProgress(bool hideProgressWhenDone)
+    {
+        _Unreferenced_parameter_(hideProgressWhenDone);
```
JohnMcPMS, line 148: Remove name from signature.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/COMContext.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    NullStream::NullStream()
+    {
+        null_Out.reset(new std::ostream(&nullStreamBuf));
+        null_In.reset(new std::istream(&nullStreamBuf));
+    }
+
+    void COMContext::BeginProgress()
+    {
+        m_comProgressCallback(ReportType::BeginProgress, m_current, m_maximum, m_type, m_executionPhase);
```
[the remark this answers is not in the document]
sachintaMSFT: I do reset progress values in SetExecutionStage. But I guess if there are multiple separate & sequential progresses in same ExecutionStage (which I think Discover stage has), that doesn't help BeginProgress send reset progress values. And if that progress is super quick that OnProgress isn't called and EndProgress get called directly, then the leak goes into EndProgress as well. Accept your suggestion.

JohnMcPMS (approved):

Issue #149, opened 2020-05-15
User vs. System installation (non MSIX)


denelon:
As a user I want to be able to specify whether the software installs for me or any user on my machine.

Note:
The manifest needs to support specifications for both.
The client should default to user mode "out of the box" if no preference is supplied in the manifest and both are possible.
The user should be able to specify user or machine based installation with the client regardless of the manifest default if both are possible.

thlac:
It would also be great if it was possible to run this in context of the localsystem user specifically for machine-based installations. Not sure if this is possible due to being bundled in DesktopAppInstaller, but this would be a great for extending with group policy based deployments or other unattended means of deployment.

JohnMcPMS:
I have added the user control, but this issue still needs work per-installer type to determine if we can dynamically specify scope to the underlying installer. For instance, MSI supports specifying it, as well as embedding the value. And MSIX can be installed for the machine, but requires additional code to convert the user install to a machine install.

So for each type, we need to understand:

1. Are there command line options that let us choose the scope dynamically?
2. Is there metadata embedded in the installer that will tell us the scope?
3. If 2, can it be overridden or would that lead to a dysfunctional package?

With this we could then improve the default behaviors, as currently we don't allow unknown scopes to meet the requirement of a particular given one.

denelon:
#965

PR #819, opened 2021-03-26
Settings and command line argument for specifying scope

JohnMcPMS:
Furthers #149, but needs work per-installer type to determine if we can dynamically specify scope to the underlying installer.  For instance, MSI supports specifying it, as well as embedding the value.  And MSIX can be installed for the machine, but requires additional code to convert the user install to a machine install.

## Change
This change adds settings and a command line argument to `install` to enable control over the scope of the package install.  Scope is the difference between a user or machine install of a package.

The settings allow both the preference and requirement of scope to be set.  A preference affects the sort order for selecting installers, while a requirement filters the installers.  The default behavior (with no settings configured) is equivalent to:
```json
    "installBehavior": {
        "preferences": {
            "scope": "user"
        }
    },
```

The command line argument is `--scope` and accepts the values `user` and `machine`.  This is treated as a requirement, overriding any requirement set in the settings file.

The default value for Scope in a manifest is being changed from implicitly `User` to just `Unknown`.  If a requirement is specified (which cannot be `Unknown`), it will not match the default value of `Unknown`.  This is necessary to ensure that we can meet the requirement.  It will also affect what is a valid manifest; an unknown scope value will now conflict with any other scope value to prevent specifying an unknown and known value together.

Finally, to implement the actual behavior change, `ManifestComparator` is completely rewritten to be a more dynamic configuration of filters and comparators.  This will allow for more behavior to be added in the future in a straightforward manner.  It also fixes a bug that previously existed where the sorting was not consistent (ie. would not produce a properly sorted list).  It will now only use a lower priority sort when the higher priority sort considers the two installers equal.

## Validation
Added tests for `ManifestComparator`, as well as some infrastructure for overriding the user settings in unit tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/819)


src/AppInstallerCLI.sln
```diff
@@ -78,6 +78,15 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "spelling", "spelling", "{2A
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "cpprestsdk", "cpprestsdk\cpprestsdk.vcxproj", "{866C3F06-636F-4BE8-BC24-5F86ECC606A1}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "schemas", "schemas", "{92637527-6CDA-4F4A-84FD-858793776777}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "JSON", "JSON", "{F2149997-295A-4593-9282-4C675DFEB670}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "settings", "settings", "{1487DFBB-7C53-4BD3-9B2C-9B94C6C91528}"
```
yao-msft, line 85: Is it intentional to create 3 new projects here?

yao-msft:
>             Argument{ "force", Argument::NoAlias, Args::Type::Force, Resource::String::InstallForceArgumentDescription, ArgumentType::Flag },

So this could also be Utility::LocIndView like above?

---
Refers to: src/AppInstallerCLICore/Commands/InstallCommand.cpp:41 in 8838fcd. [](commit_id = 8838fcd9b31caa3fa8ea94e0aa769c2a040b4a56, deletion_comment = False)

src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -78,23 +85,31 @@ namespace AppInstaller::CLI
...
+        {
+            throw CommandException(Resource::String::BothManifestAndSearchQueryProvided);
+        }
+
+        if (execArgs.Contains(Args::Type::InstallScope))
         {
-            throw CommandException(Resource::String::BothManifestAndSearchQueryProvided, "");
+            if (ConvertToScopeEnum(execArgs.GetArg(Args::Type::InstallScope)) == Manifest::ScopeEnum::Unknown)
+            {
+                throw CommandException(Resource::String::InvalidArgumentValueError, s_ArgumentName_Scope, "user', 'machine");
```
yao-msft, line 107: >"user', 'machine" [](start = 106, length = 17)

Missing single quote and the leading space?  " 'user', 'machine'"

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -199,6 +200,51 @@ namespace AppInstaller::Settings
...
+            if (Utility::CaseInsensitiveEquals(value, s_scope_user))
+            {
+                return ScopePreference::User;
+            }
+            else if (Utility::CaseInsensitiveEquals(value, s_scope_machine))
+            {
+                return ScopePreference::Machine;
+            }
+
+            return {};
```
yao-msft, line 218: >{} [](start = 19, length = 2)

Do we have a way to output warnings for unrecognize settings?

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -3,93 +3,244 @@
...
-            if (installedType != Manifest::InstallerTypeEnum::Unknown &&
-                !Manifest::IsInstallerTypeCompatible(installer.InstallerType, installedType))
+            bool IsApplicable(const Manifest::ManifestInstaller& installer) override
             {
+                return Utility::IsApplicableArchitecture(installer.Arch) != Utility::InapplicableArchitecture;
+            }
+
+            std::string ExplainInapplicable(const Manifest::ManifestInstaller& installer) override
+            {
+                std::string result = "Machine is compatible with ";
```
yao-msft, line 52: >is [](start = 46, length = 2)

not

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -3,93 +3,244 @@
...
+            }
+
+            bool IsApplicable(const Manifest::ManifestInstaller& installer) override
+            {
+                return m_requirement == Manifest::ScopeEnum::Unknown || installer.Scope == m_requirement;
+            }
+
+            std::string ExplainInapplicable(const Manifest::ManifestInstaller& installer) override
+            {
+                std::string result = "Installer scope does not matched required scope: ";
```
yao-msft, line 191: >matched [](start = 63, length = 7)

match

yao-msft (approved):
:shipit:

src/AppInstallerCLI.sln
JohnMcPMS: They are just folders in the solution.

---
In reply to: [602805965](https://github.com/microsoft/winget-cli/pull/819#discussion_r602805965) [](ancestors = 602805965)

src/AppInstallerCLICore/Commands/InstallCommand.cpp
JohnMcPMS: This is because the output of this is value wraps it in single quotes currently.  Maybe I should move it to be a vector of values, then I wouldn't need to do this here.

---
In reply to: [602806765](https://github.com/microsoft/winget-cli/pull/819#discussion_r602806765) [](ancestors = 602806765)

src/AppInstallerCommonCore/UserSettings.cpp
JohnMcPMS: Validate takes care of that when it gets a nullopt back from these functions.

---
In reply to: [602807223](https://github.com/microsoft/winget-cli/pull/819#discussion_r602807223) [](ancestors = 602807223)

JohnMcPMS:
>             Argument{ "force", Argument::NoAlias, Args::Type::Force, Resource::String::InstallForceArgumentDescription, ArgumentType::Flag },

Taken care of by @lechacon in his GP change.

---
In reply to: [808818926](https://github.com/microsoft/winget-cli/pull/819#issuecomment-808818926) [](ancestors = 808818926)

---
Refers to: src/AppInstallerCLICore/Commands/InstallCommand.cpp:41 in 8838fcd. [](commit_id = 8838fcd9b31caa3fa8ea94e0aa769c2a040b4a56, deletion_comment = False)

PR #821, opened 2021-03-26
Rest endpoint helper and json object field check - Bug fix

ashpatil-msft:
Fixed two bugs in the Rest source code:
- Added check to format Rest API Uri when attempting to get Information from Rest source
- Added check to see if field is present before getting value for some fields that got missed.

Manually tested and verified basic install and manifest deserialization.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/821)


src/AppInstallerRepositoryCore/Rest/Schema/RestHelper.h
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+
+namespace AppInstaller::Repository::Rest::Schema
+{
+    // Rest source helper.
+    struct RestHelper
```
JohnMcPMS, line 8: You don't have to put every function inside a class type.

JohnMcPMS (approved):
So we obviously need tests eventually, and I think we do need to do the http client injection for the unit tests so that we can easily test the corner cases.

PR #836, opened 2021-03-31
Fix date formatting

martincostello:
I noticed during a presentation that the date format for the exported JSON files wasn't formatted correctly (the space just before the `3` for March instead of a `0`):

image

This pull request updates the formatting to ensure that the day and month are padded with zeroes like the time already is.

Generating an export JSON locally, I get something similar to the following before and after the change.

## Before

```json
{
    "CreationDate" : "2021- 3-31 13:18:58.648"
}
```

## After

```json
{
    "CreationDate" : "2021-03-31 13:17:55.408"
}
```

There's a test that the date is output below, but it doesn't validate anything beyond it being present. As the clock isn't injectable in some way, I wasn't sure on the best way to add a test for it that would verify the format was always correct, rather than just on days that should have a leading zero, so I haven't updated any test for this fix.

https://github.com/microsoft/winget-cli/blob/9e104ebe699617ef003676eb2ab9837136d799b9/src/AppInstallerCLITests/PackageCollection.cpp#L66

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/836)

### DIFF
--- a/src/AppInstallerCommonCore/DateTime.cpp
+++ b/src/AppInstallerCommonCore/DateTime.cpp
@@ -16,8 +16,8 @@ namespace AppInstaller::Utility
 
         stream
             << std::setw(4) << (1900 + localTime.tm_year) << '-'
-            << std::setw(2) << (1 + localTime.tm_mon) << '-'
-            << std::setw(2) << localTime.tm_mday << (useRFC3339 ? 'T' : ' ')
+            << std::setw(2) << std::setfill('0') << (1 + localTime.tm_mon) << '-'
+            << std::setw(2) << std::setfill('0') << localTime.tm_mday << (useRFC3339 ? 'T' : ' ')
             << std::setw(2) << std::setfill('0') << localTime.tm_hour << ':' 
             << std::setw(2) << std::setfill('0') << localTime.tm_min << ':' 
             << std::setw(2) << std::setfill('0') << localTime.tm_sec << '.';


lechacon:
Good catch.

That field in the JSON file is supposed to be a [date-time](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3.1) per the [schema](https://github.com/microsoft/winget-cli/blob/ee7dd86f827b220b5f5d20c0c4ec208c93cf7517/schemas/JSON/packages/packages.schema.1.0.json#L22). I would have thought that schema validation would have caught anything like this, but it seems valijson does not support date-time. I'm also not too worried that it doesn't because that field is mostly informational.

I tried to check what other things I missed to have the right format. I think I missed the T separator and maybe the time zone. (Not clear on whether the time zone is needed.) The T separator can be added by adding a `true` parameter to the call to [`OutputTimePoint`](https://github.com/microsoft/winget-cli/blob/9879f0765c80a2e2819891d6e488a285751b3ffb/src/AppInstallerCLICore/PackageCollection.cpp#L95).

I'm not sure if we should try to validate this field. Regardless, it may be good to add a test with the bad format to ensure we accept or reject it.

I'm fine taking this change as-is and looking at the other things I said later myself, unless you want to try your hand at it @martincostello

martincostello:
> I'm fine taking this change as-is and looking at the other things I said later myself, unless you want to try your hand at it @martincostello

I don't really have any C++ chops, so if you're happy to follow up I'd be fine to leave it to you if that's OK 😄

I spotted the TODO at the call site regarding using a time zone or not, but I guess that's more of an implementation decision on your side to make a call.

lechacon (approved):

PR #838, opened 2021-03-31
Rest source search functionality and other todos

ashpatil-msft:
- Added search serializer to be able to send in the query, inclusions and filters from users.
- Implemented version support based on the versions returned by information endpoint.
- Added logic to handle different http response status codes.
- Fixed other Todos.
- Added unit tests for RestHelper, JsonHelper and RestClient methods.

Todo:
- Will add unit tests for other features in rest source in the upcoming PRs.

Manual testing:
- Verified winget show, search and install command work.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/838)


.github/actions/spelling/expect.txt
```diff
@@ -31,6 +32,7 @@ bomgar
 BOMs
 brk
 Buf
+Bversion
```
jsoref, line 35: Add this to the `patterns.txt` file instead:
```
REQUIRE\(RestHelper::GetRestAPIBaseUri\(".*"\) == L".*"

# URL escaped characters
\%[0-9A-F]{2}
```

The second is from https://github.com/check-spelling/check-spelling/wiki/Configuration-Examples:-patterns

src/AppInstallerRepositoryCore/Rest/Schema/RestHelper.cpp
```diff
@@ -2,16 +2,74 @@
 // Licensed under the MIT License.
 #include "pch.h"
 #include "RestHelper.h"
+#include "cpprest/uri_builder.h"
+#include "cpprest/http_client.h"
+#include "Rest/Schema/Json/JsonHelper.h"
+#include "Rest/Schema/Json/CommonRestConstants.h"
+using namespace AppInstaller::Repository::Rest::Schema::Json;
```
JohnMcPMS, line 9: Separate `using namespace` from the `#include`s.

src/AppInstallerRepositoryCore/Rest/Schema/RestHelper.cpp
```diff
@@ -2,16 +2,74 @@
...
+    {
+        web::uri_builder builder(restApiUri);
+        builder.append_path(path, true);
+        return builder.to_string();
+    }
+
+    utility::string_t RestHelper::MakeQueryParam(std::string_view queryName, const std::string& queryValue)
+    {
+        std::string queryParam;
+        queryParam.append(queryName).append("=").append(queryValue);
```
JohnMcPMS, line 47: >queryParam.append(queryName).append("=").append(queryValue); [](start = 8, length = 60)

What if `queryValue` contains characters that need to be encoded?

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -129,13 +130,14 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
 
     IRestClient::SearchResult Interface::OptimizedSearch(const SearchRequest& request) const
     {
-        // TODO: Send in VersionLatest = true query param.
         SearchResult searchResult;
-        std::vector<Manifest::Manifest> manifests = GetManifests(request.Filters[0].Value, {}, {});
+        std::map<std::string_view, std::string> queryParams;
+        queryParams.emplace(LatestVersionQueryParam, "true");
```
JohnMcPMS, line 135: >queryParams.emplace(LatestVersionQueryParam, "true"); [](start = 8, length = 53)

Per-offline discussions, we want to remove this.  And I don't think it is actually appropriate here anyway, as we don't know the caller's intention RE: versions at this point.  I know I was the one who said to add it, but I can be wrong.

src/AppInstallerRepositoryCore/Rest/Schema/Json/SearchRequestSerializer.h
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "pch.h"
+#include "cpprest/json.h"
+#include "Rest/Schema/IRestClient.h"
+
+namespace AppInstaller::Repository::Rest::Schema::Json
```
JohnMcPMS, line 8: Shouldn't these serialize/deserialize functions be in the 1_0 directory?  They seem like the primary thing that would change between versions.

src/AppInstallerRepositoryCore/Rest/Schema/Json/SearchRequestSerializer.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+            if (searchRequest.Query)
+            {
+                auto& requestMatch = searchRequest.Query.value();
+                web::json::value requestMatchObject = web::json::value::object();
+                requestMatchObject[JsonHelper::GetUtilityString(RequestMatch)] = GetRequestMatchJsonObject(requestMatch);
+                json_body[JsonHelper::GetUtilityString(Query)] = requestMatchObject;
+            }
+
+            if (searchRequest.Filters.size() > 0)
```
JohnMcPMS, line 59: >searchRequest.Filters.size() > 0 [](start = 16, length = 32)

While not a big deal for a vector, generally it is more efficient to use `!container.empty()` for a check like this.  Other containers don't necessarily store their size, and calculating it can be expensive compared to the empty check.

src/AppInstallerCLITests/RestClient.cpp
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "TestCommon.h"
+#include <Rest/RestClient.h>
```
JohnMcPMS, line 5: I don't feel that these tests are sufficient.  I think we can have fairly extensive test coverage on the serialize/deserialize by writing our own http client for rest-cpp.  That will likely require creating a separate lib for the current http client code to prevent a linker conflict, but then I would hope that we can make a test one that allows us to inject our own test responses.

It may also be good experience for future work that we have, depending on the current availability of metadata about the connection.

src/AppInstallerRepositoryCore/Rest/Schema/Json/SearchRequestSerializer.h
ashpatil-msft: I agree, I will move them to 1_0 dir

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerRepositoryCore/Rest/Schema/RestHelper.cpp
ashpatil-msft: Yes, they need to be. This method is called by AppendQueryParams, which does the encoding part when it appends the query param to the uri.

---
In reply to: [605982807](https://github.com/microsoft/winget-cli/pull/838#discussion_r605982807) [](ancestors = 605982807)

src/AppInstallerCLITests/RestClient.cpp
ashpatil-msft: As discussed offline, I will be working on it and sending a different PR along with more tests.

---
In reply to: [605993882](https://github.com/microsoft/winget-cli/pull/838#discussion_r605993882) [](ancestors = 605993882)

JohnMcPMS (approved):
:shipit:

PR #841, opened 2021-04-01
Group Policy for controlling sources

lechacon:
Adding enforcement of Group Policies for additional required sources and allowed sources. This change includes:
* Reading of source lists from Group Policy. Each source is specified by a JSON with all its details.
* A new `GroupPolicyException` type for reporting errors.
* Enforcement of the policies when adding and removing sources (only allow adding allowed sources and don't allow removing additional required sources).
* A `source export` command to produce the JSON strings used in the policies.
* Unit tests.

Still pending for Group Policies:
* E2E tests
* Use final registry key path, value names, and policy names

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/841)


src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
```diff
@@ -82,15 +83,6 @@ namespace AppInstaller::Repository
         virtual SearchResult Search(const SearchRequest& request) const = 0;
     };
 
-    // Interface extension to ISource for locally installed packages.
```
JohnMcPMS, line 85: Hey, I'm eventually going to get a chance to use that...

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -104,26 +104,110 @@ namespace AppInstaller::Settings
             // Use folding to call each policy validate function.
             (FoldHelper{}, ..., Validate<static_cast<ValuePolicy>(P)>(policiesKey, policies));
         }
+
+        // Reads a list from a Group Policy.
+        // The list is stored in a sub-key of the policies key, and each value in that key is a list item.
+        // Cases not considered by this function because we don't use them:
+        //  - When the list is in an arbitrary key, not a sub key.
+        //  - When the list values are mixed with other values and are identified by a prefix in their names.
+        //  - When the value names are relevant.
```
JohnMcPMS, line 113: How do these values end up looking?  Are the value names under the control of the IT Admin, but we simply don't care what they are?  Or are they named by the system?

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
lechacon: I'll add it back.

src/AppInstallerCommonCore/GroupPolicy.cpp
lechacon: The ADMX definition gives two options: let the IT Admin name them, or let the system name them just numbering (0, 1, ..) with an optional prefix set by us.
The PR I have for the ADMX lets the IT Admin set the names but I may change it to automatic.
In any case we don't care about the value names.

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -104,26 +104,110 @@ namespace AppInstaller::Settings
...
+            Json::String jsonErrors;
+            if (!jsonReader->parse(jsonString->c_str(), jsonString->c_str() + stringLength, &sourceJson, &jsonErrors))
+            {
+                AICLI_LOG(Core, Warning, << "Registry value does not contain a valid JSON: " << jsonErrors);
+                return std::nullopt;
+            }
+
+            SourceFromPolicy source;
+
+#define READ_SOURCE_ATTRIBUTE(_attr_) \
```
JohnMcPMS, line 164: (I originally was going to say that you shouldn't use a macro here, but I'm going to just suggest the options)

If we must use a macro here, can we not construct the string 3 times?

But if you want to make it a lambda with default ref capture and use pointer to member syntax that would be cool too.
[https://en.cppreference.com/w/cpp/language/pointer#Pointers_to_data_members](https://en.cppreference.com/w/cpp/language/pointer#Pointers_to_data_members)
```
auto L = [&](const std::string& name, std::string SourceFromPolicy::* member) { ... return true/false; }
if (!L("Name", &SourceFromPolicy::Name)) return {};
```

src/AppInstallerCommonCore/Registry.cpp
```diff
@@ -93,6 +177,87 @@ namespace AppInstaller::Registry
...
+            THROW_HR(E_UNEXPECTED);
+        }
+
+        m_value = ValueRef{ std::move(valueName), type, std::move(data) };
+    }
+
+    const ValueList::ValueRef& ValueList::const_iterator::operator*() const
+    {
+        return m_value.value();
+    }
```
JohnMcPMS, line 237: Not that it matters much for the general use case, but I would err toward making the caller keep the value while they wanted it rather than the iterator getting the value while iterating.  If I wanted to skip ahead for some reason this way reads every value anyway.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -49,6 +49,210 @@ namespace AppInstaller::Repository
             bool IsTombstone = false;
         };
 
+        namespace
+        {
```
JohnMcPMS, line 53: I don't think we need two nested, unnamed namespaces...

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -586,6 +809,10 @@ namespace AppInstaller::Repository
             case SourceOrigin::User:
                 m_sourceList.erase(FindSource(source.Name));
                 break;
+            case SourceOrigin::GroupPolicy:
+                // This should be blocked higher up for better reporting
```
JohnMcPMS, line 813: >// This should be blocked higher up for better reporting [](start = 16, length = 56)

But the comment in SourceCommand.cpp says that it should be done here...

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -709,6 +910,13 @@ namespace AppInstaller::Repository
         AICLI_LOG(Repo, Info, << "Source created with extra data: " << details.Data);
         AICLI_LOG(Repo, Info, << "Source created with identifier: " << details.Identifier);
 
+        if (!IsAddedUserSourceAllowedByPolicy(details))
```
JohnMcPMS, line 913: >if (!IsAddedUserSourceAllowedByPolicy(details)) [](start = 8, length = 47)

This check cannot be done before AddSourceFromDetails?  Type + Arg should be enough to block then since it is an allow list.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -49,6 +49,210 @@ namespace AppInstaller::Repository
...
+                    return std::nullopt;
+                }
+
+                const auto& sources = sourcesOpt->get();
+                auto source = std::find_if(
+                    sources.begin(),
+                    sources.end(),
+                    [&](const SourceFromPolicy& policySource)
+                    {
+                        return Utility::ICUCaseInsensitiveEquals(name, policySource.Name) && Utility::ICUCaseInsensitiveEquals(type, policySource.Type) && arg == policySource.Arg;
```
JohnMcPMS, line 104: >Utility::ICUCaseInsensitiveEquals(name, policySource.Name) [](start = 31, length = 58)

Should we really be requiring the name to match?  That seems restrictive.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -49,6 +49,210 @@ namespace AppInstaller::Repository
...
+                // We don't need to check sources added by policy as those have higher priority.
+                //
+                // Use the name and arg to match sources as we don't have the identifier before adding.
+
+                // Case 1:
+                //  - The source is not a tombstone and we don't need the policy to be explicitly enabled.
+                //  - Check only against the source argument as the user source may have a different name.
+                //  - Do a case insensitive check as the domain portion of the URL is case insensitive,
+                //    and we don't need case sensitivity for the rest as we control the domain.
+                if (Utility::CaseInsensitiveEquals(arg, s_Source_WingetCommunityDefault_Arg))
```
JohnMcPMS, line 143: >Utility::CaseInsensitiveEquals(arg, s_Source_WingetCommunityDefault_Arg) [](start = 20, length = 72)

It should be Type + Arg to ensure a match.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -49,6 +49,210 @@ namespace AppInstaller::Repository
...
+                //  3. The source has the same name as a default source that is explicitly enabled (to prevent shadowing)
+                //  4. Allowed sources are disabled, blocking all user sources
+                //  5. There is an explicit list of allowed sources and this source is not in it
+                //
+                // We don't need to check sources added by policy as those have higher priority.
+                //
+                // Use the name and arg to match sources as we don't have the identifier before adding.
+
+                // Case 1:
+                //  - The source is not a tombstone and we don't need the policy to be explicitly enabled.
```
JohnMcPMS, line 139: >The source is not a tombstone and we don't need the policy to be explicitly enabled. [](start = 22, length = 84)

There is no check on this to ensure that it is not a tombstone.

src/AppInstallerRepositoryCore/RepositorySource.cpp
lechacon: The policy includes the source data and identifier. (Because we need it for additional sources and it is easier to be consistent.)
Since we have the data and identifier, I think we might as well use it and verify that it actually matches. And that can only be after the factory extracts them.

I don't expect this check to fail often because I'm actually checking the Type + Arg before.

I'm making a lot of assumptions here (that we should require having the Data and Identifier, that it is possible they don't match even if the Type + Arg do, that we actually care if they don't, that we cannot verify it without adding the source). I'm happy to remove it if you don't think any of those is true.

JohnMcPMS: If the Type+Arg check is happening before we actually throw off to the source code then it should be fine.

src/AppInstallerRepositoryCore/RepositorySource.cpp
lechacon: For allowed sources, it may be restrictive but it may also be what the IT admin expects: that all the machines that have the policy and want to use the source will use the same name. E.g. so they can think in terms of "is the source foo configured?" instead of "is the REST source from foo.example.com configured?"

src/AppInstallerRepositoryCore/RepositorySource.cpp
lechacon: Uh... Do you mean "it should be fine" as in "the source should be fine and we don't need to check here" or as in "it should be fine to check here as long as we also do it before"?

JohnMcPMS: We don't need a check in this function as long as one is happening in the stack before ISourceFactory::Add.

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: True, and I suppose if they have a list for the employee of available sources they would provide the name.

JohnMcPMS (approved):
Approving with the assumption that changes to resolve the test failures will be minor.

lechacon:
Two tests failed. One was because in one test for failure adding a disabled default source I needed to add the Type in the test to fit the Arg+Type matching (where I only used the Arg before). The other was because I had a test for the Data+Identifier check after adding the source; I removed the test since that check was removed.

PR #845, opened 2021-04-02
Add docs on restoring AppInstaller

KevinLaMS:
This is draft.  We need to get URL to older package.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/845)

### DIFF
--- a/doc/restore.md
+++ b/doc/restore.md
@@ -1,0 +1,21 @@
+# Remove the Windows Package Manager preview
+
+Sometimes users trying the preview of the Windows Package Manager would like to restore their machine to a version that did not include the preview.  The following steps will restore your machine.
+
+1) Remove your machine from Windows Insider Flights.
+The Windows Package Manager is included in the Windows Insider Flights.  If you are installing a Windows Insider build, you will keep getting updated.
+2) Remove yourself from the Windows Package Manager flights.  
+If you requested to be added to the Windows Package Manager flights, then you will specifically need to ask to be removed.  You can ask to be removed by filling in the following form:
+
+Note: Make sure you use the correct email address.  It must be the address used to sign up for the Insider build.  If your email address is not the same address that you are using to log into the Windows Store, then removing it will have no impact.
+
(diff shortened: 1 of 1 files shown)


doc/restore.md
```diff
@@ -0,0 +1,21 @@
...
+1) Remove your machine from Windows Insider Flights.
+The Windows Package Manager is included in the Windows Insider Flights.  If you are installing a Windows Insider build, you will keep getting updated.
+2) Remove yourself from the Windows Package Manager flights.  
+If you requested to be added to the Windows Package Manager flights, then you will specifically need to ask to be removed.  You can ask to be removed by filling in the following form:
+
+Note: Make sure you use the correct email address.  It must be the address used to sign up for the Insider build.  If your email address is not the same address that you are using to log into the Windows Store, then removing it will have no impact.
+
+Please allow for 1 week, to be sure the form has been processed, though most likely it is processed much sooner.
+
+3) Install an older build of the [Desktop App Installer](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1).  We have placed an earlier released version [here](https://www.github.com/micrsoft/winget-cli/releases).
```
denelon, line 14: we should remove /en-us/ as the site handles localization.

jsoref:
Can you please rebase? It'll make the bot much happier.

doc/restore.md
lechacon, line 14: I think we don't want to use "here" for links? I read somewhere that it makes it harder to parse what the link refers to, both for people and tools like screen readers.

Maybe something like `Here is an [earlier released version for download](...).`

jsoref: I'd probably write `There is...` or `An [earlier version(...) is available for download.`

And yes, avoid linking `here` it's screen reader hostile.

JohnMcPMS:
I think the ship has sailed on this PR, as well as going back to older versions generally.

PR #846, opened 2021-04-03
Add None to Group Policy values enum

lechacon:
@JohnMcPMS suggested offline that the current code to get the Group Policy value associated with a setting was hard to read. Currently it's using SFINAE to pick whether the SettingMapping includes a Policy. He suggested adding a None to the policy enum whose value was always nullopt instead.

Here's my attempt at doing that. Feel free to reject if readability is actually worse.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/846)


JohnMcPMS (approved):
Not how I was expecting it to go, and not necessarily easier to follow.  The reader still has to note the conversion from `std::nullopt_t` to any `std::optional`, but that makes more sense than figuring out the SFINAE that was happening before (although I think a comment would have been enough to point it out [or lead someone to look up and understand SFINAE]).

Overall I prefer this as it doesn't rely on knowledge that a type definition is missing to drive the code path and the interesting way to effectively normalize `void` into the type system.

PR #848, opened 2021-04-05
Output sources from policy with --info

lechacon:
Adding the sources configured by Group Policy to the output of --info with the other policies. This is something I missed doing in #841 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/848)


JohnMcPMS (approved):
For future changes, it would be a good idea to include a sample command line output in the PR description.

PR #853, opened 2021-04-06
Using custom http stage for injecting tests responses

ashpatil-msft:
- Added optional stage to httpclienthelper and injected httpclienthelper into Interface.
- The optional stage can be used to intercept request message and return test responses in tests.
- Added a sample test doing the above.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/853)

### DIFF
--- a/src/AppInstallerRepositoryCore/Rest/RestClient.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/RestClient.cpp
@@ -41,11 +41,10 @@ namespace AppInstaller::Repository::Rest
         return endpoint;
     }
 
-    Version RestClient::GetSupportedVersion(const utility::string_t& restApi, const std::set<Version>& wingetSupportedVersions)
+    Version RestClient::GetSupportedVersion(const utility::string_t& restApi, const std::set<Version>& wingetSupportedVersions, HttpClientHelper&& clientHelper)
     {
         // Call information endpoint
-        HttpClientHelper httpClientHelper{ GetInformationEndpoint(restApi) };
-        std::optional<web::json::value> response = httpClientHelper.HandleGet();
+        std::optional<web::json::value> response = clientHelper.HandleGet(GetInformationEndpoint(restApi));
 
         THROW_HR_IF(APPINSTALLER_CLI_ERROR_UNSUPPORTED_RESTSOURCE, !response);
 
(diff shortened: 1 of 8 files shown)


src/AppInstallerRepositoryCore/Rest/RestClient.cpp
```diff
@@ -83,18 +82,19 @@ namespace AppInstaller::Repository::Rest
     {
         if (version == Version_0_2_0 || version == Version_1_0_0)
         {
-            return std::make_unique<Schema::V1_0::Interface>(api);
+            HttpClientHelper helper{};
```
JohnMcPMS, line 85: >HttpClientHelper helper{}; [](start = 12, length = 26)

This should be the one passed in to `Create`, no?

src/AppInstallerRepositoryCore/Rest/RestSourceFactory.cpp
```diff
@@ -19,7 +20,8 @@ namespace AppInstaller::Repository::Rest
             {
                 THROW_HR_IF(E_INVALIDARG, !Utility::CaseInsensitiveEquals(details.Type, RestSourceFactory::Type()));
 
-                RestClient restClient = RestClient::Create(details.Arg);
+                HttpClientHelper helper{};
+                RestClient restClient = RestClient::Create(details.Arg, std::move(helper));
```
JohnMcPMS, line 24: >std::move(helper) [](start = 72, length = 17)

If this is mostly for tests, either make a default value or an overload that doesn't require it.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.h
```diff
@@ -24,12 +25,14 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
         std::vector<Manifest::Manifest> GetManifests(const std::string& packageId, const std::map<std::string_view, std::string>& params = {}) const override;
    
     protected:
+        HttpClientHelper GetHttpClientHelper() const;
```
JohnMcPMS, line 28: >HttpClientHelper [](start = 8, length = 16)

&

PR #854, opened 2021-04-07
Add Group Policy E2E tests

lechacon:
Adding E2E tests for Group Policy. These tests read the ADMX to get the registry key path and the value names, and use that to set the policy. This ensures that the client code matches what the policy definition says.
The changes include a helper for reading the ADMX and set the policy, and tests that use the helper to set the policy and validate that the client respects it.

Note that running these tests will edit (and remove) the registry keys controlling the policy in the machine. Although this should be temporary if there is any actual GP set.

Also in this change:
* Updated version used of Microsoft.Msix.Utils. Only difference is that it targets .NET Standard instead of .NET Framework, so it doesn't produce a warning when building.
* Renamed some error codes in tests for consistency.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/854)

### DIFF
--- a/src/AppInstallerCLIE2ETests/GroupPolicy.cs
+++ b/src/AppInstallerCLIE2ETests/GroupPolicy.cs
@@ -1,12 +1,151 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using NUnit.Framework;
+
 namespace AppInstallerCLIE2ETests
 {
     /// <summary>
     /// Tests for enforcement of Group Policy.
+    /// Behavior is better tested in the unit tests; these tests mostly ensure match between the code and the definition.
     /// </summary>
-    public class GroupPolicy : BaseCommand
+    public class GroupPolicyTests : BaseCommand
     {
+        [SetUp]
+        public void Setup()
+        {
+            InitializeAllFeatures(false);
+            GroupPolicyHelper.DeleteExistingPolicies();
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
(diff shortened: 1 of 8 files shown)


JohnMcPMS (approved):

src/AppInstallerCLIE2ETests/GroupPolicy.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            InitializeAllFeatures(false);
+            GroupPolicyHelper.DeleteExistingPolicies();
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            InitializeAllFeatures(false);
+            GroupPolicyHelper.DeleteExistingPolicies();
```
JohnMcPMS, line 25: Post-approval comment: This could cause trouble if the user actually has policies that should be in place.  I don't know of a better way to do this other than product code changes though.

lechacon: The only trouble I think this could cause is leaving the policy not as it should be, or allowing to bypass the policy.

The configuration for registry-based policies is stored in a [Registry.pol file](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/Policy/registry-policy-file-format) in the system. I believe that if we modify the registry directly, that policy will at some point overwrite anything we change (but I'm not clear on the details), so it will all eventually be as it should.

And anybody running these tests could already have their version of `wingetdev` that ignores policy, so I don't think we need to worry about this opening the option to bypass it.

PR #856, opened 2021-04-08
Enable code analysis, sdl checks, binskim and fix related errors

yao-msft:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/856)


azure-pipelines.yml
```diff
@@ -322,7 +322,7 @@ jobs:
   - task: BinSkim@3
     displayName: 'Run BinSkim '
     inputs:
-      arguments: 'analyze "$(system.defaultWorkingDirectory)\src\$(buildPlatform)\Release\AppInstallerCLI\*.exe" "$(system.defaultWorkingDirectory)\src\$(buildPlatform)\Release\WinGetUtil\*.dll" --config default --recurse' 
+      arguments: 'analyze "$(system.defaultWorkingDirectory)\src\AppInstallerCLI*.exe" "$(system.defaultWorkingDirectory)\src\WinGetUtil*.dll" --config default --recurse' 
 
   - task: securedevelopmentteam.vss-secure-development-tools.build-task-publishsecurityanalysislogs.PublishSecurityAnalysisLogs@2
     displayName: 'Publish Security Analysis Logs'
```
jsoref, line 328: Can I convince you to add a newline at eof?

(and also one in ` .github/actions/spelling/patterns.txt` while you're at it)

src/cpprestsdk/cpprestsdk.vcxproj
```diff
@@ -317,8 +321,8 @@
     <ClCompile Include="cpprestsdk\Release\src\utilities\asyncrt_utils.cpp" />
     <ClCompile Include="cpprestsdk\Release\src\utilities\base64.cpp" />
     <ClCompile Include="cpprestsdk\Release\src\utilities\web_utilities.cpp" />
-  </ItemGroup>    
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
jsoref, line 328: And please put the one here back

yao-msft: Done

azure-pipelines.yml
yao-msft: Done, thanks.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -636,8 +636,8 @@ namespace AppInstaller::CLI::Workflow
                 changes.size(),
                 findByManifest.Matches.size(),
                 packagesInBoth.size(),
-                toLog ? static_cast<std::string_view>(toLog->GetProperty(PackageVersionProperty::Name)) : "",
-                toLog ? static_cast<std::string_view>(toLog->GetProperty(PackageVersionProperty::Version)) : "",
+                toLog ? static_cast<std::string>(toLog->GetProperty(PackageVersionProperty::Name)) : "",
```
JohnMcPMS, line 639: Why here but not the two lines below?  The difference is rvalue vs lvalue, but the lifetime of the rvalue is until the end of the statement.  So converting to string_view is just as valid.  This is either a mistake/overcaution in the analysis, or an error in my understanding.

JohnMcPMS (approved):

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
yao-msft: The code analysis error is C26449( [https://docs.microsoft.com/en-us/cpp/code-quality/c26449?view=msvc-160](https://docs.microsoft.com/en-us/cpp/code-quality/c26449?view=msvc-160) )
It specifically checks view created from temporary. It is most likely an overcaution because we immediately use the view within the statement. This is to make code analysis happy.

---
In reply to: [613644294](https://github.com/microsoft/winget-cli/pull/856#discussion_r613644294) [](ancestors = 613644294)

JohnMcPMS: I think the ternary is confusing it, since that link specifically calls out:
```
Temporaries created for function call arguments are not flagged. It is safe to pass spans from such temporaries if target functions don’t retain data pointers in external variables.
```

Issue #869, opened 2021-04-15
Provide Taskbar progress indicator in Terminal


denelon:
# Description of the new feature/enhancement

Provide OSC 9;4 sequence to provide status in Terminals for status.

Windows Terminal added support in https://github.com/microsoft/terminal/pull/8055 for https://github.com/microsoft/terminal/issues/3004 Taskbar progress indicator.

chausner:
This seems like a fun little feature I could potentially contribute, provided that @JohnMcPMS hasn't already started working on this. Does this need a spec written first?

denelon:
@chausner I see this as a small work item. I'd be open to a PR against this issue :)

We would love to have your support!

JohnMcPMS:
I have not, and I agree that it is both fun and should be a fairly small amount of work.

Watch out for how it handles cancellation with CTRL+C; a single signal should be able to end with the proper sequence but I don't know how the second one (where we just exit process) will be handled by Terminal.  Hopefully they track the source of this and end the progress when that process terminates.

I think we did want to get this in over the next two weeks, so please let us know if you've started it.

JohnMcPMS:
And since neither of us explicitly answered the spec question; no, I don't see this as needing a spec.

chausner:
Okay great, please assign me to this issue and I'll try to have a first implementation ready by next week.

> Watch out for how it handles cancellation with CTRL+C; a single signal should be able to end with the proper sequence but I don't know how the second one (where we just exit process) will be handled by Terminal. Hopefully they track the source of this and end the progress when that process terminates.

I did a quick test with
```
> python
print("\x1b]9;4;2;40\x1b\\")
quit()
```
and Terminal did not reset the progress on the task bar after the Python process exited. So when the winget process exits abnormally and if there is no event that would notify us before the process exit, there is probably no way to reset the progress. But normally, winget should not exit unexpectedly, right? And in the case of normal exit, we can always reset the progress.

JohnMcPMS:
The only "expected" exit would be the second CTRL+C, where we terminate the process because cooperative cancellation didn't happen fast enough for the user.  We could probably even handle it in that case, but only of there is a clean "turn off progress" sequence.  I think it is fine to leave that as future work since most places where we are in a progress state should also be the best at cancellation.

PR #883, opened 2021-04-21
Implement VT-based progress reporting

chausner:
This implements #869: if VT is enabled, OSC 9;4 sequences are printed in addition to the existing progress spinner and progress bar.

Whenever an indefinite spinner is printed, a corresponding OSC sequence with state "Indeterminate" is output. When the spinner is hidden, an OSC sequence to clear the progress is written. Analogously, OSC sequences for showing/updating/hiding determinate progress are printed.

I noticed that the implementation of the progress OSC sequences in Windows Terminal does not exactly match the implementation in ConEmu. In particular, ConEmu allows the progress value to be left out (e.g. to switch progress states without having to specify the current progress value again) but Windows Terminal does not recognize these sequences at the moment. It is not really an issue at all, there are just a couple of lines with workaround logic.

I added the new logic to the existing `IndefiniteSpinner` and `ProgressBar` classes, as opposed to creating a new class derived from `ProgressVisualizerBase`. The latter should work easily for determinate progress but for the spinner case, `IndefiniteSpinner` contains some logic with `Sleep` that would have to be duplicated if we want to separate the VT-logic into its own class, which is probably not so nice.

@JohnMcPMS [mentioned](https://github.com/microsoft/winget-cli/issues/869#issuecomment-824286503) potential issues with clearing the progress when a user forces a process exit by hitting Ctrl+C twice. I am able to reproduce this issue by running `winget source update` and quickly hitting Ctrl+C twice. In this case, the progress indicator in the taskbar is indeed not cleared but there are also other issues like the console foreground color not getting reset (the spinner uses a different color). So any fix for this issue should ideally not only reset the taskbar progress bar but also any other terminal state like colors.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/883)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionProgress.cpp
+++ b/src/AppInstallerCLICore/ExecutionProgress.cpp
@@ -193,7 +193,7 @@ namespace AppInstaller::CLI::Execution
             if (UseVT())
             {
                 // Additional VT-based progress reporting, for terminals that support it
-                m_out << Progress::ProgressBar(Progress::ProgressState::Indeterminate);
+                m_out << Progress::Construct(Progress::State::Indeterminate);
             }
 
             // Indent two spaces for the spinner, but three here so that we can overwrite it in the loop.
@@ -211,7 +211,7 @@ namespace AppInstaller::CLI::Execution
 
             if (UseVT())
             {
-                m_out << Progress::ProgressBar(Progress::ProgressState::None);
+                m_out << Progress::Construct(Progress::State::None);
             }
         }
 
@@ -257,7 +257,7 @@ namespace AppInstaller::CLI::Execution
                 // We always clear the VT-based progress bar, even if hideProgressWhenDone is false
                 // since it would be confusing for users if progress continues to be shown after winget exits
                 // (it is typically not automatically cleared by terminals on process exit)
-                m_out << Progress::ProgressBar(Progress::ProgressState::None);
+                m_out << Progress::Construct(Progress::State::None);
             }
 
             m_isVisible = false;
@@ -369,7 +369,7 @@ namespace AppInstaller::CLI::Execution
             }
 
             // Additional VT-based progress reporting, for terminals that support it
-            m_out << Progress::ProgressBar(Progress::ProgressState::Normal, static_cast<int>(percentage * 100));
+            m_out << Progress::Construct(Progress::State::Normal, static_cast<int>(percentage * 100));
         }
         else
         {
--- a/src/AppInstallerCLICore/ExecutionProgress.cpp
+++ b/src/AppInstallerCLICore/ExecutionProgress.cpp
@@ -19,7 +19,7 @@ namespace AppInstaller::CLI::Execution
 
         BytesFormatData s_bytesFormatData[] =
         {
-            // Multi-terabyate installers should be fairly rare for the foreseeable future...
+            // Multi-terabyte installers should be fairly rare for the foreseeable future...
             { 40, "TB"sv },
             { 30, "GB"sv },
             { 20, "MB"sv },
@@ -167,7 +167,7 @@ namespace AppInstaller::CLI::Execution
         if (!m_spinnerJob.valid() && !m_spinnerRunning && !m_canceled)
         {
             m_spinnerRunning = true;
-            m_spinnerJob = std::async(std::launch::async, (UseVT() ? &IndefiniteSpinner::ShowSpinnerInternalWithVT : &IndefiniteSpinner::ShowSpinnerInternalNoVT), this);
+            m_spinnerJob = std::async(std::launch::async, &IndefiniteSpinner::ShowSpinnerInternal, this);
         }
     }
 
@@ -180,7 +180,7 @@ namespace AppInstaller::CLI::Execution
         }
     }
 
-    void IndefiniteSpinner::ShowSpinnerInternalNoVT()
+    void IndefiniteSpinner::ShowSpinnerInternal()
     {
         char spinnerChars[] = { '-', '\\', '|', '/' };
 
@@ -188,26 +188,35 @@ namespace AppInstaller::CLI::Execution
         // showing anything, or a progress task to skip straight to progress.
         Sleep(100);
 
-        // Indent two spaces for the spinner, but three here so that we can overwrite it in the loop.
-        m_out << "   ";
-
-        for (size_t i = 0; !m_canceled; ++i)
+        if (!m_canceled)
         {
-            constexpr size_t repetitionCount = 20;
-            ApplyStyle(i % repetitionCount, repetitionCount, true);
-            m_out << '\b' << spinnerChars[i % ARRAYSIZE(spinnerChars)] << std::flush;
-            Sleep(250);
+            if (UseVT())
+            {
+                // Additional VT-based progress reporting, for terminals that support it
+                m_out << Progress::ProgressBar(Progress::ProgressState::Indeterminate);
+            }
+
+            // Indent two spaces for the spinner, but three here so that we can overwrite it in the loop.
+            m_out << "   ";
+
+            for (size_t i = 0; !m_canceled; ++i)
+            {
+                constexpr size_t repetitionCount = 20;
+                ApplyStyle(i % repetitionCount, repetitionCount, true);
+                m_out << '\b' << spinnerChars[i % ARRAYSIZE(spinnerChars)] << std::flush;
+                Sleep(250);
+            }
+
+            m_out << "\b \r";
+
+            if (UseVT())
+            {
+                m_out << Progress::ProgressBar(Progress::ProgressState::None);
+            }
         }
 
-        m_out << "\b \r";
         m_canceled = false;
         m_spinnerRunning = false;
-    }
-
-    void IndefiniteSpinner::ShowSpinnerInternalWithVT()
-    {
-        // Nothing special to do at the moment, can use NoVT version.
-        ShowSpinnerInternalNoVT();
     }
 
     void ProgressBar::ShowProgress(uint64_t current, uint64_t maximum, ProgressType type)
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLICore/VTSupport.h
```diff
@@ -133,6 +134,20 @@ namespace AppInstaller::CLI::VirtualTerminal
...
+        enum class ProgressState
+        {
+            None,
+            Indeterminate,
+            Normal,
+            Paused,
+            Error
+        };
+
+        ConstructedSequence ProgressBar(ProgressState state, std::optional<int> percentage = std::nullopt);
```
JohnMcPMS, line 149:
```suggestion
        ConstructedSequence Construct(State state, std::optional<uint32_t> percentage = std::nullopt);
```

To prevent confusion with `ProgressBar` in the `ExecutionProgress.cpp`, and reduce repetition again.  Also, since the value cannot be negative, let's just use a type without the range.

src/AppInstallerCLICore/VTSupport.cpp
```diff
@@ -164,4 +164,63 @@ namespace AppInstaller::CLI::VirtualTerminal
...
+                // Windows Terminal does not recognize the OSC sequence if the progress value is left out.
+                // As a workaround, we can specify an arbitrary value since it does not matter for None and Indeterminate states.
+                percentage = percentage.value_or(0);
+                break;
+            case ProgressState::Normal:
+            case ProgressState::Error:
+            case ProgressState::Paused:
+                // Windows Terminal does not support switching progress states without also setting a progress value at the same time,
+                // so we disallow this case for now.
+                THROW_HR_IF(E_UNEXPECTED, !percentage.has_value());
```
JohnMcPMS, line 190:
```suggestion
                THROW_HR_IF(E_INVALIDARG, !percentage.has_value());
```

src/AppInstallerCLICore/VTSupport.h
```diff
@@ -133,6 +134,20 @@ namespace AppInstaller::CLI::VirtualTerminal
         extern const Sequence EraseLineBackward;
         extern const Sequence EraseLineEntirely;
     }
+
+    namespace Progress
+    {
+        enum class ProgressState
```
JohnMcPMS, line 140:
```suggestion
        enum class State
```

To reduce repetition so that it is just `Progress::State`

src/AppInstallerCLICore/VTSupport.cpp
```diff
@@ -164,4 +164,63 @@ namespace AppInstaller::CLI::VirtualTerminal
...
+            case ProgressState::None:
+            case ProgressState::Indeterminate:
+                // Windows Terminal does not recognize the OSC sequence if the progress value is left out.
+                // As a workaround, we can specify an arbitrary value since it does not matter for None and Indeterminate states.
+                percentage = percentage.value_or(0);
+                break;
+            case ProgressState::Normal:
+            case ProgressState::Error:
+            case ProgressState::Paused:
+                // Windows Terminal does not support switching progress states without also setting a progress value at the same time,
```
JohnMcPMS, line 188: It seems like `Error` and `Paused` could also get the `value_or` treatment, since I don't think that there are any actual values that makes sense to pass in.

JohnMcPMS (commented):
I was promised this would take a week 😉 

This is exactly what I would have done, and I think will work well.  I do wonder though if we should actually set the indeterminate state as soon as we hit anything with progress, then not disable it until we exit.  By which I mean, rather than None right now, we would write Indeterminate (or nothing if we already know that is the state we are in), and have set a flag somewhere that on destruction/exit we write None to close it out.

I'm not sure what it looks like right now when it flips back and forth, or maybe those happen so fast that it isn't really noticeable.  That might just need to be something we give a chance and see if we need to improve on it.

src/AppInstallerCLICore/VTSupport.cpp
JohnMcPMS: Unless `Paused` is supposed to mean "paused at 50%", but I doubt that you are supposed to provide the HR to `Error`.

chausner: > It seems like Error and Paused could also get the value_or treatment, since I don't think that there are any actual values that makes sense to pass in.

I think values do make sense for Error and Paused in general (not speaking about winget right now where we don't even use these states at the moment). E.g. your progress might be at 60% when an error occurs. In this case you may want to set the progress to 100% and state to Error at the same time.

Using Error or Paused with value zero does not look good because the user would basically see an empty progress bar colored in red or yellow. Paused is certainly meant to be used together with values (e.g. hit pause on a Windows file copy dialog), and Error would typically be set to 100 when it occurs to indicate the end of the failed operation. At least that's how I've seen it being used in some applications.

src/AppInstallerCLICore/VTSupport.h
chausner: Applied your suggestion in c4d5c48a1c5a638d15025687a060f752bd5be3f0.

src/AppInstallerCLICore/VTSupport.h
chausner: Applied your suggestion in c4d5c48a1c5a638d15025687a060f752bd5be3f0.

chausner:
> I do wonder though if we should actually set the indeterminate state as soon as we hit anything with progress, then not disable it until we exit. By which I mean, rather than None right now, we would write Indeterminate (or nothing if we already know that is the state we are in), and have set a flag somewhere that on destruction/exit we write None to close it out.
> 
> I'm not sure what it looks like right now when it flips back and forth, or maybe those happen so fast that it isn't really noticeable. That might just need to be something we give a chance and see if we need to improve on it.

Good point. It seems it can look slightly distracting when it switches states multiple times in quick succession but most people probably wouldn't notice. Your proposal of always staying in the Indeterminate state would probably work, it just makes the code slightly less nice or might lead to issues later on (suppose winget adds some form of user interaction/confirmation in the future, if there was a spinner displayed before, it would continuously show the Intermediate state while waiting for user input. That would probably feel distracting.). 

So I suggest us to get a bit experience first how the current implementation feels in practice, then can decide if we need to tweak it or if it is fine as it is.

denelon:
I'll defer to @JohnMcPMS on the initial implementation. I'd be happy to solicit feedback from the community, and I can likely get @cinnamon-msft to help get feedback from Windows Terminal users. Just let me know what scenarios we want to ask for the community to test.

src/AppInstallerCLICore/VTSupport.cpp
JohnMcPMS: I see, so the state and progress amount are somewhat independent.

src/AppInstallerCLICore/ExecutionProgress.cpp
```diff
@@ -19,7 +19,7 @@ namespace AppInstaller::CLI::Execution
 
         BytesFormatData s_bytesFormatData[] =
         {
-            // Multi-terabyate installers should be fairly rare for the foreseeable future...
+            // Multi-terabyte installers should be fairly rare for the foreseeable future...
```
jsoref, line 22: Oops. Thanks for catching this. Initial spell-checking passes are best effort, and a goal is for people to incrementally improve.

JohnMcPMS (approved):
Thanks, I look forward to trying it out.

denelon:
@chausner I'm hoping to get this out in a build next week. I'll tweet about it as well. Would you like me to @mention you or not? 

If you would just let me know your Twitter handle.

chausner:
> @chausner I'm hoping to get this out in a build next week. I'll tweet about it as well.

Cool 👍 

> If you would just let me know your Twitter handle.

Thanks but I am not on Twitter. But feel free to mention my GitHub name in the next blog post in case you mention the feature there. :)

JohnMcPMS:
The only part of the build that failed was publishing the test result artifacts, so I'm going to go ahead and merge.

src/AppInstallerCLICore/VTSupport.cpp
```diff
@@ -164,4 +164,63 @@ namespace AppInstaller::CLI::VirtualTerminal
...
+            case State::None:
+            case State::Indeterminate:
+                // Windows Terminal does not recognize the OSC sequence if the progress value is left out.
+                // As a workaround, we can specify an arbitrary value since it does not matter for None and Indeterminate states.
+                percentage = percentage.value_or(0);
+                break;
+            case State::Normal:
+            case State::Error:
+            case State::Paused:
+                // Windows Terminal does not support switching progress states without also setting a progress value at the same time,
```
DHowett, line 188: Somebody shoulda filed this bug on us! 😉

src/AppInstallerCLICore/VTSupport.cpp
```diff
@@ -164,4 +164,63 @@ namespace AppInstaller::CLI::VirtualTerminal
...
+
+    namespace Progress
+    {
+        ConstructedSequence Construct(State state, std::optional<uint32_t> percentage)
+        {
+            // See https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
+
+            THROW_HR_IF(E_BOUNDS, percentage.has_value() && percentage > 100u);
+
+            // Workaround some quirks in the Windows Terminal implementation of the progress OSC sequence
```
DHowett, line 176: Y'all, we work at the same company. If I keep finding "quirk for Windows Terminal" comments I'm gonna have to glare at you from across the lunch room. 😁

JohnMcPMS: I will keep that in mind @DHowett , although this one is more of an external report through us, so maybe we can still be friends?

chausner: I am to blame for writing that mean comment, sorry. 😁 I wanted to check whether this was a known issue/limitation (I think I saw an issue with some follow-up tasks related to the initial implementation in Windows Terminal) but forget it in the end.

I see you haven't yet created an issue report so let me do it.

chausner: I created an issue for this in the Windows Terminal repo: https://github.com/microsoft/terminal/issues/9960

DHowett: Thanks a ton!

(sorry -- my joke probably came across very poorly over the medium of text. Thanks for being understanding! 😁)

PR #901, opened 2021-04-26
Rest source search result continuation token, PFN and product codes

ashpatil-msft:
- [ Y] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [ N] Are you working against an Issue?

-----

- Implemented logic to accept continuation token and maximum results in Rest source search method.
- Added usage of source identifier from Rest Information endpoint.
- Added Package family names and Product codes and implemented required methods that use these fields.
- Added check to verify search request package match fields and match types and fixed some other small Todos.
- Added unit tests for the above changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/901)

### DIFF
--- a/src/AppInstallerRepositoryCore/Rest/RestClient.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/RestClient.cpp
@@ -21,7 +21,7 @@ namespace AppInstaller::Repository::Rest
     std::set<Version> WingetSupportedContracts = { Version_0_2_0, Version_1_0_0 };
 
     RestClient::RestClient(std::unique_ptr<Schema::IRestClient> supportedInterface, std::string sourceIdentifier)
-        : m_interface(std::move(supportedInterface)), m_sourceIdentifier(sourceIdentifier)
+        : m_interface(std::move(supportedInterface)), m_sourceIdentifier(std::move(sourceIdentifier))
     {
     }
 
--- a/src/AppInstallerRepositoryCore/Rest/RestClient.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/RestClient.cpp
@@ -20,8 +20,8 @@ namespace AppInstaller::Repository::Rest
     // Supported versions
     std::set<Version> WingetSupportedContracts = { Version_0_2_0, Version_1_0_0 };
 
-    RestClient::RestClient(std::unique_ptr<Schema::IRestClient> supportedInterface)
-        : m_interface(std::move(supportedInterface))
+    RestClient::RestClient(std::unique_ptr<Schema::IRestClient> supportedInterface, std::string sourceIdentifier)
+        : m_interface(std::move(supportedInterface)), m_sourceIdentifier(sourceIdentifier)
     {
     }
 
@@ -35,13 +35,18 @@ namespace AppInstaller::Repository::Rest
         return m_interface->Search(request);
     }
 
+    std::string RestClient::GetSourceIdentifier() const
+    {
+        return m_sourceIdentifier;
+    }
+
     utility::string_t RestClient::GetInformationEndpoint(const utility::string_t& restApiUri)
     {
         utility::string_t endpoint = RestHelper::AppendPathToUri(restApiUri, JsonHelper::GetUtilityString(InformationGetEndpoint));
         return endpoint;
     }
 
-    Version RestClient::GetSupportedVersion(const utility::string_t& restApi, const std::set<Version>& wingetSupportedVersions, const HttpClientHelper& clientHelper)
+    IRestClient::Information RestClient::GetInformation(const utility::string_t& restApi, const HttpClientHelper& clientHelper)
     {
         // Call information endpoint
         std::optional<web::json::value> response = clientHelper.HandleGet(GetInformationEndpoint(restApi));
@@ -51,10 +56,7 @@ namespace AppInstaller::Repository::Rest
         Json::InformationResponseDeserializer responseDeserializer;
         IRestClient::Information information = responseDeserializer.Deserialize(response.value());
 
-        std::optional<Version> latestCommonVersion = GetLatestCommonVersion(information, wingetSupportedVersions);
-        THROW_HR_IF(APPINSTALLER_CLI_ERROR_UNSUPPORTED_RESTSOURCE, !latestCommonVersion);
-
-        return latestCommonVersion.value();
+        return information;
     }
 
     std::optional<Version> RestClient::GetLatestCommonVersion(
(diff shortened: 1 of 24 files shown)


src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/SearchRequestSerializer.cpp
```diff
@@ -20,6 +20,48 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0::Json
         constexpr std::string_view MatchType = "MatchType"sv;
         constexpr std::string_view PackageMatchField = "PackageMatchField"sv;
         constexpr std::string_view FetchAllManifests = "FetchAllManifests"sv;
+
+        std::optional<std::string_view> ConvertPackageMatchFieldToString(AppInstaller::Repository::PackageMatchField field)
+        {
+            // Rest API schema supports only a subset of fields.
+            switch (field)
```
ashpatil-msft, line 27: @JohnMcPMS - Should the rest source also support NormalizedNameAndPublisher and any other match types apart from the ones below?
I am skipping the search request items that contain the match fields or match types not in the Rest API schema. Is this okay? If not, what is the expected behavior on the client side?

JohnMcPMS: Yes, it must support NormalizedNameAndPublisher in order to allow for local packages to be correlated with the remote ones.  I don't see why we wouldn't support all of the `PackageMatchField` and `MatchType` values here.

Why would those values not be in the schema?  The remote side is not obliged to implement them, but if they cannot even be sent it will create issues if/when we do want to use them.

In general, the behavior should be "fail as early as possible unless there is a reason not to".

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -74,18 +73,47 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
...
+                size_t insertElements = !request.MaximumResults ? currentResult.Matches.size() :
+                    std::min(currentResult.Matches.size(), request.MaximumResults - results.Matches.size());
+
+                std::move(currentResult.Matches.begin(), std::next(currentResult.Matches.begin(), insertElements), std::inserter(results.Matches, results.Matches.end()));
+                ct = RestHelper::GetContinuationToken(jsonObject.value()).value_or(L"");
+            }
+
+            continuationToken = ct;
+
+        } while (!continuationToken.empty() && (!request.MaximumResults || results.Matches.size() < request.MaximumResults));
```
JohnMcPMS, line 109: This really suggests a need to refactor the interface to allow for an actual iteration of the results rather than a simple list containing all of them.  We need to make sure that the COM API is set up to handle that from the start.

@sreadingMSFT , should we change the `FindPackagesResult` to have an `IIterable` rather than an `IVectorView` for future consideration?

JohnMcPMS (approved):

src/AppInstallerRepositoryCore/Rest/RestClient.cpp
```diff
@@ -20,8 +20,8 @@ namespace AppInstaller::Repository::Rest
     // Supported versions
     std::set<Version> WingetSupportedContracts = { Version_0_2_0, Version_1_0_0 };
 
-    RestClient::RestClient(std::unique_ptr<Schema::IRestClient> supportedInterface)
-        : m_interface(std::move(supportedInterface))
+    RestClient::RestClient(std::unique_ptr<Schema::IRestClient> supportedInterface, std::string sourceIdentifier)
+        : m_interface(std::move(supportedInterface)), m_sourceIdentifier(sourceIdentifier)
```
yao-msft, line 24: >sourceIdentifier [](http://example.com/codeflow?start=73&length=16)

could std::move

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
```diff
@@ -144,8 +178,25 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
...
+                {
+                    if (!installer.PackageFamilyName.empty())
+                    {
+                        packageFamilyNames.emplace_back(installer.PackageFamilyName);
+                    }
+
+                    if (!installer.ProductCode.empty())
+                    {
+                        productCodes.emplace_back(installer.ProductCode);
+                    }
```
yao-msft, line 194: Do you need to check duplicates before add? Or you have dedupe logic later?

src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h
```diff
@@ -26,9 +26,11 @@ namespace AppInstaller::Repository::Rest::Schema
     {
         AppInstaller::Utility::VersionAndChannel VersionAndChannel;
         std::optional<Manifest::Manifest> Manifest;
+        std::vector<std::string> PackageFamilyNames;
+        std::vector<std::string> ProductCodes;
 
-        VersionInfo(AppInstaller::Utility::VersionAndChannel versionAndChannel, std::optional<Manifest::Manifest> manifest)
-            : VersionAndChannel(versionAndChannel), Manifest(manifest) {}
+        VersionInfo(AppInstaller::Utility::VersionAndChannel versionAndChannel, std::optional<Manifest::Manifest> manifest, std::vector<std::string> packageFamilyNames = {}, std::vector<std::string> productCodes = {})
+            : VersionAndChannel(versionAndChannel), Manifest(manifest), PackageFamilyNames(packageFamilyNames), ProductCodes(productCodes) {}
```
yao-msft, line 33: >VersionAndChannel(versionAndChannel), Manifest(manifest), PackageFamilyNames(packageFamilyNames), ProductCodes(productCodes) [](http://example.com/codeflow?start=14&length=124)

I think we can use std::move on these

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/ManifestDeserializer.cpp
```diff
@@ -82,6 +82,17 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0::Json
 
         constexpr std::string_view Capabilities = "Capabilities"sv;
         constexpr std::string_view RestrictedCapabilities = "RestrictedCapabilities"sv;
+
+        std::vector<Manifest::string_t> ConvertToManifestStringArray(const std::vector<std::string>& values)
+        {
+            std::vector<Manifest::string_t> result;
+            for (auto& value : values)
```
yao-msft, line 89: >auto [](http://example.com/codeflow?start=17&length=4)

auto const

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
yao-msft: Or maybe we don't care..

yao-msft:
>         : PackageIdentifier(packageIdentifier), PackageName(packageName), Publisher(publisher) {}

We might not catch these last time, looks like all structs in this file uses copy in constructor so they all can be moved

---
Refers to: src/AppInstallerRepositoryCore/Rest/Schema/IRestClient.h:22 in a43d242. [](commit_id = a43d242e63cbb66dec92f99ad1ecdc57b812dec0, deletion_comment = False)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
yao-msft: In GetMultiProperty() implementation in RestSource.cpp, there's no dedupe logic, I think it's better to dedupe here

ashpatil-msft: Added dedupe logic in the two places where pfn and product codes were added.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Interface.cpp
ashpatil-msft: Sounds reasonable. Let me know whenever this needs to be updated/refactored for COM Api's use.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/SearchRequestSerializer.cpp
ashpatil-msft: I have added the others as well.

PR #903, opened 2021-04-26
Multiple locale support for winget workflows

yao-msft:
## Change
This change implements support for multi locale in winget workflows.
- Exposed locale option in command line as a requirement
- Exposed locale option in winget settings as requirement/preference similar to how Scope works
- Installer selection uses both locale requirement or preference if applicable
- Manifest localization merging(for show) only use locale preference

## Validation
- Added tests for workflows and ManifestComparato
- Tried local run and works as expected

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/903)


doc/Settings.md
```diff
@@ -67,6 +67,20 @@ The `scope` behavior affects the choice between installing a package for the cur
     },
 ```
 
+### Locale
+
+The `locale` behavior affects the choice of installer based on installer locale. The matching parameter is `--locale`, and uses bcp47 language tag.
+
+Note the locale preference will also be applied to `winget show` command if applicable.
```
JohnMcPMS, line 74: Should they?  It feels like a separate setting should exist for display language.

src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -107,6 +108,14 @@ namespace AppInstaller::CLI
                 throw CommandException(Resource::String::InvalidArgumentValueError, s_ArgumentName_Scope, { "user"_lis, "machine"_lis });
             }
         }
+
+        if (execArgs.Contains(Args::Type::Locale))
+        {
+            if (!Utility::IsWellFormedBcp47Tag(execArgs.GetArg(Args::Type::Locale)))
+            {
+                throw CommandException(Resource::String::InvalidArgumentValueError, s_ArgumentName_Locale, { "bcp47 language tags"_lis });
```
JohnMcPMS, line 116: >s_ArgumentName_Locale [](http://example.com/codeflow?start=84&length=21)

If the Arg is defined in another file, this value should come from finding the Arg and using it's name.

JohnMcPMS, line 116: >"bcp47 language tags" [](http://example.com/codeflow?start=109&length=21)

This is not a localization independent string and it wouldn't fit well into this exception constructor anyway.  I think it should just be dropped so the user gets "invalid arg: locale" and can go look at the help on that.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -298,8 +298,8 @@ They can be configured through the settings file 'winget settings'.</value>
   <data name="InvalidNameError" xml:space="preserve">
     <value>Argument name was not recognized for the current command</value>
   </data>
-  <data name="LanguageArgumentDescription" xml:space="preserve">
-    <value>Language to install (if supported)</value>
+  <data name="LocaleArgumentDescription" xml:space="preserve">
+    <value>Locale to install (if supported)</value>
```
JohnMcPMS, line 302: >Locale to install (if supported) [](http://example.com/codeflow?start=11&length=32)

What about: `Locale to use (BCP 47 format)`

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -216,6 +216,134 @@ namespace AppInstaller::CLI::Workflow
...
+
+                // Preference will come from winget settings or Preferred Languages settings. winget settings takes precedence.
+                preference = Settings::User().Get<Settings::Setting::InstallLocalePreference>();
+                if (preference.empty())
+                {
+                    auto preferredList = Utility::GetUserPreferredLanguages();
+                    if (!preferredList.empty())
+                    {
+                        // TODO: we only take the first one for now
+                        preference = preferredList.at(0);
```
JohnMcPMS, line 282: This suggests that the setting should also be an array.  What are the complications with taking the whole list?  The only one that comes to mind is that the sort is harder, but I don't think that is insurmountable.

src/AppInstallerCommonCore/Public/winget/Locale.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include <string>
+#include <vector>
+
+namespace AppInstaller::Utility
```
JohnMcPMS, line 7: >Utility [](http://example.com/codeflow?start=24&length=7)

Locale?

src/AppInstallerCommonCore/Manifest/Manifest.cpp
```diff
@@ -2,13 +2,53 @@
...
+        if (targetLocale.empty())
+        {
+            targetLocale = Settings::User().Get<Settings::Setting::InstallLocalePreference>();
+            if (targetLocale.empty())
+            {
+                auto preferredList = Utility::GetUserPreferredLanguages();
+                if (!preferredList.empty())
+                {
+                    // TODO: we only take the first one for now
+                    targetLocale = preferredList.at(0);
```
JohnMcPMS, line 25: >targetLocale = preferredList.at(0); [](http://example.com/codeflow?start=20&length=35)

As discussed, we should use all of these languages and only these for now.

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.h
```diff
@@ -15,42 +15,42 @@ namespace AppInstaller::Repository::Microsoft
     struct ARPHelper
     {
         // See https://docs.microsoft.com/en-us/windows/win32/msi/uninstall-registry-key for details.
-        std::wstring SubKeyPath{ L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall" };
+        inline static const std::wstring SubKeyPath{ L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall" };
```
JohnMcPMS, line 18: >static [](http://example.com/codeflow?start=15&length=6)

Please revert to instance.

JohnMcPMS:
>                 result += m_requirement;

is not one of: en-us, es-mx, etc. ?

---
Refers to: src/AppInstallerCLICore/Workflows/ManifestComparator.cpp:316 in aa5db9b. [](commit_id = aa5db9b7230b7110bd9103b453aacd49d078896b, deletion_comment = True)

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -216,6 +216,140 @@ namespace AppInstaller::CLI::Workflow
...
+                {
+                    return false;
+                }
+
+                for (auto const& preferredLocale : m_preference)
+                {
+                    double firstScore = first.Locale.empty() ? Locale::UnknownLanguageDistanceScore : Locale::GetDistanceOfLanguage(preferredLocale, first.Locale);
+                    double secondScore = second.Locale.empty() ? Locale::UnknownLanguageDistanceScore : Locale::GetDistanceOfLanguage(preferredLocale, second.Locale);
+
+                    if (firstScore > secondScore && firstScore >= Locale::MinimumDistanceScoreAsCompatibleMatch)
```
JohnMcPMS, line 338: >if (firstScore > secondScore && firstScore >= Locale::MinimumDistanceScoreAsCompatibleMatch) [](http://example.com/codeflow?start=20&length=92)

The search through m_preference needs to stop whenever either score is compatible.  So:
```
if (firstScore >= Compat || secondScore => Compat)
  return firstScore > secondScore;
```
This is required or you could have IsFirstBetter return true for both orderings of first and second.

JohnMcPMS (approved):
:shipit:

Issue #151, opened 2020-05-15
Support Delivery Optimization


denelon:
Some enterprise tools integrate with delivery optimization (DO) to minimize network traffic.  We need to be good citizen and be DO aware.

As an enterprise administrator I want to have the benefit of reduced traffic to the Internet so I can reduce the cost and load on my network.

# Proposed technical implementation details (optional)

Provide the ability to use DO when downloading resources from the Internet when using the Windows Package Manager.

Edited: description.

jszabo98:
What other enterprise tools integrate with delivery optimization?

denelon:
The reference to #528 is a reminder to look at eliminating additional redundant downloads of the source based on different per user TTL values.

skycommand:
> Most enterprise tools integrate with delivery optimization (DO)

"Most enterprise tools"? Really? If you don't mind, please name them all.

ecovio1:
@skycommand 

Is there any need to be offensive ? that question doesn't like a sincere one.
think of that line as a justification for feature efforts and that's a good sign.

denelon:
@skycommand I edited the description. as @ecovio1 mentioned, it was an initial justification for the feature. I was multi-tasking during the few days ahead of the preview announcement, and didn't always have as much time or focus on crisp descriptions as I would have liked.

skycommand:
@denelon Thanks. I guess I was excited for no reason. And to make matter worse, my excitement led @ecovio1 to assume bad faith.

Bummer.

denelon:
@skycommand no harm was done :smile:. It's hard to convey intent via text, and English is a particularly messy and complicated language. I am always looking for ways to improve the product and to try and meet people where they are. It's also hard sometimes to remember humans are on the other end of the Internet. The Windows Package Manager is a very exciting product I have the privilege to be able to work on. I look forward to any suggestions you may have or other features that would make things better.

skycommand:
Hmm... Your last post got me thinking and I've formed an initial idea. I think adding Delivery Optimization support has no cons. It's a positive. But I think it is ineffective.

Consider the case of delivering Windows updates via Delivery Optimization:

- Each update is no bigger than 1 GB each month.
- Updates are cumulative, so delivering one (the latest) is enough.
- All Windows 10 computers in one site need them.
- Even if the site computers don't need it, the rest of the Internet can benefit from it. Windows 10 has a large installation base.

Now, the same thing cannot be said about the delivery of large apps, like Blender:

[...]

PR #909, opened 2021-04-29
Implement Delivery Optimization

JohnMcPMS:
Initial work on #151 

## Change
This change adds code to leverage the Delivery Optimization (DO) service to download packages.  Only packages are downloaded through DO at this time, as other things that we download do not fit in with the design of DO.

Currently one must still opt-in to using DO via settings, which I used rather than an experimental feature because it makes sense to allow the control long term.  Eventually, we will make DO the default value here.

## Settings
Two settings are added:
1. Control over which downloader is used for packages (`.network.downloader`)
2. A timeout for controlling how long to wait before DO indicates any progress has been made (`.network.doProgressTimeoutInSeconds`)

## Validation
Only manual validation has been done currently, but when we move DO to be the default it will automatically be tested by the E2E tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/909)


src/AppInstallerCommonCore/DODownloader.cpp
```diff
@@ -0,0 +1,395 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "DODownloader.h"
+#include "Public/AppInstallerLogging.h"
+#include "Public/AppInstallerSHA256.h"
+#include "Public/AppInstallerStrings.h"
+#include "winget/UserSettings.h"
+
+// Until it is in the Windows SDK, get it from here
```
yao-msft, line 10: >Until [](http://example.com/codeflow?start=3&length=5)

a TODO: for tracking?

src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -128,11 +128,25 @@ namespace AppInstaller::Utility
         return result;
     }
 
+    std::optional<std::vector<BYTE>> DownloadToStream(
```
yao-msft, line 131: >DownloadToStream [](http://example.com/codeflow?start=37&length=16)

Do we expect to make this use DO in the future, currently it's only used by manifest downloading and this extra level of wrapping seems unnecessary.

src/AppInstallerCommonCore/Public/AppInstallerDownloader.h
```diff
@@ -31,8 +43,10 @@ namespace AppInstaller::Utility
     std::optional<std::vector<BYTE>> Download(
         const std::string& url,
         const std::filesystem::path& dest,
+        DownloadType type,
         IProgressCallback& progress,
-        bool computeHash = false);
+        bool computeHash = false,
+        std::string_view downloadIdentifier = {});
```
yao-msft, line 49: >downloadIdentifier [](http://example.com/codeflow?start=25&length=18)

Should we add this to the above comments? Something like, only used in DO as content id

src/AppInstallerCommonCore/Downloader.cpp
JohnMcPMS: Maybe; I had originally implemented the split here, but I did not understand how to make DO do an in memory download of the full file.  I do now, but I don't see a reason to change back here.  I do want to have the type continue to come in to the download though, so we can properly categorize them.

.github/actions/spelling/expect.txt
```diff
@@ -6,6 +6,7 @@ agg
 aicli
 AICLIC
 ajor
+ALLOC
```
yao-msft, line 9: We should probably put these in allow.txt

JohnMcPMS: The only difference I know of is expect == warning and allow == no warning.  But I have never seen these warnings.

src/AppInstallerCommonCore/DODownloader.cpp
```diff
@@ -0,0 +1,395 @@
...
+                }
+
+                return false;
+            }
+
+        private:
+            IProgressCallback& m_progress;
+            std::mutex m_statusMutex;
+            std::condition_variable m_statusCV;
+            DO_DOWNLOAD_STATUS m_currentStatus{};
```
yao-msft, line 331: >{} [](http://example.com/codeflow?start=46&length=2)

 = {} for code style consistency?

yao-msft (approved):
:shipit:

.github/actions/spelling/expect.txt
jsoref: You'll see them if an item is no longer used and someone adds an additional misspelling.

Since these are terms, it's reasonable to move them to `allow`.

`ajor` for instance is a good example of something that should be in `expect`. It's there because you're currently using `[M]ajor`, but `ajor` isn't really a word, so if you rewrite `[M]ajor` some other way (or mask it w/ a pattern), you wouldn't want someone else to accidentally write `ajor` (as a typo by dropping the `m`).

You could for instance change:
https://github.com/microsoft/winget-cli/blob/741955e48ce2f069ab323f71e40dd1326654e7ab/src/AppInstallerRepositoryCore/Microsoft/ARPHelper.h#L26
to:
```
        // REG_DWORD (ex. 0xMMmmbbbb, MM: major, mm: minor, bbbb: build)
```
At which point the bot would tell you it noticed `ajor` is no longer present (when it notices someone adding a new unrecognized word, which happens often enough).

src/AppInstallerCommonCore/DODownloader.cpp
```diff
@@ -0,0 +1,395 @@
...
+            {
+                SetProperty(DODownloadProperty_Uri, uri);
+            }
+
+            void ContentId(std::string_view contentId)
+            {
+                SetProperty(DODownloadProperty_ContentId, contentId);
+            }
+
+            void DisplayName(std::string_view displayName)
```
shishirb-MSFT, line 113: >DisplayName [](http://example.com/codeflow?start=17&length=11)

Can you please add a call to this and specify a name? We suggest "MS WinGet Download".

src/AppInstallerCommonCore/DODownloader.cpp
```diff
@@ -0,0 +1,395 @@
...
+            {
+                SetProperty(DODownloadProperty_LocalPath, localPath.wstring());
+            }
+
+            void CorrelationVector(std::string_view correlationVector)
+            {
+                SetProperty(DODownloadProperty_CorrelationVector, correlationVector);
+            }
+
+            void NoProgressTimeoutSeconds(uint32_t noProgressTimeoutSeconds)
```
shishirb-MSFT, line 128: >NoProgressTimeoutSeconds [](http://example.com/codeflow?start=17&length=24)

Unused. Remove it to avoid someone from introducing a call while still implementing the timeout in DODownloadStatusCallback::Wait?

shishirb-MSFT: Ah, I see Wait() actually waits indefinitely after the first progress update. Keep in mind DO's default timeouts are 30 minutes for foreground, 3 hours for background.

Are you ok with these defaults?

Issue #163, opened 2020-05-15
Support for package dependencies


denelon:
# Description of the new feature/enhancement

Some packages like IDEs require the language to be installed separately. Most installers take care of their own dependencies, but there should be a way to support this situation. Maybe it's just a package collection, but we would still need to figure out the right order to install packages like this

# Proposed technical implementation details (optional)

https://github.com/microsoft/winget-cli/blob/master/doc/specs/%23163%20-%20Dependencies.md

---

Edited: Added link to specification document.

Conan-Kudo:
You should consider using [openSUSE/libsolv](https://github.com/openSUSE/libsolv) for this. It's a battle-tested dependency resolver library, used by [DNF](https://dnf.readthedocs.io/en/latest/), Zypper, opkg, mamba, and other package managers.

adelcast:
Couldn't agree more with @Conan-Kudo. I currently maintain opkg, a tiny package manager used for embedded devices. It used to rely on its own dependency engine, which was the most complicated/buggy code in the project. We added support for libsolv, which smoothed things out a lot (speed them up too, since libsolv is a really fast SAT solver). Now dependency bugs are a tiny fraction of the overall bug backlog.

Another benefit is that by sharing the same dependency model as rpm/dpkg (Recommends, Depends, Suggest, Conflicts, etc) you get a model that has been successful for 20+ years, so there is no need to reinvent the wheel. It will also bring winget closer to other package managers that already have large tooling around them. Some of that tooling could then be leveraged by winget. Certainly an option worth giving some serious thought.

adamroach:
Another example of a dependency I would expect `winget` to understand is: `powertools` depends on `.NET Core` being installed, but `winget` will happily install `powertools` in what is effectively an unusable state (since it doesn't install `.NET Core` automatically).

dodgepong:
Another example would be any GPL application that depends on VC++ runtimes, such as [OBS Studio](https://github.com/microsoft/winget-pkgs/blob/master/manifests/OBSProject/OBSStudio/25.0.8.yaml). The GUI installer for OBS prompts the user to go download the VC++ redists if they are not detected on the machine, but when installed via `winget`, since the installer is run with the `/S` flag, that prompt is never shown to the user.

GPL applications are allowed to link against VC++ runtimes via the System Library Exception, but are forbidden from distributing those libraries, or else the exception does not apply.

denelon:
[...]

PR #954, opened 2021-05-10
Create #163 - Dependencies.md

denelon:

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/954)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -75,6 +75,7 @@ deigh
 deleteifnotneeded
 dirs
 dnld
+dustojnikhummer
 dvinns
 dw
 endian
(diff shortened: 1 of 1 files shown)


jsoref:
Fwiw, the latest version will do a better job of annotating misspellings (you'll get 10 instead of duplicates).

JohnMcPMS (approved):
I think this is a fine starting point, but probably needs more details added.  One thing that would be good to add now though would be what we are not planning to do.  For instance, there has been discussion around a produces/consumes model as well, where a package could state that it produces "JRE" and another that it consumes "JRE", but I don't think that we are planning on doing that here.

denelon:
@jsoref some of the misspellings appear to be part of URLs. Is that expected?

jsoref:
@denelon: people make typos in urls, yes :-).

Also, a word that appears in a url if it's real is likely to appear elsewhere.

PR #955, opened 2021-05-11
Name an Publisher correlation in CompositeSource

JohnMcPMS:
## Change
This change leverages the normalized name and publisher to increase the number of packages that can be matched.  It also reworks the way the the correlation occurs internally, albeit only slightly.

All operations that rely on this data should now work better; this includes `import`, `list`, `upgrade`, and `uninstall`.

## Validation
The existing tests cover a lot of the scenarios, and a test is added to ensure that we do in fact send the name and publisher along in the search.  In large part though, actually matching based on this data is a source responsibility.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/955)

### DIFF
--- a/src/AppInstallerRepositoryCore/Rest/RestSource.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/RestSource.cpp
@@ -34,7 +34,7 @@ namespace AppInstaller::Repository::Rest
         {
             PackageVersion(
                 const std::shared_ptr<const RestSource>& source, IRestClient::PackageInfo packageInfo, IRestClient::VersionInfo versionInfo)
-                : SourceReference(source), m_packageInfo(packageInfo), m_versionInfo(versionInfo) {}
+                : SourceReference(source), m_packageInfo(std::move(packageInfo)), m_versionInfo(std::move(versionInfo)) {}
 
             // Inherited via IPackageVersion
             Utility::LocIndString GetProperty(PackageVersionProperty property) const override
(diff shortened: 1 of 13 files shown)


src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -114,6 +144,24 @@ namespace AppInstaller::Repository::Rest
...
+            void BuildPackageVersionMultiPropertyWithFallback(std::vector<Utility::LocIndString>& result) const
+            {
+                result.emplace_back(m_versionInfo.Manifest->DefaultLocalization.Get<Field>());
+                for (const auto& loc : m_versionInfo.Manifest->Localizations)
+                {
+                    auto f = loc.Get<Field>();
+                    if (f.empty())
+                    {
+                        result.emplace_back(loc.Get<Field>());
+                    }
```
yao-msft, line 157: We should just ignore if empty?

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -114,6 +144,24 @@ namespace AppInstaller::Repository::Rest
...
+                    else
+                    {
+                        result.emplace_back(std::move(f));
```
yao-msft, line 160: >result.emplace_back(std::move(f)); [](http://example.com/codeflow?start=24&length=34)

I think we should dedupe here

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
yao-msft: I saw what you're trying to do in PackageVersionMultiProperty comments

yao-msft (approved):
:shipit:

PR #959, opened 2021-05-13
Ensure chain of trust with packaged index

JohnMcPMS:
## Change
This change better ensures the chain of trust when using a packaged index by:
1. Ensuring that the index package does not change family names
2. Ensuring that the default source is Store signed
3. Using signed data from the package to verify the manifest files that we download separately

## Validation
New tests are added, but many existing tests also covered the scenarios.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/959)


src/AppInstallerCommonCore/Public/winget/Manifest.h
```diff
@@ -43,5 +44,8 @@ namespace AppInstaller::Manifest
 
         // Get all commands across installers
         std::vector<string_t> GetAggregatedCommands() const;
+
+        // If not empty, the SHA256 hash of the manifest stream itself.
+        Utility::SHA256::HashBuffer StreamSHA256;
```
yao-msft, line 49: >SHA256 [](http://example.com/codeflow?start=42&length=6)

nit: we use Sha256 across the manifest

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -90,21 +90,35 @@ namespace AppInstaller::Repository::Microsoft
...
+                }
 
                 auto lock = Synchronization::CrossProcessReaderWriteLock::LockForWrite(CreateNameForCPRWL(details));
 
-                UpdateInternal(packageLocation, details, progress);
+                UpdateInternal(packageLocation, packageInfo, details, progress);
             }
 
-            virtual void UpdateInternal(std::string packageLocation, const SourceDetails& details, IProgressCallback& progress) = 0;
+            virtual void UpdateInternal(const std::string& packageLocation, Msix::MsixInfo& packageInfo, const SourceDetails& details, IProgressCallback& progress) = 0;
```
yao-msft, line 121: >Msix::MsixInfo& [](http://example.com/codeflow?start=76&length=15)

const?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_3/Interface_1_3.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    std::optional<std::string> Interface::GetPropertyByManifestIdInternal(const SQLite::Connection& connection, SQLite::rowid_t manifestId, PackageVersionProperty property) const
+    {
+        switch (property)
+        {
+        case AppInstaller::Repository::PackageVersionProperty::ManifestSHA256Hash:
+        {
+            SQLite::blob_t hash = std::get<0>(V1_0::ManifestTable::GetIdsById<HashVirtualTable>(connection, manifestId));
+            return hash.empty() ? std::optional<std::string>{} : Utility::SHA256::ConvertToString(hash);
```
yao-msft, line 86: >std::optional<std::string>{} [](http://example.com/codeflow?start=34&length=28)

{}

yao-msft (approved):
:shipit:

src/AppInstallerRepositoryCore/Microsoft/Schema/1_3/Interface_1_3.cpp
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "Microsoft/Schema/1_3/Interface.h"
+
+#include "Microsoft/Schema/1_0/ManifestTable.h"
+
+#include "Microsoft/Schema/1_3/HashVirtualTable.h"
+
```
yao-msft, line 9: nit: extra empty lines

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
JohnMcPMS: Not without changing all of the methods to also be const correct.  And as this is internal to it I'm not concerned about it not being const.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_3/Interface_1_3.cpp
JohnMcPMS: I didn't actually try that, but I suspect that would create an empty string to match the return type of the other half of the ternary.  Then instead of an empty optional, we would return a filled optional with an empty string, which is not what we want.

PR #968, opened 2021-05-17
Create v2.0 packages schema with changes requested in #940

### LINKED ISSUE #940: Updates to packages.json schema.

The packages.json schema shouldn't need the "WinGetVersion" as a required attribute since the schema itself is versioned.

The "CreationDate" should include an example for the proper format.

The "Id" field in the "Packages" node should be changed to "PackageIdentifier" so it conforms to the other schemas.

Update the packages.json schema and documentation.

Thanks @MehediH for the suggestions.

JohnMcPMS:
Fixes #940 

## Change
Create a new v2.0 packages schema as the changes required a breaking version (renaming package identifier from `Id` to `PackageIdentifier`).  It also makes `WinGetVersion` optional, and adds a comment to `CreationDate` with an example of the format (to both 1.0 and 2.0 schemas).

The current pattern is that winget will only ever produce the latest schema.  If we decide to allow a command line option to select the output format, it should be a simple change to be able to produce 1.0 files in the future.

Also fixes a bug in the 1.0 schema that was not actually requiring the `Id` property to be present.  Packages without an `Id` would still pass schema validation, but they would not function as the code would be looking for a package with an empty string as the identifier.  Although this is technically a breaking change in terms of what would and wouldn't pass validation, since it does not actually change what will and won't function I'm not treating it as anything more than a bug fix.

## Validation
Updates the unit and functional tests to include packages with the new schema.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/968)


yao-msft (approved):
:shipit:

PR #969, opened 2021-05-17
Make a lot of experimental things release

JohnMcPMS:
## Change
Make `list`, `upgrade`, `uninstall`, and `export` commands no longer experimental.  Also make the REST source implementation not experimental, as well as making Delivery Optimization the default downloader.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/969)

### DIFF
--- a/src/AppInstallerCLIE2ETests/ImportCommand.cs
+++ b/src/AppInstallerCLIE2ETests/ImportCommand.cs
@@ -11,15 +11,12 @@ namespace AppInstallerCLIE2ETests
         [SetUp]
         public void Setup()
         {
-            InitializeAllFeatures(false);
-            ConfigureFeature("export", true);
             CleanupTestExe();
         }
 
         [TearDown]
         public void TearDown()
         {
-            InitializeAllFeatures(false);
         }
 
         [Test]
@@ -119,7 +116,6 @@ namespace AppInstallerCLIE2ETests
 
         private void UninstallTestExe()
         {
-            ConfigureFeature("uninstall", true);
             TestCommon.RunAICLICommand("uninstall", Constants.ExeInstallerPackageId);
         }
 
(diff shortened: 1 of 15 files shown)


src/AppInstallerCLIE2ETests/ImportCommand.cs
```diff
@@ -11,15 +11,12 @@ public class ImportCommand : BaseCommand
         [SetUp]
         public void Setup()
         {
-            InitializeAllFeatures(false);
-            ConfigureFeature("export", true);
             CleanupTestExe();
         }
 
         [TearDown]
         public void TearDown()
```
yao-msft, line 18: >public void TearDown() [](http://example.com/codeflow?start=8&length=22)

Should we just delete the tear down?

yao-msft: Applies to the other 2 tests' setup and teardown if you decide to do so

yao-msft (approved):
:shipit:

src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -164,18 +164,21 @@ namespace AppInstaller::Utility
...
+                if (std::filesystem::exists(dest))
+                {
+                    ApplyMotwIfApplicable(dest, URLZONE_INTERNET);
+                }
+                return result;
 
-                // While DO still requires an explicit opt-in, we will let failures through.
-                // When DO becomes the default, we may choose to catch exceptions and fall back to WinINet below.
+                // If DO becomes an issue, we may choose to catch exceptions and fall back to WinINet below.
+                // We would need to be careful not to bypass metered networks or other reasons that might
```
yao-msft, line 180: >other reasons [](http://example.com/codeflow?start=81&length=13)

Curious what the other reasons could be?
Is metered networks the reason we do not fall back to wininet for now?

PR #973, opened 2021-05-20
Fixes for a few crashes, change DO timeout to 60 seconds default, update version that I forgot

JohnMcPMS:
## Change
Fixes a few crashes that were happening, increase the default DO timeout for no progress to 60 seconds, and bump the minor version for the previous command releases.

## Validation
Force the telemetry event code to run by changing the global value to true.  This will cause more code to be exercised by the existing tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/973)


yao-msft (approved):
:shipit:

PR #889, opened 2021-04-22
Draft API proposal for #888 Support for UWP callers and a non command line interface API

sreadingMSFT:
I'll be sharing this with the API design review team and will update this when that review has a scheduled date.

Issue #888, opened 2021-04-22
Support for UWP callers and a non command line interface API


sreadingMSFT:
# Description of the new feature/enhancement

The Windows Package Manager currently exposes a command line interface to search for packages, install them, view progress, and more. This request is for a COM API so that callers have a more programmable way to make use of that functionality. The COM API would be preferred by callers that want to receive progress and completion events, and UWP packages that do not have permission to launch command line processes.

# Proposed technical implementation details (optional)

The winget package could expose an out of proc COM server with an API surface that allows install of packages.
UWP apps and other callers would use the COM server to activate the objects.
The calling apps would include the winmd of the API in their project in order to handle marshalling data between processes.

sreadingMSFT:
I added a draft proposal here https://github.com/microsoft/winget-cli/pull/889 . I do not believe I have permission to link the draft pull request to this item.

jantari:
Sorry for the partly unrelated question, but I am not super familiar with COM and many others will possibly wonder the same thing:

Will this automatically enable remote activation of the interface via DCOM or would you have to specifically support DCOM?

denelon:
@doctordns this is some of the required work to build PowerShell cmdlets.

sreadingMSFT:
@jantari The proposed feature wouldn't have DCOM support automatically in the sense that people asking that question would mean. DCOM at some point did refer to any out of process scenario such that COM over RPC could have been called DCOM. But since it was introduced the cross process stuff is now so built in that most everyone would just consider that COM which leaves the cross machine scenarios as what everyone would mean when asking about DCOM. Those scenarios are more complex than regular COM and now pretty unusual. For cross-machine scenarios I would speculate you'd be more likely to get support by a CSP feature request for something like https://docs.microsoft.com/en-us/windows/client-management/mdm/enterprisemodernappmanagement-csp but I have no insight into whether anything like that is planned. If you follow through to the draft proposal you can see that the proposal is an out of proc com server that can return winrt objects to workaround Windows limitations for packaged com (no out of proc winrt servers), so not intended to support remote scenarios.

jantari:
[...]

PR #984, opened 2021-05-24
Add support for callers using programmatic, non command line interface #888

sreadingMSFT:
This is the implementation for the spec at: https://github.com/microsoft/winget-cli/pull/889 for issue https://github.com/microsoft/winget-cli/issues/888 "Support for UWP callers and a non command line interface API"

As proposed:
The winget package exposes an out of proc COM server with an API surface that allows install of packages.
UWP apps and other callers can use the COM server to activate the objects.
The calling apps need to include the winmd of the API in their project in order to handle marshalling data between processes.

Per the feedback in the design review we may later provide a client dll that re-implements this interface but provides more optimized client-server communication by adding methods to the server idl that get and set multiple property values at once.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/984)


src/AppInstallerCLIInterfaceServer/pch.h
```diff
@@ -0,0 +1,18 @@
+﻿#pragma once
+
+#include <windows.h>
+#ifdef GetCurrentTime
+#undef GetCurrentTime
+#endif
+#include <unknwn.h>
+//#include <mutex>
+//#include "cppwinrtprocess.h"
```
cwruss, line 9: nit:remove if unused

src/Microsoft.Management.Deployment/pch.h
```diff
@@ -0,0 +1,7 @@
+﻿#pragma once
+//#include <mutex>
+//#include <combaseapi.h>
+//#include "cppwinrtprocess.h"
```
cwruss, line 4: nit:remove if unused

src/AppInstallerCLIInterfaceServer/AppInstallerCLIInterfaceServer.vcxproj
```diff
@@ -0,0 +1,176 @@
+<?xml version="1.0" encoding="utf-8"?>
```
yao-msft, line 1: Per compliance, for product code, we need to enable sdl check, code analysis and binskim. You can see how to enable them in this pr: [https://github.com/microsoft/winget-cli/pull/856](https://github.com/microsoft/winget-cli/pull/856)

src/AppInstallerCLI/AppInstallerCLI.vcxproj
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(SolutionDir)\packages\Microsoft.Windows.CppWinRT.2.0.200729.8\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('$(SolutionDir)\packages\Microsoft.Windows.CppWinRT.2.0.200729.8\build\native\Microsoft.Windows.CppWinRT.props')" />
+  <Import Project="..\packages\Microsoft.Windows.CppWinRT.2.0.210427.2\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('..\packages\Microsoft.Windows.CppWinRT.2.0.210427.2\build\native\Microsoft.Windows.CppWinRT.props')" />
```
yao-msft, line 3: >.. [](http://example.com/codeflow?start=19&length=2)

nit: I think we preferred $(SolutionDir) so we don't rely on relative paths, is that not the case anymore?

yao-msft:
New files are missing copyright headers

src/AppInstallerCLI/AppInstallerCLI.vcxproj
sreadingMSFT: Would assume it still is - it must just not be the default when you update the cppwinrt version. Will fix in every project that changed.

src/AppInstallerCLI.sln
```diff
@@ -105,6 +108,21 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		CodeAnalysis.ruleset = CodeAnalysis.ruleset
 	EndProjectSection
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "AppInstallerCLIInterfaceServer", "AppInstallerCLIInterfaceServer\AppInstallerCLIInterfaceServer.vcxproj", "{2B00D362-AC92-41F3-A8D2-5B1599BDCA01}"
```
cwruss, line 111: Maybe consider removing CLI from the names since this is a generic interface that is not specific to CLI?

JohnMcPMS: In addition, we are trying to use `WinGet` rather than `AppInstaller` for new things.

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -1,44 +1,62 @@
...
+<Package 
+  xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  xmlns:uap3="http://schemas.microsoft.com/appx/manifest/uap/windows10/3" 
+  xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+  xmlns:com="http://schemas.microsoft.com/appx/manifest/com/windows10"
+  xmlns:com2="http://schemas.microsoft.com/appx/manifest/com/windows10/2"
+  IgnorableNamespaces="uap uap3 uap5 rescap">
+  <Identity Name="WinGetDevCLI" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="0.0.16.0" />
```
JohnMcPMS, line 11: >16 [](http://example.com/codeflow?start=138&length=2)

Are you using this for anything more than an easy way to verify whether someone has the right package?

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -1,44 +1,62 @@
...
+          </uap3:AppExtensionHost>
+        </uap3:Extension>
+        <uap5:Extension Category="windows.appExecutionAlias">
+          <uap5:AppExecutionAlias>
+            <uap5:ExecutionAlias Alias="WinGetDev.exe" />
+          </uap5:AppExecutionAlias>
+        </uap5:Extension>
+        <com:Extension Category="windows.comServer">
+          <com:ComServer>
+            <com:ExeServer Executable="AppInstallerCLIInterfaceServer\AppInstallerCLIInterfaceServer.exe" DisplayName="AppInstaller Interface Server" LaunchAndActivationPermission="O:SYG:SYD:(A;;11;;;WD)(A;;11;;;RC)(A;;11;;;AC)(A;;11;;;AN)S:P(ML;;NX;;;S-1-16-0)">
```
JohnMcPMS, line 41: >AppInstaller [](http://example.com/codeflow?start=119&length=12)

This probable needs to be `Windows Package Manager`; should get @denelon to weigh in on this.

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -1,44 +1,62 @@
...
+              <com:Class Id ="C53A4F16-787E-42A4-B304-29EFFB4BF597" DisplayName="AppInstaller Server">
```
JohnMcPMS, line 42: >C53A4F16-787E-42A4-B304-29EFFB4BF597 [](http://example.com/codeflow?start=30&length=36)

This is going to conflict with the real package.  Is there any way to make this a separate interface, like we do with `wingetdev`?  I don't really see how we can acheive that, but throwing it out there in case someone else has ideas.

JohnMcPMS: Two possible solutions discussed:
1. `#ifdef` our way to having two sets of GUIDs and a duplicate IDL file.
2. Similar, but simply support both sets of GUIDs all of the time.

src/AppInstallerCLIPackage/Package.appxmanifest
sreadingMSFT: I'm not even using it for that. I can revert it back down to 0.0.3.0 or 0.0.2.1 or whatever else you want. I assume it probably should go up, but the way VS increments this when making a package locally doesn't make much sense for checking it in to git.

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
```diff
@@ -101,6 +101,9 @@ namespace AppInstaller::Repository
     // Gets the details for a single source.
     std::optional<SourceDetails> GetSource(std::string_view name);
 
+    // Gets the details for a single source by identifier
+    std::optional<SourceDetails> GetSourceByIdentifier(std::string_view identifier);
```
JohnMcPMS, line 105: >std::string_view identifier [](http://example.com/codeflow?start=55&length=27)

The way that this is being used, it should be a `PredefinedSource`.  If that is confusing because not all values are being exposed via COM, then another enum name would be fine (although I don't think that is necessary).

src/Microsoft.Management.Deployment/AppCatalog.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "FindPackagesResult.h"
+#include "MatchResult.h"
+#include "CatalogPackage.h"
+#include "PackageMatchFilter.h"
+#include "Microsoft/PredefinedInstalledSourceFactory.h"
+#include <wil\cppwinrt_wrl.h>
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    void PackageCatalog::Initialize(winrt::Microsoft::Management::Deployment::PackageCatalogInfo info, std::shared_ptr<::AppInstaller::Repository::ISource> source)
```
JohnMcPMS, line 15: Was this file copied and abandoned?  It seems like it should be PackageCatalog.

src/Microsoft.Management.Deployment/CatalogPackage.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "PackageVersionId.h"
+#include <wil\cppwinrt_wrl.h>
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    void CatalogPackage::Initialize(
+        std::shared_ptr<const ::AppInstaller::Repository::ISource> source,
+        std::shared_ptr<::AppInstaller::Repository::IPackage> package)
+    {
+        m_source = source;
```
JohnMcPMS, line 17: >source [](http://example.com/codeflow?start=19&length=6)

std::move these both

src/Microsoft.Management.Deployment/CatalogPackage.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return winrt::to_hstring(m_package->GetProperty(::AppInstaller::Repository::PackageProperty::Name));
+    }
+    Microsoft::Management::Deployment::PackageVersionInfo CatalogPackage::InstalledVersion()
+    {
+        if (!m_installedVersion)
+        {
+            // InstalledVersion hasn't been created yet, create and populate it.
+            std::shared_ptr<::AppInstaller::Repository::IPackageVersion> installedVersion = m_package.get()->GetInstalledVersion();
+            if (installedVersion)
```
JohnMcPMS, line 34: >if (installedVersion) [](http://example.com/codeflow?start=12&length=21)

If there is no installed version we keep running this over and over.

Also, this isn't thread safe, but I'm not sure what model you gave this server.

JohnMcPMS: Applies generally to numerous instances below.  In addition to thread safety for the members of the class, we should ensure that our SQLite connection is in serialized mode.  We don't specify it currently, and while the default is serialized, I could not find documentation on which setting was used to compile winsqlite3.

src/Microsoft.Management.Deployment/PackageCatalog.cpp
```diff
@@ -0,0 +1,115 @@
...
+                continue;
+            }
+            ::AppInstaller::Repository::MatchType packageFieldMatchOption = GetRepositoryMatchType(filter.Option());
+            ::AppInstaller::Repository::PackageMatchField matchField = GetRepositoryMatchField(filter.Field());
+
+            if (isSelector)
+            {
+                if (filter.Field() == Microsoft::Management::Deployment::PackageMatchField::CatalogDefault)
+                {
+                    searchRequest->Query = ::AppInstaller::Repository::RequestMatch(packageFieldMatchOption, winrt::to_string(filter.Value()));
```
JohnMcPMS, line 59: Error if second "query" value found?

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <wil\cppwinrt_wrl.h>
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    void PackageCatalogReference::Initialize(
+        ::AppInstaller::Repository::SourceDetails sourceDetails,
+        winrt::Microsoft::Management::Deployment::PackageCatalogInfo packageCatalogInfo)
+    {
+        m_sourceDetails = sourceDetails;
+        m_info = packageCatalogInfo;
```
JohnMcPMS, line 19: The details and the info are the same data; I assume that this was just to prevent the need to convert an info back into a details?

src/Microsoft.Management.Deployment/CatalogPackage.cpp
JohnMcPMS: Applies to a *lot* of other places.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -0,0 +1,82 @@
...
+                // TODO: this will need to be updated when adding the Installing catalog.
+                installedSource = ::AppInstaller::Repository::OpenPredefinedSource(::AppInstaller::Repository::PredefinedSource::Installed, progress);
+            }
+            
+            // Create the composite source.
+            m_source = ::AppInstaller::Repository::CreateCompositeSource(installedSource, remoteSources, searchBehavior);
+        }
+        else
+        {
+            m_source = ::AppInstaller::Repository::OpenSourceByIdentifier(winrt::to_string(m_info.Id()), progress).Source;
```
JohnMcPMS, line 65: >OpenSourceByIdentifier [](http://example.com/codeflow?start=51&length=22)

As discussed, `OpenSourceByDetails` is the better solution for most of this method.

src/Microsoft.Management.Deployment/PackageInstaller.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include "PackageInstaller.g.h"
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    [uuid("C53A4F16-787E-42A4-B304-29EFFB4BF597")]
+    struct PackageInstaller : PackageInstallerT<PackageInstaller>
```
JohnMcPMS, line 7: >PackageInstaller [](http://example.com/codeflow?start=11&length=16)

PackageManager

src/Microsoft.Management.Deployment/PackageInstaller.idl
```diff
@@ -0,0 +1,525 @@
...
+        /// Used by a caller to correlate the install with a caller's data.
+        /// IMPLEMENTATION NOTE: This area is in flux but right now it's expected that this would allow a caller to 
+        /// use JSON to encode multiple arguments.
+        String CorrelationData;
+        /// A string that will be passed to the source server if using a REST source
+        String AdditionalPackageCatalogArguments;
+    }
+
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 1)]
+    runtimeclass PackageInstaller
```
JohnMcPMS, line 472: >PackageInstaller [](http://example.com/codeflow?start=17&length=16)

PackageManager

src/Microsoft.Management.Deployment/PackageMatchFilter.h
```diff
@@ -0,0 +1,29 @@
...
+        void Initialize(::AppInstaller::Repository::PackageMatchFilter matchFilter);
+
+        winrt::Microsoft::Management::Deployment::PackageFieldMatchOption Option();
+        void Option(winrt::Microsoft::Management::Deployment::PackageFieldMatchOption const& value);
+        winrt::Microsoft::Management::Deployment::PackageMatchField Field();
+        void Field(winrt::Microsoft::Management::Deployment::PackageMatchField const& value);
+        hstring Value();
+        void Value(hstring const& value);
+    private:
+        hstring m_value = L"";
```
JohnMcPMS, line 19: >L"" [](http://example.com/codeflow?start=26&length=3)

Isn't this the default value?  I thought hstring was not nullable.

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -1,44 +1,62 @@
...
+              </com:Class>
+              <com:Class Id ="572DED96-9C60-4526-8F92-EE7D91D38C1A" DisplayName="FindPackagesOptions Server">
```
JohnMcPMS, line 44: >com:Class [](http://example.com/codeflow?start=15&length=9)

How is it decided which of these needs to be exported?

src/Microsoft.Management.Deployment/AppCatalog.cpp
sreadingMSFT: Yeah, git is being really annoying about renames for some reason, not sure what happened. Deleted.

src/AppInstallerCLIPackage/Package.appxmanifest
sreadingMSFT: Depends on whether the object needs to be creatable by the caller or if it's only created by the server and then passed back to the client.

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
sreadingMSFT: This got removed.

src/AppInstallerCLIPackage/Package.appxmanifest
sreadingMSFT: Option 2 is working in the latest version of this. AppInstaller package would register the other set of guids in the appxmanifest.

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
JohnMcPMS: Appears to still be here in the latest commit.

src/AppInstallerCLI.sln
sreadingMSFT: Made this change.

src/AppInstallerCLIPackage/Package.appxmanifest
sreadingMSFT: Made this change.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1009,6 +1009,31 @@ namespace AppInstaller::Repository
...
 
+    std::shared_ptr<ISource> CreateCompositeSource(
+        const std::shared_ptr<ISource>& installedSource,
+        const std::vector<std::shared_ptr<ISource>>& availableSources,
+        CompositeSearchBehavior searchBehavior)
+    {
+        if (availableSources.empty())
+        {
+            THROW_HR(E_UNEXPECTED);
+        }
```
JohnMcPMS, line 1020: This seems like it should be fine to allow to happen, even if it is useless.  If it were to fail, it should be invalid arg.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -436,92 +436,105 @@ namespace AppInstaller::Repository
...
-        // Search installed source
-        SearchResult installedResult = m_installedSource->Search(request);
-        result.Truncated = installedResult.Truncated;
-
-        for (auto&& match : installedResult.Matches)
+        // If the search behavior is for AllPackages or Installed then the result can contain packages that are
+        // only in the Installed source, but do not have an AvailableVersion.
+        // If the search behavior is for AvailablePackages, then most queries can be resolved by just querying the available source
+        // and then correlating it with the installed source, but if the query is for everything then query the installed catalog
+        // and only add installed packages that have an AvailableVersion.
```
JohnMcPMS, line 443: I don't think that this makes sense, if you ask for Available only, it should not go down this path at all.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -436,92 +436,105 @@ namespace AppInstaller::Repository
...
                     }
 
-                    // We found some matching packages here, don't keep going
-                    break;
+                    compositePackage->SetAvailablePackage(std::move(availablePackage));
                 }
 
-                compositePackage->SetAvailablePackage(std::move(availablePackage));
+                // If using CompositeSearchBehavior::AvailablePackages only add it if an available package was found.
+                if ((m_searchBehavior == CompositeSearchBehavior::AllPackages || m_searchBehavior == CompositeSearchBehavior::Installed) || (compositePackage->GetAvailablePackage() != nullptr))
```
JohnMcPMS, line 526: >if ((m_searchBehavior == CompositeSearchBehavior::AllPackages || m_searchBehavior == CompositeSearchBehavior::Installed) || (compositePackage->GetAvailablePackage() != nullptr)) [](http://example.com/codeflow?start=16&length=177)

This shouldn't be needed if the outer `if` is crafted correctly.

src/Microsoft.Management.Deployment/CatalogPackage.cpp
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return winrt::to_hstring(m_package->GetProperty(::AppInstaller::Repository::PackageProperty::Id).get());
+    }
+    hstring CatalogPackage::Name()
+    {
+        return winrt::to_hstring(m_package->GetProperty(::AppInstaller::Repository::PackageProperty::Name));
+    }
+    Microsoft::Management::Deployment::PackageVersionInfo CatalogPackage::InstalledVersion()
+    {
+        std::scoped_lock lock(m_installedVersionMutex);
```
JohnMcPMS, line 33: >std::scoped_lock lock(m_installedVersionMutex); [](http://example.com/codeflow?start=8&length=47)

I don't think that it is really a performance difference, but from a lines of code perspective using `std::call_once` would be better for these.  Only 1 member instead of two in the header, and then less bookkeeping for every use, for instance here it would be:
```
        std::call_once(m_installedVersionOnceFlag,
        [&]()
        {
std::shared_ptr<::AppInstaller::Repository::IPackageVersion> installedVersion = m_package.get()->GetInstalledVersion();
            if (installedVersion)
            {
                auto installedVersionImpl = winrt::make_self<wil::details::module_count_wrapper<
                    winrt::Microsoft::Management::Deployment::implementation::PackageVersionInfo>>();
                installedVersionImpl->Initialize(std::move(installedVersion));
                m_installedVersion = *installedVersionImpl;
            }
        });
        return m_installedVersion;
```

src/WinGetServer/WinMain.cpp
```diff
@@ -0,0 +1,54 @@
...
+wil::unique_event _comServerExitEvent;
+
+// Routine Description:
+// - Called back when COM says there is nothing left for our server to do and we can tear down.
+static void _releaseNotifier() noexcept
+{
+    _comServerExitEvent.SetEvent();
+}
+
+int __stdcall wWinMain(_In_ HINSTANCE, _In_opt_ HINSTANCE, _In_ LPWSTR, _In_ int)
```
JohnMcPMS, line 31: >wWinMain [](http://example.com/codeflow?start=14&length=8)

Needs experimental feature and to respect the overall "can use WPM" policy.

src/AppInstallerRepositoryCore/CompositeSource.cpp
sreadingMSFT: It's consistent with the early return when request.IsForEverything() and search behavior is AllPackages. Thought that was by design. The feature of "get me everything that's currently installed from the available catalogs" is good and needed somehow.

src/WinGetServer/WinMain.cpp
JohnMcPMS: The overall policy is here:
[https://github.com/microsoft/winget-cli/blob/862a2af3cd836f682a95093919e4f189f01e4c94/src/AppInstallerCommonCore/Public/winget/GroupPolicy.h#L34](https://github.com/microsoft/winget-cli/blob/862a2af3cd836f682a95093919e4f189f01e4c94/src/AppInstallerCommonCore/Public/winget/GroupPolicy.h#L34)

src/Microsoft.Management.Deployment/CatalogPackage.cpp
sreadingMSFT: Fancy.

src/WinGetServer/WinMain.cpp
sreadingMSFT: Done.

src/AppInstallerRepositoryCore/RepositorySource.cpp
sreadingMSFT: Allowed it.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -0,0 +1,82 @@
...
+            }
+            
+            // Create the composite source.
+            source = ::AppInstaller::Repository::CreateCompositeSource(installedSource, remoteSources, searchBehavior);
+        }
+        else
+        {
+            winrt::Microsoft::Management::Deployment::implementation::PackageCatalogInfo* catalogInfoImpl = get_self<winrt::Microsoft::Management::Deployment::implementation::PackageCatalogInfo>(m_info);
+            ::AppInstaller::Repository::SourceDetails sourceDetails = catalogInfoImpl->GetSourceDetails();
+            source = ::AppInstaller::Repository::OpenSource(sourceDetails.Name, progress).Source;
```
JohnMcPMS, line 69: >source = ::AppInstaller::Repository::OpenSource(sourceDetails.Name, progress).Source; [](http://example.com/codeflow?start=12&length=85)

OpenSourceByDetails?

src/AppInstallerCLIPackage/Package.appxmanifest
JohnMcPMS: How do we generate the two different headers then?

src/AppInstallerRepositoryCore/CompositeSource.cpp
JohnMcPMS: I don't think that was thought through; the search behavior has been tacked onto the original design and was only ever intended for use with import.  In that case it was never for everything.

I think it is better to make it make sense now than try to preserve an accidental behavior.

JohnMcPMS:
I would like some tests in the E2E category to ensure that it is functioning.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
sreadingMSFT: There's a cross-package locking issue when using that that I need more time to debug.

doc/Settings.md
```diff
@@ -133,3 +133,13 @@ Microsoft Store App support in WinGet is currently implemented as an experimenta
        "experimentalMSStore": true
    },
 ```
+
+### experimentalPackagedAPI
+
+Support in WinGet for packaged callers is currently implemented as an experimental feature. It allows other programs on Windows to use WinGet. You can enable the feature as shown below.
```
JohnMcPMS, line 139: >WinGet [](http://example.com/codeflow?start=135&length=6)

"the Windows Package Manager"

doc/Settings.md
```diff
@@ -133,3 +133,13 @@ Microsoft Store App support in WinGet is currently implemented as an experimenta
        "experimentalMSStore": true
    },
 ```
+
+### experimentalPackagedAPI
```
JohnMcPMS, line 137: >experimental [](http://example.com/codeflow?start=4&length=12)

Don't put experimental in the name; I missed it once but never again.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -0,0 +1,536 @@
...
+        /// Directs the logging to a log file. If provided, the installer must have have write access to the file 
+        String LogOutputPath;
+        /// Continues the install even if the hash in the catalog does not match the linked installer.
+        Boolean AllowHashMismatch;
+        /// A string that will be passed to the installer. 
+        /// IMPLEMENTATION NOTE: maps to "--override" in the winget cmd line
+        String ReplacementInstallerArguments;
+
+        /// Used by a caller to correlate the install with a caller's data.
+        /// JSON encoded strings are allowed.
```
JohnMcPMS, line 477: >allowed [](http://example.com/codeflow?start=37&length=7)

I think they are actually required.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -436,92 +436,100 @@ namespace AppInstaller::Repository
...
-            // Move the installed result into the composite result
-            result.Matches.emplace_back(std::move(compositePackage), std::move(match.MatchCriteria));
-        }
-
-        // Optimization for the "everything installed" case, no need to allow for reverse correlations
-        if (request.IsForEverything())
-        {
-            return std::move(result);
+            // Optimization for the "everything installed" case, no need to allow for reverse correlations
+            if (request.IsForEverything())
```
JohnMcPMS, line 529: >if (request.IsForEverything()) [](http://example.com/codeflow?start=12&length=30)

This probably needs to not happen for `m_searchBehavior == CompositeSearchBehavior::AllPackages`, right?

src/WinGetServer/WinMain.cpp
```diff
@@ -0,0 +1,79 @@
...
+        winrt::check_hresult(globalOptions->Set(COMGLB_RO_SETTINGS, COMGLB_FAST_RUNDOWN));
+    }
+
+    AppInstaller::COMContext::SetLoggers();
+
+    _comServerExitEvent.create();
+    auto& module = ::Microsoft::WRL::Module<::Microsoft::WRL::ModuleType::OutOfProc>::Create(&_releaseNotifier);
+    try
+    {
+        if (!IsServerEnabled())
```
JohnMcPMS, line 65: >if (!IsServerEnabled()) [](http://example.com/codeflow?start=8&length=23)

From a debugging and usability standpoint, this check should probably happen in the factory function rather than simply refusing to start the process.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
JohnMcPMS: Should have been fixed by one of my changes.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -0,0 +1,316 @@
...
+        }
+        return catalogs.GetView();
+    }
+    winrt::Microsoft::Management::Deployment::PackageCatalogReference PackageManager::GetPredefinedPackageCatalog(winrt::Microsoft::Management::Deployment::PredefinedPackageCatalog const& predefinedPackageCatalog)
+    {
+        switch (predefinedPackageCatalog)
+        {
+        case winrt::Microsoft::Management::Deployment::PredefinedPackageCatalog::OpenWindowsCatalog:
+            // TODO: Mapping of enum to sources should link directly to definition, not copy string.
+            return GetPackageCatalogByName(L"winget");
```
JohnMcPMS, line 57: Should be implemented with a new `OpenWellKnownSource` function (or similar name).  Could also use the existing `OpenPredefinedSource` and add new enum values, as well as changing the internals to not assume that the result is of the type Predefined.

JohnMcPMS (changes requested):
:clock1:

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -964,31 +951,44 @@ namespace AppInstaller::Repository
...
-                        {
-                            sourceList.SaveMetadata();
-                        }
-                    }
-                    catch (...)
-                    {
-                        AICLI_LOG(Repo, Warning, << "Failed to update source: " << (*source).Name);
-                        result.SourcesWithUpdateFailure.emplace_back(*source);
-                    }
+                    sourceList.SaveMetadata();
```
JohnMcPMS, line 971: This doesn't do anything of value; the details passed in may be updated but they will have no effect on the source list.

I hadn't thought of this aspect of things, and thus my previous suggestions may have been a little off of what we actually need.

It might be possible to prevent a rewrite of the whole details/metadata design by having the metadata saved for all details, including tombstones.  But this function would have to first look up the real details from the `SourceListInternal` and use those to update/open.

I actually prefer instead to go with a Predefined model, where you can get the details but still have to open them by enum.  I realize that makes things more complex on the COM method, but I foresee issues with someone creating some arbitrary source details and then wanting to open them and it not working correctly.

If not one of those two, then I think it is a rewrite of the details system to be more object oriented.  Basically making it more like the reference system in the COM API.  It would certainly make the COM API simpler, but I don't think it should be done as part of this change.

For now I would suggest just matching the details with the right value in the source list and using that.  And comment about matching the tombstones.  I'm working on a refactor of the metadata code anyway, I might be able to also move to the OO model at the same time.

JohnMcPMS (changes requested):

JohnMcPMS (approved):

PR #1006, opened 2021-05-28
Enable COM APIs to log diagnostic and telemetry events and Log Corela…

sachintaMSFT:
…tion Vector Json to existing events

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1006)


src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -11,6 +11,15 @@
 #define AICLI_TraceLoggingStringView(_sv_,_name_) TraceLoggingCountedUtf8String(_sv_.data(), static_cast<ULONG>(_sv_.size()), _name_)
 #define AICLI_TraceLoggingWStringView(_sv_,_name_) TraceLoggingCountedWideString(_sv_.data(), static_cast<ULONG>(_sv_.size()), _name_)
 
+#define AICLI_TraceLoggingWriteActivity(_eventName_,...) TraceLoggingWriteActivity(\
+g_hTraceProvider,\
+_eventName_,\
+GetActivityId(false),\
+nullptr,\
+TraceLoggingCountedString(m_caller.c_str(),  static_cast<ULONG>(m_caller.size()), "Caller"),\
+TraceLoggingPackedFieldEx(GetTelemetryCorelationJsonW().c_str(), static_cast<ULONG>((GetTelemetryCorelationJsonW().size() + 1) * sizeof(wchar_t)), TlgInUNICODESTRING, TlgOutJSON, "CvJson"),\
```
JohnMcPMS, line 20: This should use the member, or the function should be changed to return a `const &`.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -91,16 +109,54 @@ namespace AppInstaller::Logging
...
+        std::string errors;
+        std::wstring jsonStrW{ jsonStr_view };
+        std::string jsonStr = ConvertToUTF8(jsonStrW.c_str());
+
+        bool result = jsonReader->parse(jsonStr.c_str(),
+            jsonStr.c_str() + jsonStr.size(),
+            pJsonValue.get(),
+            &errors);
+
+        delete jsonReader;
```
JohnMcPMS, line 128: This should be a `std::unique_ptr`

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -91,16 +109,54 @@ namespace AppInstaller::Logging
...
+
+        if (result)
+        {
+            m_telemetryCorelationJsonW = jsonStrW;
+            AICLI_LOG(Core, Info, << "Passed in Corelation Json is valid: " << jsonStr);
+        }
+        else
+        {
+            AICLI_LOG(Core, Info, << "Passed in Corelation Json is invalid: " << jsonStr << "; Error: " << errors);
+            m_telemetryCorelationJsonW = L"{}";
```
JohnMcPMS, line 138: The default value should already be this; plus we should make this an error, not a hidden swallow.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -133,33 +189,31 @@ namespace AppInstaller::Logging
...
 
-        AICLI_LOG(Core, Info, << "WinGet, version [" << version << "], activity [" << *GetActivityId() << ']');
+        AICLI_LOG(Core, Info, << "WinGet, version [" << version << "], activity [" << *GetActivityId(false) << ']');
         AICLI_LOG(Core, Info, << "OS: " << Runtime::GetOSVersion());
         AICLI_LOG(Core, Info, << "Command line Args: " << Utility::ConvertToUTF8(GetCommandLineW()));
         if (Runtime::IsRunningInPackagedContext())
         {
             AICLI_LOG(Core, Info, << "Package: " << packageVersion);
         }
+        AICLI_LOG(Core, Info, << "IsCOMCall:" << isCOMCall << "; Caller: " << m_caller << "; CallerCj: ");
```
JohnMcPMS: Use the same name as the event, not `cj`.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
sachintaMSFT: This was mistakenly left over string. removed that. we already log it when setting CvJson.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
sachintaMSFT: Done.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
sachintaMSFT: Used member instead.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
sachintaMSFT: Json::CharReader has virtual methods and is abstract class which cannot be instantiated in make_unique. The derived class OurCharReader isn't abstract whose object pointer newCharReader() is actually returning but it is not declared in Json.h to make unique_ptr to it. How to use std::unique_ptr here ?

sachintaMSFT: Right. Thanks.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -11,6 +11,15 @@
 #define AICLI_TraceLoggingStringView(_sv_,_name_) TraceLoggingCountedUtf8String(_sv_.data(), static_cast<ULONG>(_sv_.size()), _name_)
 #define AICLI_TraceLoggingWStringView(_sv_,_name_) TraceLoggingCountedWideString(_sv_.data(), static_cast<ULONG>(_sv_.size()), _name_)
 
+#define AICLI_TraceLoggingWriteActivity(_eventName_,...) TraceLoggingWriteActivity(\
+g_hTraceProvider,\
+_eventName_,\
+GetActivityId(false),\
+nullptr,\
+TraceLoggingCountedString(m_caller.c_str(),  static_cast<ULONG>(m_caller.size()), "Caller"),\
```
JohnMcPMS, line 19: `TraceLoggingCountedUtf8String`

src/AppInstallerCommonCore/FileLogger.cpp
```diff
@@ -13,7 +13,7 @@ namespace AppInstaller::Logging
     using namespace std::string_view_literals;
     using namespace std::chrono_literals;
 
-    static constexpr std::string_view s_fileLoggerDefaultFilePrefix = "WinGet-"sv;
+    static constexpr std::string_view s_fileLoggerDefaultFilePrefix = "WPM-"sv;
```
JohnMcPMS, line 16: Ah, I meant change the prefix *for the COM server logs*.

src/AppInstallerCommonCore/Public/AppInstallerTelemetry.h
```diff
@@ -119,6 +125,9 @@ namespace AppInstaller::Logging
 
         bool IsTelemetryEnabled() const noexcept;
 
+        // Check for valid Json string and return a valid Json
+        std::wstring GetTelemetryCorelationJsonW() const;
```
JohnMcPMS, line 129: I still think this would be better as returning a `const std::wstring&`

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/FileLogger.cpp
sachintaMSFT: Got it.

src/AppInstallerCommonCore/Public/AppInstallerTelemetry.h
sachintaMSFT: Ok. Yes.

JohnMcPMS (approved):
I just went ahead and did these to prevent any more back and forth.  If you approve of my changes, it can be merged.

src/AppInstallerCommonCore/AppInstallerLogging.cpp
```diff
@@ -130,6 +130,21 @@ namespace AppInstaller::Logging
         }
     }
 
+    std::string_view DiagnosticLogger::GetLogFileNamePrefix()
```
JohnMcPMS, line 133: And if I did feel it was necessary to change the prefix of the installer log file, I would want to use something cleaner than this.  If there is no file log, then installer log files have no prefix?

I think that your last two commits should be reverted except for your one change that was unrelated.

If you do want to force the installer log file to have the same prefix, then it needs to be done another way.  The first thing that comes to mind is storing the default as a member on `DiagnosticLogger`.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/AppInstallerLogging.cpp
sachintaMSFT: Undoing the change.

JohnMcPMS (approved):
I thought I had approved this, sorry.

PR #1098, opened 2021-06-03
Make CrossProcessReaderWriteLock resilient to process termination

JohnMcPMS:
## Change
The previous implementation (using a semaphore) was not resilient to unexpected process termination, as Windows does not track semaphore ownership.  This could lead to a deadlock in a waiting process, and at least one user even experienced an ongoing deadlock despite having killed all of the active processes (which should have destroyed and thus reset the named semaphore).

This version uses a set of named mutexes instead, which Windows will release for us in the event of abandonment.  As part of the change I have also added the ability to pass in an `IProgressCallback` to allow for cancellation of the wait.

On top of that change, I realized the the source management functions were not properly indicating that they were cancelled.  While a CTRL+C in the middle of a `source add` would in fact cancel the action internally, it would still add the source information into the settings stream.  I have updated the stack to handle cancellation now, including the output from top level now saying `Cancelled` rather than `Done` to indicate that CTRL+C actually did stop the action.

## Validation
All existing tests pass with these changes; a new test is added for `IProgressCallback` cancellation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1098)

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -257,6 +257,7 @@ msrc
 Multifile
 Multimatch
 mutex
+mutexes
 namespace
 namespaces
 Nelon
@@ -530,6 +531,7 @@ woah
 wofstream
 workaround
 workflow
+wostringstream
 wostream
 wpfn
 wrl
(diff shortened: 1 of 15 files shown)


yao-msft (approved):
:shipit:

Issue #1012, opened 2021-05-28
Show dependencies


fzanollo:
# Description of the new feature/enhancement
Some packages may have required dependencies, as of now nothing is done with them. This functionality informs the user about required dependencies before continuing.

# Proposed technical implementation details (optional)
[https://github.com/fzanollo/winget-cli/blob/dependencies/doc/specs/%231012%20-%20Show%20dependencies.md](https://github.com/fzanollo/winget-cli/blob/dependencies/doc/specs/%231012%20-%20Show%20dependencies.md)

fzanollo:
Part of the feature discussed in #163

PR #1114, opened 2021-06-04
Specs for show dependencies' feature #1012

fzanollo:
Specs for show dependencies feature, see issue #1012.
This is a step towards managing dependencies #163.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1114)


doc/specs/#1012 - Show dependencies.md
```diff
@@ -0,0 +1,111 @@
...
+Package had dependencies that may not be needed anymore:
+  - Windows Feature: Hyper-V
+  - Package: Microsot.WindowsTerminal
+Starting package uninstall...
+Successfully uninstalled
+```
+Dependencies will not be uninstalled.
+
+### validate
+For this step dependencies will not be validated.
```
jedieaston, line 91: I think that validate should attempt to resolve the dependencies against the repositories available to see if they exist or not, it'd save on typos and moderation headaches.

denelon: @jedieaston just a heads-up. We're not planning on supporting cross-repository dependencies yet. There would only be "one" source to look for dependencies (the same one the manifest came from). The "-m" option would allow you to specify the source to check for those dependencies when you are testing a local manifest.

denelon: I guess this could be something like:

 `winget validate -m <pathToManifest> -s winget`

Validation could at least (in theory) check to see if the PackageIdentifier values for the dependencies exist at validation time for the given source.

JohnMcPMS: Yes, I agree that `validate` doing what checks it can is a good idea.  This iteration of the feature is just acknowledging that dependencies exist by informing the user though, so at most I think we would do something similar to `show` w/r/t dependencies.

@fzanollo is ramping up here so we are taking a very iterative approach.

doc/specs/#163 - Dependencies.md
```diff
@@ -58,7 +58,7 @@ These include other packages. The restriction on these dependencies is that they
 ### External Dependencies
 These include dependencies from outside of the source the original package is distributed. In some cases suitable items may exist in the same source, but for licensing or personal preference, no explicit package should be required. One example is Java. Many vendors offer JREs and JDKs so it may be more reasonable to have a user informed, and allow the user to confirm the presence of a dependency.
 
-
+As a first step the Windows Package Manager should show dependency information to the user, see issue [1012](github.com/microsoft/winget-cli/issues/1012) with [spec](github.com/fzanollo/winget-cli/blob/dependencies/doc/specs/%231012%20-%20Show%20dependencies.md).
```
JohnMcPMS, line 61: As @jsoref pointed out on the other iteration of this, you should use a relative path, not an absolute.  I believe something like `/doc/specs/%231012%20-%20Show%20dependencies.md` will work, but I am certainly no master of GitHub MD.

doc/specs/#1012 - Show dependencies.md
```diff
@@ -0,0 +1,111 @@
...
+Starting package uninstall...
+Successfully uninstalled
+```
+Dependencies will not be uninstalled.
+
+### validate
+For this step dependencies will not be validated. 
+
+### import
+Will show dependencies before installing each of the specified packages on the import; i.e., it will not gather all the dependencies and show them at once before starting.
```
JohnMcPMS, line 94: Why not gather all dependencies?  I agree that it isn't strictly needed, but it seems to go against the goal of having the majority of the code path laid out by this change.

JohnMcPMS:
Additional note; please place this behind an experimental feature (yes, even just the information only phase).

doc/specs/#163 - Dependencies.md
jsoref: I'd suggest relative 
```suggestion
As a first step the Windows Package Manager should show dependency information to the user, see issue [1012](github.com/microsoft/winget-cli/issues/1012) with [spec](./%231012%20-%20Show%20dependencies.md).
```

doc/specs/#1012 - Show dependencies.md
fzanollo: Yes, thank you John.
I thought about informing dependencies on validate but since it doesn't show any manifest information wasn't sure how to go about it. It may be weird to show only dependencies info and not package name or else. What do you think?

doc/specs/#163 - Dependencies.md
fzanollo: Oh nice, thank you @jsoref didn't know you could use links like that on github.

doc/specs/#1012 - Show dependencies.md
fzanollo: Well it's something I wanted to ask. Import should install packages only if it can install them all? If so then yes, gathering all the dependencies is the best approach. Otherwise, it may be better to check and install one by one. 

Although now that I think about it import has an option for what to do right?(--ignore-unavailable) maybe it can be different depending on that?

jsoref (commented):
The other items should be added to expect.txt

doc/specs/#1012 - Show dependencies.md
JohnMcPMS: If I remember the default behavior correctly, `import` will fail before doing any work if it cannot provide one of the packages.  To maintain this same behavior, all dependencies would also need to be verified before attempting any installations.

doc/specs/#1012 - Show dependencies.md
```diff
@@ -0,0 +1,111 @@
...
+Moniker: notepad++
+Description: Notepad++ is a free (as in “free speech” and also as in “free beer”) source code editor and Notepad replacement that supports several languages. Running in the MS Windows environment, its use is governed by GNU General Public License.
+Homepage: https://notepad-plus-plus.org/
+License: GPL-2.0-only
+License Url: https://raw.githubusercontent.com/notepad-plus-plus/notepad-plus-plus/v7.9.5/LICENSE
+Installer:
+  Type: Nullsoft
+  Locale: en-US
+  Download Url: https://github.com/notepad-plus-plus/notepad-plus-plus/releases/download/v7.9.5/npp.7.9.5.Installer.x64.exe
+  SHA256: 4881548cd86491b453520e83c19292c93b9c6ce485a1f9eb9301e3913a9baced
```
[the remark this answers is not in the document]
fzanollo: Changed in [d3794bd](https://github.com/microsoft/winget-cli/pull/1114/commits/d3794bdbd0eb69ef474719d69e865f19121ffdf1). Thank you @jsoref

.github/actions/spelling/expect.txt
```diff
@@ -25,6 +26,7 @@ AType
 AUrl
 Aysnc
 azurewebsites
+baced
```
jsoref, line 29: This isn't necessary, but it's ok, it'll get cleaned out eventually:
https://github.com/microsoft/winget-cli/runs/2783873246#step:3:61

.github/actions/spelling/expect.txt
```diff
@@ -341,4 +347,5 @@ Wunused
 WZDNCRFJ
 XPLATSTR
 xsi
-zy
+Zanollo
+zy
```
jsoref, line 351: Git/Diff tools/Unix tools in general appreciate having a new line at the end of files. (You can take the line that's currently being inserted at the beginning of the file 😉.)

(The tool itself doesn't care much these days, but the next person to touch the file will probably be blamed for the last line in the file, just as you're currently taking blame for `zy` which was added by someone else.)

n.b.. in this case, that's going to be short-lived anyway, I need to send a PR to upgrade check-spelling, at which point two-letter words will be ignored entirely so they'll be dropped no later than that PR.

fzanollo: Sorry, since the file was alphabetically ordered I thought it needed to be that way. Will have that into account for next time.

jsoref: It wants to be sorted, yes. It doesn't need to be (I understand that people don't like running arbitrary code, but are willing to make hand edits). That said, sorting it is generally appreciated. (At the very least, sorting makes it easier for a reader to spot related items with some typo.)

The commands it suggests will clean out stale entries, order it, and ensure a new line at the of file.

It's ok to do any/all of these by hand.

I have a feature that should be generally usable soon that will let you ask the action to do this cleanup instead of doing it by hand, but I'm letting it bake with some other almost done features. Hoping to ship it soon.

denelon: @jsorf, I hear lots of complaints from people until the dictionary gets cleaned up a "bit". It's an awesome tool after one has adapted! In the winget-pkgs directory we will have plenty of manifests! Is there a way to ignore files by extension, or better yet, support localization "to some extent"?

jsoref: @denelon: `excludes.txt` should let you exclude things fairly easily. Certainly, ignoring files by extension or path (or both).

Also, there's an optional advice file which would help:

https://github.com/check-spelling/spell-check-this/blob/main/.github/actions/spelling/advice.md

(It's content is configurable, this is just a sample. Note that in older versions of the action the file needs to be `.txt`, so test, ...)

I'm generally including it when I send PRs to update the action.

doc/specs/#1012 - Show dependencies.md
fzanollo: Validate will show dependencies of all installers found.

doc/specs/#1012 - Show dependencies.md
```diff
@@ -0,0 +1,124 @@
...
+  - Windows Feature: Hyper-V
+  - Package: Microsoft.WindowsTerminal
+Manifest validation succeeded.
+```
+
+### import
+Will gather all the dependencies and show them together before starting.
+```
+The packages found in this import have the following dependencies:
+  - Windows Feature: Hyper-V, Containers
```
JohnMcPMS, line 103: I'm not a huge fan of this model of having them all on one line, but I can see the value for import.

Also, you need to have examples of at least External dependencies to see how they fit into this model.  I don't think that putting them on one line makes sense.

fzanollo: Ok, something like a nested bullet point will be better?

JohnMcPMS: No, I think that a whole line for the type is wasted.  I would rather have each on its own line, or multiple on the line (like you had it).  I wouldn't expect many packages to have loads of dependencies, although a graph could explode.

Own line option:
```
Windows Feature: Hyper-V
Windows Feature: Containers
Package: Foo.Foo
```

ljkt7334 (approved):

JohnMcPMS (approved):

PR #1146, opened 2021-06-10
Move Temp path to the user's location

### LINKED ISSUE #189: Discord does not install
When trying to install Discord via winget, installer throws an error about lack of permission to copy file. Doesn't matter if this is done via Windows Terminal launched as admin or not (in the latter case, error left in console is different and only second error from attached screenshot is displayed).

error

# Steps to reproduce
1. Launch Windows Terminal as admin
2. Issue `winget install Discord` command

# Expected behavior
Discord installs properly

# Actual behavior
Installation fails

# Environment
```
Windows Package Manager v0.1.41331
Windows: Windows.Desktop v10.0.19041.264
Package: Microsoft.DesktopAppInstaller v1.0.41331.0
```

JohnMcPMS:
Fixes #189 , and maybe others with strange permissions errors.

## Change
Move the temp directory that is used from the package specific temp to the user's temp location.  This should prevent the Mandatory Low ACL that was apparently causing #189 from being inherited.  While this is more avoiding the problem than fixing it, given the relatively low rate of occurrence it might be hard to ensure that it was truly fixed by actually changing the ACLs (and would require much more time and code to write).

Additionally changes the behavior of `Runtime::GetPathTo` to delete non-directories when they are encountered rather than just erroring directly.

## Validation
Manually validated that installers are downloaded to the new location; do not have a repro of #189 to try against, but given the root cause it is unlikely that it wouldn't be fixed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1146)

### DIFF
--- a/src/AppInstallerCommonCore/Runtime.cpp
+++ b/src/AppInstallerCommonCore/Runtime.cpp
@@ -88,6 +88,16 @@ namespace AppInstaller::Runtime
             return knownFolder.get();
         }
 
+        // Gets the user's temp path
+        std::filesystem::path GetPathToUserTemp()
+        {
+            wchar_t tempPath[MAX_PATH + 1];
+            DWORD tempChars = GetTempPathW(ARRAYSIZE(tempPath), tempPath);
+            THROW_LAST_ERROR_IF(!tempChars);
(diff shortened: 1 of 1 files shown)


yao-msft (approved):
:shipit:

PR #1165, opened 2021-06-14
Show dependencies feature #1012

### LINKED ISSUE #1012: Show dependencies

fzanollo:
Contains:
- implementation of show dependencies feature for commands: install, show, upgrade, uninstall, validate, import.
- everything under a new experimental feature 
- unit testcases for each of the commands
Closes #1012 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1165)


src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -123,4 +123,70 @@ namespace AppInstaller::CLI::Workflow
 
         context.Reporter.Info() << Resource::String::UninstallFlowUninstallSuccess << std::endl;
     }
+
+    void ReportDependencies(Execution::Context& context)
+    {
+        context <<
+            Workflow::GetManifest <<
```
JohnMcPMS, line 130: `GetManifest` is not the right thing to do here.  I still think that this should use the unified context data item, so this particular task wouldn't worry about it.

But when you do get the dependencies, it should be something like this:
```
auto p = context.Get<Execution::Data::Package>();
if (matchingAvailableVersion = p->GetAvailableVersion({"", p->GetInstalledVersion()->GetProperty(Version), "" }))
{
  ChooseInstallerBasedOnInstalledVersion(matchingAvailableVersion->GetManifest()).Dependencies;
}
```

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -123,4 +123,70 @@ namespace AppInstaller::CLI::Workflow
...
+            
+        const auto& installer = context.Get<Execution::Data::Installer>();
+        if (installer)
+        {
+            auto dependencies = installer->Dependencies;
+            if (dependencies.HasAny())
+            {
+                context.Reporter.Info() << "This package had dependencies that may not be needed anymore:" << std::endl;
+
+                auto windowsFeaturesDep = dependencies.WindowsFeatures;
```
JohnMcPMS, line 141: Definitely too many copies of this floating around.  A single workflow task should be able to handle the reporting here.  I would keep it as `Report*` as it will need to stay that long term, then the action tasks can call `Report*` for now and eventually be replaced with the actual action.

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
```diff
@@ -111,7 +111,9 @@ namespace AppInstaller::CLI::Workflow
 
             updateAllFoundUpdate = true;
 
-            updateContext << InstallPackageInstaller;
+            updateContext << 
+                ManageDependencies <<
```
JohnMcPMS, line 115: See how much cleaner the callsite is when you don't have to check the feature?  Also, you would have forgotten to check the feature here if you needed to.

src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
```diff
@@ -511,6 +511,26 @@
     <CopyFileToFolders Include="TestData\Manifest-Bad-Channel-NotSupported.yaml">
       <DeploymentContent>true</DeploymentContent>
     </CopyFileToFolders>
+    <CopyFileToFolders Include="TestData\Manifest-Good-AllDependencyTypes.yaml">
+      <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</DeploymentContent>
```
JohnMcPMS, line 515: Remove `Condition` and `FileType`.

src/AppInstallerCLITests/AppInstallerCLITests.vcxproj.filters
```diff
@@ -450,5 +450,10 @@
     <CopyFileToFolders Include="TestData\ImportFile-Good-MachineScope.json">
       <Filter>TestData</Filter>
     </CopyFileToFolders>
+    <CopyFileToFolders Include="TestData\Manifest-Good-AllDependencyTypes.yaml" />
```
JohnMcPMS, line 453: These should be in the `TestData` filter.

src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -782,6 +815,30 @@ TEST_CASE("ShowFlow_SearchAndShowAppVersion", "[ShowFlow][workflow]")
...
+    ShowCommand show({});
+    show.Execute(context);
+    INFO(showOutput.str());
+
+    // Verify all types of dependencies are printed
+    REQUIRE(showOutput.str().find("Dependencies") != std::string::npos);
+    REQUIRE(showOutput.str().find("WindowsFeaturesDep") != std::string::npos);
+    REQUIRE(showOutput.str().find("WindowsLibrariesDep") != std::string::npos);
+    // PackageDep1 has minimum version (1.0), PackageDep2 doesn't (shouldn't show [>=...])
+    REQUIRE(showOutput.str().find("Package.Dep1-x64\x1b[0m [>= 1.0]") != std::string::npos);
```
JohnMcPMS, line 836: I think we should explicitly be looking not to output these extraneous characters.  I'm going to guess that they are being output as the "reset" state for the Info stream.  Instead of using `Info()` repeatedly, simply capturing the object for reuse prevents that:
```
auto info = context.Info();
```

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -124,6 +124,11 @@ namespace AppInstaller::Manifest
         std::vector<string_t> WindowsLibraries;
         std::vector<PackageDependency> PackageDependencies;
         std::vector<string_t> ExternalDependencies;
+
+        bool HasAny() const {
```
JohnMcPMS, line 128: Regarding moving to a unified dependency model, I would suggest having the manifest simply have a `std::vector<Dependency>` as well.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -407,13 +407,111 @@ namespace AppInstaller::CLI::Workflow
...
+
+            ManageWindowsLibrariesDependencies(dependencies.WindowsLibraries, context);
+
+            ManagePackageDependencies(dependencies.PackageDependencies, context);
+
+            ManageExternalDependencies(dependencies.ExternalDependencies, context);
+        }
+    }
+
+    void ManageWindowsFeatureDependencies(std::vector<AppInstaller::Manifest::string_t>& windowsFeaturesDep, AppInstaller::CLI::Execution::Context& context)
```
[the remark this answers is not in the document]
fzanollo: You mean a vector with all dependencies together or divided by type? because Dependency is a struct with a vector for each type.

src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
fzanollo: This were auto-generated

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -35,39 +35,96 @@ namespace AppInstaller::CLI::Workflow
...
+                    context.Reporter.Info() << Execution::ManifestInfoEmphasis << "  Dependencies: " << std::endl;
+
+                    auto windowsFeaturesDep = dependencies.WindowsFeatures;
+                    if (!windowsFeaturesDep.empty())
+                    {
+                        context.Reporter.Info() << "    - Windows Features: ";
+                        for (size_t i = 0; i < windowsFeaturesDep.size(); i++)
+                        {
+                            context.Reporter.Info() << windowsFeaturesDep[i];
+                            if (i < windowsFeaturesDep.size() - 1) context.Reporter.Info() << ", ";
```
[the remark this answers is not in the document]
fzanollo: I'll show them one on each new line

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -35,39 +35,96 @@ namespace AppInstaller::CLI::Workflow
...
+            if (Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::EFExperimentalShowDependencies)) {
+                auto dependencies = installer->Dependencies;
+                if (dependencies.HasAny())
+                {
+                    context.Reporter.Info() << Execution::ManifestInfoEmphasis << "  Dependencies: " << std::endl;
+
+                    auto windowsFeaturesDep = dependencies.WindowsFeatures;
+                    if (!windowsFeaturesDep.empty())
+                    {
+                        context.Reporter.Info() << "    - Windows Features: ";
```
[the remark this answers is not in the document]
fzanollo: Only for ShowFlow right?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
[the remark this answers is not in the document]
JohnMcPMS: I mean a new type to hold a generic dependency; sorry if the name was already taken.

```C++
enum class DependencyType
{
  WindowsFeature,
  Package,
  ...
}

struct Dependency
{
  DependencyType Type;
  std::string Name;
  std::optional<Utility::Version> Version;
}
```

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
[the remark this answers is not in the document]
JohnMcPMS: Yes, only `show` should use the manifest names directly (for now).  The rest of the time we should use a localized, friendly name.

jjvrmorales367 (approved):
approve

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -407,13 +407,111 @@ namespace AppInstaller::CLI::Workflow
...
+        {
+            context << Workflow::ManageDependencies;
+        }
+        
+        context << Workflow::InstallPackageInstaller;
+    }
+
+    void ManageDependencies(Execution::Context& context)
+    {
+        const auto& installer = context.Get<Execution::Data::Installer>();
```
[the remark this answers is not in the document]
fzanollo: I created a separate context data item called Dependency which is of type DependencyList

jjvrmorales367 (approved):
It's worth a trial

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -408,12 +409,41 @@ namespace AppInstaller::CLI::Workflow
...
+            Workflow::ManageDependencies <<
             Workflow::InstallPackageInstaller;
     }
 
+    void ManageDependencies(Execution::Context& context)
+    {
+        if (Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::Dependencies))
+        {
+            const auto& installer = context.Get<Execution::Data::Installer>();
+            if (installer && installer->Dependencies.HasAny())
```
JohnMcPMS, line 421: All dependencies tasks should separate creating a list of dependencies and using that list.  I would not expect to see a task that both adds the list of dependencies and then later consumes them (or calls another task that consumes them).  All of the top level flows should be able to do something like:
```
context << DetermineDependencies << ActOnDependencies;
```
Where the "determine" and "act on" tasks may be different for each flow, but always completely independent from each other.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -429,7 +459,51 @@ namespace AppInstaller::CLI::Workflow
             // TODO: In the future, it would be better to not have to convert back and forth from a string
             installContext.Args.AddArg(Execution::Args::Type::InstallScope, ScopeToString(package.PackageRequest.Scope));
 
-            installContext << InstallPackageVersion;
+            installContext <<
+                Workflow::SelectInstaller;
+            
+            if (installContext.IsTerminated())
+            {
+                allSucceeded = false;
```
JohnMcPMS, line 467: If the context is terminated, we should be stopping this iteration of the loop.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -408,12 +409,41 @@ namespace AppInstaller::CLI::Workflow
...
+
+        std::optional<AppInstaller::Manifest::ManifestInstaller> Installer;
+        AppInstaller::CLI::Execution::PackageToInstall Package;
+    };
+
     void InstallMultiple(Execution::Context& context)
     {
         bool allSucceeded = true;
+
+        DependencyList allDependencies;
```
JohnMcPMS, line 443: No need to have empty lines between these.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -429,7 +459,51 @@ namespace AppInstaller::CLI::Workflow
...
+            Execution::Context& installContext = *installContextPtr;
+
+            // set data needed for installing
+            installContext.Add<Execution::Data::PackageVersion>(package.PackageVersion);
+            installContext.Add<Execution::Data::Manifest>(package.PackageVersion->GetManifest());
+            installContext.Args.AddArg(Execution::Args::Type::InstallScope, ScopeToString(package.PackageRequest.Scope));
+            installContext.Add<Execution::Data::Installer>(installer);
+
+            installContext <<
+                Workflow::EnsureApplicableInstaller <<
```
JohnMcPMS, line 504: This should definitely be happening in the upper loop.

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -35,39 +35,65 @@ namespace AppInstaller::CLI::Workflow
...
+
+            if (Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::Dependencies)) {
+                auto info = context.Reporter.Info();
+                const auto& dependencies = installer->Dependencies;
+
+                if (dependencies.HasAny())
+                {
+                    info << Execution::ManifestInfoEmphasis << "  Dependencies: " << std::endl;
+
+                    if (dependencies.HasAnyOf(Manifest::DependencyType::WindowsFeature)) info << "    - WindowsFeatures: " << std::endl;
```
JohnMcPMS, line 81: `if` style

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -122,5 +122,15 @@ namespace AppInstaller::CLI::Workflow
         }
 
         context.Reporter.Info() << Resource::String::UninstallFlowUninstallSuccess << std::endl;
+    }
+    
+    void GetDependencies(Execution::Context& context)
+    {
+        const auto& installer = context.Get<Execution::Data::Installer>();
```
JohnMcPMS, line 129: There is no `Installer` for `uninstall`.  I commented previously on how this would need to be done.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -112,20 +112,58 @@ namespace AppInstaller::Manifest
...
     {
-        std::vector<string_t> WindowsFeatures;
-        std::vector<string_t> WindowsLibraries;
-        std::vector<PackageDependency> PackageDependencies;
-        std::vector<string_t> ExternalDependencies;
+        DependencyType Type;
+        string_t Id;
+        std::optional<string_t> MinVersion;
+
+        Dependency(DependencyType type, string_t id, std::optional<string_t> minVersion) : Type(type), Id(id), MinVersion(minVersion) {}
```
JohnMcPMS, line 129: The parameter here should be `string_t minVersion` as there is no need to call this overload if you don't have a value for it.

All of these should use `std::move` to initialize the string value from the parameter.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -112,20 +112,58 @@ namespace AppInstaller::Manifest
...
+        Dependency(DependencyType type, string_t id) : Type(type), Id(id) {}
+        Dependency(DependencyType type) : Type(type) {}
     };
 
+    struct DependencyList
+    {
+        std::vector<Dependency> dependencies;
```
JohnMcPMS, line 136: Put at end of struct and make private.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -112,20 +112,58 @@ namespace AppInstaller::Manifest
...
+
+        DependencyList() {}
+
+        void Add(Dependency dependency) { dependencies.push_back(dependency); }
+        void Add(DependencyList otherDependencyList)
```
JohnMcPMS, line 141: `const DependencyList&`

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -112,20 +112,58 @@ namespace AppInstaller::Manifest
...
+        Dependency(DependencyType type) : Type(type) {}
     };
 
+    struct DependencyList
+    {
+        std::vector<Dependency> dependencies;
+
+        DependencyList() {}
+
+        void Add(Dependency dependency) { dependencies.push_back(dependency); }
```
JohnMcPMS, line 140: `const Dependency&` and/or `Dependency&&`.  Probably both overloads are worthwhile.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -112,20 +112,58 @@ namespace AppInstaller::Manifest
...
+        Dependency(DependencyType type, string_t id, std::optional<string_t> minVersion) : Type(type), Id(id), MinVersion(minVersion) {}
+        Dependency(DependencyType type, string_t id) : Type(type), Id(id) {}
+        Dependency(DependencyType type) : Type(type) {}
     };
 
+    struct DependencyList
+    {
+        std::vector<Dependency> dependencies;
+
+        DependencyList() {}
```
JohnMcPMS, line 138: If you want a default behavior for constructors et al., use `DependencyList() = default;`

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
JohnMcPMS: This should also be checking for validity; and probably in different configurations:
1. Manifest validation would require uniqueness of dependencies; so error if a second PackageDependency shows up for the same package name.
2. Relaxed mode would be used for things like `import` and loading manifests (not for validation).  It would not error on a second PackageDependency, but rather take the higher minimum version.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -112,20 +112,58 @@ namespace AppInstaller::Manifest
...
+        bool HasAnyOf(DependencyType type) const 
+        {
+            for (const auto& dependency : dependencies)
+            {
+                if (dependency.Type == type) return true;
+            };
+            return false;
+        }
+
+        void ApplyToType(DependencyType type, std::function<void(Dependency)> func) const
```
JohnMcPMS, line 159: `std::function<void(const Dependency&)>`

src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
```diff
@@ -71,6 +71,6 @@ namespace AppInstaller::Manifest
         // For msix only
         std::vector<string_t> RestrictedCapabilities;
 
-        Dependency Dependencies;
+        DependencyList Dependencies;
```
JohnMcPMS, line 74: Are these properly being applied to all installers if they are in the root of the manifest?  I don't remember if the manifest schema allows that; but if it does there is extra work to ensure that it actually happens.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/ManifestDeserializer.cpp
```diff
@@ -428,18 +428,32 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0::Json
...
         }
 
-        Manifest::Dependency dependency;
+        Manifest::DependencyList dependencyList;
 
-        dependency.WindowsFeatures = ConvertToManifestStringArray(JsonHelper::GetRawStringArrayFromJsonNode(dependenciesObject, JsonHelper::GetUtilityString(WindowsFeatures)));
-        dependency.WindowsLibraries = ConvertToManifestStringArray(JsonHelper::GetRawStringArrayFromJsonNode(dependenciesObject, JsonHelper::GetUtilityString(WindowsLibraries)));
-        dependency.ExternalDependencies = ConvertToManifestStringArray(JsonHelper::GetRawStringArrayFromJsonNode(dependenciesObject, JsonHelper::GetUtilityString(ExternalDependencies)));
+        const auto& wfIds = ConvertToManifestStringArray(JsonHelper::GetRawStringArrayFromJsonNode(dependenciesObject, JsonHelper::GetUtilityString(WindowsFeatures)));
+        for (auto id : wfIds)
```
JohnMcPMS, line 441: The temporary `wfIds` should not be a temporary, and this should then use `auto&&` and below should use `std::move`.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
```diff
@@ -128,6 +129,8 @@ namespace AppInstaller::CLI
         context <<
             Workflow::GetInstalledPackageVersion <<
             Workflow::GetUninstallInfo <<
+            Workflow::GetDependencies <<
```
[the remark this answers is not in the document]
fzanollo: Changed it for GetDependenciesInfo

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -408,12 +409,41 @@ namespace AppInstaller::CLI::Workflow
         context <<
             Workflow::SelectInstaller <<
             Workflow::EnsureApplicableInstaller <<
+            Workflow::ManageDependencies <<
             Workflow::InstallPackageInstaller;
     }
 
+    void ManageDependencies(Execution::Context& context)
```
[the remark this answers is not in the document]
fzanollo: Actually as of now it will not manage any dependencies and it's only gathering dependencies info from installer. So instead of moving it to DependenciesFlow I changed its name to GetInstallerDependencies

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
fzanollo: Agree but for import (InstallMultiple) it's hard to divide as the task is not completely divided into workflows.

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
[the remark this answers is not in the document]
JohnMcPMS: I mean more specifically for `uninstall`.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -409,20 +409,20 @@ namespace AppInstaller::CLI::Workflow
...
     {
         if (Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::Dependencies))
         {
             const auto& installer = context.Get<Execution::Data::Installer>();
             if (installer && installer->Dependencies.HasAny())
             {
                 context.Add<Execution::Data::Dependencies>(installer->Dependencies);
-                context.Reporter.Info() << "This package requires the following dependencies:" << std::endl;
-                context << Workflow::ReportDependencies;
+                context.Reporter.Info() << Resource::String::InstallAndUpgradeCommandsReportDependencies << std::endl;
```
JohnMcPMS, line 425: It should not be reporting anything either; the "determine" step from the other comment should be simply a transform that collects the dependencies and places them on the context.

If you have different header statements for different flows, consider using a workflow task object that is constructed with the desired string resource for the output.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -124,13 +124,16 @@ namespace AppInstaller::CLI::Workflow
         context.Reporter.Info() << Resource::String::UninstallFlowUninstallSuccess << std::endl;
     }
     
-    void GetDependencies(Execution::Context& context)
+    void GetDependenciesInfo(Execution::Context& context)
     {
-        const auto& installer = context.Get<Execution::Data::Installer>();
-        if (installer && installer->Dependencies.HasAny())
+        const auto& installedPackageVersion = context.Get<Execution::Data::InstalledPackageVersion>();
+        const auto& installerOptions = installedPackageVersion->GetManifest().Installers;
```
JohnMcPMS, line 130: This doesn't work (yet); you need to follow the pattern I gave in an older comment for the current codebase to get a manifest to work with.

src/AppInstallerCLICore/Workflows/ValidateFlow.cpp
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "pch.h"
+#include "ValidateFlow.h"
+
+namespace AppInstaller::CLI::Workflow
+{
+    void GetDependenciesFromManifest(Execution::Context& context) {
```
JohnMcPMS, line 9: All dependencies related functions should be in the dependencies files.  I get that the names of the files make it seem like they are related to the commands, but they are not tied to them.  I would much rather group them up by their relation to certain features, and dependencies is certainly large enough to warrant its own file.

Please move this one and the one from InstallFlow to the dependencies file.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -921,4 +921,32 @@ Configuration is disabled due to Group Policy.</value>
...
+  </data>
+  <data name="PackageDependencies" xml:space="preserve">
+    <value>Packages</value>
+  </data>
+  <data name="UninstallCommandReportDependencies" xml:space="preserve">
+    <value>This package had dependencies that may not be needed anymore:</value>
+    <comment>Uninstall command sentence showed before reporting dependencies</comment>
+  </data>
+  <data name="ValidateCommandReportDependencies" xml:space="preserve">
+    <value>Manifest has the following dependencies that were not validated, ensure that they are valid:</value>
```
JohnMcPMS, line 943:
```suggestion
    <value>Manifest has the following dependencies that were not validated; ensure that they are valid:</value>

```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -921,4 +921,32 @@ Configuration is disabled due to Group Policy.</value>
   <data name="Cancelled" xml:space="preserve">
     <value>Cancelled</value>
   </data>
+  <data name="ExternalDependencies" xml:space="preserve">
+    <value>Externals</value>
```
JohnMcPMS, line 925: "External Dependencies" or just "External"

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -126,16 +126,14 @@ namespace AppInstaller::Manifest
         string_t Id;
         std::optional<string_t> MinVersion;
 
-        Dependency(DependencyType type, string_t id, std::optional<string_t> minVersion) : Type(type), Id(id), MinVersion(minVersion) {}
-        Dependency(DependencyType type, string_t id) : Type(type), Id(id) {}
+        Dependency(DependencyType type, string_t id, string_t minVersion) : Type(std::move(type)), Id(std::move(id)), MinVersion(std::move(minVersion)) {}
```
JohnMcPMS, line 129: `type` doesn't need `std::move` as there are no external resources.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -156,13 +154,41 @@ namespace AppInstaller::Manifest
             return false;
         }
 
+        bool HasDependency(DependencyType type, string_t id, string_t minVersion = "")
+        {
+            for (const auto& dependency : dependencies)
+            {
+                if (dependency.Type == type && dependency.Id == id)
```
JohnMcPMS, line 161: Id will need to be compared with ICU case insensitive equality.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
fzanollo: Is there some configuration I can check or should I just use a parameter to choose between each?

JohnMcPMS: The list object should take a parameter to choose; the only place that it would be set to use strict mode would be the validate command (aka the `fullValidation` bool is set).

fzanollo: I think as for this step (show/report dependencies) it will be better to just output the dependencies to the user. Although merging dependencies, i.e., take the higher min version if the dependency exists, is a nice idea to have now.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -158,7 +178,19 @@ namespace AppInstaller::Manifest
...
-        bool HasDependency(DependencyType type, string_t id, string_t minVersion = "")
+        Dependency* HasDependency(const Dependency& dependencyToSearch)
+        {
+            for (auto& dependency : dependencies) {
+                if (dependency.Type == dependencyToSearch.Type && ICUCaseInsensitiveEquals(dependency.Id, dependencyToSearch.Id))
+                {
+                    return &dependency;
+                }
+            }
+            return NULL;
```
JohnMcPMS, line 189: style: `nullptr`

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -135,16 +135,36 @@ namespace AppInstaller::Manifest
...
         { 
-            dependencies.push_back(dependency); 
+            Dependency* existingDependency = this->HasDependency(newDependency);
+
+            if (existingDependency != NULL) {
+                if (newDependency.MinVersion) 
+                {
+                    if (existingDependency->MinVersion)
+                    {
+                        const auto& newDependencyVersion = AppInstaller::Utility::Version(newDependency.MinVersion.value());
```
JohnMcPMS, line 147: I think it would also be fine to simply hold the MinVersion as `std::optional<Version>`; but that can be an optimization to think about in a future change.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -135,16 +135,36 @@ namespace AppInstaller::Manifest
...
+                        const auto& existingDependencyVersion = AppInstaller::Utility::Version(existingDependency->MinVersion.value());
+                        existingDependency->MinVersion.value() = newDependencyVersion <= existingDependencyVersion ? existingDependencyVersion.ToString() : newDependencyVersion.ToString();
```
JohnMcPMS, line 149: Given that there are allocations required, I would structure this as "write only if needed":
```
if (existing < new)
  existing->MV = new.MV;
```

src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
fzanollo: @denelon said the common behavior is: 
- Root applies when none are specified on an installer node. 
- Installer nodes override root. 

Added tests for this, change it to work like that and put a comment on this features' specs.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
fzanollo: Agree, as of now that is the only place we are in need of a Version object but it will surely change in the future.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -146,7 +146,13 @@ namespace AppInstaller::Manifest
                     {
                         const auto& newDependencyVersion = AppInstaller::Utility::Version(newDependency.MinVersion.value());
                         const auto& existingDependencyVersion = AppInstaller::Utility::Version(existingDependency->MinVersion.value());
-                        existingDependency->MinVersion.value() = newDependencyVersion <= existingDependencyVersion ? existingDependencyVersion.ToString() : newDependencyVersion.ToString();
+                        if (newDependencyVersion <= existingDependencyVersion) {
+                            existingDependency->MinVersion.value() =  existingDependencyVersion.ToString();
```
JohnMcPMS, line 150: It already is this value though; there is no need to update it in this case.

JohnMcPMS (approved):

PR #1170, opened 2021-06-15
Setting DefaultLocale's default value to "en-US"

palenshus:
Nice to have a default value, and enables our codegen'd object model to have a default value set for this property

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1170)

### DIFF
--- a/schemas/JSON/manifests/v1.0.0/manifest.defaultLocale.1.0.0.json
+++ b/schemas/JSON/manifests/v1.0.0/manifest.defaultLocale.1.0.0.json
@@ -32,6 +32,7 @@
     },
     "PackageLocale": {
       "type": "string",
+      "default": "en-US",
       "pattern": "^([a-zA-Z]{2}|[iI]-[a-zA-Z]+|[xX]-[a-zA-Z]{1,8})(-[a-zA-Z]{1,8})*$",
       "maxLength": 20,
       "description": "The package meta-data locale"
--- a/schemas/JSON/manifests/v1.0.0/manifest.version.1.0.0.json
+++ b/schemas/JSON/manifests/v1.0.0/manifest.version.1.0.0.json
@@ -18,6 +18,7 @@
     },
     "DefaultLocale": {
       "type": "string",
+      "default": "en-US",
       "pattern": "^([a-zA-Z]{2}|[iI]-[a-zA-Z]+|[xX]-[a-zA-Z]{1,8})(-[a-zA-Z]{1,8})*$",
       "maxLength": 20,
       "description": "The default package meta-data locale"


palenshus:
@denelon, thoughts?

denelon:
I've seen one or two packages submitted with other defaultLocale values. I don't mind having a default as long as the user can change it if they wish.

JohnMcPMS:
> Nice to have a default value, and enables our codegen'd object model to have a default value set for this property

I don't think we actually do want a default here; I would rather it be required to be provided rather than accidentally forgotten (and then potentially wrong).

palenshus:
> I don't think we actually do want a default here; I would rather it be required to be provided rather than accidentally forgotten (and then potentially wrong).

Setting a default [doesn't mean it's not required](https://json-schema.org/understanding-json-schema/reference/generic.html#:~:text=The%20default%20keyword%20specifies%20a%20default%20value%20for,in%20which%20it%20resides%2C%20but%20that%20isn%E2%80%99t%20required.), validation will still fail if the property is missing altogether. This just gives a nice default value to codegen tools like ours

jjvrmorales367 (approved):
I'm gonna ive it a trial

JohnMcPMS:
> > I don't think we actually do want a default here; I would rather it be required to be provided rather than accidentally forgotten (and then potentially wrong).
> 
> Setting a default [doesn't mean it's not required](https://json-schema.org/understanding-json-schema/reference/generic.html#:~:text=The%20default%20keyword%20specifies%20a%20default%20value%20for,in%20which%20it%20resides%2C%20but%20that%20isn%E2%80%99t%20required.), validation will still fail if the property is missing altogether. This just gives a nice default value to codegen tools like ours

Did you verify that `validate` still properly fails when a file is missing this value?

yao-msft:
> > > I don't think we actually do want a default here; I would rather it be required to be provided rather than accidentally forgotten (and then potentially wrong).
> > 
> > 
> > Setting a default [doesn't mean it's not required](https://json-schema.org/understanding-json-schema/reference/generic.html#:~:text=The%20default%20keyword%20specifies%20a%20default%20value%20for,in%20which%20it%20resides%2C%20but%20that%20isn%E2%80%99t%20required.), validation will still fail if the property is missing altogether. This just gives a nice default value to codegen tools like ours
> 
> Did you verify that `validate` still properly fails when a file is missing this value?

IIRC, the default field is not used during schema validation. So it won't break the "validate". We have semantic validation that the DefaultLocale field in version manifest must match PackageLocale field in defaultLocale manifest. Shall we update the defaultLocale schema to default to en-US to be consistent too?

palenshus:
@yao-msft is correct, we have many other default values already, which winget validate complains if they aren't set (for example, ManifestType, ManifestVersion, etc). And Yao yes I think it would make sense to default the PackageLocale to that as well

palenshus:
@yao-msft, @JohnMcPMS, @denelon, ping

denelon:
@palenshus I don't see this as a problem, we just need to be very careful about schema changes, and we're currently working on the v1.1 schema so we should make sure to be consistent across all of the product area.

yao-msft:
> @yao-msft, @JohnMcPMS, @denelon, ping

@palenshus, can you send a new commit changing the defaultLocale schema to default to en-US as well, so I can just approve and merge?

yao-msft (approved):

Issue #1009, opened 2021-05-28
First usage feedback: multiple non-silent fails, one crash, and two types of repeats


Master-Guy:
# Brief description of your issue
Sorry if this shouldn't be reported this way, but I'm not yet up-to-speed on the Microsoft ways-of-work on Github.
After installing winget-cli this morning, and running `winget upgrade --all`, I got the following feedback:
- Multiple non-silent fails
- One crash of winget
- Looping installation after fail
- Upgrades of up-to-date applications

# Steps to reproduce
Have an old 32-bit version of Teamviewer installed on your 64 machine, together with any old version of the other apps listed below under _Actual behaviour_.
Then run `winget upgrade --all` as administrator in a Windows Terminal session running Powershell as default.

# Expected behavior
Winget to update these packages, and kill a running download when pressing CTRL+C.

# Actual behavior
While downloading the Unity update of 2GB, I pressed CTRL+C because I needed my bandwidth for a Teams meeting, but it seems the download kept running even though I got my Powershell prompt back.

Multiple packages (also a number of packages not listed here) have visible installers. It wouldn't be too much of an issue if these would be running in the background, but they are being pushed as top window, and get the focus of Windows, which makes it possible to accidentally cancel an installation, otherwise interrupt an installation, or close an error message.

- Battle.Net: Keeps re-installing the update (and launching after installation).
- Discord: Installer fails to install, because Discord.exe was still running. 2nd attempts installs a version that still needs to update when running for the first time.
- LockHunter: Keeps re-installing the update (and popping up their website after installation).
- Parsec: Installer hash does not match; this cannot be overridden when running as admin.
- PyCharm Community Edition: Installation fails, because it attempts to install into a non-empty directory. Then it re-tries, gives the same error, and continues to the next package. As long as the error is open, _winget_ does not continue.
- Speccy: Keeps re-installing the update.
- Spotify: Fails because it can only be installed under user level, and not as administrator.
- TrackMania Nations Forever: Keeps re-installing the update (500MB slow download with visible installer).
- TeamViewer: Returns error 0x80072f7c and exits _winget_. After manually updating, it successfully skips this package.
- CORSAIR iCUE 4 Software: Installer freezes after spawning multiple C++ Application Development Frameworks (QtWebEngineProcess.exe). _winget_ reported error 1603 and continued after I killed the installer.
- euroscope: Installer failed with exit code: 1603 (no user-interaction happened)
[...]

PR #1172, opened 2021-06-16
Make CTRL signal able to take multiple contexts

JohnMcPMS:
Related to #1009 , fixes one of the problems but there are many listed.

## Change
Update the CTRL signal handler to be able to fan out to multiple contexts, and make all cases of using sub-contexts both connected to the signal handler and better handle cancellation.

Additionally, improve the COM API install method to properly cancel any in progress task rather than just terminating the context.

## Validation
Manually ensured that CTRL+C cancels various multi-step commands.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1172)

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -424,6 +424,7 @@ STRINGIFY
 STRINGIZE
 stringstream
 strstr
+subcontext
 SUBLANG
 subresource
 subselect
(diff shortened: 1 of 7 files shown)


jjvrmorales367 (approved):
I give ut a trial and see if anything changes

yao-msft (approved):
:shipit:

PR #1175, opened 2021-06-17
Fix a potential infinite loop due to file read errors during hashing

JohnMcPMS:
## Change
The `std::istream` was not handled properly here, as only `eof` determined the loop termination, when any of the state bits should have.  Now requires `good` and additionally lets `bad` throw exceptions directly, while throwing our own generic exception if `fail` is set without `eof`.

## Validation
Exiting tests and manual validation of golden path.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1175)


JohnMcPMS:
> Can you also do the mapping to DownloadError in Converters.cpp in GetInstallResultStatus

I don't think that the error strategy there should be to require every possible error to be mapped.  It would be better to have the current phase set the default error, and then allow for specific errors to bubble up through as necessary.  In that way, this would already be a download error (unless it happens in a different phase, which it could).

sreadingMSFT:
> > Can you also do the mapping to DownloadError in Converters.cpp in GetInstallResultStatus
> 
> I don't think that the error strategy there should be to require every possible error to be mapped. It would be better to have the current phase set the default error, and then allow for specific errors to bubble up through as necessary. In that way, this would already be a download error (unless it happens in a different phase, which it could).

Ok, I like that idea. I'll make the change separately.

sreadingMSFT (approved):

yao-msft (approved):
:shipit:

PR #1182, opened 2021-06-18
Add more error handling for com callers and prevent concurrent installs

sreadingMSFT:
This change has error mapping changes to make error mapping more reliable when new errors are added by using the execution stage to determine the expected status if it's not specifically overridden. It also adds an explicit check for concurrent comcallers in the context, based on the knowledge that those callers can currently only be installs, and returns a known failure for that scenario that callers can check for while the overall scenario is still in development for queuing com installs from callers like powershell.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1182)


src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -1,97 +1,121 @@
...
+            // Specifically, Rest sources do not have the Ids filled in m_info since they only get the id from the rest server after being Opened.
+            auto packageCatalogInfo = winrt::make_self<wil::details::module_count_wrapper<winrt::Microsoft::Management::Deployment::implementation::PackageCatalogInfo>>();
+            packageCatalogInfo->Initialize(source->GetDetails());
+            auto connectResult = winrt::make_self<wil::details::module_count_wrapper<winrt::Microsoft::Management::Deployment::implementation::ConnectResult>>();
+            auto packageCatalog = winrt::make_self<wil::details::module_count_wrapper<winrt::Microsoft::Management::Deployment::implementation::PackageCatalog>>();
+            packageCatalog->Initialize(*packageCatalogInfo, source, (m_compositePackageCatalogOptions != nullptr));
+            connectResult->Initialize(winrt::Microsoft::Management::Deployment::ConnectResultStatus::Ok, *packageCatalog);
+            return *connectResult;
+        }
+        catch (const wil::ResultException&)
```
JohnMcPMS, line 102: Do you plan on doing anything with these individual exception types?  If you want to just swallow all exceptions the last one is sufficient.

JohnMcPMS (approved):

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
sreadingMSFT: Yeah, thats true. I added them thinking I'd add logging before I realized the hresult can't be set anywhere and its already being logged. Removed.

PR #1198, opened 2021-06-22
Add Capability checks for callers to the packagedAPI

sreadingMSFT:
Check for the packageManagement capability for caller apps. All Medium-IL processes have all capabilities so this will pass for powershell callers. I originally tried to add these checks in the activation factories but the calling process in that case was a svchost process.

I also updated the telemetry to log the caller.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1198)


.github/actions/spelling/allow.txt
```diff
@@ -1,559 +1,561 @@
...
-XElement
-XName
-XResource
-xsd
-XTOKEN
-yaml
-yml
-yy
-zh
+abi
```
JohnMcPMS, line 1: I wonder if this file having every line changed is what has lead to so many spelling failures.  I suspect that github has merged this file and mucked about with the line endings.

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -0,0 +1,92 @@
...
+        if ((rpcStatus != RPC_S_NO_CALL_ACTIVE) &&
+            !((rpcStatus == RPC_S_OK) && HandleToULong(callAttributes.ClientPID) == GetCurrentProcessId()))
+        {
+            DWORD callerProcessId = HandleToULong(callAttributes.ClientPID);
+            return callerProcessId;
+        }
+        return {};
+    }
+
+    HRESULT EnsureProcessHasCapability(DWORD callerProcessId)
```
JohnMcPMS, line 29: More specific name since this is for just one capability.

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // Best effort at getting caller info. This should only be used for logging.
+    std::wstring TryGetCallerProcessInfo(DWORD callerProcessId)
+    {
+        wil::unique_process_handle processHandle(OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, callerProcessId));
+        if (processHandle)
+        {
+            HANDLE token;
+            if (OpenProcessToken(processHandle.get(), TOKEN_QUERY, &token))
```
JohnMcPMS, line 66: The token HANDLE needs to be closed; use wil.

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -0,0 +1,92 @@
...
+            {
+                std::wstring appUserModelID = GetAppUserModelIdFromToken(token);
+                if (!appUserModelID.empty())
+                {
+                    // The AppUserModelID is of the format <PackageFamilyName>!<ApplicationName>, only the package family name is of interest
```
JohnMcPMS, line 71: There is no `GetPackageFamilyNameFromToken` API?

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    HRESULT EnsureProcessHasCapability(DWORD callerProcessId)
+    {
+        // Get the caller process id and use it to check if the caller has permissions to access the feature.
+        winrt::Windows::Security::Authorization::AppCapabilityAccess::AppCapabilityAccessStatus status = winrt::Windows::Security::Authorization::AppCapabilityAccess::AppCapabilityAccessStatus::DeniedBySystem;
+
+        auto capability = winrt::Windows::Security::Authorization::AppCapabilityAccess::AppCapability::CreateWithProcessIdForUser(nullptr, L"packageManagement", callerProcessId);
+        status = capability.CheckAccess();
+        RETURN_HR_IF(E_ACCESSDENIED, status != winrt::Windows::Security::Authorization::AppCapabilityAccess::AppCapabilityAccessStatus::Allowed);
```
JohnMcPMS, line 36: This check works if the caller is packaged but Medium IL? (AKA Centennial)

src/Microsoft.Management.Deployment/PackageCatalog.cpp
```diff
@@ -107,70 +108,81 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
     {
         winrt::Microsoft::Management::Deployment::FindPackagesResultStatus::Ok;
         bool isTruncated = false;
         Windows::Foundation::Collections::IVector<Microsoft::Management::Deployment::MatchResult> matches{ winrt::single_threaded_vector<Microsoft::Management::Deployment::MatchResult>() };
         ::AppInstaller::Repository::SearchRequest searchRequest;
 
-        HRESULT hr = PopulateSearchRequest(&searchRequest, options);
-        if (SUCCEEDED(hr))
+        HRESULT hr = S_OK;
+        if (FAILED(hr = EnsureComCallerHasCapability()))
```
JohnMcPMS, line 129: Why is this the place to require the capability?  Why is it not the object factory method?

I would think that installing and other modifying APIs would require `packageManagement`, while searching would require `packageQuery`.  So maybe it does make sense to do the checks in the functions themselves.  But then we are requiring the wrong capabilities.

.github/actions/spelling/allow.txt
jsoref: There's a 0.0.18. I need to send a PR.

I thought that I had fixed it to complain but tolerate this case:
```
winget-cli/.github/actions/spelling/allow.txt:    ASCII text, with CRLF line terminators
winget-cli/.github/actions/spelling/excludes.txt: ASCII text
winget-cli/.github/actions/spelling/expect.txt:   ASCII text
winget-cli/.github/actions/spelling/patterns.txt: ASCII text
winget-cli/.github/actions/spelling/reject.txt:   ASCII text
```

Anyway, dos2unix on the file does fix this. https://github.com/jsoref/winget-cli/actions/runs/962433068

src/Microsoft.Management.Deployment/PackageCatalog.cpp
sreadingMSFT: Read the description! Fine, probably I should add a comment too. The ActivateInstance calls seem to be coming from a com service svchost, so the process id isn't useful for doing the check.

src/Microsoft.Management.Deployment/Helpers.cpp
sreadingMSFT: OOPS! Yeah, there very much is.

.github/actions/spelling/allow.txt
jsoref: For reference, c0f1081ae5def579847148c3bee7926c1bc1d6c6 changed the line endings (so it wasn't a merge).

jsoref: Indeed, in 0.0.18 the tool would have tolerated the changed line endings:https://github.com/check-spelling/winget-cli/runs/2890223690.
> Warning: .github/actions/spelling/allow.txt: line 1, columns 3-4, Warning - entry has unexpected whitespace (whitespace-in-dictionary)
...
(Yes, it's very noisy...)

That said, I'm not sure you want the line ending changes commits...

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -0,0 +1,97 @@
...
+        if ((rpcStatus != RPC_S_NO_CALL_ACTIVE) &&
+            !((rpcStatus == RPC_S_OK) && HandleToULong(callAttributes.ClientPID) == GetCurrentProcessId()))
+        {
+            DWORD callerProcessId = HandleToULong(callAttributes.ClientPID);
+            return callerProcessId;
+        }
+        return {};
+    }
+
+    std::wstring GetStringForCapability(Capability capability)
```
JohnMcPMS, line 30: Better to return a `std::wstring_view` in these cases to prevent an unnecessary allocation.

src/Microsoft.Management.Deployment/PackageCatalog.cpp
```diff
@@ -107,70 +108,78 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            isTruncated = searchResult.Truncated;
         }
-        auto findPackagesResult = winrt::make_self<wil::details::module_count_wrapper<
-            winrt::Microsoft::Management::Deployment::implementation::FindPackagesResult>>();
-        // TODO: Add search timeout and error code.
-        winrt::Microsoft::Management::Deployment::FindPackagesResultStatus status = FindPackagesResultStatus(hr);
-        findPackagesResult->Initialize(status, isTruncated, matches);
-
-        return *findPackagesResult;
+        // Exceptions that may occur in the process of executing an arbitrary command
```
JohnMcPMS, line 161: It seems like we should just have a function to handle this:
```
template <typename T>
HRESULT ExecuteWithExceptionTrap(T&& t)
{
  HRESULT result = S_OK;
  try { t(); }
  catch ... // All of these different types
  return result;
}
```
or copy WIL and create our own `WINGET_CATCH_RETURN()` and/or `WINGET_CATCH_STORE(hr)` macro.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -37,10 +38,22 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        connectResult->Initialize(winrt::Microsoft::Management::Deployment::ConnectResultStatus::CatalogError, nullptr);
+        return *connectResult;
+    }
     winrt::Microsoft::Management::Deployment::ConnectResult PackageCatalogReference::Connect()
     {
         try
         {
+            if (FAILED(EnsureComCallerHasCapability(Capability::PackageQuery)))
+            {
+                // TODO: When more error codes are added, this should go back as something other than CatalogError.
```
JohnMcPMS, line 53: Why not `E_ACCESSDENIED` right now?

src/Microsoft.Management.Deployment/Helpers.cpp
sreadingMSFT: Yes, tested using a hardcoded process id with a centennial app from the store.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
sreadingMSFT: To throw? Would prefer not to throw. Needs an idl update to return more errors unfortunately.

FATHE9876 (approved):

sachintaMSFT (approved):

src/AppInstallerCLICore/COMContext.h
```diff
@@ -56,6 +56,8 @@ namespace AppInstaller
         //Execution::Context
         void SetExecutionStage(CLI::Workflow::ExecutionStage executionPhase, bool);
 
+        CLI::Workflow::ExecutionStage GetExecutionStage() const { return m_executionStage; }
```
JohnMcPMS, line 59: This should be on the base `Context`; not sure why you moved it.

JohnMcPMS (approved):

src/AppInstallerCLICore/COMContext.h
sreadingMSFT: ComContext apparently has its own m_executionStage that it updates in SetExecutionStage, base class doesn't get called, so getting it from base class is useless apparently.

PR #1201, opened 2021-06-22
Add basic synchronization for the install queue to prevent installs from colliding

sreadingMSFT:
Remove the temporary error code that was being returned when this scenario hit. Added a static mutex and removed async calls that would cause the mutex to be released too early or on the wrong thread. The mutex needs to be static because Install is a class method and callers can create as many PackageManager's as they want.

Tested multiple simultaneous installs and verified the wait behavior. 

I also moved the Queued call down below where the arguments were verified. I noticed in my tests that even though the progress and cancel tokens are co_awaited the Queued event was getting sent so fast that my caller wasn't even receiving it because the event handler wasn't hooked up. This can be fixed on the client side by setting the operation to not start immediately, but that seems overly complex, and I don't like the idea of Queued being returned before it's actually waiting, when the call may still be rejected due to invalid arguments.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1201)


sreadingMSFT:
I guess I need to let update this to let Cancel through while it's waiting.

JohnMcPMS (approved):

Issue #142, opened 2020-05-14
Change default sort order on search.


denelon:
I would like the ability to customize the sort order when multiple results are returned.

In addition to arguments like `winget search _foo_ -sort Name`, I would like settings to specify my default preferences.

`winget list -sort Version`

In addition, I would like to have settings to handle results from multiple sources so they can be displayed in a coalesced manner vs. the default where we sort by source based on their order in `winget source list`.

>Note: The current sort order is based on the fields matched for search. This may vary by source as REST sources are allowed to return results in any order they prefer.

jhoff80:
Same for the 'show' command.  I don't want it to be sorted by the ID but the name.

weitzhandler:
Currently, is there any way to display all apps alphabetically at all?

chausner:
I started an implementation for this feature in https://github.com/microsoft/winget-cli/pull/1205.

JohnMcPMS:
There is currently a sort order defined by the search itself (as mentioned in the OP).  It is currently sorted by (applies to the `winget` source; arbitrary REST sources can sort arbitrarily):

1. The "quality" of the match
2. The field that matched
3. Arbitrary ordering inherent to the internal data representation

Quality here means whether the match was exact, a substring, etc.  So this means that `winget search git` puts "Git" first as it is the best match due to the case insensitive equality between the name and the search term.  Always sorting the entire output by name blindly will lead to results with far less relevance being placed first.  For instance, when I run this locally I am seeing a result that is accidentally related because "digital" happens to contain "git" (and tags are lower for sort criteria 2 than names):
```
FL Studio                            ImageLine.FLStudio                    20.8.3.2304   Tag: digital audio workstation
```
This would actually be placed before "Git" in the results with name sorting only.  I would imagine that the large majority of users searching for "git" are not looking for this at all though.  These 4 also fall into the same bucket:
```
Duplicate Cleaner Pro                DigitalVolcanoSoftware.DuplicateClea… 4.1.4
Duplicate Cleaner Free               DigitalVolcanoSoftware.DuplicateClea… 4.1.2
APDigitalExams                       CollegeBoard.APDigitalExams           0.9.1
2021 Digital AP Exams                CollegeBoard.2021DigitalAPExams       0.9.4.5
```
If users still want results to be sorted purely based on name, I would think it best to have a setting rather than change the default  (or have a setting and change the default).

denelon:
[...]

PR #1205, opened 2021-06-22
Sort the output of "winget search" and other commands by name

chausner:
This implements #142 by extending `TableOutput` to optionally do a case-insensitive sort on one of the columns. I enabled sorting not only for the results of `winget search` as suggested in the issue but also for a few other outputs where I deemed it appropriate.

Since `TableOutput` needs to fully buffer all table rows in memory before it can sort and output the table, the `sizingBuffer` argument of the constructor has no effect when `sortColumn` is set, so at the moment I am just passing zero there.

In case sorting performance turns out to be an issue for large tables (for some reason, I was not able to build in Release mode so could not measure the performance difference), we could implement `Utility::ICUCaseInsensitiveCompare` more intelligently as it is called twice in the sort predicate. I left it simple for now.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1205)

### DIFF
--- a/src/AppInstallerCLICore/TableOutput.h
+++ b/src/AppInstallerCLICore/TableOutput.h
@@ -4,7 +4,9 @@
 #include "ExecutionReporter.h"
 #include "Resources.h"
 
+#include <algorithm>
 #include <array>
+#include <optional>
 #include <ostream>
 #include <string>
 #include <vector>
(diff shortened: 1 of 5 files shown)


doctordns:
I can't help thinking that building in formatting INTO Winget is just the wrong answer. It is re-inventing wheels that have been in use for nearly 2 decades.  Instead of wasting time, why can't I do this:
```powershell
Get-Winget-Package | Format-Table name, hash
```

Please do not reinvent perfectly usable functionality, just because you can. The sooner you get onto the cmdlet bandwagon the better as it stops you from reinventing wheels and more.

denelon:
@doctordns, I think we will likely keep a table for the `winget` CLI output as that's our current convention, but I agree 100% with the PowerShell implementation. We're discussing how we want to have PowerShell objects output. One possibility that has been discussed is an optional --json parameter to have the client format output as JSON so the PowerShell wrapper can use the "same" output from the client or the COM interface.

JohnMcPMS:
Closing due to inactivity.

PR #1216, opened 2021-06-24
Add Microsoft Store REST Source as default option and fix telemetry gaps

sachintaMSFT:
Add Microsoft Store REST Source as default option
Fix gaps found in telemetry from data validation

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1216)

### DIFF
--- a/src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
+++ b/src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
@@ -179,6 +179,8 @@ namespace AppInstaller::Logging
             packageVersion = Runtime::GetPackageVersion();
         }
 
+        Logging::SetActivityId();
+
         if (IsTelemetryEnabled())
         {
             AICLI_TraceLoggingWriteActivity(
--- a/src/AppInstallerRepositoryCore/RepositorySource.cpp
+++ b/src/AppInstallerRepositoryCore/RepositorySource.cpp
@@ -51,6 +51,10 @@ namespace AppInstaller::Repository
         {
             // If true, this is a tombstone, marking the deletion of a source at a lower priority origin.
             bool IsTombstone = false;
+
+            SourceDetailsInternal() = default;
+
+            SourceDetailsInternal(SourceDetails details) : SourceDetails(details) {};
         };
 
         // Checks whether a default source is enabled with the current settings.
@@ -405,14 +409,7 @@ namespace AppInstaller::Repository
             {
                 if (IsWingetCommunityDefaultSourceEnabled())
                 {
-                    SourceDetailsInternal details;
-                    details.Name = s_Source_WingetCommunityDefault_Name;
-                    details.Type = Microsoft::PreIndexedPackageSourceFactory::Type();
-                    details.Arg = s_Source_WingetCommunityDefault_Arg;
-                    details.Data = s_Source_WingetCommunityDefault_Data;
-                    details.Identifier = s_Source_WingetCommunityDefault_Identifier;
-                    details.TrustLevel = SourceTrustLevel::Trusted | SourceTrustLevel::StoreOrigin;
-                    result.emplace_back(std::move(details));
+                    result.emplace_back(std::move(GetWellKnownSourceDetails(WellKnownSource::WinGet)));
                 }
 
                 if (IsWingetMSStoreDefaultSourceEnabled())
@@ -429,13 +426,7 @@ namespace AppInstaller::Repository
 
                 if (IsMSStoreDefaultSourceEnabled())
                 {
-                    SourceDetailsInternal details;
-                    details.Name = s_Source_MSStoreDefault_Name;
-                    details.Type = Rest::RestSourceFactory::Type();
-                    details.Arg = s_Source_MSStoreDefault_Arg;
-                    details.Identifier = s_Source_MSStoreDefault_Identifier;
-                    details.TrustLevel = SourceTrustLevel::Trusted;
-                    result.emplace_back(std::move(details));
+                    result.emplace_back(std::move(GetWellKnownSourceDetails(WellKnownSource::MicrosoftStore)));
                 }
             }
             break;
@@ -962,7 +953,11 @@ namespace AppInstaller::Repository
                             result.SourcesWithUpdateFailure.emplace_back(source);
                         }
                     }
-                    aggregatedSource->AddAvailableSource(CreateSourceFromDetails(source, progress));
+
+                    if (source.get().Restricted != true)
+                    {
+                        aggregatedSource->AddAvailableSource(CreateSourceFromDetails(source, progress));
+                    }
                 }
 
                 if (sourceUpdated)
@@ -1101,7 +1096,8 @@ namespace AppInstaller::Repository
             details.Type = Rest::RestSourceFactory::Type();
             details.Arg = s_Source_MSStoreDefault_Arg;
             details.Identifier = s_Source_MSStoreDefault_Identifier;
-            details.TrustLevel = SourceTrustLevel::Trusted | SourceTrustLevel::StoreOrigin;
+            details.TrustLevel = SourceTrustLevel::Trusted;
+            details.Restricted = true;
             return details;
         }
         }
--- a/src/AppInstallerRepositoryCore/RepositorySource.cpp
+++ b/src/AppInstallerRepositoryCore/RepositorySource.cpp
@@ -926,7 +926,15 @@ namespace AppInstaller::Repository
             else if (currentSources.size() == 1)
             {
                 AICLI_LOG(Repo, Info, << "Default source requested, only 1 source available, using the only source: " << currentSources[0].get().Name);
-                return OpenSource(currentSources[0].get().Name, progress);
+                // Restricted sources may not support the full set of functionality
+                if (currentSources[0].get().Restricted)
+                {
+                    return {};
+                }
+                else
+                {
+                    return OpenSource(currentSources[0].get().Name, progress);
+                }
             }
             else
             {
--- a/src/AppInstallerRepositoryCore/RepositorySource.cpp
+++ b/src/AppInstallerRepositoryCore/RepositorySource.cpp
@@ -54,7 +54,7 @@ namespace AppInstaller::Repository
 
             SourceDetailsInternal() = default;
 
-            SourceDetailsInternal(SourceDetails details) : SourceDetails(details) {};
+            SourceDetailsInternal(const SourceDetails& details) : SourceDetails(details) {};
         };
 
         // Checks whether a default source is enabled with the current settings.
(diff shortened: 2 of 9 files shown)


src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -179,6 +179,8 @@ namespace AppInstaller::Logging
             packageVersion = Runtime::GetPackageVersion();
         }
 
+        Logging::SetActivityId();
```
JohnMcPMS, line 182: Why is this moved here?

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -417,6 +426,16 @@ namespace AppInstaller::Repository
                     details.TrustLevel = SourceTrustLevel::Trusted | SourceTrustLevel::StoreOrigin;
                     result.emplace_back(std::move(details));
                 }
+
+                if (IsMSStoreDefaultSourceEnabled())
```
JohnMcPMS, line 430: I don't want this included in the default list if it doesn't support search.  It also needs to be defined in the appropriate "get details" rather than here.

The easiest way to make it work all up would be a new bool on the details that allows it to be marked as hidden (maybe a better word to indicate that it requires direct targeting), then filter that out when creating the default composite source and returning details.  That lets it be targeted directly without making it part of the default set, as well as being able to update the details if needed.

@denelon, what would be a good word to describe a source that is not included with `winget search foo`, but can still be accessed via `winget search foo -s source`?  Hidden?  Shadow?  Explicit?

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -417,6 +426,16 @@ namespace AppInstaller::Repository
...
                     result.emplace_back(std::move(details));
                 }
+
+                if (IsMSStoreDefaultSourceEnabled())
+                {
+                    SourceDetailsInternal details;
+                    details.Name = s_Source_MSStoreDefault_Name;
+                    details.Type = Rest::RestSourceFactory::Type();
+                    details.Arg = s_Source_MSStoreDefault_Arg;
+                    details.TrustLevel = SourceTrustLevel::Trusted | SourceTrustLevel::StoreOrigin;
```
JohnMcPMS, line 436: `StoreOrigin` is only useful for packaged index sources.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -417,6 +426,16 @@ namespace AppInstaller::Repository
                     details.TrustLevel = SourceTrustLevel::Trusted | SourceTrustLevel::StoreOrigin;
                     result.emplace_back(std::move(details));
                 }
+
+                if (IsMSStoreDefaultSourceEnabled())
+                {
+                    SourceDetailsInternal details;
```
JohnMcPMS, line 432: Need to set the `Id`.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
sachintaMSFT: Earlier commit missed setting activity Id for winget-cli. I did that in our preview package, though. So, commiting it here and Instead of setting ActivityID at two different spots for COM and cli paths, I felt it's better to do at one spot, here.

sreadingMSFT:
> 

Probably might as well pull in the com interface changes as well in this checkin.

---
Refers to: src/AppInstallerRepositoryCore/RepositorySource.cpp:1098 in 5e6c3af. [](commit_id = 5e6c3afb0962446db1e9ac0ef476150a6f8aab77, deletion_comment = False)

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
JohnMcPMS: It is probably fine for now; we should just make it part of the object construction in the "make not global" PR.

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: I think I like `Restricted` the most from @denelon 's suggestions.  This obviously isn't permanent anyway, but let's use it as the starting point.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
sachintaMSFT: Sure. That is the plan with that pending code change.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -417,6 +426,17 @@ namespace AppInstaller::Repository
...
+
+                if (IsMSStoreDefaultSourceEnabled())
+                {
+                    SourceDetailsInternal details;
+                    details.Name = s_Source_MSStoreDefault_Name;
+                    details.Type = Rest::RestSourceFactory::Type();
+                    details.Arg = s_Source_MSStoreDefault_Arg;
+                    details.Identifier = s_Source_MSStoreDefault_Identifier;
+                    details.TrustLevel = SourceTrustLevel::Trusted;
+                    result.emplace_back(std::move(details));
```
JohnMcPMS, line 438: This should all come from just calling `GetWellKnownSourceDetails` rather than being implemented twice.

JohnMcPMS (changes requested):
We also need to make the COM API work by default on Win11 insider builds to continue to enable seemless integration.  I would suggest just making the experimental feature return true if the OS version is high enough (there is a function to check that in Runtime).

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1073,6 +1094,17 @@ namespace AppInstaller::Repository
...
             return details;
         }
+        case WellKnownSource::MicrosoftStore:
+        {
+            SourceDetailsInternal details;
+            details.Name = s_Source_MSStoreDefault_Name;
+            details.Type = Rest::RestSourceFactory::Type();
+            details.Arg = s_Source_MSStoreDefault_Arg;
+            details.Identifier = s_Source_MSStoreDefault_Identifier;
+            details.TrustLevel = SourceTrustLevel::Trusted | SourceTrustLevel::StoreOrigin;
```
sreadingMSFT, line 1104: > | SourceTrustLevel::StoreOrigin [](http://example.com/codeflow?start=58&length=32)

Should remove it here too.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1073,6 +1089,18 @@ namespace AppInstaller::Repository
             details.TrustLevel = SourceTrustLevel::Trusted | SourceTrustLevel::StoreOrigin;
             return details;
         }
+        case WellKnownSource::MicrosoftStore:
+        {
+            SourceDetailsInternal details;
+            details.Name = s_Source_MSStoreDefault_Name;
```
JohnMcPMS, line 1095: Should have `Origin` set.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -84,6 +92,11 @@ namespace AppInstaller::Repository
             return IsDefaultSourceEnabled(s_Source_WingetMSStoreDefault_Name, ExperimentalFeature::Feature::ExperimentalMSStore, onlyExplicit, TogglePolicy::Policy::MSStoreSource);
         }
 
+        bool IsMSStoreDefaultSourceEnabled(bool onlyExplicit = false)
+        {
+            return IsDefaultSourceEnabled(s_Source_MSStoreDefault_Name, ExperimentalFeature::Feature::None, onlyExplicit, TogglePolicy::Policy::MSStoreSource);
```
JohnMcPMS, line 97:
```suggestion
            return IsDefaultSourceEnabled(s_Source_MSStoreDefault_Name, ExperimentalFeature::Feature::ExperimentalMSStore, onlyExplicit, TogglePolicy::Policy::MSStoreSource);

```

JohnMcPMS (changes requested):
Per discussion, make the source part of the store EF.

The problem with this is it more clearly shows the issue with `OpenSourceFromDetails`.  That function needs to go away, but a larger redesign of the interaction is needed.

You will need to make `OpenSourceFromDetails` handle not finding the source that it is looking for.  Right now I would suggest that be by simply skipping the update step.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -926,6 +937,11 @@ namespace AppInstaller::Repository
                 bool sourceUpdated = false;
                 for (auto& source : currentSources)
                 {
+                    if (source.get().Restricted == true)
```
sreadingMSFT, line 940: > if (source.get().Restricted == true) [](http://example.com/codeflow?start=19&length=37)

Can you add a comment here like:
// Restricted sources may not support the full set of functionality so they shouldn't be included in the default aggregated source.

sreadingMSFT (approved):
:shipit:

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1026,7 +1039,8 @@ namespace AppInstaller::Repository
                 AICLI_LOG(Repo, Info, << "Named source no longer found. Source was tombstoned: " << details.Name);
                 return {};
             }
-            if (ShouldUpdateBeforeOpen(*source))
+            if (source->Restricted != true &&
```
JohnMcPMS, line 1042: nit: binary operators on booleans are more difficult to understand (for me at least)
```
source->Restricted != true
```
vs
```
!source->Restricted
```

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1026,7 +1039,8 @@ namespace AppInstaller::Repository
                 AICLI_LOG(Repo, Info, << "Named source no longer found. Source was tombstoned: " << details.Name);
```
JohnMcPMS, line 1039: I know this isn't part of your change, but I now realize that this results in the ability to remove the store source despite the fact that we want the well known sources to *always* be present.  I would suggest removing the Tombstone block.

JohnMcPMS (approved):

mohammedpanamai:
> Add Microsoft Store REST Source as default option Fix gaps found in telemetry from data validation
> 
> * [ ]  Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
> * [ ]  Are you working against an Issue?
> 
> ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1216)

PR #1275, opened 2021-07-13
Upgrade check-spelling to v0.0.19

jsoref:

-----

* general expect maintenance for drift
  * expect jitter should go away in a version in the near future (I think I have that in my pending stack.
* excludes are sorted here as a near future version will automatically sort them just as it maintains expect.
* some minor template updates are included (I'm skipping adding a README and advice).
* `shortest_word: 2` is set for compatibility with 0.0.17. As of 0.0.18, the default is `3` as two letter items have turned out to be more noise than useful value. But moving to `3` is a bit noisy, so, I'm leaving that for later.

v0.0.19 is really a minor bump to fix:
https://github.com/check-spelling/check-spelling/security/advisories/GHSA-g86g-chm8-7r2p

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1275)


.github/workflows/spelling2.yml
```diff
@@ -0,0 +1,27 @@
+# spelling.yml is blocked per https://github.com/check-spelling/check-spelling/security/advisories/GHSA-g86g-chm8-7r2p
```
denelon, line 1: I tried to open the URL, but got a 404 error.

jsoref: I'm waiting to publish the GHSA until after I get the majority of users updated.

JohnMcPMS (approved):

JohnMcPMS:
After merging this, is it safe to re-enable the action?  Will the existing branches continue to run the older version, or is `spelling.yaml` truly blocked?

jsoref:
You should leave `spelling.yml` blocked forever, merging this with the new name should automatically start it running. There are too many branches in this repository for it to make sense to try to upgrade all branches with the old file name.

Going forward, I have a plan to for a way to make this migration problem more manageable, but I haven't written the code yet.

JohnMcPMS:
> You should leave `spelling.yml` blocked forever

Where is that file blocked?  If it is on your side, that is great.  If it is on ours then I want to know which bug red button to not press 😃

jsoref:
It's on your side:
https://github.com/microsoft/winget-cli/actions/workflows/spelling.yml
image

And yes, it's kinda unfortunate that the button is quite large and clickable. Thankfully you're unlikely to visit this page. Although it's definitely an attractive nuisance. I'm trying to give feedback to GitHub about my experience too...

PR #1336, opened 2021-08-02
Allow Restricted source be always available to callers

sachintaMSFT:
1. Restricted source to always be available to Store even if not visible in winget list
2. Fix to set ExecutionStage to ensure telemetry reports it correct

### DIFF
--- a/src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
@@ -2,10 +2,8 @@
 // Licensed under the MIT License.
 #include "pch.h"
 #include "COMInstallCommand.h"
-#include "Workflows/CompletionFlow.h"
 #include "Workflows/InstallFlow.h"
 #include "Workflows/WorkflowBase.h"
-#include "Resources.h"
 
 using namespace AppInstaller::CLI::Execution;
 using namespace AppInstaller::CLI::Workflow;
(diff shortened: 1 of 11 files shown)


src/AppInstallerCLICore/Commands/COMInstallCommand.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "Command.h"
+
+namespace AppInstaller::CLI
+{
+    struct COMInstallCommand final : public Command
```
yao-msft, line 8: > COMInstallCommand  [](http://example.com/codeflow?start=10&length=19)

I think we should better extend the current Install command(and other commands COM supports in the future, but not now) to have a flag indicating "it's from COM" rather than creating a duplicate command here.

By the way, I need that flag too with my current work to support source agreements

yao-msft: or maybe even simpler, something like IsCom() exposed in context if it works

src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -197,6 +197,8 @@ namespace AppInstaller::CLI::Execution
 
     void Context::SetExecutionStage(Workflow::ExecutionStage stage, bool allowBackward)
     {
+        Logging::SetExecutionStage(static_cast<uint32_t>(stage));
```
yao-msft, line 200: >Logging::SetExecutionStage(static_cast<uint32_t>(stage)); [](http://example.com/codeflow?start=8&length=57)

this should probably go to the end, after error checking

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Commands/COMInstallCommand.h
JohnMcPMS: I disagree; this type of large scale context sensitivity is what we want to avoid with workflows.  I suggested potentially making a different base object to hold the concept of a composed workflow though.

I'm not following the part about source agreements though.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1025,7 +1025,9 @@ namespace AppInstaller::Repository
         // Get the details again by name from the source list because SaveMetadata only updates the LastUpdateTime
         // if the details came from the same instance of the list that's being saved.
         // Some sources that do not need updating like the Installed source, do not have Name values.
-        if (!details.Name.empty())
+        // Restricted sources don't have full functionality
+        if (!details.Name.empty() &&
+            !details.Restricted)
```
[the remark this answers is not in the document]
sreadingMSFT: Problem being addressed is: not all sources that can be opened appear in the sourcelist (specifically the one source that is marked as Restricted right now). Constraint is: I think we do want to fail in some cases if the source is not in the source list any more - i.e. scenario: App caller gets list of sources on app open, user removes a source from the command line, app caller eventually tries to Connect to the removed source but it's no longer in the source list. Decision has to be made here whether to fail or succeed. If succeed is ok, then this code can just continue to do the Open even if it doesn't find it in the list. If it needs to fail, then we need a way of knowing whether it's expected that the source isn't in the source list. Right now Restricted sources are expected to not be in the source list, and not need to be updated..

src/AppInstallerCLICore/ExecutionContext.cpp
sachintaMSFT: Thinking was that telemetry should reflect the reality as is even in error cases. But I guess we are more interested in knowing the last valid execution stage from failure telemetry events rather than the invalid execution stage it is asked to set to.

src/AppInstallerCLICore/Commands/COMInstallCommand.h
yao-msft: Chatted with John offline, I was not aware of the required differences between COM workflow and the regular flow. I got the wrong first impression by the name ComInstallCommand, instead it's sort of a different set of workflow tasks.

I think we should clean up any nonapplicable code from this file and maybe name it differently?

Regarding source agreements, we'll use a specific flag say SKIP_AGREEMENTS in context flags.

src/AppInstallerCLICore/ExecutionContext.cpp
yao-msft: yep, we care more about last valid stage so I think it should be after error checking?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -648,6 +648,12 @@ namespace AppInstaller::CLI::Workflow
         ReportIdentity(context, manifest.CurrentLocalization.Get<Manifest::Localization::PackageName>(), manifest.Id);
     }
 
+    void LogManifestFields(Execution::Context& context)
+    {
+        const auto& manifest = context.Get<Execution::Data::Manifest>();
+        Logging::Telemetry().LogManifestFields(manifest.Id, manifest.DefaultLocalization.Get<Manifest::Localization::PackageName>(), manifest.Version);
```
yao-msft, line 654: >LogManifestFields [](http://example.com/codeflow?start=29&length=17)

nit: usually this gets logged when manifest is generated(you can find them in all other places). I think we could remove this task and just do the same in PackageManager.cpp as well, or at least update all other places to call this workflow task.

src/AppInstallerCLICore/Commands/COMInstallCommand.h
sachintaMSFT: @JohnMcPMS, @yao-msft,  does the more precise InstallFromManifestCommand suit better for the name ?

src/AppInstallerRepositoryCore/RepositorySource.cpp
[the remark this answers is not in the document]
sachintaMSFT: Yes. sreadingMSFT said it right. Since winget list should not display Restricted source, it is not available in SourceList and hence, it fails there without this change.

src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
```diff
@@ -14,114 +12,11 @@ using namespace AppInstaller::Utility::literals;
...
-        if (execArgs.Contains(Args::Type::Locale))
-        {
-            if (!Locale::IsWellFormedBcp47Tag(execArgs.GetArg(Args::Type::Locale)))
-            {
-                throw CommandException(Resource::String::InvalidArgumentValueErrorWithoutValidValues, Argument::ForType(Args::Type::Locale).Name(), {});
-            }
-        }
-    }
-
+    // IMPORTANT: To use this command, the caller should have already retrieved the package manifest (GetManifest()) and added it to the Context Data
```
JohnMcPMS, line 15: I really meant a very precise set of context data items; this does not specify that.  It simply says "put a manifest in" which could theoretically be into any data item that can contain a manifest.  There could be more than one in the future.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -193,10 +201,6 @@ namespace winrt::Microsoft::Management::Deployment::implementation
             context.SetLoggerContext(options.CorrelationData(), ::AppInstaller::Utility::ConvertToUTF8(callerProcessInfoString));
 
             // Convert the options to arguments for the installer.
-            context.Args.AddArg(::AppInstaller::CLI::Execution::Args::Type::Id, ::AppInstaller::Utility::ConvertToUTF8(package.Id()));
-            context.Args.AddArg(::AppInstaller::CLI::Execution::Args::Type::Version, ::AppInstaller::Utility::ConvertToUTF8(packageVersionInfo.Version()));
-            context.Args.AddArg(::AppInstaller::CLI::Execution::Args::Type::Channel, ::AppInstaller::Utility::ConvertToUTF8(packageVersionInfo.Channel()));
-            context.Args.AddArg(::AppInstaller::CLI::Execution::Args::Type::Source, ::AppInstaller::Utility::ConvertToUTF8(packageVersionInfo.PackageCatalog().Info().Name()));
             context.Args.AddArg(::AppInstaller::CLI::Execution::Args::Type::Exact);
```
JohnMcPMS, line 204: This one can be removed as well.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -240,9 +244,7 @@ namespace winrt::Microsoft::Management::Deployment::implementation
                 }
             }
 
-            ::AppInstaller::Manifest::Manifest manifest;
-            manifest = GetManifestFromPackageVersionInfo(packageVersionInfo);
-            context.Add<::AppInstaller::CLI::Execution::Data::Manifest>(manifest);
+            AddPackageManifestToContext(packageVersionInfo, context);
```
JohnMcPMS, line 247: Adding the PackageVersion will become necessary to support dependencies in the future; maybe my last comment shouldn't have been a "nit" so much as a "forward facing request".

JohnMcPMS (approved):
The code is fine for now to get the fix checked in, but these changes would be appreciated at some point.

PR #1346, opened 2021-08-05
Fix WinGetDev guids to prevent collision when DesktopAppInstaller is also installed.

sreadingMSFT:

-----

Users were hitting occasional collisions when both wingetdev and desktopappinstaller were installed due to both packages registering both the dev and prod guids once they ran, even though each package only put one set of guids in the manifest. This change uses build props to set a value so that only one set is defined and the server only registers the set that its package declares.  It will need to be done in sync with a change in DesktopAppInstaller source which I've tested that adds a Directory.Build.props file in the root to define the UseProdCLSIDs parameter so that the build for that package uses the other set of guids.

Since the currently shipping version of the desktopappinstaller package is still registering the existing dev guids, I've replaced all the dev guids, which makes this a breaking change for com callers expecting to use the com interface from the wingetdev package. I don't think there are any of those users at the moment though. If the guids can't be updated it will take longer to resolve the issue since it would require everyone getting a new flight of DesktopAppInstaller.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1346)


JohnMcPMS (approved):
I don't think that the CLSID changes were actually necessary; but you are correct that they will speed up our ability to leverage this change (so do keep them since you already did the change).

PR #1373, opened 2021-08-16
Add projected winrt wrapper for server com interface and add tests

sreadingMSFT:

-----
The server com interface is wrapped with a winrt dll that just overrides the factory implementation to create the server com classes. Given that implementation there is currently no optimization of client\server communication. If necessary that will need to be explored in a separate checkin. This change also creates the framework for adding tests, with a new test project that runs inside a package where the winrt interface can be registered for the caller. I added a few simple ones here for demonstration purposes but will do the rest in a separate checkin.


src/AppInstallerCLIE2ETests/AppInstallerCLIE2ETests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
```
sreadingMSFT, line 4: Oops. Changes in this file are from an attempt to re-use the existing tests. Will revert.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -533,5 +533,9 @@ namespace Microsoft.Management.Deployment
         interface Windows.Foundation.Collections.IVectorView<PackageVersionId>;
         interface Windows.Foundation.Collections.IVector<PackageVersionInfo>;
         interface Windows.Foundation.Collections.IVectorView<PackageVersionInfo>;
+        interface Windows.Foundation.Collections.IVector<PackageVersionId>;
```
sreadingMSFT, line 536: Bad merge I guess. Will revert.

src/AppInstallerCLI.sln
```diff
@@ -145,20 +150,26 @@ Global
 		ManifestSchema\ManifestSchema.vcxitems*{fb313532-38b0-4676-9303-ab200aa13576}*SharedItemsImports = 4
 	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
JohnMcPMS, line 153: We don't build "Any CPU", so I'm not sure how this is going to work.

We do have C# projects that are built without it though; if there are problems you should be able to follow the E2E tests (among others).

src/Microsoft.Management.Deployment.Client/readme.txt
```diff
@@ -0,0 +1,23 @@
+========================================================================
+    C++/WinRT Microsoft.Management.Deployment Project Overview
```
JohnMcPMS, line 2: Remove (and from the server project if it made it in there).

src/PackagedTests/ComInterfaceUnitTest.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.Management.Deployment;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace PackagedUnitTests
+{
+    [TestClass]
+    public class ComInterfaceUnitTests
+    {
+        [TestMethod]
+        public void OpenPredefinedCatalog()
```
JohnMcPMS, line 13: Hopefully a few more tests could be added leveraging the existing local source infrastructure to exercise the API more extensively.

JohnMcPMS (changes requested):

src/WinGetServer/WinGetServer.vcxproj
```diff
@@ -1,176 +1,176 @@
...
+      <AdditionalIncludeDirectories>$(ProjectDir)..\AppInstallerCLICore;$(ProjectDir)..\AppInstallerRepositoryCore;$(ProjectDir)..\AppInstallerRepositoryCore\Public;$(ProjectDir)..\AppInstallerCommonCore\Public;$(ProjectDir)..\JsonCppLib\json;$(ProjectDir)..\JsonCppLib;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+    </ClCompile>
+    <Link>
+      <SubSystem>Windows</SubSystem>
+      <GenerateWindowsMetadata>false</GenerateWindowsMetadata>
+      <AdditionalLibraryDirectories>$(OutDir)..\Microsoft.Management.Deployment;$(OutDir)..\AppInstallerCLICore;$(OutDir)..\JsonCppLib;$(OutDir)..\AppInstallerRepositoryCore;$(OutDir)..\YamlCppLib;$(OutDir)..\AppInstallerCommonCore;$(OutDir)..\cpprestsdk;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalDependencies>Microsoft.Management.Deployment.Server.lib;AppInstallerCLICore.lib;AppInstallerCommonCore.lib;AppInstallerRepositoryCore.lib;JsonCppLib.lib;YamlCppLib.lib;cpprestsdk.lib;wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;winhttp.lib;onecoreuap.lib;msi.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
sreadingMSFT, line 109: >Server [](http://example.com/codeflow?start=62&length=6)

This change probably isn't necessary anymore, was doing it when trying to build the test project as part of the same solution. But I do still think it's clearer this way than having two libs with the same name floating around.

PR #1385, opened 2021-08-20
Fix GetInstallProgress to return null directly when an install is not in progress for that package

sreadingMSFT:

-----
This api is doc'd as returning null directly when there's no install in progress. To do that synchronously I separated out the lock in the orchestrator into an inner and outer lock, so that the GetInstallProgress calls will not be blocked on callers making calls to the SQL index.
Since I was already moving the com capability check code I also fixed an issue where callers with PackageQuery only, and not PackageManagement, were able to Cancel the async operations.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1385)


src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -439,24 +417,38 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            bool cancelWaitEventFired = false;
+            HANDLE operationEvents[3];
             operationEvents[0] = progressEvent.get();
             operationEvents[1] = queueItem->GetCompletedEvent().get();
-            while (!completionEventFired)
+            operationEvents[2] = cancelWaitEvent.get();
+            while (!completionEventFired && !cancelWaitEventFired)
             {
                 DWORD dwEvent = WaitForMultipleObjects(
                     2 /* number of events */,
```
JohnMcPMS, line 454: Maybe that should be `ARRAY_SIZE(operationEvents)`.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -481,30 +473,97 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+    winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::InstallResult, winrt::Microsoft::Management::Deployment::InstallProgress> GetEmptyAsynchronousResultForInstallOperation(
+        HRESULT hr,
+        hstring correlationData)
+    {
+        // If a function uses co_await or co_return (i.e. if it is a co_routine), it cannot use return directly.
+        // This helper helps a function that is not a coroutine itself to return errors asynchronously.
+        co_return GetInstallResult(::Workflow::ExecutionStage::Initial, hr, correlationData, false);
+    }
+
+#define WINGET_RETURN_INSTALL_RESULT_HR_IF(hr, boolVal) { if(boolVal) { return GetEmptyAsynchronousResultForInstallOperation(hr, correlationData); }}
```
JohnMcPMS, line 509: Why make a macro that wraps a function?  It seems like this could just be:
```suggestion
#define WINGET_RETURN_INSTALL_RESULT_HR_IF(hr, boolVal) { if(boolVal) { co_return GetInstallResult(::Workflow::ExecutionStage::Initial, hr, correlationData, false); }}
```

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -481,30 +473,97 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            // This must be done before any co_awaits since it requires info from the rpc caller thread.
+            auto [hrGetCallerId, callerProcessId] = GetCallerProcessId();
+            WINGET_RETURN_INSTALL_RESULT_HR_IF_FAILED(hrGetCallerId);
+            WINGET_RETURN_INSTALL_RESULT_HR_IF_FAILED(EnsureProcessHasCapability(Capability::PackageManagement, callerProcessId));
+            callerProcessInfoString = TryGetCallerProcessInfo(callerProcessId);
+        }
+        WINGET_CATCH_STORE(hr, APPINSTALLER_CLI_ERROR_COMMAND_FAILED);
+        WINGET_RETURN_INSTALL_RESULT_HR_IF_FAILED(hr);
+
+        return GetInstallOperation(true /*canCancelQueueItem*/, nullptr /*queueItem*/, package, options, std::move(callerProcessInfoString));
```
JohnMcPMS, line 533: I assume if you took my suggestion above this would have to become `co_return`, or `co_return co_await`.  I'm not super familiar with how the winrt machinery handles that.  But I can see why you did what you did above here, so you can just no-op these two comments if you want.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -481,30 +473,97 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
     }
 
     winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::InstallResult, winrt::Microsoft::Management::Deployment::InstallProgress> PackageManager::GetInstallProgress(winrt::Microsoft::Management::Deployment::CatalogPackage package, winrt::Microsoft::Management::Deployment::PackageCatalogInfo catalogInfo)
     {
-        return GetInstallOperation(false, package, nullptr, catalogInfo);
+        hstring correlationData;
+        WINGET_RETURN_INSTALL_RESULT_HR_IF(APPINSTALLER_CLI_ERROR_INVALID_CL_ARGUMENTS, package == nullptr);
```
JohnMcPMS, line 539:
```suggestion
        WINGET_RETURN_INSTALL_RESULT_HR_IF(APPINSTALLER_CLI_ERROR_INVALID_CL_ARGUMENTS, !package);
```
To stay consistent with the check in the function above.

JohnMcPMS (changes requested):

src/Microsoft.Management.Deployment/PackageManager.cpp
sreadingMSFT: :|

src/AppInstallerCLICore/ContextOrchestrator.cpp
```diff
@@ -42,14 +42,20 @@ namespace AppInstaller::CLI::Execution
...
+
+            THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INSTALL_ALREADY_RUNNING), FindById(item->GetId()));
+            m_queueItems.push_back(item);
+        }
 
         // Add the package to the Installing source so that it can be queried using the ISource interface.
         const auto& manifest = item->GetContext().Get<Execution::Data::Manifest>();
         m_installingWriteableSource->AddPackageVersion(manifest, std::filesystem::path{ manifest.Id + '.' + manifest.Version });
+
+        item->GetContext().EnableCtrlHandler();
```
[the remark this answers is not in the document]
sreadingMSFT: Ok. We'll want to fix that soon to enable the multiple downloads I guess. Reverting.

src/AppInstallerCLICore/ContextOrchestrator.cpp
```diff
@@ -42,14 +42,20 @@ namespace AppInstaller::CLI::Execution
     
     void ContextOrchestrator::EnqueueItem(std::shared_ptr<OrchestratorQueueItem> item)
     {
-        std::lock_guard<std::mutex> lock{ m_queueLock };
+        std::lock_guard<std::mutex> lock{ m_queueOuterLockForInstallingSource };
```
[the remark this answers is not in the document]
sreadingMSFT: Reworked this to remove the double lock.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -366,45 +366,29 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
-        #define WINGET_RETURN_INSTALL_RESULT_HR(hr) { WINGET_RETURN_INSTALL_RESULT_IF(GetInstallResult(executionStage, hr, correlationData, false), true) }
-        #define WINGET_RETURN_INSTALL_RESULT_HR_IF(hr, boolVal) { if(boolVal) { WINGET_RETURN_INSTALL_RESULT_HR(hr) }}
-        #define WINGET_RETURN_INSTALL_RESULT_HR_IF_FAILED(hr) { WINGET_RETURN_INSTALL_RESULT_HR_IF(hr, FAILED(hr)) }
-
-        // options and catalog can both be null, package must be set.
-        WINGET_RETURN_INSTALL_RESULT_HR_IF(APPINSTALLER_CLI_ERROR_INVALID_CL_ARGUMENTS, !package);
-
         try
         {
+            // re-scope the parameter to inside the try block to avoid lifetime management issues.
```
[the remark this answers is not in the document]
sreadingMSFT: It won't be cleaned up as long as the coroutine holds it, it holds the context, and the context holds the callback back into this coroutine for progress. coroutine has to release it before returning.

JohnMcPMS (approved):

PR #1395, opened 2021-08-24
Add version comparison and architecture control to COM API

JohnMcPMS:
## Change
This change adds an API, `CompareToVersion`, to compare a version string against a `PackageVersionInfo`.  The result is a value that indicates how the `PackageVersionInfo` object's version relates to the given value.  Thus if you want to check if the installed version is at least some value:
```C++
if (packageVersionInfo.CompareToVersion(targetVersion) == CompareResult::Lesser)
{
// Time to upgrade
}
```

It also adds a property to `InstallOptions`; `AllowedArchitectures`.  This vector of `ProcessorArchitecture` values contains both the allowed installer architectures that may be chosen when installing, as well as the order in which to prioritize the choice.  The value `Unknown` may be used to place the remaining supported architectures for the machine afterward (in the default preference order).  For example, `{ X86, Unknown }` would expand to `{ X86, X64, Neutral }` on an X64 machine.

This change also fixes the default scope for `InstallOptions` to be `Any` as it is for the command line.

Finally, a new well known source is added that will contain a controlled set of "framework" packages for future use.  The packages within this source are not yet stabilized, so taking a dependency on it is not supported yet.

## Validation
As we do no yet have the COM interface test infrastructure, automated tests were only added for the installer selection code.  Manual validation was done on both APIs.


src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -891,6 +967,11 @@ namespace AppInstaller::Repository
         auto source = sourceList.GetCurrentSource(name);
         THROW_HR_IF(APPINSTALLER_CLI_ERROR_SOURCE_NAME_ALREADY_EXISTS, source != nullptr);
 
+        // Check for a non-user tombstone; hidden source data that we don't want to collide.
```
yao-msft, line 970: >non-user tombstone [](http://example.com/codeflow?start=23&length=18)

I don't remember we have code that supports user tombstone source, would just calling sourceList.GetSource be enough?

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -480,6 +506,16 @@ namespace Microsoft.Management.Deployment
         String CorrelationData;
         /// A string that will be passed to the source server if using a REST source
         String AdditionalPackageCatalogArguments;
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 2)]
+        {
+            // The set of allowed Architectures, in preference order, that will be considered for
+            // the install operation.  The default set is empty, which allows any architecture
+            // to be used.  Unknown can also be used to set a preference rather than a requirement,
+            // such as { X86, Unknown }, which will prefer X86 if it is available and applicable,
```
yao-msft, line 515: >Unknown [](http://example.com/codeflow?start=30&length=7)

The name "Unknown" seems not accurate to denote "I'm ok with other defaults" and conflicts with the "Architecture::unknown"meaning in winget , looks like we are redefining the behavior of what Architecture::Unknown means

yao-msft (approved):
:shipit:

src/Microsoft.Management.Deployment/PackageManager.idl
JohnMcPMS: There is not another value within `Windows.System.ProcessorArchitecture` that is not also a value in our internal `Architecture` enum that represents a legitimate installer architecture.  `Neutral` could also work, but is an actual installer architecture.

I had originally created my own enum and used the value `Any` in this role, but I don't like creating yet another enum for this when multiple exist.

PR #1397, opened 2021-08-25
Add deferred registration for MSIX

JohnMcPMS:
## Change
Prior to this change, MSIX install (and upgrade) would forcibly shut down the application so that it could upgrade the package during the command.  After, the new package will be staged and the upgrade will complete when the application is restarted.

This is done by first attempting `RequestAddPackageAsync` if SmartScreen is necessary.  Then if it that fails due to the package being in use (or it is not needed at all), we stage (place files onto the machine) then register (install package for user).  If the package is in use at this point, we detect it and change the success message to indicate that the user must restart the application:
```
Successfully installed. Restart the application to complete the upgrade.
```

## Validation
Manual validation of both SmartScreen and non-SmartScreen paths.  Automated validation would require significant test infrastructure to launch and monitor applications while performing actions.  Existing tests also serve as regression for cases where the package is not in use.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1397)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
@@ -10,6 +10,9 @@
 #include "WorkflowBase.h"
 #include "Workflows/DependenciesFlow.h"
 
+#include <AppInstallerDeployment.h>
+#include <AppInstallerMsixInfo.h>
+
 namespace AppInstaller::CLI::Workflow
 {
     using namespace winrt::Windows::ApplicationModel::Store::Preview::InstallControl;
@@ -408,26 +411,26 @@ namespace AppInstaller::CLI::Workflow
 
     void MsixInstall(Execution::Context& context)
     {
-        Uri uri = nullptr;
+        std::string uri;
         if (context.Contains(Execution::Data::InstallerPath))
         {
-            uri = Uri(context.Get<Execution::Data::InstallerPath>().c_str());
+            uri = context.Get<Execution::Data::InstallerPath>().u8string();
         }
         else
         {
-            uri = Uri(Utility::ConvertToUTF16(context.Get<Execution::Data::Installer>()->Url));
+            uri = context.Get<Execution::Data::Installer>()->Url;
         }
 
         context.Reporter.Info() << Resource::String::InstallFlowStartingPackageInstall << std::endl;
 
+        bool registrationDeferred = false;
+
         try
         {
-            DeploymentOptions deploymentOptions =
-                DeploymentOptions::ForceApplicationShutdown |
-                DeploymentOptions::ForceTargetApplicationShutdown;
-
-            context.Reporter.ExecuteWithProgress(std::bind(Deployment::AddPackage, uri, deploymentOptions,
-                WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerTrusted), std::placeholders::_1));
+            registrationDeferred = context.Reporter.ExecuteWithProgress([&](IProgressCallback& callback)
+            {
+                return Deployment::AddPackageWithDeferredFallback(uri, WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerTrusted), callback);
+            });
         }
         catch (const wil::ResultException& re)
         {
@@ -438,7 +441,14 @@ namespace AppInstaller::CLI::Workflow
             AICLI_TERMINATE_CONTEXT(re.GetErrorCode());
         }
 
-        context.Reporter.Info() << Resource::String::InstallFlowInstallSuccess << std::endl;
+        if (registrationDeferred)
+        {
+            context.Reporter.Warn() << Resource::String::InstallFlowRegistrationDeferred << std::endl;
+        }
+        else
+        {
+            context.Reporter.Info() << Resource::String::InstallFlowInstallSuccess << std::endl;
+        }
     }
 
     void RemoveInstaller(Execution::Context& context)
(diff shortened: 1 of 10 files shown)


src/AppInstallerCLICore/pch.h
```diff
@@ -49,11 +49,9 @@
 
 #include <AppInstallerArchitecture.h>
 #include <AppInstallerDateTime.h>
-#include <AppInstallerDeployment.h>
```
yao-msft, line 52: Is there a reason we specifically move these 2 out of pch to individual files?

yao-msft (approved):
:shipit:

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1019,4 +1019,7 @@ Do you agree to the terms?</value>
   <data name="PromptOptionYes" xml:space="preserve">
     <value>Yes</value>
   </data>
-</root>
+  <data name="InstallFlowRegistrationDeferred" xml:space="preserve">
+    <value>Successfully installed. Restart the application to complete the upgrade.</value>
```
ranm-msft, line 1023: this only applies to upgrade?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -408,26 +411,26 @@ namespace AppInstaller::CLI::Workflow
 
     void MsixInstall(Execution::Context& context)
     {
-        Uri uri = nullptr;
+        std::string uri;
         if (context.Contains(Execution::Data::InstallerPath))
         {
-            uri = Uri(context.Get<Execution::Data::InstallerPath>().c_str());
+            uri = context.Get<Execution::Data::InstallerPath>().u8string();
```
ranm-msft, line 417: u8 is for utf8?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -408,26 +411,26 @@ namespace AppInstaller::CLI::Workflow
...
         }
         else
         {
-            uri = Uri(Utility::ConvertToUTF16(context.Get<Execution::Data::Installer>()->Url));
+            uri = context.Get<Execution::Data::Installer>()->Url;
         }
 
         context.Reporter.Info() << Resource::String::InstallFlowStartingPackageInstall << std::endl;
 
+        bool registrationDeferred = false;
```
ranm-msft, line 426: isn't the term deferredRegistration?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
yao-msft: I think if the package is in use, then it could only be "upgrading"

src/AppInstallerCommonCore/Deployment.cpp
```diff
@@ -94,6 +107,92 @@ namespace AppInstaller::Deployment
...
+        if (!skipSmartScreen)
+        {
+            // The only way to get SmartScreen is to use RequestAddPackageAsync, so we will have to start with that.
+            size_t id = GetDeploymentOperationId();
+            AICLI_LOG(Core, Info, << "Starting RequestAddPackageAsync operation #" << id << ": " << uri);
+
+            DeploymentOptions options = DeploymentOptions::None;
+            // Optimization to keep files if the package is in use. Only available in a newer OS per:
+            // https://docs.microsoft.com/en-us/uwp/api/Windows.Management.Deployment.DeploymentOptions
+            if (Runtime::IsCurrentOSVersionGreaterThanOrEqual(Utility::Version{ "10.0.18362.0" }))
```
ranm-msft, line 139: should we keep consts with all versions we reference and applicable names (i.e. win11 or 20H1 etc.)?

ranm-msft (approved):

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -408,26 +411,26 @@ namespace AppInstaller::CLI::Workflow
...
+
         try
         {
-            DeploymentOptions deploymentOptions =
-                DeploymentOptions::ForceApplicationShutdown |
-                DeploymentOptions::ForceTargetApplicationShutdown;
-
-            context.Reporter.ExecuteWithProgress(std::bind(Deployment::AddPackage, uri, deploymentOptions,
-                WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerTrusted), std::placeholders::_1));
+            registrationDeferred = context.Reporter.ExecuteWithProgress([&](IProgressCallback& callback)
```
ranm-msft, line 430: we only defer registration if package is in use?

src/AppInstallerCLICore/pch.h
JohnMcPMS: I've realized that even cross project it is wasteful to have our own files in the pch.  I didn't want to move everything and have the size of this change explode, so I moved the ones that I was touching to make my builds faster.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: Any install action on MSIX uses this path; I used the verb upgrade to stay consistent with our own command but it could happen on install or import as well.

src/AppInstallerCommonCore/Deployment.cpp
JohnMcPMS: Possibly.  I was intentionally tying it to the documented version in which this particular flag was made available rather than the OS release name/timeframe.  If we need this kind of check more in the future I think it would be fine to implement then.

PR #1399, opened 2021-08-26
Update Loggers to support multiple Win32 app installs in single process

sachintaMSFT:
Added support to maintain both ProcessGlobal and ThreadGlobal loggers and used them to support processing of multiple Win32 app Install requests in single WPM Server instance. Ensured Trace Provider is registered only once and unregistered only once in the lifetime of the host process.

Tested real world install of Win32 apps from Winget command line as well as from Store.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1399)


src/AppInstallerCLICore/COMContext.cpp
```diff
@@ -51,22 +51,33 @@ namespace AppInstaller::CLI::Execution
...
 
-    void COMContext::SetLoggerContext(const std::wstring_view telemetryCorrelationJson, const std::string& caller)
+    void COMContext::SetContextLoggers(const std::wstring_view telemetryCorrelationJson, const std::string& caller)
     {
         m_correlationData = telemetryCorrelationJson;
-        Logging::Telemetry().SetTelemetryCorrelationJson(telemetryCorrelationJson);
-        Logging::Telemetry().SetCaller(caller);
-        Logging::Telemetry().LogStartup(true);
+
+        std::unique_ptr<AppInstaller::ThreadLocalStorage::PreviousThreadGlobals> setThreadGlobalsToPreviousState;
```
JohnMcPMS, line 58: Two comments here:
1. It is better to use `auto` for these RAII lifetime holders as you aren't ever going to interact with them directly.  That allows the return type to change tomorrow without updating every call site.
2. You are in the type that owns the TLS object itself; there is no need to set the TLS pointer then turn around and retrieve it back.  I get that you are also reusing `SetLoggers` to initialize the global state, but I would not be shocked to have that diverge in the future.  And there are ways to reuse that code without using the TLS state.

Neither of these necessitates a change, but they are things to think about for the future.

src/AppInstallerCLICore/COMContext.cpp
```diff
@@ -51,22 +51,33 @@ namespace AppInstaller::CLI::Execution
...
+
     std::wstring_view COMContext::GetCorrelationJson()
     {
         return m_correlationData;
     }
 
+    void COMContext::SetGlobalLoggers()
+    {
+        SetLoggers();
+        AppInstaller::Logging::Telemetry().Initialize();
```
JohnMcPMS, line 75: We discussed having the global `Initialize` happen naturally as part of the TLS fallback.  What was the issue with that?

Without that change, there are missing calls to `Initialize` on the global object for the other binaries.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -122,6 +122,9 @@ namespace AppInstaller::CLI::Execution
 
         virtual void SetExecutionStage(Workflow::ExecutionStage stage, bool);
 
+        // Get Globals for Current Thread
+        AppInstaller::ThreadLocalStorage::ThreadGlobals& GetThreadGlobalsActive();
```
JohnMcPMS, line 126: This has nothing to do with threads; it is simply the `Context`'s `ThreadGlobals` object:
```suggestion
        AppInstaller::ThreadLocalStorage::ThreadGlobals& GetThreadGlobals();
```

src/AppInstallerCommonCore/Public/winget/ThreadGlobals.h
```diff
@@ -0,0 +1,52 @@
...
+        ThreadGlobals() = default;
+        ~ThreadGlobals() = default;
+
+        AppInstaller::Logging::DiagnosticLogger& GetDiagnosticLogger();
+
+        AppInstaller::Logging::TelemetryTraceLogger& GetTelemetryLogger();
+
+        // Set Globals for Current Thread
+        // Return RAII object with it's ownership to set the AppInstaller ThreadLocalStorage back to previous state
+        AppInstaller::ThreadLocalStorage::PreviousThreadGlobals* SetForCurrentThread();
```
JohnMcPMS, line 23: Here is why `auto` wouldn't work for my previous comment; this should be returning the RAII object itself (not a type that does RAII but itself has no lifetime semantic):
```suggestion
        std::unique_ptr<AppInstaller::ThreadLocalStorage::PreviousThreadGlobals SetForCurrentThread();
```

src/AppInstallerCommonCore/Public/winget/ThreadGlobals.h
```diff
@@ -0,0 +1,38 @@
...
+        // Clear Globals
+        void Clear();
+
+        // Return Globals for Current Thread
+        static ThreadGlobals* GetForCurrentThread();
+
+    private:
+
+        // Set and return Globals for Current Thread
+        static ThreadGlobals* SetOrGetThreadGlobals(bool setThreadGlobals, ThreadGlobals* pThreadGlobals = nullptr);
```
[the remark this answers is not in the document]
JohnMcPMS: No seriously; moving it to be a cpp detail also means that `PreviousThreadGlobals` doesn't need to be a friend.

src/AppInstallerCommonCore/ThreadGlobals.cpp
```diff
@@ -0,0 +1,73 @@
...
+    void ThreadGlobals::Initialize()
+    {
+        try
+        {
+            std::call_once(loggerInitOnceFlag, [this]()
+                {
+                    m_pDiagnosticLogger = std::make_unique<DiagnosticLogger>();
+                    m_pTelemetryLogger = std::make_unique<TelemetryTraceLogger>();
+
+                    if (m_pTelemetryLogger)
```
JohnMcPMS, line 36: nit: `make_unique` will throw or there will be an object; this if statement is always going to be true.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/COMContext.cpp
sachintaMSFT: The issue was that the TraceLogger write is accessing Telemetry object to log activityId as part of AICLI_LOGs which is causing an indefinitely recursion due to initialize being part of the global telemetry object's constructor. But as proposed in our call, I will instead not log activityId in TraceLogger and let it be a problem to solve for future. For now, I am making changes as per your comment.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -551,8 +559,8 @@ namespace AppInstaller::Logging
         }
         else
         {
-            static TelemetryTraceLogger processGlobalTelemetry;
-            return processGlobalTelemetry;
+            static GlobalTelemetryTraceLogger processGlobalTelemetry;
```
JohnMcPMS, line 562: You made things so much more complicated than necessary, and still don't seem to be calling `Initialize`.  This would have been sufficient (no header involvement; this type doesn't need to be public):
```C++
struct GlobalTelemetryTraceLogger : public TelemetryTraceLogger
{
    GlobalTelemetryTraceLogger() { Initialize(); }
}

JohnMcPMS (approved):

Issue #674, opened 2020-12-29
When do we begin to see the PowerShell cmdlets?


doctordns:
# Description of the new feature/enhancement

We have previously asked for good PowerShell cmdlet support from the WinGet project.
The team have agreed and there is a milestone.  This is good!

Thus far, I have seen no progress, no initial thoughts on how this might work, and no detailed object model or suggested commands.
If the team is to meet that milestone commitment then time is running short (although Covid has probably made any older milestones much more "flexible".

# Proposed technical implementation details (optional)

I suggest some draft thoughts and maybe some proto-commands. Not dissimilar to what happened with Hyper-V cmdlets. An early module was published (that was terrible!) but was quickly transformed into a great tool Iand adopted by the Hyper-V team who did an amazing job). Publish the early thinking, and let the PowerShell community help you to do another great job.

megamorf:
Duplicate of #221.

doctordns:
It is nor really a duplicate. My issue is asking for early work to be published. Issue #221 is asking for the completion of a fully supported PowerShell module.

Of course, both issues are related - but nevertheless where is the progress against #221? The road map says end-April for full PowerShell support. To have a hope of releasing a good version by then, work has to have started (probably a while ago).  This issue is asking the team to expose their thinking one what they plan to deliver (and by when). 

This issue can be closed once and an initial set of cmdlets are ready, or when the 1st draft of their specification is available. 

And for the avoidance of doubt, my goal is to see a great set of cmdlets that enable IT Professionals to manage the underlying winget feature set. To have good cmdlets with all the qualities of PowerShell but thus far lacking in winget.exe is not an easy task.

denelon:
As we get "closer" to the feature set intended for 1.0 we will be ready to look at the proper syntax for cmdlets. I'll likely propose some sort of draft via a gist.

denelon:
I also believe we will be able to iterate fairly quickly once we have clear scope. We will also need to continue revising the syntax as new features are developed.

doctordns:
I realise you may be new to PowerShell and cmdlets. From nearlhy 17 years experience with PowerShell, I can tell you that building a huge .exe with it's attendant infrastructure, and then trying to back port it to be a great PowerShell modules is going to be very hard work. It's not just a few get/set commands. And the more you bake into the exe, with commands switches etc all designed without thought of the IT Pro), the hard it will hard to do a great job.

[...]

PR #1407, opened 2021-08-28
PowerShell cmdlets

denelon:

-----

#674

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1407)


doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,434 @@
...
+-Silent
+
+-Log
+
+**Example**
+
+>Note: The example below could be executed to remove all packages with "xbox" in their name, or in other metadata fields provided in the manifest (assumes there is a match with the installed package in a configured source).
+
+```PowerShell
+Get-WingetPackage xbox | UninstallWingetPackage
```
megamorf, line 270: Typo: `s/UninstallWingetPackage/Uninstall-WingetPackage/g`

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,434 @@
...
+**PowerShell Cmdlet**
+
+```PowerShell
+Install-WinGetPackage
+```
+**Parameters**
+
+-Manifest
+
+-PackageIdentifier
```
megamorf, line 56: Since this parameter will be used quite often I think it would make sense to add a parameter alias:

```c#
[Alias("Id")]
```

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,434 @@
...
+```PowerShell
+Reset-WinGetSource
+```
+
+**Example**
+> Note: `winget source export` was added as a helper for adding approved sources to Group Policy. To get the same values with PowerShell the example blow would perform the same behavior.
+
+
+```PowerShell
+Get-WinGetSource | convertto-json -compress
```
megamorf, line 138: Since this would entirely rely on the output object structure this approach could break easily. It might make more sense to handle this via a `-Raw` parameter, e.g.
```powershell
Get-WinGetSource -Raw
```

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,434 @@
...
+```
+
+---
+
+### Client command `winget hash`
+
+**PowerShell Cmdlet**
+
+```PowerShell
+Get-WinGetHash
```
megamorf, line 280: Kind of redundant. See [Get-FileHash](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/get-filehash?view=powershell-7.1)

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,434 @@
...
+### Client command `winget export`
+
+**PowerShell Cmdlet**
+
+```PowerShell
+Export-WinGetPackageSet 
+```
+**Parameters**
+
+-Output
```
megamorf, line 354: `-Path` and `-LiteralPath` might be better parameter names for `-Output`. See [Export-Csv](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/export-csv?view=powershell-7.1)

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,434 @@
...
+
+-ImportFile
+
+-Install
+
+>Note: This would behave like the current `winget install` command.
+
+-IgnoreUnavailable
+
+-IgnoreVersions
```
megamorf, line 382: `Skip` could be used as prefix instead of `Ignore`.

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,434 @@
...
+>Note: The current behavior for `winget show vscode` would look like the example below.
+
+```PowerShell
+Find-WinGetPackage -Moniker vscode | Get-WinGetPackageVersion -Detail
+```
+
+>Note: The current behavior for `winget show vscode --versions` would look like the example below.
+```PowerShell
+Find-WinGetPackage -Moniker vscode | Get-WinGetPackageVersion -All
+```
```
megamorf, line 109: This is valid PowerShell code but needlessly complicated from a user experience perspective. Many users will expect these actions to be done with a single command. It'll be interesting to see what feedback you'll get here :-)

megamorf (commented):
Added my feedback :+1:

jantari:
Looks like a solid first draft overall.

I have noticed there is no mention of ParameterSets, but some cmdlets parameters should probably be mutually exclusive e.g. these three wherever they are used:

    -PackageIdentifier
    -Name
    -Moniker

I think a user should not be able to specify more than one of these for any given command.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: surely it should be Name?

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,172 @@
...
+
+-Location
+
+### `winget show`
+
+Show-WindowsPackage
+
+### `winget source`
+
+Add-WindowsPackageManagerSource
```
doctordns: This, like other cmdlet names is long. Can you use a shorter Noun Prefix?

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,172 @@
...
+Remove-WindowsPackageManagerSource
+
+### `winget search`
+
+Find-WindowsPackage
+
+### `winget list`
+
+Show-InstalledWindowsPackages
```
doctordns: Should be Get- not Show-?

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,172 @@
...
+Get-WindowsPackageSha256
+Get-WindowsPackageSignatureSha256
+
+### `winget validate`
+
+Test-WindowsPackageManifest
+
+### `winget settings`
+
+Edit-WindowsPackageManagerSettings
```
doctordns: Should use the SET verb

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,172 @@
...
+
+### `winget features`
+
+Show-WindowsPackageManagerFeatures
```
doctordns: should be
Get-WindowsPackageManagerExperimentalFeatures

doc/specs/#674 - PowerShell cmdlets.md
jantari: Package ID and Name are two different things. The Name parameter is two lines below

doc/specs/#674 - PowerShell cmdlets.md
doctordns: The cmdlets shoudl emit either .NET Objects (for consumption by other PowerShell cmdlets along with JSON where necessary.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: Find-Package should, uhh, find packages and return all the information. C/F Find-Module. Please do not introduce needless complexity where it can be avoided.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: Can you have two packages with the same name? Likewise Package-ID?  What is the difference?

jantari: Take a look at the schema for an explanation of the schema: https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.0.0/manifest.singleton.1.0.0.json

Both are arbitrary strings, only the ID has to be unique. It's the reason you basically can't install anything with winget unless you use `--id` because it always finds multiple results

jedieaston:
Just checking, since it's not in this spec file, these cmdlets are targeting Windows PowerShell (5.1), not PowerShell Core, right?

doc/specs/#674 - PowerShell cmdlets.md
jantari: I disagree. AFAIK winget repositories use a cached index to speed up searches. It makes a ton of sense to return only the minimal, cached set of data first and then get all the information (which is an expensive operation) only for the package you're really interested in. Other existing cmdlets such as Get-ADUser and Get-WindowsDriver take a similar approach for performance and server-load reasons.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: Is Get-FileHash doing the same thing as the proposed get-wingethash?

doc/specs/#674 - PowerShell cmdlets.md
doctordns: Also - there should be a noun prefix here.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: and - there should probably be a 'Set-WIndowsPackageManagerExperimentalFeature

note the singular noun.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: If the approach is to return a minimal amount of properties, then two things:
1. The Find-WinGetPackage should return enough properties, by default, to be useful. As you mantion, like Get-ADuser etc.
2. If there are more properties not returned by default - then use a -Property parameter to specify which optional properties to return. 
In this case, I'd argue that the version of a package is important and probably should be returned by default. 

There are two issues therefore: what properties get returned by default and how to get more. I do not support adding cmdlets in the proposed manner.

> Just checking, since it's not in this spec file, these cmdlets will be targeting Windows PowerShell (5.1), not PowerShell Core, right? We can't guarantee that PowerShell Core is installed since it's not shipped with Windows (and it isn't usually installed on systems that aren't used for development unless you work at a really hip company).

I sure hope not. It would be crazy, this long after PowerShell 7 has shipped to be so backwards as to ONLY target WIndows PowerShell.

doctordns:
> Just checking, since it's not in this spec file, these cmdlets will be targeting Windows PowerShell (5.1), not PowerShell Core, right? We can't guarantee that PowerShell Core is installed since it's not shipped with Windows (and it isn't usually installed on systems that aren't used for development unless you work at a really hip company).

These cmdlets should be validated and fully tested for both Windows PowerShell and PowerShell 7.x and should NOT rely on the PowerShell 7 compatibility mechanism. MUCH more importantly, the team should not use techniques that _only_ work with       Windows PowerShell. The WSUS team's cmdlets and a few others, use classes that exist in .NET Framework, but don't in (open source) .NET.   That makes it much more complex for IT Pros who do use PowerShell 7 when it doesn't need to be that way. Please ensure the cmdlets can run on both versions of PowerShell. 

That is not to say that many (most?) users at first would be using Windows PowerShell. But PowerShell 7 has been around for a couple of years (7.0, 7.1, and very soon 7.2). But if my book sales are anything to go by, there are many IT pros investigating and moving across to PowerShell 7 and it would be wrong to just not to take advantage.  And in enterprise arenas, there are many PowerShell 7 features that would be useful in a Package Management Scenairo.

doc/specs/#674 - PowerShell cmdlets.md
jedieaston: `winget hash` just gives you the hash of a file, so I think so, yeah.

jedieaston:
> MUCH more importantly, the team should not use techniques that only work with Windows PowerShell.

I agree with this. 

> there are many IT pros investigating and moving across to PowerShell 7 and it would be wrong to just not to take advantage.

Kind of off-topic, but why the heck isn't it shipped with Windows then? I don't mind the changes in Core, but I don't use it because I can't guarantee it will be there when I run something. I can guarantee Windows PowerShell is there, and if I need to install a language, why not one I like? (I kid. but something like PowerShell needs to be in the box to be of the best use).

>  And in enterprise arenas, there are many PowerShell 7 features that would be useful in a Package Management Scenairo.

Which ones are you thinking of? Just curious.

doctordns:
> Kind of off-topic, but why the heck isn't it shipped with Windows then? I don't mind the changes in Core, but I don't use it because I can't guarantee it will be there when I run something. I can guarantee Windows PowerShell is there, and if I need to install a language, why not one I like? (I kid. but something like PowerShell needs to be in the box to be of the best use).
> 

I can only agree, but if it were that simple, they would have done it a long time ago. I suspect it's mainly due to licensing. .NET and PowerShell are both open source and not a proprietary Microsoft Product. I can only imagine the number of lawyers it would take to get this done. I suspect it will happen, just not soon. There is also t;he issue of support - if it;s in the box, MSFT needs to support it. So NOT simple - In the meantime, there are numerous ways to install it (including WInget). And despite this, I hope that the team doesn't do anything that would stop winget's ability to work with PowerShell 7.

> > And in enterprise arenas, there are many PowerShell 7 features that would be useful in a Package Management Scenairo.
> 
> Which ones are you thinking of? Just curious.

Using ``Foreach-Object -Parallel`` to install or even update a set of packages. You can therefore install individual packages potentially in parallel.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: BUt they are different objects returned.
So Get-FileHash returns a .NET object  like this:
```powershellconsole
PS C:\Foo> get-filehash .\2019-01-03_19-15-45.png | gm

   TypeName: Microsoft.PowerShell.Commands.FileHashInfo

Name        MemberType Definition
----        ---------- ----------
Equals      Method     bool Equals(System.Object obj)
GetHashCode Method     int GetHashCode()
GetType     Method     type GetType()
ToString    Method     string ToString()
Algorithm   Property   string Algorithm {get;set;}
Hash        Property   string Hash {get;set;}
Path        Property   string Path {get;set;}
```

whereas Winget Hash returns a string, like this:
PS C:\Foo> winget hash -f .\aaa-airport.txt
Sha256: 1fcfecf92e8917f6bbfaa2668ec17357cd32c2c419db229cc0435e55b1e25485
PS C:\Foo> winget hash -f .\aaa-airport.txt | gm

   TypeName: System.String

Name                 MemberType            Definition
----                 ----------            ----------
Clone                Method                System.Object Clone(), System.Object ICloneable.Clone()
CompareTo            Method                int CompareTo(System.Object value), int CompareTo(string strB), int ICompar…
....
```

So they are not the same objects and contain different properties.

File: doc/specs/#674 - PowerShell cmdlets.md (hunk 5 above)
Comment (denelon) [reply doc/specs/#674 - PowerShell cmdlets.md:354]: @megamorf what is the difference between the two?

Comment (jantari) [reply doc/specs/#674 - PowerShell cmdlets.md:363]: Per convention, `-Path` accepts wildcards and globs. LiteralPath does not. There's probably official docs for this as it's very standard

**Comment by denelon** [comment]
Parallel install is a challenge. MSIX supports up to five concurrent installs (I believe). MSI is a singleton scenario due to msiexec. We also don't know what would happen with parallel installs in a dependency required scenario.

File: doc/specs/#674 - PowerShell cmdlets.md [hunk 12]
```diff
@@ -0,0 +1,443 @@
...
+---
+
+### Client command `winget show`
+
+**PowerShell Cmdlet**
+```PowerShell
+Get-WinGetPackageVersion
+```
+
+-Detail
```
JamesWTruher, line 93: it would be good to elaborate what these different parameters do. since the name of the cmdlet includes *version* i'm a little puzzled why you need a *version* parameter

doc/specs/#674 - PowerShell cmdlets.md
megamorf: Correct. The official documentation says the following:
> Unlike **Path**, the value of the **LiteralPath** parameter is used exactly as it is typed. No characters are interpreted as wildcards. If the path includes escape characters, use single quotation marks. Single quotation marks tell PowerShell not to interpret any characters as escape sequences.

Why is that important? Let's consider that you have a folder with square brackets or an asterisk in its name. `-Path` will treat that as an expression rather than a true representantion of the path. Users have learned to use `-LiteralPath` when they literally want PowerShell to treat the user specified path as it is.

doc/specs/#674 - PowerShell cmdlets.md
JustSomeGuyNZ: > I sure hope not. It would be crazy, this long after PowerShell 7 has shipped to be so backwards as to ONLY target WIndows PowerShell.

I'd disagree. For all that Microsoft want PowerShell 7 to be the future, I don't know a single enterprise that's using it, precisely because it's not a backwards compatible, drop in replacement for PowerShell 5. In every current, supported version of Windows, PowerShell 7 is a side by side addon and the chance of MS actually ever releasing an in place upgrade from PS5 to PS7+ is slim to none, so PowerShell 5 is going to continue to be what is used by enterprise admins for years to come. 
So yeah, IMHO this module absolutely needs to be PS 5.1 compatible.

doctordns: I would agree that the module should work in Windows PowerShell - I am sorry if you thought I meant otherwise. Probably targeting Windows PowerShell 5.1.

But it must also work, and be validated, using PowerShell 7. I expect the Winget team to do NOTHING to stop it from being fully natively in PowerShell 7 as well. We don't need more teams at Microsoft ignoring PowerShell 7 if it can be avoided.

As for your comments - there is a great deal of backwards compatibility built-in. The Compatibility mechanism works well for the Modules shipped, in particular, by the Windows Server team (for the most part). Some teams have taken the opportunity to update their modules to be .NET Core compatible (c/f ActiveDirectory module). I am just wanting to make sure that the winget team does not deliver a PowerShell module that _only_ runs in Windows PowerShell (c/f WSUS, BestPractices, and the PSScheduledJob modules) and has to be blocked in PowerShell 7.  If the Winget team use APIs that are supported both in .NET and the .NET Framework then I can have it in PowerShell 7, and you can use it with  Windows PowerShell without issue. Seems to me that in 2021, this should be a design goal for this and all new modules.

JustSomeGuyNZ: Fair play, I had misread your post as wanting it to not be PS5 compatible. I've got absolutely no objection to things being both PS5 and PS7 compatible, but this tool could be of great use to those who work in enterprise environments where hell is likely to freeze over before PS 7 is widely deployed.

denelon: I've been engaging the PowerShell team in the discussions with our design, and the crescendo team to aid in rapid prototyping. We will follow their guidance with respect to which versions of PowerShell behave differently so we can maintain the greatest compatibility level. I am certainly not a PowerShell expert, so I'm learning from the experts on that front. I wanted to get this draft out quickly so we could get community feedback. I'm building on the list of concerns brought up here so they can be clearly addressed in the document.

denelon:
> I think a user should not be able to specify more than one of these for any given command.

This is an interesting position. If I think about this in text, it seemed logical for one to say "I want software by Adobe with Acrobat in the name". This seems counter to that proposal. It would also be a breaking change to implement this mutual exclusion in the command line interface. I'd like to avoid breaking changes until we're ready to make a 2.0 release. I also want to clarify it's not a function of a date on a calendar, but more about the right time to take in a series of breaking changes. We should ensure that the 1.x versions are stable enough to go into a bug fix support mode when we make a semantically different release.

doc/specs/#674 - PowerShell cmdlets.md
denelon: OK, so rather than "-Output" I believe what I'm reading is we should support both "-Path" and "-LiteralPath". Is it safe to assume PowerShell handles the wildcards when "-Path" is used?

doc/specs/#674 - PowerShell cmdlets.md
Karl-WE: @justsomeguynz I agree with your point that there are companies that will never allow PS 7 to be deployed. I do not get the reason to be honest. I consult customers to deploy it widely just for the same of improved commandlets like test-connection etc. 
It's an MSI package, easy to deploy with a small mst. What is their issues with PS 7? Do you have more information about it? I just know it is a controversy because of patchmanagement.org and also some ConfigMgr MEMCM pals.

jantari:
> This is an interesting position. If I think about this in text, it seemed logical for one to say "I want software by Adobe with Acrobat in the name". This seems counter to that proposal. It would also be a breaking change to implement this mutual exclusion in the command line interface. I'd like to avoid breaking changes until we're ready to make a 2.0 release. I also want to clarify it's not a function of a date on a calendar, but more about the right time to take in a series of breaking changes. We should ensure that the 1.x versions are stable enough to go into a bug fix support mode when we make a semantically different release.

I, uh, wasn't even aware you could currently do a query like this: `winget search --moniker vscode --name insider --id Microsoft`. I do want to say that your usecase of "I want software by Adobe with Acrobat in the name" hinges entirely on people including the software publisher in the id-string of their packages, which is not mandatory and is also not really the point of an ID (companies/publishers change names, an ID is supposed to stay unique and ideally constant so t hat automation doesn't break). I am not sure why the package ID wasn't just forced to be a GUID in the schema, but it's too late for that now I guess... Once we have PowerShell cmdlets we can filter packages properly by their publisher without the need to rely on magic conventions in ID-strings anyway, so we really don't need to (and should not encourage) the use of multiple of these parameters together (`--id` and `--name` for example). The proper way to do it would be `Get-WingetPackage --Name "*code*" | Where-Object { $_.Publisher -eq 'Microsoft' }` anyway, and we maybe shouldn't permit nonsensical parameter-combinations in v1 if we know we're going to remove it in v2 again. There is no expectation of the PowerShell cmdlets working exactly the same way as the winget.exe command, in fact if that was the case we wouldn't need them - the point is that they're a *better* (and windows-native) interface to wingets features than the existing winget.exe. my opinion though.

doctordns:
A better approach might be to do this:
```
Get-WingetPackage -Filter {Name -like "*code*"  and Publisher -eq 'Microsoft' } 
```

doc/specs/#674 - PowerShell cmdlets.md
JustSomeGuyNZ: It's not a drop in replacement, and it breaks backwards compatibility. In my experience enterprise IT is pretty resistant to  change that breaks backwards compatibility and unforced changes lol.

doctordns:
> Parallel install is a challenge. MSIX supports up to five concurrent installs (I believe). MSI is a singleton scenario due to msiexec. We also don't know what would happen with parallel installs in a dependency required scenario.

True - but you could (again in an Enterprise scenairo) install packages on multiple systems in parallel - and given that the actual installation time is the bottle neck, updating multiple machines at once could deliver significantly faster installations.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: NOte - if the intention is to handle cmdlets using Crescendo - then this can be met, _provided_ the team do a good job at writing the necessary output handelers. And for the record, with Cresendo, this is NOT an easy task unless you are outstandingly good with regular expressions.

With that said, the cmdlets must emit objects - and some of those objects can contain json where json is relevant.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: So far as I can recall, -Path takes a wildcarded string and the cmdlet does the variable expansion.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: > I've been engaging the PowerShell team in the discussions with our design, and the crescendo team to aid in rapid prototyping. We will follow their guidance with respect to which versions of PowerShell behave differently so we can maintain the greatest compatibility level. I am certainly not a PowerShell expert, so I'm learning from the experts on that front. I wanted to get this draft out quickly so we could get community feedback. I'm building on the list of concerns brought up here so they can be clearly addressed in the document.

This is excellent news and much to be welcomed. Knowing what I do, Crescendo should provide a good way forward and helps to move this forward and makes the best out of the decision to do this all in C++. But we are where we are, and Crescendo does look promising as a way forward. It also gives the winget team to re-write certain cmdlets in C# where that makes business sense. 

As I stated a while ago, creating the object model is very important. With Crescendo, as I understand it, you use regular expressions to create objects that the Crescendo-generated cmdlet emits. As I have said elsewhere, this is going to be a very difficult task to do well and completely - string to object translation is not an exact science. Thus my first concern is the objects emitted. It should be a goal for winget-ps to emit objects that are as rich and functional as those emitted by other modules and usable in rich cross-silo scenarios. 

The second concern is over verb usage. As mentioned a long time ago, the subcommand of winget do not line up completely with PowerShell standard verb naming.  For example, the show command in winget should probably be get. And hash should be something else. I would like to see the command to verb translation table - and please, please, please only uses approved verbs in all cmdlets. 

I look forward to seeing how this progresses.

denelon: I've been going through the process of learning the PowerShell verbs and continuing to expand the draft to others so we can get a sense of the best verbs and behaviors. First impressions are very valuable in this area. There is definitely some impedance to overcome between the .exe cli syntax and PowerShell. We're heads down on the v1.1 release, and then I'll have more time to dedicate to fleshing this out. We have been discussing the rich object model needed for a good PowerShell experience.

doctordns: The verb choice is a vital component of the sacred vow. IT Pros would expect to use certain verbs and certain parameter names names. To create a truly  great Powershell module means using approved verbs. :-)

jantari:
> A better approach might be to do this:
> 
> ```
> Get-WingetPackage -Filter {Name -Like "*code*"  -And Publisher -Eq 'Microsoft' } 
> ```

That'd be quite a lot of extra work, to implement a querying and filtering syntax for winget. The `exe` doesn't have something like this either currently. The specific syntax you proposed also doesn't make a ton of sense if we dig into it, the filter should probably be a string instead of a scriptblock and as a query language something like JMESPath would make a lot of sense because the winget packages are in JSON-compatible YAML format already. Azure CLI also uses JMESPath expressions, and there are ready-made libraries for it. Your syntax looks similar to the one the ActiveDirectory module implemented, but that's something entirely different and that filter-implementation was likely entirely custom. WMI and CIM also have their own query language ( WQL ) that looks like this `-Filter 'Name LIKE "%search%"'` - I say enough with the proprietary syntax, just give us JMESPath support in v2 -
this `-Filter` parameter and it's `.exe` CLI-equivalent could be introduced at any point later and doesn't have to be in v1.

> > Parallel install is a challenge. MSIX supports up to five concurrent installs (I believe). MSI is a singleton scenario due to msiexec. We also don't know what would happen with parallel installs in a dependency required scenario.
>
>True - but you could (again in an Enterprise scenairo) install packages on multiple systems in parallel - and given that the actual installation time is the bottle neck, updating multiple machines at once could deliver significantly faster installations.

Respectfully, I think you're getting lost and arguing about details that don't actually matter. Even the old versions of PowerShell can run arbitrary commands against a list of computers in parallel just using the old `Invoke-Command` cmdlet. There is no need for `foreach -parallel` or a new PowerShell version to be able to do this.

---

Let's focus on getting a good and extensible v1 of the winget PowerShell cmdlets finalized before we dive into fantasies about what we could possibly tack on in the future. The cmdlets should align somewhat with the capabilities of the winget.exe program and there should probably not be too much custom functionality directly embedded into them that the exe doesn't have. - just for consitency. Anything else we can use the existing capabilities of PowerShell for to accomplish, e.g. parallel runs on many computers, complex filter logic or runspaces or whatever.

doc/specs/#674 - PowerShell cmdlets.md
jantari: This was marked as resolved, but resolved with what decision?

`winget.exe hash -f .\file.txt` may return a string but how is that relevant when the same string can be accessed with:

```powershell
(Get-FileHash -LiteralPath .\file.txt -Algorithm SHA256).Hash
```

? The `.Hash` property of the `Microsoft.PowerShell.Commands.FileHashInfo` object that `Get-FileHash` returns contains the hash-string and if someone really wants to recreate the exact format that winget outputs they could do so:

```powershell
$hash = Get-FileHash -LiteralPath .\get-pip.py -Algorithm SHA256
"{0}: {1}" -f $hash.Algorithm, $hash.Hash.ToLower()
```

I agree with @megamorf here, it's redundant and the generic `Get-FileHash` cmdlet is preferred.

doctordns:
Yes, the filter parameter is based on existing methods built into PowerShell (and referencing late vs early filtering). The Get-ChildItem cmdlet, for example, uses a -Filter parameter, although it uses a string vs a script block, A string, as you mention, is a better choice especially as a string is serializable which might be relevant for cross-host action. If we are building PowerShell cmdlets, I would like to see the filter cmdlet follow PowerShell syntax - it's less for the IT Pro to learn and is an approach in keeping with the sacred vow. While Other query mechanisms  might be interesting, they impose a learning curve which can be avoided)., 

I also take the point about v1 vs v2. I agree totally with getting a decent set of cmdlets out to begin with and improve them in V2.

How about this - a) make sure that V1` produce all the properties necessary to pass to Where-Object et al (ie later filtering). And define a -Filter parameter that does filtering in V2. The PowerShell team might be able to point to some of the similar code as a head start on the coding.

Karl-WE:
In the context of PowerShell commandlets for WinGet will there be a method to following scenario?
- having an input file as text with one line per application
- having an input file as text formatted as CSV 
- having an input file as formatted as XML

Winget can read the import file 
I imagine this could be a pipe command like Get-Content or similar that pipes the input file to winget.
Alternatively, winget has own commandlets to specify such an import file


**txt file:**
get-content -path c:\inputfile.txt | Install-WinGetPackage -ErrorAction SilentlyContinue

**example of inputfile.txt:**

inputfile.txt
`OpenWhisperSystems.Signal
CPUID.CPU-Z
CrystalDewWorld.CrystalDiskInfo
KCSoftwares.SUMo
Logitech.Options
Nlitesoft.NTLite
WiresharkFoundation.Wireshark
Intel.IntelDriverAndSupportAssistant
Microsoft.VC++2015-2019Redist-x64
GitHub.cli
Zoom.Zoom
Microsoft.PowerToys
Microsoft.VisualStudioCode
Microsoft.dotnetRuntime.5-x64`

**CSV file**
import-csv -path c:\inputfile.csv | Install-WinGetPackage -ErrorAction SilentlyContinue

**XML file:**
import-clixml -path c:\inputfile.xml -encoding Unicode | Install-WinGetPackage -ErrorAction SilentlyContinue


Result: this will install (or alternatively) upgrade all packages that contained in the file.
 It will continue on error.
 
**Usecase:** 
automated installation of software based on an input file / central distribution / standardized installation and application patch management

denelon:
> In the context of PowerShell commandlets for WinGet will there be a method to following scenario?
> 
> * having an input file as text with one line per application
> * having an input file as text formatted as CSV
> * having an input file as formatted as XML
> 
> Winget can read the import file
> I imagine this could be a pipe command like Get-Content or similar that pipes the input file to winget.
> Alternatively, winget has own commandlets to specify such an import file
> 
> **txt file:**
> get-content -path c:\inputfile.txt -encoding Unicode | Install-WinGetPackage -ErrorAction SilentlyContinue
> 
> **example of inputfile.txt:**
> 
> inputfile.txt
> `OpenWhisperSystems.Signal`
> `CPUID.CPU-Z`
> `CrystalDewWorld.CrystalDiskInfo`
> `KCSoftwares.SUMo`
> `Logitech.Options`
> `Nlitesoft.NTLite`
> `WiresharkFoundation.Wireshark`
> `Intel.IntelDriverAndSupportAssistant`
> `Microsoft.VC++2015-2019Redist-x64`
> `GitHub.cli`
> `Zoom.Zoom`
> `Microsoft.PowerToys`
> `Microsoft.VisualStudioCode`
> `Microsoft.dotnetRuntime.5-x64`
> 
> **CSV file**
> import-csv -path c:\inputfile.csv -encoding Unicode | Install-WinGetPackage -ErrorAction SilentlyContinue
> 
> **XML file:**
> import-clixml -path c:\inputfile.xml | Install-WinGetPackage -ErrorAction SilentlyContinue
> 
> Result: this will install (or alternatively) upgrade all packages that contained in the file.
> It will continue on error.
> 
> **Usecase:**
> automated installation of software based on an input file / central distribution / standardized installation and application patch management

I'm not a PowerShell expert, but I would expect you could iterate over the list and run the install command for each of them. The "import" format for a collection of packages is JSON with a schema, so that is the "current" preferred method, but I believe PowerShell will open up a myriad of new possibilities.

doc/specs/#674 - PowerShell cmdlets.md
denelon: I'm fine with removing the Windows Package Manager specific hash command if the community agrees with using Get-FileHash. It reduces the surface area, and uses the standard PowerShell tool based on what I'm reading here. It was originally added to the CLI to help author manifests. The same holds true for calculating the MSIX Signature SHA256. I'm not sure if we would be able to remove that, or if there is another PowerShell incantation to determine the signature hash.

doctordns:
> > In the context of PowerShell commandlets for WinGet will there be a method to following scenario?
> > 
> > * having an input file as text with one line per application
> > * having an input file as text formatted as CSV
> > * having an input file as formatted as XML

I would expect Install-WingetPackage to have a -Name parameter which is of the type [string[]].  

Rather than ask Winget to worry about XML, CSV,(and probably JSON, let _me_ workout where the strings come from and just pass them to the cmdlet as strings.

doc/specs/#674 - PowerShell cmdlets.md
doctordns: So long as Winget can accept the hash created with Get-FIleHash - seems a no brainer to remove it.  The approach should always be to not reinvent wheels wherever possible.  

It also removes the need to worry about how to convert it to a cmdlet, name wise! :-)

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,443 @@
...
+-All
+
+---
+
+### Client command `winget uninstall`
+
+**PowerShell Cmdlet**
+
+```PowerShell
+Uninstall-WinGetPackage
```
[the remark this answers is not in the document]
doctordns: In a perfect world, ``Install-WinGetPackage`` AND ``Uninstall-WInGetPackage` would support all the common parameters, including both ``-WhatIf`` and ``-Confirm``.

What I don't know is how that would work with Crescendo. The Crescendo wrapper _may_ be able to say "We are going to call winget now" but the internal winget processing probably would not be easy to expose to the ShouldProcess mechanism.

doc/specs/#674 - PowerShell cmdlets.md
```diff
@@ -0,0 +1,443 @@
...
+-Moniker
+
+-Version
+
+-Source
+
+-Exact
+
+-Interactive
+
```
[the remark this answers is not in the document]
doctordns: Are multiple parameter sets possible with Crescendo?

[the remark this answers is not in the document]
denelon: Yes, that's my current understanding.

doc/specs/#674 - PowerShell cmdlets.md
megamorf: @denelon As far as I know there is no built-in command to achieve the same as 
```
PS> winget hash -f example.msix -m
Sha256: 6bfa8f184e0afeb38dc7112e920d06df4704b4b1068968167c1972a97c1cc5ee
SignatureSha256: 6149697e47bdb4a29cbc07a62a79ac57c0d7007062a3514e63322d00be179d7c
```

One approach in PowerShell would be to extract the package and then extract the information:

```
PS> Get-FileHash -Path example.msix

Algorithm       Hash                                                                   Path
---------       ----                                                                   ----
SHA256          6BFA8F184E0AFEB38DC7112E920D06DF4704B4B1068968167C1972A97C1CC5EE       C:\Users\megamorf\Downloads\example.msix

# Expand archive expects .zip extension
PS> Rename-Item -Path example.msix -NewName example.zip
PS> Expand-Archive -Path example.zip -DestinationPath $env:TEMP\MsixExtracted
PS> Get-FileHash -Path $env:TEMP\MsixExtracted\AppxSignature.p7x -Algorithm SHA256

Algorithm       Hash                                                                   Path
---------       ----                                                                   ----
SHA256          6149697E47BDB4A29CBC07A62A79AC57C0D7007062A3514E63322D00BE179D7C       C:\Users\megamorf\AppData\Local\Temp\MsixExtracted\AppxSignature.p7x
```

This would be kind of like this here:
```powershell
function Get-AppxSignatureHash
{
    [CmdletBinding()]
    Param
    (
        [Parameter(Mandatory=$true)]
        $Path
    )

    Add-Type -AssemblyName System.IO.Compression.FileSystem
    $Guid = [System.Guid]::NewGuid()
    $OutputPath = Join-Path $env:TEMP $Guid

    try {
        $zip = [System.IO.Compression.ZipFile]::OpenRead($Path)
        $SigFile = $zip.Entries | Where-Object { $_.FullName -like "AppxSignature.p7x" }
        [System.IO.Compression.ZipFileExtensions]::ExtractToFile($SigFile, $OutputPath, $true)
        $zip.Dispose()

        Get-FileHash -Path $OutputPath -Algorithm SHA256
    }
    catch {
        [PSCustomObject]@{Hash=""}
    }
    finally {
        Remove-Item -Path $TempDir -ErrorAction SilentlyContinue
    }
}

$Path = "C:\Users\nem\megamorf\example.msix"

[PSCustomObject]@{
    Sha256 = (Get-FileHash -Path $Path -Algorithm SHA256).Hash.ToLower()
    SignatureSha256 = (Get-AppxSignatureHash -Path $Path -Verbose).Hash.ToLower()
} | Format-List


Sha256          : 6bfa8f184e0afeb38dc7112e920d06df4704b4b1068968167c1972a97c1cc5ee
SignatureSha256 : 6149697e47bdb4a29cbc07a62a79ac57c0d7007062a3514e63322d00be179d7c
```

doctordns: What role does ``winget hash`` play in the sceme of things? Does it _just_ get the hash of a wingetted program?  If I understand it, winget itself could just drop the hash command and tell folks to just use Get-FileHash?

denelon: It was a way to get the file hash and the signature hash (for an MSIX package) to make manifest creation a bit easier. We now have wingetcreate for that so it's a bit redundant, and will likely be removed when wingetcreate is 1.0 stable.

doc/specs/#674 - PowerShell cmdlets.md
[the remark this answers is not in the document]
leoniDEV: Maybe for `-Silent` and `-Interactive` would be better to have a single switch parameter

denelon:
I'm going to try and update this PR and convert to a draft.

denelon:
Here is the Fork/Branch with the "Fix, Hack, Learn" project for a PowerShell module. This is where I'm doing some learnings.

https://github.com/denelon/winget-cli/tree/PowerShellFHL/PowerShell/WinGet

jedieaston:
I don’t think there needs to be two separate cmdlets. Most of the Get-* ones I’ve used have search options available and options to get one specific thing. For example, [Get-AzureADUser](https://docs.microsoft.com/en-us/powershell/module/azuread/get-azureaduser?view=azureadps-2.0) will give you a specific user if you give it an ID or all of the users if you don’t (well, you have to pass `-All`). 

`-Installed` to only look locally is how a lot of cmdlets do type filters (like `Get-ChildItem -Directory`). That makes sense to me.

denelon:
It looks like we're going to go with Get-WinGetPackage for "list" a.k.a. installed packages and Find-WinGetPackages for "search" a.k.a. packages available in sources.

We're still debating some options for "show". We're not really displaying an entire manifest, but we're returning an object that is a bit of a subset of a manifest. We're still looking at parameters and verbs that might fit the bill better in this scenario.

JohnMcPMS:
> We're going to have "Get-WinGetManifest" for operations against a REST source. `winget show` is really pulling in a subset of a manifest isolated to the locale, machine architecture, and scope. We were thinking the object was more about the installer than a manifest.

Remember to not be limited by how the current `winget` CLI operates.  PowerShell is more like the COM API than the command line, and it will be far easier to make them work together the closer they are.  I realize that makes an intermediate Crescendo based exploration more difficult to implement fully, but maybe it can be half measures until fully hooked up with COM.  For instance, a `Manifest` object exposing a subset of the fields, but also *all* of the installers.  For now you can only have one with Crescendo, but still make it an array that we can put all of them into it in the future.

Also, the source type should not ever be a consideration for API design at this level.  They are all black box repositories of packages.  Yes, there might be some special flags that only affect a subset of types, but that is acceptable so long as the underlying code is handling it, not this layer.

doctordns:
It seems to me that instead of emitting mangled strings, there should be an option at least for Crescendo, that send the output ion JSON, and enables Crescendo to do the conversion into objects.

When will the team publish the object model?

denelon:
We still have to take what we get from the client (strings in most cases) and build objects. Crescendo ended up not being quite as useful as we had hoped due to the difference between the command line verbs/switches and PowerShell. One exception was the `winget source export` command returned JSON.

Since we ended up with four modules, we're going to create a new repository to get the PowerShell build pipelines set up, then we will move the "child" modules back to their respective repositories.

You can see the [work in progress](https://github.com/denelon/winget-cli/tree/PowerShellFHL/PowerShell) on my fork.

leoniDEV:
@denelon I see your repo and I find 2 big issues:

1) [don't use `Write-Host`](https://www.jsnover.com/blog/2013/12/07/write-host-considered-harmful/)!

2) Collect all the resultant objects in an array and then return the array, defeat the purpose to support the pipeline

denelon:
The Microsoft.WinGet.Source PowerShell module was released at https://github.com/microsoft/winget-cli-restsource/releases as a preview version. I've created a [discussion](https://github.com/microsoft/winget-cli-restsource/discussions/92) to gather feedback.

megamorf:
> The Microsoft.WinGet.Source PowerShell module was released at https://github.com/microsoft/winget-cli-restsource/releases as a preview version. I've created a [discussion](https://github.com/microsoft/winget-cli-restsource/discussions/92) to gather feedback.

That's great to hear. I'll check the code base and cmdlet behaviour this weekend and add my feedback 👍

denelon:
We will be releasing the other modules containing cmdlets via another open source repository here at GitHub. They will be migrated to their respective repositories as they become a bit more complete. We will keep the other repository for the parent module designed to import all three of the modules we've defined:

Microsoft.WinGet
* Microsoft.WinGet.Client
* Microsoft.WinGet.Create
* Microsoft.WinGet.Source

We're currently dotting the "I"s and crossing the "T"s to release the project.

floh96:
@denelon why did you not merge the spec pr?

denelon:
@flo96 I broke my branches yesterday. I just submitted https://github.com/microsoft/winget-cli/pull/1760 yesterday after restoring one of them. I'll try to get this one back as well.

ryfu-msft (approved):
:shipit:

PR #1420, opened 2021-08-31
Expose pass through header through com interface.

sreadingMSFT:

-----
This allows callers to set the pass through header on any PackageCatalogReference which means that the header will be sent on any call that the catalog\source makes during and after Connect. For composite catalogs I've chosen to have the value override any values set on individual references in the composite, if it is set. I did not think it made sense to try to update each internal PackageCatalogReference with the value that was being set since the caller could just set the internal one to a different value later which would make returning the value from the composite catalog very confusing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1420)


src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -64,8 +63,14 @@ namespace winrt::Microsoft::Management::Deployment::implementation
                 {
                     auto catalog = m_compositePackageCatalogOptions.Catalogs().GetAt(i);
                     winrt::Microsoft::Management::Deployment::implementation::PackageCatalogInfo* catalogInfoImpl = get_self<winrt::Microsoft::Management::Deployment::implementation::PackageCatalogInfo>(catalog.Info());
-                    ::AppInstaller::Repository::SourceDetails sourceDetails = catalogInfoImpl->GetSourceDetails();
-                    std::shared_ptr<::AppInstaller::Repository::ISource> remoteSource = ::AppInstaller::Repository::OpenSourceFromDetails(sourceDetails, progress).Source;
+                    
+                    ::AppInstaller::Repository::SourceDetails details = catalogInfoImpl->GetSourceDetails();
+                    // If the AdditionalPackageCatalogArguments value has been explicitly set on the composite, then override the value on each individual catalog.
```
JohnMcPMS, line 68: I think we should fail to set the value on the Reference rather than support this.  If the caller truly wants to pass it to all of the other sources they can do so programmatically easily enough.

JohnMcPMS (changes requested):

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -115,4 +113,33 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        }
+
+        return {};
+    }
+    void PackageCatalogReference::AdditionalPackageCatalogArguments(hstring const& value)
+    {
+        if (IsComposite())
+        {
+            // Can't set AdditionalPackageCatalogArguments on a composite. Callers should set it on each non-composite PackageCatalogReference in the composite.
+            throw winrt::hresult_error(ERROR_INVALID_STATE);
```
JohnMcPMS, line 137: Does this automatically convert a Win32 error to an HRESULT?  It doesn't look like it from the docs.  Also, it looks like you should use either `winrt::hresult_illegal_state_change` or `winrt::throw_hresult(E_ILLEGAL_STATE_CHANGE)` given that they have those defined over this Win32 error we thought of off the cuff.

JohnMcPMS (approved):

PR #1421, opened 2021-08-31
Add expected installer return codes

lechacon:
Adding expected return codes for installer. Expected return codes are translated to specific HResults when found and show a custom error message.
* Updated v1.1 manifest schema with new field for expected return codes. Updated file would look like this:
```
Installers: 
    - InstallerUrl: https://example.com
      InstallerType: exe
      ExpectedReturnCodes:
          PackageInUse: 1
          InstallInProgress: 2
```
* Updated manifest parsing and validation for the new fields.
* Refactored common code for checking installer exit code from MSI and ShellExecute installs into a workflow task that now also checks the expected return codes.

Not in this PR: Changes for REST sources

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1421)


schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
```diff
@@ -132,18 +132,50 @@
         }
       }
     },
+    "ReturnCode": {
```
yao-msft, line 135: >ReturnCode [](http://example.com/codeflow?start=5&length=10)

nit: InstallerReturnCode

schemas/JSON/manifests/v1.1.0/manifest.singleton.1.1.0.json
```diff
@@ -165,18 +165,52 @@
...
+        "$ref": "#/definitions/InstallerReturnCode"
       },
       "maxItems": 16,
       "uniqueItems": true,
       "description": "List of additional non-zero installer success exit codes other than known default values by winget"
     },
+    "ExpectedReturnCodes": {
+      "type": "object",
+      "properties": {
+        "PackageInUse": {
```
JohnMcPMS, line 187: RebootRequired?

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -80,6 +80,15 @@
 #define APPINSTALLER_CLI_ERROR_LICENSE_NOT_ACCEPTED             ((HRESULT)0x8a150041)
 #define APPINSTALLER_CLI_ERROR_PROMPT_INPUT_ERROR               ((HRESULT)0x8a150042)
 #define APPINSTALLER_CLI_ERROR_INVALID_MSIEXEC_ARGUMENT         ((HRESULT)0x8a150043)
+#define APPINSTALLER_CLI_ERROR_INSTALL_PACKAGE_IN_USE           ((HRESULT)0x8a150044)
```
JohnMcPMS, line 83: I wonder if we shouldn't jump to some higher range here to intentionally keep these separate from the rest and give us the ability to add more on the future.

src/AppInstallerCommonCore/Public/AppInstallerLanguageUtilities.h
```diff
@@ -138,7 +138,7 @@ namespace AppInstaller
         const typename Variant::variant_t& GetVariant(Enum e) const
         {
             auto itr = m_data.find(e);
-            THROW_HR_IF_MSG(E_NOT_SET, itr == m_data.cend(), "GetVariant(%d)", e);
+            THROW_HR_IF_MSG(E_NOT_SET, itr == m_data.cend(), "GetVariant(%d)", static_cast<int>(e));
```
JohnMcPMS, line 141: @sreadingMSFT Apparently you were not the only one to see it even if it didn't give me any issues.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -427,15 +462,17 @@ namespace AppInstaller::CLI::Workflow
         context <<
             GetInstallerArgs <<
             RenameDownloadedInstaller <<
-            ShellExecuteInstallImpl;
+            ShellExecuteInstallImpl <<
+            ReportInstallerResult("ShellExecute"sv);
```
JohnMcPMS, line 466: I think that MSIX should also be normalized to flow through this path.

JohnMcPMS (commented):
We need to make sure we have the full set of cases covered; "reboot required" being missing suggests that we have some to get still.

Also, the arguably more important part of this change is having a healthy set of default values per installer type and thorough thought put into how those would be overridden by the manifest.  Requiring every MSI to output the fairly standardized set of values is not helpful.

This also plays into my comment on MSIX, as we can normalize the HRESULT returns from deployment into the set of values that we have here to make all package types easier to work with in a uniform fashion.

src/AppInstallerCommonCore/Public/AppInstallerLanguageUtilities.h
sreadingMSFT: Great. Glad the fix was easy enough, the build error was not particularly helpful.

yao-msft: Ashwini and I both saw this yesterday after we merge with latest, we fixed it the same way. Looks like some new code called into this const path.

lechacon:
Updated with default values per installer type.
I changed the way this is represented in the manifest because I hadn't considered multiple return codes mapping to the same response. The format is now
```
      ExpectedReturnCodes:
          - InstallerReturnCode: 1
            ReturnResponse: PackageInUse
```

schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
```diff
@@ -132,20 +132,56 @@
         }
       }
     },
+    "InstallerReturnCode": {
+      "type": "integer",
+      "not": {
+        "enum": [ 0 ]
+      },
+      "minimum": -2147483648,
+      "maximum": 429496729,
```
yao-msft, line 141: >429496729 [](http://example.com/codeflow?start=17&length=9)

Did you accidentally change the value, the previous 4294967295 is the max for uint32

lechacon: Yes, it was accidental. I'm not even sure how I managed to get a copy-paste wrong 😆

schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
```diff
@@ -132,20 +132,56 @@
...
+    "ExpectedReturnCodes": {
+      "type": [ "array", "null" ],
+      "items": {
+        "type": "object",
+        "properties": {
+          "InstallerReturnCode": {
+            "$ref": "#/definitions/InstallerReturnCode"
+          },
+          "ReturnResponse": {
+            "type": [ "string", "null" ],
```
yao-msft, line 162: >"null" [](http://example.com/codeflow?start=32&length=6)

This probably can not be null since we don't have defaults for them, let's remove the null from type

schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
```diff
@@ -132,20 +132,56 @@
...
+            "enum": [
+              "PackageInUse",
```
yao-msft, line 164: >PackageInUse [](http://example.com/codeflow?start=15&length=12)

nit: our enum values use camel case throughout the winget manifest

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -468,11 +519,9 @@ namespace AppInstaller::CLI::Workflow
         }
         catch (const wil::ResultException& re)
         {
-            const auto& manifest = context.Get<Execution::Data::Manifest>();
-            Logging::Telemetry().LogInstallerFailure(manifest.Id, manifest.Version, manifest.Channel, "MSIX", re.GetErrorCode());
-
-            context.Reporter.Error() << GetUserPresentableMessage(re) << std::endl;
-            AICLI_TERMINATE_CONTEXT(re.GetErrorCode());
+            context.Add<Execution::Data::InstallerReturnCode>(re.GetErrorCode());
+            context << ReportInstallerResult("MSIX", re.GetErrorCode(), /* isHResult */ true);
```
yao-msft, line 523: >" [](http://example.com/codeflow?start=50&length=1)

nit: sv to match above?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1097,4 +1097,58 @@ Do you agree to the terms?</value>
...
     <value>Admin setting disabled.</value>
   </data>
+  <data name="InstallFlowReturnCodePackageInUse" xml:space="preserve">
+    <value>Application is currently running. Exit the application then try again.</value>
+  </data>
+  <data name="InstallFlowReturnCodeInstallInProgress" xml:space="preserve">
+    <value>Another installation is already in progress. Try again later.</value>
+  </data>
+  <data name="InstallFlowReturnCodeFileInUse" xml:space="preserve">
+    <value>One or more file is being used. Exit the application then try again.</value>
```
yao-msft, line 1107: >One or more file is [](http://example.com/codeflow?start=11&length=19)

nit: One or more files are ...

src/AppInstallerRepositoryCore/Rest/Schema/1_1/Json/ManifestDeserializer_1_1.cpp
```diff
@@ -123,6 +126,27 @@ namespace AppInstaller::Repository::Rest::Schema::V1_1::Json
...
             }
+
+            // Expected return codes
+            std::optional<std::reference_wrapper<const web::json::array>> expectedReturnCodesNode = JsonHelper::GetRawJsonArrayFromJsonNode(installerJsonObject, JsonHelper::GetUtilityString(ExpectedReturnCodes));
+            if (expectedReturnCodesNode)
+            {
+                for (auto& returnCodeNode : expectedReturnCodesNode.value().get())
+                {
+                    ExpectedReturnCodeEnum returnResponse = Manifest::ConvertToExpectedReturnCodeEnum(JsonHelper::GetRawStringValueFromJsonNode(returnCodeNode, JsonHelper::GetUtilityString(ReturnResponse)).value_or(""));
+                    DWORD installerReturnCode = JsonHelper::GetRawIntValueFromJsonNode(returnCodeNode, JsonHelper::GetUtilityString(InstallerReturnCode)).value_or(0);
```
yao-msft, line 137: >JsonHelper::GetRawIntValueFromJsonNode [](http://example.com/codeflow?start=48&length=38)

Code analysis did not complain about converting from signed to unsigned? I think we need a static cast here

schemas/JSON/manifests/v1.1.0/manifest.singleton.1.1.0.json
```diff
@@ -165,20 +165,57 @@
...
+              "RebootInitiated",
+              "CancelledByUser",
+              "AlreadyInstalled",
+              "Downgrade",
+              "BlockedByPolicy"
+            ]
+          }
+        }
+      },
+      "maxItems": 16,
```
yao-msft, line 216: >16, [](http://example.com/codeflow?start=18&length=3)

Since we allow many to one mapping now, shall we increase the maxItems? I'd think 128

Also, please make the definition in this file and the other file consistent, the installer schema did not have maxItems

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -79,20 +79,38 @@
...
+#define APPINSTALLER_CLI_ERROR_INSTALL_INSUFFICIENT_MEMORY      ((HRESULT)0x8a150106)
+#define APPINSTALLER_CLI_ERROR_INSTALL_NO_NETWORK               ((HRESULT)0x8a150107)
+#define APPINSTALLER_CLI_ERROR_INSTALL_CONTACT_SUPPORT          ((HRESULT)0x8a150108)
+#define APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_FINISH  ((HRESULT)0x8a150109)
+#define APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_INSTALL ((HRESULT)0x8a150110)
+#define APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_INITIATED         ((HRESULT)0x8a150111)
+#define APPINSTALLER_CLI_ERROR_INSTALL_CANCELLED_BY_USER        ((HRESULT)0x8a150112)
+#define APPINSTALLER_CLI_ERROR_INSTALL_ALREADY_INSTALLED        ((HRESULT)0x8a150113)
+#define APPINSTALLER_CLI_ERROR_INSTALL_DOWNGRADE                ((HRESULT)0x8a150114)
+#define APPINSTALLER_CLI_ERROR_INSTALL_BLOCKED_BY_POLICY        ((HRESULT)0x8a150115)
```
yao-msft, line 105: >APPINSTALLER_CLI_ERROR_INSTALL_BLOCKED_BY_POLICY [](http://example.com/codeflow?start=8&length=48)

nit: these error codes could be added to e2etests error list for completeness

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -432,15 +481,17 @@ namespace AppInstaller::CLI::Workflow
...
     }
 
     void DirectMSIInstall(Execution::Context& context)
     {
         context <<
             GetInstallerArgs <<
             RenameDownloadedInstaller <<
-            DirectMSIInstallImpl;
+            DirectMSIInstallImpl <<
+            ReportInstallerResult("MsiInstallProduct"sv, APPINSTALLER_CLI_ERROR_MSI_INSTALL_FAILED, /* isHResult */ true);
```
yao-msft, line 494: >/* isHResult */ true [](http://example.com/codeflow?start=100&length=20)

MsiInstallProduct returns win32 error codes(like 1603), not hresults. So this should be false.

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -282,6 +282,74 @@ namespace AppInstaller::Manifest
         }
     }
 
+    ExpectedReturnCodeEnum ConvertToExpectedReturnCodeEnum(const std::string& in)
+    {
+        ExpectedReturnCodeEnum result = ExpectedReturnCodeEnum::Unknown;
+
+        if (Utility::CaseInsensitiveEquals(in, "PackageInUse"))
```
yao-msft, line 289: >Utility::CaseInsensitiveEquals [](http://example.com/codeflow?start=12&length=30)

nit: since there're more than a couple strings, we might want to do toLower than do direct compare, like what we did for InstallerTypeEnum

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -410,4 +478,48 @@ namespace AppInstaller::Manifest
...
+            {
+                { 2, ExpectedReturnCodeEnum::CancelledByUser },
+                { 5, ExpectedReturnCodeEnum::CancelledByUser },
+                { 8, ExpectedReturnCodeEnum::RebootRequiredForInstall },
+            };
+        case InstallerTypeEnum::Msix:
+            // See https://docs.microsoft.com/en-us/windows/win32/appxpkg/troubleshooting
+            return
+            {
+                { ERROR_INSTALL_PREREQUISITE_FAILED, ExpectedReturnCodeEnum::MissingDependency },
```
yao-msft, line 513: >ERROR_INSTALL_PREREQUISITE_FAILED [](http://example.com/codeflow?start=18&length=33)

For MSIX apis, the returned error codes are HResults, so you'll have to convert these win 32 error codes to HResults,
otherwise, they may not get matched. Use below in allmsix return codes
HRESULT_FROM_WIN32(ERROR_INSTALL_PREREQUISITE_FAILED)

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -162,6 +162,20 @@ namespace AppInstaller::Manifest
...
             }
+
+            // Check expected return codes for duplicates between successful and expected error codes
+            std::set<DWORD> returnCodeSet;
+            returnCodeSet.insert(installer.InstallerSuccessCodes.begin(), installer.InstallerSuccessCodes.end());
+            for (const auto& code : installer.ExpectedReturnCodes)
+            {
+                if (!returnCodeSet.insert(code.first).second)
+                {
+                    resultErrors.emplace_back(ManifestError::DuplicateReturnCodeEntry);
```
yao-msft, line 173: >DuplicateReturnCodeEntry [](http://example.com/codeflow?start=61&length=24)

A negative test case for semantic checks would be ideal

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -625,6 +639,31 @@ namespace AppInstaller::Manifest
         return resultErrors;
     }
 
+    ValidationErrors ManifestYamlPopulator::ProcessExpectedReturnCodesNode(const YAML::Node& returnCodesNode)
```
yao-msft, line 642: >ProcessExpectedReturnCodesNode [](http://example.com/codeflow?start=44&length=30)

As of convention and completeness, please add the new manifest entries in MultiFileManifestV1_1 and ManifestV1_1-Singleton.yaml and update the VerifyManifestContent in unit tests.

src/AppInstallerCLIE2ETests/Constants.cs
```diff
@@ -134,6 +134,25 @@ public class ErrorCode
             public const int ERROR_SOURCE_OPEN_FAILED = unchecked((int)0x8a150045);
             public const int ERROR_SOURCE_AGREEMENTS_NOT_ACCEPTED = unchecked((int)0x8a150046);
             public const int ERROR_CUSTOMHEADER_EXCEEDS_MAXLENGTH = unchecked((int)0x8a150047);
+            public const int ERROR_MISSING_RESOURCE_FILE = unchecked((int)0x8a150048);
+            public const int ERROR_MSI_INSTALL_FAILED = unchecked((int)0x8a150049);
+            public const int ERROR_INVALID_MSIEXEC_ARGUMENT = unchecked((int)0x8a150050);
```
yao-msft, line 139: >50 [](http://example.com/codeflow?start=85&length=2)

just noticed this should be 4A

src/AppInstallerCLIE2ETests/Constants.cs
```diff
@@ -134,6 +134,25 @@ public class ErrorCode
...
+            public const int ERROR_INSTALL_PACKAGE_IN_USE = unchecked((int)0x8a150101);
+            public const int ERROR_INSTALL_INSTALL_IN_PROGRESS = unchecked((int)0x8a150102);
+            public const int ERROR_INSTALL_FILE_IN_USE = unchecked((int)0x8a150103);
+            public const int ERROR_INSTALL_MISSING_DEPENDENCY = unchecked((int)0x8a150104);
+            public const int ERROR_INSTALL_DISK_FULL = unchecked((int)0x8a150105);
+            public const int ERROR_INSTALL_INSUFFICIENT_MEMORY = unchecked((int)0x8a150106);
+            public const int ERROR_INSTALL_NO_NETWORK = unchecked((int)0x8a150107);
+            public const int ERROR_INSTALL_CONTACT_SUPPORT = unchecked((int)0x8a150108);
+            public const int ERROR_INSTALL_REBOOT_REQUIRED_TO_FINISH = unchecked((int)0x8a150109);
+            public const int ERROR_INSTALL_REBOOT_REQUIRED_TO_INSTALL = unchecked((int)0x8a150110);
```
yao-msft, line 150: >10 [](http://example.com/codeflow?start=95&length=2)

0A, 0B ...

src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
```diff
@@ -337,6 +340,12 @@
     <CopyFileToFolders Include="TestData\Manifest-Bad-DuplicateKey.yaml">
       <DeploymentContent>true</DeploymentContent>
     </CopyFileToFolders>
+    <CopyFileToFolders Include="TestData\Manifest-Bad-DuplicateReturnCode-ExpectedCodes.yaml">
```
yao-msft, line 343: >Manifest-Bad-DuplicateReturnCode-ExpectedCodes.yaml [](http://example.com/codeflow?start=41&length=51)

Looks like the new test files forgot tobe included in the commit causing test failure

yao-msft (approved):
:shipit:

Issue #1243, opened 2021-07-01
Manifest Schema v1.1


denelon:
# Description of the new feature/enhancement

We will be adding multiple new nodes to the schema to support future features. The schema suggestions are in the following features:

* #200
* #713
* #873
* #967
* #970
* #993
* #1073
* #1145
* #1163
* #1227
* #1241
* #1242
* #1278
* #1284

# Proposed technical implementation details (optional)

These schema values will be added, but the features may not be implemented in the next release. This will help consolidate the complex efforts related to schema changes across the Windows Package Manager client, the Windows Package Manager Manifest Creator, the validation pipelines, the publishing pipelines, and our periodic validation services.

Karl-WE:
Are the new schemas reflected in the YAML store app UI?
source: https://www.microsoft.com/en-us/p/yaml-generator-for-windows-package-manager/9p3n60fs22k5?activetab=pivot:overviewtab

ItzLevvie:
You'll have to contact the application developer to allow creation of 1.0 and above manifests.

denelon:
I contacted the author to let them know about the schema changes.

ptorr-msft:
Thanks - I have been on leave for most of the year so wasn't keeping up. I have temporarily hidden the app and will work with @denelon to see if it's worth updating or not.

denelon:
> Are the new schemas reflected in the YAML store app UI?
> source: https://www.microsoft.com/en-us/p/yaml-generator-for-windows-package-manager/9p3n60fs22k5?activetab=pivot:overviewtab

I don't believe anything other than the Windows Package Manager client and the reference implementation for the REST API have the new schema at this time.

Karl-WE:
@denelon the project of YAML generator seems not to be active atm
https://github.com/ptorr-msft/WinGetYamlGenerator/graphs/code-frequency
@ptorr-msft

denelon:
It was removed now that we have [wingetcrate](https://github.com/microsoft/winget-create). Be sure to 👍 microsoft/winget-create/issues/159 if you want a GUI.

ptorr-msft:
Right, I de-listed it from the Store. I'll update the GitHub readme as well.

Issue #967, opened 2021-05-17
Max Installer nodes 1024


denelon:
# Description of the new feature/enhancement

The maximum number of installer nodes should be increased. ~[Firefox has 231](https://github.com/microsoft/winget-pkgs/discussions/13266#discussioncomment-748775).~ [Firefox has 291](https://github.com/microsoft/winget-pkgs/discussions/13266#discussioncomment-748844).

# Proposed technical implementation details (optional)

Increase the maximum number of installer nodes to ~256~ ~512~ 1024.
This needs to be implemented in the:
* Client validation - https://github.com/microsoft/winget-cli/issues/967
* Manifest schema - https://github.com/microsoft/winget-cli/issues/967
* PR validation - https://github.com/microsoft/winget-pkgs/issues/13373
* REST schema

PR #1427, opened 2021-09-02
Add support for remaining manifest v1.1 fields

yao-msft:
## Change
- Added more fields to manifest v1.1 schema
- Added yaml parsing support for new fields
- Added rest support for new fields

## Validation
Added unit tests and also tested manually.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1427)

### DIFF
--- a/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
+++ b/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
@@ -293,22 +293,23 @@
     },
     "Markets": {
       "description": "The installer markets",
+      "type": [ "object", "null" ],
       "oneOf": [
         {
-          "type": "object",
           "properties": {
             "AllowedMarkets": {
               "$ref": "#/definitions/MarketArray"
             }
-          }
+          },
+          "required": [ "AllowedMarkets" ]
         },
         {
-          "type": "object",
           "properties": {
             "ExcludedMarkets": {
               "$ref": "#/definitions/MarketArray"
             }
-          }
+          },
+          "required": [ "ExcludedMarkets" ]
         }
       ]
     },
--- a/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
+++ b/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
@@ -386,6 +386,15 @@
       },
       "description": "List of ARP entries."
     },
+    "ElevationRequirement": {
+      "type": [ "string", "null" ],
+      "enum": [
+        "elevationRequired",
+        "elevationProhibited",
+        "elevatesSelf"
+      ],
+      "description": "The installer's elevation requirement"
+    },
     "Installer": {
       "type": "object",
       "properties": {
@@ -487,6 +496,9 @@
         },
         "AppsAndFeaturesEntries": {
           "$ref": "#/definitions/AppsAndFeaturesEntries"
+        },
+        "ElevationRequirement": {
+          "$ref": "#/definitions/ElevationRequirement"
         }
       },
       "required": [
@@ -579,13 +591,16 @@
     "AppsAndFeaturesEntries": {
       "$ref": "#/definitions/AppsAndFeaturesEntries"
     },
+    "ElevationRequirement": {
+      "$ref": "#/definitions/ElevationRequirement"
+    },
     "Installers": {
       "type": "array",
       "items": {
         "$ref": "#/definitions/Installer"
       },
       "minItems": 1,
-      "maxItems": 128
+      "maxItems": 1024
     },
     "ManifestType": {
       "type": "string",
--- a/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
+++ b/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
@@ -138,7 +138,9 @@
         "type": "integer",
         "not": {
           "enum": [ 0 ]
-        }
+        },
+        "minimum": -2147483648,
+        "maximum": 4294967295
       },
       "maxItems": 16,
       "uniqueItems": true,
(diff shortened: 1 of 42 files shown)


schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
```diff
@@ -426,6 +558,27 @@
...
+      "$ref": "#/definitions/RequireExplicitUpgrade"
+    },
+    "UnsupportedOSArchitectures": {
+      "$ref": "#/definitions/UnsupportedOSArchitectures"
+    },
+    "AppsAndFeaturesEntries": {
+      "$ref": "#/definitions/AppsAndFeaturesEntries"
+    },
     "Installers": {
       "type": "array",
```
jedieaston, line 583: Should the length of the installers array be bumped to 1024 here, per https://github.com/microsoft/winget-cli/issues/967?

yao-msft: Yes, thanks for catching that!

schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json
```diff
@@ -275,6 +277,115 @@
...
+    "ReleaseDate": {
+      "type": [ "string", "null" ],
+      "format": "date",
+      "description": "The installer release date"
+    },
+    "InstallLocationRequired": {
+      "type": [ "boolean", "null" ],
+      "description": "Indicates whether the installer requires an install location provided"
+    },
+    "RequireExplicitUpgrade": {
```
yao-msft, line 329: >RequireExplicitUpgrade [](http://example.com/codeflow?start=5&length=22)

Also add RequireElevation field

JohnMcPMS: I think we need an enum for that, but I cannot remember exactly the reasoning for that.  But at least at a minimum there are both elevation required installers and elevation rejecting installers.  So it would be better to have an enum value that we can set `ElevationBehavior` or something like that.

JohnMcPMS: More thoughts on values for it:
```
ElevationRequired // intended to indicate that the caller must be elevated because the installer will not elevate itself
ElevatesSelf
RestrictsSelf // Better word than restrict?
RestrictedRequired
```
Then we could add more as we find need of them.  For instance `ElevationRequiredForSilent` was a problem for MSI before we fixed it, but I know that for instance WebView2 falls into this category currently.  It self elevates for its normal unattended pattern, but does not for a silent install (basically same as msiexec behavior).

yao-msft: Per offline chat, we'll do an enum of elevationRequired, elevationProhibited and elevatesSelf

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -249,6 +263,23 @@ namespace AppInstaller::Manifest
                     std::move(v1InstallerFields.begin(), v1InstallerFields.end(), std::inserter(result, result.end()));
                 }
             }
+
+            if (manifestVersion >= ManifestVer{ s_ManifestVersionV1_1 })
+            {
+                std::vector<FieldProcessInfo> fields_v1_1 =
```
JohnMcPMS, line 269: I don't see anywhere that we are limiting what the server will accept.  Letting manifests into winget-pkgs when the feature behind them has not yet been implemented will likely lead to failures.  Either now because someone relies on it for a package and it doesn't actually do anything or in the future when we start using it and find that we wanted validation to prevent some case that is now already present.

It feels like we should have:
1. Add new version to client to allow for local testing; full validation fails if any 1.1 fields are present even if manifest is 1.1
2. As we add functionality and validation, we start allowing fields that are ready

JohnMcPMS (commented):
Didn't look at REST code; letting @ashpatil-msft look at that part.

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
yao-msft: Evaluated 1 by 1 offline and we're ok with them for now

src/AppInstallerRepositoryCore/Rest/Schema/1_1/Json/ManifestDeserializer_1_1.cpp
```diff
@@ -12,6 +12,30 @@ namespace AppInstaller::Repository::Rest::Schema::V1_1::Json
...
+        constexpr std::string_view DisplayName = "DisplayName"sv;
+        constexpr std::string_view Publisher = "Publisher"sv;
+        constexpr std::string_view DisplayVersion = "DisplayVersion"sv;
+        constexpr std::string_view ProductCode = "ProductCode"sv;
+        constexpr std::string_view UpgradeCode = "UpgradeCode"sv;
+        constexpr std::string_view InstallerType = "InstallerType"sv;
+        constexpr std::string_view Markets = "Markets"sv;
+        constexpr std::string_view AllowedMarkets = "AllowedMarkets"sv;
+        constexpr std::string_view ExcludedMarkets = "ExcludedMarkets"sv;
+        constexpr std::string_view ElevationRequirement = "ElevationRequirement"sv;
```
ashpatil-msft, line 30: >ElevationRequirement [](http://example.com/codeflow?start=35&length=20)

Didnt see this in the Rest Schema 1.1. Was this missed and needs to be added there?

yao-msft: Yes, this(From comments above) and expected return codes(Flor's) are 2 new fields added after our v1.1 schema review meeting, we'll have to update the rest spec later.

src/AppInstallerRepositoryCore/Rest/Schema/1_1/Json/ManifestDeserializer_1_1.cpp
```diff
@@ -35,6 +59,110 @@ namespace AppInstaller::Repository::Rest::Schema::V1_1::Json
...
+            }
+
+            // Markets
+            std::optional<std::reference_wrapper<const web::json::value>> marketsNode = JsonHelper::GetJsonValueFromNode(installerJsonObject, JsonHelper::GetUtilityString(Markets));
+            if (marketsNode && !marketsNode.value().get().is_null())
+            {
+                installer.Markets.ExcludedMarkets = V1_0::Json::ManifestDeserializer::ConvertToManifestStringArray(
+                    JsonHelper::GetRawStringArrayFromJsonNode(marketsNode.value().get(), JsonHelper::GetUtilityString(ExcludedMarkets)));
+                installer.Markets.AllowedMarkets = V1_0::Json::ManifestDeserializer::ConvertToManifestStringArray(
+                    JsonHelper::GetRawStringArrayFromJsonNode(marketsNode.value().get(), JsonHelper::GetUtilityString(AllowedMarkets)));
```
ashpatil-msft, line 124: Do we thro if both are given?

yao-msft: Yes, it's in the manifestvalidation.cpp

ashpatil-msft (approved):
:shipit:

denelon:
Related to #1243

PR #1432, opened 2021-09-04
No package name provided

### LINKED ISSUE #1131: Winget install/show without arguments should not produce a list of all packages
When typing `winget install` or  `winget show` the CLI currently outputs a list of all packages with a mention there have been too many packages found. This floods the terminal and is not user-friendly. When no arguments are given it should default to the help message or simply a warning the user has to provide a package name. 

If `install`or `show` have no arguments, the terminal should respond with: `Please provide a package name` or refer to the help message of the command.

ChungZH:
close #1131

-----
image
image
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1432)

### DIFF
--- a/src/AppInstallerCLICore/Commands/InstallCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/InstallCommand.cpp
@@ -122,6 +122,11 @@ namespace AppInstaller::CLI
 
     void InstallCommand::ExecuteInternal(Context& context) const
     {
+        if (context.Args.GetArg(Execution::Args::Type::Query).empty())
+        {
+            context.Reporter.Info() << Resource::String::NoPackageNameProvided;
+            return;
+        }
         context <<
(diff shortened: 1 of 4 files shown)


src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -122,6 +122,11 @@ namespace AppInstaller::CLI
 
     void InstallCommand::ExecuteInternal(Context& context) const
     {
+        if (context.Args.GetArg(Execution::Args::Type::Query).empty())
```
yao-msft, line 125: Unfortunately this will block usages like "winget install --id Microsoft.Terminal"

We might need a more sophisticated logic to detect "there's nothing in the search request".

Also I think this should apply to uninstall command as well, in this case, we may want to move the logic to base command.

PR #1440, opened 2021-09-08
Add colons after agreement labels

lechacon:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1440)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ShowFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShowFlow.cpp
@@ -8,6 +8,7 @@
 
 using namespace AppInstaller::Repository;
 using namespace AppInstaller::CLI;
+using namespace AppInstaller::Utility::literals;
 
 namespace AppInstaller::CLI::Workflow
 {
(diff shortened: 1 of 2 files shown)


JohnMcPMS (approved):

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -97,7 +97,7 @@ namespace AppInstaller::CLI::Workflow
             {
                 if (!agreement.Label.empty())
                 {
-                    info << Execution::ManifestInfoEmphasis << agreement.Label << " ";
+                    info << Execution::ManifestInfoEmphasis << agreement.Label << ": ";
```
yao-msft, line 100: If possible, can you add the colons to source agreement labels too? I think I followed the package agreements style when doing source agreements work

lechacon: Added.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -158,7 +158,7 @@ namespace AppInstaller::CLI::Workflow
             {
                 if (!agreement.Label.empty())
                 {
-                    context.Reporter.Info() << Execution::SourceInfoEmphasis << Utility::LocIndString{ agreement.Label } << " ";
+                    context.Reporter.Info() << Execution::SourceInfoEmphasis << Utility::LocIndString{ agreement.Label } << ": ";
```
JohnMcPMS, line 161: Realized that they should really have `_liv` for some magical future where I get around to enforcing localization.

lechacon: Added it to all these strings in Show. I realized there are still some non-localized strings for showing dependencies. I'll update these on my next change.

PR #1441, opened 2021-09-08
Support for x64 on ARM64

### LINKED ISSUE #666: Revisit the logic in Architecture.cpp for X64 emulation of ARM64 devices.

Given the X64 emulation of ARM64 devices went publicly preview this week. We'd better add X64 to ApplicableArchitecturesVector for ARM64 devices. 

It should be a trivial change by adding an entry to ApplicableArchitecturesVector. However I don't know how to check if the system supports X64 emulation.

JohnMcPMS:
## Issue
Support for x64 emulation on ARM64 was [added to Windows 10 Dev builds last year](https://blogs.windows.com/windows-insider/2020/12/10/introducing-x64-emulation-in-preview-for-windows-10-on-arm-pcs-to-the-windows-insider-program/).  In order for us to detect the presence of that support, a new API was needed.

## Change
This change fixes #666 by implementing dynamic loading of the API and invoking it specifically for x64 when the machine is ARM64.  The API can replace our current use of `IsWow64GuestMachineSupported`, but there is no need to take the additional risk of that change at this time.

## Validation
As we do not have any automated test bed for ARM64, I was only able to test it manually on a local device.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1441)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -130,6 +130,7 @@ hhx
 HINSTANCE
 hkey
 hlocal
+hmodule
 hre
 hresults
 htm
@@ -246,6 +247,7 @@ pkindex
 PMS
 positionals
 powertoys
+processthreads
 productcode
 pseudocode
 pvk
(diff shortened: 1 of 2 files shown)


yao-msft (approved):
:shipit:

PR #1453, opened 2021-09-10
Disable local manifest by default

yao-msft:
## Change
Due to security concerns, this change disables local manifest files by default. User can still turn on local manifest files through group policy, or through running 'winget settings --enable LocalManifestFiles' as administrator.

## Validation
Added tests. Also tested manually.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1453)

### DIFF
--- a/src/AppInstallerCommonCore/AdminSettings.cpp
+++ b/src/AppInstallerCommonCore/AdminSettings.cpp
@@ -1,0 +1,183 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "AppInstallerLogging.h"
+#include "AppInstallerStrings.h"
+#include "winget/Settings.h"
+#include "winget/AdminSettings.h"
+#include "winget/GroupPolicy.h"
+#include "winget/Yaml.h"
+
+namespace AppInstaller::Settings
+{
+    using namespace std::string_view_literals;
+    using namespace Utility;
+
+    namespace
+    {
+        constexpr std::string_view s_AdminSettingsYaml_LocalManifestFiles = "LocalManifestFile"sv;
+
+        // Attempts to read a single scalar value from the node.
+        template<typename Value>
+        bool TryReadScalar(const YAML::Node& rootNode, std::string_view name, Value& value)
+        {
+            YAML::Node valueNode = rootNode[std::string{ name }];
+
+            if (!valueNode || !valueNode.IsScalar())
+            {
+                AICLI_LOG(Core, Info, << "Admin setting '" << name << "' was not found or did not contain the expected format");
+                return false;
+            }
+
+            value = valueNode.as<Value>();
+            return true;
+        }
+
+        struct AdminSettingValues
+        {
+            bool LocalManifestFiles = false;
+        };
+
+        struct AdminSettingsInternal
+        {
+            AdminSettingsInternal();
+
+            void SetAdminSetting(AdminSetting setting, bool enabled);
+
+            bool GetAdminSettingBoolValue(AdminSetting setting) const;
+
+        private:
+            void LoadAdminSettings();
+            void SaveAdminSettings() const;
+
+            AdminSettingValues m_settingValues;
+        };
+
+        AdminSettingsInternal::AdminSettingsInternal()
+        {
+            LoadAdminSettings();
+        }
+
+        void AdminSettingsInternal::SetAdminSetting(AdminSetting setting, bool enabled)
+        {
+            switch (setting)
+            {
+            case AdminSetting::LocalManifestFiles:
+                m_settingValues.LocalManifestFiles = enabled;
+                break;
+            default:
+                return;
+            }
+
+            SaveAdminSettings();
+        }
+
+        bool AdminSettingsInternal::GetAdminSettingBoolValue(AdminSetting setting) const
+        {
+            switch (setting)
+            {
+            case AdminSetting::LocalManifestFiles:
+                return m_settingValues.LocalManifestFiles;
+            default:
+                return false;
+            }
+        }
+
+        void AdminSettingsInternal::LoadAdminSettings()
+        {
+            auto stream = Settings::GetSettingStream(Settings::Streams::AdminSettings);
+            if (!stream)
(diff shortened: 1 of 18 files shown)


denelon:
@ImJoakim
@ItzLevvie
@jedieaston
@KaranKad
@OfficialEsco
@quhxl

Hey, I just wanted to let you know this change is coming. The PowerShell scripts will likely need to be modified for testing local manifest files once this change is released. We had a security review, and the decision was made to make this disabled by default. Most users will not need to execute local manifest files. This requires administrative access to modify, so it's not in the local settings file.

OfficialEsco:
> Most users will not need to execute local manifest files.

Seems fair since there is just about max 10 active contributors that might have to test a package locally..
Might have to make it a bit more user friendly for the people who submit one or two manifests

```
[ ] Have you tested your manifest locally with winget install --manifest <path>?
```

What if there is a way to temporary allow local manifests instead of toggling it once to use it one time?

denelon:
That's an interesting approach. Are you thinking about having the command just work for the "next" execution rather than an "on/off" switch?

yao-msft:
> What if there is a way to temporary allow local manifests instead of toggling it once to use it one time?
For now, the toggle setting stays until it's changed explicitly by user later. If users want to test local manifests, they'll run 'winget settings --enable LocalManifestFiles', after the testing is done, they could use 'winget settings --disable LocalManifestFiles' to disable it.

If there's need in the future, we can add an option to make the toggle be effective only once.

OfficialEsco:
> Are you thinking about having the command just work for the "next" execution rather than an "on/off" switch?

I have PowerShell's execution policy on my mind, however since winget is a appx package there is a lot less to worry about, we do need a on/off switch, however i believe it would be nice for people to have a next switch, i was thinking of `Set-ExecutionPolicy -Scope Process`, however since it requires Administrator to change the policy `winget install` would run as Administrator which we do not prefer when users test their manifest..

yao-msft:
Given the current code structure, I think I can easily add an option '_winget settings --enable LocalManifestFiles --once_', which will allow next local manifest invocation(could be either non elevated or elevated), then the feature is disabled after invocation.

Does that sort of satisfy the "temporarily enable" ask? Thanks.

jedieaston:
Two questions:

- Can we have a GPO to configure this? 
- What was the security concern brought up by local manifests? I don't understand what you can do with winget that you can't do with `Invoke-WebRequest` or a browser. It seems like organizations concerned about this vector would already be using [AppLocker](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-control/applocker/applocker-overview) or something since that applies to all executables a user tries to run, not just ones they wrote a winget manifest for. I guess it doesn't really matter, but I'm curious.

OfficialEsco:
> (could be either non elevated or elevated)

Doesn't this defeat the purpose of a security risk?

If can answer, what is the security concerns you guys are thinking about?

yao-msft:
>Can we have a GPO to configure this?

Yes, there's an existing one named LocalManifestFiles. https://github.com/microsoft/winget-cli/blob/fee797bb5a51238e0b37b9590ee12d4f1acace7c/doc/admx/DesktopAppInstaller.admx#L46

>Doesn't this defeat the purpose of a security risk?

The 'winget settings --enable LocalManifestFiles --once' still needs to be invoked as elevated. The next run with local manifest input could be non elevated.

Regarding questions about what security concerns, I think mainly we don't want to be the entrypoint to install random apps potentially without user knowledge, but I guess I'll let @denelon elaborate.

jedieaston:
Shoot, that's what I get for not checking the admx file again. I remember that being there now, thanks!

JohnMcPMS:
> Given the current code structure, I think I can easily add an option '_winget settings --enable LocalManifestFiles --once_', which will allow next local manifest invocation(could be either non elevated or elevated), then the feature is disabled after invocation.
> 
> Does that sort of satisfy the "temporarily enable" ask? Thanks.

That can't work; it would require admin to set the "consumed" state or it would be possible to just set it back to unconsumed.

While we don't run with any more permissions than a potential attacker, we don't want to make it any easier to download and run some binary either.  Given that 99% of users will never run `winget` (even that is optimistic when we we will have a billion machines with the code at some point), and then 99% of the ones who do will never need the local manifest, it is just safer to disable this potential attack surface by default.  Happy to make it as easy as possible for anyone to turn it back on if they want, so long as they have the permissions.

jedieaston:
Alrighty, that makes sense. Just going to have to be another consideration when running local manifest tests. 

Thanks for looping me in @denelon.

yao-msft:
> it would require admin to set the "consumed" state or it would be possible to just set it back to unconsumed.

I think it could work. I was thinking to make the LocalManifestFiles setting be a tri-state value {Enabled, EnabledOnce, Disabled}. There'll not be an extra field to save 'consumed' state. This LocalManifestFiles value can only be set with admin as we are currently doing. Then during command execution, if we see a local manifest as input and the value is EnabledOnce, we just set the value to Disabled. This is the only scenario a non-elevated invocation will change the setting, and it's only EnabledOnce -> Disabled.

But anyway, as you said, given the rare use cases for this feature, I'm totally fine with an on/off switch only.

src/AppInstallerCommonCore/AdminSettings.cpp
```diff
@@ -0,0 +1,183 @@
...
+        AdminSettingsInternal adminSettingsInternal;
+        adminSettingsInternal.SetAdminSetting(setting, false);
+    }
+
+    bool IsAdminSettingEnabled(AdminSetting setting)
+    {
+        AdminSettingsInternal adminSettingsInternal;
+        bool isEnabled = adminSettingsInternal.GetAdminSettingBoolValue(setting);
+
+        // For some admin settings, even if it's disabled, if the corresponding policy is enabled then override it.
```
JohnMcPMS, line 174: I would put this above the code that just read in the settings stream and return directly so that we don't waste cycles doing that.

JohnMcPMS (approved):

PR #1455, opened 2021-09-10
Show source identifier with `source list`

lechacon:
* Show source identifier when doing `source list <name>`. This can be useful for creating json files for import.
* Get the source identifier when adding a REST source to ensure it is added to the sources yaml. We were getting it only when opening the source so it wouldn't get saved and could not show it with source list. Existing sources are not updated.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1455)


yao-msft (approved):
:shipit:

PR #1463, opened 2021-09-14
Color contamination fix

hackean-msft:
Reset console to default color when user force quits(hits ctrl + c twice) during execution. 
Channel all console output write operations only through OutputStream/BaseStream wrappers. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1463)


src/AppInstallerCLICore/ChannelStreams.h
```diff
@@ -45,24 +44,38 @@ namespace AppInstaller::CLI::Execution
...
         BaseStream& operator<<(const VirtualTerminal::Sequence& sequence);
         BaseStream& operator<<(const VirtualTerminal::ConstructedSequence& sequence);
 
+        void DisableVT() { m_VTEnabled = false; }
+
+        void Close();
+
     private:
+        template <typename T>
+        void Write(const T&t, bool byPassifNotEnabled)
```
yao-msft, line 69: >byPassif [](http://example.com/codeflow?start=35&length=8)

spellcheck failed here: bypassIf

src/AppInstallerCLICore/ExecutionReporter.cpp
```diff
@@ -209,6 +211,12 @@ namespace AppInstaller::CLI::Execution
         }
     }
 
+    void Reporter::RestoreDefault() 
+    {
+        EndProgress(true);
+        m_standardOutputStream.Close();
```
JohnMcPMS, line 217: I don't think that a function called `RestoreDefault` should close the output stream.  Without knowing, I would think that it resets the state, not that it prevents future output.

src/AppInstallerCLICore/ExecutionReporter.cpp
```diff
@@ -45,7 +46,7 @@ namespace AppInstaller::CLI::Execution
 
     OutputStream Reporter::GetOutputStream(Level level)
     {
-        OutputStream result = GetBasicOutputStream();
+        OutputStream result = m_standardOutputStream;
```
JohnMcPMS, line 49: By making all of these streams a copy of the original, you are not respecting changes to `m_channel` and VT enablement state.  You also are not gaining what I suspect you might think you are gaining; that closing `m_standardOutputStream` also closes all of the active streams.  Because they are all copies that will not happen.

It would be possible to create a single base object that is in between all attempts to write to the stream that the reporter is given, but my suggestion was to try to do that by operating inside the STL infrastructure.  Because you only care about being able to control whether output happens or not, it shouldn't be that hard.  Alternately, wrapping it would be fine too, but that should be done in a functional way that integrates with the existing code.

src/AppInstallerCLICore/ExecutionReporter.cpp
```diff
@@ -76,6 +77,7 @@ namespace AppInstaller::CLI::Execution
     void Reporter::SetChannel(Channel channel)
     {
         m_channel = channel;
+        m_standardOutputStream.DisableVT();
```
JohnMcPMS, line 80: This isn't correct.

src/AppInstallerCLICore/ChannelStreams.cpp
```diff
@@ -9,8 +9,8 @@ namespace AppInstaller::CLI::Execution
     using namespace Settings;
     using namespace VirtualTerminal;
 
-    BaseStream::BaseStream(std::ostream& out, bool enabled, bool VTEnabled) :
-        m_out(out), m_enabled(enabled), m_VTEnabled(VTEnabled) {}
+    BaseStream::BaseStream(std::ostream& out, bool enabled, bool VTEnabled, VirtualTerminal::Sequence sequence) :
```
JohnMcPMS, line 12:
```suggestion
    BaseStream::BaseStream(std::ostream& out, bool enabled, bool VTEnabled, VirtualTerminal::Sequence defaultSequence) :

```

src/AppInstallerCLICore/ChannelStreams.cpp
```diff
@@ -39,6 +39,15 @@ namespace AppInstaller::CLI::Execution
         return *this;
     }
 
+    void BaseStream::Close() 
+    {
+        m_enabled = false;
```
JohnMcPMS, line 44: If this were a single shared object that every `OutputStream` had a reference to, at least `m_enabled` would need to be updated to be `std::atomic_bool` to prevent a race at `Close`.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/ExecutionReporter.cpp
hackean-msft: The function name is the one thing I see changing here to something that clearly indicates that we're closing the output stream, since every write to output stream happens through the reporter. Also, restoring the default doesn't prevent a background asynchronous operation from altering the output stream. The goal is that if we ever get to this point, the reporter is about to be terminated and we should close.

JohnMcPMS (changes requested):
I think that my casual use of the word "base" in a previous statement may have lead to this design, which I don't think is the right pattern.  It is also bogged down by a lot of unnecessary name changes which make it a bit harder to see the actual functional changes.

My suggestion is to have the stream object that can be disabled as a singular member of the `Reporter`, replacing the current `std::ostream m_out`.  That object is given out in place of the current `m_out` and treated almost identically otherwise.  That keeps the code change minimal while achieving the goal of allowing all output to be disabled through a central mechanism.

Issue #1437, opened 2021-09-06
0x801901f4 : Internal server error (500) when I try to execute winget list and winget install (Source is down)


Fukubi:
# Brief description of your issue

When I try to execute winget list or winget install command I get an 0x801901f4 : Internal error in server (500) error.
# Steps to reproduce

Execute winget list, winget install.
```
[winget list]
Ocorreu um erro inesperado ao executar o comando:
0x801901f4 : Erro interno de servidor (500).

[winget install Mozilla.Firefox]
Ocorreu um erro inesperado ao executar o comando:
0x801901f4 : Erro interno de servidor (500).
```

# Expected behavior

List installed apps (with winget list) or install a new app (with winget install)

# Actual behavior

Throw an internal server error

# Environment

```none
[winget --info]
Windows Package Manager v1.0.11694
Copyright (c) Microsoft Corporation. Todos os direitos reservados.

Windows: Windows.Desktop v10.0.22000.176
Pacote: Microsoft.DesktopAppInstaller v1.15.11694.0

Registros: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir
```

jedieaston:
https://github.com/microsoft/winget-cli/issues/1349#issuecomment-913651947

Fukubi:
With ```--source winget``` it works perfectly

jedieaston:
Can you reopen this? It's the only issue noting that the source is down.

Fukubi:
Yeah, no problem.

leochras:
I just turned on my machine and also got this error..
![image](https://user-images.githubusercontent.com/16163571/132316909-9d10fad4-8ff6-41b3-a7d3-cb751c643a5b.png)

Caticer:
I have same issue + I can't download apps from microsoft store too

feoh:
Same here and it's super frustrating. Reproduced on both a System76 Thelio and a Lenovo Thinkpad T15 gen2 both running Windows 11 Insiders Beta/latest update.

I submitted feedback at: https://aka.ms/AAdr7rh

yooakim:
I have the same issue as well, frustrating.

```
PS C:\Users\joakim.westin> winget --info
Windows Package Manager v1.0.11694
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22000.176
[...]

PR #1470, opened 2021-09-15
Source resilience

### LINKED ISSUE #1349: Any REST source being unavailable completely breaks winget install/show/list/search/upgrade.

If a REST source has been added to winget and is unavailable, you are no longer able to use `winget install`, `show`, `list` , `upgrade` or `search` even if you have other sources which are available.

# Steps to reproduce

1. Add a rest source to winget via `winget source add --type "Microsoft.Rest" -n <name> <address of repository>` (in addition to the default source[s]).
2. Once the source is added successfully, shut down the server for the source or otherwise make it unavailable.
3. Try to use `winget install`, `show`, `list`, `upgrade` or `search` with valid arguments.

# Expected behavior

The command executes successfully since you have other sources available to query.

# Actual behavior

All commands fail since they cannot reach the source you added, even if other sources are available. The log file claims that winget is attempting to build an aggregated source to query against since there are multiple to choose from, but when the first one can't be contacted it bails out.

```
PS C:\Users\easton> winget install Microsoft.WindowsTerminal
Failed to open the source; try the 'source reset' command if the problem persists.
An unexpected error occurred while executing the command:
WinHttpSendRequest: 12029: A connection with the server could not be established

PS C:\Users\easton> winget search Microsoft.WindowsTerminal
Failed to open the source; try the 'source reset' command if the problem persists.
An unexpected error occurred while executing the command:
WinHttpSendRequest: 12029: A connection with the server could not be established

PS C:\Users\easton> winget upgrade
Failed to open the source; try the 'source reset' command if the problem persists.
An unexpected error occurred while executing the command:
WinHttpSendRequest: 12029: A connection with the server could not be established

PS C:\Users\easton> # Oh-no!
```

# Environment

```none
[winget --info]
Windows Package Manager v1.0.11694

Windows: Windows.Desktop v10.0.22000.100
Package: Microsoft.DesktopAppInstaller v1.15.11694.0

Links
--------------------------------------------------------

Any other software? This is running on ARM, although since it's the same package for all architectures I doubt that's relevant.
```

JohnMcPMS:
Fixes #1349, #1437 (although #1437 seems to have been re-opened more as a "the store is down" than "the client doesn't handle that well").

## Change
Improves handling of errors from one source when multiple are configured by capturing the errors and including them as part of the source open and search results.  Some commands will continue to fail if one of the sources has an error to prevent said error from affecting a change to the system.  The commands that still fail are `install`, `import`, and `upgrade --all|<specific>` and they now indicate which source failed.

Examples of new behavior:
```
> wingetdev search terminal
Failed when searching source; results will not be included: msstore
Name                          Id                               Version     Match             Source
---------------------------------------------------------------------------------------------------
Windows Terminal              Microsoft.WindowsTerminal        1.10.2383.0 Moniker: terminal winget
...

> wingetdev install terminal
Failed when searching source: msstore
An unexpected error occurred while executing the command:
WinHttpSendRequest: 12007: The server name or address could not be resolved

The following packages were found among the working sources.
Please specify one of them using the `--source` option to proceed.
Name             Id                        Source
-------------------------------------------------
Windows Terminal Microsoft.WindowsTerminal winget
```

This change also replaces the experimental `msstore` source with the more official but still green `msstore` source (previously called `storepreview`).  This source is now capable of providing basic search functionality against the entire Store catalog.  It does not provide any correlation for installed packages however.  In addition, the store source is being moved out of experimental.

## Validation
Added test source type that can be configured to fail in specific ways to enable easier validation of changes.
Unit tests added to ensure proper error handling.
E2E test added using test source type.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1470)


jedieaston:
It seems to me, as a user, `winget install` should still work without adding a source argument even if one of the sources is down. If I understand the PR description correctly, if I do `winget install terminal`, and the Microsoft Store is down (as an example), it doesn't matter if there's a great version of Windows Terminal available from the default repo, I just get an error message saying that the msstore is down, which I then use as a hint to start doing `winget search` so I can find a repo that is up and has it. 

I would think the expected behavior (coming from other package managers), would be that winget still installs whatever the closest thing it can find from available sources, since I presumably trust all of the sources I've added equally (or else I wouldn't have them). 

If that's too unsafe, then at the very least a more helpful message under the error would be nice, like:

```
> wingetdev install terminal
Failed to open a source; try the 'source reset' command if the problem persists: msstore
An unexpected error occurred while executing the command:
WinHttpSendRequest: 12007: The server name or address could not be resolved

Note: A match was found in the "winget" source. To install it, run winget install terminal --source winget.
(or, if multiple sources have a match, it shows the table)
```


which would save the trouble of me having to search for someone with the package myself, since presumably `winget install` already knows everyone who should have it.

denelon:
@JohnMcPMS what do you think about the suggestion from @jedieaston?

JohnMcPMS:
> If that's too unsafe

I don't think of it as a security issue, but rather a conservative take that we can't know what the user intent was.  So if there is a `terminal` in multiple sources, but one of those is down, how are we to know that the user didn't actually want the unavailable one?  Rather than performing a change to the system, I would rather fail "closed" for now.  We can consider changes to that based on broad feedback when it is in everyone's hands.

> maybe a hint under the error would be nice

This is a good idea and I will add it.

JohnMcPMS:
> This is a good idea and I will add it.

Ok, it is going to require a bit more work than I expected, but I think it will end up being a better experience overall anyway.

JohnMcPMS:
@jedieaston , I updated the PR description with the behavior of showing the search results from the not failing sources.

jedieaston:
That looks awesome! Thanks!

yao-msft:
>         }

nit, unrelated: Dependencies is missing

---
Refers to: schemas/JSON/settings/settings.schema.0.2.json:121 in c360e5d. [](commit_id = c360e5d278a7da771163c6effc39ab442cbe06da, deletion_comment = False)

yao-msft:
>                 Workflow::SearchSourceUsingManifest <<

I think this also performs a search against the sources, this is also used in upgrade command

but maybe it's ok we do not perform sophisticated handling for -m cases

---
Refers to: src/AppInstallerCLICore/Commands/UninstallCommand.cpp:119 in c360e5d. [](commit_id = c360e5d278a7da771163c6effc39ab442cbe06da, deletion_comment = False)

src/AppInstallerRepositoryCore/Public/AppInstallerRepositorySource.h
```diff
@@ -95,14 +95,14 @@ namespace AppInstaller::Repository
...
-        bool Restricted = false;
-
         // Custom header for Rest sources
         std::optional<std::string> CustomHeader;
 
         // Source information containing source agreements, required/unsupported match fields.
         SourceInformation Information;
+
+        // Prevent correlation against this source if true.
+        bool DoNotCorrelateAgainst = false;
```
yao-msft, line 105: >DoNotCorrelateAgainst [](http://example.com/codeflow?start=13&length=21)

SupportsCorrelation?

src/AppInstallerCLICore/Commands/ExportCommand.cpp
```diff
@@ -53,11 +53,14 @@ namespace AppInstaller::CLI
...
     void ExportCommand::ExecuteInternal(Execution::Context& context) const
     {
+        context.SetFlags(Execution::ContextFlag::TreatSourceFailuresAsWarning);
+
         context <<
             Workflow::ReportExecutionStage(Workflow::ExecutionStage::Discovery) <<
             Workflow::OpenSource <<
             Workflow::OpenCompositeSource(Repository::PredefinedSource::Installed) <<
             Workflow::SearchSourceForMany <<
+            Workflow::HandleSearchResultFailures <<
```
yao-msft, line 63: >HandleSearchResultFailures [](http://example.com/codeflow?start=22&length=26)

This is called after every SearchSourceForMany or SearchSourceForSingle, shall we consolidate them inside?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -454,6 +527,54 @@ namespace AppInstaller::CLI::Workflow
...
+            else
+            {
+                HRESULT overallHR = S_OK;
+                auto error = context.Reporter.Error();
+                for (const auto& failure : searchResult.Failures)
+                {
+                    error << Resource::String::SearchFailureError << ' ' << failure.Source->GetDetails().Name << std::endl;
+                    HRESULT failureHR = HandleException(context, failure.Exception);
+
+                    // Just take first failure for now
```
yao-msft, line 553: >Just take first failure for now [](http://example.com/codeflow?start=23&length=31)

Should we have a new hr for this instead of taking the first one?

src/AppInstallerCLIE2ETests/InstallCommand.cs
```diff
@@ -134,6 +134,27 @@ public void InstallMSIXWithSignatureHashMismatch()
...
+
+            try
+            {
+                var installDir = TestCommon.GetRandomTestDir();
+                var result = TestCommon.RunAICLICommand("install", $"AppInstallerTest.TestExeInstaller --silent -l {installDir}");
+                Assert.AreEqual(unchecked((int)0x80070002), result.ExitCode);
+                Assert.True(result.StdOut.Contains("Failed when searching source: failSearch"));
+                Assert.True(result.StdOut.Contains("AppInstallerTest.TestExeInstaller"));
+                Assert.False(result.StdOut.Contains("Successfully installed"));
+                Assert.False(VerifyTestExeInstalled(installDir, "/execustom"));
```
yao-msft, line 150: > "/execustom" [](http://example.com/codeflow?start=63&length=13)

not needed, we expect the installed file does not exist

src/AppInstallerCLICore/AppInstallerCLICore.vcxproj
```diff
@@ -226,6 +226,11 @@
       <SubSystem Condition="'$(Configuration)|$(Platform)'=='Release|x64'">Windows</SubSystem>
     </Link>
   </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(WingetDisableTestHooks)'=='true'">
```
yao-msft, line 229: >WingetDisableTestHooks [](http://example.com/codeflow?start=37&length=22)

Is this turned on in our official AppInstaller release build?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -467,7 +488,32 @@ namespace AppInstaller::Repository
...
+                        {
+                            if (result.AddFailureIfSourceNotPresent({ source, std::current_exception() }))
+                            {
+                                LOG_CAUGHT_EXCEPTION();
+                                AICLI_LOG(Repo, Warning, << "Failed to search source for correlation: " << source->GetDetails().Name);
+                            }
+                        }
+
+                        // Move failures into the single result
+                        for (SearchResult::Failure& failure : availableResult.Failures)
```
yao-msft, line 513: >availableResult.Failures [](http://example.com/codeflow?start=62&length=24)

I think this is from a single source(not composite), curious in what scenario will this have Failures entries?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -532,7 +578,21 @@ namespace AppInstaller::Repository
...
+            }
+            catch (...)
+            {
+                LOG_CAUGHT_EXCEPTION();
+                AICLI_LOG(Repo, Warning, << "Failed to search source: " << source->GetDetails().Name);
+                result.AddFailureIfSourceNotPresent({ source, std::current_exception() });
+            }
+
+            // Move failures into the single result
+            std::move(availableResult.Failures.begin(), availableResult.Failures.end(), std::back_inserter(result.Failures));
```
yao-msft, line 595: >std::move [](http://example.com/codeflow?start=12&length=9)

Should this use AddFailureIfSourceNotPresent to check duplicates?

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1139,19 +1116,45 @@ namespace AppInstaller::Repository
...
+
+                    try
+                    {
+                        openedSource = CreateSourceFromDetails(source, progress);
+                    }
+                    catch (...)
+                    {
+                        LOG_CAUGHT_EXCEPTION();
+                        AICLI_LOG(Repo, Warning, << "Failed to open source: " << source.get().Name);
+                        openExceptionProxies.emplace_back(std::make_shared<OpenExceptionProxy>(source, std::current_exception()));
```
yao-msft, line 1135: >openExceptionProxies.emplace_back(std::make_shared<OpenExceptionProxy>(source, std::current_exception())); [](http://example.com/codeflow?start=24&length=106)

nit: I think we can just directly add the proxy to the aggregatedSource directly here instead of saving to a list and add all later?

yao-msft: never mind, there's a none available check below

src/AppInstallerCLIE2ETests/InstallCommand.cs
```diff
@@ -134,6 +134,27 @@ public void InstallMSIXWithSignatureHashMismatch()
             Assert.False(VerifyTestMsixInstalledAndCleanup());
         }
 
+        [Test]
+        public void InstallExeWithAlternateSourceFailure()
+        {
+            TestCommon.RunAICLICommand("source add", "failSearch \"{ \"\"SearchHR\"\": \"\"0x80070002\"\" }\" Microsoft.Test.Configurable --header \"{}\"");
```
yao-msft, line 140: >\"\" [](http://example.com/codeflow?start=69&length=4)

I may be missing something but why double quotes for json key and value here, I thought it's regular json

src/AppInstallerRepositoryCore/CompositeSource.cpp
yao-msft: Ah, I saw this carries the OpenSource failure.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -1139,19 +1116,45 @@ namespace AppInstaller::Repository
...
+                    }
                 }
 
                 if (sourceUpdated)
                 {
                     sourceList.SaveMetadata();
                 }
 
+                // If all sources failed to open, then throw an exception that is specific to this case.
+                THROW_HR_IF(APPINSTALLER_CLI_ERROR_FAILED_TO_OPEN_ALL_SOURCES, !aggregatedSource->HasAvailableSource());
```
yao-msft, line 1150: >HasAvailableSource [](http://example.com/codeflow?start=98&length=18)

nit: there is GetAvailableSources().empty()

src/AppInstallerCLICore/Commands/ExportCommand.cpp
JohnMcPMS: I had considered that, but I'm not sure.  I think that there is probably a case to be made to have an uber search task (or two) that takes a few well defined behavior flags and roll search + handle failure + ensure* since they are very much linked.

But given our timeframe I'm going to put that on the debt rather than change it here.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
JohnMcPMS: I don't want to create a new failure bucket that is "multiple failures" yet.

src/AppInstallerCLICore/AppInstallerCLICore.vcxproj
JohnMcPMS: It will be after this is pulled in.

src/AppInstallerRepositoryCore/CompositeSource.cpp
JohnMcPMS: I had originally done that, but this pattern made it cleaner to show the suggestions on a failed `install` command.

src/AppInstallerRepositoryCore/CompositeSource.cpp
JohnMcPMS: Probably, although I put that in mostly as a guard against unbounded correlation failures.

src/AppInstallerCLIE2ETests/InstallCommand.cs
JohnMcPMS: The double double-quotes are to convince `cmd` to actually pass in a single double-quote to the process.

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: It returns a copy of the vector and I didn't need that.

yao-msft (dismissed):
:shipit:

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -94,7 +93,7 @@ namespace AppInstaller::Repository
                 details.Arg = s_Source_MSStoreDefault_Arg;
                 details.Identifier = s_Source_MSStoreDefault_Identifier;
                 details.TrustLevel = SourceTrustLevel::Trusted;
-                details.Restricted = true;
+                details.SupportCorrelation = true;
```
yao-msft, line 96: >true [](http://example.com/codeflow?start=45&length=4)

should be false here

yao-msft (approved):

PR #1473, opened 2021-09-16
Ensure available packages match with a single installed package

### LINKED ISSUE #983: Import does not work when some packages names could correspond to multiple packages.
winget import fail for some packages because of multiple entries that could match the name of the package.

# Steps to reproduce

Create a pacakge.json file with a package (Microsoft.Teams or Microsoft.Office) that has a name that match multple packages. 
```json
				{
					"PackageIdentifier": "Microsoft.Teams"
				}
			],
			"SourceDetails" : 
			{
				"Argument" : "https://winget.azureedge.net/cache",
				"Identifier" : "Microsoft.Winget.Source_8wekyb3d8bbwe",
				"Name" : "winget",
				"Type" : "Microsoft.PreIndexed.Package"
			}
```
Do a `winget import package.json`

# Expected behavior

Installation happen without error.

# Actual behavior

Logs can be found in this feedback: https://aka.ms/AAco2lu
image
error is in french but it basically just says Microsoft.Teams name match multiple packages.

# Environment

```none
[winget --info]
Windows Package Manager v0.4.11391
Windows: Windows.Desktop v10.0.19042.985
Package: Microsoft.DesktopAppInstaller v1.11.11391.0

Any other software?
```

lechacon:
When searching for installed packages matching an available package, if there were multiple matches we would create multiple results for the search. This caused commands that require a single match, like import or upgrade, to fail. Modified to select a single installed package as match if it has a strong match (product code or package family name).

This partially fixes #983 in that import and upgrade now won't fail for packages like Teams, but does not address the underlying issue of how we should deal with packages that have multiple ARP entries.

---

Also, modified dependencies code to check the experimental feature only once.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1473)

### DIFF
--- a/src/AppInstallerRepositoryCore/CompositeSource.cpp
+++ b/src/AppInstallerRepositoryCore/CompositeSource.cpp
@@ -568,27 +568,23 @@ namespace AppInstaller::Repository
                         for (auto&& crossRef : installedCrossRef.Matches)
                         {
                             AICLI_LOG(Repo, Info, << "  Checking match with package id: " << crossRef.Package->GetProperty(PackageProperty::Id));
-
-                            if (!result.ContainsInstalledPackage(crossRef.Package.get()))
+                            if (IsStrongMatchField(crossRef.MatchCriteria.Field))
                             {
-                                if (IsStrongMatchField(crossRef.MatchCriteria.Field))
+                                if (!installedPackage)
                                 {
-                                    if (!installedPackage)
-                                    {
-                                        installedPackage = std::move(crossRef.Package);
-                                    }
-                                    else
-                                    {
-                                        AICLI_LOG(Repo, Info, << "  Found multiple packages with strong match fields");
-                                        installedPackage.reset();
-                                        break;
-                                    }
+                                    installedPackage = std::move(crossRef.Package);
+                                }
+                                else
+                                {
+                                    AICLI_LOG(Repo, Info, << "  Found multiple packages with strong match fields");
+                                    installedPackage.reset();
+                                    break;
                                 }
                             }
                         }
                     }
 
-                    if (installedPackage)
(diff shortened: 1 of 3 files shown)


src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -553,14 +553,46 @@ namespace AppInstaller::Repository
...
+                                        break;
+                                    }
+                                }
+                            }
                         }
                     }
+
+                    if (installedPackage)
+                    {
+                        foundInstalledMatch = true;
```
JohnMcPMS, line 593: The test failure is due to the single result path not ever calling `ContainsInstalledPackage`.  I'm not sure that the call to it in the multi result path is correct either.  It should probably just be moved here, as the correlation itself should not be concerned with whether another source has already correlated.

With your change, one source correlating with an installed package could lead a package that had two (and thus was ignored) suddenly having only one since one could be ignored due to a prior correlation.  It should just check all options and only succeed if a single string correlation was found.

JohnMcPMS (approved):

DrCaos:
we have that problem when we want uninstall teams ;(
image

denelon:
@lechacon, does this also address the issue with uninstall also?

lechacon:
I think it should. The change was at the source search level, so it would affect anything that calls into it, including uninstall. I just tried it and uninstall worked for me.

@DrCaos , what version of winget are you using? (Output of `winget --info`)

gulshan:
Also facing the multiple teams issue. `winget` cannot upgrade teams-
image

PR #1480, opened 2021-09-17
Minor tweaks to spelling to fix issues

denelon:
I've seen a few issues mentioned by the spell checker. I'm attempting to clean some of those issues up with this PR.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1480)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -33,6 +33,7 @@ Aysnc
 azurewebsites
 badbit
 Baz
+bcp
 Beigi
 bfd
 bght
@@ -78,6 +79,7 @@ contractversion
 count'th
 countof
 countryregion
+convertto
 createmanifestmetadata
 cstdint
 ctc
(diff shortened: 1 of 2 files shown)


JohnMcPMS (approved):

.github/actions/spelling/expect.txt
```diff
@@ -33,6 +33,7 @@ Aysnc
 azurewebsites
 badbit
 Baz
+bcp
```
yao-msft, line 36: This actually fits in allow.txt since there's api we use with bcp as partial of the api name

denelon: I was seeing strange behavior in the allow file when we had bcp47. I don't think spelling likes having numbers in words. Do you think I should move this back to allow?

yao-msft (approved):

PR #1483, opened 2021-09-18
Handle single unknown version from REST source

JohnMcPMS:
## Change
If a REST source returns a single version that is `Unknown`, look up all versions when the manifest is requested.

Unfortunately I moved the code around due to the implementation and this looks like a lot more change than it is.  Things I actually changed are:
1. Fold `PackageBase` into `AvailablePackage` as there are no other derived types.
2. Made `PackageVersion` hold a `std::shared_ptr<AvailablePackage>` rather than the info.
3. Put lock around access to package versions data in `AvailablePackage`.
4. Added method and calls to update versions in this case.

The update is triggered by a `GetManifest` call that meets the criteria (version is `Unknown` and manifest is missing).  When we see that, and the version info in the package is consistent, we perform an optimized search that should return the versions and manifests.  Then we overwrite the data from the `PackageVersion` that called us and it returns the manifest.

## Validation
Manually ensured that the failing scenario is functioning now.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1483)

### DIFF
--- a/src/AppInstallerRepositoryCore/Rest/RestSource.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/RestSource.cpp
@@ -29,12 +29,160 @@ namespace AppInstaller::Repository::Rest
             std::weak_ptr<const RestSource> m_source;
         };
 
+        // The IPackage implementation for Available packages from RestSource.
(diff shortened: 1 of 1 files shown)


ashpatil-msft (approved):
:shipit:

PR #1487, opened 2021-09-20
Add option to check fields requiring verified publisher in manifest p…

yao-msft:
## Change
Extend manifest validation option to support validating fields requiring verified publishers.

## Validation
Added tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1487)

### DIFF
--- a/src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
+++ b/src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
@@ -392,7 +392,7 @@ namespace AppInstaller::Manifest
             {
                 std::vector<FieldProcessInfo> fields_v1_1 =
                 {
-                    { "Agreements", [this](const YAML::Node& value)->ValidationErrors { return ProcessAgreementsNode(value); } },
+                    { "Agreements", [this](const YAML::Node& value)->ValidationErrors { return ProcessAgreementsNode(value); }, true },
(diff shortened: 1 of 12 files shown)


src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -664,9 +671,15 @@ namespace AppInstaller::Manifest
         return resultErrors;
     }
 
-    ValidationErrors ManifestYamlPopulator::PopulateManifestInternal(const YAML::Node& rootNode, Manifest& manifest, const ManifestVer& manifestVersion, bool fullValidation)
+    ValidationErrors ManifestYamlPopulator::PopulateManifestInternal(
+        const YAML::Node& rootNode,
+        Manifest& manifest,
+        const ManifestVer& manifestVersion,
+        bool fullValidation,
+        bool errorOnVerifiedPublisherFields)
```
JohnMcPMS, line 679: For the long term it probably makes more sense to pass the `ManifestValidateOption` around rather than individual bools.  That would also mean moving it into the `Manifest` namespace rather than being in YAML (which makes sense really).

JohnMcPMS: When I wrote that I thought that it was a struct; I still stand by it being easier to update though.  You only need to add a new helper function (if that even) to call `WI_IsFlagSet` and you have a new control.

JohnMcPMS (approved):

Issue #797, opened 2021-03-15
Silent install of "winget install git.git" is not working


jarrettgsmith:
# Environment

Windows Package Manager v0.2.10191 Preview
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.18363.1440
Package: Microsoft.DesktopAppInstaller v1.11.10191.0

Logs: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

# Issue

I am trying to use `winget install` to install git remotely.  When running through a remote session the install hangs because I suspect the installer is waiting for an interaction as it has not gone silent.  When I run the silent mode on a local powershell with administrator I get the dialog pop.  How can I guarantee silent mode for all potential apps?     

I have tried...

winget isntall git.git -h
winget isntall git.git --silent
winget isntall git.git /silent
winget isntall git.git /s
winget install --id Git.Git -e -h
winget install --id Git.Git -e --silent

The dropbox link shows my winget --info as well as the install command failing to run silently and popping up the installer dialog.  I am using a powershell in Administrator mode.  Eventually I want this to work over a remote session but Im surprised to run up against this immediately. I also tried Notepad++ with the same issue.  

# Screen Capture

[https://www.dropbox.com/s/wwvl068kfvx98y1/winGetInstallSilentIssue.png?dl=0](url)

Thank you,

Jarrett

jedieaston:
Is SmartScreen enabled on the target machine? If it is disabled, the prompts appear under certain scenarios regardless of what flags are passed to winget (since it's Windows putting them up, not winget).

#703 has some discussion of this issue, but the problem should have been fixed in the last release of winget (which you have)...

jarrettgsmith:
So I checked SmartScreen settings on all systems.  They were all turned off but a few of them had a warning link to the top right of the toggle button that I clicked which popup up a security dialog asking me to confirm.  Once I did this I clearly confirmed `winget isntall git.git --silent` worked on one computer.  Then I did the same check on another computer and it didnt fix the issue.  I noticed the user interface for this section was slightly different on each computer - so I updated everything to 20H2.

So now I have all systems up to date but 1 computer still forces the popup.  Here is an image of the security toggles and the shell output.

![winget-git](https://user-images.githubusercontent.com/6508266/111244989-23c18700-85c1-11eb-936c-075cd441dfd4.png)

[...]

PR #1491, opened 2021-09-21
Remove motw before calling IAttachmentExecute::Save

yao-msft:
This fixes the "winget downloaded installers do not honor url zone settings" issue brought up in #797

## Change
IAttachmentExecute::Save() api does not clear existing Motw if the url is in trusted zone. This change puts the downloaded file in a clean state before calling the IAttachmentExecute::Save() method.

## Validation
Manually set break points and inspect Motw values are as expected and url zone settings are honored.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1491)

### DIFF
--- a/src/AppInstallerCommonCore/Downloader.cpp
+++ b/src/AppInstallerCommonCore/Downloader.cpp
@@ -284,7 +284,7 @@ namespace AppInstaller::Utility
         AICLI_LOG(Core, Info, << "Finished removing motw");
     }
 
-    HRESULT ApplyMotwUsingIAttachmentExecuteIfApplicable(const std::filesystem::path& filePath, const std::string& source)
+    HRESULT ApplyMotwUsingIAttachmentExecuteIfApplicable(const std::filesystem::path& filePath, const std::string& source, URLZONE zoneIfScanFailure)
     {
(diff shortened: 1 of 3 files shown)


src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -281,6 +302,8 @@ namespace AppInstaller::Utility
             RETURN_IF_FAILED(CoCreateInstance(CLSID_AttachmentServices, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&attachmentExecute)));
             RETURN_IF_FAILED(attachmentExecute->SetLocalPath(filePath.c_str()));
             RETURN_IF_FAILED(attachmentExecute->SetSource(Utility::ConvertToUTF16(source).c_str()));
+            // IAttachmentExecute::Save() expects the local file to be clean(i.e. it won't clear existing motw if it thinks the source url is trusted)
+            RemoveMotwIfApplicable(filePath);
             aesSaveResult = attachmentExecute->Save();
             RETURN_IF_FAILED(aesSaveResult);
```
JohnMcPMS, line 308: Maybe reapply it if this did fail?

JohnMcPMS (approved):

PR #1492, opened 2021-09-21
Fix correlation issue

JohnMcPMS:
## Change
Update the name of the `SupportCorrelation` field to better reflect the intention and correct the use to not allow available source searching when installed only semantics are being used.

Intentionally does allow available/all package searching to correlate against local packages to enable `import` to detect when packages are already present.  Also updates `import` to work better with the REST optimized search path.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1492)


yao-msft (approved):
:shipit:

PR #1494, opened 2021-09-21
Minor bug fixes

JohnMcPMS:
## Change
Use the current set of experimentals in the tests.

Fix a bug that would cause the "Done" string to be output in certain cases where the `source add` failed (requiring elevation for instance).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1494)


yao-msft (approved):
:shipit:

PR #1496, opened 2021-09-22
Switch to hardlink for retry if rename does not succeed

sreadingMSFT:

-----

User reports of rename failing due to unknown issue (thread here: https://github.com/microsoft/PowerToys/issues/13271 ). I had seen this previously and before the repro went away, switching it to a hardlink copy was successful. Using that solution here.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1496)


src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -256,8 +256,9 @@ namespace AppInstaller::CLI::Workflow
         if (retry)
         {
             std::this_thread::sleep_for(250ms);
-            // If it fails again, let that one throw
-            std::filesystem::rename(installerPath, renamedDownloadedInstaller);
+            // If retry fails, try copying with hardlink. The file at installerPath will not be cleaned up after installation,
```
yao-msft, line 259: >retry [](http://example.com/codeflow?start=18&length=5)

if rename fails?

yao-msft (approved):
:shipit:

sreadingMSFT:
Please don't merge for me. Still trying to do some more testing.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -256,8 +256,9 @@ namespace AppInstaller::CLI::Workflow
         if (retry)
         {
             std::this_thread::sleep_for(250ms);
-            // If it fails again, let that one throw
-            std::filesystem::rename(installerPath, renamedDownloadedInstaller);
+            // If retry fails, try copying with hardlink. The file at installerPath will not be cleaned up after installation,
+            // but it is in a temp folder.
+            std::filesystem::copy(installerPath, renamedDownloadedInstaller, std::filesystem::copy_options::create_hard_links);
```
JohnMcPMS, line 261: I was thinking to make it retry every 100ms for 500ms, and only then copy.  Since the file won't get cleaned up we would really prefer the rename.

sreadingMSFT: ok.

JohnMcPMS:
Sorry, I drank your milkshake 🥛

PR #1497, opened 2021-09-22
Make rename retry more frequently for longer, then try making a hardlink

JohnMcPMS:
## Change
We already had a single retry after a 250ms wait for our `rename` of the installer to make it shell actionable with the correct extension.  This change makes it try every 100ms for 500ms, then use `copy_file` with a hard link option as a fall back.

The hard link is better than a true copy, as it won't take up more space.  But it won't get cleaned up on a successful install either.

Long term we should put in some background cleanup to remove old installers that were abandoned due to failures.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1497)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
@@ -181,6 +181,61 @@ namespace AppInstaller::CLI::Workflow
 
             return args;
         }
+
+        // Complicated rename algorithm due to somewhat arbitrary failures.
+        // 1. First, try to rename.
+        // 2. Then, create an empty file for the target, and attempt to rename.
+        // 3. Then, try repeatedly for 500ms in case it is a timing thing.
+        // 4. Finally, attempt to use a hard link if available or copy if not.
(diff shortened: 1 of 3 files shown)


yao-msft (approved):
:shipit:

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -181,6 +181,61 @@ namespace AppInstaller::CLI::Workflow
...
+                return;
+            }
+            CATCH_LOG();
+
+            // 2. Then, create an empty file for the target, and attempt to rename.
+            //    This seems to fix things in certain cases, so we do it.
+            try
+            {
+                std::ofstream targetFile{ to };
+                std::filesystem::rename(from, to);
```
sreadingMSFT, line 206: >std::filesystem::rename(from, to); [](http://example.com/codeflow?start=16&length=34)

This isn't creating a file.

sreadingMSFT (approved):
:shipit:

PR #1526, opened 2021-09-29
Fix test task overrides

lechacon:
Fix test context to always use the task overrides, not only after cloning the context.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1526)

### DIFF
--- a/src/AppInstallerCLITests/WorkFlow.cpp
+++ b/src/AppInstallerCLITests/WorkFlow.cpp
@@ -306,9 +306,6 @@ namespace
             Override(wto);
         }
 
-        // For clone
-        TestContext(std::ostream& out, std::istream& in, std::shared_ptr<std::vector<WorkflowTaskOverride>> overrides) : TestContext(out, in, true, overrides) {}
-
         ~TestContext()
         {
             if (!m_isClone)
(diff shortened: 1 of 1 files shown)


yao-msft (approved):
:shipit:

src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -305,25 +307,7 @@ namespace
...
-                }
-                else
-                {
-                    itr->Used = true;
-                    itr->Override(*this);
-                    return false;
-                }
-            };
-        }
+        TestContext(std::ostream& out, std::istream& in, std::shared_ptr<std::vector<WorkflowTaskOverride>> overrides) : TestContext(out, in, true, overrides) {}
```
yao-msft, line 310: >TestContext(std::ostream& out, std::istream& in, std::shared_ptr<std::vector<WorkflowTaskOverride>> overrides) : TestContext(out, in, true, overrides) {} [](http://example.com/codeflow?start=8&length=153)

nit: this could be removed then?

PR #1528, opened 2021-09-30
Split COM install command into download and install stages

lechacon:
Based on @sreadingMSFT 's #1431.

This splits the COM install command into download and install stages as a first step for implementing concurrent downloads.
* Added check for already existing files when downloading.
* Items in the context orchestrator queue now can contain multiple commands. After executing a queue item, it is immediately requeued if it contains more commands.
* Queue items for install contain a download command and an install command. The download command executes up to downloading the file. The install command executes starting from downloading the file, which should simply see that it already exists unless it was modified or deleted.
* Changed removal of motw to prevent it from failing if motw is already not present. This can now happen because a file may go through the removal twice if it is reused.

Still pending:
* Update thread management logic to do downloads in parallel
* Improve the hash re-check to consider the installer file renaming to be able to reuse after failures

### DIFF
--- a/src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 #include "pch.h"
 #include "COMInstallCommand.h"
+#include "Workflows/DependenciesFlow.h"
 #include "Workflows/InstallFlow.h"
 #include "Workflows/WorkflowBase.h"
 
@@ -15,11 +16,17 @@ namespace AppInstaller::CLI
     // IMPORTANT: To use this command, the caller should have already retrieved the package manifest (GetManifest()) and added it to the Context Data
     void COMDownloadCommand::ExecuteInternal(Context& context) const
     {
+        // TODO: Remove duplication with InstallFlow
         context <<
             Workflow::ReportExecutionStage(ExecutionStage::Discovery) <<
             Workflow::SelectInstaller <<
             Workflow::EnsureApplicableInstaller <<
-            Workflow::DownloadPackageVersion;
+            Workflow::ReportIdentityAndInstallationDisclaimer <<
+            Workflow::ShowPackageAgreements(/* ensureAcceptance */ true) <<
+            Workflow::GetDependenciesFromInstaller <<
+            Workflow::ReportDependencies(Resource::String::InstallAndUpgradeCommandsReportDependencies) <<
+            Workflow::ReportExecutionStage(ExecutionStage::Download) <<
+            Workflow::DownloadInstaller;
     }
 
     // IMPORTANT: To use this command, the caller should have already retrieved the package manifest (GetManifest()) and added it to the Context Data
@@ -27,15 +34,5 @@ namespace AppInstaller::CLI
     {
         context <<
             Workflow::InstallPackageInstaller;
-    }
-
-    bool COMDownloadCommand::IsCommandAllowedToRunNow(std::map<std::string, UINT32>&, UINT32 runningCommandsOfCurrentType) const
-    {
-        return (runningCommandsOfCurrentType < 5);
-    }
-    
-    bool COMInstallCommand::IsCommandAllowedToRunNow(std::map<std::string, UINT32>&, UINT32 runningCommandsOfCurrentType) const
-    {
-        return (runningCommandsOfCurrentType == 0);
     }
 }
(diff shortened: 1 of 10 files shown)


src/AppInstallerCLICore/ContextOrchestrator.cpp
```diff
@@ -59,6 +59,13 @@ namespace AppInstaller::CLI::Execution
         }
     }
 
+    void ContextOrchestrator::RequeueItem(OrchestratorQueueItem& item)
+    {
+        std::lock_guard<std::mutex> lockQueue{ m_queueLock };
+
+        item.SetState(OrchestratorQueueItemState::Queued);
```
sreadingMSFT, line 66: >item.SetState(OrchestratorQueueItemState::Queued); [](http://example.com/codeflow?start=8&length=50)

Just a note here for when you do try to enable multiple downloads - one case my code didn't attempt to handle is that it doesn't create a new thread when doing a requeue when it maybe should.
Example scenario:
Simultaneous download limit is 4 and 4 apps are downloading while 2 other apps are waiting.
One of the apps finishes downloading and is requeued. Now that one of the apps is done downloading it probably should create a new thread to do another download to keep it at the 4 download limit while the current thread is installing. If the code doesn't create a new thread here then once the simultaneous download stuff comes in you can end up in a case where the queue looks like:
1. Installing, 2. Downloaded but waiting for Install spot, 3. Downloaded but waiting for Install spot, 4. Downloaded but waiting for install spot, 5. Not started, 6. Not Started.
which seems wrong.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -247,21 +247,57 @@ namespace AppInstaller::CLI::Workflow
...
+            auto existingFileHash = SHA256::ComputeHash(inStream);
+
+            if (std::equal(installer.Sha256.begin(), installer.Sha256.end(), existingFileHash.begin()))
+            {
+                AICLI_LOG(CLI, Info, << "Hash matches. Will use existing installer.");
+                context.Add<Execution::Data::HashPair>(std::make_pair(installer.Sha256, existingFileHash));
+                return;
+            }
+
+            AICLI_LOG(CLI, Info, << "Hash mismatch. Will download installer.");
```
JohnMcPMS, line 293: Remove the existing file in this case as well (best effort, aka catch).

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
```diff
@@ -13,12 +14,25 @@ using namespace AppInstaller::Utility::literals;
...
+            Workflow::ReportDependencies(Resource::String::InstallAndUpgradeCommandsReportDependencies) <<
+            Workflow::ReportExecutionStage(ExecutionStage::Download) <<
+            Workflow::DownloadInstaller;
+    }
+
+    // IMPORTANT: To use this command, the caller should have already executed the COMDownloadCommand
+    void COMInstallCommand::ExecuteInternal(Context& context) const
+    {
+        context <<
+            Workflow::InstallPackageInstaller;
```
sreadingMSFT, line 36: >InstallPackageInstaller [](http://example.com/codeflow?start=22&length=23)

I see that this still tries to do a download in cases where the file no longer exists or the hash is mismatched. I think that will probably not be the best idea once simultaneous downloads comes in since part of the goal of that feature is to not have installs need to wait for other downloads to finish.

src/AppInstallerCLICore/ContextOrchestrator.cpp
lechacon: Thanks! I hadn't thought of that case. I'll keep it in mind.

sreadingMSFT: Ah actually I got it wrong. You can't end up with that case above. The worst you can get is three apps downloading while one installs instead of four downloading while one installs, since the threads that see that the install is blocked don't just sit there waiting they'd move on to trying to do a download for the other item. Depending on the download limit that may be ok to just sometimes be doing one less than the max number of downloads since the system will still be busy doing an install (which i think for some installers also involves some downloading). But yeah, something to add a comment about probably on which behavior gets chosen.

src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
JohnMcPMS: I agree that in this case we don't want to perform the download on the second command; just verify the file and proceed to install.

PR #1529, opened 2021-09-30
Fix unit tests in pipeline

JohnMcPMS:
## Change
This fixes:
- REST tests that were deadlocking due to not setting the response body in the test helpers.
- The `-logto` unit test argument to now properly be the actual log path again.
- The test script to fail on timeout rather than simply ending.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1529)


yao-msft (approved):

JohnMcPMS:
Confirmed that the unit tests are now showing in the results and the log file is no longer empty.

PR #1533, opened 2021-09-30
Change format to long for installer success codes

ryfu-msft:
Resolves an issue where the installer success codes for some manifests were larger than the range of an int datatype, which can cause issues for winget-create's model as it is expecting a valid int. 

 https://github.com/microsoft/winget-create/issues/155

Changes the integer format to long in the v1.0 and v1.1 schema to increase the range of values.

Verification:
- Passes all unit tests 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1533)


yao-msft:
Please update the singleton schemas too for consistency

yao-msft (approved):

PR #1534, opened 2021-10-01
Settings streams exchange semantics

### LINKED ISSUE #463: Settings stream access is not synchronized cross process

The base settings stream set/get functions are not synchronized cross process (nor in process either, but since things are largely single threaded, not a big deal).  This probably does not apply to the app setting container streams, but the file streams created by std::fstream and kin are not exclusive.

The first step would be to simply prevent corrupt data via use of a CrossProcessReaderWriteLock when the fstream is open.

The next step would be to change the setting stream functionality to be more object oriented, allowing for read then write functionality while holding the exclusive lock.

# Steps to reproduce

Attempting to manage sources from multiple processes could lead to corrupted data.  This includes running winget concurrently, triggering multiple automatic updates that

JohnMcPMS:
Fixes #463 

## Change
The goal of this change is to add exchange semantics to settings stream writes.  Currently they do not have any interprocess synchronization.  This does not lead to corrupted streams as the underlying systems prevent simultaneous writes, but it can lead to data loss if updates from one process are not merged with those from another.

Because holding any kind of lock would likely lead to contention amongst the processes, we use exchange semantics instead.  Writes are only permitted if the stream has not changed since it was last read.  This required moving to a more object oriented model for the settings streams, which then required updating the consumers of it to also maintain an object and respond to failed write attempts.

The update to `AdminSettings` was fairly straightforward, requiring just the necessary interface updates and the exchange retry loop.

For `SourceListInternal`, I also refactored the code out of `RepositorySource.cpp` which was getting quite large.  I moved the group policy related functions to their own file (`SourcePolicy.cpp`) and the source enumeration related functions to theirs (`SourceList.cpp`).  The renamed `SourceList` type now maintains the stream objects for its lifetime and manages the necessary exchange retry attempts for the various writes that are required.

## Validation
New unit tests are added to validate the code under controlled circumstances.  Functional tests would be impractical to write as the actual interactions are races between various writes, although the configurable test source that was added recently would aid in this area.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1534)


src/AppInstallerCommonCore/Settings.cpp
```diff
@@ -128,66 +111,143 @@ namespace AppInstaller::Settings
...
+                m_hash.reset();
+            }
+
+            std::filesystem::path PathTo() override
+            {
+                return m_container->PathTo();
+            }
+
+        protected:
+            static bool AreHashesEqual(const SHA256::HashBuffer& a, const SHA256::HashBuffer& b)
```
yao-msft, line 200: >static bool AreHashesEqual(const SHA256::HashBuffer& a, const SHA256::HashBuffer& b) [](http://example.com/codeflow?start=12&length=84)

nit: this could be in SHA256.h so other code can use this as well

src/AppInstallerCommonCore/Settings.cpp
```diff
@@ -128,66 +111,143 @@ namespace AppInstaller::Settings
...
+                THROW_HR_IF(E_UNEXPECTED, value.size() > std::numeric_limits<uint32_t>::max());
+
+                // If Set is called without ever reading the value, then we can assume that caller wants
+                // to overwrite it regardless. Also, we don't have any previous value to compare against
+                // anyway so the only other option would be to always reject it.
+                if (m_hash)
                 {
-                    result /= name.parent_path();
+                    std::optional<SHA256::HashBuffer> currentHash;
+                    std::ignore = GetInternal(currentHash);
```
yao-msft, line 167: >GetInternal(currentHash) [](http://example.com/codeflow?start=34&length=24)

I thought we don't like out parameter? Maybe it's ok for internal implementation

src/AppInstallerRepositoryCore/SourcePolicy.h
```diff
@@ -0,0 +1,26 @@
...
+    // that already existed when the Group Policy came into effect.
+    Settings::TogglePolicy::Policy GetPolicyBlockingUserSource(std::string_view name, std::string_view type, std::string_view arg, bool isTombstone);
+
+    // Helper that converts the result of GetPolicyBlockingUserSource into a bool.
+    bool IsUserSourceAllowedByPolicy(std::string_view name, std::string_view type, std::string_view arg, bool isTombstone);
+
+    // Determines if a well known source is enabled; if onlyExplicit is true, it must be explicitly enabled by group policy.
+    bool IsWellKnownSourceEnabled(WellKnownSource source, bool onlyExplicit = false);
+
+    // Checks that the specified source is removable per policy; throwing if it is not.
```
yao-msft, line 24: >throwing if it is not. [](http://example.com/codeflow?start=65&length=22)

looks like this is sort of a helper function, should we return bool instead of throwing?

yao-msft (approved):
:shipit:

src/AppInstallerCommonCore/Settings.cpp
JohnMcPMS: It has its places, and I think that in this internal case it works well to enable the side effect to be controlled without adding type overhead.

src/AppInstallerRepositoryCore/SourcePolicy.h
JohnMcPMS: Certainly we should change the function rather than adding a catch at some point in the future, but for now this will flow up to the top level catch as intended.  Anyone calling this function will know that if the preconditions are met the workflow will be ending.

PR #1551, opened 2021-10-05
Added vsconfig file to automatically install required workloads.

jedieaston:

-----

Something I noticed when trying to see if I could contribute anywhere in the project was that it's kind of hard to get all of the dependencies running in Visual Studio (at least, if you're like me and try to avoid VS like the plague). 

This PR adds a .vsconfig file to the root of the repository, which causes Visual Studio 2019 to automatically prompt you when opening the repo to install any missing workloads (and automatically tells you which ones you need). I tried to keep it to the minimum required, but that's still quite a few things (.NET Desktop Development, C++ Development, UWP, the Spectre Mitigation  libraries, Windows SDK 18362 and 19041, etc). You can't list extensions (so the Visual Studio Installer Projects extension still has to be installed manually), but this makes it a heck of a lot easier than going back to the installer every time the project won't build for something else.

I tested this in a VM and a sandbox, both worked great (i.e. the project built). 

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1551)


JohnMcPMS (approved):
I would probably have just excluded the .vsconfig from spell checking since it does not contain actual words for the most part.  But thanks for the addition 👍

PR #1563, opened 2021-10-08
Icelandic localization tool description wording

GudniNatan:
Changed two words in the icelandic localization for the toolDescription string:
- The word "of" should be "og" (the word of is like "too much" which makes no sense) and, 
- "úr" should be "frá" (The difference is in grammar, úr is like "out of", whereas frá is simply from. You are not installing out of the command line)

-----


JohnMcPMS:
Thank you for the feedback; unfortunately, we have had to cut localization down to a smaller set of languages (see #1761).

Issue #1640, opened 2021-10-25
Upgrade when installer type changes


denelon:
![image](https://user-images.githubusercontent.com/77048997/138540316-1d5ceded-f42a-4541-a13c-c0ae8f34574b.png)

![image](https://user-images.githubusercontent.com/77048997/138540326-e3e75ca8-a88f-4f92-b102-a63c04aae3b0.png)

_Originally posted by @MagicFirefly in https://github.com/microsoft/winget-cli/issues/1191#issuecomment-950047548_

Trenly:
I experienced this issue today with Microsoft Edge. Log is attached
```raw
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

Try the new cross-platform PowerShell https://aka.ms/pscore6

PS C:\WINDOWS\system32> winget upgrade
Name                              Id               Version          Available        Source
-------------------------------------------------------------------------------------------
Microsoft Edge                    Microsoft.Edge   95.0.1020.40     95.0.1020.44     winget
Microsoft 365 Apps for enterprise Microsoft.Office 16.0.14326.20508 16.0.14527.20234 winget
2 upgrades available.
PS C:\WINDOWS\system32> winget upgrade Microsoft.Edge
No applicable update found.
PS C:\WINDOWS\system32> winget upgrade --id Microsoft.Edge
No applicable update found.
PS C:\WINDOWS\system32> winget show Microsoft.Edge
Found Microsoft Edge [Microsoft.Edge]
Version: 95.0.1020.44
Publisher: Microsoft Corporation
Publisher Url: https://www.microsoft.com/en-US/
Publisher Support Url: https://support.microsoft.com/en-US/
Author: Microsoft Corporation
Moniker: msedge
Description: World-class performance with more privacy, more productivity, and more value while you browse.
Homepage: https://www.microsoft.com/en-us/edge
License: MIT License
License Url: https://www.microsoft.com/en-us/servicesagreement/
Privacy Url: https://privacy.microsoft.com/en-US/privacystatement
Copyright: Copyright (C) Microsoft Corporation
Copyright Url: https://www.microsoft.com/en-us/servicesagreement/
Installer:
  Type: Msi
  Download Url: https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/93862b4a-15ae-40a0-b501-f969633661e2/MicrosoftEdgeEnterpriseX64.msi
[...]

Issue #1191, opened 2021-06-19
Winget upgrade package, No applicable update found


kdpuvvadi:
# Brief description of your issue

 if i run `winget upgraded packagename` & update available, it should update the package but it shows No applicable update found even though it's showing the newer version available. 

# Steps to reproduce

`winget upgrade ` list of the available updates
`winget upgrade --all` or `winget upgrade Python.Python.3` 

# Expected behavior

It should update the actual selected package or all incase of `--all`

# Actual behavior

throws `No applicable update found` and if i run `winget install packagename` then installs the update. 

# Environment

Windows 10 Enterprise 21H1 build 19043.1055
Experience Pack 120.2212.2020.0

```none
[winget --info]
Windows Package Manager v1.0.11451
Windows: Windows.Desktop v10.0.19043.1055
Package: Microsoft.DesktopAppInstaller v1.11.11451.0
```

KaranKad:
This issue might be related to [winget-pkgs](https://github.com/microsoft/winget-pkgs)
@kdpuvvadi Did you first install install Firefox from it's website using exe installer or from winget?

I was able to replicate this issue by installing old version of Firefox using `exe` installer and then trying `winget upgrade Mozilla.Firefox`
```
❯ winget upgrade
Name                              Id                      Version      Available   Source
-----------------------------------------------------------------------------------------
Mozilla Firefox                   Mozilla.Firefox         89.0         89.0.1      winget
Winamp                            Radionomy.Winamp        5.8          5.8         winget
Microsoft Windows Desktop Runtime Microsoft.dotnetRuntime 5.0.6.30021  5.0.7.30113 winget
TechPowerUp GPU-Z                 TechPowerUp.GPU-Z       Unknown      2.40.0      winget
Microsoft Windows Desktop Runtime Microsoft.dotnetRuntime 3.1.15.30014 5.0.7.30113 winget

❯ winget upgrade Mozilla.Firefox
No applicable update found.
```
WinGet's Firefox manifest only contains msi installer so that might be the issue.

denelon:
@KaranKad thanks for figuring this one out. We do try to avoid updating a package with a different "InstallerType".

OfficialEsco:
winget CLI need to explain this instead of `No applicable update found.`. A better reporting would most likely help solve most of these issues since we could narrow it down.

denelon:
https://github.com/microsoft/winget-cli/issues/1196

kdpuvvadi:
[...]

PR #1649, opened 2021-10-27
Print message when install technology is different in upgrade scenarios

msftrubengu:

It is possible that the winget shows an available update via `winget upgrade` but doing `winget upgrade <id>` results in "No applicable update found." The details of why already shown in the logs.

Based on #1191, one of the common patterns we've seen is when the installed technology of the installed version is different to the update version. This change bubbles up the error to make it more visible to users. 

The current recommendation is to uninstall and install the new version. In the future, #1640 will provide a way to do it automatically if desired.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1649)


src/AppInstallerCLICore/Workflows/ManifestComparator.h
```diff
@@ -62,10 +62,10 @@ namespace AppInstaller::CLI::Workflow
         };
     }
 
-    struct InstallerAndInapplicability
+    struct InstallerAndInapplicabilities
     {
         std::optional<Manifest::ManifestInstaller> installer;
-        InapplicabilityFlags inapplicabilityFlags;
+        std::vector<InapplicabilityFlags> inapplicabilitiesInstaller;
```
JohnMcPMS, line 68: It probably doesn't matter right now, but it seems odd to not know which installer each goes with.

JohnMcPMS (approved):

PR #1666, opened 2021-11-02
Added an option to select architecture to install.

### LINKED ISSUE #906: Specify installer architecture
# Add an argument and settings to specify installer architecture

As a user I would like to be able to specify which architecture an installer should use on my machine.

This could be an argument like: `winget install foo.bar --architecture x86`

There could also be settings to specify a "preferred" architecture or a "required" architecture.

Given a "required" architecture, the search results could be trimmed to include only packages with manifests supporting my "required" architecture.

jedieaston:

-----
Related to #906.

This PR adds a option, `--architecture`, to `winget install`. This option allows the user to (you guessed it) tell winget what architecture they want to install. If the architecture isn't available, winget does the same thing it would do if there wasn't a installer entry for the user's platform (No applicable installer error). The user is only allowed to provide architectures their system can run (if they ask for arm64 on x64, a `CommandException` is thrown). 

This may have implications for upgrades (I'm not 100% sure on whether upgrade checks the architecture of the currently installed version. If it doesn't, winget might upgrade a 32-bit version to a 64-bit version even if the user didn't want that). If it breaks too much stuff I can close this.

Tested: manually. If I need to write unit tests I can. 

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1666)

### DIFF
--- a/src/AppInstallerCLICore/Commands/InstallCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/InstallCommand.cpp
@@ -17,6 +17,7 @@ namespace AppInstaller::CLI
     namespace
     {
         constexpr Utility::LocIndView s_ArgumentName_Scope = "scope"_liv;
+        constexpr Utility::LocIndView s_ArgumentName_Architecture = "arch"_liv;
     }
 
     std::vector<Argument> InstallCommand::GetArguments() const
@@ -31,6 +32,7 @@ namespace AppInstaller::CLI
             Argument::ForType(Args::Type::Channel),
             Argument::ForType(Args::Type::Source),
             Argument{ s_ArgumentName_Scope, Argument::NoAlias, Args::Type::InstallScope, Resource::String::InstallScopeDescription, ArgumentType::Standard, Argument::Visibility::Help },
+            Argument{ s_ArgumentName_Architecture, Argument::NoAlias, Args::Type::InstallArchitecture, Resource::String::InstallArchitectureArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help },
             Argument::ForType(Args::Type::Exact),
             Argument::ForType(Args::Type::Interactive),
             Argument::ForType(Args::Type::Silent),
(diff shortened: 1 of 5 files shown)


denelon:
Issue #906 was intended to include adding this to settings so a user wouldn't necessarily need to specify on the command line. I wouldn't close that issue until we have a setting (preference and requirement) for architecture. We also may have some pain if we run into a situation where we have an x64 package that requires an x86 dependency and the user has a setting to require only x64 architecture.

Masamune3210:
Absolutely love the original title

jedieaston:
> Issue #906 was intended to include adding this to settings so a user wouldn't necessarily need to specify on the command line.

My memory was dim, I thought that issue was purely for adding the argument. Unlinked, sorry.

> Absolutely love the original title

TIL that GitHub uses the branch name to name a PR if you didn't specify a title 😄

JohnMcPMS:
> This may have implications for upgrades (I'm not 100% sure on whether upgrade checks the architecture of the currently installed version. If it doesn't, winget might upgrade a 32-bit version to a 64-bit version even if the user didn't want that).

I wanted upgrade to take that into account, but is difficult to determine the true architecture of an installed package based on ARP data alone.  This is actually something we should store in the tracking catalog that I've recently added; the preference/requirement and the architecture itself.  That way we can at try to honor the original request during upgrade.  I don't expect you to do that as part of this change though.

@denelon, are you ok with taking just the command line argument, or do you want the settings implemented as well so as to have a more complete feature?

denelon:
I think this is a great enhancement al by itself. I'd suggest accepting the feature as implemented, and we will keep the feature open until we build the settings and integrate that with the other work being done.

@jedieaston you rock! This is a fantastic enhancement, and I'm sure will be very well received. I've noticed the ask for 64bit adobe in another manifest. I love being able to eliminate the x64 and x86 variants of the "same" package. We still need to think about what this means in terms of an upgrade when the preferred architecture isn't available in the latest version, but that is a solvable problem.

JohnMcPMS (approved):

PR #1667, opened 2021-11-02
Add Package Dependencies to index. 

hackean-msft:
This pull request creates a new dependencies table  in the index an provides an interface via winget util to allow the addition, update and validation of manifest dependencies to the index. 

A new API is exposed to allow the verification of the dependency graph structure when a delete operation is performed for a specific manifest. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1667)


src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/Interface.h
```diff
@@ -36,6 +36,10 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_0
         // Version 1.2
         Utility::NormalizedName NormalizeName(std::string_view name, std::string_view publisher) const override;
 
+        // Validate the dependencies of the given manifest.
+        bool ValidateManifest(SQLite::Connection& connection, const Manifest::Manifest& manifest) const override;
```
JohnMcPMS, line 40: I don't like this design of putting the validation inside the ISQLiteIndex.  It does not fit with the rest of the design; the index is a data store and the only reason to put the validation code inside would be that the data was potentially going to change so much that it was no longer convertible.

It also means that this is less flexible.  For instance, if we wanted to keep this table and use it during `install`, we would have to have new methods to actually extract the dependency information.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_3/Interface_1_3.cpp
```diff
@@ -44,6 +46,8 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_3
             V1_0::ManifestTable::UpdateValueIdById<HashVirtualTable>(connection, manifestId, manifest.StreamSha256);
         }
 
+        DependenciesTable::AddDependencies(connection, manifest, manifestId);
```
JohnMcPMS, line 49: You can't add this new table to the existing schema.  You have to create 1.4

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/SQLiteStatementBuilder.cpp
```diff
@@ -267,8 +267,8 @@ namespace AppInstaller::Repository::SQLite::Builder
 
     StatementBuilder& StatementBuilder::Select(details::rowcount_t)
     {
-        m_stream << "SELECT COUNT(*)";
-        return *this;
+m_stream << "SELECT COUNT(*)";
```
JohnMcPMS, line 270: ?

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <Microsoft/Schema/1_4/DependenciesTable.h>
+
+namespace AppInstaller::Repository
+{
+	using namespace Microsoft::Schema::V1_4;
+
+	bool PackageDependenciesValidation::ValidateManifestDependencies(const std::string& indexPath, const Manifest::Manifest& manifest)
+	{
+		SQLite::Connection connection(SQLite::Connection::Create(indexPath, SQLite::Connection::OpenDisposition::ReadOnly, SQLite::Connection::OpenFlags::None));
+		return DependenciesTable::ValidateDependencies(connection, manifest);
```
JohnMcPMS, line 18: This is not at all how this should be done.  You are reaching directly into a SQLite database to do this validation; one that might not even have your tables present.  It must be done through the SQLiteIndex interface.  As I suggested in a previous comment, you need to expose two functions: one to get the dependency information for a given manifest, and one to find all manifests that are dependent on a given id.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
```diff
@@ -146,6 +146,8 @@ namespace AppInstaller::Repository::Microsoft
         // Largely a utility function; should not be used to do work on behalf of the index by the caller.
         Utility::NormalizedName NormalizeName(std::string_view name, std::string_view publisher) const;
 
+        bool ValidateManifest(SQLite::Connection& connection, const Manifest::Manifest& manifest);
```
JohnMcPMS, line 149: See previous request for changes and new comment; expose the information through this interface, don't do the validation inside it.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+#include "Microsoft/Schema/1_0/OneToOneTable.h"
+#include "Microsoft/Schema/1_0/IdTable.h"
+#include "Microsoft/Schema/1_0/ManifestTable.h"
+#include "Microsoft/Schema/1_0/VersionTable.h"
+#include "Microsoft/Schema/1_0/Interface.h"
+#include "Microsoft/Schema/1_0/ChannelTable.h"
+
+namespace AppInstaller::Repository::Microsoft::Schema::V1_4
+{
+	using namespace AppInstaller;
```
yao-msft, line 16: >	 [](http://example.com/codeflow?start=0&length=1)

nit: please replace tabs with 4 spaces in this file, maybe you'd want to set the VS editor settings to do that for you so you won't need to worry about it in the future

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
```diff
@@ -146,6 +146,12 @@ namespace AppInstaller::Repository::Microsoft
         // Largely a utility function; should not be used to do work on behalf of the index by the caller.
         Utility::NormalizedName NormalizeName(std::string_view name, std::string_view publisher) const;
 
+        // Get all the dependencies for a specific manifest.
+        std::map<Manifest::Dependency, SQLite::rowid_t> GetDependenciesByManifestRowId(SQLite::rowid_t manifestRowId) const;
```
JohnMcPMS, line 150:
```suggestion
        std::vector<std::pair<IdType, Utility::Version>> GetDependenciesByManifestId(IdType manifestId) const;
```
We shouldn't build the dependency internally; that can be done with the existing interface.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
```diff
@@ -146,6 +146,12 @@ namespace AppInstaller::Repository::Microsoft
         // Largely a utility function; should not be used to do work on behalf of the index by the caller.
         Utility::NormalizedName NormalizeName(std::string_view name, std::string_view publisher) const;
 
+        // Get all the dependencies for a specific manifest.
+        std::map<Manifest::Dependency, SQLite::rowid_t> GetDependenciesByManifestRowId(SQLite::rowid_t manifestRowId) const;
+        std::vector<std::pair<Manifest::Manifest, Utility::Version>> GetDependenciesByPackageId(AppInstaller::Manifest::string_t packageId) const;
```
JohnMcPMS, line 151:
```suggestion
        std::vector<std::pair<IdType, Utility::Version>> GetDependentsById(IdType id) const;
```
Where the pair is `{ Manifest ID, Version Required }`.  Really "highest version required" would be sufficient, but for reporting purposes we want to me able to call out the actual dependent.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
```diff
@@ -146,6 +146,12 @@ namespace AppInstaller::Repository::Microsoft
         // Largely a utility function; should not be used to do work on behalf of the index by the caller.
         Utility::NormalizedName NormalizeName(std::string_view name, std::string_view publisher) const;
 
+        // Get all the dependencies for a specific manifest.
+        std::map<Manifest::Dependency, SQLite::rowid_t> GetDependenciesByManifestRowId(SQLite::rowid_t manifestRowId) const;
+        std::vector<std::pair<Manifest::Manifest, Utility::Version>> GetDependenciesByPackageId(AppInstaller::Manifest::string_t packageId) const;
+
+        // Get the latest version of a package.
+        std::optional<std::pair<SQLite::rowid_t, Utility::Version>> GetPackageLatestVersion(Manifest::string_t packageId, std::set<Utility::Version> versionsToExclude = {}) const;
```
JohnMcPMS, line 154: This is not needed.  A combination of `GetVersionKeysById` and `GetManifestIdByKey` already meet this need.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+			result.emplace(select.GetColumn<std::string>(1), select.GetColumn<SQLite::rowid_t>(0));
+		}
+
+		return result;
+	}
+
+	void DependenciesTable::Create(SQLite::Connection& connection)
+	{
+		using namespace SQLite::Builder;
+		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "createDependencyTable_v1_3");
```
JohnMcPMS, line 187:
```suggestion
		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "createDependencyTable_v1_4");
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+		std::vector<DependenciesTableColumnInfo> dependenciesColumns =
+		{ 
+			{ s_DependenciesTable_Manifest_Column_Name },
+			{ s_DependenciesTable_MinVersion_Column_Name },
+			{ s_DependenciesTable_PackageId_Column_Name }
+		};
+
+		for (const DependenciesTableColumnInfo& value : dependenciesColumns)
+		{
+			createTableBuilder.Column(ColumnBuilder(value.Name, Type::Int64).NotNull());
```
JohnMcPMS, line 202:
```suggestion
			createTableBuilder.Column(ColumnBuilder(value.Name, SQLite::rowid_t).NotNull());
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+		}
+
+		createTableBuilder.EndColumns();
+
+		createTableBuilder.Execute(connection);
+
+		StatementBuilder createPKIndexBuilder;
```
JohnMcPMS, line 209: The actual primary key index would be on manifest id and package id both and would be unique.
The index we need for add and potentially for client use is on manifest id.
The index needed for remove is on package id.

I would probably add all three so that it is easier to decide not to drop them in the future (although would still require a minor version update).

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+		createPKIndexBuilder.CreateIndex(s_DependenciesTable_Index_Name).On(s_DependenciesTable_Table_Name).Columns({ s_DependenciesTable_Manifest_Column_Name });
+		createPKIndexBuilder.Execute(connection);
+
+		savepoint.Commit();
+	}
+
+
+	void DependenciesTable::AddDependencies(SQLite::Connection& connection, const Manifest::Manifest& manifest, SQLite::rowid_t manifestRowId)
+	{
+		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "add_dependencies");
```
JohnMcPMS, line 219:
```suggestion
		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "add_dependencies_v1_4");
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+		}
+
+		InsertManifestDependencies(connection, manifestRowId, dependencies);
+
+		savepoint.Commit();
+	}
+
+	bool DependenciesTable::UpdateDependencies(SQLite::Connection& connection, const Manifest::Manifest& manifest, SQLite::rowid_t manifestRowId)
+	{
+		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "update_dependencies");
```
JohnMcPMS, line 234:
```suggestion
		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "update_dependencies_v1_4");
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+		const auto dependencies = GetDependencies(manifest, Manifest::DependencyType::Package);
+		const std::set<Manifest::Dependency> dependenciesSet(dependencies.begin(), dependencies.end());
+		if (!dependencies.size())
```
JohnMcPMS, line 238: What if there were dependencies but they were removed?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+		savepoint.Commit();
+	}
+
+	bool DependenciesTable::UpdateDependencies(SQLite::Connection& connection, const Manifest::Manifest& manifest, SQLite::rowid_t manifestRowId)
+	{
+		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "update_dependencies");
+
+		const auto dependencies = GetDependencies(manifest, Manifest::DependencyType::Package);
+		const std::set<Manifest::Dependency> dependenciesSet(dependencies.begin(), dependencies.end());
```
JohnMcPMS, line 237: Just return a `set` from `GetDependencies`?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+		InsertManifestDependencies(connection, manifestRowId, toAddDependencies);
+		RemoveDependenciesByRowIds(connection, toRemoveDependencies);
+		savepoint.Commit();
+
+		return true;
+	}
+
+	void DependenciesTable::RemoveDependencies(SQLite::Connection& connection, SQLite::rowid_t manifestRowId)
+	{
+		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "remove_dependencies_by_manifest");
```
JohnMcPMS, line 278:
```suggestion
		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "remove_dependencies_by_manifest_v1_4");
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+				select.GetColumn<SQLite::rowid_t>(0));
+		}
+
+		return resultSet;
+
+	}
+
+	void DependenciesTable::PrepareForPackaging(SQLite::Connection& connection)
+	{
+		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "prepareForPacking_V1_3");
```
JohnMcPMS, line 372:
```suggestion
		SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "prepareForPacking_V1_4");
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+			return dependencies;
+		}
+
+		void RemoveDependenciesByRowIds(SQLite::Connection& connection, std::vector<SQLite::rowid_t> dependencyRowId)
+		{
+			using namespace SQLite::Builder;
+			SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "remove_dependencies_by_rowid");
+
+			if (dependencyRowId.size() > 0)
```
JohnMcPMS, line 54: If you are going to check the size, just check for `empty` and get out up front rather than creating a savepoint.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+				[&](Manifest::Dependency dep) { minVersions.emplace_back(dep.MinVersion.value().ToString()); }
+			);
+
+			std::for_each(
+				dependencies.begin(),
+				dependencies.end(),
+				[&](Manifest::Dependency dep) { packageIds.emplace_back(dep.Id); }
+			);
+
+			ManifestTable::ExistsById(connection, manifestRowId);
```
JohnMcPMS, line 106: ?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+				notFoundVersion.end(),
+				[&](Utility::NormalizedString& version) { versionsResultSet[version] = VersionTable::EnsureExists(connection, version); });
+
+			StatementBuilder insertBuilder;
+			insertBuilder.InsertInto(s_DependenciesTable_Table_Name)
+				.Columns({ s_DependenciesTable_Manifest_Column_Name, s_DependenciesTable_MinVersion_Column_Name, s_DependenciesTable_PackageId_Column_Name })
+				.Values(Unbound, Unbound, Unbound);
+			SQLite::Statement insert = insertBuilder.Prepare(connection);
+
+			// TODO: use insert with multiple values, rather than run the query in a loop.
```
JohnMcPMS, line 139: Honestly I find this function much harder to read than if it were a simple loop to handle each one individually.  Why does it need to do all of each step?  How many dependencies do you expect to be inserting at one time that the performance gains outweigh the maintainability cost of this structure?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,384 @@
...
+			}
+		}
+	}
+
+	std::string_view DependenciesTable::TableName()
+	{
+		return s_DependenciesTable_Table_Name;
+	}
+
+	std::map<std::string, SQLite::rowid_t> SelectIdsByValues(const SQLite::Connection& connection, std::string_view tableName, std::string_view columnName, const std::vector<Utility::NormalizedString>& values)
```
JohnMcPMS, line 157: You ask tables for their values rather than reaching in.  Remove this in favor of calling the existing methods of getting values.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,192 @@
...
+		using namespace Manifest;
+
+		Dependency rootId(DependencyType::Package, manifest.Id, manifest.Version);
+		std::vector<ValidationError> dependenciesError;
+		bool foundErrors = false;
+
+		DependencyGraph graph(rootId, [&](const Dependency& node) {
+
+			DependencyList depList;
+			if (node.Id == rootId.Id)
```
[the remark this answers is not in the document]
hackean-msft: The special case is needed, it can be handled in the lambda function or outside(before the dependencies graph initialization),  it just makes sense to have the logic to handle dependencies evaluation in single place irrespective of whether it is root or child dependency .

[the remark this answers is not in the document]
JohnMcPMS: Because it is only checking for `Id` equality, a loop in the graph could get back to this special case.  That is itself very strange and I'm not even sure what we should do in that case at the moment.  If you are adding the highest version, then it is correct.  If you are not, then we have a strange situation.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
hackean-msft: The actual dependent is the manifest returned by GetDependents as different manifests within the same package might not have dependencies on the current package,  which is why the manifest is returned for reporting sake. 

Also, there are two versions here, the manifest version is useful for reporting on the service and the min version associated with dependency relationship between the manifest and the package. Can you clarify what highest version required mean in this case?

JohnMcPMS: By that I mean that the only thing you really need to verify is that the highest version specified as a dependency will still be satisfied after a delete.  So you only need to know the highest version.

It is also interesting to know the set of manifests that would be impacted by the removal though, so returning the manifest id and the version of the dependency that it specifies allows you to determine all impacted manifests instead of just "is the set of impacted manifests empty?".

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
hackean-msft: see line 223 - 233 and 236

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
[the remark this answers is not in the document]
hackean-msft: The dependency graph never process a node twice(so loops are not an issue) so the lambda is never called to the same node. Also, we're adding the highest version.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
hackean-msft: The set of manifest that will be impacted is addressed in 215 - 232.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -184,7 +184,7 @@ bool HasExtension(std::string_view extension) const;
 
         bool operator <(const Dependency& rhs) const
         {
-            return Id < rhs.Id;
+            return FoldCase(Id) < FoldCase(rhs.Id);
```
JohnMcPMS, line 187: >FoldCase [](http://example.com/codeflow?start=19&length=8)

This is a somewhat expensive function on an operator that needs to be used a lot for sorted containers.  Please have:
```
private:
std::string m_foldedId;
```
And use it in both operators to improve the performance.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
JohnMcPMS: Why is this a `map`?
Please use `IdType`.
Why are you returning a second row id?  What is it?  The comment should indicate what the ids are that you are returning.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
JohnMcPMS: Give this it's own comment that describes what the rowid is.
The input should be `IdType id` which is the package id's rowid.
I realize that it makes it harder given the way the DependencyGraph works now, but that should be fixed, not this made to conform.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        {
+            using namespace SQLite::Builder;
+            if (dependencyRowIds.empty())
+            {
+                return;
+            }
+            SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "remove_dependencies_by_rowid");
+
+            SQLite::Builder::StatementBuilder builder;
+            builder.DeleteFrom(s_DependenciesTable_Table_Name).Where(SQLite::RowIDName).In(dependencyRowIds.size());
```
JohnMcPMS, line 59: >builder.DeleteFrom(s_DependenciesTable_Table_Name).Where(SQLite::RowIDName).In(dependencyRowIds.size()); [](http://example.com/codeflow?start=12&length=104)

I'm not sure that you are optimizing this like you think you are.  No need to change it now, but I think you did extra work over just creating a statement and executing it N times that probably doesn't help.

Even if it is faster it feels like premature optimization.  No blame here, the struggle is real, but we still need to fight against it.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        void ThrowOnMissingPackageNodes(std::vector<Manifest::Dependency> missingPackageNodes)
+        {
+            if (!missingPackageNodes.empty())
+            {
+                std::string missingPackages{ missingPackageNodes.begin()->Id };
+                std::for_each(
+                    missingPackageNodes.begin() + 1,
+                    missingPackageNodes.end(),
+                    [&](auto& dep) { missingPackages.append(", " + dep.Id);  });
+                THROW_HR_MSG(APPINSTALLER_CLI_ERROR_MISSING_PACKAGE, "Missing packages: %s", missingPackages.c_str());
```
JohnMcPMS, line 83: >%s [](http://example.com/codeflow?start=88&length=2)

`%hs`, these are tricky because internally they use a wide string format function.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        {
+            using namespace SQLite::Builder;
+            using namespace Schema::V1_0;
+
+            std::vector<Manifest::Dependency> missingPackageNodes;
+            std::map<Utility::Version, SQLite::rowid_t> versionsMap;
+            std::map<Manifest::Manifest::string_t, SQLite::rowid_t> idsMap;
+            for (const auto& dep : dependencies)
+            {
+                auto depMinVersion = dep.MinVersion.value();
```
JohnMcPMS, line 100: >dep.MinVersion.value(); [](http://example.com/codeflow?start=37&length=23)

Throws if no MinVersion, which is legitimate.  You don't have any tests for that?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+                    }
+                );
+
+                if (result == manifestVersions.end())
+                {
+                    VersionTable::EnsureExists(connection, depMinVersion.ToString());
+                }
+
+                idsMap[packageId] = packageRowId.value();
+                versionsMap[depMinVersion] = VersionTable::SelectIdByValue(connection, depMinVersion.ToString()).value();
```
JohnMcPMS, line 127: >SelectIdByValue [](http://example.com/codeflow?start=59&length=15)

`EnsureExists` does everything for you, including returning the rowid of the version string.  Line 110 through 124 can be removed and this should just become `VersionTable::EnsureExists(connection, depMinVersion.ToString())`

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        void InsertManifestDependencies(
+            SQLite::Connection& connection,
+            SQLite::rowid_t manifestRowId,
+            std::set<Manifest::Dependency>& dependencies)
+        {
+            using namespace SQLite::Builder;
+            using namespace Schema::V1_0;
+
+            std::vector<Manifest::Dependency> missingPackageNodes;
+            std::map<Utility::Version, SQLite::rowid_t> versionsMap;
```
JohnMcPMS, line 96: >std::map<Utility::Version, SQLite::rowid_t> versionsMap; [](http://example.com/codeflow?start=12&length=56)

Don't use maps, use a single vector containing the information you are going to insert (both ids).

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+                insert.Bind(2, versionsMap[dep.MinVersion.value().ToString()]);
+                insert.Bind(3, idsMap[dep.Id]);
+
+                insert.Execute(true);
+            }
+        }
+
+        std::optional<std::pair<SQLite::rowid_t, Utility::NormalizedString>> GetPackageRowIdAndVersion(SQLite::Connection& connection, const Manifest::Dependency& dependency)
+        {
+            std::optional<std::pair<SQLite::rowid_t, Utility::NormalizedString>> row;
```
JohnMcPMS, line 151: >std::optional<std::pair<SQLite::rowid_t, Utility::NormalizedString>> row; [](http://example.com/codeflow?start=12&length=73)

Hint on `optional`, you can just return the value type when you have something and `{}` when you want it to be empty.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        std::vector<DependenciesTableColumnInfo> dependenciesColumns =
+        {
+            { s_DependenciesTable_Manifest_Column_Name },
+            { s_DependenciesTable_MinVersion_Column_Name },
+            { s_DependenciesTable_PackageId_Column_Name }
+        };
+
+        for (const DependenciesTableColumnInfo& value : dependenciesColumns)
+        {
+            createTableBuilder.Column(ColumnBuilder(value.Name, Type::RowId).NotNull());
```
JohnMcPMS, line 187: >NotNull [](http://example.com/codeflow?start=77&length=7)

MinVersion is optional.  This could be modeled by an empty string for the version, but it isn't clear that you are doing so.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        createTableBuilder.EndColumns();
+
+        createTableBuilder.Execute(connection);
+
+        StatementBuilder createPKIndexBuilder;
+        createPKIndexBuilder.CreateUniqueIndex(s_DependenciesTable_Index_Name).On(s_DependenciesTable_Table_Name).Columns({ s_DependenciesTable_Manifest_Column_Name, s_DependenciesTable_MinVersion_Column_Name, s_DependenciesTable_PackageId_Column_Name });
```
JohnMcPMS, line 195: >s_DependenciesTable_MinVersion_Column_Name [](http://example.com/codeflow?start=166&length=42)

The unique index should not include the version.  And the other indexes I suggested will help a lot for performance.  Is there a reason not to add them?

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+        // Get dependencies to add.
+        std::set<Manifest::Dependency> toAddDependencies;
+        std::copy_if(
+            dependencies.begin(),
+            dependencies.end(),
+            std::inserter(toAddDependencies, toAddDependencies.begin()),
+            [&](Manifest::Dependency dep)
+            {
+                auto depRow = GetPackageRowIdAndVersion(connection, dep);
```
JohnMcPMS, line 232: >auto depRow = GetPackageRowIdAndVersion(connection, dep); [](http://example.com/codeflow?start=16&length=57)

Why not just convert to this in `GetDependencies`?

JohnMcPMS: It would also mean that you have only one location that has to deal with package ids that don't exist.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+            {
+                if (dependenciesPackageIdAndVersion.find(row.first) == dependenciesPackageIdAndVersion.end())
+                {
+                    toRemoveDependencies.emplace_back(row.second);
+                }
+            }
+        );
+
+        InsertManifestDependencies(connection, manifestRowId, toAddDependencies);
+        RemoveDependenciesByRowIds(connection, toRemoveDependencies);
```
JohnMcPMS, line 272: >RemoveDependenciesByRowIds [](http://example.com/codeflow?start=8&length=26)

Given that the PK should be { manifest, package id }, you could just as easily remove by that instead of forcing the use of the rowid here.  Then you wouldn't need the public API to return the unnecessary dependency rowid.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        return true;
+    }
+
+    void DependenciesTable::RemoveDependencies(SQLite::Connection& connection, SQLite::rowid_t manifestRowId)
+    {
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, std::string{ s_DependenciesTable_Table_Name } + "remove_dependencies_by_manifest_v1_4");
+        SQLite::Builder::StatementBuilder builder;
+        builder.DeleteFrom(s_DependenciesTable_Table_Name).Where(s_DependenciesTable_Manifest_Column_Name).Equals(manifestRowId);
+
+        builder.Prepare(connection).Execute(true);
```
JohnMcPMS, line 284: >builder.Prepare(connection).Execute(true); [](http://example.com/codeflow?start=8&length=42)

```suggestion
builder.Execute(connection);
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,375 @@
...
+        savepoint.Commit();
```
JohnMcPMS, line 285: >Commit [](http://example.com/codeflow?start=18&length=6)

Since you are also the potential owner of the version ids, you need to collect them and remove if they are no longer needed.

This also brings up the point that you need to update `OneToOneTableDeleteIfNotNeededById` to handle the fact that you are now another potential reference point.

I would suggest changing the code there slightly to allow a more generic call out to each of the potential owners to check for reference.  That way dependencies can simply ignore any non-version calls as no reference.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/Interface.h
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+#pragma once
+#include "Microsoft/Schema/ISQLiteIndex.h"
+#include "Microsoft/Schema/1_3/Interface.h"
+#include "Microsoft/Schema/1_2/Interface.h"
+
+namespace AppInstaller::Repository::Microsoft::Schema::V1_4
+{
+    // Interface to this schema version exposed through ISQLiteIndex.
+    struct Interface : public V1_3::Interface
```
JohnMcPMS, line 11: >Interface [](http://example.com/codeflow?start=11&length=9)

You also need to implement `CheckConsistency` here and call into the same for the dependencies table.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+    namespace
+    {
+        std::vector<AppInstaller::Manifest::Dependency> GetDependencies(
+            const Manifest::Manifest& manifest, AppInstaller::Manifest::DependencyType dependencyType)
+        {
+            std::vector<AppInstaller::Manifest::Dependency>  dependencies;
+
+            for (const auto& installer : manifest.Installers)
+            {
+                installer.Dependencies.ApplyToAll([&](AppInstaller::Manifest::Dependency dependency)
```
JohnMcPMS, line 28: >ApplyToAll [](http://example.com/codeflow?start=39&length=10)

There is already an `ApplyToType` I believe.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+            }
+
+            return dependencies;
+        }
+
+        std::optional<std::pair<SQLite::rowid_t, Utility::Version>> GetPackageLatestVersion(
+            SQLiteIndex* index, Manifest::string_t packageId, std::set<Utility::Version> exclusions = {})
+        {
+            SearchRequest request;
+            request.Filters.emplace_back(PackageMatchField::Id, MatchType::Exact, packageId);
```
JohnMcPMS, line 44: >Exact [](http://example.com/codeflow?start=75&length=5)

`CaseInsensitive` unless you know that every caller is getting the string from the index to begin with.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+            for (auto& v : vac)
+            {
+                auto currentVersion = v.GetVersion();
+                if (exclusions.find(currentVersion) != exclusions.end())
+                {
+                    continue;
+                }
+
+                if (maxVersion.IsUnknown() || currentVersion > maxVersion)
```
JohnMcPMS, line 68: >if (maxVersion.IsUnknown() || currentVersion > maxVersion) [](http://example.com/codeflow?start=16&length=58)

1. Unkown always compares as lower.
2. The version come back in sorted order, highest first, already.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+                {
+                    maxVersion = currentVersion;
+                }
+            }
+
+            if (!maxVersion.IsUnknown())
+            {
+                auto manifestRowId = index->GetManifestIdByKey(packageRowId, maxVersion.ToString(), "");
```
JohnMcPMS, line 76: >"" [](http://example.com/codeflow?start=100&length=2)

Should really take channel into account, although doing so is likely very complicated, both in the code and conceptually.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+        Dependency rootId(DependencyType::Package, manifest.Id, manifest.Version);
+        std::vector<ValidationError> dependenciesError;
+        bool foundErrors = false;
+
+        DependencyGraph graph(rootId, [&](const Dependency& node) {
+
+            DependencyList depList;
+            if (node.Id == rootId.Id)
+            {
+                const auto dependencies = GetDependencies(manifest, DependencyType::Package);
```
JohnMcPMS, line 97: >GetDependencies [](http://example.com/codeflow?start=42&length=15)

Why can't you just `return GetDependencies` in this block?

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+                error.append(" ").append(node.Id);
+                dependenciesError.emplace_back(ValidationError(error));
+                foundErrors = true;
+                return depList;
+            }
+
+            auto packageLatestDependencies = index->GetDependenciesByManifestRowId(packageLatest.value().first);
+            std::for_each(
+                packageLatestDependencies.begin(),
+                packageLatestDependencies.end(),
```
JohnMcPMS, line 129: range-for is much easier to read.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+            return true;
+        }
+
+        std::vector<std::pair<Manifest::Manifest, Utility::Version>> manifestToVersionPair;
+        std::for_each(
+            dependentsSet.begin(),
+            dependentsSet.end(),
+            [&](std::pair<SQLite::rowid_t, Utility::Version> current)
+            {
+                Manifest::Manifest manifest;
```
JohnMcPMS, line 174: >Manifest::Manifest manifest; [](http://example.com/codeflow?start=16&length=28)

Shadowing a parameter will only lead to pain.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+    {
+        auto dependentsSet = index->GetDependentsById(manifest.Id);
+
+        if (!dependentsSet.size())
+        {
+            // all good this manifest is not a dependency of any manifest.
+            return true;
+        }
+
+        std::vector<std::pair<Manifest::Manifest, Utility::Version>> manifestToVersionPair;
```
JohnMcPMS, line 168: >Manifest::Manifest [](http://example.com/codeflow?start=30&length=18)

Why are you using a huge Manifest struct when all you care about is Id and Version?  And the Version in the manifest is the same as the version in the pair.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+            THROW_HR(APPINSTALLER_CLI_ERROR_MISSING_PACKAGE);
+        }
+
+        if (Utility::Version(manifest.Version) < packageLatest.value().second)
+        {
+            // all good, since it's min version the criteria is still satisfied.
+            return true;
+        }
+
+        auto nextLatestAfterDelete = GetPackageLatestVersion(index, manifest.Id, { packageLatest.value().second });
```
JohnMcPMS, line 196: >GetPackageLatestVersion [](http://example.com/codeflow?start=37&length=23)

It seems like it would be easier for this to just return all of the manifests and their versions.  Or at least between 0 and 2 of them.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
JohnMcPMS: Actually, the way you are doing it now that doesn't take channel into account is going to have bugs if we ever do implement channel.  If you instead assume that they are sorted you will be good-ish.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
JohnMcPMS: And it should probably filter out anything that isn't empty channel for now.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+        if (!nextLatestAfterDelete.has_value())
+        {
+            auto itrStart = manifestToVersionPair.begin();
```
JohnMcPMS, line 200: >auto itrStart = manifestToVersionPair.begin(); [](http://example.com/codeflow?start=12&length=46)

Don't copy this code, have a function that handles throwing the exception from the list of affected manifests.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -0,0 +1,253 @@
...
+                    APPINSTALLER_CLI_ERROR_DEPENDENCIES_VALIDATION_FAILED));
+
+        }
+
+        std::vector<std::pair<Manifest::Manifest, Utility::Version>> breakingManifests;
+
+        std::copy_if(
+            manifestToVersionPair.begin(),
+            manifestToVersionPair.end(),
+            std::back_inserter(breakingManifests),
```
JohnMcPMS, line 224: range-for is much easier to read.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
hackean-msft: No, they are not, the manifest struct presented an alternative to tuples or pairs, also I can't stress how much I don't like tuples because of their ugly, nonmaintainable and unreadable nature. It was a decision between creating a new struct for the sake of data transfer or using the manifest object.  Went with the latter, now switched to an internal struct for data transfer.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
hackean-msft: I've had no contention so far with using collections related methods for any sort of data aggregation for elegance and to take advantage of whatever performance gain is associated with the underlying framework, I don't see why it should be any different in this case. Unless there is some sort of unwritten rule about avoiding them in this codebase, because I don't see any difference in readability between this and what is suggested, besides it's plastered almost everywhere in the dependencies implementation.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
hackean-msft: Same as previous comment.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
hackean-msft: The intent for GetDependencies it to get the dependencies in a manifest and not do any form of database interaction for clarity sake. Retrieving dependencies from manifest and retrieving from db seems to be like 2 distinct operations. I see the potential for tight coupling in what is being suggested here.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
hackean-msft: Can you explain:
**_I would suggest changing the code there slightly to allow a more generic call out to each of the potential owners to check for reference. That way dependencies can simply ignore any non-version calls as no reference._**

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
hackean-msft: GetPackageLatest approach is much consistent with the install flow implementation and it's clearer that the latest version is being selected by just reading the code for manitainability sake. Getting a list of manifest might seem "easier"(I don't see how), it just isn't consistent with the install flow implementation. Btw, GetPackageLatest has been implemented to handle channels.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+        std::vector<DependenciesTableColumnInfo> notNullableDependenciesColumns =
+        {
+            { s_DependenciesTable_Manifest_Column_Name },
+            { s_DependenciesTable_MinVersion_Column_Name },
+        };
+
+        std::vector<DependenciesTableColumnInfo> nullableDependenciesColumns =
+        {
+            { s_DependenciesTable_PackageId_Column_Name }
```
JohnMcPMS, line 184: Looks like you swapped package id (which cannot be null) and min version (which can) in these.

General note on `vector`: since it requires allocations it isn't great for a fixed set of known values contained within a specific scope.  Either a `std::array` or implicit `initializer_list` in range-for would be more efficient.

```
for (auto& thing : { a, b, c })
    ...
```

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
JohnMcPMS: This is the lowest layer of the implementation where a manifest and database are involved and a function that is local to this file.  What I'm suggesting does not change the coupling of the data itself.  While it does indeed create coupling in the function, that is perfectly fine when it is contained to this file.

If there are other use cases to get just the dependencies from a manifest, you can always create a layered approach; where `GetAndLinkDependencies` uses `GetDependencies` to extract the set from the manifest and then links them with the database (again, creating only a single location that needs to deal with package ids that don't exist).  If there is no reason to need both right now, that minor refactoring could be left for when it is needed.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
JohnMcPMS: In the function [`OneToOneTableDeleteIfNotNeededById`](https://github.com/microsoft/winget-cli/blob/2e693852fe7fae7f1643fdb799653645cdf0034f/src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/OneToOneTable.cpp#L145), the only thing that is checked right now is whether the manifest table references the given value or not.

You will also need a similar "does the dependencies table reference the version value" check in that function.  But rather than adding that directly, I'm suggesting a change to that function to call new method like `ManifestTable::IsValueReferenced` and then also call `DependenciesTable::IsValueReferenced` so that they can do the check themselves.  Dependencies would not even check the database unless the value was a version or package id.  That sets a better pattern for extension than putting special logic to handle the dependencies directly into the other function.

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
JohnMcPMS: There is nothing wrong with using `algorithm`, but it likely generates the exact same code.  It is just that the way you have it written:
```
std::for_each(
    container.begin(),
    container.end(),
    [&] (const auto& value)
    {
        <body>
    });
```
Is several more lines and one indentation deeper, containing a bunch of boilerplate that just adds noise.  The same thing with range-for is one line to indicate the loop:
```
for (const auto& value : container)
{
    <body>
}
```
I'm not going to reject your change solely on use of `for_each`, but I'm very comfortable with my stance on readability difference of those two. 😄

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -170,35 +166,56 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_4
...
 
+        // Primary key index by package rowid and manifest rowid.
         StatementBuilder createPKIndexBuilder;
-        createPKIndexBuilder.CreateUniqueIndex(s_DependenciesTable_Index_Name).On(s_DependenciesTable_Table_Name).Columns({ s_DependenciesTable_Manifest_Column_Name, s_DependenciesTable_MinVersion_Column_Name, s_DependenciesTable_PackageId_Column_Name });
+        createPKIndexBuilder.CreateUniqueIndex(s_DependenciesTable_Index_Name).On(s_DependenciesTable_Table_Name).Columns({ s_DependenciesTable_Manifest_Column_Name, s_DependenciesTable_PackageId_Column_Name });
         createPKIndexBuilder.Execute(connection);
 
+        // Index of dependency by Manifest id. 
+        StatementBuilder createIndexByManifestIdBuilder;
+        createIndexByManifestIdBuilder.CreateIndex(dependencyIndexByManifestId).On(s_DependenciesTable_Table_Name).Columns({ s_DependenciesTable_Manifest_Column_Name });
```
JohnMcPMS, line 208: Since the manifest is the first column on the primary key index, this one isn't helping any.  But you probably do want one on the version to make the "is referenced" function faster.

src/AppInstallerRepositoryCore/SQLiteStatementBuilder.h
```diff
@@ -216,6 +216,7 @@ namespace AppInstaller::Repository::SQLite::Builder
         // Begin a filter clause on the given column.
         StatementBuilder& Where(std::string_view column);
         StatementBuilder& Where(const QualifiedColumn& column);
+        StatementBuilder& Or(const QualifiedColumn& column);
```
JohnMcPMS, line 219: Probably better next to the `And` (unless I'm misremembering and there is no `And`).

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/Microsoft/Schema/1_3/Interface_1_3.cpp
```diff
@@ -87,4 +87,9 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_3
             return V1_2::Interface::GetPropertyByManifestIdInternal(connection, manifestId, property);
         }
     }
+
+    bool Interface::CheckConsistency(const SQLite::Connection& connection, bool log) const
+    {
+        return V1_2::Interface::CheckConsistency(connection, log);
```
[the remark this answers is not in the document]
hackean-msft: If there is a need to add a CheckConsistency in v1_3 interface in the future, I don't believe the change will take effect if there is a direct call from V1_4 to V1_2.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -137,7 +137,12 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_4
             {
                 insert.Reset();
                 insert.Bind(1, dep.m_manifestRowId);
-                insert.Bind(2, dep.m_versionRowId.has_value() ? std::to_string(dep.m_versionRowId.value()) : "");
+
+                if (dep.m_versionRowId.has_value())
```
JohnMcPMS, line 141: I'm not exactly sure on what the bind state is after a `Reset`, but I think it doesn't change.  You want to `Bind(2, nullptr)` to set the value to null.

Also note that you have to use `GetColumnIsNull` when reading the rows to check for a DB null value before you can `GetColumn<std::string>`.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/Interface_1_4.cpp
```diff
@@ -66,6 +67,18 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_4
         savepoint.Commit();
     }
 
+    bool Interface::NotNeeded(const SQLite::Connection& connection, std::string_view tableName, std::string_view valueName, SQLite::rowid_t id) const
+    {
+        bool result = V1_0::Interface::NotNeeded(connection, tableName, valueName, id);
+
+        if (tableName == Schema::V1_0::VersionTable::TableName())
```
JohnMcPMS, line 74: I would put the filtering responsibility on the `DependenciesTable` as it is the one to know what it references.  In addition, it knows the name of the column; copying the string out to `Interface` is not good.

src/AppInstallerRepositoryCore/Microsoft/Schema/ISQLiteIndex.h
```diff
@@ -108,6 +108,8 @@ namespace AppInstaller::Repository::Microsoft::Schema
         virtual std::set<std::pair<SQLite::rowid_t, Utility::NormalizedString>> GetDependenciesByManifestRowId(const SQLite::Connection& connection, SQLite::rowid_t manifestRowId) const = 0;
 
         virtual std::vector<std::pair<SQLite::rowid_t, Utility::NormalizedString>> GetDependentsById(const SQLite::Connection& connection, AppInstaller::Manifest::string_t packageId) const = 0;
+
+        virtual bool NotNeeded(const SQLite::Connection& connection, std::string_view tableName, std::string_view valueName, SQLite::rowid_t id) const = 0;
```
JohnMcPMS, line 112: This doesn't need to be public, so it can be a `protected` function on the base class rather than on the interface.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/Interface_1_4.cpp
```diff
@@ -66,6 +67,18 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_4
         savepoint.Commit();
     }
 
+    bool Interface::NotNeeded(const SQLite::Connection& connection, std::string_view tableName, std::string_view valueName, SQLite::rowid_t id) const
```
JohnMcPMS, line 70: I can't put a comment on `RemoveManifestById`, but it isn't removing the versions that are no longer needed.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/Microsoft/Schema/1_4/DependenciesTable.cpp
```diff
@@ -423,32 +423,61 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_4
...
+
+        std::array<std::string_view, 1> columns = { s_DependenciesTable_MinVersion_Column_Name };
+        bool referenced = false;
 
-        auto result = std::find(dependenciesColumns.begin(), dependenciesColumns.end(), valueName);
-        if (result == std::end(dependenciesColumns))
+        for(auto column: columns)
         {
-            THROW_HR(APPINSTALLER_CLI_ERROR_INVALID_TABLE_COLUMN);
+            builder.Select(SQLite::RowIDName).From(s_DependenciesTable_Table_Name).Where(column).Equals(Unbound).Limit(1);
```
JohnMcPMS, line 440: nit: You can pass `valueRowId` instead of `Unbound` to achieve the same thing here.  Not need to change it, just a note in case you weren't aware.

JohnMcPMS (approved):

PR #1671, opened 2021-11-03
Use package tracking data to correlate in CompositeSource

JohnMcPMS:
## Change
This change leverages the tracking data that is stored per source to correlate installed packages to available packages.  This will currently provide no additional benefit with the `winget` source, as we aren't doing the work yet to determine the best fit system artifact when none of our current correlation attempts succeed.  It will help the `msstore` source to correlate as that is currently lacking in support to search system references.

## Validation
Existing regression tests and new tests added for the tracking catalog.  Modified the tracking catalog to enable the existing test infrastructure to inject custom data.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1671)

### DIFF
--- a/src/AppInstallerRepositoryCore/CompositeSource.cpp
+++ b/src/AppInstallerRepositoryCore/CompositeSource.cpp
@@ -30,6 +30,138 @@ namespace AppInstaller::Repository
             return false;
         }
 
+        // Move returns if there is only one package in the matches that is strong; otherwise returns an empty value.
+        std::shared_ptr<IPackage> FindOnlyStrongMatchFieldResult(std::vector<ResultMatch>& matches)
+        {
+            std::shared_ptr<IPackage> result;
+
+            for (auto&& match : matches)
+            {
+                AICLI_LOG(Repo, Info, << "  Checking match with package id: " << match.Package->GetProperty(PackageProperty::Id));
+
+                if (IsStrongMatchField(match.MatchCriteria.Field))
+                {
+                    if (!result)
+                    {
+                        result = std::move(match.Package);
+                    }
+                    else
+                    {
+                        AICLI_LOG(Repo, Info, << "  Found multiple packages with strong match fields");
+                        result.reset();
+                        break;
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        // Gets a single matching package from the results
+        template <typename MultipleIntro, typename Indeterminate>
+        std::shared_ptr<IPackage> GetMatchingPackage(std::vector<ResultMatch>& matches, MultipleIntro&& multipleIntro, Indeterminate&& indeterminate)
+        {
+            if (matches.empty())
+            {
+                return {};
+            }
+            else if (matches.size() == 1)
+            {
+                return std::move(matches[0].Package);
+            }
+            else
+            {
+                multipleIntro();
+
+                auto result = FindOnlyStrongMatchFieldResult(matches);
+
+                if (!result)
+                {
+                    indeterminate();
+                }
+
+                return result;
+            }
+        }
+
+        // For a given package from a tracking catalog, get the latest write time.
+        // Look at all versions rather than just the latest to account for the potential of downgrading.
+        std::chrono::system_clock::time_point GetLatestTrackingPackageWriteTime(const std::shared_ptr<IPackage>& trackingPackage)
+        {
+            std::chrono::system_clock::time_point result{};
+
+            for (const auto& key : trackingPackage->GetAvailableVersionKeys())
+            {
+                auto version = trackingPackage->GetAvailableVersion(key);
+                if (version)
+                {
+                    auto metadata = version->GetMetadata();
+                    auto itr = metadata.find(PackageVersionMetadata::TrackingWriteTime);
+                    if (itr != metadata.end())
+                    {
+                        std::int64_t unixEpoch = 0;
+                        try
+                        {
+                            unixEpoch = std::stoll(itr->second);
+                        }
+                        CATCH_LOG();
+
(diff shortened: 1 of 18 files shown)


src/AppInstallerRepositoryCore/CompositeSource.h
```diff
@@ -57,7 +52,7 @@ namespace AppInstaller::Repository
         SearchResult SearchAvailable(const SearchRequest& request) const;
 
         std::shared_ptr<ISource> m_installedSource;
```
yao-msft, line 54: >std::shared_ptr<ISource> [](http://example.com/codeflow?start=8&length=24)

Should this one be updated to use Source too since you are updating the available ones already?

src/AppInstallerRepositoryCore/PackageTrackingCatalog.cpp
```diff
@@ -27,6 +29,108 @@
...
+            {
+                return m_details.Identifier;
+            }
+
+            SourceInformation GetInformation() override
+            {
+                return {};
+            }
+
+            // Set custom header. Returns false if custom header is not supported.
```
yao-msft, line 52: >// Set custom header. Returns false if custom header is not supported. [](http://example.com/codeflow?start=12&length=70)

nit: this override is not needed unless you wanted it to be explicit

src/AppInstallerRepositoryCore/PackageTrackingCatalog.cpp
```diff
@@ -27,6 +29,108 @@
...
+            {
+                return m_details;
+            }
+
+            std::string GetIdentifier() override
+            {
+                return m_details.Identifier;
+            }
+
+            SourceInformation GetInformation() override
```
yao-msft, line 47: >SourceInformation GetInformation() override [](http://example.com/codeflow?start=12&length=43)

nit: this override is not needed either

src/AppInstallerRepositoryCore/Public/winget/PackageTrackingCatalog.h
```diff
@@ -1,30 +1,37 @@
...
         PackageTrackingCatalog& operator=(PackageTrackingCatalog&&) noexcept;
         ~PackageTrackingCatalog();
 
-        // Creates or opens the tracking catalog for the given source.
-        // TODO: Make creation exclusive to the refactored Source type.
-        static PackageTrackingCatalog CreateForSource(const Source& source);
-
-        // Removes the package tracking catalog for a given source.
+        // Removes the package tracking catalog for a given source identifier.
         static void RemoveForSource(const std::string& identifier);
```
yao-msft, line 30: >RemoveForSource [](http://example.com/codeflow?start=20&length=15)

I saw this is only used in unit tests now, will there be usage in the product code in the future?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -30,6 +30,138 @@ namespace AppInstaller::Repository
             return false;
         }
 
+        // Move returns if there is only one package in the matches that is strong; otherwise returns an empty value.
```
yao-msft, line 33: >Move returns [](http://example.com/codeflow?start=11&length=12)

nit: Returns?

yao-msft: never mind..

src/AppInstallerRepositoryCore/Public/winget/PackageTrackingCatalog.h
JohnMcPMS: Not sure, but since I wanted to use it in the tests I implemented it here.  We could call it as part of removing a source, but I wasn't sure if that was the right thing to do.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -574,31 +829,61 @@ namespace AppInstaller::Repository
...
+                                "] in source [" << source.GetIdentifier() << "] when searching for [" << systemReferenceSearch.ToString() << "]");
+                        }, [&] {
+                            AICLI_LOG(Repo, Warning, << "  Appropriate installed package could not be determined");
+                        });
+
+                    if (installedPackage && !result.ContainsInstalledPackage(installedPackage.get()))
+                    {
+                        auto compositePackage = std::make_shared<CompositePackage>(
+                            std::move(installedPackage),
+                            GetTrackedPackageFromAvailableSource(result, source, match.Package->GetProperty(PackageProperty::Id)));
```
yao-msft, line 870: >GetTrackedPackageFromAvailableSource [](http://example.com/codeflow?start=28&length=36)

This might return null, in that case, we prevent the later available package to be associated with this installed package, I guess it'll be fixed when we have support for 1 -> n installed to available mapping in the future

yao-msft (approved):
:shipit:

src/AppInstallerRepositoryCore/CompositeSource.cpp
yao-msft: Actually, this might cause some upgrade cases to report "no applicable upgrades found" incorrectly. Looks like we may want to fix the 1 -> n mapping sooner.

JohnMcPMS: We need to think about the behavior in that case.  We haven't really laid down how the association will work.  In my opinion, we should not attempt to upgrade in the case that the original source no longer contains the package, but another one does.  We should instead inform the user of that fact so that they can decide what to do.  If they did install/upgrade into the other source, then the association would be updated and we would start upgrading from the new location in the future.

PR #1674, opened 2021-11-04
Prevent circular dependency in global telemetry logger creation

yao-msft:
Currently there is a potential circular dependency between the telemetry object and the logger object due to the need to open settings to finish telemetry construction.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1674)

### DIFF
--- a/src/WinGetUtil/Exports.cpp
+++ b/src/WinGetUtil/Exports.cpp
@@ -10,6 +10,7 @@
 #include <AppInstallerTelemetry.h>
 #include <Microsoft/SQLiteIndex.h>
 #include <winget/ManifestYamlParser.h>
+#include <winget/ThreadGlobals.h>
 
 using namespace AppInstaller::Utility;
 using namespace AppInstaller::Manifest;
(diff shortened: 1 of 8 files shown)


JohnMcPMS (changes requested):
This does not solve the problem, which is telemetry > settings > failure > telemetry.  The only solution that I see to that is disabling telemetry until settings are loaded in order to be able to respect the potential for the setting that disables telemetry.

src/WinGetUtil/Exports.cpp
```diff
@@ -21,8 +22,13 @@ extern "C"
     {
         THROW_HR_IF(E_INVALIDARG, !logPath);
 
-        static std::once_flag s_initLogging;
-        std::call_once(s_initLogging, []() {
+        thread_local AppInstaller::ThreadLocalStorage::ThreadGlobals threadGlobals;
+        thread_local std::once_flag initLogging;
+
+        std::call_once(initLogging, []() {
+            std::unique_ptr<AppInstaller::ThreadLocalStorage::PreviousThreadGlobals> previous = threadGlobals.SetForCurrentThread();
```
JohnMcPMS, line 29: While this may solve the problem when the code stays on one thread, it may make other problems that we are not aware of.  I would rather we fix it completely by creating a set of APIs that give more explicit control over the thread globals so that callers can actually control what is happening.

JohnMcPMS (approved):

PR #1677, opened 2021-11-04
Add product code and PFN to enable existing searches

JohnMcPMS:
## Change
Add `ProductCode` and `PackageFamilyName` values to enum to enable the existing search capabilities through COM.

Also updates the sample caller to have the `packageQuery` capability as I found that it did not during some manual testing.

## Validation
Manual update to sample app to try it out.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1677)

### DIFF
--- a/tools/SampleWinGetUWPCaller/AppInstallerCaller/Package.appxmanifest
+++ b/tools/SampleWinGetUWPCaller/AppInstallerCaller/Package.appxmanifest
@@ -26,6 +26,7 @@
   </Applications>
   <Capabilities>
     <Capability Name="internetClient" />
+    <rescap:Capability Name="packageQuery" />
     <rescap:Capability Name="packageManagement" />
   </Capabilities>
 </Package>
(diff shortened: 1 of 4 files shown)


tools/SampleWinGetUWPCaller/AppInstallerCaller/Package.appxmanifest
```diff
@@ -26,6 +26,7 @@
   </Applications>
   <Capabilities>
     <Capability Name="internetClient" />
+    <rescap:Capability Name="packageQuery" />
```
yao-msft, line 29: >packageQuery [](http://example.com/codeflow?start=29&length=12)

packageManagement should already include packageQuery in semantics, did you find places that only work with packageQuery but not packageManagement? If yes, we may need to fix those.

yao-msft (approved):
:shipit:

tools/SampleWinGetUWPCaller/AppInstallerCaller/Package.appxmanifest
JohnMcPMS: Ok, it looks like the code is doing that and I was just seeing the failure to directly find `packageQuery` while debugging.  I can revert this and update the code to not log a failure here.

PR #1683, opened 2021-11-08
Make the text for the Microsoft Store disclaimer yellow

### LINKED ISSUE #1682: Make the text for the Microsoft Store disclaimer yellow.

There's been a little confusion (https://github.com/microsoft/winget-pkgs/issues/34313, but I've seen it other places too) about the difference between the msstore source and the winget source from a user perspective. 

I think part of this is because the msstore source spews 30 lines to stdout which nobody slows down to read. However, the second line, `This package is provided through Microsoft Store. winget may need to acquire the package from Microsoft Store on behalf of the current user.` is really important for users when they want to know where to ask questions.

I think this text should be made yellow so that it pops out to the user. (If we have agreement about this, I can make the PR. Literally a one line, five second change.)

Change `context.Reporter.Info()` to `context.Reporter.Warn()` [here](https://github.com/microsoft/winget-cli/blob/467adfa34ede4cc7291022a28eef3b9a5d8c69d9/src/AppInstallerCLICore/Workflows/InstallFlow.cpp#L126).

vedantmgoyal2009:

* Try to resolve #1682

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1683)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
@@ -123,7 +123,7 @@ namespace AppInstaller::CLI::Workflow
 
         if (installerType == InstallerTypeEnum::MSStore)
         {
-            context.Reporter.Info() << Resource::String::InstallationDisclaimerMSStore << std::endl;
+            context.Reporter.Warn() << Resource::String::InstallationDisclaimerMSStore << std::endl;
         }
         else
         {


jedieaston:
Based on the conversation in that issue, this is not the correct solution. We need to bold this text.

yao-msft:
I feel this should be an emphasized Info() instead of a Warn() semantically as well.

Issue #1242, opened 2021-07-01
Complex installers


denelon:
# Description of the new feature/enhancement

As we get closer to supporting .zip and standalone .exe installers, we need some additional mechanisms to identify a few different scenarios. These include what kind of an installer is inside of .zip files. It also includes scenarios where a package is distributed with a .exe installer, but it also contains an .msi installer (PowerToys). 

# Proposed technical implementation details (optional)

We would add a new node to the Installers called "NestedInstallerType".

This example would help fix the upgrade scenario related to PowerToys:
```YAML
Installers:
  - InstallerType: exe
    NestedInstallerType: msi
    InstallerArchitecture: x64
    InstallerSHA256: 12341234123412341234123412341234123412341234
    InstallerURL: https://www.contoso.com/awesomeapp/1.0.0/awesomeapp.x64.exe
```

This example is for a nullsoft installer distributed inside of a .zip file, but this could also be .exe, inno, wix, etc.
```YAML
Installers:
  - InstallerType: zip
    NestedInstallerType: nullsoft
    InstallerArchitecture: x64
    InstallerSHA256: 12341234123412341234123412341234123412341234
    InstallerURL: https://www.contoso.com/awesomeapp/1.0.0/awesomeapp.x64.zip
```

This example is for a standalone / portable executable (these are not yet supported) inside of a .zip file.

```YAML
Installers:
  - InstallerType: zip
    NestedInstallerType: standalone
    InstallerArchitecture: x64
    InstallerSHA256: 12341234123412341234123412341234123412341234
    InstallerURL: https://www.contoso.com/awesomeapp/1.0.0/awesomeapp.x64.zip
```

Issue #182, opened 2020-05-19
Support for installation of portable/standalone apps


KiruyaMomochi:
# Description of the new feature/enhancement

There are programs like [starship](https://github.com/starship/starship/) or [aria2](https://github.com/aria2/aria2), which only provide portable *archived* files.
There maybe also some apps just releases an *standalone* .exe file.
Although we can request they to build an msi/msix package, it will be more convenient if we support to *install* them directly. 

For the portable apps, the *install* may mean extract the archive to *Program Files*, then create a [shim](https://github.com/chocolatey/shimgen)-like redirection for the program, put it into a specific directory, and add the directory to PATH.

Related: 
* https://github.com/microsoft/winget-pkgs/issues/174
* https://github.com/microsoft/winget-cli-restsource/issues/116

Edited by @denelon

timvisee:
Yes, I'd like to see this as well. Many other tools you might use in the CLI (it is a CLI package installer after all) are provided as standalone binaries. I'd like to package [ffsend](https://github.com/timvisee/ffsend).

Existing package managers for Windows such as scoop or Chocolatey allow the installation of standalone binaries as well. Some inspiration might be taken from there.

RobCannon:
I would love to be able to install tools like terraform, packer, kubectl, etc..  I could use scoop.sh, but I would love to have one tool to handle everything.  winget should be able to install command line tools to a common directory in the user's directory but also make sure that directory is part of the path.

marcomorain:
I'm looking for this support to package the CircleCI CLI tool.

👍

koppor:
This is useful for the [mob tool](https://github.com/remotemobprogramming/mob#swift-git-handover-with-mob) which is fundamental for [remote mob programming](https://www.remotemobprogramming.org/). A [release](https://github.com/remotemobprogramming/mob/releases) for windows is a single `.tar.gz` file containing a single `.exe` file.

darrenj:
Also useful for the [ngrok](https://ngrok.com) tool for HTTP tunnelling.

karsten3s:
To piggy back off of this, how would we go about installing applications like [https://github.com/ggreer/the_silver_searcher](url) where there will be no direct developer support for windows? There are unofficial windows builds but can those be considered?

zubairq:
I would like to be able to package a standalone app too, Yazz Pilot which is a single .exe file. Any estimate on when this will be coming?

ajeetdsouza:
[...]

Issue #140, opened 2020-05-14
Support installing .zip files


denelon:
I want to be able to have winget install apps inside of .zip files.

Related to:
* https://github.com/microsoft/winget-cli-restsource/issues/117

# Proposed technical implementation details (optional)

utybo:
This would be great! Gradle has always required a package manager on Windows (I'm currently using it with Scoop) and having it on winget would be fantastic, but it's a .zip file.

Also, wouldn't zip files (and standalone .exe apps) require some more complicated setup steps that would have to be described in winget's YAML file? e.g. adding things to PATH, creating a desktop link, creating a Start menu entry etc.

zuozishi:
>

> This would be great! Gradle has always required a package manager on Windows (I'm currently using it with Scoop) and having it on winget would be fantastic, but it's a .zip file.
> 
> Also, wouldn't zip files (and standalone .exe apps) require some more complicated setup steps that would have to be described in winget's YAML file? e.g. adding things to PATH, creating a desktop link, creating a Start menu entry etc.

And environment variable.

DennisGaida:
I think there should be a default _Apps_ folder maybe within the %USERPROFILE%. Unzipped zips should be per default copied there maybe within a subfolder. The same may also happen for PowerShell Scripts (PS1) and EXEs.

PedroLamas:
I was about to create a manifest for Paint.net when I noticed that .zip is not supported... yet! 🙂

pashagolub:
A lot of .msi installers are packed into .zip files because you cannot execute .msi downloaded from internet. ;-) 

I wanted to add PostgreSQL support, but lack of zip extraction is a problem here.

maxandersen:
related to https://github.com/microsoft/winget-pkgs/issues/580

Right now Winget can't be used to package/install anything without complex installers.

chausner:
I think there are actually two issues or requests related to ZIP:
1. Support software that is provided as ZIP archives for plain xcopy deployment
2. Support installers (MSI, EXE, etc.) that are made available for download in a ZIP file (e.g. https://mh-nexus.de/de/downloads.php?product=HxD20)

timsneath:
@aclinick, ideally it would be a) unzip a ZIP file; b) (optionally) run a batch file script to perform some configuration. Would be reasonable if there's also an uninstall script that reverses the latter. 

We'd use this for installing Dart and Flutter on Windows.

marypcbuk:
[...]

PR #1717, opened 2021-11-16
Added ZIP support spec

midoriiro:
Added ZIP support spec for community review as draft

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1717)

### DIFF
--- a/doc/specs/#140 - ZIP Support.md
+++ b/doc/specs/#140 - ZIP Support.md
@@ -11,26 +11,26 @@ For [#140](https://github.com/microsoft/winget-cli/issues/140)
 
 ## Abstract
 
-This feature adds to winget-cli the ability to install application delivered through a ZIP archive.
+This feature adds to winget-cli the ability to install application deserved thought a ZIP archive.
 
 ## Inspiration
 
 A bunch of applications are only provided by developer's through a ZIP archive. There are three cases:
-1) The archive contain only one binary that need do to be extracted and add to PATH env var
-2) The archive contain an installer that need to be executed and let do his job
-3) The archive contain a complex hierarchy that need do to be extracted and add specific folder to PATH env var
+1) The archive contains only one binary (a portable / standalone executable).
+2) The archive contains only one supported installer (msix, msi, or exe).
+3) The archive contains more than one file.
 
 ## Solution Design
 
-A new installation flow need to be created to execute different installation strategies regarding these three cases above.
+A new installation flow need to be created to execute different installation strategies regarding the three cases above.
 
-First thing first, the archive need to be extracted in a specific path (TEMP folder can be a good candidate).
+First thing first, the archive needs to be extracted in a specific path (TEMP folder can be a good candidate).
 
-Regarding case 1, installation flow check if TEMP folder contain a single binary and repack it
+Regarding case 1, installation flow check if TEMP folder contain a single binary and repack that binary
 into an MSI package (including setting PATH env var). Then, MSI installation flow can take control.
 
 Regarding case 2, installation flow check if TEMP folder contain an installer (.exe, .msi, .msix) and pass control
-to MSI/Other installation flow.
+to MSI installation flow.
 
 Regarding case 3, installation flow check if TEMP folder does not contain an installer nor a single binary.
 This case is the complex one, example with 'vagrant' hierarchy:
@@ -92,7 +92,7 @@ ManifestType: installer
 ManifestVersion: 1.0.0
 ```
 
-This is not compliant with the actual schema v1.1.0. We need to specify which binary to add into
+This is not compliant with the actual schema v1.1.0. We need to specify which binary to add in
 PATH env var. Therefore, schema need to be updated to take knowledge of that information.
 
 ### Repack ZIP into MSI package
--- a/doc/specs/#140 - ZIP Support.md
+++ b/doc/specs/#140 - ZIP Support.md
@@ -1,7 +1,7 @@
 ---
 author: Alexis Bekhdadi @midoriiro
 created on: 2021-11-16
-last updated: 2021-11-17
+last updated: 2021-11-30
 issue id: 140
 ---
 
@@ -26,11 +26,11 @@ A new installation flow need to be created to execute different installation str
 
 First thing first, the archive needs to be extracted in a specific path (TEMP folder can be a good candidate).
 
-Regarding case 1, installation flow check if TEMP folder contain a single binary and repack that binary
-into an MSI(X) package (including setting PATH env var). Then, MSI(X) installation flow can take control.
+Regarding case 1, installation flow check if TEMP folder contain a single binary and pass instruction to 
+installation flow (including setting PATH env var).
 
-Regarding case 2, installation flow check if TEMP folder contain an installer (.exe, .msi, .msix) and pass control
-to MSI(X) installation flow.
+Regarding case 2, installation flow check if TEMP folder contain an installer (.exe, .msi, .msix, etc) and pass control
+to proper installation flow.
 
 Regarding case 3, installation flow check if TEMP folder does not contain an installer nor a single binary.
 This case is the complex one, example with 'vagrant' hierarchy:
@@ -42,15 +42,15 @@ This case is the complex one, example with 'vagrant' hierarchy:
     - bin
     - some other folders
 
-This hierarchy can be packed into an MSI(X) installer, but how can we specify which binary to add 
-in PATH env var (in that case bin/vagrant.exe)?
+This hierarchy can be packed into an MSI(X) installer or register the application in ARP at the end of installation 
+flow (see section below about repack or ARP entry).
 
 This specification take on consideration schema on version 1.1.0. 
 A schema update should be necessary to take all the cases in consideration. 
 
 Take a look into the following manifest examples
 
-### Case 1|2 YAML manifest
+### Case 1 YAML manifest
 
 ```
 PackageIdentifier: Editor.App
@@ -73,7 +73,7 @@ ManifestVersion: 1.0.0
 
 This is compliant with the actual schema v1.1.0
 
-### Case 2.1 YAML manifest
+### Case 2 YAML manifest
 
 If the archive contains an installer but not located at root folder (e.g. archive.zip/somefolder/installer.exe)
 
@@ -93,12 +93,17 @@ Installers:
   InstallerUrl: https://editor.com/app/1.0.zip
   InstallerSha256: ...
   InstallerPath: somefolder/installer.exe
+  NestedInstallerType: msi
 ManifestType: installer
 ManifestVersion: 1.0.0
 ```
 
 This is not compliant with the actual schema v1.1.0.
-```InstallerPath``` attribute should be added to schema.
+```InstallerPath``` key should be added to schema. 
+This optional key define where to find installer file after extraction if not located at root.
+
+To specify installer type, ```NestedInstallerType``` key must be defined.
+As mentioned in issue [#1242](https://github.com/microsoft/winget-cli/issues/1242).
 
 ### Case 3 YAML manifest
 
@@ -126,19 +131,51 @@ This is not compliant with the actual schema v1.1.0.
 We need to specify which binary folder to add in PATH env var.
 ```BinaryPath``` attribute should be added to schema.
 
-### Repack uncompressed files into a supported package format
+### Processing uncompressed files after extraction
 
-There are many tool to repack an MSI(X) package, some expansive and some apparently free (Advanced Installer ?).
-However, Windows SDK provide a CLI tool called MakeAppx in order to create MSIX package.
+Two solutions can be taken in consideration. Repack uncompressed files into an installer or add an ARP entry in
+registry.
+
+For case 2, the manifest should contain all information needed to other installation flow. There is no need to repack 
+or add an ARP entry.
+
+For all others cases, repacking a file structure might not be a good solution. This process require installing on client
+side a third party repacking solution or installing Windows SDK that provide MakeAppx that repack our file structure
+into a MSIX installer. But not everyone needs to install Windows SDK and repacking can take time and resources.
+
+Adding an ARP entry into registry can be done more easily, actually winget-cli already read ARP entries through 
+[ARPHelper](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp) 
+struct. 
+
+Windows [documentation](https://docs.microsoft.com/en-us/windows/win32/msi/uninstall-registry-key) define properties for
+an ARP entry (these properties are also defined in ARPHelper struct).
+
+An ARP entry can be defined in two location:
+ - HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Uninstall (scope machine)
+ - HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall (scope user)
+
+Uninstall key contains sub keys identified by the application product code, but for our case 1 and 3 a product code
+might not exist. Two solutions can be considered, the package maintainer can generate a random GUID but this identifier 
+should be reused on each manifest for a particular application. The second solution is to set the sub key name by the 
+application name which can be retrieved on the manifest with the ```PackageName``` key.
+
+Then the installation flow should register an ARP entry with the proper properties defined in manifest.
+
+Our file structure should be moved into a proper installation folder.
+Depending on the scope and the architecture we have different folders:
+  - %HOMEDRIVE%%HOMEPATH%/AppData/Local (scope user)
+  - %PROGRAMFILES% (scope machine, architecture x64)
+  - %PROGRAMFILES(X86)% (scope machine, architecture x86)
+
+> Above section might need more specification about where to install the file structure
+
+Finally, if manifest key ```BinaryPath``` is defined the installation flow should add to PATH environment variable the
+resolved binary path folder.
 
 ## UI/UX Design
 
-One downside of this feature is the package will be built on the client side.
-if the uncompressed files need to be packaged into an MSI(X) installer, that mean installing on client side
-an external or internal tool to repacking the uncompressed files is needed. In the case of MakeAppx, that mean
-installing a Windows SDK.
+If we exclude repacking solutions and prefer add/edit ARP entry, that should be transparent to the end-user.
 
-But with a proper MSI(X) repacking solution, this operation should be transparent to the end-user.
 - For case 2, the existing installation flow will take control and behave as expected
 - For case 1 and 3, the new installation flow need to behave like other installation flow (showing steps and progress) 
 
@@ -150,7 +187,7 @@ This should have no direct impact on accessibility.
 
 ### Security
 
-Need more specification
+> Need more specification
 
 ### Reliability
 
@@ -165,15 +202,15 @@ No code breaking, reuse existing code and add code to support zip installation.
 
 ### Performance, Power, and Efficiency
 
-Using an internal/external repacking MSI(X) installer solution will impact client device performance. 
+Add/edit an ARP entry should not be an issue for performance, power and efficiency 
 
 ## Potential Issues
 
-Need more specification
+> Need more specification
 
 ## Future considerations
 
-Initial implementation will only take in consideration .zip archive. However, when the feature will be stable
+Initial implementation will only take in consideration .zip archive. However, when the feature will be stable enough,
 we can consider other types of archives (e.g. *.gz. *.7z, etc). 
 
 ## Resources
--- a/doc/specs/#140 - ZIP Support.md
+++ b/doc/specs/#140 - ZIP Support.md
@@ -1,0 +1,135 @@
+---
+author: Alexis Bekhdadi @midoriiro
+created on: <2021-11-16>
+last updated: <2021-11-16>
+issue id: 140
+---
+
+# Spec Title
+
+For [#140](https://github.com/microsoft/winget-cli/issues/140)"
+
+## Abstract
+
+This feature adds to winget-cli the ability to install application deserved thought a ZIP archive.
+
+## Inspiration
+
+A bunch of applications are only provided by developer's through a ZIP archive. There are three cases:
+1) The archive contain only one binary that need do to be extracted and add to PATH env var
+2) THe archive contain an installer that need to be executed and let do his job
+3) The archive contain a complex hierarchy that need do to be extracted and a specific folder to PATH env var
+
+## Solution Design
+
+A new installation flow need to be created to execute different installation strategies regarding the three cases above.
+
+First thing first, the archive need to be extracted in a specific path (TEMP folder can be a good candidate).
+
+Regarding case 1, installation flow check if TEMP folder contain a single binary and repack that binary
+into an MSI package (including setting PATH env var). Then, MSI installation flow can take control.
+
+Regarding case 2, installation flow check if TEMP folder contain an installer (.exe, .msi, .msix) and pass control
+to MSI installation flow.
+
+Regarding case 3, installation flow check if TEMP folder does not contain an installer nor a single binary.
+This case is the complex one, example with 'vagrant' hierarchy:
+- root
+  - bin 
+    - vagrant.exe
+    - other.exe
+  - embedded
+    - bin
+    - some other folders
+
+This hierarchy can be packed into an MSI installer, but how can we specify which binary to add 
+in PATH env var (in that case bin/vagrant.exe)?
+
+Take a look into the following manifest examples (v1.1.0)
+
+### Case 1|2 YAML manifest
+
+```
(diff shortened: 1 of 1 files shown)


ItzLevvie (changes requested):
Looks like this PR's touching files (467,403 additions and 467,269 deletions across 1314 different files) that shouldn't which makes reviewing changes a bit difficult.

midoriiro:
> 
> 
> Looks like this PR's modifying files (467,403 additions and 467,269 deletions across 1314 different files) which makes reviewing changes a bit difficult.

Exact, did not notice that. I just clone with git bash. I'll try to fix that.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+For [#140](https://github.com/microsoft/winget-cli/issues/140)
+
+## Abstract
+
+This feature adds to winget-cli the ability to install application delivered through a ZIP archive.
+
+## Inspiration
+
+A bunch of applications are only provided by developer's through a ZIP archive. There are three cases:
+1) The archive contain only one binary that need do to be extracted and add to PATH env var
```
denelon, line 19:
```suggestion
1) The archive contains only one binary (a portable / standalone executable).
```

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+2) The archive contain an installer that need to be executed and let do his job
```
denelon, line 20:
```suggestion
2) The archive contains only one supported installer (msix, msi, or exe).
```

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+3) The archive contain a complex hierarchy that need do to be extracted and add specific folder to PATH env var
```
denelon, line 21:
```suggestion
3) The archive contains more than one file,
```

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Solution Design
+
+A new installation flow need to be created to execute different installation strategies regarding these three cases above.
```
denelon, line 25:
```suggestion
A new installation flow needs to be created to execute different installation strategies regarding these three cases above.
```

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+First thing first, the archive need to be extracted in a specific path (TEMP folder can be a good candidate).
+
+Regarding case 1, installation flow check if TEMP folder contain a single binary and repack it
+into an MSI package (including setting PATH env var). Then, MSI installation flow can take control.
```
denelon, line 30: Packaging a standalone / portable application is an interesting approach. I'm not sure we had considered this specific approach. We may prefer the concept of repackaging as an MSIX as that is the latest technology, but there may be different concerns to discuss here.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+This case is the complex one, example with 'vagrant' hierarchy:
+- root
+  - bin 
+    - vagrant.exe
+    - other.exe
+  - embedded
+    - bin
+    - some other folders
+
+This hierarchy can be packed into an MSI installer, but how can we specify which binary to add
```
denelon, line 45: This use case is why packaging an MSIX may be better than an MSI, but we will need to dig a bit deeper here.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+    - vagrant.exe
+    - other.exe
+  - embedded
+    - bin
+    - some other folders
+
+This hierarchy can be packed into an MSI installer, but how can we specify which binary to add 
+in PATH env var (in that case bin/vagrant.exe)?
+
+Take a look into the following manifest examples (v1.1.0)
```
denelon, line 48: We should stick with only supporting this in the 1.1.0 or later schema.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+Installers:
+- Architecture: x64
+  InstallerUrl: https://editor.com/app/1.0.zip
+  InstallerSha256: ...
+  BinaryPath: bin/binary.exe
+ManifestType: installer
+ManifestVersion: 1.0.0
+```
+
+This is not compliant with the actual schema v1.1.0. We need to specify which binary to add into
```
denelon, line 95: If we need additional metadata, this would be a newer schema than 1.1.0. Once we sufficiently capture all the use cases, we can certainly look at extending the schema. The clients lacking support for the 1.1.0 or later schema must be able to determine this is not going to be a valid scenario, so the .zip would not be a supported installer type on those clients.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+PATH env var. Therefore, schema need to be updated to take knowledge of that information.
+
+### Repack ZIP into MSI package
```
denelon, line 98:
```suggestion
### Repack uncompressed files into a supported package format
```

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+This section need more specifications. There are many tool to repack an MSI package, some expansive and
+some apparently free (Advanced Installer ?).
+
+## UI/UX Design
+
+One downside of this feature is the MSI repackaging will be executed on client side.
```
denelon, line 105:
```suggestion
One downside of this feature is the package will be built on the client side.
```

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+Which mean for each installation the same process will be executed and remain the same for an
```
denelon, line 106: This will certainly have performance implications. We also need to ensure rational versioning so the upgrade flow also works as expected.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+This section need more specifications. There are many tool to repack an MSI package, some expansive and
+some apparently free (Advanced Installer ?).
+
+## UI/UX Design
+
+One downside of this feature is the MSI repackaging will be executed on client side.
+Which mean for each installation the same process will be executed and remain the same for an 
+application x.y.z.
+
+But with a proper MSI repacking solution, this operation should be transparent to the end-user.
```
denelon, line 109: We would likely have some output for the user with the packaging step so they at least get an indication of progress.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Capabilities
+
+### Accessibility
+
+This should have no direct impact on accessibility.
+
+### Security
+
+No direct impact on security.
```
denelon, line 119: There would most certainly be security implications. We may even need a SHA256 hash for each of the files to be packaged. We don't want to be in a state where the files are downloaded and modified prior to the packaging steps. This will also have a substantial impact on the validation flow so we can verify the hashes are valid at validation time, and correctly included in the manifest for any changes to the manifest.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Reliability
+
+This is not expected to impact reliability.
+
+### Compatibility
+
+No code breaking, reuse existing code and add code to support zip installation.
```
denelon, line 127: This would not be a backward compatible behavior. Only clients with the support for this feature would be able to install packages of this type. We would need to add something like "zip" to the enumeration of installer types. The initial implementation would also be an experimental feature until we have covered enough of the different use cases to confirm viability of the solution.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Performance, Power, and Efficiency
```
denelon, line 129: Building the installer is likely to take additional time and resources on the client. There will be performance considerations here with the client in the packaging phase.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Potential Issues
```
denelon, line 131: MSIX packages are typically signed. There would likely be security issues related to signing. If the MSI route or some other mechanism is used for the package, we would also need to look at those considerations. Even if we just move the standalone / portable application (or set of files) to a directory path, any changes to those files could have unintended consequences. Some packages also generate additional files in their directories so those may need to be considered during an upgrade or uninstall flow.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Future considerations
```
denelon, line 133: Other compression formats may also need to be considered in the future. The initial implementation is likely to be limited to .zip files.

denelon:
Thanks for taking the time to write this draft PR! Please be patient as we're dealing with limited resources during this time of year. We also need time for additional internal design review, and have a security review performed to make sure we aren't facing any other additional security concerns. There may also be licensing or copyright concerns to consider if the publisher or the artifacts aren't governed by an open source license.

doc/specs/#140 - ZIP Support.md
JohnMcPMS: I'm not sure that writing code to transform the archive into an installable type is better than simply writing the code to manage archives directly.  All of the proposals require unpacking the archive, which is itself the primary step in management.  Afterward, adding to the `%PATH%` is fairly straightforward and the issues are more about lifetime management.  Injecting our own ARP entry is largely sufficient to achieve that.

doc/specs/#140 - ZIP Support.md
JohnMcPMS: Currently you would get an `ERROR_NOT_SUPPORTED` result.  If you think that isn't a good experience, we should come up with a better one and implement now to increase the chances of the downlevel client having it.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+InstallModes: ...
+InstallerSwitches:
+  Silent: ...
+  SilentWithProgress: ...
+UpgradeBehavior: install
+Installers:
+- Architecture: x64
+  InstallerUrl: https://editor.com/app/1.0.zip
+  InstallerSha256: ...
+  BinaryPath: bin/binary.exe
```
JohnMcPMS, line 90: If this is intended to be the `%PATH%` update location, it should be to a directory and not a file.  The default would be either be no `%PATH%` update or the root.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,135 @@
...
+  - embedded
+    - bin
+    - some other folders
+
+This hierarchy can be packed into an MSI installer, but how can we specify which binary to add 
+in PATH env var (in that case bin/vagrant.exe)?
+
+Take a look into the following manifest examples (v1.1.0)
+
+### Case 1|2 YAML manifest
```
JohnMcPMS, line 50: I would argue for case 2 to be the one that includes a relative path to the installer, in the event that there are support files included within the archive.

doc/specs/#140 - ZIP Support.md
JohnMcPMS: The cases I see:
1. Archive is the set of files that should be placed on the system (both your 1 and 3)
2. Archive is a container for the installer, and is itself temporary (your 2)
3. [New] Archive is basically both of the above; the files themselves need to be kept, but there is some registration step that is needed.

doc/specs/#140 - ZIP Support.md
jedieaston: MSIX packaging client side is impossible right now due to the signing requirements, no? Right now, you either have to sign in with Azure AD to get a short lived (48 hours?) cert, or buy a longer one.

doc/specs/#140 - ZIP Support.md
jedieaston: > if we just move the standalone / portable application (or set of files) to a directory path, any changes to those files could have unintended consequences. 

which is true of any package winget currently installs, as with admin rights (or for user scoped packages, normal rights) I can go mess around with anything the installer created. I think that might be a separate issue.

doc/specs/#140 - ZIP Support.md
midoriiro: @denelon I'm not an expert concerning Windows packaging application. The main idea of this specification is to package the content of a zip file (except for case 2) in order to let winget managing the lifecycle of the new package. 

I saw Windows SDK offers a set of CLI command to create MSIX package, but that mean installing Windows SDK.

@JohnMcPMS  I don't know what is ARP, but if I understood your suggestion. How you do you handle uninstall/upgrade of the package ? That the goal of an ARP entry ?

doc/specs/#140 - ZIP Support.md
midoriiro: That could be tricky to add all checksum for all files in the manifest. If the ZIP archive is validated can we just lock the extracted files before packaging to prevent any modifications by other process ?

doc/specs/#140 - ZIP Support.md
midoriiro: Yes, you are right. For the moment this draft is only intended to schema 1.1.0. Like i suggested, we maybe need to update the schema (1.1.1?) to handle case 3 and other new cases.

doc/specs/#140 - ZIP Support.md
midoriiro: Like someone said on issue #140, if a file is modified is not the packager responsibility but it's an user fault. If files are created by a process delivered by the package, these files should be referenced somewhere to prevent overwriting them, or rather we can specify to overwrite theses files during an upgrade. Typically for config files, user need to edit them, that could be a problem if on each upgrade these files are overwritten. 

Concerning signing a MSIX package, A self-signed package could be acceptable ?

doc/specs/#140 - ZIP Support.md
midoriiro: Yes, I noticed my mistake before your write your comment ^^".  You are right, BinaryPath should be a relative path from the root archive.

doc/specs/#140 - ZIP Support.md
midoriiro: Case 2 expect to have an installer at the root archive. But if the installer is located to 'archive.zip/somefolder/installer.exe' that should be a new case in my opinion. To avoid having more case I suggest to add a new attribute like BinaryPath.

midoriiro:
@denelon I know this feature is not planned to be released shortly. However, I'll take on consideration your feedback (@JohnMcPMS included) and push a new revision soon, so we can discuss it when you will be available ;-)

doc/specs/#140 - ZIP Support.md
midoriiro: You mean that information (like version, name, etc) from the winget manifest should be linked to built MSI(X) package ?

doc/specs/#140 - ZIP Support.md
denelon: Add Remove Programs (ARP) is where many of the entries for programs installed on Windows 10 and Windows 11 get recorded. We use that metadata for the list command and as the source for values we used to correlate with manifests in configured sources. We don't necessarily need to create an "insaller" per se'. It's possible to just write entries to ARP (now part of Apps & Features). The Windows Package Manager could act as an "installer" proxy to create the entries and modify the path environment variable. We're looking at pros and cons to several approaches.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,180 @@
...
+InstallModes: ...
+InstallerSwitches:
+  Silent: ...
+  SilentWithProgress: ...
+UpgradeBehavior: install
+Installers:
+- Architecture: x64
+  InstallerUrl: https://editor.com/app/1.0.zip
+  InstallerSha256: ...
+  BinaryPath: bin (e.g. bin folder where vagrant.exe is located in my example)
```
Trenly, line 120: Could this be simplified to "InstallerPath" as well, and have the cli determine if it is an installer or binary?

midoriiro: Case 2 cover installer (.exe, .msi, etc). You're pointing case 3, which is for complex hierarchy that does not include installer, only binary with complex folders structure. The base idea is to implement all the cases. Nonetheless, "BInaryPath" maybe not a good name, any idea ?

doc/specs/#140 - ZIP Support.md
midoriiro: I see, it's make sense. I looked if an API exist about managing an ARP entry, but I did found nothing. If winget-cli already manage that kind of tasks, @JohnMcPMS made a good point.

I don't know what is the state of art about ARP in Windows 10/11 now. But, ten years ago I wrote an application that find all installed applications on the system through the registry, the main goal was to uninstall silently a batch of applications. That was tricky to handle all the cases. 

If an API already exist, the repacking solution would not be the best approach(in my opinion) as you explained both of you.

I'll try to find some documentation on the internet and this repository to study how to manage an ARP entry.

doc/specs/#140 - ZIP Support.md
Trenly: I’m saying that for case 2, you are suggesting that a key be added to define the path to the exe/msi/etc.

Why can’t that same key be used for case 3?

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,180 @@
...
+InstallModes: ...
+InstallerSwitches:
+  Silent: ...
+  SilentWithProgress: ...
+UpgradeBehavior: install
+Installers:
+- Architecture: x64
+  InstallerUrl: https://editor.com/app/1.0.zip
+  InstallerSha256: ...
+  InstallerPath: somefolder/installer.exe
```
Trenly, line 95: I think we may also need a key defining the secondary installer type.

It is likely that some of the exe’s shipped inside zip packages are actually nullsoft or inno, which should be preferred to exe since the CLI already handles their silent switches without having to define them

doc/specs/#140 - ZIP Support.md
midoriiro: I got confused you was quoting case 3. On case 3, BinaryPath key is for indicate to register a path after the zip extraction into user or system environment variables (need to be defined how to). 

Correct me if I'm wrong, but from my point of view regarding case 2 (excluding my example manifest 2.1). Case 2 assuming that the zip archive only contains one file, the installer. Case 2 also assuming that the installer will set the proper environment variables if needed (specifically $PATH) or not. that depend of the nature of the application and how editors package their applications. I don't know, how do you manage that kind of thing, winget-cli is agnostic regarding this aspect ?

To answer your question, case 3 does not handle installer file. That the job of case 2 and 2.1. Case 2 does not need to define InstallerPath key because, the archive only contains the installer at the root archive. On the other hand, case 2.1 zip archive maybe not contains only one file and maybe not located at root, so InstallerPath key is needed to indicate where to find the installer after the extraction and execute it.

To summarize, BinaryPath and InstallerPath are optional keys that depend on the context and they should be set properly by the package maintainer. To avoid messing up with these settings, updating the documentation and the helper script YAMLCreate.ps1 should be needed, in my opinion. For all cases, if the manifest is miss-configured the procedure should fail in an organic way. 

I don't know if I'm clear, my english is not perfect. Maybe I should clarify these aspects on the specification.

doc/specs/#140 - ZIP Support.md
midoriiro: A key definition already exist on schema [1.1.0](https://github.com/midoriiro/winget-cli/blob/03f8a555693003129e0199759135d874a580a0b5/schemas/JSON/manifests/v1.1.0/manifest.installer.1.1.0.json#L459) (line 459)

Maybe we can use it to tell which installation flow to choose?

Trenly: That key, by definition, would be "zip" since the primary installer is inside of a .zip file. The issue I'm trying to get ahead of is that the zip file could contain different types of exes which have their own installation flow

midoriiro: The key I'm pointing could be resolve that issue with something like this:

```
PackageIdentifier: Editor.App
PackageVersion: 1.0
MinimumOSVersion: 10.0.0.0
InstallerType: zip
Scope: ...
InstallModes: ...
InstallerSwitches:
  Silent: ...
  SilentWithProgress: ...
UpgradeBehavior: install
Installers:
- Architecture: x64
  InstallerUrl: https://editor.com/app/1.0.zip
  InstallerSha256: ...
  InstallerPath: somefolder/installer.exe
  InstallerType: inno # or something else
ManifestType: installer
ManifestVersion: 1.0.0
```

Trenly: I don't believe that is valid. Having the key at the manifest level is not compatible with the key at the installer level. One of them will override the other.

midoriiro: I see, which component validate the manifest and could be responsible of overriding these key ? An other component or the installation flow ?

Trenly: I'm not sure I understand your question.

The overriding is an inherent property of the CLI. If a key is specified at the installer level, it is overridden by the manifest level. You can't have them defined at both levels
```
PackageIdentifier: ...
PackageVersion: ...
InstallerType: zip <----- THIS IS THE MANIFEST LEVEL
Installers:
- Architecture: ...
  InstallerUrl: ...
  InstallerSha256: ...
  InstallerType: nullsoft<----- THIS IS THE INSTALLER LEVEL
ManifestType: installer
ManifestVersion: 1.0.0
```

midoriiro: Ok i see, that's a dead end. About the issue, can you provide a concrete example with multiple installers included in a zip archive ? That could help to better understanding and had a better overview on this issue.

Trenly: I’m not saying multiple installers may be in a .Zip archive. I’m saying a manifest may have different types of installers, and a zip archive may be one of them.

For example - the 7-zip manifest has installers for both exe and msi. Both are in the same manifest.

Some publishers pubish exe’s, msi’s, and zip files, which should all be able to be in a single manifest

midoriiro: Ok I understand now. And I agree with you about defining another key to specify the installer type. Maybe InstallerSubType ?

I need to clarify some points on the specification about previous conversations. I'll commit soon.

denelon: https://github.com/microsoft/winget-cli/issues/1242 is how we're planning to capture some of the "nested installer type" issues. For example, PowerToys is distributed as an .exe, but it installs as an .msi. The installer is a wrapper to help with dependencies.

Trenly: Perfect, I think that as long as that issue is referenced in the spec, then there won't be an issue here

midoriiro: That should close this issue. Thank for your feedback.

doc/specs/#140 - ZIP Support.md
JohnMcPMS: I don't think that there are APIs; we certainly do not use them and I looked for them when we implemented reading.  It is basically a registry data contract.

midoriiro: Yes, I saw that yesterday. I updated the spec. There some points to discuss about it but, especially where to install the zip content. The ARP section seem good for you ?

parithon: Perhaps we treat ZIP installs the same way visual studio code is installed (i.e. user installer) within the AppData folder.

midoriiro: That I suggested in the last revision of the spec. For user scope, application will be installed in AppData/Local/myapplication. But maybe there is a better location for user scope ? 

In any ways, scope and location can be overrides by users when calling winget install. Just found that, that should be take in consideration on the specification.

midoriiro:
What's the state about my last commits? Someone has time to review them? I know you're busy, I'm just checking. Thanks

denelon:
We're getting ramped up on how we're going to tackle .zip support. We will take as much of what has been drafted as possible.

ryfu-msft:
Hey @midoriiro,

The team has begun working on a design for adding .ZIP installer support. As this spec is a great starting point and a lot of important discussions have taken place, we plan to discuss and incorporate key components that you have brought to our attention into our design review. We will continue to update the community as we make progress. Great work :)

denelon:
We're getting closer with support for portable applications:

* #182 

Once that is "code complete", we will turn our attention back to .zip.
I just wanted to provide an update 😊

JohnMcPMS:
Superseded by #2270

PR #1718, opened 2021-11-16
Make telemetry logger per context and send summary event on destruction

yao-msft:
## Change
- Change TelemetryTraceLogger to be per context
- Add parent activity id concept to replace the subExecutionId
- Send summary event upon each TelemetryTraceLogger destruction

## Validation
Due to lack of test infra support for telemetry, I manually run and see telemetry events and also set breakpoints to verify the summary events.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1718)


src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -132,6 +175,33 @@ namespace AppInstaller::Logging
...
+    std::unique_ptr<TelemetryTraceLogger> TelemetryTraceLogger::CreateSubTraceLogger() const
+    {
+        THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !this->m_isInitialized);
+
+        auto subTraceLogger = std::make_unique<TelemetryTraceLogger>();
+
+        subTraceLogger->m_parentActivityId = this->m_activityId;
+        subTraceLogger->m_subExecutionId = s_subExecutionId++;
+
+        // Copy remaining fields from parent.
```
JohnMcPMS, line 192: We should start with the copy constructor, then modify from there.  That way we don't miss new fields.  If necessary, override the copy constructor for the summary event to simply default construct instead.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -570,14 +808,20 @@ namespace AppInstaller::Logging
         }
 #endif
         ThreadLocalStorage::ThreadGlobals* pThreadGlobals = ThreadLocalStorage::ThreadGlobals::GetForCurrentThread();
-        if (pThreadGlobals)
+        if (pThreadGlobals && pThreadGlobals->ContainsTelemetryLogger())
```
JohnMcPMS, line 811: In what cases is this desired to be null?

src/AppInstallerCLICore/Core.cpp
```diff
@@ -46,6 +46,8 @@ namespace AppInstaller::CLI
     {
         init_apartment();
 
+        Logging::UseGlobalTelemetryLoggerActivityIdOnly();
```
JohnMcPMS, line 49: Is this intended to stay until the back end has switched to the summary event?

src/AppInstallerCLICore/Commands/CompleteCommand.cpp
```diff
@@ -53,8 +53,10 @@ namespace AppInstaller::CLI
             }
 
             // Create a new Context to execute the Complete from
-            auto subContextPtr = context.Clone();
+            auto subContextPtr = context.CreateSubContext();
             Context& subContext = *subContextPtr;
+            auto previousThreadGlobals = subContext.GetThreadGlobals().SetForCurrentThread();
```
JohnMcPMS, line 58:
```suggestion
            auto contextActive = subContext.SetActive();
```
As well as creating the appropriate `SetActive`.  I think it is easier to follow.

src/AppInstallerCLICore/Core.cpp
```diff
@@ -61,7 +63,7 @@ namespace AppInstaller::CLI
         // Initiate the background cleanup of the log file location.
         Logging::BeginLogFileCleanup();
 
-        Execution::Context context{ std::cout, std::cin };
+        Execution::Context context{ std::cout, std::cin, nullptr /* Command execution always use global diagnostic logger */, Logging::Telemetry().CreateSubTraceLogger() };
```
JohnMcPMS, line 66: If we are moving to the CLI using context specific objects, we should go all the way and not ever use the global one.  I would prefer that we didn't have a global one anyway, but getting rid of it was too difficult at the time.

I can see the desire to have a single file logger, but I think that the telemetry should be rooted here, not a child of the global telemetry.

And then set this as the active context of course.

src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -192,19 +192,19 @@ namespace AppInstaller::CLI::Execution
...
         }
-        else if (m_executionStage > stage && !allowBackward)
+        else if (m_executionStage > stage)
         {
             THROW_HR_MSG(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), "Reporting ExecutionStage to an earlier Stage without allowBackward as true");
         }
 
         m_executionStage = stage;
-        Logging::SetExecutionStage(static_cast<uint32_t>(m_executionStage));
+        Logging::Telemetry().SetExecutionStage(static_cast<uint32_t>(m_executionStage));
```
JohnMcPMS, line 207: This seems like madness.  The owner of the thread specific telemetry object requesting it via the thread locals rather than getting it directly.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Core.cpp
```diff
@@ -51,19 +51,25 @@ namespace AppInstaller::CLI
...
+        auto previousThreadGlobals = context.SetForCurrentThread();
+        context.EnableCtrlHandler();
 
         // Enable all logging for this phase; we will update once we have the arguments
         Logging::Log().EnableChannel(Logging::Channel::All);
         Logging::Log().SetLevel(Logging::Level::Info);
         Logging::AddFileLogger();
         Logging::EnableWilFailureTelemetry();
 
+#ifndef AICLI_DISABLE_TEST_HOOKS
```
JohnMcPMS, line 66: This should be done before everything but COM initialization.

src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -184,6 +184,7 @@ namespace AppInstaller::CLI::Execution
     void Context::SetTerminationHR(HRESULT hr)
     {
         m_terminationHR = hr;
+        m_isTerminated = FAILED(hr);
```
JohnMcPMS, line 187:
```suggestion
        m_isTerminated = true;
```
Should stay consistent with `Terminate`.

JohnMcPMS (changes requested):

JohnMcPMS (approved):

PR #1720, opened 2021-11-17
Updated solution to build under Visual Studio 2022

jedieaston:

-----

This PR ended up being way bigger than I thought it would be, but in general, it makes the necessary changes for winget to build using Visual Studio 2022's toolchain (MSVC v143). 

Here's what I did:
- Changed all of the project files to conditionally use the v143 compiler if you're using Visual Studio 2022 (same way that v142 was conditionally used for many of the project files for 2019).
- ~~Updated Valijson to version 0.6. This is the vast majority of the changes, and maybe should be a separate PR (up to y'all). The reason for this is that Valijson made a change to their CMake configuration [here](https://github.com/tristanpenman/valijson/commit/cb778b6bb60d609a18a0ebc48c79e8ee6d4bd92d#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a) to allow it to work properly with Visual Studio. Given that the version currently in the repo is from January, it seemed reasonable to do an update (and since there's a big "External - Do Not Modify" on the vcxitems file, I'm assuming that I shouldn't go cherrypick the change).~~ (edit: this change is actually in #1721 now)
- `std::ifstream` has been modified in v143 to be a little more explicit about what its constructor would accept (see https://github.com/microsoft/STL/pull/1968 and the linked C++ spec). As such, several of the usages throughout the project had to be changed to be more explicit (using `TestCommon::TestDataFile.GetPath()` explicitly, and in one instance converting a `string_view` to a string). These conversions were happening implicitly before if I understand the spec correctly.
- I had to override a warning ([C26449](https://docs.microsoft.com/en-us/cpp/code-quality/c26449?view=msvc-170)) in Downloader.cpp, as the usages of `AICLI_LOG()` were making it cranky. I think it was because the string literals were being converted and so the compiler couldn't 100% guarantee they would still exist? I'm not 100% sure, but if there's a way to actually fix it then I can do it. 

If this PR is too big, I can try to move the Valijson changes out (but I couldn't get the project to build without the update, so that would have to go in first). I know review time is limited in this part of the year, so if there's not bandwidth now, I can write down what I did and bring it back sometime later (the project configuration changes are prone to merge conflicts, so I don't know if this could sit in a draft for too too long). 

Tested: Ran all tests and builds under Visual Studio 2022 (with workloads configured via .vsconfig file) and Visual Studio 2019 (none of the changes made any difference to v142). 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1720)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
@@ -163,7 +163,7 @@ namespace AppInstaller::CLI::Workflow
 
     void ReadImportFile(Execution::Context& context)
     {
-        std::ifstream importFile{ context.Args.GetArg(Execution::Args::Type::ImportFile) };
+        std::ifstream importFile(std::string(context.Args.GetArg(Execution::Args::Type::ImportFile)));
         THROW_LAST_ERROR_IF(importFile.fail());
 
         Json::Value jsonRoot;
--- a/src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
@@ -163,7 +163,7 @@ namespace AppInstaller::CLI::Workflow
 
     void ReadImportFile(Execution::Context& context)
     {
-        std::ifstream importFile(std::string(context.Args.GetArg(Execution::Args::Type::ImportFile)));
+        std::ifstream importFile(Utility::ConvertToUTF16(context.Args.GetArg(Execution::Args::Type::ImportFile)));
         THROW_LAST_ERROR_IF(importFile.fail());
 
         Json::Value jsonRoot;
--- a/src/AppInstallerCommonCore/Downloader.cpp
+++ b/src/AppInstallerCommonCore/Downloader.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+
 #include "pch.h"
 #include "Public/AppInstallerErrors.h"
 #include "Public/AppInstallerRuntime.h"
@@ -22,6 +23,9 @@ namespace AppInstaller::Utility
         IProgressCallback& progress,
         bool computeHash)
     {
+        // For AICLI_LOG usages with string literals.
+        #pragma warning(disable:26449)
+
         AICLI_LOG(Core, Info, << "WinINet downloading from url: " << url);
 
         wil::unique_hinternet session(InternetOpenA(
@@ -125,6 +129,8 @@ namespace AppInstaller::Utility
         }
 
         AICLI_LOG(Core, Info, << "Download completed.");
+
+        #pragma warning(disable:26449)
 
         return result;
     }
--- a/src/AppInstallerCommonCore/Downloader.cpp
+++ b/src/AppInstallerCommonCore/Downloader.cpp
@@ -24,6 +24,7 @@ namespace AppInstaller::Utility
         bool computeHash)
     {
         // For AICLI_LOG usages with string literals.
+        #pragma warning(push)
         #pragma warning(disable:26449)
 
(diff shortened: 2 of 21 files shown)


JohnMcPMS:
First, thanks for what was likely a lot of tedious work.

I would like to keep the subtree pulls as separate PRs.  In order for them to work correctly we have to not squash them.  I'm not sure that it was done in the past for Valijson, but even if it wasn't we can make it be so now.  I'm assuming that it all still works under 2019 and so pulling in 0.6 should be ok without the rest of your changes.

Honestly we should take a look at all of our subtrees and see if we should pull in updates.  So you can either wait for us to do that or split Valijson out by itself.  If you want to wait, we can prioritize Valijson first.

denelon:
Wow! let me echo John's comment of thank you for all of that effort!

jedieaston:
>  I'm assuming that it all still works under 2019 and so pulling in 0.6 should be ok without the rest of your changes.

It works fine under 2019, so I can update the subtree in another PR, and once that gets merged I should be able to merge that change into this branch so it only contains the other changes. I'll mark this as draft until that's done.

jedieaston:
I rebased the branch on the current master commit (now that Valijson is updated). Much cleaner now :)

src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -22,6 +23,9 @@ namespace AppInstaller::Utility
         IProgressCallback& progress,
         bool computeHash)
     {
+        // For AICLI_LOG usages with string literals.
+        #pragma warning(disable:26449)
```
JohnMcPMS, line 27: Do you know which line is causing this?  I would just like to understand it better.

Also the typical pattern would be:
```
#pragma warning(push)
#pragma warning(disable:26449)
<code>
#pragma warning(pop)
```

jedieaston: It's being caused by the usages on lines 61, 65, 77, 93, 125 (in Downloader.cpp). I can't tell what those lines have in common that 30 does not.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -163,7 +163,7 @@ namespace AppInstaller::CLI::Workflow
 
     void ReadImportFile(Execution::Context& context)
     {
-        std::ifstream importFile{ context.Args.GetArg(Execution::Args::Type::ImportFile) };
+        std::ifstream importFile(std::string(context.Args.GetArg(Execution::Args::Type::ImportFile)));
```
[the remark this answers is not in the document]
jedieaston: I made that change, and am happy to do the test data files too, but since the ifstreams in the tests are being constructed with `std::filesystem::path` do they still need to be converted to UTF16?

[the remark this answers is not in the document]
JohnMcPMS: Ok, it looks like there would only be one place that is likely to need updating, and that is really only used by the tests on the build agent.  The default scenario should work already.  No need to bother with it unless an issue actually arises.

src/AppInstallerCommonCore/Downloader.cpp
JohnMcPMS: Hmm, feels like an erroneous warning if it is on the `AICLI_LOG` lines.  There is a `string` -> `string_view` function call in the macro, but the docs page for 26449 says it shouldn't trigger on those.  Strange...

JohnMcPMS (approved):

JohnMcPMS:
@jedieaston , I will merge when the build finishes and you approve / comment that you are done.

jedieaston:
Everything looks good to me. Assuming everything passes, I think this is ready to merge. 

Thanks for your help!

PR #1722, opened 2021-11-18
Change download file name to match the installer URL

JohnMcPMS:
## Change
Change the installer download file from:
```
%TEMP%\WinGet\<ID>.<Version>.<Installer Extension>
```
to
```
%TEMP%\WinGet\<ID>.<Version>\<Installer URL File Stem>.<Installer Extension>
```

Example:
```
"%TEMP%\WinGet\Notepad++.Notepad++.8.1.9.1.exe"
```
becomes
```
"%TEMP%\WinGet\Notepad++.Notepad++.8.1.9.1\npp.8.1.9.1.Installer.x64.exe"
```

In addition, this change uses:
```
%TEMP%\WinGet\<ID>.<Version>\<Installer SHA256 hash>
```
as the name of the file before hash validation.  This prevents the potential for any user input to create an executable file name since the SHA256 hash must be a hexadecimal string.

## Validation
Existing regression tests and manual runs to ensure consistency for golden path.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1722)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
@@ -20,8 +20,22 @@ namespace AppInstaller::CLI::Workflow
         std::filesystem::path GetInstallerBaseDownloadPath(Execution::Context& context)
         {
             const auto& manifest = context.Get<Execution::Data::Manifest>();
+
             std::filesystem::path tempInstallerPath = Runtime::GetPathTo(Runtime::PathName::Temp);
             tempInstallerPath /= Utility::ConvertToUTF16(manifest.Id + '.' + manifest.Version);
+
+            // Get file name from download URI
+            std::filesystem::path filename = GetFileNameFromURI(context.Get<Execution::Data::Installer>()->Url);
+
+            // Assuming that we find a stem value in the URI, use it
+            if (filename.has_stem())
+            {
+                std::filesystem::create_directories(tempInstallerPath);
+
+                // Name the file the same as it is in the download URI
+                tempInstallerPath /= filename.stem();
+            }
+
             return tempInstallerPath;
         }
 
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
```diff
@@ -20,8 +20,22 @@ namespace AppInstaller::CLI::Workflow
...
             const auto& manifest = context.Get<Execution::Data::Manifest>();
+
             std::filesystem::path tempInstallerPath = Runtime::GetPathTo(Runtime::PathName::Temp);
             tempInstallerPath /= Utility::ConvertToUTF16(manifest.Id + '.' + manifest.Version);
+
+            // Get file name from download URI
+            std::filesystem::path filename = GetFileNameFromURI(context.Get<Execution::Data::Installer>()->Url);
+
+            // Assuming that we find a stem value in the URI, use it
+            if (filename.has_stem())
```
yao-msft, line 31: >filename [](http://example.com/codeflow?start=16&length=8)

If the filename happens to be ".exe", has_stem() will return true and stem() will just return ".exe", I think that's not what we wanted?

JohnMcPMS: Are you concerned about the "it already has an executable extension" or the "that is a weird file name" or the "we will create an even weirder file name"?

I don't think we care about weird file names; this is more in line with the name that the developer created (although maybe not perfectly).

As to the file having an executable extension before being hashed, you could already do that by setting `PackageVersion: exe`.  If we want to truly avoid that then we will have to change the code a bit more to use the hash as the temp file name.  But then the "rename with extension" code will have to generate the whole name.  In any case I see it as defense in depth of the general case, not a true security barrier to a dedicated attacker.  But if you feel strongly we can discuss it more.

yao-msft: I don't feel strongly about the concern. It just does not match what the comment says at the top of the method. So just wanted to confirm we are ok with the less common cases.

yao-msft (approved):

yao-msft:
Thanks for fixing! Looks good to me.

PR #1727, opened 2021-11-18
Added settings to select preferred/required architecture.

### LINKED ISSUE #906: Specify installer architecture

jedieaston:

-----

Resolves #906.

This is the other half of #906, adding settings to select a preferred or required architecture. The required architecture behaves exactly like the `--architecture` flag, where `AllowedArchitectures` only contains the architecture that is in the setting. The preferred architecture behaves much like the preferred scope setting, where the installer entry with the preferred architecture is selected over all others, but if there isn't an installer entry with the preferred architecture, the normal rationale is used.

The fields are validated by determining if the architecture selected is compatible with the current system. By default the setting is set to `neutral`. (although if there is a reason we need to allow users to require neutral manifests, that could just as easily be `unknown`. I figured there were few scenarios where a user wouldn't want a installer that only had the relevant bits for their system over a neutral one with all of the bits.)

The only other note I have is that I decided to make these just strings instead of arrays since that's what it sounded like in the original ticket. If there's a usecase where someone wants to prefer multiple architectures, that's a quick fix (although there's only ever at most three options, so I doubt that comes up much?).

Tested: manually. I can write unit tests if necessary. 

image

Edit: also, sorry for the big diff on UserSettings.cpp. Visual Studio got upset that the line endings were mixed. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1727)

### DIFF
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -55,17 +55,21 @@
           "minItems": 1,
           "maxItems": 10
         },
-        "architecture": {
-          "description": "The architecture for a package install",
-          "type": "string",
-          "enum": [
-            "neutral",
-            "x64",
-            "x86",
-            "arm64",
-            "arm"
-          ],
-          "default": "neutral" 
+        "architectures": {
+          "description": "The architecture(s) to use for a package install",
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
+              "neutral",
+              "x64",
+              "x86",
+              "arm64",
+              "arm"
+            ],
+            "minItems": 1,
+            "maxItems": 4
+          }
         }
       }
     },
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -54,6 +54,18 @@
           },
           "minItems": 1,
           "maxItems": 10
+        },
+        "architecture": {
+          "description": "The architecture for a package install",
+          "type": "string",
+          "enum": [
+            "neutral",
+            "x64",
+            "x86",
+            "arm64",
+            "arm"
+          ],
+          "default": "neutral" 
         }
       }
     },
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -59,6 +59,7 @@
           "description": "The architecture(s) to use for a package install",
           "type": "array",
           "items": {
+            "uniqueItems": "true",
             "type": "string",
             "enum": [
               "neutral",
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -46,11 +46,10 @@ namespace AppInstaller::CLI::Workflow
 
         struct MachineArchitectureComparator : public details::ComparisonField
         {
-            MachineArchitectureComparator(Utility::Architecture preference) :
-        	details::ComparisonField("Machine Architecture"), m_preference(preference) {}
+            MachineArchitectureComparator() : details::ComparisonField("Machine Architecture") {}
 
-            MachineArchitectureComparator(std::vector<Utility::Architecture> allowedArchitectures, Utility::Architecture preference) :
-                details::ComparisonField("Machine Architecture"), m_allowedArchitectures(std::move(allowedArchitectures)), m_preference(preference)
+            MachineArchitectureComparator(std::vector<Utility::Architecture> allowedArchitectures) :
+                details::ComparisonField("Machine Architecture"), m_allowedArchitectures(std::move(allowedArchitectures))
             {
                 AICLI_LOG(CLI, Verbose, << "Architecture Comparator created with allowed architectures: " << GetAllowedArchitecturesString());
             }
@@ -58,7 +57,6 @@ namespace AppInstaller::CLI::Workflow
             // TODO: At some point we can do better about matching the currently installed architecture
             static std::unique_ptr<MachineArchitectureComparator> Create(const Execution::Context& context, const Repository::IPackageVersion::Metadata&)
             {
-                Utility::Architecture preference = Settings::User().Get<Settings::Setting::InstallArchitecturePreference>();
                 if (context.Contains(Execution::Data::AllowedArchitectures))
                 {
                     const std::vector<Utility::Architecture>& allowedArchitectures = context.Get<Execution::Data::AllowedArchitectures>();
@@ -97,11 +95,11 @@ namespace AppInstaller::CLI::Workflow
                             }
                         }
 
-                        return std::make_unique<MachineArchitectureComparator>(std::move(result), preference);
+                        return std::make_unique<MachineArchitectureComparator>(std::move(result));
                     }
                 }
 
-                return std::make_unique<MachineArchitectureComparator>(preference);
+                return std::make_unique<MachineArchitectureComparator>();
             }
 
             InapplicabilityFlags IsApplicable(const Manifest::ManifestInstaller& installer) override
@@ -131,10 +129,6 @@ namespace AppInstaller::CLI::Workflow
 
             bool IsFirstBetter(const Manifest::ManifestInstaller& first, const Manifest::ManifestInstaller& second) override
             {
-                if (m_preference != Utility::Architecture::Neutral)
-                {
-                    return (first.Arch == m_preference && second.Arch != m_preference);
-                }
                 auto arch1 = CheckAllowedArchitecture(first.Arch);
                 auto arch2 = CheckAllowedArchitecture(second.Arch);
 
@@ -176,7 +170,6 @@ namespace AppInstaller::CLI::Workflow
             }
 
             std::vector<Utility::Architecture> m_allowedArchitectures;
-            Utility::Architecture m_preference;
         };
 
         struct InstalledTypeComparator : public details::ComparisonField
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -46,10 +46,11 @@ namespace AppInstaller::CLI::Workflow
 
         struct MachineArchitectureComparator : public details::ComparisonField
         {
-            MachineArchitectureComparator() : details::ComparisonField("Machine Architecture") {}
+            MachineArchitectureComparator(Utility::Architecture preference) :
+        	details::ComparisonField("Machine Architecture"), m_preference(preference) {}
 
-            MachineArchitectureComparator(std::vector<Utility::Architecture> allowedArchitectures) :
-                details::ComparisonField("Machine Architecture"), m_allowedArchitectures(std::move(allowedArchitectures))
+            MachineArchitectureComparator(std::vector<Utility::Architecture> allowedArchitectures, Utility::Architecture preference) :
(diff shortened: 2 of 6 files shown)


schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -54,6 +54,18 @@
           },
           "minItems": 1,
           "maxItems": 10
+        },
+        "architecture": {
+          "description": "The architecture for a package install",
+          "type": "string",
```
yao-msft, line 60: Shall we consider making it a list to be more flexible(like local above) ?

jedieaston: The ticket made it sound like you’d select a preferred (single) architecture instead of multiple, but I can do that.

denelon: I was also thinking we would have a "single" preferred or required architecture. I can't think of a valid scenario where there would be three architectures that could all work and you would prefer (a), accept (b), and reject (c). With locale, I could see a set you would like in some kind of order so you could essentially have a bit of control when one of the languages you used was available, but it wasn't a logical close match otherwise. Let's keep this as a singleton for now.

denelon: Just after I hit send, I remembered some emulation might be valid, but I think the "requires" forces it to be what I want and "prefers" will take anything else.

jedieaston: That is how it works. Prefers will fallback to the normal logic if the architecture you prefer is not available. If your required architecture is not available you get an error.

denelon: Yeah, I was just not completing thoughts last night. Days have bee feeling like weeks lately. I'm trying to close a lot of gaps. We have some heavy investments coming in related to Intune integration for the replacement of store for business, and Win32 support for the store. The next couple of quarters are going to be busy.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -46,17 +46,19 @@ namespace AppInstaller::CLI::Workflow
 
         struct MachineArchitectureComparator : public details::ComparisonField
         {
-            MachineArchitectureComparator() : details::ComparisonField("Machine Architecture") {}
+            MachineArchitectureComparator(Utility::Architecture preference) :
+        	details::ComparisonField("Machine Architecture"), m_preference(preference) {}
 
-            MachineArchitectureComparator(std::vector<Utility::Architecture> allowedArchitectures) :
-                details::ComparisonField("Machine Architecture"), m_allowedArchitectures(std::move(allowedArchitectures))
+            MachineArchitectureComparator(std::vector<Utility::Architecture> allowedArchitectures, Utility::Architecture preference) :
```
JohnMcPMS, line 52: This shouldn't have needed to change to fully support preferences and requirements.  An entry of `Utility::Architecture::Unknown` in `Execution::Data::AllowedArchitectures` indicates "and all the rest"; aka that the previous architectures are preferences.  Thus the only difference between the preferences and requirements settings are whether to append `Unknown`.

schemas/JSON/settings/settings.schema.0.2.json
JohnMcPMS: I would rather have it be an array to start.  At some point someone will want an array and then we will have two settings (or have to implement it allowing both scalar and array).  Even requirements makes sense to support multiple, just like locales.  I might be an ARM purist on my ARM64 machine and so require `{ arm64, arm }`.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -962,14 +962,20 @@ namespace AppInstaller::CLI::Workflow
...
         if (isUpdate)
         {
             installationMetadata = context.Get<Execution::Data::InstalledPackageVersion>()->GetMetadata();
         }
         if (context.Args.Contains(Execution::Args::Type::InstallArchitecture))
         {
+            // arguments override settings.
             context.Add<Execution::Data::AllowedArchitectures>({ Utility::ConvertToArchitectureEnum(std::string(context.Args.GetArg(Execution::Args::Type::InstallArchitecture))) });
         }
+        else if (requiredArchitecture != Utility::Architecture::Neutral)
```
JohnMcPMS, line 975: I would make this an unconditional `else` and the check the settings (both once you revert the `ManifestComparator` changes) inside it.  That way we don't do unnecessary work getting the setting(s) if there is an argument passed in.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -1,334 +1,348 @@
...
+        WINGET_VALIDATE_PASS_THROUGH(EFExperimentalCmd)
+        WINGET_VALIDATE_PASS_THROUGH(EFExperimentalArg)
+        WINGET_VALIDATE_PASS_THROUGH(EFDependencies)
+        WINGET_VALIDATE_PASS_THROUGH(TelemetryDisable)
+        WINGET_VALIDATE_PASS_THROUGH(EFDirectMSI)
+        WINGET_VALIDATE_PASS_THROUGH(EnableSelfInitiatedMinidump)
+        WINGET_VALIDATE_SIGNATURE(InstallArchitecturePreference)
+        {
+            Utility::Architecture arch = Utility::ConvertToArchitectureEnum(value);
+	        if (Utility::IsApplicableArchitecture(arch) != Utility::InapplicableArchitecture)
```
JohnMcPMS, line 236: You have tabs here; please convert to 4 spaces.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -1,334 +1,348 @@
...
-            return {};
-        }
-
-        WINGET_VALIDATE_PASS_THROUGH(EFExperimentalCmd)
-        WINGET_VALIDATE_PASS_THROUGH(EFExperimentalArg)
-        WINGET_VALIDATE_PASS_THROUGH(EFDependencies)
-        WINGET_VALIDATE_PASS_THROUGH(TelemetryDisable)
-        WINGET_VALIDATE_PASS_THROUGH(EFDirectMSI)
-        WINGET_VALIDATE_PASS_THROUGH(EnableSelfInitiatedMinidump)
-
```
JohnMcPMS, line 231: What did whitespace do to you?

JohnMcPMS (changes requested):

jedieaston:
I fixed all of the complaints (array is used, ManifestComparator isn't messed with, whitespace fixed). All of the changes to `AllowedArchitectures` are now dealt with in `SelectInstaller`. 

Sorry for the delay, I was traveling for Thanksgiving.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -962,7 +962,8 @@ namespace AppInstaller::CLI::Workflow
         bool isUpdate = WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerExecutionUseUpdate);
 
         IPackageVersion::Metadata installationMetadata;
-        Utility::Architecture requiredArchitecture = Settings::User().Get<Settings::Setting::InstallArchitectureRequirement>();
+        std::vector<Utility::Architecture> requiredArchitectures = Settings::User().Get<Settings::Setting::InstallArchitectureRequirement>();
```
JohnMcPMS, line 965: I meant that these settings should only be retrieved in the `else` that needs to read them, thus preventing the work to get them if an argument was given.  Minor optimization, but also an easy one.  I'll stop commenting on it though, so up to you.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -962,14 +962,34 @@ namespace AppInstaller::CLI::Workflow
...
         }
         if (context.Args.Contains(Execution::Args::Type::InstallArchitecture))
         {
+            // arguments override settings.
             context.Add<Execution::Data::AllowedArchitectures>({ Utility::ConvertToArchitectureEnum(std::string(context.Args.GetArg(Execution::Args::Type::InstallArchitecture))) });
         }
+        else 
+        {
+            std::set<Utility::Architecture> settingArchitectures;
+            if (!optionalArchitectures.empty())
```
JohnMcPMS, line 979: I think the logic here should be:
```
if required exists
    use required
else if preferences exists
    use preferences + Unknown
else
    don't set AllowedArchitectures
```
There shouldn't be a need to combine requirements and preferences.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
jedieaston: I was thinking that if you wanted only arm or arm64, but of those you may prefer arm64, that may be a reason to have both. But that’s probably taken care of by the existing logic for selecting architectures.

jedieaston:
I noticed something when working on that last commit, which is that adding the Unknown architecture to AllowedArchitectures makes the ManifestComparator use its traditional logic instead of accepting the user's preference above all else:

settings.json:
```
{
    "$schema": "https://aka.ms/winget-settings.schema.json",

    // For documentation on these settings, see: https://aka.ms/winget-settings
    // "source": {
    //    "autoUpdateIntervalInMinutes": 5
    // },
    "installBehavior": {
        "preferences": {
            "architectures": ["x86"]
        }
    },
}
```

Relevant logs from `wingetdev install Google.Chrome`

```
2021-11-30 09:27:59.146 [YAML] Detected UTF-8
2021-11-30 09:27:59.153 [CLI ] Manifest fields: Name [Google Chrome], Version [96.0.4664.45]
2021-11-30 09:28:39.117 [CLI ] Locale Comparator created with Required Locales: [] , Preferred Locales: [en-US]
2021-11-30 09:28:39.117 [CLI ] Architecture Comparator created with allowed architectures: X64, X86, Neutral
2021-11-30 09:28:39.117 [CLI ] Starting installer selection.
2021-11-30 09:28:39.117 [CLI ] Installer [X64,msi,Machine,] is current best choice
2021-11-30 09:28:39.117 [CLI ] Installer [X64,msi,Machine,] is better than [X86,msi,Machine,] due to: Machine Architecture
2021-11-30 09:28:39.117 [CLI ] Completed installer selection.
```

Am I missing something here or do I still need to do something to ManifestComparator to make sure that the preferences are honored above the traditional logic?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
jedieaston: I moved it, that makes more sense.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
JohnMcPMS: The order of the requirements is the order of preference as well, which should take care of that.

JohnMcPMS:
> Am I missing something here or do I still need to do something to ManifestComparator to make sure that the preferences are honored above the traditional logic?

Are you putting the `Unknown` value at the back of the vector?  I would double check in the allowed architectures vector coming in is `{ x86, Unknown }`.

jedieaston:
That was exactly it, thanks! Now it's using emplace_back, and all is right with the world.

JohnMcPMS (approved):

PR #1731, opened 2021-11-19
Update source agreement market message to be more clear

yao-msft:
There's been feedback item filed claiming the source agreement market message is not clear enough so making explicit that we are sending the geographic region info to the source's backend server.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1731)

### DIFF
--- a/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
+++ b/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
@@ -1098,7 +1098,7 @@ Do you agree to the terms?</value>
     <value>One or more of the source agreements were not agreed to. Operation cancelled. Please accept the source agreements or remove the corresponding sources.</value>
   </data>
   <data name="SourceAgreementsMarketMessage" xml:space="preserve">
-    <value>The source requires current machine's geographic region to be sent to the backend service to function properly.</value>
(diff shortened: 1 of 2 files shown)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1098,7 +1098,7 @@ Do you agree to the terms?</value>
     <value>One or more of the source agreements were not agreed to. Operation cancelled. Please accept the source agreements or remove the corresponding sources.</value>
   </data>
   <data name="SourceAgreementsMarketMessage" xml:space="preserve">
-    <value>The source requires current machine's geographic region to be sent to function properly.</value>
+    <value>The source requires current machine's geographic region to be sent to the backend service to function properly.</value>
```
JohnMcPMS, line 1101: Should this include an example or explanation of the amount of data being sent?

```suggestion
    <value>The source requires the current machine's 2-letter geographic region to be sent to the backend service to function properly (ex. "US").</value>
```

JohnMcPMS (approved):
Is there an issue that can be linked against?  I'm good with this but I was hoping that whomever raised the issue could also approve it.  If it was only feedback, then I guess we just have to hope this is satisfactory.

yao-msft:
It's feedback item, not a github issue.

Issue #147, opened 2020-05-15
Support different release channels [released, beta, alpha]


denelon:
I'd like to be able to use the package manager for different channels of software. This could be beta, or unstable branches, or other kinds of constructs.

Semantic versioning (https://semver.org) has a specification we may follow for pre-release software.
Microsoft has channels.
Many open source projects have branches.

>Note: Some packages support "side by side" install where a stable and a pre-release version may both be on a system at the same time. These are not considered release channels in the Windows Package Manager, but they are treated as different packages. Release channels are considered in the context of the Windows Package Manager for packages that do not support "side by side" installation.

utybo:
Java (AdoptOpenJdk) is a good example for this. There are two LTS releases at the moment (versions 8 and 11) and the current release is 14. This feature would be useful for having at least the option to install version 8 or 11.

It might also be possible to  just have different packages (e.g. openjdk8 and openjdk11, that's what Ubuntu is using I believe) but that feels less clean.

ocdtrekkie:
The ideal scenario (which is familiar to Linux folks, I believe), is that `openjdk` or `openjdk-latest` should install the newest supported release, and `openjdk-8` or `openjdk-11` or the like should be maintained packages respectively, so that people can script opening the release version they need.

The best solution, I think, is to require major version numbers for packages with multiple supported releases, and then winget needs to support "shortcuts", such that running `winget install openjdk` is understood to install `openjdk-14`. And when a newer OpenJDK package comes out, that shortcut can be pointed to the newer release, without breaking users who are explicitly trying to install openjdk-14.

cnshenj:
Chocolatey: `--version`
NuGet: `-Version`
Docker: `name:tag`
npm: `name@tag`

Personally I like the npm format.

rashil2000:
Another example of this would be the Node.js package. Many times the recommended version is `nodejs-lts`, but when installing Node through winget, it installs the current (latest) release, which is not very widely supported. The same is the case with .Net Core. The default installation version should be the Long Term Support version, or we should have some way of specifying it, instead of manually feeding the version numbers every time.

KorbenDev:
I want to support this. Today my script tried to download and install a 7Zip alpha(!) version. First I want to question the need to distribute alpha/beta versions over package management in general. I don't think that's a good idea.
And if you really want to make alpha/beta state applications available, they should be tagged very clearly!
[...]

PR #1742, opened 2021-11-23
Spec #147 - Support different release channels [released, beta, alpha]

Trenly:
- [x] Are you working against an Issue? See #147 

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1742)

### DIFF
--- a/doc/specs/#147 - Support different release channels [released, beta, alpha].md
+++ b/doc/specs/#147 - Support different release channels [released, beta, alpha].md
@@ -24,7 +24,7 @@ The Windows Package Manager manifest v1.1.0 schema has provided a key for declar
 > The `Channel` key should be removed from the manifest v1.1.0 schema.
 
 Instead, the use of multiple installer manifest files will be used to define and manage the various channels available for a package in a way similar to locales. This will ensure that the client remains backwards compatible with old manifest versions, and will make the creation and maintenance of the various channels easier for contributors in the community repository. This will require a key be added to the  manifest schema. Channels should not be compatible with singleton manifests.
-> The `DefaultChannel` key should be added to the manivest v1.1.0 schema for version manifests
+> The `DefaultChannel` key should be added to the manifest v1.1.0 schema for version manifests
 
 An example of the manifest file structure, with installer channel manifests:
 ```raw
--- a/doc/specs/#147 - Support different release channels [released, beta, alpha].md
+++ b/doc/specs/#147 - Support different release channels [released, beta, alpha].md
@@ -1,5 +1,5 @@
 ---
-author: Kaleb Luedtke Trenly/trenlymc@gmail.com
+author: Kaleb Luedtke - Trenly <trenlymc@gmail.com>
 created on: 2021-11-18
 last updated: 2021-11-20
 issue id: 147
@@ -15,14 +15,16 @@ Several packages have different channels for releasing different versions of the
 
 ## Inspiration
 
-Some other package managers have this capability. In the current state, a new package identifier must be created for each release channel which can cause conflicts when the versions are not side-by-side compatible.
+Some other package managers have this capability. In the current state, a new package identifier must be created for each release channel which can cause conflicts when the versions are *not* side-by-side compatible.
 
 ## Solution Design
 
 ### Manifest Structure
-The Windows Package Manager manifest v1.1.0 schema has provided a key for declaring which channel an installer belongs to. However, this has the potential to make the installer manifests extremely complex when there are multiple release channels with multiple installers each. This key should be removed from the manifest v1.1.0 schema.
+The Windows Package Manager manifest v1.1.0 schema has provided a key for declaring which channel an installer belongs to. However, this has the potential to make the installer manifests extremely complex when there are multiple release channels with multiple installers each.
+> The `Channel` key should be removed from the manifest v1.1.0 schema.
 
-Instead, the use of multiple installer manifest files will be used to define and manage the various channels available for a package in a way similar to locales. This will ensure that the client remains backwards compatible with old manifest versions, and will make the creation and maintenance of the various channels easier for contributors in the community repository. This will require a `DefaultChannel` key be added to the version manifest schema. Channels shold not be compatible with singleton manifests.
+Instead, the use of multiple installer manifest files will be used to define and manage the various channels available for a package in a way similar to locales. This will ensure that the client remains backwards compatible with old manifest versions, and will make the creation and maintenance of the various channels easier for contributors in the community repository. This will require a key be added to the  manifest schema. Channels should not be compatible with singleton manifests.
+> The `DefaultChannel` key should be added to the manivest v1.1.0 schema for version manifests
 
 An example of the manifest file structure, with installer channel manifests:
 ```raw
@@ -58,9 +60,9 @@ An example of the manifest file structure when `DefaultChannel` is unset -
 ```
 
 ### CLI Behavior
-Where not specified by the user, the default release channel should be used/shown first or exclusively, then other channels in ASCII sort order. If the package manifest has a default channel specified and no versions of that package have an installer manifest matching the default channel, then the next available channel should be used/shown. Channels should be treated as case-insensitive.
+Where not specified by the user, the default release channel should be used/shown first or exclusively, then other channels in ASCII sort order. If the package manifest has a default channel specified and no versions of that package have an installer manifest matching the default channel, then the next available channel should be used/shown. Channels should be treated as *case-insensitive*.
 
-The channel `Default` shall be treated as a reserved token. When used in commands, default should refer to the default channel specified in the version manifest. In command output, when the default channel is named, the name should be shown. If the default channel is null, then `Default` shold be shown.
+The channel `Default` shall be treated as a reserved token. When used in commands, default should refer to the default channel specified in the version manifest. In command output, when the default channel is named, the name should be shown. If the default channel is null, then `Default` should be shown. Manifest validation should throw an error when `DefaultChannel` is `Default`
 
 #### winget search
 In order to avoid package duplication in search results, only the default channel should be shown unless otherwise specified. Users should be able to search for packages by release channel in addition to all other search filters.
@@ -79,14 +81,35 @@ When the user specifies `--versions`, the versions for all channels along with t
 When the user specifies `--channels`, the available channels for the package should be shown. If a package contains a named default channel and unnamed (null) channels, only the named channels should be shown
 
 #### winget install
-[Comment] Install should support --channel
-[Comment] Install should keep track of which channel was installed
+When the user does not specify `--channel`, the default channel should be used. If the user does specify `--channel`, the specified channel should be used. The latest version for the channel should be installed, unless a user specified the version using the `--version` parameter. Additionally, the Windows Package Manager should record the package channel in some manner to ensure that upgrades to the package are found using the correct channel.
+
+Optionally, the user should be able to specify the channel as part of the package identifier using the `@` symbol. For example, `winget install Google.Chrome@dev` would be equivalent to `winget install Google.Chrome --channel dev`
 
 #### winget upgrade
-[Comment] If we can't determine channel, ask user. If --all was specified, throw a warning and skip the upgrade
+When listing the packages available for upgrade, the Windows Package Manger should look only for packages matching the release channel of the installed package. 
+When installing packages available for upgrade, the Windows Package Manager should only install new versions matching the release channel of the installed package.
+
+If the package channel cannot be determined, there are several options. 
+1) Ask the user for input.
+2) Throw a warning and skip that package
+3) Use the default channel for that package
+
+The Windows Package Manager should never leave a package in a state where it cannot be upgraded simply because the package is unknown. To account for this, there must be some way for the user to specify a release channel. Therefore, when `winget upgrade --all` is used, and the release channel cannot be determined, a warning should be shown and the package should be skipped. For all other cases, the user should be asked to select a release channel on a per-package basis; the user's choice for that package should persist.
 
 #### winget export/import
-[Comment] When exporting, specify the release channel. When importing, honor the release channel. If release channel cannot be honored on import, use default release channel
+When exporting the list of installed packages, where possible, the release channel should be included with the export.
+When importing the list of installed packages, where possible, the release channel should be honored. If the release channel is not able to be honored, a warning should be thrown, and the default release channel should be used
+
+### Settings
+The user should be able to specify a set of preferred release channels in their settings file. The `channel` parameter should be added under `installBehavior`. The matching parameter is `--channel`. The order of the preferences should be considered when installing packages such that the first value is the highest preference.
+```json
+"installBehavior": {
+        "preferences": {
+            "locale": [ "en-US", "fr-FR" ],
+            "channel": ["alpha","beta","dev","canary","nightly"]
+        }
+    }
+```
 
 ## UI/UX Design
 ### Searching for package version with release channel
@@ -192,7 +215,7 @@ Google Chrome                           Google.Chrome                           
 
 ## Capabilities
 
-Many packages use different structures for release channels. The Windows Package Manager should acommodate as many of these as is practicable.
+Many packages use different structures for release channels. The Windows Package Manager should accommodate as many of these as is practicable.
 
 ### Accessibility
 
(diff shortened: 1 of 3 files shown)


doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+  | Google.Chrome.installer.Stable.yaml
+  | Google.Chrome.locale.en-US.yaml
+  | Google.Chrome.locale.nb-NO.yaml
+  | Google.Chrome.yaml
+/manifests/g/Google/Chrome/2.0
+  | Google.Chrome.installer.canary.yaml
+  | Google.Crome.installer.dev.yaml
+  | Google.Chrome.locale.en-US.yaml
+  | Google.Chrome.locale.nb-NO.yaml
+  | Google.Chrome.yaml
```
Comment (jedieaston) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:44]: The problem I have with this is that it relies on the folder/file naming setup of the community repo, which the client is not aware of (since it is designed to work with any source). How would channels be expressed without a schema field in REST sources, or other pre-indexed sources?

I think we keep the channel field in the schema, and add the `DefaultChannel` to the version schema. The channel can be given per installer entry, which I'm aware will get unwieldy (but hopefully we don't have to use it often enough for it to be a problem).

jedieaston (commented):
Thanks for writing this up :D

I think we need to consider what these changes mean for other repos, like the Microsoft Store or "bob's homegrown organic REST source".

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:44]: If I’m not mistaken, for other sources, the channel would still be included in the package schema. Since the package schema is different than the manifest schema, I believe this would still be supported?

Either that, or I would suggest having it both ways. That way the community repo and other sources  could still be kept relatively organized by taking advantage of splitting the manifests, but other sources which aren’t able to could still support channels

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+
+Some other package managers have this capability. In the current state, a new package identifier must be created for each release channel which can cause conflicts when the versions are *not* side-by-side compatible.
+
+## Solution Design
+
+### Manifest Structure
+The Windows Package Manager manifest v1.1.0 schema has provided a key for declaring which channel an installer belongs to. However, this has the potential to make the installer manifests extremely complex when there are multiple release channels with multiple installers each.
+> The `Channel` key should be removed from the manifest v1.1.0 schema.
+
+Instead, the use of multiple installer manifest files will be used to define and manage the various channels available for a package in a way similar to locales. This will ensure that the client remains backwards compatible with old manifest versions, and will make the creation and maintenance of the various channels easier for contributors in the community repository. This will require a key be added to the  manifest schema. Channels should not be compatible with singleton manifests.
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:26]: Everything needs be compatible with any manifest representation, we just impose certain patterns on `winget-pkgs`.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (JohnMcPMS) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:44]: The original thought (pre multi-file manifests) was to have a structure like:
```
publisher\package\version
publisher\package\non-default-channel\version
```
With `Channel` being in the installer schema, your design can work too.  I think it might be a bit more confusing though, as it is harder to see what is in each channel at a glance.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+  | Google.Chrome.yaml
+/manifests/g/Google/Chrome/2.0
+  | Google.Chrome.installer.canary.yaml
+  | Google.Crome.installer.dev.yaml
+  | Google.Chrome.locale.en-US.yaml
+  | Google.Chrome.locale.nb-NO.yaml
+  | Google.Chrome.yaml
+```
+
+To maintain backwards compatability, `null` should be treated as a valid `DefaultChannel`. If the default channel is not specified, `<PackageIdentifier>.installer.yaml` would be treated as the default. Additional channels may still be added to the package, and all functionality would still be present.
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:47]: The code already expects that an empty value is the default channel.  Having a user defined default channel value will complicate things.  I don't see the value in allowing it even if it was easy either.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+PS> winget search --channel <channel> # Return all packages which have a channel matching <channel>
+PS> winget search --channel "" # Return all packages which have a null channel
+PS> winget search <query> --channel <channel> # Return packages matching <query> which have a channel matching <channel>
+PS> winget search --tag <tag> --channel <channel> # Return packages with a tag matching <tag> which have a channel matching <channel>
+...
+```
+
+#### winget show
+When the user specifies `--versions`, the versions for all channels along with the channels should be shown. The versions should be grouped by channel, sorted by channel, then sorted by version descending
+When the user specifies `--channels`, the available channels for the package should be shown. If a package contains a named default channel and unnamed (null) channels, only the named channels should be shown
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:81]: I'm not following this part, but I assert that there should be only one empty channel, and any number of non-empty channels.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+
+#### winget install
+When the user does not specify `--channel`, the default channel should be used. If the user does specify `--channel`, the specified channel should be used. The latest version for the channel should be installed, unless a user specified the version using the `--version` parameter. Additionally, the Windows Package Manager should record the package channel in some manner to ensure that upgrades to the package are found using the correct channel.
+
+Optionally, the user should be able to specify the channel as part of the package identifier using the `@` symbol. For example, `winget install Google.Chrome@dev` would be equivalent to `winget install Google.Chrome --channel dev`
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:86]: Unfortunately we didn't reserve any characters for this kind of thing.  We could still support it, but it would probably need to check for packages named `Google.Chrome` and `Google.Chrome@dev` and then continue depending on what is found.

If we do decide to go down this route, we should also consider supporting other identifiers (well, mostly one, version).  Maybe `id@channel#version`?  Then we would probably want to reserve these on `winget-pkgs` even if the client needed to handle it all.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+If the package channel cannot be determined, there are several options. 
+1) Ask the user for input.
+2) Throw a warning and skip that package
+3) Use the default channel for that package
+
+The Windows Package Manager should never leave a package in a state where it cannot be upgraded simply because the package is unknown. To account for this, there must be some way for the user to specify a release channel. Therefore, when `winget upgrade --all` is used, and the release channel cannot be determined, a warning should be shown and the package should be skipped. For all other cases, the user should be asked to select a release channel on a per-package basis; the user's choice for that package should persist.
+
+#### winget export/import
+When exporting the list of installed packages, where possible, the release channel should be included with the export.
+When importing the list of installed packages, where possible, the release channel should be honored. If the release channel is not able to be honored, a warning should be thrown, and the default release channel should be used
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:101]: I think our default stance is to fail the import without doing anything if we know up front that we can't do what is asked.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+Beta
+Canary
+Dev
+
+# When the default channel is unnamed
+PS> winget show 7zip.7zip --channels
+Found 7-zip [7zip.7zip]
+Channel
+-------
+Default
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:155]: If the default channel is always "empty", then there is no need to have a reserved value or even show it in this case.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+
+There should be no security impact directly, although we must remember that different sources may not guarantee the safety of packages. The Windows Package Manager community repository performs static and dynamic analysis, and in some cases additional manual validation before accepting a package.
+
+### Reliability
+
+The Windows Package Manager community repository has many packages that may be affected by release channels. The ease of keeping packages updated must be a consideration when designing other tools for creating and updating manifests.
+
+### Compatibility
+
+The current implementation for different release channels is to have separate package identifiers for each package. Packages which have channel-specific identifiers currently existing in the community repository /should/ be updated to use this feature, but will not be required to.
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:234]: If they do not support side-by-side installation, they really should be moved to this pattern.

If they do support it, I think that is a whole other design spec to deal with them better.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+
+* If a user installs a package outside of the Windows Package Manager, we may not be able to determine the channel. This has potential to cause conflicts if an attempt is made to upgrade applications from one channel to another.
+* A package may have a `DefaultChannel` set and there be no versions available for the specified channel. When a user runs a search or upgrade for the package, the behavior may be unpredictable unless fallback behavior is accounted for in the implementation. 
+* Some packages may use multiple different names to refer to the same channel. Example - `Release`, `LTS`, `latest`, `current`, and `production` may all refer to the same channel. It may be wise to add an additional list parameter to the schema for `ChannelAliases`. This would allow the Windows Package Manager to identify when there is a matching package even though the user specified a release channel.
+
+## Future considerations
+
+* Implementation of the shorthand channel specifier may also allow for a shorthand version specifier such as `:`. This would enable something like `winget install Google.Chrome:96.0.4664.45@dev` for quick installation of specific versions and channels
```
Comment (JohnMcPMS) [inline doc/specs/#147 - Support different release channels [released, beta, alpha].md:250]: Guess I should have read the whole thing before making my comment on versions here. 😄

doc/specs/#147 - Support different release channels [released, beta, alpha].md
```diff
@@ -0,0 +1,259 @@
...
+Several packages have different channels for releasing different versions of their software. The Windows Package Manager should be able to support packages of the same Package Identifier where the versions come from different release channels. The Windows Package Manager should be also able to inform a user about which channels are available.
+
+## Inspiration
+
+Some other package managers have this capability. In the current state, a new package identifier must be created for each release channel which can cause conflicts when the versions are *not* side-by-side compatible.
+
+## Solution Design
+
+### Manifest Structure
+The Windows Package Manager manifest v1.1.0 schema has provided a key for declaring which channel an installer belongs to. However, this has the potential to make the installer manifests extremely complex when there are multiple release channels with multiple installers each.
```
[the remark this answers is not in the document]
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:23]: I think that each installer should have a channel associated with it. The main issue I see here is that a single channel could have many many installers or just a single installer. Providing the support to have different channels as different installer manifests would remove this issue. 

The key could potentially be left as it is to allow for the case where there is one installer for each channel too

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:101]: Thats a fair stance. I’ll update this if needed

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:81]: I think this goes back to your point about whether or not a named channel could or should be allowed to be default

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:47]: > The original thought (pre multi-file manifests) was to have a structure like:
> 
> ```
> publisher\package\version
> publisher\package\non-default-channel\version
> ```
>  With Channel being in the installer schema, your design can work too. I think it might be a bit more confusing though, as it is harder to see what is in each channel at a glance.

That is the way it works currently, but based on the validation over at winget-pkgs, these would be completely separate package Identifiers. If I’m not mistaken, the whole point of channels is to combine package identifiers to be better representative of how packages are published, since some publishers do not allow side-by-side compatibility for different channels and so a user may not realize that Example.Package would be overwritten if they install Example.Package.Dev

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:234]: I think the point of channels is to have them be combined to a single identifier where possible. The packages which do support side by side are the best to have in the Example.Package / Example.Package.Dev format. It is when they do not support side by side that they should have the channel specified for the installer to ensure the proper channel is used for updates and installation

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:44]: > The original thought (pre multi-file manifests) was to have a structure like:
> 
> ```
> publisher\package\version
> publisher\package\non-default-channel\version
> ```
>  With Channel being in the installer schema, your design can work too. I think it might be a bit more confusing though, as it is harder to see what is in each channel at a glance.

That is the way it works currently, but based on the validation over at winget-pkgs, these would be completely separate package Identifiers. If I’m not mistaken, the whole point of channels is to combine package identifiers to be better representative of how packages are published, since some publishers do not allow side-by-side compatibility for different channels and so a user may not realize that Example.Package would be overwritten if they install Example.Package.Dev

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:26]: I'm not entirely sure what you mean by this comment. Is this just saying that singletons should support channels?

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:47]: One that I could think of is Ubuntu. Canonical.Ubuntu `16.04`, `18.04`, and `20.04` all have their own release cadence, but so does the un-versioned channel. In this case, `20.04` would make the most sense to be the default channel.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (jedieaston) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:26]: Manifests from any source, i.e. REST sources that don’t use the manifest schema, preindexed sources (which are generated from the YAML files but converted to a different format), and local (-m on a folder) manifests is what I think that means.

Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:26]: I see; In that case, there still shouldn’t be an issue, since the rest sources should still be able to specify the channel, pre-indexed should work as it would just require them to support building from the proposed mutiple-installer-manifests yaml structure, and local would still be handled just fine since the CLI would be able to find the channels based on the filenames or the specified value in the manifest file

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (JohnMcPMS) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:44]: We would update winget-pkgs validation to handle non-empty channels in this way so that both would still be `Example.Package`.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (JohnMcPMS) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:47]: Is the thinking that the default channel will march forward with the latest release?  So currently it would be `20.04` but when `22.04` is released it would change to that, while still leaving the `20.04` channel as is.  I see the value there; if we used an empty channel then to achieve the same thing would require moving/renaming all of the manifests.

I think it still makes sense (and I think what you are saying the spec) to allow for the empty channel to be the default default.  Then *only* allow for `DefaultChannel` when there is no empty channel.

doc/specs/#147 - Support different release channels [released, beta, alpha].md
Comment (Trenly) [reply doc/specs/#147 - Support different release channels [released, beta, alpha].md:44]: That would work too. If that is the preferred approach, I will update the spec. Just so long as we capture the work that is needed

denelon:
Would you mind updating the allow.txt for valid words and expected.txt for URL fragments and programming vernacular?

Trenly:
I'm closing this to allow MSFT to better align on a spec internally, as this feature has several complexities and is likely not in the next several client releases. To ensure that the spec stays up to date with the client, it is best for this to be closed and be used as a reference for future ideas and considerations

vedantmgoyal2009:
Why can't we just include a `Channel` key in the version manifest?

I mean WinGet finally merges the manifests, for e.g. see `https://winget.azureedge.net/cache/manifests/m/Microsoft/Teams/1.5.00.8070/d9c7-Microsoft.Teams.yaml` and note `ManifestType: merged`

It's rare that the publisher publishes the "same version" to two channels and for that, we can keep the `Channel` field as an array.

The Upgrade Cases may be complex because: `In the current state, a new package identifier must be created for each release channel which can cause conflicts when the versions are not side-by-side compatible.` For this, we can modify the existing `UpgradeBehavior` key which can be set to: 
- When the Beta version can be installed over a Stable version [`install`] + user will have to specify `--force` parameter to **look for updates** as well as **upgrade** across different channels.
- When the Beta version can be installed side-by-side with the Stable version [`installStrict`]. 
Note: When the package has an upgrade available in the Stable channel, it should not modify the installation of Beta version.
- When a Stable version has to be uninstalled before installing the Beta version [`uninstallPrevious`]

This may make the implementation of channels into winget-cli somewhat complex, but it would be easy on the winget-pkgs repository side and for users/orgs who manage their own rest-source. 

Apologies in advance if I'm misunderstanding something but I really don't understand why it has to be too complex.

Trenly:
The reason I was looking at the structure of manifests is because the CLI will have to be able to search for, and differentiate, between channels even with all other things about the package being the same

Because there is not a set way for publishers to do things, there are many cases where the only differentiator between a prerelease and an actual release is quite literally what the publisher says it is. Take winget for example, If winget were looking at the ARP entries of the PreRelease vs the Latest release, they would be identical. So, the client has to have some way to track that on install. 

This is also a reason why the channel would likely have to be an Installer property, because there are publishers who do use the same version within their channels. Google Chrome is a great example, since the version starts in Canary and then gets promoted through to stable. Now, just because it gets promoted doesn’t mean that it can’t also be the latest beta or latest stable, the same version can be in multiple channels at the same time (even with different URLS to download from)

Then, when upgrading a package, winget should choose the latest version that is within the channel. As I was thinking about how the client may accomplish this, it seemed to me that instead of loading each manifest into the index it would be easier to identify specific files. Honestly, this is partly because (at the time) I didn’t know much about the different source types and how each are created. 

However, I think there are still complexities in that sometimes a user will specify to install a version without specifying a channel, sometimes they may specify a channel and a version, and they can do all sorts of other weird things we may not expect. 

-----

All of that to say - Channels impacts the core functionality of how packages are searched for, installed, and upgraded. Not only that, but they impact almost every other feature within the client. Additionally, it impacts validation from the client all the way through the pipelines into the validation of what is installed on users machines currently.

My first draft of a spec (which you will notice I closed in favor of letting the Experts at MSFT create) was trying to account for all the complexities but still didn’t take into consideration the range of impact this feature has.

PR #1761, opened 2021-12-01
Reduce localized languages per localization team guidance

yao-msft:
## Validation
Created a localized package with languages removed, manually verified the AppInstaller ui keeps the winget removed language and winget defaults to other languages in the App Language settings list.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1761)


JohnMcPMS (approved):
Explaining a bit more:
The Microsoft localization team has determined that winget should be treated as other IT/dev tools rather than an end customer application.  Because we had been shipping with the Desktop App Installer, we were following its set of languages.  But now winget will have the smaller set, while the App Installer will keep the larger.

Issue #1692, opened 2021-11-10
winget upgrade should have the option to hide packages with Unknown version from output


MrBeeMovie:
### Description of the new feature / enhancement

winget upgrade currently shows packages with a version of Unknown as being out of date. Since the version is Unknown the package may or may not be out of date. It would be useful to have the option to hide packages with an Unknown version or allow for the option to list only unknown packages with the upgrade command defaulting to not showing these Unknown version packages.

### Proposed technical implementation details

No suggestion.

jedieaston:
Maybe this should just be a argument, `--include-unknown`, which enables unknown packages to be upgraded? By default if `winget upgrade` finds a installed app that doesn't publish version info to the ARP table it won't attempt an upgrade.

Wouldn't be too hard to implement if there's agreement.

MrBeeMovie:
> Maybe this should just be a argument, `--include-unknown`, which enables unknown packages to be upgraded? By default if `winget upgrade` finds a installed app that doesn't publish version info to the ARP table it won't attempt an upgrade.
> 
> Wouldn't be too hard to implement if there's agreement.

I would be fine with this solution.

felipecrs:
I even think that by default such packages should be ignored on `winget upgrade --all`, but still shown on `winget upgrade` so that the user is at least aware of it.

jedieaston:
Perhaps a setting? I would (personally) prefer winget not to show unknown version apps in the upgrade table by default, since 99/100 times "Unknown" ends up meaning "It updates itself", but I can understand that some people may think differently. 

It also may be a confusing user experience for a app to be listed for upgrade, then when the user wants to upgrade it and runs `winget upgrade --all` as they usually would, it doesn't get upgraded since they didn't pass the additional option. A setting that tells winget to use the current behavior (when in doubt, upgrade) that is enforced for both flows would make a little more sense to me.

MrBeeMovie:
> I even think that by default such packages should be ignored on `winget upgrade --all`, but still shown on `winget upgrade` so that the user is at least aware of it.

I disagree since the point of `winget upgrade` is to see a list of packages that have available updates. If the version is "Unknown" it is not certain whether it has an available update or not. In regards to whether or not the user should be aware, they can use the `--include-unknown` flag mentioned by @jedieaston if they wish to know about these packages.

felipecrs:
I mean like a third column:

```console
$ winget upgrade
Name         Id            Version Available  Source Auto-update   
[...]

Issue #1763, opened 2021-12-02
Import command option to not upgrade installed packages


jpawlowski:
### Description of the new feature / enhancement

`winget import` currently will also install newer versions of packages that are already installed.

There shall be the option to ignore packages that are already installed on the system, no matter the version. This will safe a lot of time when one just wants to install any missing software and not mix this up with upgrades.

### Proposed technical implementation details

Implement a command line option `--ignore-installed` to skip re-installing(=upgrading) a package.

jedieaston:
Related: https://github.com/microsoft/winget-cli/issues/929

Trenly:
@denelon - #2655

denelon:
@Trenly did this make it into 1.4 Release Candidate (2)?

Trenly:
> @Trenly did this make it into 1.4 Release Candidate (2)?

![image](https://user-images.githubusercontent.com/12611259/212379825-3c2fce62-8fbe-47a8-bb86-5d279def72e4.png)

PR #1765, opened 2021-12-02
Added argument to control whether to upgrade packages if they have "unknown" versions

jedieaston:

-----

Related to https://github.com/microsoft/winget-cli/issues/1692.

This PR adds a argument, `--include-unknown`, to `winget upgrade`. This argument controls whether to evaluate packages for upgrade if they have "Unknown" versions (basically, if the version is not set in the ARP table). This resolves a lot of user confusion where every time they run `winget upgrade --all`, packages will be upgraded regardless of whether they actually need it because winget is not sure if it needs an upgrade or not. 

I know nobody from MS commented on that issue, so if anyone has issues with this solution, I'm happy to redo this. The issue also has some discussion on whether or not upgrades for unknown packages should be shown by default in the upgrade table. The way I have it, the argument must be used for the upgrade to show in the table. 

Tested: manually. As always, I'm happy to write unit tests if needed.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1765)


src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -20,7 +20,8 @@ namespace AppInstaller::CLI
         bool ShouldListUpgrade(Context& context)
         {
             return context.Args.Empty() ||
-                (context.Args.GetArgsCount() == 1 && context.Args.Contains(Execution::Args::Type::Source));
+                (context.Args.GetArgsCount() == 1 && (context.Args.Contains(Execution::Args::Type::Source) || context.Args.Contains(Execution::Args::Type::IncludeUnknown))) ||
```
JohnMcPMS, line 23: My gut on the right way to do this is to enable:
```
for (Execution::Args::Type type : context.Args.GetTypes())
{
    if (type != Execution::Args::Type::Source && type != Execution::Args::Type::IncludeUnknown)
    {
        return false;
    }
}
return true;
```
I definitely don't like keeping the count in the condition and referencing types multiple times.

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -46,6 +47,7 @@ namespace AppInstaller::CLI
             Argument::ForType(Args::Type::AcceptSourceAgreements),
             Argument::ForType(Execution::Args::Type::CustomHeader),
             Argument{ "all", Argument::NoAlias, Args::Type::All, Resource::String::UpdateAllArgumentDescription, ArgumentType::Flag },
+            Argument{ "include-unknown", Argument::NoAlias, Args::Type::IncludeUnknown, Resource::String::IncludeUnknownArgumentDescription, ArgumentType::Flag }
```
JohnMcPMS, line 50: @denelon for argument name review.

JohnMcPMS (commented):
@denelon , does this seem like fine behavior to you?  Nothing springs to mind as to why we wouldn't want it.

@lechacon , does there need to be additional work done for import?  It does upgrade-like things but I don't think it uses this code path.  If it would only attempt to upgrade in the event that a specific version was given in the import file, then I think it is probably ok to leave as is, even if that means attempting to re-install.

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
```diff
@@ -86,16 +89,25 @@ namespace AppInstaller::CLI::Workflow
...
-            else
+        }
+        else
+        {
+            // the package has a unknown version and the user did not request to upgrade it anyway.
+            if (m_reportUpdateNotFound) 
             {
-                // Any following versions are not applicable
-                break;
+                context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl;
```
lechacon, line 107: I think this message could be more specific so a user knows that they can still update if they add the flag.

lechacon:
> @lechacon , does there need to be additional work done for import? It does upgrade-like things but I don't think it uses this code path. If it would only attempt to upgrade in the event that a specific version was given in the import file, then I think it is probably ok to leave as is, even if that means attempting to re-install.

Yes, there needs to be some work done for import as it doesn't use this code path. If there is no specific version in the file, I think import will try to update to the latest.

Import does call `EnsureUpdateVersionApplicable`, so we could do the unknown version check there and import would also pick it up. There are also some upgrade flows that don't use `SelectLatestApplicableUpdate` but do use `EnsureUpdateVersionApplicable`, like updating from a manifest or to a specific version; so maybe we need it there regardless.

denelon:
The current behavior lists everything. 

This PR could be considered a breaking change. I'm leaning in the direction of agreeing to do this since most of the packages listed from the "msstore" source today have `version: "Unknown"` and are expected to upgrade themselves. We are working towards getting versions from the Microsoft Store, so this will improve over time.

@JohnMcPMS does this code path get used for `winget list`? We should be consistent in both places.

I'm ok with "--include-unknown". I think this will help much more than hurt or confuse.

It might help to include a count for the number of packages with an unknown version.
```
PS C:\Users\denelon>winget upgrade
No installed package found matching input criteria. <n> packages specify "Unknown" version. Use "--include unknown" to see all results.
```

I would add the "<n> packages... Use..." at the end of the `winget upgrade --all` as well. That will help make this option more discoverable. If we do this for `winget list` it should be there as well.

>Note, we should ensure the value "Unknown" is appropriately defined for a private REST source so the behavior extends there as well @RDMacLachlan.

denelon:
@lechacon we should consider this for import as well.

* #1763

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
lechacon: Personally, I find the argument name somewhat awkward for a single app: `winget upgrade SomeApp --include-unknown`
For `upgrade --all` I understand it as "include apps with unknown version", but I don't quite get what we are "including" for a single app.

jedieaston:
> does this code path get used for `winget list`? We should be consistent in both places.

It does, but since the argument doesn't exist for `winget list` it will not show an available upgrade for applications with no version number. The app still is shown in the list, it just doesn't check for an upgrade. I can add the argument there too if you want to keep it consistent.

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
jedieaston: In the latest commit: 
```
PS C:\Users\easton\Downloads> wingetdev upgrade GOG.Galaxy
This package's version number cannot be determined. To upgrade it anyway, add the argument --include-unknown to your previous command.
PS C:\Users\easton\Downloads>
```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1250,4 +1250,10 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="InstallArchitectureArgumentDescription" xml:space="preserve">
     <value>Select the architecture to install</value>
   </data>
+  <data name="IncludeUnknownArgumentDescription" xml:space="preserve">
+    <value>Upgrade packages even if their current version cannot be determined</value>
+  </data>
+  <data name="UpgradeUnknownVersionExplanation" xml:space="preserve">
+    <value>This package's version number cannot be determined. To upgrade it anyway, add the argument --include-unknown to your previous command.</value>
```
lechacon, line 1257: Add a comment to make sure the localization team does not change the argument name:
```suggestion
    <value>This package's version number cannot be determined. To upgrade it anyway, add the argument --include-unknown to your previous command.</value>
    <comment>{Locked="--include-unknown"}</comment>
```

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
jedieaston: Resolved in latest commit.

denelon:
"Unknown" is considered to be a sentinel value meaning the lowest version of a package. When the installed version is also Unknown, no upgrade is shown as available.

So, there shouldn't be a case where an available package with version "Unknown" is displayed as a potential upgrade in list when there is a version present in Apps & Features.

I am just trying to make sure we capture all of the scenarios:
1. An installed version is known, and the only available package version is "Unknown"
2. An installed version is known, and the latest version of an available package is lower.
3. An installed version is known, and the latest version of an available package is higher.
4. An installed version is Unknown.

I believe the work here is for the last scenario.

@JohnMcPMS to keep me honest, or at least correct me if I'm mistaken.

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
denelon: I think if the user explicitly specifies an app, `--include-unknown` would/should be redundant.

I also think `winget upgrade AwesomePackage --all` would be similarly strange.

jedieaston:
@denelon, does this look good to you?

image

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1250,4 +1250,19 @@ Please specify one of them using the `--source` option to proceed.</value>
...
   </data>
+  <data name="IncludeUnknownArgumentDescription" xml:space="preserve">
+    <value>Upgrade packages even if their current version cannot be determined</value>
+  </data>
+  <data name="UpgradeUnknownVersionExplanation" xml:space="preserve">
+    <value>This package's version number cannot be determined. To upgrade it anyway, add the argument --include-unknown to your previous command.</value>
+    <comment>{Locked="--include-unknown"}</comment>
+  </data>
+  <data name="UpgradeUnknownCount" xml:space="preserve">
+    <value>package's version numbers cannot be determined. Use "--include-unknown" to see all results.</value>
```
lechacon, line 1261: packages'

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1250,4 +1250,19 @@ Please specify one of them using the `--source` option to proceed.</value>
...
+    <comment>{Locked="--include-unknown"}</comment>
+  </data>
+  <data name="UpgradeUnknownCountSingle" xml:space="preserve">
+    <value>package's version number cannot be determined. Use "--include-unknown" to see all results.</value>
```
lechacon, line 1265: I don't think this formatting ([n] + " package's version number") works for all languages. For example, in Spanish I would probably say this as "El número de versión de [n] paquetes no pudo ser determinada.", which doesn't have the number at the beginning. You can probably play with the phrasing to make it work (e.g. "[n] paquetes tienen números de versión que no pudieron ser determinados"), but I'm not sure if you can do that in all languages. At least we should add a comment about how this string is used so that the localization team can make it work.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -766,6 +774,10 @@ namespace AppInstaller::CLI::Workflow
                 context.Reporter.Info() << availableUpgradesCount << ' ' << Resource::String::AvailableUpgrades << std::endl;
```
lechacon, line 774: I see that you are using different strings to report the count depending on whether there is a single package or multiple, but here we were using the same string for both cases. I think we should be doing the same for both messages; although I'm not sure which one would be better. I would lean towards using a single string regardless of the number of packages as I don't know if there are any languages that make any distinction beyond 1/many or that don't make distinctions at all.

lechacon: I looked a bit into pluralization and it's a whole rabbit hole one can go into. I found it interesting to look at the [rules](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html) for all languages...
I found some internal references to string formats for putting all variants in a single string and using libraries for handling that. @JohnMcPMS does the current setup support something like that? @JohnMcPMS , @denelon , is this something we would want to handle?

jedieaston: I saw that too, but figured it was a separate PR... thanks for pointing it out :D

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: We have a general problem with this, and a bug from the localization team to prove it 😄 The right thing to do is actually to include format specifiers (not formally defined, but think like C#'s `{0}`) to our strings and replace them at runtime with the actual values.  I wanted it to be a generalized approach when we fix it, hopefully making it easy at the call sites to use consistent syntax.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
JohnMcPMS: If we wanted to do something to handle plurals programmatically, then we could probably do it at the same time as the dynamic string replacement mentioned above.  I don't see it as important as the dynamic strings though, so depending on the complexity we might just leave a nice place for it.  I don't know if the english standard of making it "optional" like:
```
N installed package version(s) cannot be determined.
```
applies to the other supported languages.  But it might be an option.

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
jedieaston: The current behavior is that `upgrade <ID>` doesn’t upgrade a package if already at latest. I had the argument behavior stay the same so the expectation of winget not doing an upgrade unless it has to holds.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
jedieaston: Should I go ahead and use a optional plural for now? `SourceOpenFailedSuggestion` uses it ("Failed when opening source(s); try the 'source reset' command if the problem persists.")...

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
Trenly: Community opinion here; I would think that if I specify an ID, it should attempt to be upgraded even if the version is unknown. I would think that even if the version doesn't change, from a user perspective, an upgrade was at least attempted. It could be very confusing if the user knows there is an update, and they try to upgrade it only to be told that no applicable update was found.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
Trenly: What if the message were more general?
```
Some packages' versions could not be determined. Use "--include-unknown" to see all results"
```
That would be easily localizable and avoid having to count the number of packages, saving on processing time and memory for systems with large numbers of packages installed

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
jedieaston: There's also a message that prints:
```
PS C:\Users\easton> wingetdev upgrade GOG.Galaxy
This package's version number cannot be determined. To upgrade it anyway, add the argument --include-unknown to your previous command.
PS C:\Users\easton>
```
There was some discussion in the issue about whether or not there should be a setting to restore the current behavior (when in doubt, upgrade), but by default I think it should prompt the user to make a decision. The infinite reupgrading is causing more confusion than this is, IMO.

Trenly: Ahh, I must have missed that. I would be fine with that approach, since it is clear and informative

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
jedieaston: The count was added after a request from @denelon: https://github.com/microsoft/winget-cli/pull/1765#issuecomment-985060905

Trenly: I wasn’t sure if denelon meant more that it would be helpful to have a message (since looking through the commits, it seems like it was only if you tried to upgrade a specific package?) or if he specifically meant to have a count

jedieaston: The design he gave left a spot for a count (`.... <n> have unknown version...`).

There are quite a few spots where the format specifier solution will need to be used (where strings are broken up into pieces to fit the localized bits, like in [`ReportIdentity`](https://github.com/microsoft/winget-cli/blob/60db65945217445651b8139d264d1326e9c23b54/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp#L39)). I think the phrasing change should be okay for now (I made it more verbose based on this review and added a comment for localization).

denelon: In either case, a message should be displayed, and the count is better.

jedieaston:
@lechacon Do you want me to add the argument to `import` as well? The only case where `upgrade` only uses `EnsureUpdateVersionApplicable` and not `SelectLatestApplicableUpdate` is where the user specifies a version, and in that scenario, they probably want to get to that version and winget telling them "I don't know if that's an upgrade or not..." wouldn't be helpful. But if we need the logic for import, might as well make it work the same way in both places...

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -19,8 +19,19 @@ namespace AppInstaller::CLI
     {
         bool ShouldListUpgrade(Context& context)
         {
-            return context.Args.Empty() ||
-                (context.Args.GetArgsCount() == 1 && context.Args.Contains(Execution::Args::Type::Source));
+            if (!context.Args.Empty())
```
lechacon, line 22: I don't think you need to check for empty here. If there isn't any, it should immediately go past the `for` and return true.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
lechacon: That sounds good to me.

lechacon (approved):
:shipit:

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
jedieaston: Fixed in latest commit!

PR #1780, opened 2021-12-06
Bug fix: PackageTrackingCatalog and SQLiteIndexUpdate

hackean-msft:

**PackageTrackingCatalog**:
When dependency nodes are not found in the index, an attempt to add or update manifest by the catalog tracking feature causes a failure. In some scenarios this might be valid as the dependencies of a package might already be present in the user's machine. This fix totally disable dependencies for catalog tracking feature by adding an option to not create dependencies table and also checking for the presence of the table when there is an attempt to perform any operation on the table.  

**SQLiteIndexUpdate**:
Bug in dependencies udpate manifest causes the method to always return true even if the index wasn't modified. This pull request fixes the issue, and also extends the update manifest tests to capture 1_4 schema update.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1780)

### DIFF
--- a/src/AppInstallerRepositoryCore/Microsoft/Schema/ISQLiteIndex.h
+++ b/src/AppInstallerRepositoryCore/Microsoft/Schema/ISQLiteIndex.h
@@ -42,7 +42,7 @@ namespace AppInstaller::Repository::Microsoft::Schema
         enum class CreateOptions
         {
             // Standard
-            None = 0,
+            None = 0x0,
             // Enable support for passing in nullopt values to Add/UpdateManifest
             SupportPathless = 0x1,
             // Disable support for dependencies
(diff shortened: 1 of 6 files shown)


src/AppInstallerRepositoryCore/Microsoft/Schema/ISQLiteIndex.h
```diff
@@ -42,9 +42,11 @@ namespace AppInstaller::Repository::Microsoft::Schema
         enum class CreateOptions
         {
             // Standard
-            None,
+            None = 0,
```
yao-msft, line 45: >0 [](http://example.com/codeflow?start=19&length=1)

nit: 0x0 since you are using hex representation

yao-msft (approved):
:shipit:

Issue #1788, opened 2021-12-10
Implement better validation of how CLI arguments are used together


florelis:
### Description of the new feature / enhancement

The current validation of arguments seems to miss many combinations of arguments that don't make sense together. For example, the `upgrade` commands allows `winget upgrade --all -v 1.0` (version is used only when upgrading single package, not all of them) or `winget upgrade -m myManifest.yaml --accept-source-agreements` (sources are not used with local manifest, so there are no agreements).

We should improve that validation, preferably in a way that makes it easier to extend as we add more arguments.

### Proposed technical implementation details

I think we could create "categories" of arguments and define the rules on top of them. For example, for upgrade there are arguments for selecting the package (e.g. id, version), arguments for selecting multiple packages (`--all`, `--include-unknown`), and arguments for the sources (source name, REST source header, agreements). We could define rules on top of them like disallowing arguments for both single and multiple packages, or for sources when including a local manifest. Adding a new argument to one category would include it in all existing rules easily.

JohnMcPMS:
Indeed, Powershell's parameter sets was something I had considered when this was originally written.  At the time it wasn't really worth it, but I assumed we could just add it later.  I still think we could do so easily.

I would argue for keeping them separate as in sets, rather than tagging each argument with flags.  That way, each set can have custom documentation (if needed) for the argument.

Trenly:
@florelis - Is this resolved with the above PR?

florelis:
> @florelis Is this resolved with the above PR?

I think the PR does what I wanted so I'll close this. Maybe in the future we'll want to do what @johnmcpms suggested about sets, but that can be a different issue.

Issue #476, opened 2020-07-01
Pin a package


denelon:
# Description of the new feature/enhancement

Users should be able to prevent the Windows Package Manager from updating a package (assumes the package doesn't have it's own auto-update).

`winget pin <package>` and the corollary `winget unpin <package>`.

This was mentioned by:

@megamorf https://github.com/microsoft/winget-cli/issues/120#issuecomment-635174060
@rodalpho https://github.com/microsoft/winget-cli/issues/120#issuecomment-635520051
@aetos382 https://github.com/microsoft/winget-cli/issues/120#issuecomment-635741018
@kmindi https://github.com/microsoft/winget-cli/issues/120#issuecomment-637343639

# Proposed technical implementation details (optional)

There should also be a mechanism to display the packages (and the version) that have been "pinned". This might be a function of list or a function of pin.  Additionally, if packages are known to self-update (like Visual Studio Code) an additional meta-data entry in the manifest could help users understand when they may not be able to pin a package.

Edit: Clarifying behavior for ambiguity suggested in other Issues.

Pinning should enable users to specify the exact version (likely the default case with no parameters).
It should also allow users to specify what portion of a version should be pinned so bug fixes could be applied, or even minor version bumps.

>Note: the syntax below is just suggestive.

Assume the user has installed "Awesome App" version "1.2.3".

`winget pin "Awesome App"`
This would pin to version 1.2.3 and would not be upgraded when any newer version is released.

`winget pin "Awesome App" --version 1.2`
This would pin version to anything less than version 1.3. If version 1.2.4 were released this would be a valid upgrade.

`winget pin "Awesome App" --version 1`
This would pin the version to anything less than version 2. If version 1.2.4 were released this would be a valid upgrade. If version 1.3.0 were released this would also be a valid upgrade.

Additional example per @brainz80
>E.g. Having nodejs-lts version 12.1.1 installed and running winget pin --all nodejs-lts --version 12.x should make it so that when later winget upgrade --all is ran it'd check if there is a update available for nodejs-lts what matches 12.x - if the latest update is e.g 16.1.1, but there is also a 12.1.2 it'd update nodejs-lts to that version.

megamorf:
From my experience the most common scenario for why you'd want to pin a package version is when there is a known issue with a newer version of that package itself or with one if its dependencies. So on Linux that meant I pinned the docker package because one of its dependencies (containerd) could not be installed. 

[...]

PR #1795, opened 2021-12-13
Add support for `RequireExplicitUpgrade` manifest element

### LINKED ISSUE #1163: Enhancements to upgrade (Apps that upgrade themselves)

Several packages manage upgrades themselves. One example is Visual Studio Code.

Add a key to the manifest to identify packages with this capability.
>Placeholder value: "PackageAutoUpgrades" (optional boolean key)

When the user executes `winget upgrade` this package would either not be displayed (as having an upgrade available), or it would be displayed in a manner that indicates this behavior.

When the user executes `winget upgrade --all` this package would not be upgraded.
To be in line with `--include-unknown` we should add `--include-automatic`.
We may also want to consider `--force` to apply all `--include-<*>` options.

>Note: Schema changes have a large surface area: (Client, Validation Pipelines, Publishing Pipelines, REST API, Manifest Creator). This key may be added to the schema ahead of implementing the behavior in the client.

lechacon:
Adding support for skipping packages marked as `RequireExplicitUpgrade` during an `upgrade --all`, and an argument `--include-explicit` to upgrade these packages. This is similar to what #1765 did for unknown installed versions.

* Updated argument validation for upgrade to prevent some argument combinations that don't work together (e.g. `--manifest` with `--header`) and allow some that should work together (e.g. `winget upgrade` with `--accept-source-agreements`). This is related to what I mentioned in #1788, but other commands probably still accept weird combinations.
* Added logic to `upgrade --all` to skip packages marked as requiring explicit upgrade unless the argument was specified. Did not do the same for `upgrade` (list available) as this requires having identified the installer to use, which we don't currently do in that flow.
* Added unit tests for the new `--include-explicit` argument.

Closes #1163 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1795)


src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -52,7 +103,8 @@ namespace AppInstaller::CLI
             Argument::ForType(Args::Type::AcceptSourceAgreements),
             Argument::ForType(Execution::Args::Type::CustomHeader),
             Argument{ "all", Argument::NoAlias, Args::Type::All, Resource::String::UpdateAllArgumentDescription, ArgumentType::Flag },
-            Argument{ "include-unknown", Argument::NoAlias, Args::Type::IncludeUnknown, Resource::String::IncludeUnknownArgumentDescription, ArgumentType::Flag }
+            Argument{ "include-unknown", Argument::NoAlias, Args::Type::IncludeUnknown, Resource::String::IncludeUnknownArgumentDescription, ArgumentType::Flag },
+            Argument{ "include-explicit", Argument::NoAlias, Args::Type::IncludeExplicit, Resource::String::IncludeExplicitArgumentDescription, ArgumentType::Flag },
```
lechacon, line 107: @denelon Is this argument name okay?

denelon: I don't know of anything better. Can you think of a better term for something that represents packages that update themselves?

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
```diff
@@ -176,6 +179,18 @@ namespace AppInstaller::CLI::Workflow
                 continue;
             }
 
+            // Filter out packages that require explicit upgrades.
+            // Note that this is filtering based on the updated installer, not the installed version. So there
```
JohnMcPMS, line 183: We could store this information in the tracking catalog so that we can make the decision based on the "current" version.  I envisioned that it is the same location we would store the pinning data, which is basically a user driven version of this feature.

There will always be problems if the state changes, whether it be an update to the current version or a newer version changing it.  We could potentially solve it by not even storing the value but always looking it up from the manifest for the "most likely version" (something like "greatest version that is less than or equal to the current version").

I think either of these options is better than what you have, which will guarantee a gap in the experience if the state changes across versions.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1257,12 +1257,18 @@ Please specify one of them using the `--source` option to proceed.</value>
...
     <comment>{Locked="--include-unknown"} This string is preceded by a (integer) number of packages that do not have notated versions. </comment>
   </data>
-  <data name="UpgradeUnknownCountSingle" xml:space="preserve">
-    <value>package has a version number that cannot be determined. Use "--include-unknown" to see all results.</value>
-    <comment>{Locked="--include-unknown"} This string is preceded by a (integer) number of packages that do not have notated versions.</comment>
+  <data name="IncludeExplicitArgumentDescription" xml:space="preserve">
+    <value>Upgrade packages that are pinned by default</value>
+  </data>
+  <data name="UpgradeRequireExplicitCount" xml:space="preserve">
+    <value>package(s) is pinned by default and needs to be explicitly upgraded. Use "--include-explicit" to see all results.</value>
```
JohnMcPMS, line 1268: Maybe we shouldn't call out `by default` and just treat the manifest property and pinning as the same feature (to the user).  Internally we can still track the difference so that we can allow unpinning something that is marked as explicit, but externally we just treat it no differently than if the user had requested pinning.

JohnMcPMS (changes requested):
The determination of whether explicit upgrade is required needs to be sorted out.  I would lean towards using the tracking catalog because it makes the delta to having user driven pinning very small.  The drawback being that if the package isn't installed through us, or was installed prior to this feature, or an update to the manifest changes it, we won't have the correct information.  But I think those are acceptable, and even recoverable by the user with pinning, situations.

glen-84:
> Did not do the same for upgrade (list available) as this requires having identified the installer to use, which we don't currently do in that flow.

Is there an issue to fix this that I can follow? I usually use `upgrade` to see what upgrades are available, and then update applications one by one. It'll be annoying to have to keep ignoring certain packages like `Microsoft Edge WebView2 Runtime`.

glen-84:
@denelon I'm referring specifically to the quoted text. Closing this PR will close #1163, but it won't be applied to `winget upgrade`, only `winget upgrade --all`. I wanted to track when it might be supported there as well.

denelon:
I believe` winget upgrade` will show "everything" that would be upgraded if you ran `winget upgrade --all`. The exception here is for packages that upgrade themselves like Visual Studio Code.

We were thinking that "pin" would exclude packages from being upgraded even if an upgrade was available unless you unpinned them. Users would explicitly pin things they don't want upgraded.

The "Unknown" version problem is being handled slightly differently. We're adding additional arguments to be able to display packages that don't report a version to Windows Apps & Features, and to go ahead and "try" to upgrade them. We still wouldn't know if it was an upgrade, so it would be more like just running `winget upgrade` on those specific packages and hoping the version we have in the source is newer. There really isn't any way to know that until the installers actually report the version installed.

glen-84:
> I believe` winget upgrade` will show "everything" that would be upgraded if you ran `winget upgrade --all`. The exception here is for packages that upgrade themselves like Visual Studio Code.

I understood @lechacon's description to mean that `winget upgrade` will still list packages marked as `RequireExplicitUpgrade`.

It might be useful to have the option to filter those out. For example, if VSC or WebView2 will be upgraded separately, you don't necessarily need to see them listed as requiring an upgrade. They could still be listed by default though, but perhaps it should indicate which packages require an explicit upgrade, so that you can at least mentally ignore them when looking at the list.

JohnMcPMS:
I agree with @glen-84 , we should still show them by default with `winget upgrade` but separate them to indicate that they won't be affected by `upgrade --all`.  I would rather split them out than mark them with a `*` or something.  My suggestion is something like:

```
> winget upgrade
Name                                          Id                                     Version       Available     Source
Windows Package Manager Manifest Creator      Microsoft.WingetCreate                 0.1.0.6       0.4.4.1       winget
Visual Studio Enterprise 2019                 Microsoft.VisualStudio.2019.Enterprise 16.11.7       16.11.9       winget
Microsoft Visual C++ 2015-2019 Redistributab… Microsoft.VC++2015-2019Redist-x64      14.29.30133.0 14.29.30139.0 winget
Microsoft Visual C++ 2015-2019 Redistributab… Microsoft.VC++2015-2019Redist-x86      14.29.30133.0 14.29.30139.0 winget
Microsoft Visual Studio Code                  Microsoft.VisualStudioCode             1.62.3        1.63.2        winget
Adobe Acrobat Reader DC                       Adobe.Acrobat.Reader.32-bit            21.007.20091  21.011.20039  winget
PowerToys (Preview)                           Microsoft.PowerToys                    0.45.0        0.53.3        winget

The following packages have an upgrade available, but require explicit targeting for upgrade:
Name                                          Id                                     Version       Available     Source
Microsoft Azure CLI                           Microsoft.AzureCLI                     2.27.0        2.32.0        winget
```

denelon:
We should add another table for "pinned" packages along with that feature.

denelon:
> > We should add another table for "pinned" packages along with that feature.
> > #476
> 
> I would prefer to treat manually pinned and default pinned through `RequiresExplicitUpgrade` the same to the user. Thus they would both appear in the lower section and could all be "unpinned" if desired.

I think of "pinning" an explicit request made by a user. They are saying "don't" upgrade this package. If the package updates itself as specified in the manifest which is what this is covering, the user wouldn't be able to "pin" it since we couldn't guarantee to honor the request. Visual Studio Code is the example for this. It doesn't make sense to have users upgrade it with the Windows Package Manager since the program will update itself the next time they launch.

There may be some confusion in using the term "pin" in the description.

JohnMcPMS:
> I think of "pinning" an explicit request made by a user. They are saying "don't" upgrade this package. If the package updates itself as specified in the manifest which is what this is covering, the user wouldn't be able to "pin" it since we couldn't guarantee to honor the request. Visual Studio Code is the example for this. It doesn't make sense to have users upgrade it with the Windows Package Manager since the program will update itself the next time they launch.
> 
> There may be some confusion in using the term "pin" in the description.

We don't know or control what happens after installation begins.  We can't honor any request to pin because of this.  Thus, semantically these are the same thing; don't attempt to upgrade this package unless the user explicitly tells us to.  But I agree that we shouldn't use the term "pin" in our usage anywhere except the command as a verb.  You pin a package, but it is not necessarily pinned.  It could just as well be:
```
winget pin <ID>
Pin a package.  This will exclude the package from winget upgrade --all, requiring an explicit
winget upgrade <ID> command in order for an upgrade to be attempted.
```

werrnerben (approved):

lechacon:
Reviving this PR with changes per the comments..

* Changed the parameter to `--include-pinned`
* Included the list of pinned packages in the list of available upgrades
* Changed the logic for determining whether a package requires an explicit upgrade to use a value from the metadata in the tracking catalog instead of the manifest being installed. I made the flag thinking of being able to add a state "pinned by user"

denelon:
We don't have pinning, and this isn't the same thing. I would use --include-pinned in this use case.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -751,22 +783,32 @@ namespace AppInstaller::CLI::Workflow
...
-                         shouldShowSource ? sourceName : ""s
-                    });
-                   
-                   
+                         shouldShowSource ? sourceName : Utility::LocIndString()
+                    );
+
+                    auto pinnedState = ConvertToPackagePinnedStateEnum(installedVersion->GetMetadata()[PackageVersionMetadata::PinnedState]);
+                    bool requiresExplicitUpgrade = m_onlyShowUpgrades &&
+                        (pinnedState == PackagePinnedState::PinnedByManifest);
```
JohnMcPMS, line 796: I think you are going to want to either create a semantic check function here, or always use `pinnedState != Unpinned` to enable other pinning reasons to automatically be included.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1272,6 +1272,17 @@ Please specify one of them using the `--source` option to proceed.</value>
     <value>package has a version number that cannot be determined. Using "--include-unknown" may show more results.</value>
     <comment>{Locked="--include-unknown"} This string is preceded by a (integer) number of packages that do not have notated versions.</comment>
   </data>
+  <data name="UpgradeUnknownVersionCount" xml:space="preserve">
+    <value>package(s) have version numbers that cannot be determined. Use "--include-unknown" to see all results.</value>
```
JohnMcPMS, line 1276: @AmelBawa-msft is working to fix our lack of placeholders in localization strings.  This family of strings will need to be updated and this comment is mostly a note for him.

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -178,11 +178,24 @@ namespace AppInstaller::Repository
...
     };
 
     // Convert a PackageVersionMetadata to a string.
     std::string_view ToString(PackageVersionMetadata pvm);
 
+    // Possible pinned states for a package.
+    // Pinned packages need to be explicitly updated (i.e., are not included in `upgrade --all`)
+    enum class PackagePinnedState
+    {
+        Unknown,
```
JohnMcPMS, line 192: I would suggest that unknown and "not pinned" are the same state, then use `NotPinned` as the name to better indicate that.

JohnMcPMS (approved):

JohnMcPMS:
> We don't have pinning, and this isn't the same thing. I would use --include-pinned in this use case.

I assume you meant "I would *not* use...".

But the manifest entry *is* implicit pinning, and I don't think we want to create a parallel feature.  Adding end-user controlled pinning should simply allow the default to be overridden.  The default is currently "not pinned ever", this PR enables the default to be package/installer specific.

denelon:
@JohnMcPMS you are correct. I think the confusion would come from using "--include-pinned" for `winget upgrade --all`. Truly pinned packages wouldn't be upgraded unless the pin were removed. These packages could be included for upgrades, we're just excluding them from `winget upgrade --all` when they are known to upgrade themselves.

denelon:
The example packages I was thinking about are Visual Studio Code and Edge Web View 2 Runtime. These packages upgrade themselves and may not work and play well when manually upgraded. I'm sure there are others. I've also seen some cases with other browsers and tools that have more than one package being upgraded with automatic updates.

JohnMcPMS:
> Truly pinned packages wouldn't be upgraded unless the pin were removed.

I disagree, manually pinned packages should be able to be explicitly upgraded.  I don't see pinning as setting read-only on the package, just moving to a separate list to prevent normal flow from upgrading it.

We could make an additional level of pinning that prevents any modifying interaction, but how far does that guarantee need to go?  I don't think it is actually necessary.

Back to `--include-pinned`; I would have no problem simply removing the entire concept from the PR.  I'm not sure that there is actually a large use case for it, since blindly upgrading any "pinned" package defeats the entire point of the pinning.  Forcing every package to be named individually would prevent a new package with `RequireExplicitUpgrade` (or manually pinned) from suddenly being included and breaking things.

PR #1796, opened 2021-12-14
Allow upgrades in packages that register a different installer type

### LINKED ISSUE #1242: Complex installers

lechacon:
* Modified the applicability check in `InstalledTypeComparator` to also consider the installer types listed under `AppsAndFeaturesEntries`
* Extended string returned from `ExplainInapplicable` to also mention the installed type and other types accepted by the manifest.

Closes #1242. Validated that upgrading PowerToys (mentioned in that issue) worked correctly using a local manifest edited to include the `AppsAndFeaturesEntries` and installed type `msi`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1796)

### DIFF
--- a/src/AppInstallerCLITests/WorkFlow.cpp
+++ b/src/AppInstallerCLITests/WorkFlow.cpp
@@ -210,6 +210,21 @@ namespace
                         PackageMatchFilter(PackageMatchField::Id, MatchType::Exact, "AppInstallerCliTest.TestExeInstaller")));
             }
 
+            if (input == "TestExeInstallerWithDifferentInstalledType")
+            {
+                auto manifest = YamlParser::CreateFromPath(TestDataFile("InstallFlowTest_Exe.yaml"));
+                auto manifest2 = YamlParser::CreateFromPath(TestDataFile("UpdateFlowTest_Exe_ARPInstallerType.yaml"));
+                result.Matches.emplace_back(
+                    ResultMatch(
+                        TestPackage::Make(
+                            manifest,
+                            TestPackage::MetadataMap{ { PackageVersionMetadata::InstalledType, "Msix" } },
+                            std::vector<Manifest>{ manifest2, manifest },
+                            shared_from_this()
+                        ),
+                        PackageMatchFilter(PackageMatchField::Id, MatchType::Exact, "AppInstallerCliTest.TestExeInstaller")));
+            }
+
             if (input == "TestExeInstallerWithNothingInstalled")
             {
                 auto manifest = YamlParser::CreateFromPath(TestDataFile("InstallFlowTest_Exe.yaml"));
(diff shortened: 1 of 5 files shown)


jedieaston:
Another test case for this is iTunes, it packages a msi inside a exe.

If you add this ARP entry:
```
  AppsAndFeaturesEntries:
    - DisplayName: iTunes
      Publisher: Apple Inc.
      DisplayVersion: 12.12.1.1
      InstallerType: msi
```
to the manifest, it works great!

```
PS C:\Users\easton\Downloads> winget upgrade -m .\12.12.1.1\
No applicable installer found; see logs for more details.
PS C:\Users\easton\Downloads> wingetdev upgrade -m .\12.12.1.1\
Found iTunes [Apple.iTunes] Version 12.12.1.1
Starting package install...
Successfully installed
PS C:\Users\easton\Downloads>
```

denelon:
This addresses at least one version of Adobe Acrobat, and PowerToys (@crutkas).

src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -210,6 +210,21 @@ namespace
                         PackageMatchFilter(PackageMatchField::Id, MatchType::Exact, "AppInstallerCliTest.TestExeInstaller")));
             }
 
+            if (input == "TestExeInstallerWithDifferentInstalledType")
```
lechacon, line 213: I'm not a fan of how long this function is getting. Many tests seem to use only a single case of all this and most are used in a single test, so we probably could add a constructor that took the manifests and always returned that. It would also put the search results closer to where it's used. Putting this here to see what others think of doing that..

JohnMcPMS: Agree, making it easy to create specific new cases rather than extending this would be good.

JohnMcPMS (approved):

PR #1806, opened 2021-12-15
Add support for markets

### LINKED ISSUE #1241: Add Markets to manifest

The new store source we are working on requires information regarding the market a given user is in. We need to extend the manifest to we can support behaviors related to markets.

This would be optional, and a manifest could specify either an allow list, or a deny list. 

Markets will have an "Allow" list and a "Exclusion" list.
 
If the client receives both lists and any individual market is in both lists the client will not proceed.
&nbsp;And the client will inform the user the manifest is invalid due to markets.
 
If the client receives either (but only one list) the client will only evaluate based on the list it received
If we receive an allow list:
&nbsp;If the client is in the market included in the allow list, then the install will proceed.
&nbsp;If the client is not in the market included in the allow list then the install will not proceed.
&nbsp;&nbsp;And the client will inform the user that the package is not allowed in their market.

If we receive an Exclusion list:
&nbsp;If the client is in the market included in the exclusion list, then the install will not proceed.
&nbsp;&nbsp;And the client will inform the user that the package is not allowed in their market.
&nbsp;If the client is not in the market included in the exclusion list, then the install will proceed.
		 
If the client receives both an allow list and an exclusion list the client will behave as though it only received an allow list.

```YAML
Installers:
  - InstallerType: exe
    InstallerArchitecture: x64
    Markets:
      AllowedMarkets:
       - DE
      ExcludedMarkets:
       - US
```

lechacon:
Adding support for filtering installers based on markets in the manifest.
* Added new field filter for markets to `ManifestComparator`. Filter uses the current market as returned by `GetUserDefaultGeoName()`.
* Added unit tests for new filter.

Closes #1241

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1806)


src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -508,12 +508,99 @@ namespace AppInstaller::CLI::Workflow
...
+            }
+
+        private:
+            Manifest::string_t m_market;
+
+            static Manifest::string_t GetCurrentMarket()
+            {
+                // TODO: Manifest uses 2-letter codes (ISO 3166?) but GetUserDefaultGeoName can return
+                // either a 2-letter ISO-3166 code or a numeric UN M.49 code.
+                auto geoNameSize = GetUserDefaultGeoName(nullptr, 0);
```
lechacon, line 562: I'm not actually 100% sure whether this matches what was intended with this manifest field, but it should be easy to change if it's not :)
I'm somewhat worried about the fact that this function can return 3-digit numeric codes, which are not supported in the manifest. I couldn't find anything about when would it return ISO 3166 or UN M.49.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -508,12 +508,99 @@ namespace AppInstaller::CLI::Workflow
...
+                std::string result = "Current market '" + m_market + "' does not match installer markets." +
+                    " Allowed markets: " + GetMarketsListAsString(installer.Markets.AllowedMarkets) +
+                    " Excluded markets: " + GetMarketsListAsString(installer.Markets.ExcludedMarkets);
+                return result;
+            }
+
+        private:
+            Manifest::string_t m_market;
+
+            static Manifest::string_t GetCurrentMarket()
```
JohnMcPMS, line 558: We already need to get the market for passing to the store; it seems like this should already by a common function (or should have been made so then).  If not, please move to a common location and use by both.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
JohnMcPMS: I would hope that the manifest field and the REST market were considered at the same time and should be the same format.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
lechacon: Changed to use the existing code which uses the 2 letter format.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -508,12 +508,99 @@ namespace AppInstaller::CLI::Workflow
...
+                std::vector<wchar_t> geoName(geoNameSize);
+                geoNameSize = GetUserDefaultGeoName(geoName.data(), geoNameSize);
+                THROW_LAST_ERROR_IF(geoNameSize == 0);
+
+                return Manifest::string_t{ geoName.data() };
+            }
+
+            std::string GetMarketsListAsString(const std::vector<Manifest::string_t>& markets)
+            {
+                // TODO: Refactor to merge with GetLocalesListAsString and GetAllowedArchitecturesString
```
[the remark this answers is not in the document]
lechacon: Done. I had to make it work for `Manifest::string_t`, `NormalizedString` and `Architecture` but I'm not that good at template magic, so it probably could be improved.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -508,12 +508,99 @@ namespace AppInstaller::CLI::Workflow
...
+                return std::make_unique<MarketFilter>(GetCurrentMarket());
+            }
+
+            InapplicabilityFlags IsApplicable(const Manifest::ManifestInstaller& installer) override
+            {
+                // If both allowed and excluded lists are provided, we only need to check the allowed markets.
+                if (!installer.Markets.AllowedMarkets.empty())
+                {
+                    // Inapplicable if NOT found
+                    if (installer.Markets.AllowedMarkets.end() == std::find(installer.Markets.AllowedMarkets.begin(), installer.Markets.AllowedMarkets.end(), m_market))
```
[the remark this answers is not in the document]
lechacon: I made the comparison case insensitive, but after re-reading your comment I'm not sure if that was what you meant. The manifest schema requires the markets to be two uppercase letters, so we shouldn't need any normalization there, although I'm not sure if the region from  the OS could not be uppercase.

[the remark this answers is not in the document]
JohnMcPMS: Unless this is a performance critical path and we find this to be an issue, I would leave it as case insensitive (also unless it turns out that case sensitivity is actual part of the definition of markets).  We don't own the validation on both sides, so better to be safe.

src/AppInstallerCommonCore/Public/AppInstallerStrings.h
```diff
@@ -145,4 +145,36 @@ namespace AppInstaller::Utility
...
+            if (firstItem)
+            {
+                firstItem = false;
+            }
+            else
+            {
+                strstr << ", ";
+            }
+
+            strstr << toString(item);
```
JohnMcPMS, line 168: Which type(s) were not being output properly by the `<<`?  I would expect any stringish type we use to work without a coercion function.

Not critical, but I like to understand 😄

JohnMcPMS (approved):

src/AppInstallerCommonCore/Public/AppInstallerStrings.h
lechacon: `Utility::Architecture` for the applicable architectures filter. It was coming out as an integer.

PR #1807, opened 2021-12-16
Add support for UnsupportedOSArchitectures manifest element

### LINKED ISSUE #993: Identify installers that don't work under emulation

Some installers do not work under emulation. The Windows Package Manager client would try to install these under emulation. There should be a mechanism to identify known architectures where a package shouldn't be installed or wouldn't work correctly.

lechacon:
Adding support for `UnsupportedOSArchitectures` manifest entry by extending `MachineArchitectureComparator` to add a check that the system architecture is not listed as unsupported.

Closes #993

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1807)


ranm-msft (approved):
:shipit:

PR #1813, opened 2021-12-17
Add new build configuration for WinGetUtil

lechacon:
Adding a new build configuration called WinGetUtilRelease for building the WinGetUtil. Some things we referenced in the project are not available in the OS we are using the WinGetUtil in, causing DLL load errors. The new build configuration includes some defines that allows us to exclude code we don't want in the winget util.

The configuration is as follows:
* Debug builds WinGetUtil as it always has
* Release does not build WinGetUtil anymore
* WinGetUtilRelease builds WinGetUtil and its dependencies (CommonCore and RepositoryCore), which can define something to exclude what we don't need.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1813)


src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -373,6 +373,11 @@ namespace AppInstaller::Settings
 
     UserSettings::UserSettings() : m_type(UserSettingsType::Default)
     {
+// When building WinGetUtil, we don't want to use user settings so we don't do anything here.
```
ranm-msft, line 376: >so we don't do [](http://example.com/codeflow?start=64&length=14)

please add a note as to why we don't

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -373,6 +373,11 @@ namespace AppInstaller::Settings
 
     UserSettings::UserSettings() : m_type(UserSettingsType::Default)
     {
+// When building WinGetUtil, we don't want to use user settings so we don't do anything here.
+// Some OS APIs related to architecture that we use when reading the settings are not
+// available on Windows 10 1607, which is where we use the util. We avoid compiling that
```
ranm-msft, line 378: >on Windows 10 1607 [](http://example.com/codeflow?start=13&length=18)

please add a note that this is the configuration for the Azure build machine which causes the errors

ranm-msft (approved):
:shipit:

lechacon:
Closing in favor of #1844

PR #1839, opened 2022-01-07
Add a default user agent to REST source calls

JohnMcPMS:
## Change
Provides a default user agent header in the http helper class that the REST source code uses.  It can be overridden by providing the user agent header in the headers argument.

The default format is:
```
winget-cli WindowsPackageManager/1.2.3 DesktopAppInstaller/10.11.12
```
where `1.2.3` is the client version as would be output by `winget -v` (without the `v` at the beginning) and `10.11.12` is the package version of the MSIX.

Also updates the WinINET downloader to use the same user agent string, but it is not overridable.

## Validation
Added a unit test to ensure that the default was being set properly for both `GET` and `POST`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1839)


yao-msft (approved):

jedieaston:
The WinINet downloader uses the user-agent `winget-cli` (see [here](https://github.com/microsoft/winget-cli/blob/2a749eb7de4d1d6350e8fa4de544e85cba2717bc/src/AppInstallerCommonCore/Downloader.cpp#L33)), should that be switched to `WPM/<version>` too?

JohnMcPMS:
> The WinINet downloader uses the user-agent `winget-cli` (see [here](https://github.com/microsoft/winget-cli/blob/2a749eb7de4d1d6350e8fa4de544e85cba2717bc/src/AppInstallerCommonCore/Downloader.cpp#L33)), should that be switched to `WPM/<version>` too?

I will take @denelon 's 👍 as an indication that he thinks so.  I was indifferent, but will update the PR accordingly.

JohnMcPMS:
Based on some internal discussion with @denelon, I've updated the actual user agent string (see PR description for new format).  It will still be used in the same locations.

PR #1852, opened 2022-01-13
Split pipeline build job into x86 and x64

lechacon:
# Change:
This changes the CI build pipeline to build/test x86 and x64 in two separate parallel jobs.

# Context:
On #1813 the build failed because the build machine was running out of space. We are using Microsoft Hosted agents for running the pipelines, which have only 10GB free space (according to docs; in practice it seems to be around 12GB). Upon investigation, building the solution was already taking about 9GB since we are building both x86 and x64 at the same time. Out of these, about 4GB were from pre-compiled headers; the biggest offender was the one for CommonCore, with over 800MB for each x86,x64.

I also considered trying to reduce the size of the pre-compiled headers or delete them after compilation (but didn't know how); or setting up a custom agent pool with more space (but setting it up and maintaining it seemed complicated). The idea of splitting the jobs by platform is from @yao-msft :)

By splitting into two jobs, building the solution for each one now takes only about 5GB, which puts us a safe distance from the 10GB limit.

# Change details:
* Added a matrix strategy for the build job to run the same steps for both platforms
* Renamed all build artifacts that contained the platform in the file name to now be grouped under a single folder per platform
* Removed duplicated test runs (there was one per platform)
* Moved some files generated by the pipeline from the MSBuild output directory to the pipeline's artifact staging directory
* Changed tests to run only if the build succeeded by removing the `succeededOrFailed()` condition. Added `continueOnError: true` to the tests so that it still makes it to the publish test results step if they fail.
* Also added publishing of MSBuild binlogs for diagnosing build errors

Opening as draft first to validate that the pipeline works correctly.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1852)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -105,7 +105,7 @@ jobs:
       platform: '$(buildPlatform)'
       solution: '$(solution)'
       configuration: '$(buildConfiguration)'
-      msbuildArgs: '/p:AppxBundlePlatforms="$(buildPlatform)"
+      msbuildArgs: '/p:AppxBundlePlatforms="x86|x64"
                     /p:AppxPackageDir="$(appxPackageDir)"
                     /p:AppxBundle=Always
                     /p:UapAppxPackageBuildMode=StoreUpload
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -200,7 +200,7 @@ jobs:
     name: HTTPSDevCert
     displayName: 'Download Kestrel Certificate'
     inputs:
-      secureFile: 'AppInstallerTest.pfx'
+      secureFile: 'HTTPSDevCertV2.pfx'
 
   - task: MSBuild@1
     displayName: Build MSIX Test Installer File
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -105,22 +105,15 @@ jobs:
       platform: '$(buildPlatform)'
       solution: '$(solution)'
       configuration: '$(buildConfiguration)'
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -253,145 +181,100 @@ jobs:
...
-
-  - task: PublishBuildArtifacts@1
-    displayName: Publish E2E Tests Packaged x64 Log
-    inputs:
-      PathtoPublish: 'C:\Users\VssAdministrator\AppData\Local\Temp\E2ETestLogs'
-      ArtifactName: 'E2ETestPackagedx64Log'
-      publishLocation: 'Container'
-    condition: succeededOrFailed()
 
+  # TODO: This does not seem to be matching the logs for x86
```
yao-msft, line 230: What does this mean by not matching?

lechacon: The logs for x86 are not under %temp%\E2ETestLogs, so it's not really copying anything. I couldn't figure out why as the tests should be copying winget logs to that path. I had left it as a TODO as it's still publishing the test results, so it's not critical.
But just now I'm realizing that the x86 used to have a different path from x64. I'll try with that.

lechacon: The logs for x86 were not there because the test cleanup was failing before getting to that part. I removed this TODO, and should have fixed the cleanup error (but still waiting for the build)

JohnMcPMS (approved):

PR #1853, opened 2022-01-14
Add upgrade functionality in Com api

yao-msft:
## Change
- Added upgrade Com api
- Fixed Com install issue for msstore type
- Fixed Com install issue when an installer has dependencies

## Validation
Since the Com packaged tests are not ready yet, I manually verified the fixed Com install api and the newly added upgrade api works on wnget source and msstore rest source.

Next I'll work on enabling Com tests as part of the e2e tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1853)


src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
```diff
@@ -27,8 +27,7 @@ namespace AppInstaller::CLI
     void COMInstallCommand::ExecuteInternal(Context& context) const
     {
         context <<
-            Workflow::GetInstallerHash <<
-            Workflow::VerifyInstallerHash <<
+            Workflow::ReverifyInstallerHash <<
```
JohnMcPMS, line 30: nit: This seems overly aggressive on combining tasks.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -341,6 +341,9 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         Microsoft::Management::Deployment::PackageVersionInfo packageVersionInfo = GetPackageVersionInfo(package, options);
         AddPackageManifestToContext(packageVersionInfo, context.get());
 
+        // If the installer has dependencies, DependencySource needs to be set.
```
JohnMcPMS, line 344: Is there not a way that we can set the actual Source object rather than adding an argument?

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -422,6 +426,30 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+                        // Set upgrade flag
+                        comContext->SetFlags(AppInstaller::CLI::Execution::ContextFlag::InstallerExecutionUseUpdate);
+                        // Add installed version
+                        winrt::Microsoft::Management::Deployment::implementation::PackageVersionInfo* installedVersionInfoImpl = get_self<winrt::Microsoft::Management::Deployment::implementation::PackageVersionInfo>(package.InstalledVersion());
+                        std::shared_ptr<::AppInstaller::Repository::IPackageVersion> internalInstalledVersion = installedVersionInfoImpl->GetRepositoryPackageVersion();
+                        comContext->Add<AppInstaller::CLI::Execution::Data::InstalledPackageVersion>(internalInstalledVersion);
+                    }
+                    else
+                    {
+                        co_return GetInstallResult(executionStage, APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE, correlationData, false);
```
JohnMcPMS, line 449: I think that we might need to break down the specific cases more for these errors.  Specifically, I'm concerned about overloading this error (based on what I think is returning it, so I could be wrong).  If the update does not have an applicable installer, do we get this error?

If the caller gives us a version that is less than or equal to the current version, should we have a different error?  If the caller does not give us a version, but the latest version is already installed, what error will be produced, if any?

src/Microsoft.Management.Deployment/PackageManager.h
```diff
@@ -21,9 +21,12 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         winrt::Microsoft::Management::Deployment::PackageCatalogReference CreateCompositePackageCatalog(winrt::Microsoft::Management::Deployment::CreateCompositePackageCatalogOptions const& options);
         winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::InstallResult, winrt::Microsoft::Management::Deployment::InstallProgress> 
             InstallPackageAsync(winrt::Microsoft::Management::Deployment::CatalogPackage package, winrt::Microsoft::Management::Deployment::InstallOptions options);
-        //Contract 2.0
+        // Contract 2.0
         winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::InstallResult, winrt::Microsoft::Management::Deployment::InstallProgress> 
             GetInstallProgress(winrt::Microsoft::Management::Deployment::CatalogPackage package, winrt::Microsoft::Management::Deployment::PackageCatalogInfo catalogInfo);
+        // Contract 4.0
```
JohnMcPMS, line 27: What happened to 3?  Was that the enum value I added?

JohnMcPMS (changes requested):

src/Microsoft.Management.Deployment/PackageManager.cpp
yao-msft: It's my test manifest is pointing to invalid dependencies. I'll remove this line as it's unnecessary.

src/AppInstallerCLICore/Commands/COMInstallCommand.cpp
yao-msft: We'll leave it as is for bug fix now

src/Microsoft.Management.Deployment/Converters.cpp
```diff
@@ -199,6 +199,8 @@ namespace winrt::Microsoft::Management::Deployment::implementation
             resultStatus = winrt::Microsoft::Management::Deployment::InstallResultStatus::NoApplicableInstallers;
             break;
         case APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE:
+        case APPINSTALLER_CLI_ERROR_UPGRADE_VERSION_UNKNOWN:
+        case APPINSTALLER_CLI_ERROR_UPGRADE_VERSION_NOT_NEWER:
```
JohnMcPMS, line 203: I'm not sure what the existing interactions look like, but we might need to separate these into new `InstallResultStatus` values as well.

It looks like we didn't split the "expected return codes" out in the `InstallResultStatus`, which itself seems like it might be an issue.

JohnMcPMS (approved):

PR #1871, opened 2022-01-21
[ImgBot] Optimize images

vedantmgoyal2009:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1871)


JohnMcPMS:
Seems fine to me, but I will let @denelon take a look to make sure that the actual images we care about are good.

The images in the subtrees might get stomped in the future, but I don't think we need any of them anyway (could get 100% size reduction by just deleting them really).

denelon:
I'm fine with the space saved here. I don't know if there is a decent way to determine if any of these can safely be deleted or not.

JohnMcPMS:
I think the only and best way would be to simply delete them in a PR and see.

JohnMcPMS (approved):

PR #1881, opened 2022-01-27
Add query or option args check for install and search command

ashpatil-msft:

-----
Issue:
Currently winget install and search commands return a list of all the packages from the sources if no query or filter options are specified. Some sources might or might not support this.

Fix:
Check if winget install or search commands are followed with query or filters and prompt user to specify at least one of those.

Tested both of the commands and ensured proper error message is being displayed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1881)

### DIFF
--- a/src/AppInstallerCLICore/Commands/SearchCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/SearchCommand.cpp
@@ -66,6 +66,18 @@ namespace AppInstaller::CLI
         return "https://aka.ms/winget-command-search";
     }
 
+    void SearchCommand::ValidateArgumentsInternal(Args& execArgs) const
+    {
+        if (!execArgs.Contains(Args::Type::Query) &&
+            !execArgs.Contains(Args::Type::Id) &&
+            !execArgs.Contains(Args::Type::Name) &&
+            !execArgs.Contains(Args::Type::Moniker) &&
+            !execArgs.Contains(Args::Type::Tag) &&
+            !execArgs.Contains(Args::Type::Command))
+        {
+            throw CommandException(Resource::String::NoSearchPackageSpecified);
+        }
+    }
     void SearchCommand::ExecuteInternal(Context& context) const
     {
         context.SetFlags(Execution::ContextFlag::TreatSourceFailuresAsWarning);
--- a/src/AppInstallerCLIE2ETests/GroupPolicy.cs
+++ b/src/AppInstallerCLIE2ETests/GroupPolicy.cs
@@ -29,7 +29,7 @@ namespace AppInstallerCLIE2ETests
         public void PolicyEnableWinget()
         {
             GroupPolicyHelper.EnableWinget.Disable();
-            var result = TestCommon.RunAICLICommand("search", string.Empty);
+            var result = TestCommon.RunAICLICommand("search package", string.Empty);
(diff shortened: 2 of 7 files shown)


yao-msft:
> Currently winget install and search commands return a list of all the packages from the sources if no query or filter options are specified.

I think this is expected behavior when we initially implemented this (i.e. empty query means list all). Maybe mark unsupported sources and return empty if query is IsForEverything() ?

ashpatil-msft:
As far as I know, Demitrius/John wanted to show help whenever install/search command is issued without query/filters for any source. So why not show help to the user instead of empty results, as to why the command didn't work?

src/AppInstallerCLICore/Commands/SearchCommand.cpp
```diff
@@ -66,6 +66,18 @@ namespace AppInstaller::CLI
...
     }
 
+    void SearchCommand::ValidateArgumentsInternal(Args& execArgs) const
+    {
+        if (!execArgs.Contains(Args::Type::Query) &&
+            !execArgs.Contains(Args::Type::Id) &&
+            !execArgs.Contains(Args::Type::Name) &&
+            !execArgs.Contains(Args::Type::Moniker) &&
+            !execArgs.Contains(Args::Type::Tag) &&
+            !execArgs.Contains(Args::Type::Command))
```
ashpatil-msft, line 76: I am not sure if we want to support -n option here as there is a possibility user will specify a large number which may cause querying the entire catalog. Thoughts?

Also, if there are any other options, you'll want me to include/relax in install/search, let me know.

yao-msft:
> As far as I know, Demitrius/John wanted to show help whenever install/search command is issued without query/filters for any source. So why not show help to the user instead of empty results, as to why the command didn't work?

Looks like I was not involved in the discussion of this behavior change. I'm totally fine if this is already agreed upon by the team.

src/AppInstallerCLIE2ETests/GroupPolicy.cs
```diff
@@ -29,7 +29,7 @@ public void TearDown()
         public void PolicyEnableWinget()
         {
             GroupPolicyHelper.EnableWinget.Disable();
-            var result = TestCommon.RunAICLICommand("search", string.Empty);
+            var result = TestCommon.RunAICLICommand("search package", string.Empty);
```
yao-msft, line 32: nit: this would be
var result = TestCommon.RunAICLICommand("search", "package");
Semantically the parameters are (Command, Args)

denelon:
I think we will make some users unhappy if they don't have a way to retrieve the list of "all" packages. We should introduce an option to show all packages here. I'm not sure if the behavior should be the same in the PowerShell cmdlets, but there are users piping the list to some form of output.

How about introducing `winget search --all`?

Note, the cmdlet should be updated as well if we're going to remove the ability to search without a "--query" argument.

src/AppInstallerCLICore/Commands/SearchCommand.cpp
denelon: I added a comment below regarding searching the entire catalog. I think "-n" should still be supported.

Issue #1513, opened 2021-09-27
`ReplacedBy` key for manifests


Trenly:
# Description of the new feature/enhancement
Some programs have multiple versions where a newer version replaces an older version, but both packages are maintained separately. One example of this would be `Corsair.iCUE.3` and `Corsair.iCUE.4`, where they are two separate packages but iCUE.3 is replaced by iCUE.4. Additionally, some older software that is no longer maintained may be picked up by a different user and published under a different package name. If there were a manifest key to indicate that a different package may be available than the package they elected to install, it would be possible to inform users of potential upgrades.

# Proposed technical implementation details (optional)
Add a manifest key for `ReplacedBy: <Package.Identifier>

When running an install or an upgrade to the package, if the manifest contains the ReplacedBy key, display an informational message to the user that a newer version of this software may be available under a different name and display the Package Identifier.  Continue to install the package they requested to install

When running winget list or upgrade with no arguments, display an informational table following the existing output showing which packages may have newer versions available under a different name.

## For Consideration
Would we want to treat such a key as a redirect to the other package for upgrades? If this is something that would be wanted, then winget upgrade should clearly show that new versions will install a different package. Example -
| Name  | Id | Version  | Available | Source  |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| Corsair iCUE Software  | Corsair.iCUE.3  | 3.38.61  | [Corsair.iCUE.4] 4.15.153 | Winget  |
| Snipping Tool | Microsoft.SnippingTool | 1.0.0.1 | [Microsoft.SnipAndSketch] 10.2008.2277.0 |  Winget |

In such cases, it would probably be best to notify the user and have them decide they want to manually run `winget install <Package.Identifier>` to upgrade to the new package

denelon:
@Trenly I just added #1889 does this cover the scenario here well enough?

denelon:
These two issues are cross linked. If you don't mind, close this one and be sure to add your concerns to that one so we can make sure to capture the scenario. I guess the problem or challenge is how one would be able to determine that the newer version would work for them, and what a conditional redirect might look like.

I would be sad if I needed 3.x and mistakenly upgrade to 4.x and my stuff broke.

I would also be sad if 4.x suited my needs better and I wasn't informed.

PR #1894, opened 2022-02-02
Add spec for package pinning.

jedieaston:

-----

Related to #476.

I have written a spec for package pinning, since I thought it was one of the big features where I understood how it had to work. Please rip it to shreds, since this is my first spec here and I didn't know how much detail to go into. I tried to include examples where necessary, but if you want more I can definitely write some! 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1894)

### DIFF
--- a/doc/specs/#476 - Pinned packages.md
+++ b/doc/specs/#476 - Pinned packages.md
@@ -24,11 +24,14 @@ A table of packages (by Name, Version and Product Code if available) that are cu
 
 ## UI/UX Design
 
+Several commands will need to be added to support pinning packages. Where applicable, equivalent names (or new arguments) for PowerShell cmdlets will also be given.
+
 The main interface for the pinning feature will be a new command, `winget pin`, with a couple of options:
 
 
 ```
 winget pin
+PowerShell: Get-WinGetPinnedPackage
 ```
 
 Running `winget pin` with no arguments will show a table of currently pinned packages:
@@ -42,7 +45,8 @@ iTunes                                               Apple.iTunes               
 ```
 
 ```
-winget pin --set <package> 
+winget pin --set <package>
+PowerShell: Set-WinGetPinnedPackage
 ```
 
 `<package>` correlates to a query, similar to `winget list`. When running this command, the package information will be stored in the table in the tracking catalog. This command should be able to accept Package Identifers (checked against all repositories) as well as names, tags, etc. If the package does not have a version number locally but is available from a source, then the version number from the source should be used.
@@ -51,6 +55,7 @@ If there is no package directly available for software a user wants to pin, but 
 
 ```
 winget pin --clear <package>
+PowerShell: Remove-WinGetPinnedPackage
 ```
 
 It functions similarly to `winget pin --set`, but it clears the pin. This should automatically happen on `winget uninstall`, or if it is detected the package is no longer present on the system.
@@ -59,6 +64,7 @@ Enhancements will also need to be made to other commands in the Windows Package 
 
 ```
 winget upgrade --include-pinned
+PowerShell: Get-WinGetPackage -UpgradeAvailable -Pinned
 ```
 
 With the new `--include-pinned` argument, Upgrade should show a separate table in addition to the regular table which shows possible upgrades to pinned packages.
@@ -75,6 +81,7 @@ Clippy for Azure 2011 Datacenter                     Microsoft.Clippy.2011.Datac
 
 ```
 winget <upgrade/install> --include-pinned <package>
+PowerShell: Install-WinGetPackage -Pinned / Upgrade-WinGetPackage -Pinned
 ```
 
 Upgrade and Install will now need to use the same argument to bypass the pin for a package temporarily, in which case the pin is updated to pin the new version after the installation completes.
@@ -84,7 +91,6 @@ winget install --pin <package>
 ```
 
 Install should also be extended to allow the user to pin a package at install time.
-
 
 
 
--- a/doc/specs/#476 - Pinned packages.md
+++ b/doc/specs/#476 - Pinned packages.md
@@ -46,7 +46,7 @@ iTunes                                               Apple.iTunes               
 
 ```
 
-The "Reason" field serves to explain why a application was pinned. If a package designates the `RequiresExplictUpgrade` property in their manifest, it should be explained here so that the user has a way to diagnose why it isn't automatically upgraded during `winget upgrade --all`.
+The "Reason" field serves to explain why a application was pinned. If a package designates the `RequiresExplicitUpgrade` property in their manifest, it should be explained here so that the user has a way to diagnose why it isn't automatically upgraded during `winget upgrade --all`.
 
 ```
 winget pin <package>
--- a/doc/specs/#476 - Pinned packages.md
+++ b/doc/specs/#476 - Pinned packages.md
@@ -157,7 +157,7 @@ There should not be any breaking changes to the code. Although there could be a 
 
 ### Performance, Power, and Efficiency
 
-There should not be any noticible performance changes.
+There should not be any notable performance changes.
 
 
 ## Potential Issues
--- a/doc/specs/#476 - Pinned packages.md
+++ b/doc/specs/#476 - Pinned packages.md
@@ -90,7 +90,7 @@ PowerShell: Install-WinGetPackage / Upgrade-WinGetPackage
 
 Upgrade and Install should work as normal if a pinned package is requested, but print a warning to remind the user that it was previously pinned.
 
-```
+
 ```
 PS C:\Users\billg> winget upgrade Microsoft.Bob
 WARNING: This package was previously pinned at version 2.32.0. The pin will be updated to represent the new version number.
--- a/doc/specs/#476 - Pinned packages.md
+++ b/doc/specs/#476 - Pinned packages.md
@@ -5,9 +5,9 @@ last updated: 2022-02-02
 issue id: 476
 ---
 
-# Pinning a package.
+# Pinning a package
 
-[comment]: # Link to issue: "For [#476](https://github.com/microsoft/winget-cli/issues/476)"
+For [#476](https://github.com/microsoft/winget-cli/issues/476)
 
 ## Abstract
 
@@ -24,8 +24,6 @@ A table of packages (by Name, Version and Product Code if available) that are cu
 
 ## UI/UX Design
 
-[comment]: # What will this fix/feature look like? How will it affect the end user?
-
 The main interface for the pinning feature will be a new command, `winget pin`, with a couple of options:
 
 
@@ -36,7 +34,8 @@ winget pin
 Running `winget pin` with no arguments will show a table of currently pinned packages:
 
 ```
-Name                                                 Id                                Version    ------------------------------------------------------------------------------------------------
+Name                                                 Id                                Version    
+------------------------------------------------------------------------------------------------
 Microsoft Bob                                        Microsoft.Bob                     2.35.0
 iTunes                                               Apple.iTunes                      12.0.199.4
 
@@ -70,7 +69,7 @@ Upgrades to pinned packages:
 
 Name                                                 Id                                Version       Available     Source
 -------------------------------------------------------------------------------------------------------------------------
-Clippy for Azure 2008 Datacenter                     Microsoft.Clippy.2008.Datacenter  2.35.0        2.35.1.2      winget
+Clippy for Azure 2011 Datacenter                     Microsoft.Clippy.2011.Datacenter  2.35.0        2.35.1.2      winget
 ```
 
 
@@ -78,7 +77,7 @@ Clippy for Azure 2008 Datacenter                     Microsoft.Clippy.2008.Datac
 winget <upgrade/install> --include-pinned <package>
 ```
 
-Upgrade and Install will now require arguments to bypass the pin for a package temporarily, in which case the pin is updated to pin the new version after the installation completes.
+Upgrade and Install will now need to use the same argument to bypass the pin for a package temporarily, in which case the pin is updated to pin the new version after the installation completes.
 
 ```
 winget install --pin <package>
@@ -134,56 +133,44 @@ This package requires the following dependencies:
   - Packages
       Microsoft.Bob, version 2.39.0.
 
-Microsoft Bob is currently pinned at version 2.32.0, but Rover the Dog 5.1 requires at least version 2.39.0. To upgrade Microsoft Bob and other pinned dependencies, add the argument `--include-pinned` to your previous command.
+Microsoft Bob is currently pinned at version 2.32.0, but Rover the Dog 5.1 requires at least version 2.39.0. 
+To upgrade Microsoft Bob and other pinned dependencies, add the argument --include-pinned to your previous command.
 ```
 
 ## Capabilities
 
-[comment]: # Discuss how the proposed fixes/features impact the following key considerations:
-
 ### Accessibility
-
-[comment]: # How will the proposed change impact accessibility for users of screen readers, assistive input devices, etc.
 
 Accessibility should not be impacted by this change. There will be a few more tables printed to the terminal in certain cases, but they should use the current table implementation used by `winget upgrade` and `winget list`.
 
 ### Security
 
-[comment]: # How will the proposed change impact security?
-
 Security of the Windows Package Manager should not be impacted by this change. However, security of user's software may be, as if they pin a insecure version of a package they currently will not be notified of important security updates within the Windows Package Manager.
 
 ### Reliability
-
-[comment]: # Will the proposed change improve reliability? If not, why make the change?
 
 The change will improve reliability, as users will be able to have fine grained control of the Windows Package Manager's upgrade functionality to ensure their workflow is not disrupted. 
 
 ### Compatibility
 
-[comment]: # Will the proposed change break existing code/behaviors? If so, how, and is the breaking change "worth it"?
-
 There should not be any breaking changes to the code. Although there could be a mild breaking change to the behavior of `upgrade --all` (not all packages are upgraded anymore since pinned ones are skipped), this is purely opt-in from the user's perspective at this time (if they do not pin software, there should not be a change).
 
 ### Performance, Power, and Efficiency
 
-## Potential Issues
+There should not be any noticible performance changes.
 
-[comment]: # What are some of the things that might cause problems with the fixes/features proposed? Consider how the user might be negatively impacted.
+
+## Potential Issues
 
 Outside of missing security updates and having possible dependency resolution failures due to pins, there should not be any other impacts. Considering the opt-in nature of pinning, it should not affect users unless they want to use it.
 
 ## Future considerations
-
-[comment]: # What are some of the things that the fixes/features might unlock in the future? Does the implementation of this spec enable scenarios?
 
 - Group Policy or MDM control of pinned packages would be a natural addition to the defined functionality. Administrators may want to restrict upgrades for certain pieces of conflicting software if it will cause issues for some or all of their users.
 
 - Packages may want to define pins for their dependencies, if they can't use anything above a certain version of a dependency. These pins could be merged into the user's current pins, making sure that the packages currently installed keep working.
 
 ## Resources
-
-[comment]: # Be sure to add links to references, resources, footnotes, etc.
 
 - [Brew - How do I stop certain formulae from being upgraded?](https://docs.brew.sh/FAQ#how-do-i-stop-certain-formulae-from-being-updated)
 
--- a/doc/specs/#476 - Pinned packages.md
+++ b/doc/specs/#476 - Pinned packages.md
@@ -11,7 +11,7 @@ For [#476](https://github.com/microsoft/winget-cli/issues/476)
 
 ## Abstract
 
-This spec describes the functionality behind "pinning" a package (freezing a package at a certain version so that it will be not be automatically upgraded).
+This spec describes the functionality behind "pinning" a package (freezing a package at a certain version so that it will be not be automatically upgraded via `winget upgrade --all`).
 
 ## Inspiration
 
@@ -98,7 +98,7 @@ In addition to commands, there should be a new `Pinned` key in exported package 
         {
           "PackageIdentifier": "Microsoft.Bob",
           "PackageVersion" : 2.35.0,
-          "Pinned" : "true"
+          "Pinned" : true
         },
       ],
     }
--- a/doc/specs/#476 - Pinned packages.md
+++ b/doc/specs/#476 - Pinned packages.md
@@ -21,6 +21,8 @@ This functionality is inspired by functionality in other package managers, as we
 
 A table of packages (by Name, Version and Product Code if available) that are currently pinned will be kept in the local tracking catalog. To be repository independent, these values should be correlated from local sources (Add and Remove Programs, Appx, etc). This also means that the user should be able to pin packages that were not installed via the Windows Package Manager, so that they won't be upgraded via `winget upgrade --all`. 
 
+Some applications would also like to opt-out of `winget upgrade --all` if they manage their own upgrades or often make breaking changes. As such, the `RequiresExplicitUpgrade` key was added in manifest schema v1.1. Given that this creates the same end result for the user, packages that opt-out of upgrades in this manner should behave like pinned packages to the user, and they should be able to override the behavior listed in the manifest if they choose.
+
 
 ## UI/UX Design
 
@@ -37,15 +39,17 @@ PowerShell: Get-WinGetPinnedPackage
 Running `winget pin` with no arguments will show a table of currently pinned packages:
 
 ```
-Name                                                 Id                                Version    
-------------------------------------------------------------------------------------------------
-Microsoft Bob                                        Microsoft.Bob                     2.35.0
-iTunes                                               Apple.iTunes                      12.0.199.4
+Name                                                 Id                                Version     Reason
+-------------------------------------------------------------------------------------------------------------
+Microsoft Bob                                        Microsoft.Bob                     2.35.0      Manual
+iTunes                                               Apple.iTunes                      12.0.199.4  Via manifest
 
 ```
 
+The "Reason" field serves to explain why a application was pinned. If a package designates the `RequiresExplictUpgrade` property in their manifest, it should be explained here so that the user has a way to diagnose why it isn't automatically upgraded during `winget upgrade --all`.
+
 ```
-winget pin --set <package>
+winget pin <package>
 PowerShell: Set-WinGetPinnedPackage
 ```
 
@@ -58,7 +62,7 @@ winget pin --clear <package>
 PowerShell: Remove-WinGetPinnedPackage
 ```
 
-It functions similarly to `winget pin --set`, but it clears the pin. This should automatically happen on `winget uninstall`, or if it is detected the package is no longer present on the system.
+It functions similarly to `winget pin <package>`, but it clears the pin. This should automatically happen on `winget uninstall`.
 
 Enhancements will also need to be made to other commands in the Windows Package Manager, including:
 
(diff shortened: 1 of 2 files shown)


doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,180 @@
...
+      [
+        {
+          "PackageIdentifier": "Microsoft.Bob",
+          "PackageVersion" : 2.35.0,
+          "Pinned" : "true"
+        },
+      ],
+    }
+// ...
+```
```
vedantmgoyal2009, line 106: Shouldn't the key data type be boolean?

jedieaston: Yes, yes it should. I wasn't 100% awake when I was working on this.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,180 @@
...
+issue id: 476
+---
+
+# Pinning a package
+
+For [#476](https://github.com/microsoft/winget-cli/issues/476)
+
+## Abstract
+
+This spec describes the functionality behind "pinning" a package (freezing a package at a certain version so that it will be not be automatically upgraded).
```
denelon, line 14: I'd clarify that it wouldn't be automatically upgraded via `winget upgrade --all`.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Inspiration
+
+This functionality is inspired by functionality in other package managers, as well as community feedback. Many packages (software development tools and libraries commonly, but other software as well) introduce breaking changes that users may not want integrate into their workflow quite yet.
+
+## Solution Design
+
+A table of packages (by Name, Version and Product Code if available) that are currently pinned will be kept in the local tracking catalog. To be repository independent, these values should be correlated from local sources (Add and Remove Programs, Appx, etc). This also means that the user should be able to pin packages that were not installed via the Windows Package Manager, so that they won't be upgraded via `winget upgrade --all`.
```
denelon, line 22: With the traditional Microsoft Store packages, the Microsoft Store automatically updates programs. Users would not be able to pin those packages. I need to see if there is a way to pin packages from being automatically updated by the Microsoft Store.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,180 @@
...
+```
+Name                                                 Id                                Version    
+------------------------------------------------------------------------------------------------
+Microsoft Bob                                        Microsoft.Bob                     2.35.0
+iTunes                                               Apple.iTunes                      12.0.199.4
+
+```
+
+```
+winget pin --set <package>
```
denelon, line 45: We should cover the intent for the PowerShell cmdlets also. I'd expect several cmdlets. One to show what's pinned, one to pin, and one to "unpin".

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,180 @@
...
+Upgrades to pinned packages: 
+
+Name                                                 Id                                Version       Available     Source
+-------------------------------------------------------------------------------------------------------------------------
+Clippy for Azure 2011 Datacenter                     Microsoft.Clippy.2011.Datacenter  2.35.0        2.35.1.2      winget
+```
+
+
+```
+winget <upgrade/install> --include-pinned <package>
```
denelon, line 77: We're working towards a slightly different behavior for "install" when a version is already on the system. We've been discussing essentially switching to "upgrade" unless the user does something like `winget install <package> --force`

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,180 @@
...
+```
+winget install --pin <package>
+```
+
+Install should also be extended to allow the user to pin a package at install time.
+
+
+
+
+In addition to commands, there should be a new `Pinned` key in exported package lists, allowing users to move their pins to different systems:
```
denelon, line 91: We have several other enhancements to add to the packages.json schema. We've been discussing supporting essentially all the install switches in the schema so running `winget import packages.json` would have as much flexibility as though the user was performing discrete installs. The export command likely will be somewhat limited on what it can detect in terms of switches used during an earlier install or upgrade.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,180 @@
...
+PS C:\Users\billg> winget install --pin Microsoft.RoverTheDog
+Found Rover the Dog [Microsoft.RoverTheDog] Version 5.1
+This application is licensed to you by its owner.
+Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
+This package requires the following dependencies:
+  - Packages
+      Microsoft.Bob, version 2.39.0.
+
+Microsoft Bob is currently pinned at version 2.32.0, but Rover the Dog 5.1 requires at least version 2.39.0. 
+To upgrade Microsoft Bob and other pinned dependencies, add the argument --include-pinned to your previous command.
```
denelon, line 137: I've gone back and forth between just "informing" users and giving an interactive [Y/N] sort of dialog to see if the user wants to perform suggested actions.

denelon (changes requested):
Fantastic! This is almost exactly what we were thinking :)

I know the PowerShell cmdlets aren't ready for prime time, but I do want to treat them as first-class citizens in our specifications. We still need to wire up testing and a build pipeline for them.

Thanks so much for taking the time to write this!!!

doc/specs/#476 - Pinned packages.md
jedieaston: That's a really good idea, but in that case I think the argument should still be present.

doc/specs/#476 - Pinned packages.md
jedieaston: A [Y/N] prompt would be more user-friendly, but it would make it harder to use winget in automation scenarios. Is there a way on Windows to detect if a user is interactively using a shell vs a script? I know the way to tell apt on linux not to ask questions is via an environment variable, but that seems kludgy.

doc/specs/#476 - Pinned packages.md
jedieaston: I was trying to be as repo neutral as possible, but the Microsoft Store will definitely need special treatment. We need to be able to pin packages even if they are coming from REST sources.

doc/specs/#476 - Pinned packages.md
denelon: Any of the [Y/N] style prompts are likely suitable for a setting to specify a default.

doc/specs/#476 - Pinned packages.md
jedieaston: Resolved!

doc/specs/#476 - Pinned packages.md
jedieaston: Resolved!

doc/specs/#476 - Pinned packages.md
vedantmgoyal2009: > I know the way to tell apt on linux not to ask questions is via an environment variable, but that seems kludgy.

Sorry for being a little off topic. Just want to make sure you're talking about `export DEBIAN_FRONTEND=noninteractive`

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+## UI/UX Design
+
+Several commands will need to be added to support pinning packages. Where applicable, equivalent names (or new arguments) for PowerShell cmdlets will also be given.
+
+The main interface for the pinning feature will be a new command, `winget pin`, with a couple of options:
+
+
+```
+winget pin
+PowerShell: Get-WinGetPinnedPackage
```
jedieaston, line 34: @denelon Should this be a separate cmdlet or a argument for `Get-WinGetPackage`? I couldn't decide.

doc/specs/#476 - Pinned packages.md
jedieaston: I went through and added sample PowerShell cmdlet names (or arguments to existing cmdlets, where applicable), if you want to take a peek.

doc/specs/#476 - Pinned packages.md
vedantmgoyal2009: I would be voting for `Get-WinGetPackage -Pinned`😁

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+
+Name                                                 Id                                Version       Available     Source
+-------------------------------------------------------------------------------------------------------------------------
+Clippy for Azure 2011 Datacenter                     Microsoft.Clippy.2011.Datacenter  2.35.0        2.35.1.2      winget
+```
+
+
+```
+winget <upgrade/install> --include-pinned <package>
+PowerShell: Install-WinGetPackage -Pinned / Upgrade-WinGetPackage -Pinned
```
vedantmgoyal2009, line 84: With `Install-WinGetPackage` the switch should be `Pin`? Since the package "will be" pinned rather than "already being pinned".

`Upgrade-WinGetPackage -Pinned`: Will the argument look for updates for "only" pinned packages or "including" pinned packages?

doc/specs/#476 - Pinned packages.md
vedantmgoyal2009: At line 90 (raw), you've written `winget install --pin <package>` but here, it has been written `winget <upgrade/install> --include-pinned <package>`.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+`<package>` correlates to a query, similar to `winget list`. When running this command, the package information will be stored in the table in the tracking catalog. This command should be able to accept Package Identifers (checked against all repositories) as well as names, tags, etc. If the package does not have a version number locally but is available from a source, then the version number from the source should be used.
+
+If there is no package directly available for software a user wants to pin, but it is [included as part of another package](https://github.com/microsoft/winget-cli/issues/1073), the package which includes said piece of software should be treated as pinned unless a newer version doesn't change the version of the pinned software. 
+
+```
+winget pin --clear <package>
+PowerShell: Remove-WinGetPinnedPackage
+```
+
+It functions similarly to `winget pin --set`, but it clears the pin. This should automatically happen on `winget uninstall`, or if it is detected the package is no longer present on the system.
```
JohnMcPMS, line 61: > if it is detected the package is no longer present on the system.

A nice goal, but difficult for us to achieve.  The actual implementation might rely on future installs detecting an abandoned tracking catalog entry and erasing it before installation.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+Clippy for Azure 2011 Datacenter                     Microsoft.Clippy.2011.Datacenter  2.35.0        2.35.1.2      winget
+```
+
+
+```
+winget <upgrade/install> --include-pinned <package>
+PowerShell: Install-WinGetPackage -Pinned / Upgrade-WinGetPackage -Pinned
+```
+
+Upgrade and Install will now need to use the same argument to bypass the pin for a package temporarily, in which case the pin is updated to pin the new version after the installation completes.
```
JohnMcPMS, line 87: I don't think that this is necessary.  The goal is that the standard `upgrade --all` ignores these, but specifically targeted packages are how it should be done.  Placing yet another hurdle seems likely to just trip people rather than actually preventing accidental upgrades.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+          "Pinned" : true
+        },
+      ],
+    }
+// ...
+```
+
+Lastly, there should be good error messaging throughout the Windows Package Manager in order to explain to users when something doesn't go quite right. 
+
+As an example, consider the situation where a user has pinned their favorite package, Microsoft Bob:
```
JohnMcPMS, line 116: > Microsoft Bob

😎

doc/specs/#476 - Pinned packages.md
JohnMcPMS: I've requested that @lechacon do more extensive interactivity controls for another review.  The goal would be to have a common option that disables interactivity, as well as a setting.  Interactivity in general is not a good way to handle these cases, because the COM (and thus PowerShell eventually) entrypoints cannot use that pattern.  But with a more generic and extensive set of controls and thoughts about non-interactive defaults, we can use them more (within reason).

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+winget pin --set <package>
+PowerShell: Set-WinGetPinnedPackage
+```
+
+`<package>` correlates to a query, similar to `winget list`. When running this command, the package information will be stored in the table in the tracking catalog. This command should be able to accept Package Identifers (checked against all repositories) as well as names, tags, etc. If the package does not have a version number locally but is available from a source, then the version number from the source should be used.
+
+If there is no package directly available for software a user wants to pin, but it is [included as part of another package](https://github.com/microsoft/winget-cli/issues/1073), the package which includes said piece of software should be treated as pinned unless a newer version doesn't change the version of the pinned software. 
+
+```
+winget pin --clear <package>
```
JohnMcPMS, line 57: The spec should include discussion of how user driven pinning and `RequiresExplicitUpgrade` interact.  I assert that these are effectively the same thing, with the manifest field driving a default pinned state.  The user should also be able to clear that default pin if they so desire.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+PowerShell: Remove-WinGetPinnedPackage
```
Trenly, line 58: I know this may be very opinion based, but to me, this powershell cmdlet naming makes me think it will uninstall the package that has been pinned. 

I would suggest any of the followng (and encourage discussion)-
`Lock-WinGetPackageVersion` and `Unlock-WinGetPackageVersion` - From [Common verbs](https://docs.microsoft.com/en-us/dotnet/api/System.Management.Automation.VerbsCommon?view=powershellsdk-1.1.0)
`Disable-WinGetPackageUpgrade` and `Enable-WinGetPackageUpgrade` - From [Lifecycle verbs](https://docs.microsoft.com/en-us/dotnet/api/System.Management.Automation.VerbsLifeCycle?view=powershellsdk-1.1.0)
`Suspend-WingetPackageUpgrade` and `Resume-WinGetPackageUpgrade` - From [Lifecycle verbs](https://docs.microsoft.com/en-us/dotnet/api/System.Management.Automation.VerbsLifeCycle?view=powershellsdk-1.1.0)

doc/specs/#476 - Pinned packages.md
Trenly: I would try to remain consistent here, and suggest using `-Pinned` for both

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+### Performance, Power, and Efficiency
+
+There should not be any notable performance changes.
+
+
+## Potential Issues
+
+Outside of missing security updates and having possible dependency resolution failures due to pins, there should not be any other impacts. Considering the opt-in nature of pinning, it should not affect users unless they want to use it.
+
+## Future considerations
```
Trenly, line 173: Another consideration. As verified publishers start adding their packages, they may want to force override pins / require an update. Do we need a provision for this?

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+```
+Name                                                 Id                                Version    
+------------------------------------------------------------------------------------------------
+Microsoft Bob                                        Microsoft.Bob                     2.35.0
+iTunes                                               Apple.iTunes                      12.0.199.4
+
+```
+
+```
+winget pin --set <package>
```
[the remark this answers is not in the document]
jedieaston: What command would you want to use to list pinned packages? I used `winget pin` for that, since it was reminiscent of `winget upgrade`.

doc/specs/#476 - Pinned packages.md
jedieaston: `--pin` is to create a new pin, `--include-pinned` is acknowledging that you want to do something to a package that you've already pinned. Perhaps the distinction isn't relevant.

doc/specs/#476 - Pinned packages.md
denelon: The PowerShell convention is more likely to be discrete cmdlets for pinning and unpinning. Looking at approved PowerShell verbs: https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.2

We may also consider Lock and Unlock, but that might be confusing when compared with the "pin" name.

I've thrown a few suggestions below. I'll also mention @RDMacLachlan to get some input here:

Get-WinGetPinnedPackage - shows packages that have been pinned.
Set-WinGetPinnedPackage - pins a package
Clear-WinGetPinnedPackage - unpins a package

We may also want to consider packages that upgrade themselves as not pinnable.

doc/specs/#476 - Pinned packages.md
Trenly: I think we could just standardize the same way the powershell one is and use `--pinned`

doc/specs/#476 - Pinned packages.md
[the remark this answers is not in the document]
denelon: That does seem consistent with upgrade.

doc/specs/#476 - Pinned packages.md
jedieaston: I believe some kind of updating will be necessary of those values regardless, as if a package is pinned because it manages updates itself, whenever the update happens the pinned version in the table will be out of sync with reality.

I'm happy to remove it for now, but it is something to think about.

denelon: I'd put it in future considerations. When the user displays what is pinned or an individual package that is pinned, we could perform some kind of check, but I agree it's likely future work.

doc/specs/#476 - Pinned packages.md
jedieaston: My opinion would be that publishers shouldn't be able to override pins via the manifests, as the main reason a user would want to pin a package is because they disagree with the ISVs patching schedule. If they wanted to get around it they could add an automatic updater to their piece of software (as many have chosen to do).

doc/specs/#476 - Pinned packages.md
Trenly: I had a few other suggestions below earlier; I like `Clear` more than `Remove`. I know you already mentioned lock/unlock, but I think different nouns could be used.

`Lock-WinGetPackageVersion` and `Unlock-WinGetPackageVersion` - From [Common verbs](https://docs.microsoft.com/en-us/dotnet/api/System.Management.Automation.VerbsCommon?view=powershellsdk-1.1.0)
`Disable-WinGetPackageUpgrade` and `Enable-WinGetPackageUpgrade` - From [Lifecycle verbs](https://docs.microsoft.com/en-us/dotnet/api/System.Management.Automation.VerbsLifeCycle?view=powershellsdk-1.1.0)
`Suspend-WingetPackageUpgrade` and `Resume-WinGetPackageUpgrade` - From [Lifecycle verbs](https://docs.microsoft.com/en-us/dotnet/api/System.Management.Automation.VerbsLifeCycle?view=powershellsdk-1.1.0)

Note: After more consideration, I do like Set and Clear the best, since it maintains consistency

doc/specs/#476 - Pinned packages.md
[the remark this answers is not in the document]
Trenly: I don't think it would have to be a specific command; I think it can function exactly like upgrade, where `winget pin` would list the pinned packages and `winget pin <package>` would set the pin on that package

doc/specs/#476 - Pinned packages.md
Trenly: Maybe this is an opportunity to look at what exactly the RequiresExplicitUpgrade field is supposed to show. If it is intended to show the default pinned state, I would suggest we consider renaming the field to something like `Pinning:` and make it an enum rather than a boolean so that packages could be marked as unable to be pinned without requiring a separate field.

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Potential Issues
+
+Outside of missing security updates and having possible dependency resolution failures due to pins, there should not be any other impacts. Considering the opt-in nature of pinning, it should not affect users unless they want to use it.
+
+## Future considerations
+
+- Group Policy or MDM control of pinned packages would be a natural addition to the defined functionality. Administrators may want to restrict upgrades for certain pieces of conflicting software if it will cause issues for some or all of their users.
+
+- Packages may want to define pins for their dependencies, if they can't use anything above a certain version of a dependency. These pins could be merged into the user's current pins, making sure that the packages currently installed keep working.
```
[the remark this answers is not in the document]
Trenly: I would be in favor of a maximum version, since if a package requires a specific version then the minimum and maximum could be set to the same thing. This isn't a rare scenario, but it certainly isn't as common as a minimum version.

doc/specs/#476 - Pinned packages.md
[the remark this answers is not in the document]
jedieaston: Ah, I see. `winget pin` with no args shows the table, `winget pin <package query>` sets the pin, and `winget pin --clear <package query>` unpins. That would mean you wouldn't be able to filter the pin table, which probably doesn't happen that often but would make it inconsistent with install/list/upgrade.

doc/specs/#476 - Pinned packages.md
jedieaston: > unable to be pinned

That may not be necessary, since winget can't guarantee that a package won't be upgraded anyway, it can only guarantee that `winget upgrade` won't upgrade the package without being explicitly asked.

Trenly: > > unable to be pinned
> 
> That may not be necessary, since winget can't guarantee that a package won't be upgraded anyway, it can only guarantee that `winget upgrade` won't upgrade the package without being explicitly asked.

True, but it would be an opportunity to keep the user informed if we know of specific packages that auto-upgrade. Just an idea, but probably outside the realm of this spec anyways

doc/specs/#476 - Pinned packages.md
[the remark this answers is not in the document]
JohnMcPMS: > Ah, I see. `winget pin` with no args shows the table, `winget pin <package query>` sets the pin, and `winget pin --clear <package query>` unpins. That would mean you wouldn't be able to filter the pin table, which probably doesn't happen that often but would make it inconsistent with install/list/upgrade.

That would be consistent with `upgrade`, but not `list`.  An option to allow filtering would be `winget pin <query> --list`.  Considering it is a modifying command, I think it makes sense to remain consistent with `upgrade`.  `--list` could even be added to `upgrade`, although that behavior is already equivalently exposed via `list` itself.

doc/specs/#476 - Pinned packages.md
denelon: RequiresExplicitUpgrade was about apps that update themselves and was something added to the manifest to somewhat declutter `winget upgrade` for packages that handle upgrades themselves or for ISVs to prefer their own upgrade mechanism. https://github.com/microsoft/winget-cli/issues/1163. The description in the schema probably shouldn't use the word "pinned" as pinning is more of a user driven behavior. Although logically it causes apps to be excluded from `winget upgrade --all` which pinning also does.

Pinning a package is more about a user saying, "I don't want this upgraded". It's often the case for enterprises who have to support specific versions of applications, and developers wanting to have specific versions of packages for their needs.

We may also need to consider the case where packages won't be able to support pinning. Some packages are going to update no matter what "we" want. This could help us to inform users when we can't control the software upgrade mechanism.

denelon:
@Trenly and I were discussing another scenario in #1513 

Pinning may need to be a bit more sophisticated in light of the possibility of redirecting manifests.

I might be OK with pinning a part of a version. If I'm on 3.x and I'm happy with upgrades, but I wouldn't want to get redirected to an upgrade for 4.x I'd like to be able to pin part of a version. If we use [semver](https://semver.org) as an example, I might want to pin to 3.x or even 3.8.x, but nothing higer.

Trenly:
> @Trenly and I were discussing another scenario in #1513
> 
> Pinning may need to be a bit more sophisticated in light of the possibility of redirecting manifests.
> 
> I might be OK with pinning a part of a version. If I'm on 3.x and I'm happy with upgrades, but I wouldn't want to get redirected to an upgrade for 4.x I'd like to be able to pin part of a version. If we use [semver](https://semver.org) as an example, I might want to pin to 3.x or even 3.8.x, but nothing higer.

Speaking of this - and I'm just thought dumping here - What if a package is side-by-side compatible? Would pinning also prevent the installation of other versions if a package is required as a dependency? The Visual C++ Redistributables are a good example of this. I could pin `Microsoft.VC++2015-2019Redist-x64` at version 14.28.29913.0, but if a package requires a newer version, it won't uninstall the pinned version. What if it explicitly requires an older version?

I think we have considered the idea of a package which upgrades in place thoroughly, but I don't know if the cases for side-by-side apps have been considered, or if those would even need pinning as I believe the upgrade behavior is still being defined?

doc/specs/#476 - Pinned packages.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Solution Design
+
+A table of packages (by Name, Version and Product Code if available) that are currently pinned will be kept in the local tracking catalog. To be repository independent, these values should be correlated from local sources (Add and Remove Programs, Appx, etc). This also means that the user should be able to pin packages that were not installed via the Windows Package Manager, so that they won't be upgraded via `winget upgrade --all`. 
+
+Some applications would also like to opt-out of `winget upgrade --all` if they manage their own upgrades or often make breaking changes. As such, the `RequiresExplicitUpgrade` key was added in manifest schema v1.1. Given that this creates the same end result for the user, packages that opt-out of upgrades in this manner should behave like pinned packages to the user, and they should be able to override the behavior listed in the manifest if they choose.
```
jedieaston, line 24: Does everyone like this RequiresExplicitUpgrade behavior? I believe it continues the trend of "sane defaults, but we'll let you break it"

doc/specs/#476 - Pinned packages.md
jedieaston: Resolved in latest commit. I said that it should warn that you are modifying a pinned package, but continue anyway.

doc/specs/#476 - Pinned packages.md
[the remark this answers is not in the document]
jedieaston: Updated in latest commit. `winget pin` with no args shows the list, `winget pin <package>` pins, and `winget pin --clear <package>` removes the pin.

doc/specs/#476 - Pinned packages.md
denelon: That aligns with my desire to have safe and rational defaults, but to let "power users" drive the behavior they want with arguments and/or settings. Developers know what they want, and we should let them determine how the Windows Package Manger should behave. We should call out some of the concerns like unintended consequences when updating a package that may be breaking to other things installed on their system.

denelon:
@jedieaston would you mind converting this to a draft while we iterate?

jedieaston:
Yep! Should’ve done that to start with, sorry.

doc/specs/#476 - Pinned packages.md
RichiCoder1: At least with chocolatey, I ended up pinning for two reasons:
* I really want a specific version
* The application has it's own update mechanism

So the behavior of "the package may still auto update, and is just excluded from `winget upgrade --all` makes a lot of sense to me.

jedieaston:
Closing in favor of #2611 :)

PR #1909, opened 2022-02-08
Add uninstall functionality to Com api

yao-msft:
## Change
Added uninstall functionality to Com api

## Validation
Since the Com packaged tests are not ready yet, I manually verified the newly added uninstall api works on packages from winget source and msstore rest source.

Next I'll work on enabling Com tests as part of the e2e tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1909)


src/AppInstallerCLICore/ContextOrchestrator.cpp
```diff
@@ -23,6 +26,17 @@ namespace AppInstaller::CLI::Execution
                 queue->RunItem(queueItem->GetId());
             }
         }
+
+        // Get command queue name based on command name.
+        std::string_view GetCommandQueueName(std::string_view commandName)
+        {
+            if (commandName == "install" || commandName == "uninstall")
```
JohnMcPMS, line 33: These strings should come from the command via a static function rather than be copied here.

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -48,6 +48,8 @@
               </com:Class>
               <com:Class Id ="3F85B9F4-487A-4C48-9035-2903F8A6D9E8" DisplayName="PackageMatchFilter Server">
               </com:Class>
+              <com:Class Id ="AA2A5C04-1AD9-46C4-B74F-6B334AD7EB8C" DisplayName="UninstallOptions Server">
```
JohnMcPMS, line 51: Reminder that we need to add it to the other package on pull.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -87,6 +87,68 @@ namespace Microsoft.Management.Deployment
         }
     }
 
+    /// State of the uninstall.
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 4)]
+    enum PackageUninstallProgressState
+    {
+        /// The uninstall is queued but not yet active. Cancellation of the IAsyncOperationWithProgress in this 
+        /// state will prevent the package from downloading or installing.
```
JohnMcPMS, line 95:
```suggestion
        /// state will prevent the package from uninstalling.
```

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -87,6 +87,68 @@ namespace Microsoft.Management.Deployment
...
+        Uninstalling,
+        /// The uninstaller has completed and cleanup actions are in progress. Cancellation of the 
+        /// IAsyncOperationWithProgress in this state will not stop cleanup or roll back the uninstall.
+        PostUninstall,
+        /// The operation has completed.
+        Finished,
+    };
+
+    /// Progress object for the uninstall
+    /// DESIGN NOTE: percentage for the uninstall as a whole is purposefully not included as there is no way to
```
JohnMcPMS, line 108: Remove copied design note as there is a progress value that would satisfy this (if there is any point at which it actually means anything).

JohnMcPMS (approved):

JohnMcPMS:
Forgot to add that you should update the sample caller app to include uninstall and integration with the progress.

yao-msft:
> Forgot to add that you should update the sample caller app to include uninstall and integration with the progress.

I can do that in a separate pr later.

src/AppInstallerCLIPackage/Package.appxmanifest
yao-msft: Yep, I'll do that on next pull.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -87,6 +87,66 @@ namespace Microsoft.Management.Deployment
...
+        CatalogError,
+        InternalError,
+        InvalidOptions,
+        UninstallError,
+        ManifestError,
+    };
+
+    /// Result of the install
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 4)]
+    runtimeclass UninstallResult
```
ranm-msft, line 133: >UninstallResult [](http://example.com/codeflow?start=17&length=15)

we should make sure when this feature gets checked in we FCIB the DesktopAppInstaller package to SV3, so that dependent callers (backup / restore) are not broken

Issue #1925, opened 2022-02-10
winget --info should print the system architecture


jedieaston:
### Description of the new feature / enhancement

I would like it if `winget --info` printed what the system's architecture is, so I don't have to ask the user when I'm looking at a bug report. Sometimes I'm wondering why a odd match happened, and this narrows it down a lot.

### Proposed technical implementation details

`context.Reporter.Info() << "Architecture :                                      " << architecture << endl` 

(obviously with the fancy spacing function and the function to get architecture. this is not indicative of the actual work required 🙂)

PR #1937, opened 2022-02-14
Added system architecture to `winget --info`.

jedieaston:

-----

Resolves https://github.com/microsoft/winget-cli/issues/1925 (yes, I know it was my issue. just had to leave myself a reminder :D)

This PR adds the system architecture to `winget --info`, cutting out the first question we usually ask when troubleshooting.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1937)

### DIFF
--- a/src/AppInstallerCLICore/Commands/RootCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/RootCommand.cpp
@@ -173,7 +173,7 @@ namespace AppInstaller::CLI
             info << std::endl <<
                 "Windows: "_liv << Runtime::GetOSVersion() << std::endl;
 
-            info << Resource::String::SystemArchitecture << ": "_liv << Utility::LocIndView(Utility::ToString(Utility::GetSystemArchitecture())) << std::endl;
+            info << Resource::String::SystemArchitecture << ": "_liv << Utility::ToString(Utility::GetSystemArchitecture()) << std::endl;
 
             if (Runtime::IsRunningInPackagedContext())
             {
--- a/src/AppInstallerCLICore/Commands/RootCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/RootCommand.cpp
@@ -173,6 +173,8 @@ namespace AppInstaller::CLI
             info << std::endl <<
                 "Windows: "_liv << Runtime::GetOSVersion() << std::endl;
 
+            info << Resource::String::Architecture << ": "_liv << Utility::ToString(Utility::GetSystemArchitecture()) << std::endl;
+
             if (Runtime::IsRunningInPackagedContext())
             {
                 info << Resource::String::Package << ": "_liv << Runtime::GetPackageVersion() << std::endl;
--- a/src/AppInstallerCLICore/Commands/RootCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/RootCommand.cpp
@@ -173,7 +173,7 @@ namespace AppInstaller::CLI
             info << std::endl <<
                 "Windows: "_liv << Runtime::GetOSVersion() << std::endl;
 
-            info << Resource::String::Architecture << ": "_liv << Utility::ToString(Utility::GetSystemArchitecture()) << std::endl;
+            info << Resource::String::SystemArchitecture << ": "_liv << Utility::LocIndView(Utility::ToString(Utility::GetSystemArchitecture())) << std::endl;
 
             if (Runtime::IsRunningInPackagedContext())
             {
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1272,4 +1272,7 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="InvalidArgumentWithoutQueryError" xml:space="preserve">
     <value>The arguments provided can only be used with a query.</value>
   </data>
+  <data name="Architecture" xml:space="preserve">
+    <value>Architecture</value>
```
JohnMcPMS, line 1276:
```suggestion
    <value>System Architecture</value>
```

src/AppInstallerCLICore/Commands/RootCommand.cpp
```diff
@@ -173,6 +173,8 @@ namespace AppInstaller::CLI
             info << std::endl <<
                 "Windows: "_liv << Runtime::GetOSVersion() << std::endl;
 
+            info << Resource::String::Architecture << ": "_liv << Utility::ToString(Utility::GetSystemArchitecture()) << std::endl;
```
yao-msft, line 176: We haven't enforced the output rule yet, but I think the Utility::ToString(Utility::GetSystemArchitecture()) needs to be wrapped inside LocIndString

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1272,4 +1272,7 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="InvalidArgumentWithoutQueryError" xml:space="preserve">
     <value>The arguments provided can only be used with a query.</value>
   </data>
+  <data name="Architecture" xml:space="preserve">
```
yao-msft, line 1275: nit: shall we match the variable name with the value so we don't misuse this in the future?

src/AppInstallerCLICore/Commands/RootCommand.cpp
JohnMcPMS: `Utility::ToString(Architecture)` could probably be changed to return a loc independent view.  If it turns out we need to localize it in the future, that can be its own problem.

jedieaston:
Sorry for the delay, I made it localization independent when the command is called. I tried to do as @JohnMcPMS suggested and just change the type to LocIndView, but since LocIndView is also defined inside of that project, it seemed to create a circular dependency somehow (although since LocIndView wasn't relying on anything in AppInstallerArchitecture.h, I'm not sure why). Does anyone have any ideas on how to fix that?

yao-msft:
> Sorry for the delay, I made it localization independent when the command is called. I tried to do as @JohnMcPMS suggested and just change the type to LocIndView, but since LocIndView is also defined inside of that project, it seemed to create a circular dependency somehow (although since LocIndView wasn't relying on anything in AppInstallerArchitecture.h, I'm not sure why). Does anyone have any ideas on how to fix that?

I just tried adding `#include "winget/LocIndependent.h"` to AppInstallerArchitectures.h and modified the declaration and definition of the method. I did not see circular dependency in rebuilding. If somehow this still won't work for you, I'm fine with merging with current code. Just let me know. Thanks.

jedieaston:
It was the missing header :(

Thanks @yao-msft.

src/AppInstallerCommonCore/Architecture.cpp
```diff
@@ -186,23 +186,23 @@ namespace AppInstaller::Utility
...
     }
 
-    std::string_view ToString(Architecture architecture)
+    LocIndView ToString(Architecture architecture)
     {
         switch (architecture)
         {
         case Architecture::Neutral:
-            return "Neutral"sv;
+            return LocIndView("Neutral"sv);
```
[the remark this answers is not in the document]
jedieaston: I tried to use them but was missing the namespace. Oops, my header game is weak this week.

denelon:
Does this work in the Sandbox? I submitted an [Issue](https://github.com/NickeManarin/ScreenToGif/issues/1075#issuecomment-1039749711) for "Screen To Gif". I wanted to make sure we're checking the host architecture, for this scenario or at least be aware of limitations.

yao-msft (approved):

jedieaston:
> Does this work in the Sandbox? I submitted an [Issue](https://github.com/NickeManarin/ScreenToGif/issues/1075#issuecomment-1039749711) for "Screen To Gif". I wanted to make sure we're checking the host architecture, for this scenario or at least be aware of limitations.

Sorry I didn't get to this sooner, yes it works in the Sandbox.

PR #1948, opened 2022-02-16
Add support for installRoot and addToPath install behavior in settings

ryfu-msft:
Changes:
- Add support for InstallRoot and AddToPathEnvironment fields to settings.
- Added validation of InstallRoot value to check if the provided path already exists. If not, create the directories

Tests:
- Added tests to verify that the values for the newly added fields are processed successfully. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1948)

### DIFF
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -79,8 +79,8 @@
       "type": "string",
       "default":  "%LOCALAPPDATA%/WinGet"
     },
-    "AddToPathVariable": {
-      "description": "Boolean value indicating whether the installed location should be added to the PATH environment variable. Applies to the portable installer type.",
+    "AddToPathEnvironment": {
+      "description": "Boolean value indicating whether the installed location should be added to the PATH environment. Applies to the portable installer type.",
       "type": "boolean",
       "default": true
     },
@@ -90,8 +90,8 @@
       "properties": {
         "preferences": { "$ref": "#/definitions/InstallPrefReq" },
         "requirements": { "$ref": "#/definitions/InstallPrefReq" },
-        "binaryFilesInstallLocation": { "$ref": "#/definitions/BinaryFilesInstallLocation" },
-        "addToPathEnvironmentVariable": { "$ref": "#/definitions/AddToPathEnvironmentVariable" }
+        "binaryFilesInstallLocation": { "$ref": "#/definitions/InstallRoot" },
+        "addToPathEnvironmentVariable": { "$ref": "#/definitions/AddToPathEnvironment" }
       }
     },
     "Telemetry": {
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -74,12 +74,24 @@
         }
       }
     },
+    "InstallRoot": {
+      "description": "The root directory where packages are installed to. Applies to the portable installer type.",
+      "type": "string",
+      "default":  "%LOCALAPPDATA%/WinGet"
+    },
+    "AddToPathVariable": {
+      "description": "Boolean value indicating whether the installed location should be added to the PATH environment variable. Applies to the portable installer type.",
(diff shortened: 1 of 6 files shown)


schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -91,12 +91,24 @@
         }
       }
     },
+    "InstallRoot": {
+      "description": "The root directory where packages are installed to. Applies to the portable installer type.",
```
JohnMcPMS, line 95: We need a blanket term that applies to all packages that WPM is taking ownership of the install action.  No idea what it is.

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -91,12 +91,24 @@
         }
       }
     },
+    "InstallRoot": {
```
JohnMcPMS, line 94: Most likely we will need two; one for user scope and one for machine scope.

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -91,12 +91,24 @@
...
+      "description": "The root directory where packages are installed to. Applies to the portable installer type.",
+      "type": "string",
+      "default":  "%LOCALAPPDATA%/WinGet"
```
JohnMcPMS, line 97: `%LOCALAPPDATA%\WPM_Packages` for user scope and `???` for machine scope.  If we are planning on supporting environment variables, we should call that out.

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -91,12 +91,24 @@
...
+    },
+    "AddToPathEnvironment": {
+      "description": "Boolean value indicating whether the installed location should be added to the PATH environment. Applies to the portable installer type.",
```
JohnMcPMS, line 100: I thought we were adding to the App Path registry?  I think we should hold off on supporting %PATH% updates to see if App Path is good enough.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -130,6 +133,8 @@ namespace AppInstaller::Settings
         SETTINGMAPPING_SPECIALIZATION(Setting::NetworkDOProgressTimeoutInSeconds, uint32_t, std::chrono::seconds, 60s, ".network.doProgressTimeoutInSeconds"sv);
         SETTINGMAPPING_SPECIALIZATION(Setting::InstallLocalePreference, std::vector<std::string>, std::vector<std::string>, {}, ".installBehavior.preferences.locale"sv);
         SETTINGMAPPING_SPECIALIZATION(Setting::InstallLocaleRequirement, std::vector<std::string>, std::vector<std::string>, {}, ".installBehavior.requirements.locale"sv);
+        SETTINGMAPPING_SPECIALIZATION(Setting::InstallRoot, std::string, std::filesystem::path, {}, ".installBehavior.installRoot"sv);
```
JohnMcPMS, line 136: This should remain a string through settings, as both support for environment variable expansion and properly creating the directories will need to be handled with care (AKA not just jammed into settings).

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -229,9 +229,32 @@ namespace AppInstaller::Settings
...
 
+        WINGET_VALIDATE_SIGNATURE(InstallRoot)
+        {
+            const std::filesystem::path installRoot = value;
+
+            if (!std::filesystem::exists(installRoot))
+            {
+                try
+                {
+                    if (!std::filesystem::create_directories(installRoot))
```
JohnMcPMS, line 244: I don't think this ends well if the first time is via an elevated context and subsequent attempts are not.  We need special code to properly create directories that can be modified by the medium IL processes that will eventually follow.

PR #1986, opened 2022-03-01
Remove duplicated message when no upgrades

felipecrs:

-----

This is kind of a follow up of https://github.com/microsoft/winget-cli/pull/1982, continuing the cleanup. But this is a bit different because it also changes one message.

This is how it was before:

```console
➜ winget upgrade --all
No installed package found matching input criteria.
10 packages have version numbers that cannot be determined. Use "--include-unknown" to see all results.
No applicable update found.
```

This is how it is now:

```console
➜ winget upgrade --all
No applicable update found.
10 packages have version numbers that cannot be determined. Use "--include-unknown" to see all results.
```

Please let me know what you think.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1986)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
+++ b/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
@@ -759,12 +759,7 @@ namespace AppInstaller::CLI::Workflow
 
         if (table.IsEmpty())
         {
-            if (m_onlyShowUpgrades) {
-                context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl;
-            }
-            else {
-                context.Reporter.Info() << Resource::String::NoInstalledPackageFound << std::endl;
-            }
+            context.Reporter.Info() << Resource::String::NoInstalledPackageFound << std::endl;
         }
         else
         {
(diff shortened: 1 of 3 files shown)


src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -759,7 +759,12 @@ namespace AppInstaller::CLI::Workflow
 
         if (table.IsEmpty())
         {
-            context.Reporter.Info() << Resource::String::NoInstalledPackageFound << std::endl;
+            if (m_onlyShowUpgrades) {
+                context.Reporter.Info() << Resource::String::UpdateNotApplicable << std::endl;
```
yao-msft, line 763: I think we still want to differentiate "app not installed" vs "update not applicable". This change seems ok with winget upgrade --all, but we want clearer message for winget upgrade <specific package>

felipecrs: Got it. I restored the original message, but at least removed the duplication. What do you think?

yao-msft (approved):

Issue #1891, opened 2022-02-01
Windows Package Manager 1.2 manifest schema


denelon:
### Description of the new feature / enhancement

This is a parent issue to track outstanding schema requests for the 1.2 schema. 
Inclusion in this list is **not a commitment**. Other schema items may be added or removed.

>Note: We will be targeting the 1.2 schema with the 1.3 client.

* #1994
* #1984
* #1916
* #1889
* #1888 
* #1345 
* #1277
* #607

Related to:
* #1145
* #970 
* #182
* #140
* microsoft/winget-cli-restsource/issues/118

Also related to schema, but not implemented or targeted for the 1.3 client release:
* #163 
* #147

denelon:
See:
* #1997

Trenly:
@denelon - Is this completed?

denelon:
LOL! Yes, this was complete, I was just looking at the related open issues and saw the one at the REST source that isn't marked complete, and the features enabled by the schema that still haven't been implemented.

PR #1989, opened 2022-03-01
1.2 Schema Proposal for Portable and Zip InstallerTypes

ryfu-msft:
Related to:
* #182 
* #140 
* #1891

This PR adds the necessary schema changes to support installing portable apps and zip files. 

Changes:

<ins>**Settings schema:**</ins>
- Addition of ~~WinGetInstallerUserRoot~~ _"PortablePackageUserRoot"_ to allow the user to specify the default root directory where portable apps should be installed to under User scope.
- Addition of ~~WinGetInstallerMachineRoot~~ _"PortablePackageMachineRoot"_ to allow the user to specify the default root directory where portable apps should be installed to under Machine scope.

<ins>**InstallerManifest schema:**</ins>

Changes to 'Installer' object:
- Addition of  ~~NestedInstallers~~ _"ArchiveInstallers"_ which is an object array of ~~NestedInstaller~~ _"ArchiveInstaller"_
- Addition of _"AppendPathToProcess"_ which is utilized during the installation of portable apps. We will be registering the app using the [App Paths registry subkey](https://docs.microsoft.com/en-us/windows/win32/shell/app-registration#using-the-app-paths-subkey). This value would be used to determine whether we should create a ‘Path’ string value in the registry to indicate whether to pre-pend the ‘Path’ value to the PATH environment variable on a per-application, per-process basis

Addition of ~~NestedInstaller~~ _"ArchiveInstaller"_ object:
- Similar to the existing 'Installer' object but more refined to fit the zip scenario. 
- Addition of _"RelativeFilePath"_ to declare the relative file path to the installer/portable exe that will be installed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1989)

### DIFF
--- a/schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
+++ b/schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
@@ -63,6 +63,12 @@
       ],
       "description": "Enumeration of supported installer types. InstallerType is required in either root level or individual Installer level"
     },
+    "InstallerUrl": {
+      "type": "string",
+      "pattern": "^([Hh][Tt][Tt][Pp][Ss]?)://.+$",
+      "maxLength": 2048,
+      "description": "The installer Url"
+    },    
     "InstallerSha256": {
       "type": "string",
       "pattern": "^[A-Fa-f0-9]{64}$",
@@ -72,6 +78,17 @@
       "type": [ "string", "null" ],
       "pattern": "^[A-Fa-f0-9]{64}$",
       "description": "SignatureSha256 is recommended for appx or msix. It is the sha256 of signature file inside appx or msix. Could be used during streaming install if applicable"
+    },   
+    "Architecture": {
+      "type": "string",
+      "enum": [
+        "x86",
+        "x64",
+        "arm",
+        "arm64",
+        "neutral"
+      ],
+      "description": "The installer target architecture"
     },   
     "Scope": {
       "type": [ "string", "null" ],
@@ -463,23 +480,15 @@
           "$ref": "#/definitions/MinimumOSVersion"
         },
         "Architecture": {
-          "type": "string",
-          "enum": [
-            "x86",
-            "x64",
-            "arm",
-            "arm64",
-            "neutral"
-          ],
-          "description": "The installer target architecture"
+          "$ref": "#/definitions/Architecture"
         },
         "InstallerType": {
           "$ref": "#/definitions/InstallerType"
         },
-        "NestedInstallers": {
+        "ArchiveInstallers": {
           "type": "array",
           "items": {
-            "$ref": "#/definitions/NestedInstaller"
+            "$ref": "#/definitions/ArchiveInstaller"
           },
           "minItems": 1,
           "maxItems": 1024
@@ -491,10 +500,7 @@
           "$ref": "#/definitions/Scope"
         },
         "InstallerUrl": {
-          "type": "string",
-          "pattern": "^([Hh][Tt][Tt][Pp][Ss]?)://.+$",
-          "maxLength": 2048,
-          "description": "The installer Url"
+          "$ref": "#/definitions/InstallerUrl"
         },
         "InstallerSha256": {
           "$ref": "#/definitions/InstallerSha256"
@@ -572,30 +578,19 @@
         "InstallerSha256"
       ]
     },
-    "NestedInstaller": {
+    "ArchiveInstaller": {
       "type": "object",
       "properties": {
-        "InstallerLocale": {
-          "$ref": "#/definitions/Locale"
+        "allOf": {
+          "$ref": "#/definitions/Installer"
         },
-        "Platform": {
-          "$ref": "#/definitions/Platform"
+        "InstallerUrl": {
+          "not": {}
         },
-        "MinimumOSVersion": {
-          "$ref": "#/definitions/MinimumOSVersion"
+        "InstallerType": {
+          "not": {}
         },
-        "Architecture": {
-          "type": "string",
-          "enum": [
-            "x86",
-            "x64",
-            "arm",
-            "arm64",
-            "neutral"
-          ],
-          "description": "The installer target architecture"
-        },
-        "NestedInstallerType": {
+        "ArchiveInstallerType": {
           "type": [ "string", "null" ],
           "enum": [
             "msix",
@@ -610,90 +605,15 @@
           ],
           "description": "Enumeration of supported nested installer types that could be contained in a zip or archive file."
         },
-        "AppendPathAtProcessStart": {
-          "$ref": "#/definitions/AppendPathAtProcessStart"
-        },
         "RelativeFilePath": {
           "type": [ "string", "null" ],
           "description": "Relative path to the installer file or portable executable. Only applies to the nested installer within a Zip. "
-        },  
-        "Scope": {
-          "$ref": "#/definitions/Scope"
-        },
-        "InstallerSha256": {
-          "$ref": "#/definitions/InstallerSha256"
-        },
-        "SignatureSha256": {
-          "$ref": "#/definitions/SignatureSha256"
-        },
-        "InstallModes": {
-          "$ref": "#/definitions/InstallModes"
-        },
-        "InstallerSwitches": {
-          "$ref": "#/definitions/InstallerSwitches"
-        },
-        "InstallerSuccessCodes": {
-          "$ref": "#/definitions/InstallerSuccessCodes"
-        },
-        "ExpectedReturnCodes": {
-          "$ref": "#/definitions/ExpectedReturnCodes"
-        },
-        "UpgradeBehavior": {
-          "$ref": "#/definitions/UpgradeBehavior"
-        },
-        "Commands": {
-          "$ref": "#/definitions/Commands"
-        },
-        "Protocols": {
-          "$ref": "#/definitions/Protocols"
-        },
-        "FileExtensions": {
-          "$ref": "#/definitions/FileExtensions"
-        },
-        "Dependencies": {
-          "$ref": "#/definitions/Dependencies"
-        },
-        "PackageFamilyName": {
-          "$ref": "#/definitions/PackageFamilyName"
-        },
-        "ProductCode": {
-          "$ref": "#/definitions/ProductCode"
-        },
-        "Capabilities": {
-          "$ref": "#/definitions/Capabilities"
-        },
-        "RestrictedCapabilities": {
-          "$ref": "#/definitions/RestrictedCapabilities"
-        },
-        "Markets": {
-          "$ref": "#/definitions/Markets"
-        },
-        "InstallerAbortsTerminal": {
-          "$ref": "#/definitions/InstallerAbortsTerminal"
-        },
-        "ReleaseDate": {
-          "$ref": "#/definitions/ReleaseDate"
-        },
-        "InstallLocationRequired": {
-          "$ref": "#/definitions/InstallLocationRequired"
-        },
-        "RequireExplicitUpgrade": {
-          "$ref": "#/definitions/RequireExplicitUpgrade"
-        },
-        "UnsupportedOSArchitectures": {
-          "$ref": "#/definitions/UnsupportedOSArchitectures"
-        },
-        "AppsAndFeaturesEntries": {
-          "$ref": "#/definitions/AppsAndFeaturesEntries"
-        },
-        "ElevationRequirement": {
-          "$ref": "#/definitions/ElevationRequirement"
         }
       },
       "required": [
         "Architecture",
         "InstallerSha256",
-        "NestedInstallerType"
+        "ArchiveInstallerType"
       ]
     }
   },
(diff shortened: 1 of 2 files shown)


schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -550,6 +571,130 @@
...
+          "enum": [
+            "msix",
+            "msi",
+            "appx",
+            "exe",
+            "inno",
+            "nullsoft",
+            "wix",
+            "burn",
+            "portable"
```
JohnMcPMS, line 609: This is actually the correct PR for it, but since I'm thinking of it in the context of archives anyway...

I think we need to consider a `portableWithInit` type (or some other way of achieving it) where the package installer files are intended to stay on the device like portable, but there is also some initialization step to run.  Example: I wrote a very simple .pfx viewer that invoked the existing UI on them, but it needed to be run once in order to add the shell verb to the registry.  This would be a portable installer as the exe was needed after install, but it was more cumbersome if you would have to run it once by hand since afterward you never needed to use the exe directly again.

JohnMcPMS (changes requested):

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -550,6 +571,130 @@
...
+        },
+        "Architecture": {
+          "type": "string",
+          "enum": [
+            "x86",
+            "x64",
+            "arm",
+            "arm64",
+            "neutral"
+          ],
```
Trenly, line 595: These should probably be made into definitions

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -550,6 +571,130 @@
         "InstallerUrl",
         "InstallerSha256"
       ]
+    },
+    "NestedInstaller": {
```
Trenly, line 575: Instead of duplicating everything from the installer, why not just reference the installer entry and extend it? It's possible to negate the InstallerUrl and the InstallerType
```json
    "NestedInstaller": {
      "type": "object",
      "properties": {
        "allOf": {
          "$ref": "#/definitions/Installer"
        },
        "InstallerUrl": {
          "not": {}
        },
        "InstallerType": {
          "not": {}
        },
        "NestedInstallerType": {
          "type": [
            "string",
            "null"
          ],
          "enum": [
            "msix",
            "msi",
            "appx",
            "exe",
            "inno",
            "nullsoft",
            "wix",
            "burn",
            "portable"
          ],
          "description": "Enumeration of supported nested installer types that could be contained in a zip or archive file."
        },
        "RelativeFilePath": {
          "type": [
            "string",
            "null"
          ],
          "description": "Relative path to the installer file or portable executable. Only applies to the nested installer within a Zip. "
        }
      },
      "required": [
        "Architecture",
        "InstallerSha256",
        "NestedInstallerType"
      ]
    }
```

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -91,6 +91,16 @@
         }
       }
     },
+    "WinGetInstallerUserRoot": {
```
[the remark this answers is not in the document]
Trenly: What about `PortableAppUserRoot` ?

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -462,6 +476,17 @@
         "InstallerType": {
           "$ref": "#/definitions/InstallerType"
         },
+        "NestedInstallers": {
+          "type": "array",
```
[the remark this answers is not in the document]
JohnMcPMS: We can do all of the multi-installer scenarios by having multiple installers defined at the top level and having all of them reference the same archive, but different files within.

schemas/JSON/settings/settings.schema.0.2.json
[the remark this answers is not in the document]
denelon: Maybe "PortablePackageUserRoot".

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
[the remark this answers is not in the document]
Trenly: > We can do all of the multi-installer scenarios by having multiple installers defined at the top level and having all of them reference the same archive, but different files within.

We can, yes, but then there is a lot of duplication of data. The Installer URL and Sha256 at minimum would be repeated for however many installer nodes there are, which could be many when considering the multiplicative effect of locale and architecture.

In my mind it would make updating manifests more difficult and would complicate the flow of wingetcreate and YamlCreate even further. Whereas, having the multi-installer scenarios handled by an array here would likely result in a cleaner user experience

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -550,6 +571,130 @@
...
+        "InstallerLocale": {
+          "$ref": "#/definitions/Locale"
+        },
+        "Platform": {
+          "$ref": "#/definitions/Platform"
+        },
+        "MinimumOSVersion": {
+          "$ref": "#/definitions/MinimumOSVersion"
+        },
+        "Architecture": {
```
yao-msft, line 587: >Architecture [](http://example.com/codeflow?start=9&length=12)

shall this one and below use the common definition?

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -462,6 +476,17 @@
         "InstallerType": {
           "$ref": "#/definitions/InstallerType"
         },
+        "NestedInstallers": {
```
yao-msft, line 479: >NestedInstallers [](http://example.com/codeflow?start=9&length=16)

nit: a name that's specific to archive would be better to differentiate from the "nested" installer type concept in AppAndFeatures entries ones

schemas/JSON/settings/settings.schema.0.2.json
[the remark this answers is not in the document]
ryfu-msft: Updated name from "WinGetInstallerUserRoot" to "PortablePackageUserRoot"  and "WinGetInstallerMachineRoot" to "PortablePackageMachineRoot".

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
ryfu-msft: Created definition for Architecture

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
ryfu-msft: Thank you for the suggestion. Updated with suggested changes. Looks like a lot cleaner now :)

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
ryfu-msft: Created a common definition for Architecture and InstallerUrl for consistency

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
ryfu-msft: I changed the name to "ArchiveInstallers" as to not confuse it with the "nested" installerType concept in AppsAndFeatures

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -449,37 +480,33 @@
...
-            "arm64",
-            "neutral"
-          ],
-          "description": "The installer target architecture"
+          "$ref": "#/definitions/Architecture"
         },
         "InstallerType": {
           "$ref": "#/definitions/InstallerType"
         },
+        "ArchiveInstallers": {
```
yao-msft, line 488: Since we do decide to use multiple ArchiveInstallers, do we have decision on the installer selection behaviors? (i.e. same app with different installers vs different apps)

For example, for some apps, we may have below and we want to install both of them
app1.exe
app2.exe

For some apps, we may have below and we want to only pick 1 to install
app1_x86.exe
app1_x64.exe

I think we can only support 1 of the scenarios above without further manifest entries and making the logic more complex.

ryfu-msft: After discussing this further, we're leaning towards supporting the following scenarios for zip:

1. Single or multiple portable exes contained inside a zip, which will all be installed.
2. Single Installer inside a zip.
3. Multiple installers inside a zip, but only installing one from the selection of installers. This would ensure that we support examples such as zips having multiple architecture installers or multiple locale installers residing in the same zip. If a specific architecture or locale is not specified, winget will choose the best option to install for you.

Regarding the edge case of a zip file containing multiple installerTypes (such as msi and exe), we would need to add support for enabling a user to select which installerType they would prefer to install.

JohnMcPMS: I do not support this:
> multiple portable exes contained inside a zip, which will all be installed.

That will just complicate things unnecessarily (although given how messy things are even when we try not to be, it might actually help us to need to handle it head on).  But generally, dependencies are a better solution than having a bunch of stuff in a single archive and installing it all.  Even if all of those things are in a single archive.

Option 3 is just like our current manifest, but directly supporting the installer set expansion from a single archive rather than requiring repetition of the archive as a top level installer.

ryfu-msft: Updated the comment above. We decided that we will only install a single portable inside a zip even if the zip archive contains multiple portables. This is to ensure that we limit the complexity that comes with installing multiple packages at the same time. It is better to have that be driven through dependencies when that feature is ready.

solomoncyj:
quick q does the pr cover zips with installers + install data and zips with the application ready to extracted to install location?

ryfu-msft:
> quick q does the pr cover zips with installers + install data and zips with the application ready to extracted to install location?

Yes, this PR just covers manifest changes related to zips with installers and zips with portables (where we would extract and copy to a specified install location)

solomoncyj:
can the portable zip use a .bat or .ps1 instead on .exe for main executable?

Trenly:
> can the portable zip use a .bat or .ps1 instead on .exe for main executable?

As far as I know, no. Those are a separate installer type for their own feature

denelon:
@solomoncyj,

No, we don't support scripts as installers.

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -91,6 +91,16 @@
...
     },
+    "PortableAppUserRoot": {
+      "description": "The default root directory where packages are installed to under User scope. Applies to the portable installer type.",
+      "type": "string",
+      "default":  "%LOCALAPPDATA%/WPM_Packages/User"
+    },
+    "PortableAppMachineRoot": {
+      "description": "The default root directory where packages are installed to under Machine scope. Applies to the portable installer type.",
+      "type": "string",
+      "default":  "%LOCALAPPDATA%/WPM_Packages/Machine"
```
denelon, line 102:
```suggestion
      "default":  "%LOCALAPPDATA%/WinGet_Packages/Machine"
```

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
solomoncyj: why not just select the main portable exe, chuck a shortcut to the desktop and the folder to program files?

superusercode:
Does this current schema allow for portable apps within zips that include the files necessary to run them? i.e. multimc:
image

gerardog:
For zip files containing a group of files belonging to a portable app, and assuming that to add the app to the PATH it must either be shimmed or sym-linked (whatever one the team chooses).

- Does the schema supports to specify which files need to be added to the path? (yes, more than one)
- Does it support to symlink/shim a bash script (file with no extension), or .ps1 file? (which are part of the application, not installers).
- A possible conflict would be if two portable apps may try to create the same shim/symlink. And in that case what happens? first/last one wins?, installation error?

Examples:
- gsudo ships with a `gsudo` file (no extension) which is a bash wrapper of `gsudo.exe`, intended to be called when the app is invoked from WSL (required for improved functionality). In a similar fashion, a PowerShell wrapper also exists. Both wrappers need to be in the Path.
- Edge case: gsudo generates an alias: `sudo => gsudo` by creating a `sudo.exe` symlink to `gsudo.exe`.

denelon:
Initially, we're looking at a single portable app inside a .zip file, or a single installer inside a .zip file. The cartesian of other possibilities is a bit much to tackle with the initial implementation. We will continue to iterate to capture more of the other scenarios in future releases.

ryfu-msft:
Closing this as Portable has been implemented and ZIP schema changes will be addressed and proposed in a separate PR

PR #1995, opened 2022-03-03
Rename source auto update group policy

lechacon:
From internal review of the group policies, we are removing "InMinutes" from the name of the policy `SourceAutoUpdateIntervalInMinutes`. This updates the policy definition with the new name, and changes to read the policy from the new registry value (while respecting the old value for compatibility). Updated the tests to use the new value name, and added tests for respecting the old name.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1995)


yao-msft:
>         case TogglePolicy::Policy::Settings: return

nit: this return is misaligned

---
Refers to: src/AppInstallerCommonCore/GroupPolicy.cpp:241 in 028cbe9. [](commit_id = 028cbe976725f88cf3bf07672ecbbc0a23fefb42, deletion_comment = False)

src/AppInstallerCLITests/GroupPolicy.cpp
```diff
@@ -60,6 +60,50 @@ TEST_CASE("GroupPolicy_UpdateInterval", "[groupPolicy]")
...
+        GroupPolicy groupPolicy{ policiesKey.get() };
+
+        auto policy = groupPolicy.GetValue<ValuePolicy::SourceAutoUpdateIntervalInMinutes>();
+        REQUIRE(policy.has_value());
+        REQUIRE(*policy == 1);
+    }
+
+    SECTION("Fallback to old name")
+    {
+        SECTION("When new name has invalid data")
```
yao-msft, line 89: >When new name has invalid data [](http://example.com/codeflow?start=17&length=30)

if new name has invalid data, should we stop falling back?

JohnMcPMS: Agreed, the fallback should only happen when the new name is not present, not if it is bad data.  I think that may require additional or different registry API usage than what we have now (as I suspect that this behavior was based on how the templated function works).

JohnMcPMS (changes requested):

src/AppInstallerCLITests/GroupPolicy.cpp
lechacon: Updated to not fallback on invalid data.

JohnMcPMS (approved):

PR #1998, opened 2022-03-04
Anonymize Com caller value

yao-msft:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1998)


lechacon (approved):

JohnMcPMS:
Why are we doing this?  Abundance of caution?

yao-msft:
> Why are we doing this? Abundance of caution?

Yes, in case the caller exe locates under user path.

Issue #1977, opened 2022-02-26
Winget --version to show rendered link to log


Karl-WE:
### Description of the new feature / enhancement

When Winget detects it is run in Windows Terminal it should render the path to the logs, so it can be opened simply by ctrl+left mouse click in Windows Terminal

The link should be rendered by Windows Terminal, so a user can use ctrl+left mouse click to open it in File Explorer (this might need remote work of the Windows Terminal team).
Currently only https seems to handled, no local paths.
fyi @DHowett 

just in case  rendering the link and determination is too complicated or not worth the time, I would like to fall back to the basic idea that "winget --info" when run in a PowerShell session should show the logs folder as $env:localappdata, so a user can directly copy the path and change to this folder in the shell.

### Proposed technical implementation details

Winget shall detect if it is running in Windows Terminal (if possible)
Winget shall detect if it is running in PowerShell or cmd session

Winget -info shall output the link to logs as a rendered link (Windows Terminal would do this in the correct format)

Examples:
Winget run in native shell as cmd or PowerShell 5.x / 7.x
current behaviour:
Logs: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

Winget run in Windows Terminal (e. g. Windows 11 default shell) as cmd 
Logs: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

Winget run in Windows Terminal (e. g. Windows 11 default shell) PowerShell 5.x / 7.x
$env:localappdata\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

Karl-WE:
While technically possible It's very unlikely it is redirected to a different location indeed or on the other hand the value of the var could be parsed and displayed. Very good point @Trenly

Trenly:
@Karl-WE - Does the above linked PR for the user setting close this issue, or should it be kept open for rendering the correct styling for environment variables?

Trenly:
[Policy] Needs Author Feedback
[Policy] Area-output

Karl-WE:
Reading the PR once again I believe this is still relevant. @denelon @Trenly .

Pardon I haven't noticed this tagging earlier.

Just checked the output of --version and --info

In winget in cmd and Microsoft Terminal / PowerShell.

[...]

PR #2012, opened 2022-03-09
Portable apps specification

denelon:

-----

Related to:

*  #182

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2012)

### DIFF
--- a/doc/specs/#182 - Support for installation of portable standalone apps.md
+++ b/doc/specs/#182 - Support for installation of portable standalone apps.md
@@ -109,7 +109,7 @@ Starting package install...
 Successfully installed
 ```
 
-If the portable application is running when the user performs an upgrade, the user will be informed and given options.
+If the portable application is running when the user performs an upgrade, the user will be informed.
 
 ```text
 winget upgrade Microsoft.NuGet
--- a/doc/specs/#182 - Support for installation of portable standalone apps.md
+++ b/doc/specs/#182 - Support for installation of portable standalone apps.md
@@ -1,0 +1,188 @@
+---
+author: Demitrius Nelon @denelonms
+created on: 2022-03-09
+last updated: 2022-03-09
+issue id: 182
+---
+
+# Portable/Standalone applications
+
+"For [#182](https://github.com/microsoft/winget-cli/issues/182)"
+
+## Abstract
+
+Several packages do not have an installer. They are simply a binary executable. This specification describes how these types of programs will be treated as "packages" by the Windows Package Manager.
+
+## Inspiration
+
+Just because an installer doesn't exist for a particular package, that doesn't mean you shouldn't be able to install it with the Windows Package Manager.
+
+## Solution Design
+
+The installer type “portable” will be added to the enumerated list of installer types to support these types of packages.
+
+When one of these packages is encountered there are several aspects to “installing” these packages and supporting the “upgrade” behavior. Notable concerns include the location of the package, creating or updating an entry in Windows Apps & Features.
+
+### Command Line Arguments
+
+Users should be able to specify the location for where the program is “installed” on their machine with the “--location” argument.
+
+Some packages like GitLab Runner have a file name including extra metadata (gitlab-runner-windows-386.exe). The installation instructions suggest renaming the file after it has been downloaded. In that vein, portable / standalone executables should have their “command” value specified so the Windows Package Manager can perform a default rename behavior. In addition, a “--rename” argument should be added so the user can choose a value they prefer. To have the upgrade scenario honor the custom name, this information should be recorded in the installed packages data store.
+
+Some portable applications generate or consume other files. An additional argument "--purge" will be added for the uninstall scenario to remove all files and subsequently the directory if the user wishes. The corrolary "--preserve" will be used to preserve files if the default setting has been modified.
+
+### Settings
+
+Users should have settings to be able to specify a new default location other than the system default for the Windows Package Manager.
+
+The default path for installing these packages is "%LOCALAPPDATA%/WinGet_Packages/User" for user based installs. The default path for installing these packages is "%LOCALAPPDATA%/WinGet_Packages/Machine"for machine wide installs.
+
+The corresponding settings are "PortablePackageUserRoot" and "PortablePackageUserRoot".
+
+>Note: The "packageIdetifier" will be used to generate subdirectories for these binaries to be installed.
+
+Using Gitlab.gitlab runner as an example would result in the .exe being placed in "%LOCALAPPDATA%/WinGet_Packages/User/Gitlab.gitlab-runner/"
+If a user has configured a path to be used for portable applications that path should be honored. The user should also be able to specify if they want all portable packages placed in the same directory, or in a directory per package.
+There will be settings for specifying a default behavior related to whether the path environment variable should or should not be modified.
+
+A related setting for the uninstall scenario will be able to specify the default behavior for either "--clean" or "--purge".
+
+### Install
+
+When the portable application is being installed by the Windows Package manager, an entry will be created in Windows Apps & Features so the user will be able to see the application is installed.
+
+The "AppPath" subkey will identify the path to the application so the user will be able to execute the program from any path via command prompt.
+
+The data from “AppsAndFeaturesEntry” in the manifest will be used to specify the name of the package in Windows Apps & Features.  Constraints for manifests with portable packages are covered below in the section on validating manifests. If the “AppsAndFeaturesEntry” doesn’t have a “DisplayName” value, then we will use the “PackageName”.
+The Product Code will be populated with the "PackageIdentifier" if it is not specified in the “AppsAndFeaturesEntry”.
+
+No shortcuts or icons will be created for the package.
+
+>Note: Some portable programs may require dependencies. Dependencies are not covered in this specification and are not yet supported as of the writing of this specification.
+
+### Upgrade
+
+The package is upgraded in the same path as the installed version. The first step the Windows Package Manager will perform is an uninstall of the previous version of the package. If the file is in use, the user will be informed the package is running so they can shut it down. Optionally, the user may specify "--force" to forcefully shut the application down for upgrade.
+
+A new Windows Apps & Features entry is created to correctly report the upgraded version for future potential upgrades.
+
+There will be no support for installing multiple “side by side” versions of portable packages.
+
+### Uninstall
+
+The executable should be removed along with the entry in Apps & Features. The user should also be able to uninstall the package from Apps & Features.
+
+>Note: the default behavior for uninstalling a portable application with Windows Apps & Features will be to execute uninstall with "--purge" so any files created by the portable application will be removed if they are located in the portable applications directory.
+
+If the directory is empty after removing the portable application, the directory should also deleted. If the directory is not empty, the user should be informed that other files exist in the directory so it will not be removed.
+
+### Manifest Validation
+
+Portable package manifests will only support zero or one "command" value to be specified. In the absence of the "--rename" argument, the value specified in "command" will define the default value for renaming the portable application and subsequently the value specified in the "AppPath" subkey.
+
+## UI/UX Design
+
+### Installing a portable package
+
+```text
+winget install Microsoft.NuGet
+Found NuGet [Microsoft.NuGet] Version 6.0
+This application is licensed to you by its owner.
+Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
+Downloading https://dist.nuget.org/win-x86-commandline/v6.0.0/nuget.exe
+  ██████████████████████████████   0.1 MB /  0.1 MB
+Successfully verified installer hash
+Starting package install...
+Successfully installed
+```
+
+### Upgrading a portable package
+
+```text
+winget upgrade Microsoft.NuGet
+Found NuGet [Microsoft.NuGet] Version 7.0
+This application is licensed to you by its owner.
+Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
+Downloading https://dist.nuget.org/win-x86-commandline/v7.0.0/nuget.exe
+  ██████████████████████████████   0.1 MB /  0.1 MB
+Successfully verified installer hash
+Starting package install...
+Successfully installed
+```
+
+If the portable application is running when the user performs an upgrade, the user will be informed and given options.
+
+```text
+winget upgrade Microsoft.NuGet
+Found NuGet [Microsoft.NuGet] Version 7.0
+This application is licensed to you by its owner.
+Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
+Downloading https://dist.nuget.org/win-x86-commandline/v7.0.0/nuget.exe
+  ██████████████████████████████   0.1 MB /  0.1 MB
+Successfully verified installer hash
+Starting package install...
+Failed to install NuGet [Microsoft.Nuget].
+Package in use. Either exit the program or use "--force" to upgrade.
+```
+
+### Uninstalling a portable package
+
+```text
+winget uninstall NuGet 
+Found NuGet [Microsoft.NuGet]
+Starting package uninstall...
+Successfully uninstalled
+```
+
+If the portable application created files in the portable application's directory the user will be informed.
+
+```text
+winget uninstall NuGet 
+Found NuGet [Microsoft.NuGet]
+Starting package uninstall...
+The "--purge" argument was not specified. 
+Files still exist in "%LOCALAPPDATA%/WinGet_Packages/User/Microsoft.NuGet/". 
+Successfully uninstalled
+```
+
+## Capabilities
+
+This will allow users to manage portable applications via the Windows Package Manager.
+
+### Accessibility
+
+Any description or output text that is added by consequence of this feature will need to be localized. Moreover, this feature allows a mechanism to add accessibility settings in the future.
+
+### Security
+
+This feature will not add any new security risks. All risks associated with this feature have been resolved with the core install, import, upgrade, and uninstall features.
+
+>Note: We reccommend users leverage the Windows Sandbox when testing manifests on their local machines as the programs may not have been validated from a security perspective prior to being submitted to the Windows Package Manager Community App Repository.
+
+### Reliability
+
+This should not introduce any new reliability concerns.
+
+### Compatibility
+
+Previous versions of the Windows Package Manager will not cease to function, but they will not be able to apply arguments or settings related to this feature. They will not be able to install portable applications.
+
+An update to the Windows Package Manager is required to use this feature with the related arguments and settings.
+
+### Performance, Power, and Efficiency
+
+This should not introduce any new performande, power, or efficiency concerns in the Windows Package Manager.
+
+## Potential Issues
+
+Users may not be aware that there is no support for multiple versions of the same package on their system, so performing an upgrade may not provide they experience they were expecting.
+
+Users may not be aware that the uninstall behavior by default may leave files on their systems if they were created by a portable application.
+
+If a user performs uninstall via Windows Apps & Features the default behavior will be to run `winget uninstall package --purge`.
+
+## Future considerations
+
+## Resources
+
+[Portable / Standalone Executables · Discussion #44183 · microsoft/winget-pkgs (github.com)](https://github.com/microsoft/winget-pkgs/discussions/44183)
--- a/doc/specs/#182 - Support for installation of portable standalone apps.md
+++ b/doc/specs/#182 - Support for installation of portable standalone apps.md
@@ -15,13 +15,13 @@ Several packages do not have an installer. They are simply a binary executable. 
 
 ## Inspiration
 
-Just because an installer doesn't exist for a particular package, that doesn't mean you shouldn't be able to install it with the Windows Package Manager.
+Just because an installer doesn't exist for a particular package doesn't mean you shouldn't be able to install it with the Windows Package Manager.
 
 ## Solution Design
 
 The installer type “portable” will be added to the enumerated list of installer types to support these types of packages.
 
-When one of these packages is encountered there are several aspects to “installing” these packages and supporting the “upgrade” behavior. Notable concerns include the location of the package, creating or updating an entry in Windows Apps & Features.
+When one of these packages is encountered, there are several aspects to “installing” these packages and supporting the “upgrade” behavior. Notable concerns include the location of the package, creating or updating an entry in Windows Apps & Features.
 
 ### Command Line Arguments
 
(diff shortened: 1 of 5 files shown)


doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+Some portable applications generate or consume other files. An additional argument "--purge" will be added for the uninstall scenario to remove all files and subsequently the directory if the user wishes. The corollary argument "--preserve" will be used to preserve files if the default setting has been modified.
+
+### Settings
+
+Users should have settings to be able to specify a new default location other than the system default for the Windows Package Manager.
+
+The default path for installing these packages is "%LOCALAPPDATA%/WinGet_Packages/User" for user based installs. The default path for installing these packages is "%LOCALAPPDATA%/WinGet_Packages/Machine"for machine wide installs.
```
Trenly, line 38:
```suggestion
The default path for installing these packages is "%LOCALAPPDATA%/WinGet_Packages/User" for user based installs. The default path for installing these packages is "%PROGRAMDATA%/WinGet_Packages/Machine" for machine wide installs.
```
The localappdata for a specific user may not be accessible by all users on a machine, so I would recommend these be put into a central location. Perhaps they could all be put into programdata and the registry entries determine whether it is installed for all users or a single user?

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+>Note: The "packageIdentifier" will be used to generate subdirectories for these binaries to be installed to.
+
+Using Gitlab.gitlab runner as an example would result in the .exe being placed in "%LOCALAPPDATA%/WinGet_Packages/User/Gitlab.gitlab-runner/"
+If a user has configured a path to be used for portable applications, that path should be honored. The user should also be able to specify if they want all portable packages placed in the same directory, or in a directory per package.
+
+A related setting for the uninstall scenario will be able to specify the default behavior for either "--clean" or "--purge".
+
+### Install
+
+When the portable application is being installed by the Windows Package Manager, an entry will be created in Windows Apps & Features so the user will be able to see that the application is installed.
```
Trenly, line 51: We may need to specify that the entry here should be created in HKCU for user scope and HKLM for machine scope.

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+The "AppPath" registry subkey will identify the path to the application so the user will be able to execute the program from any path via command prompt.
+
+The data from “AppsAndFeaturesEntry” in the manifest will be used to specify the name of the package in Windows Apps & Features.  Constraints for manifests with portable packages are covered below in the section on validating manifests. If the “AppsAndFeaturesEntry” doesn’t have a “DisplayName” value, then we will use the “PackageName”.
+The Product Code will be populated with the "PackageIdentifier" if it is not specified in the “AppsAndFeaturesEntry”.
```
Trenly, line 56: I think that it would be useful to have the Product Code also be tied to a specific installer/version like it is with MSI installers. Perhaps it could be to the SHA256, or the "PackageVersion"?

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+No shortcuts or icons will be created for the package.
```
Trenly, line 58: Could this be an option in user settings? Maybe under future considerations

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+>Note: Some portable programs may require dependencies. Dependencies are not covered in this specification and are not yet supported as of the writing of this specification.
+
+### Upgrade
+
+The package is upgraded in the same path as the installed version. The first step the Windows Package Manager will perform is an uninstall of the previous version of the package. If the file is in use, the user will be informed the package is running so they can shut it down. Optionally, the user may specify "--force" to forcefully shut the application down for upgrade.
+
+A new Windows Apps & Features entry is created to correctly report the upgraded version for future potential upgrades.
+
+There will be no support for installing multiple “side by side” versions of portable packages.
```
Trenly, line 68: How does this interact with channels? I know it may be outside the scope of the spec, but I think it's important to consider. A user may want both a stable and beta version of a portable app at the same time

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Uninstall
+
+The executable should be removed along with the entry in Apps & Features. The user should also be able to uninstall the package from Apps & Features.
+
+>Note: the default behavior for uninstalling a portable application with Windows Apps & Features will be to execute uninstall with "--purge" so any files created by the portable application will be removed if they are located in the portable applications directory.
```
Trenly, line 74: Edge case - if `winget uninstall {PackageIdentifier} --purge` is set as the uninstall string, and the user no longer has winget installed, what happens?

Would it be beneficial to write the uninstall string as a string of commands to be passed directly to cmd.exe? 

(Probably ignore this comment, it is likely a very rare edge case)

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+If the directory is empty after removing the portable application, the directory should also be deleted. If the directory is not empty, the user should be informed that other files exist in the directory so it will not be removed.
+
+### Manifest Validation
+
+Portable package manifests will only support zero or one "command" value to be specified. In the absence of the "--rename" argument, the value specified in "command" will define the default value for renaming the portable application and subsequently the value specified in the "AppPath" subkey.
```
Trenly, line 80: Several applications that support portable installs and static installs support multiple commands. I think that they should support more than one command and that the first command in the array should define the default value.

What about apps with monikers? Using vscode as an example - if it were portable type, then the command is `code`, so the rename would be `code.exe` and not `vscode.exe`? Seems a bit confusing.

Trenly, line 80: As mentioned above, we may also want to consider how channels may impact the requirements of renaming files

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+Previous versions of the Windows Package Manager will not cease to function, but they will not be able to apply arguments or settings related to this feature. They will not be able to install portable applications.
+
+An update to the Windows Package Manager is required to use this feature with the related arguments and settings.
+
+### Performance, Power, and Efficiency
+
+This should not introduce any new performance, power, or efficiency concerns in the Windows Package Manager.
+
+## Potential Issues
```
Trenly, line 175: I would suggest that this issue be mentioned, as any path output in the case of remaining files may run into similar feedback -
* https://github.com/microsoft/winget-cli/issues/1977

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Uninstall
+
+The executable should be removed along with the entry in Apps & Features. The user should also be able to uninstall the package from Apps & Features.
+
+>Note: the default behavior for uninstalling a portable application with Windows Apps & Features will be to execute uninstall with "--purge" so any files created by the portable application will be removed if they are located in the portable applications directory.
+
+If the directory is empty after removing the portable application, the directory should also be deleted. If the directory is not empty, the user should be informed that other files exist in the directory so it will not be removed.
+
+### Manifest Validation
```
Trenly, line 78: Should "scope" be a valid manifest field when installer type is portable? I would imagine that portable applications should default to user scope and only use machine scope if specified using --scope

Trenly, line 78: Should there be only one `AppsAndFeaturesEntries` for portable packages? I can't imagine a scenario where there should be 2 of them

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,190 @@
...
+The data from “AppsAndFeaturesEntry” in the manifest will be used to specify the name of the package in Windows Apps & Features.  Constraints for manifests with portable packages are covered below in the section on validating manifests. If the “AppsAndFeaturesEntry” doesn’t have a “DisplayName” value, then we will use the “PackageName”.
+The Product Code will be populated with the "PackageIdentifier" if it is not specified in the “AppsAndFeaturesEntry”.
+
+No shortcuts or icons will be created for the package.
+
+>Note: Some portable programs may require dependencies. Dependencies are not covered in this specification and are not yet supported as of the writing of this specification.
+
+### Upgrade
+
+The package is upgraded in the same path as the installed version. The first step the Windows Package Manager will perform is an uninstall of the previous version of the package. If the file is in use, the user will be informed the package is running so they can shut it down. Optionally, the user may specify "--force" to forcefully shut the application down for upgrade.
```
Trenly, line 64: > The first step the Windows Package Manager will perform is an uninstall of the previous version of the package

Manifest Validation may need to warn that `UpgradeBehavior`  should be null or "uninstallPrevious" in the case of a portable installer. It may be confusing otherwise, even though the behavior will be the same.

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: @ryfu-msft thoughts?

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: That's a good point.

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: 👍

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: Future Considerstions.

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: Yeah, I'd say that's an edge case. Updated "Potential Issues".

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: added

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: @ryfu-msft

denelon: I think we discussed this before with various scenarios. I don't remember if we landed on this as a semantic requirement or not. 

@ryfu-msft thoughts?

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: @ryfu-msft should we add this as another semantic validation?

doc/specs/#182 - Support for installation of portable standalone apps.md
JohnMcPMS: I think that Program Files is the correct location for machine wide installs, but that is more my intuition than having looked at the current guidance.  I agree that the specific user's local app data is not the right location.

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,196 @@
...
+
+Some portable applications generate or consume other files. An additional argument "--purge" will be added for the uninstall scenario to remove all files and subsequently the directory if the user wishes. The corollary argument "--preserve" will be used to preserve files if the default setting has been modified.
+
+### Settings
+
+Users should have settings to be able to specify a new default location other than the system default for the Windows Package Manager.
+
+The default path for installing these packages is "%LOCALAPPDATA%/WinGet_Packages/User" for user based installs. The default path for installing these packages is "%LOCALAPPDATA%/WinGet_Packages/Machine"for machine wide installs.
+
+The corresponding settings are "PortablePackageUserRoot" and "PortablePackageUserRoot".
```
JohnMcPMS, line 40:
```suggestion
The corresponding settings are "PortablePackageUserRoot" and "PortablePackageMachineRoot".
```

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,196 @@
...
+
+>Note: The "packageIdentifier" will be used to generate subdirectories for these binaries to be installed to.
+
+Using Gitlab.gitlab runner as an example would result in the .exe being placed in "%LOCALAPPDATA%/WinGet_Packages/User/Gitlab.gitlab-runner/"
```
JohnMcPMS, line 44: Do we want to specify how a conflict between identifiers from different sources would be handled?  My first thought is that by default we wouldn't include anything about the source, but if a conflict was detected, we would prepend the source name to the second (and third, etc.) directories.

doc/specs/#182 - Support for installation of portable standalone apps.md
JohnMcPMS: And even exactly which architecture we should use when writing it (should an x86 package on an x64 device go into the `SysWow...` path).

doc/specs/#182 - Support for installation of portable standalone apps.md
JohnMcPMS: Product Codes that change over time are actually quite annoying.  I would prefer that we don't do that.

doc/specs/#182 - Support for installation of portable standalone apps.md
JohnMcPMS: I think we should not define that we will "uninstall" first; that should be left to the behavior defined in the manifest.  We should attempt to determine if files are in use though before proceeding (not that it can be perfect) and then replace any changed files.

doc/specs/#182 - Support for installation of portable standalone apps.md
JohnMcPMS: Channels are meant for use by packages that cannot be installed SxS between stable/beta.  For packages that can be, the thought was to use different package ids.

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,196 @@
...
+
+A new Windows Apps & Features entry is created to correctly report the upgraded version for future potential upgrades.
+
+There will be no support for installing multiple “side by side” versions of portable packages.
+
+### Uninstall
+
+The executable should be removed along with the entry in Apps & Features. The user should also be able to uninstall the package from Apps & Features.
+
+>Note: the default behavior for uninstalling a portable application with Windows Apps & Features will be to execute uninstall with "--purge" so any files created by the portable application will be removed if they are located in the portable applications directory.
```
JohnMcPMS, line 74: I disagree on the default here; we should not purge.  I'm not unmovable, but generally it seems like if we already inform them of the extra files, they can act.

It also makes installing to a non-empty directory very dangerous for the future.  I think generally maybe purge shouldn't work if a custom, existing or non-empty location was chosen to prevent accidental nukage.

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,196 @@
...
+
+If the directory is empty after removing the portable application, the directory should also be deleted. If the directory is not empty, the user should be informed that other files exist in the directory so it will not be removed.
```
JohnMcPMS, line 76: We should probably track whether we created the directory and decide based on that too.

doc/specs/#182 - Support for installation of portable standalone apps.md
JohnMcPMS: Yes, scope will come from the user in these cases, so it will probably be ignored in the manifest.

doc/specs/#182 - Support for installation of portable standalone apps.md
JohnMcPMS: We could allow the command(s) to define the App Path entrie(s), while not renaming the file.

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,196 @@
...
+This application is licensed to you by its owner.
+Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
+Downloading https://dist.nuget.org/win-x86-commandline/v7.0.0/nuget.exe
+  ██████████████████████████████   0.1 MB /  0.1 MB
+Successfully verified installer hash
+Starting package install...
+Successfully installed
+```
+
+If the portable application is running when the user performs an upgrade, the user will be informed and given options.
```
JohnMcPMS, line 112: "options" being actions that they could take, not "interactive command line questionnaire".

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: I think it's an edge case, but what if a user changes/removes that source? We thought about including the source in one iteration internally but decided against it.

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: The concern is if we are calling the uninstall from Apps & Features, the user wouldn't get informed. If we had a way to inform the user, then I'd agree. I just don't know what's possible here.

doc/specs/#182 - Support for installation of portable standalone apps.md
denelon: Agree, I'm trying to ensure we avoid any user input. My intent with "given options" was just the additional text in the output with the hint for "--force".

denelon, line 112:
```suggestion
If the portable application is running when the user performs an upgrade, the user will be informed.
```

doc/specs/#182 - Support for installation of portable standalone apps.md
Trenly: If we go with Program Files, then Architecture should also be respected to place them in `Program Files` of `Program Files (x86)` respectively

doc/specs/#182 - Support for installation of portable standalone apps.md
Trenly: > Product Codes that change over time are actually quite annoying. I would prefer that we don't do that.

As long as there aren't any potential issues with matching a program to an ARP entry, or disambiguation, then I would think that having static product codes would be okay. I would assume that the DisplayName and DisplayVersion would always be set, so it may not be an issue unless it conflicts with an already installed application making the package ambiguous

doc/specs/#182 - Support for installation of portable standalone apps.md
ryfu-msft: That makes sense to me. We can add an additional manifest validation rule that only applies to the "portable" installerType that verifies only one entry for ```AppsAndFeaturesEntries``` can be specified.

doc/specs/#182 - Support for installation of portable standalone apps.md
jedieaston: If a portable app expects to be able to write to the directory it is installed in (as many do, since they expect that they are installed per user), putting them in Program Files would probably break stuff since non adminstrators can't write there. 

This may open up a security can of worms, but what if when you installed a portable app for the system winget just put it in everyone's %appdata%? You'd be making several copies of it, but it would improve compatibility. (I'm thinking how the Teams Machine-Wide Installer works.) Alternatively, maybe symlinking the individual files the portable app comes with from Program Files to somewhere the user can write to (so they get the application itself from program files, but any new preferences or other data would be written to a folder they had access to).

JohnMcPMS: I feel that the expectation with a system installation is that all users (even those that have not logged in yet) will have access.  That makes it hard to do anything that involves updating every user's files (as well as being questionable in terms of respecting other users' ownership).  I think that it is also a security problem to allow any user to write to the shared location.

I don't think that we are trying to make every package work in every scenario; and this seems like an area where if the portable needs to write to its current location, it might just not work for system install and non-admin use cases.

denelon:
I believe we've aligned the specification to address the comments that have been made. We're going to have the engineering team do one more review before merging the specification in.

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+### Upgrade
+
+The package is upgraded in the same path as the installed version. The first step the Windows Package Manager will perform is to download the executable to a temporary location, and attempt to copy the exe to the specified install location. If an exe with the same name already exists, the Windows Package Manager will attempt to overwrite the file. If that process fails because the file is currently in use, the user will be informed the package is running so they can shut it down. Optionally, the user may specify "--force" to forcefully shut the application down for upgrade. Once the exe has successfully been copied to the specified install location, the "App Paths" registry and the entry in "Apps & Features" will be updated accordingly.
+
+If the "UninstallPrevious" field is specified in the manifest, then the Windows Package Manager will perform an uninstall of the previous version of the package prior to installing the newer version. 
+
+A new Windows Apps & Features entry is created to correctly report the upgraded version for future potential upgrades.
```
Trenly, line 79:
```suggestion
A new Windows Apps & Features entry is created to correctly report the upgraded version for future potential upgrades.
```
Is this a new entry, or is the existing entry from the existing package updated?

denelon: If we uninstall, then the entry is removed. If we can simply upgrade, the existing entry should get updated.

doc/specs/#182 - Support for installation of portable standalone apps.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Installation from multiple sources:
+
+If the user chooses to install the same package but from a secondary source, the Windows Package Manager will create an additional subdirectory to indicate the source that the package came from. Since entries in the registry must be unique, if the entry in the "App Paths" registry conflicts with an existing entry, then the user will be notified of the conflict and the subkey value will be updated to reflect the change in executable path. In other words, the latest installed package will take precedence. This will require utilizing the installed packages data store to determine whether a package is being installed from two different sources.
```
JohnMcPMS, line 70: I don't think that an additional subdirectory should be created; it makes it harder to visually scan the packages directory to find things.  Something more like `PackageIdentifier_Source2` would make it easier to find both the first and second package directories together.

JohnMcPMS: Additionally, I'm not sure that we should do "last writer wins (LWW)" semantics.  I think that is appropriate if we did nothing else, but I would also like to have a command like `winget app-path apply PackageFoo`.  If we had that command, I would think it better to have "first writer wins (FWW)" and allow the user to use the new command to apply the appropriate app-path.

We could start with FWW and an additional parameter to specify LWW for now, then add the app path update command in the future.

ryfu-msft: I updated the description to reflect your suggestion. I also agree that not having a separate directory just for source would make it easier to find both packages in the same directory. 

I included further clarification for having FWW be our default behavior as well as having an argument to support allowing users to have a LWW behavior if they want that using the --lww argument

ryfu-msft (approved):

PR #2024, opened 2022-03-16
PowerShell module - Support localization

BlackCatDeployment:
- [X] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [X] Are you working against an Issue?

-----

Use https://raw.githubusercontent.com/microsoft/winget-cli/master/Localization/Resources/$Language/winget.resw for getting localization data for current language in order to correctly use the PowerShell module for any languages supported

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2024)

### DIFF
--- a/tools/PowerShell/Microsoft.WinGet.Client/src/Library/Get-LanguageData.ps1
+++ b/tools/PowerShell/Microsoft.WinGet.Client/src/Library/Get-LanguageData.ps1
@@ -1,6 +1,6 @@
 Function Get-LanguageData {
     PARAM(
-        [Parameter(Position=0)] $Language = (Get-UICulture).Name
+        [Parameter(Position=0)] $Language = (Get-UICulture).Name.Substring(0,2)
     )
 
     $languageData = $(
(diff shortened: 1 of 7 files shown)


tools/PowerShell/Microsoft.WinGet.Client/src/Library/Get-LanguageData.ps1
```diff
@@ -0,0 +1,34 @@
...
+    PARAM(
+        [Parameter(Position=0)] $Language = (Get-UICulture).Name
+    )
+
+    $languageData = $(
+        $hash = @{}
+
+        $(try {
+            # We have to trim the leading BOM for .NET's XML parser to correctly read Microsoft's own files - go figure
+            ([xml](((Invoke-WebRequest -Uri "https://raw.githubusercontent.com/microsoft/winget-cli/master/Localization/Resources/$Language/winget.resw" -ErrorAction Stop ).Content -replace "\uFEFF", ""))).root.data
```
JohnMcPMS, line 11: Having the build extract just the entries that are needed and shipping those subset .resw files with the module would be better.  Even having to download every language for the ~10 strings one needs would be smaller than downloading all of the strings for 1 language.

lechacon, line 11: Will this work if somebody has the same language but different culture? Say, if I use es-MX for Windows, I would probably want it to use the es-ES resources as there is no es-MX for winget

BlackCatDeployment: Good idea!

BlackCatDeployment: I updated the module with the idea exposed. I let you check if it's clean or if there is a better way :)

JohnMcPMS:
Closing as we have created a custom PS module that doesn't rely on parsing output (and therefore doesn't need to know about the localization).

Issue #1994, opened 2022-03-02
Relax Regex Validation on Protocols


Trenly:
### Description of the new feature / enhancement

The current schema does not allow for protocols to start with a number

https://github.com/microsoft/winget-cli/blob/739aefe60aedd437dd53cccf7de45bec29532a8d/schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json#L208-L218

This causes issues when Publishers implement their own protocols or handlers such as `360chromexurl`

### Proposed technical implementation details

The regex should be relaxed to allow for custom protocols and handlers which start with numbers or removed entirely

Trenly:
Reolved in https://github.com/microsoft/winget-cli/pull/2028

Issue #1984, opened 2022-02-28
Manifest Field for User Manual / Documentation


Trenly:
### Description of the new feature / enhancement

As a user, there are times where I would like to quickly view the documentation related to a specific application. This is sometimes different than the support URL, as some publishers make their software manuals or guides available online.

### Proposed technical implementation details

Add an optional field to the Locale manifest - `DocumentationUrls`

```yaml
DocumenationUrls:
- DocumentDescription: Installation Guide
  DocumentUrl: https://github.com/microsoft/winget-cli#installing-the-client
- DocumentDescription: Settings Guide
  DocumentUrl: https://github.com/microsoft/winget-cli/blob/master/doc/Settings.md
- DocumentDescription: Troubleshooting
  DocumentUrl: https://github.com/microsoft/winget-cli/blob/master/doc/troubleshooting/README.md
```

Issue #1916, opened 2022-02-09
Increase FileExtensions limit to 512


Trenly:
### Description of the new feature / enhancement

Some applications support an inordinately large number of file extensions. The file extensions limit should be increased to allow the metadata for these packages to be accurately represented

Example with 388 extensions - 

https://nikkhokkho.sourceforge.io/static.php?page=FileOptimizer
https://github.com/microsoft/winget-pkgs/pull/45620

Issue #1889, opened 2022-02-01
Add an optional URL to Expected Return Codes


denelon:
### Description of the new feature / enhancement

In some cases, a URL would help users when they receive an Expected Return Code. This might be a page that has specific instructions on resolving issues related to the expected return code.

"ReturnResponseUrl" should be added to the schema so a user can be given additional guidance for expected return codes.

PR #2028, opened 2022-03-17
Support for 1.2 Schema Additions

ryfu-msft:
This PR updates the 1.2 schema to incorporate the proposed changes from #1891. This only adds functionality for parsing and processing the new schema additions and does not include their respective feature support. 

Also includes additional proposed arguments and settings schema changes for installing portables. Excludes some of the installer manifest changes for the portable feature as those will be included with the work for the portable install feature which will have more thorough testing.

Related to:

* #1994
* #1984
* #1916
* #1889
* #1888 
* #1345 
* #1277
* #607

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2028)


schemas/JSON/manifests/v1.2.0/manifest.defaultLocale.1.2.0.json
```diff
@@ -154,6 +169,23 @@
...
+      "minLength": 1,
+      "maxLength": 240,
+      "description": "The notes displayed to the user upon completion of a package installation."      
+    },
+    "Documentations": {
+      "type": [ "array", "null" ],
+      "items": {
+        "$ref": "#/definitions/Documentation"
+      },
+      "maxItems": 128
```
Trenly, line 187:
```suggestion
      "maxItems": 256
```
I know that this may seem a bit overkill, but given that we have seen issues created to increase the number of installer nodes and file extensions, I think making this larger off the start may be a wise choice

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -385,6 +411,19 @@
...
     },
+    "UnsupportedArguments": {
+      "type": [ "array", "null" ],
+      "uniqueItems": true,
+      "items": {
+        "type": "string",
+        "title": "UnsupportedArgument",
+        "enum": [
+          "log",
+          "location"
```
Trenly, line 422: Regarding #910 - Would it be appropriate to include `silent` in this enum?

schemas/JSON/manifests/v1.2.0/manifest.locale.1.2.0.json
```diff
@@ -149,6 +164,23 @@
...
+      "minLength": 1,
+      "maxLength": 240,
+      "description": "The notes displayed to the user upon completion of a package installation."      
+    },
+    "Documentations": {
+      "type": [ "array", "null" ],
+      "items": {
+        "$ref": "#/definitions/Documentation"
+      },
+      "maxItems": 128
```
Trenly, line 182:
```suggestion
      "maxItems": 256
```
See comments above

schemas/JSON/manifests/v1.2.0/manifest.singleton.1.2.0.json
```diff
@@ -691,6 +729,23 @@
...
+      "minLength": 1,
+      "maxLength": 240,
+      "description": "The notes displayed to the user upon completion of a package installation."
+    },
+    "Documentations": {
+      "type": [ "array", "null" ],
+      "items": {
+        "$ref": "#/definitions/Documentation"
+      },
+      "maxItems": 128
```
Trenly, line 747:
```suggestion
      "maxItems": 256
```
See comments above

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -48,6 +48,9 @@ namespace AppInstaller::Manifest
         const char* const MissingManifestDependenciesNode = "Dependency not found: ";
         const char* const NoSuitableMinVersion = "No Suitable Minimum Version: ";
         const char* const FoundLoop = "Loop found.";
+        const char* const ExceededAppsAndFeaturesEntryLimit = "Only zero or one entry for Apps and Features can be specified for InstallerType portable.";
+        const char* const ExceededCommandsLimit = "Only zero or one value for Commands can be specified for InstallerType portable.";
```
Trenly, line 52:
```suggestion
        const char* const ExceededAppsAndFeaturesEntryLimit = "Only zero or one entry for Apps and Features may be specified for InstallerType portable.";
        const char* const ExceededCommandsLimit = "Only zero or one value for Commands may be specified for InstallerType portable.";
```
nit: grammar for clarity

Trenly (commented):
Sorry for not leaving some of these comments regarding number of nodes and the enum in the original PR (#1996)

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
jedieaston: Some arguments in the Silent field are required in the schema, so I don't know if that makes sense. Manifests that don't support `--silent` shouldn't exist (hopefully).

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -48,6 +48,9 @@ namespace AppInstaller::Manifest
         const char* const MissingManifestDependenciesNode = "Dependency not found: ";
         const char* const NoSuitableMinVersion = "No Suitable Minimum Version: ";
         const char* const FoundLoop = "Loop found.";
+        const char* const ExceededAppsAndFeaturesEntryLimit = "Only zero or one entry for Apps and Features can be specified for InstallerType portable.";
```
jedieaston, line 51: I'm curious, why aren't these localized like the rest of the strings in winget (unless they are just too hard to translate)?

JohnMcPMS: We have a layering problem right now w/r/t localization.  Localization is done in CLICore, because it always has a loc file (packaged or colocated for unpackaged).  CommonCore is also used for the util binary, and we have not required that it carry around a loc file to function.

There are some changes coming that might end up requiring us to deal with this (aka move localization awareness into CommonCore), and at that point we would move these to be localized.  I think we will also need to create a compiled in backstop from the checked in .resw, so that we can continue to have the same functionality if the loc file isn't there.

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
ryfu-msft: Fixed grammar

schemas/JSON/manifests/v1.2.0/manifest.defaultLocale.1.2.0.json
ryfu-msft: Updated maxItems from 128 to 256

src/AppInstallerCommonCore/Public/winget/ExperimentalFeature.h
```diff
@@ -23,6 +23,7 @@ namespace AppInstaller::Settings
             Dependencies = 0x1,
             // Before making DirectMSI non-experimental, it should be part of manifest validation.
             DirectMSI = 0x2,
+            PortableInstall = 0x3,
```
JohnMcPMS, line 26: These are bit flags values.  Most of the code doesn't care, but `ExperimentalFeature::GetAllFeatures` does.  And being able to support querying for multiple flags at the same time might in a future where we want that (why it was done that way in the first place.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -291,6 +294,24 @@ namespace AppInstaller::Settings
             return SettingMapping<Setting::InstallLocalePreference>::Validate(value);
         }
 
+        WINGET_VALIDATE_SIGNATURE(PortableAppUserRoot)
+        {
+            if (!std::filesystem::exists(value))
+            {
+                if (!std::filesystem::create_directories(value))
```
JohnMcPMS, line 301: Settings should not be creating directories.  Create a new `PathName` for `Runtime::GetPathTo`, retrieve the setting value and create the directory there.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/UserSettings.cpp
ryfu-msft: I created a new `PathName` for `PortableAppUserRoot`, `PortableAppMachineRootX64`, and `PortableAppMachineRootX86` and create the directories there after retrieving the setting value.

src/AppInstallerCommonCore/Public/winget/ExperimentalFeature.h
ryfu-msft: Ah my mistake, I didn't realize these were bit flags, changed the flag values for `PortableInstall` and `Max` so that the experimental feature shows up for `ExperimentalFeature::GetAllFeatures`

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -73,6 +75,8 @@ namespace AppInstaller::Settings
             return ExperimentalFeature{ "Show Dependencies Information", "dependencies", "https://aka.ms/winget-settings", Feature::Dependencies };
         case Feature::DirectMSI:
             return ExperimentalFeature{ "Direct MSI Installation", "directMSI", "https://aka.ms/winget-settings", Feature::DirectMSI };
+        case Feature::PortableInstall:
+            return ExperimentalFeature{ "Portable Installation", "portableInstall", "https://aka.ms/winget-settings", Feature::PortableInstall };
```
yao-msft, line 79: >https://aka.ms/winget-settings [](http://example.com/codeflow?start=85&length=30)

remember to update the settings md doc

ryfu-msft: Updated settings.md doc for this experimental feature

.github/actions/spelling/expect.txt
```diff
@@ -287,6 +287,7 @@ powertoys
 pri
 processthreads
 productcode
+PROGRAMFILES
```
yao-msft, line 290: nit: this should probably move to allow.txt as this is well known Windows env var

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -96,7 +96,33 @@
...
+          "default": "%PROGRAMFILES%/WinGet/Packages/"
+        }
+      }
+    },
+    "UninstallBehavior": {
+      "description": "Uninstall settings",
+      "type": "object",
+      "properties": {
+        "purge": {
+          "description": "Controls whether the default behavior for uninstall should remove all files and directories relevant to this package. Only applies to the portable installerType.",
```
yao-msft, line 122: >Only applies to the portable installerType. [](http://example.com/codeflow?start=144&length=43)

I think "portableAppPurge" will be more clear and consistent with above entries

schemas/JSON/manifests/v1.2.0/manifest.defaultLocale.1.2.0.json
```diff
@@ -35,6 +35,21 @@
...
         }
       }
+    },
+    "Documentation": {
+      "type": "object",
+      "properties": {
+        "DocumentDescription": {
+          "type": [ "string", "null" ],
+          "minLength": 1,
+          "maxLength": 100,
```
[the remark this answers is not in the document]
Trenly: The intent here is more of a concise description of what the document provides. Examples that I can think of may be -
* Troubleshooting guide
* Frequently asked questions
* User manual
* Instructions on how to install and use ABC or XYZ plugin with PackageName
* System Requirements

I think 100 would be more than enough, but perhaps this could be made 256 like a short description. I don't think we need an entire document here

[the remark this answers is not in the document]
yao-msft: I guess it's more like a title rather than the contents then, we use Label in the manifests for this concept. maybe DocumentLabel?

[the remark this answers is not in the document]
Trenly: DocumentLabel
DocumentTitle
DocumentName
DocumentType

I think any of them would work; I know that denelon had suggested DocumentDescription in the original issue, but I'm sure it's open to interpretation

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -209,7 +231,7 @@
       "type": [ "array", "null" ],
       "items": {
         "type": "string",
-        "pattern": "^[a-z][-a-z0-9\\.\\+]*$",
```
yao-msft, line 212: >"pattern": "^[a-z][-a-z0-9\\.\\+]*$", [](http://example.com/codeflow?start=8&length=37)

The initial regex is actually copied from appx manifest schema, if Windows supports starting with numbers, then I guess we'll go with this change.

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
```diff
@@ -466,9 +497,7 @@
           "$ref": "#/definitions/Scope"
         },
         "InstallerUrl": {
-          "type": "string",
-          "pattern": "^([Hh][Tt][Tt][Pp][Ss]?)://.+$",
-          "maxLength": 2048,
+          "$ref": "#/definitions/Url",
```
yao-msft, line 500: >"$ref": "#/definitions/Url", [](http://example.com/codeflow?start=10&length=28)

This url definition allows null, for InstallerUrl, we should keep the original definition which does not allow null

src/AppInstallerCLICore/ExecutionArgs.h
```diff
@@ -86,6 +86,10 @@ namespace AppInstaller::CLI::Execution
             CustomHeader, // Optional Rest source header
             AcceptSourceAgreements, // Accept all source agreements
             IncludeUnknown, // Used in Upgrade command to allow upgrades of packages with unknown versions
+            Rename, // Renames the file of the executable. Only applies to the portable installerType
```
yao-msft, line 89: >Rename [](http://example.com/codeflow?start=12&length=6)

nit: some of these might better fit in Instal behavior or Uninstallbehavior rather than Other category

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -130,12 +130,15 @@ bool HasExtension(std::string_view extension) const;
         AlreadyInstalled,
         Downgrade,
         BlockedByPolicy,
+        Custom
```
yao-msft, line 133: >Custom [](http://example.com/codeflow?start=8&length=6)

nit: we usually end with , for easier merging

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -130,12 +130,15 @@ bool HasExtension(std::string_view extension) const;
...
     };
 
     struct ExpectedReturnCode
     {
-        DWORD InstallerReturnCode;
-        ExpectedReturnCodeEnum ReturnResponse;
+        DWORD InstallerReturnCode{};
+        ExpectedReturnCodeEnum ReturnResponse{};
+        string_t ReturnResponseUrl{};
+        ExpectedReturnCode() {};
```
yao-msft, line 141: >ExpectedReturnCode() {}; [](http://example.com/codeflow?start=8&length=24)

nit: above {} and constructor are probably over verbose for simple data type structs

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
Trenly:
```suggestion
          "allOf": [
            {"$ref":"#/definitions/Url"},
            {"type":"string"}
          ],
```
This keeps the validation that it matches the URL parameter, but adds the restriction that it must be a string, without duplicating URL logic

src/AppInstallerCommonCore/Public/winget/ManifestLocalization.h
```diff
@@ -38,6 +41,12 @@ namespace AppInstaller::Manifest
         string_t AgreementUrl;
     };
 
+    struct Documentation
+    {
+        string_t DocumentDescription;
```
yao-msft, line 46: >DocumentDescription [](http://example.com/codeflow?start=17&length=19)

nit: DocumentLabel better fits the entry purpose

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -83,6 +84,10 @@ namespace AppInstaller::Settings
         EFDirectMSI,
         EnableSelfInitiatedMinidump,
         LoggingLevelPreference,
+        InstallIgnoreWarnings,
+        PortableAppUserRoot,
+        PortableAppMachineRoot,
+        UninstallPurge,
```
yao-msft, line 90: >UninstallPurge [](http://example.com/codeflow?start=8&length=14)

nit: UninstallPortablAppPurge

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -309,6 +312,36 @@ namespace AppInstaller::Runtime
                 result = GetKnownFolderPath(FOLDERID_Profile);
                 create = false;
                 break;
+            case PathName::PortableAppUserRoot:
+                result = Settings::User().Get<Setting::PortableAppUserRoot>();
+                if (result.empty())
+                {
+                    result /= GetKnownFolderPath(FOLDERID_LocalAppData);
+                    result /= s_SecureSettings_Base;
```
yao-msft, line 320: >s_SecureSettings_Base [](http://example.com/codeflow?start=30&length=21)

This should be another const with same value but not s_SecureSettings_Base

src/AppInstallerCommonCore/Public/winget/ExperimentalFeature.h
```diff
@@ -23,7 +23,8 @@ namespace AppInstaller::Settings
             Dependencies = 0x1,
             // Before making DirectMSI non-experimental, it should be part of manifest validation.
             DirectMSI = 0x2,
-            Max, // This MUST always be after all experimental features
+            PortableInstall = 0x4,
+            Max = 0x8, // This MUST always be after all experimental features
```
JohnMcPMS, line 27: Setting a value for `Max` makes mistakes more likely in the future.  If you want to make the fact that these are flags clearer, put a comment on the enum.

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -309,6 +312,36 @@ namespace AppInstaller::Runtime
                 result = GetKnownFolderPath(FOLDERID_Profile);
                 create = false;
                 break;
+            case PathName::PortableAppUserRoot:
+                result = Settings::User().Get<Setting::PortableAppUserRoot>();
+                if (result.empty())
+                {
+                    result /= GetKnownFolderPath(FOLDERID_LocalAppData);
```
JohnMcPMS, line 319:
```suggestion
                    result = GetKnownFolderPath(FOLDERID_LocalAppData);
```
Even though result was just verified to be empty, it feels unsafe to add a path to it.

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -309,6 +312,36 @@ namespace AppInstaller::Runtime
...
+                    result /= s_SecureSettings_Base;
+                    result /= s_DefaultPackagesDirectory;
+                }
+                create = true;
+                break;
+            case PathName::PortableAppMachineRootX64:
+                result = Settings::User().Get<Setting::PortableAppMachineRoot>();
+                if (result.empty())
+                {
+                    result /= GetKnownFolderPath(FOLDERID_ProgramFiles);
```
JohnMcPMS, line 329:
```suggestion
                    result = GetKnownFolderPath(FOLDERID_ProgramFilesX64);
```
We should use the appropriate value here rather than just letting our process architecture drive it.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -153,6 +161,24 @@ namespace AppInstaller::Manifest
                 resultErrors.emplace_back(ManifestError::ExeInstallerMissingSilentSwitches, ValidationError::Level::Warning);
             }
 
+            if (installer.InstallerType == InstallerTypeEnum::Portable)
+            {
+                if ((manifest.DefaultInstallerInfo.AppsAndFeaturesEntries.size() > 1 && installer.AppsAndFeaturesEntries.size() == 0) ||
```
yao-msft, line 166: >manifest.DefaultInstallerInfo.AppsAndFeaturesEntries.size() > 1 && installer.AppsAndFeaturesEntries.size() == 0) [](http://example.com/codeflow?start=21&length=112)

This condition does not seem possible, installer level fields directly inherit from DefaultInstallerInfo. If you really need to verify this, just verify the installer level directly. This means we are putting a burden on manifest author to always override portable installer with empty ones.

Same for below 2

src/AppInstallerCommonCore/Public/winget/ExperimentalFeature.h
Trenly: I'd also add that it might be clearer that it's a bit flag if the bit shift syntax were used instead of the hex representation -
```
            None = 0,
            Dependencies = 1 << 0,
            // Before making DirectMSI non-experimental, it should be part of manifest validation.
            DirectMSI = 1 << 1,
            PortableInstall = 1 << 2,
            Max , // This MUST always be after all experimental features

File: .github/actions/spelling/expect.txt (hunk 10 above)
Comment (ryfu-msft) [reply .github/actions/spelling/expect.txt]: Moved this to allow.txt

File: schemas/JSON/settings/settings.schema.0.2.json (hunk 11 above)
Comment (ryfu-msft) [reply schemas/JSON/settings/settings.schema.0.2.json]: changed to `purgePortableApp` as it's a boolean and should start with a verb

File: schemas/JSON/manifests/v1.2.0/manifest.defaultLocale.1.2.0.json (hunk 12 above)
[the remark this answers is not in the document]
Comment (ryfu-msft) [reply schemas/JSON/manifests/v1.2.0/manifest.defaultLocale.1.2.0.json:45]: Thanks for the suggestions.  Decided to go with DocumentLabel as this more closely aligns with the existing pattern (i.e. AgreementLabel)

File: schemas/JSON/manifests/v1.2.0/manifest.defaultLocale.1.2.0.json [hunk 25]
```diff
@@ -154,6 +169,23 @@
...
       "description": "The package release notes url"
     },
+    "PurchaseUrl": {
+      "$ref": "#/definitions/Url",
+      "description": "The purchase url for acquiring entitlement for the package."
+    },
+    "InstallationNotes": {
+      "type": [ "string", "null" ],
+      "minLength": 1,
+      "maxLength": 240,
```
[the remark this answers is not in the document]
ryfu-msft: Changed to 256

src/AppInstallerCLICore/ExecutionArgs.h
ryfu-msft: Moved these arguments around to their appropriate categories

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
ryfu-msft: added ,

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
ryfu-msft: Removed constructor and {} from ReturnResponseUrl as it already has a default initializer.

src/AppInstallerCommonCore/Public/winget/ManifestInstaller.h
```diff
@@ -53,6 +53,8 @@ namespace AppInstaller::Manifest
 
         std::map<DWORD, ExpectedReturnCodeEnum> ExpectedReturnCodes;
```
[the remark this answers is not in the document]
ryfu-msft: Removed 2nd map and created a struct containing the ReturnResponseEnum and ReturnResponseUrl.

src/AppInstallerCommonCore/Public/winget/ManifestLocalization.h
ryfu-msft: Changed to DocumentLabel

src/AppInstallerCommonCore/Public/winget/UserSettings.h
ryfu-msft: Changed to UninstallPurgePortableApp

src/AppInstallerCommonCore/Runtime.cpp
ryfu-msft: created separate consts specific to portables

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -309,6 +312,36 @@ namespace AppInstaller::Runtime
...
+                    result /= s_DefaultTempDirectory;
```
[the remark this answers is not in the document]
ryfu-msft: Changed to using newly defined const

src/AppInstallerCommonCore/Public/winget/ExperimentalFeature.h
ryfu-msft: Removed value from `Max`

src/AppInstallerCommonCore/Runtime.cpp
ryfu-msft: Changed to suggested

src/AppInstallerCommonCore/Runtime.cpp
ryfu-msft: Changed to `FOLDERID_ProgramFilesX64`

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -96,6 +96,14 @@ namespace AppInstaller::Manifest
                 resultErrors.emplace_back(ManifestError::InvalidFieldValue, "UpdateBehavior");
             }
 
+            for (const auto& unsupportedArgument : installer.UnsupportedArguments)
+            {
+                if (unsupportedArgument == UnsupportedArgumentEnum::Unknown)
+                {
+                    resultErrors.emplace_back(ManifestError::InvalidFieldValue, "UnsupportedArguments");
+                }
+            }
```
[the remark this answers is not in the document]
ryfu-msft: Removed this block from validation as I agree that this is not absolutely needed for validation

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -399,6 +429,18 @@ namespace AppInstaller::Manifest
...
                 std::move(fields_v1_1.begin(), fields_v1_1.end(), std::inserter(result, result.end()));
             }
+
+            if (manifestVersion >= ManifestVer{ s_ManifestVersionV1_2 })
+            {
+                std::vector<FieldProcessInfo> fields_v1_2 =
+                {
+                    { "PurchaseUrl", [this](const YAML::Node& value)->ValidationErrors { m_p_localization->Add<Localization::PurchaseUrl>(value.as<std::string>()); return {}; } },
+                    { "InstallationNotes", [this](const YAML::Node& value)->ValidationErrors { m_p_localization->Add<Localization::InstallationNotes>(value.as<std::string>()); return {}; } },
+                    { "Documentations", [this](const YAML::Node& value)->ValidationErrors { return ProcessDocumentationsNode(value); }, true },
```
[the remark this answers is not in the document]
ryfu-msft: Set to not require verified publisher

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
ryfu-msft: Removed additional condition and only check value at the installer level directly.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -228,9 +228,15 @@ namespace AppInstaller::Settings
...
         WINGET_VALIDATE_PASS_THROUGH(EFDependencies)
+        WINGET_VALIDATE_PASS_THROUGH(EFPortableInstall)
         WINGET_VALIDATE_PASS_THROUGH(TelemetryDisable)
         WINGET_VALIDATE_PASS_THROUGH(EFDirectMSI)
         WINGET_VALIDATE_PASS_THROUGH(EnableSelfInitiatedMinidump)
+        WINGET_VALIDATE_PASS_THROUGH(InstallIgnoreWarnings)
+        WINGET_VALIDATE_PASS_THROUGH(UninstallPurge)
+        WINGET_VALIDATE_PASS_THROUGH(PortableAppUserRoot)
+        WINGET_VALIDATE_PASS_THROUGH(PortableAppMachineRoot)
+
```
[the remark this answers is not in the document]
ryfu-msft: Removed extra newline

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
ryfu-msft: I decided to go with Trenly's latter suggestion in order to ensure that installerUrl is not null and that we aren't duplicating definitions for Url. I did not change the descriptions as I feel that would be make it slightly confusing since not all descriptions, labels, notes are the same.

schemas/JSON/manifests/v1.2.0/manifest.locale.1.2.0.json
```diff
@@ -149,6 +164,23 @@
...
+      "minLength": 1,
+      "maxLength": 240,
+      "description": "The notes displayed to the user upon completion of a package installation."      
+    },
+    "Documentations": {
+      "type": [ "array", "null" ],
+      "items": {
+        "$ref": "#/definitions/Documentation"
+      },
+      "maxItems": 256
```
Trenly, line 182: If documentations exists, then it should not be an empty array
```suggestion
      "minItems":1,
      "maxItems": 256
```

schemas/JSON/manifests/v1.2.0/manifest.singleton.1.2.0.json
```diff
@@ -691,6 +732,23 @@
...
+      "minLength": 1,
+      "maxLength": 256,
+      "description": "The notes displayed to the user upon completion of a package installation."
+    },
+    "Documentations": {
+      "type": [ "array", "null" ],
+      "items": {
+        "$ref": "#/definitions/Documentation"
+      },
+      "maxItems": 256
```
Trenly, line 750: If documentations exists, then it should not be an empty array
```suggestion
      "minItems":1,
      "maxItems": 256
```

schemas/JSON/manifests/v1.2.0/manifest.singleton.1.2.0.json
```diff
@@ -54,6 +54,21 @@
...
+          "type": [ "string", "null" ],
+          "minLength": 1,
+          "maxLength": 100,
+          "description": "The label of the documentation for providing software guides such as manuals and troubleshooting URLs."
+        },
+        "DocumentUrl": {
+          "$ref": "#/definitions/Url",
+          "description": "The documentation URL."
+        }
+      }
```
Trenly, line 70: Documentation shouldn't be able to contain empty objects in the array
```suggestion
      },
      "required": [
        "DocumentLabel",
        "DocumentUrl"
      ]
```

schemas/JSON/manifests/v1.2.0/manifest.locale.1.2.0.json
```diff
@@ -35,6 +35,21 @@
...
+          "type": [ "string", "null" ],
+          "minLength": 1,
+          "maxLength": 100,
+          "description": "The description of the documentation for providing software guides such as manuals and troubleshooting URLs."
+        },
+        "DocumentUrl": {
+          "$ref": "#/definitions/Url",
+          "description": "The documentation URL."          
+        }
+      }
```
Trenly, line 52: Documentation shouldn't be able to contain empty objects in the array
```suggestion
      },
      "required": [
        "DocumentLabel",
        "DocumentUrl"
      ]
```

schemas/JSON/manifests/v1.2.0/manifest.locale.1.2.0.json
```diff
@@ -35,6 +35,21 @@
           "description": "The agreement URL."
         }
       }
+    },
+    "Documentation": {
+      "type": "object",
+      "properties": {
+        "DocumentLabel": {
+          "type": [ "string", "null" ],
```
Trenly, line 43: Documents should always be labeled
```suggestion
          "type": "string",
```

schemas/JSON/manifests/v1.2.0/manifest.singleton.1.2.0.json
```diff
@@ -54,6 +54,21 @@
         }
       }
     },
+    "Documentation": {
+      "type": "object",
+      "properties": {
+        "DocumentLabel": {
+          "type": [ "string", "null" ],
```
Trenly, line 61: Documents should always be labeled
```suggestion
          "type": "string",
```

schemas/JSON/manifests/v1.2.0/manifest.singleton.1.2.0.json
```diff
@@ -54,6 +54,21 @@
...
+      "type": "object",
+      "properties": {
+        "DocumentLabel": {
+          "type": [ "string", "null" ],
+          "minLength": 1,
+          "maxLength": 100,
+          "description": "The label of the documentation for providing software guides such as manuals and troubleshooting URLs."
+        },
+        "DocumentUrl": {
+          "$ref": "#/definitions/Url",
```
Trenly, line 67: Documents should always have a URL
```suggestion
          "allOf": [
            {"$ref":"#/definitions/Url"},
            {"not" : {"type":"null"}}
          ],
```

schemas/JSON/manifests/v1.2.0/manifest.locale.1.2.0.json
```diff
@@ -35,6 +35,21 @@
...
+      "type": "object",
+      "properties": {
+        "DocumentLabel": {
+          "type": [ "string", "null" ],
+          "minLength": 1,
+          "maxLength": 100,
+          "description": "The description of the documentation for providing software guides such as manuals and troubleshooting URLs."
+        },
+        "DocumentUrl": {
+          "$ref": "#/definitions/Url",
```
Trenly, line 49: Documents should always have a URL
```suggestion
          "allOf": [
            {"$ref":"#/definitions/Url"},
            {"not" : {"type":"null"}}
          ],
```

Trenly (commented):
Looking at the documentations a bit more, it seems that with the proposed schema, the Documentations (if not null) could be an empty array, or an array of empty objects.

I also think that we should force quality here, and require that each Documentation require the URL and Description, as a documentation without a URL is meaningless and a URL without a description may be confusing

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
ryfu-msft: Removed regex constraint as it is no longer needed anymore.

Trenly:
@ryfu-msft As I was looking at the schema more, I realized that there is a lot of weird things that happen because things are allowed to be null; I think I corrected these for 1.0.0 and 1,1,0 over in #2056 - it may be a good idea to do the same with the 1.2.0 schema; I didn't want to do it before this gets merged to avoid conflicts, but if you're still working on this it might make sense to roll it into one

schemas/JSON/manifests/v1.2.0/manifest.locale.1.2.0.json
```diff
@@ -149,6 +164,23 @@
...
       "description": "The package release notes url"
     },
+    "PurchaseUrl": {
+      "$ref": "#/definitions/Url",
+      "description": "The purchase url for acquiring entitlement for the package."
+    },
+    "InstallationNotes": {
+      "type": [ "string", "null" ],
+      "minLength": 1,
+      "maxLength": 240,
```
yao-msft, line 174: >240 [](http://example.com/codeflow?start=19&length=3)

nit: update here as well

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -22,6 +24,9 @@ namespace AppInstaller::Runtime
         constexpr std::string_view s_SecureSettings_Base = "Microsoft/WinGet"sv;
         constexpr std::string_view s_SecureSettings_UserRelative = "settings"sv;
         constexpr std::string_view s_SecureSettings_Relative_Unpackaged = "win"sv;
+        constexpr std::string_view s_PortableAppUserRoot = "Microsoft/WinGet"sv;
+        constexpr std::string_view s_PortableAppMachineRoot = "WinGet"sv;
+        constexpr std::string_view s_PackagesDirectory = "Packages"sv;
```
yao-msft, line 29: >s_PackagesDirectory [](http://example.com/codeflow?start=35&length=19)

nit: PortablePackagesDirectory

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -651,16 +710,15 @@ namespace AppInstaller::Manifest
...
 
         for (auto const& entry : returnCodesNode.Sequence())
         {
             ExpectedReturnCode returnCode;
             m_p_expectedReturnCode = &returnCode;
             auto errors = ValidateAndProcessFields(entry, ExpectedReturnCodesFieldInfos);
             std::move(errors.begin(), errors.end(), std::inserter(resultErrors, resultErrors.end()));
-
-            if (!returnCodes.insert({ returnCode.InstallerReturnCode, returnCode.ReturnResponse }).second)
+            if (!returnCodes.insert({ returnCode.InstallerReturnCode, ExpectedReturnCodeInfo{ returnCode.ReturnResponse, returnCode.ReturnResponseUrl} }).second)
```
yao-msft, line 721: >ExpectedReturnCodeInfo [](http://example.com/codeflow?start=70&length=22)

nit: could be removed

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -130,6 +136,10 @@ namespace AppInstaller::Settings
         SETTINGMAPPING_SPECIALIZATION(Setting::NetworkDOProgressTimeoutInSeconds, uint32_t, std::chrono::seconds, 60s, ".network.doProgressTimeoutInSeconds"sv);
         SETTINGMAPPING_SPECIALIZATION(Setting::InstallLocalePreference, std::vector<std::string>, std::vector<std::string>, {}, ".installBehavior.preferences.locale"sv);
         SETTINGMAPPING_SPECIALIZATION(Setting::InstallLocaleRequirement, std::vector<std::string>, std::vector<std::string>, {}, ".installBehavior.requirements.locale"sv);
+        SETTINGMAPPING_SPECIALIZATION(Setting::InstallIgnoreWarnings, bool, bool, false, ".installBehavior.ignoreWarnings"sv);
+        SETTINGMAPPING_SPECIALIZATION(Setting::PortableAppUserRoot, std::string, std::string, {}, ".installBehavior.portableAppUserRoot"sv);
+        SETTINGMAPPING_SPECIALIZATION(Setting::PortableAppMachineRoot, std::string, std::string, {}, ".installBehavior.portableAppMachineRoot"sv);
+        SETTINGMAPPING_SPECIALIZATION(Setting::UninstallPurgePortableApp, bool, bool, false, ".uninstallBehavior.purge"sv);
```
yao-msft, line 142: >purge [](http://example.com/codeflow?start=113&length=5)

update

src/AppInstallerRepositoryCore/Rest/Schema/1_1/Json/ManifestDeserializer_1_1.cpp
```diff
@@ -139,7 +139,7 @@ namespace AppInstaller::Repository::Rest::Schema::V1_1::Json
                     // Only add when it is valid
                     if (installerReturnCode != 0 && returnResponse != ExpectedReturnCodeEnum::Unknown)
                     {
-                        if (!installer.ExpectedReturnCodes.insert({ installerReturnCode, returnResponse }).second)
+                        if (!installer.ExpectedReturnCodes.insert({ installerReturnCode, AppInstaller::Manifest::ManifestInstaller::ExpectedReturnCodeInfo{returnResponse} }).second)
```
yao-msft, line 142: >{returnResponse} [](http://example.com/codeflow?start=154&length=16)

maybe { InstallationNotes, "" }to be clear that v1.1 does not have response string

schemas/JSON/manifests/v1.2.0/manifest.locale.1.2.0.json
ryfu-msft: updated to 256

src/AppInstallerCommonCore/Runtime.cpp
ryfu-msft: renamed to PortablePackagesDirectory

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
ryfu-msft: removed

src/AppInstallerRepositoryCore/Rest/Schema/1_1/Json/ManifestDeserializer_1_1.cpp
ryfu-msft: updated to suggestion

schemas/JSON/manifests/v1.2.0/manifest.installer.1.2.0.json
ryfu-msft: I had to revert this change back to the original. Turns out this will break wingetcreate when it tries to generate the model based on the schema. The library we use to generate the model doesn't fully support schema keywords like "allOf" and "oneOf" so its better to leave it as is.

ryfu-msft:
> Looking at the documentations a bit more, it seems that with the proposed schema, the Documentations (if not null) could be an empty array, or an array of empty objects.
> 
> I also think that we should force quality here, and require that each Documentation require the URL and Description, as a documentation without a URL is meaningless and a URL without a description may be confusing

I discussed with the team and I decided to not follow through with these changes. There are a lot of components such as WingetCreate, WingetSvc, and other 3rd parties that rely on the schema and adding more constraints could be breaking. These changes will require more thorough testing that I won't address in this PR, so I will stick with the existing pattern.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1275,4 +1275,16 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="SystemArchitecture" xml:space="preserve">
     <value>System Architecture</value>
   </data>
+  <data name="PreserveArgumentDescription" xml:space="preserve">
+    <value>Retains all files and directories created by the package (portable)</value>
+  </data>
+  <data name="PurgeArgumentDescription" xml:space="preserve">
+    <value>Deletes all files and directories created by the package (portable)</value>
```
JohnMcPMS, line 1282:
```suggestion
    <value>Deletes all files and directories in the package directory (portable)</value>
```
We cannot delete them if they were created outside of the package directory (and I don't think we should delete them unless we created the directory either).

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -130,6 +136,10 @@ namespace AppInstaller::Settings
         SETTINGMAPPING_SPECIALIZATION(Setting::NetworkDOProgressTimeoutInSeconds, uint32_t, std::chrono::seconds, 60s, ".network.doProgressTimeoutInSeconds"sv);
         SETTINGMAPPING_SPECIALIZATION(Setting::InstallLocalePreference, std::vector<std::string>, std::vector<std::string>, {}, ".installBehavior.preferences.locale"sv);
         SETTINGMAPPING_SPECIALIZATION(Setting::InstallLocaleRequirement, std::vector<std::string>, std::vector<std::string>, {}, ".installBehavior.requirements.locale"sv);
+        SETTINGMAPPING_SPECIALIZATION(Setting::InstallIgnoreWarnings, bool, bool, false, ".installBehavior.ignoreWarnings"sv);
+        SETTINGMAPPING_SPECIALIZATION(Setting::PortableAppUserRoot, std::string, std::string, {}, ".installBehavior.portableAppUserRoot"sv);
```
JohnMcPMS, line 140: I think that these settings should return a `std::filesystem::path` to prevent UTF-8 to path conversion bugs.  The path should also be verified to be good.  That should allow the settings fallback mechanism to kick in if the path is not proper; it needs to be a full path, not relative, etc.

Also, if you don't manually convert to the UTF-8 string to UTF-16, you will get the above bug (which you have in `Runtime.cpp`).

JohnMcPMS (changes requested):

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: Changed to suggested

src/AppInstallerCommonCore/Public/winget/UserSettings.h
ryfu-msft: changed settings to return a path as well as added settings validation to ensure that the path is absolute. Also added a test case checking this behavior when providing a relative path.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -234,8 +234,22 @@ namespace AppInstaller::Settings
         WINGET_VALIDATE_PASS_THROUGH(EnableSelfInitiatedMinidump)
         WINGET_VALIDATE_PASS_THROUGH(InstallIgnoreWarnings)
         WINGET_VALIDATE_PASS_THROUGH(UninstallPurgePortableApp)
-        WINGET_VALIDATE_PASS_THROUGH(PortableAppUserRoot)
-        WINGET_VALIDATE_PASS_THROUGH(PortableAppMachineRoot)
+
+        WINGET_VALIDATE_SIGNATURE(PortableAppUserRoot)
+        {
+            std::filesystem::path root = { value };
```
JohnMcPMS, line 240:
```suggestion
            std::filesystem::path root = Utility::ConvertToUTF16(value);
```

ryfu-msft: Converted to UTF16

JohnMcPMS (approved):

PR #2035, opened 2022-03-21
Add support for InProc Com invocation

yao-msft:
## Change

Added a shim dll to support in-proc Com activation of winget Com apis.

## Validation
Validated by writing a sample caller consuming winget Com apis through in-proc activation. Codes run fine with in-proc activation.

## Next
- State separation and better support for System caller
- Add Com e2e tests and remove previous workarounds(i.e. Microsoft.Management.Deployment.Client and Microsoft.Management.Deployment.Server.Test)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2035)


src/AppInstallerCLI/AppInstallerCLI.vcxproj
```diff
@@ -191,7 +191,7 @@
     <Link>
       <SubSystem>Console</SubSystem>
       <GenerateWindowsMetadata>false</GenerateWindowsMetadata>
-      <AdditionalDependencies Condition="'$(Configuration)'=='Debug'">wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;winhttp.lib;onecoreuap.lib;msi.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies Condition="'$(Configuration)'=='Debug'">%(AdditionalDependencies)</AdditionalDependencies>
```
JohnMcPMS, line 194: Isn't this now a no-op?

src/Microsoft.Management.Deployment.InProc/dllmain.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include <WindowsPackageManager.h>
+
+extern "C"
+{
+    BOOL WINDOWS_PACKAGE_MANAGER_API_CALLING_CONVENTION DllMain(
```
JohnMcPMS, line 8: I think we should be using a macro for calling convention that is more in line with what this expects, not just whatever we happen to use.

src/Microsoft.Management.Deployment.InProc/dllmain.cpp
```diff
@@ -0,0 +1,48 @@
...
+        REFCLSID rclsid,
+        REFIID riid,
+        LPVOID* ppv)
+    {
+        RETURN_HR(WindowsPackageManagerInProcModuleGetClassObject(rclsid, riid, ppv));
+    }
+
+    WINDOWS_PACKAGE_MANAGER_API DllCanUnloadNow()
+    {
+        return WindowsPackageManagerInProcModuleTerminate() ? S_OK : S_FALSE;
```
JohnMcPMS, line 46: Is this the right way to do it?  It seems like there are potential error paths here, although given the calling patterns it might be ok.

src/WindowsPackageManager/main.cpp
```diff
@@ -54,4 +56,34 @@ extern "C"
         RETURN_HR(::Microsoft::WRL::Module<::Microsoft::WRL::ModuleType::OutOfProc>::GetModule().UnregisterObjects());
     }
     CATCH_RETURN();
+
+    WINDOWS_PACKAGE_MANAGER_API WindowsPackageManagerInProcModuleInitialize() try
+    {
+        ::Microsoft::WRL::Module<::Microsoft::WRL::ModuleType::InProc>::Create();
```
JohnMcPMS, line 62: We don't need an initialization of our code?  I suppose that might be what the state separation config would do.
We don't need to `RegisterObjects`?  Is that out-of-proc only?

JohnMcPMS (approved):
It would be interesting to have the sample caller app be able to swap to this as well for test purposes.

src/AppInstallerCLI/AppInstallerCLI.vcxproj
yao-msft: Yes, I leave the node here in unlikely case we'll need something in the future.

src/Microsoft.Management.Deployment.InProc/dllmain.cpp
yao-msft: Yes, this pattern is what I found in most other com dlls we wrote in the past.

src/WindowsPackageManager/main.cpp
yao-msft: Correct. RegisterObjects is for out of proc only. In proc automatically recognizes the creator map.

Niranjan-Jyothi:
Hey, @yao-msft @denelon @Trenly I have been trying to use In-Proc Winget COM APIs directly onto my c# app (.net 8) using CsWinRT to eventually install/search applications but the initial PackageManger object creation itself fails with below message

System.TypeInitializationException: 'The type initializer for 'WinRT.ActivationFactory`1' threw an exception.'
Inner Exception
COMException: Class not registered (0x80040154 (REGDB_E_CLASSNOTREG))

image

Can someone point me the right direction on how to achieve this maybe a straightforward example code snippet or documentation. Analysing the powershell (as it uses In-Proc) was getting a bit complicated and did not see any registrations before hand in the code instead the class objects (PackageManager etc) were directly being created in the 
winget-cli\src\PowerShell\Microsoft.WinGet.Client.Engine\Helpers\**ManagementDeploymentFactory.cs** file

P.S : The reason for going with this approach was because we wanted to execute the commands in the SYSTEM context and using direct powershell modules was not approved.

PR #2042, opened 2022-03-24
Use AppsAndFeatures name and publisher

JohnMcPMS:
## Change
Places the values from the `AppsAndFeaturesEntries` into the index, and uses them in the post-install ARP entry detection.

## Validation
Added a test to ensure that the values were added and searched in a simple case.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2042)

### DIFF
--- a/src/AppInstallerCLI.sln
+++ b/src/AppInstallerCLI.sln
@@ -147,6 +147,17 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "WindowsPackageManager", "Wi
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "COMServer", "COMServer\COMServer.vcxitems", "{409CD681-22A4-469D-88AE-CB5E4836E07A}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "manifests", "manifests", "{6FF8E5FE-C2D3-4991-8519-6FD12A457DDE}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "v1.2.0", "v1.2.0", "{6FA346E3-2171-4D82-95FE-3EC55DC8FA1F}"
+	ProjectSection(SolutionItems) = preProject
+		..\schemas\JSON\manifests\v1.2.0\manifest.defaultLocale.1.2.0.json = ..\schemas\JSON\manifests\v1.2.0\manifest.defaultLocale.1.2.0.json
+		..\schemas\JSON\manifests\v1.2.0\manifest.installer.1.2.0.json = ..\schemas\JSON\manifests\v1.2.0\manifest.installer.1.2.0.json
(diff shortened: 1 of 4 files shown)


src/AppInstallerCLI.sln
```diff
@@ -147,6 +147,17 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "WindowsPackageManager", "Wi
...
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "v1.2.0", "v1.2.0", "{6FA346E3-2171-4D82-95FE-3EC55DC8FA1F}"
+	ProjectSection(SolutionItems) = preProject
+		..\schemas\JSON\manifests\v1.2.0\manifest.defaultLocale.1.2.0.json = ..\schemas\JSON\manifests\v1.2.0\manifest.defaultLocale.1.2.0.json
+		..\schemas\JSON\manifests\v1.2.0\manifest.installer.1.2.0.json = ..\schemas\JSON\manifests\v1.2.0\manifest.installer.1.2.0.json
+		..\schemas\JSON\manifests\v1.2.0\manifest.locale.1.2.0.json = ..\schemas\JSON\manifests\v1.2.0\manifest.locale.1.2.0.json
+		..\schemas\JSON\manifests\v1.2.0\manifest.singleton.1.2.0.json = ..\schemas\JSON\manifests\v1.2.0\manifest.singleton.1.2.0.json
+		..\schemas\JSON\manifests\v1.2.0\manifest.version.1.2.0.json = ..\schemas\JSON\manifests\v1.2.0\manifest.version.1.2.0.json
+	EndProjectSection
+EndProject
```
yao-msft, line 160: Is this intentional? Curious why do we list the v1.2 schemas specifically. Also we already have the Project/ManifestSchema that contains organized manifests.

src/AppInstallerRepositoryCore/Microsoft/Schema/1_2/Interface_1_2.cpp
```diff
@@ -13,27 +13,37 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_2
...
-                Utility::NormalizedString value = normalizer.NormalizeName(Utility::FoldCase(localization.Get<Manifest::Localization::PackageName>())).Name();
-                if (std::find(out.begin(), out.end(), value) == out.end())
-                {
-                    out.emplace_back(std::move(value));
-                }
+                AddNormalizedName(normalizer, localization.Get<Manifest::Localization::PackageName>(), out);
+            }
+        }
+
+        void AddNormalizedPublisher(const Utility::NameNormalizer& normalizer, const Manifest::string_t& publisher, std::vector<Utility::NormalizedString>& out)
```
yao-msft, line 33: >AddNormalizedPublisher [](http://example.com/codeflow?start=13&length=22)

nit: looks like this implementation is same as the AddNormalizedName above, shall we combine them like AddNormalizedValue?

src/AppInstallerCLI.sln
JohnMcPMS: Didn't realize they were in the shared projects, removed the whole thing then.

yao-msft (approved):
:shipit:

PR #2044, opened 2022-03-24
Add heuristics for matching packages to ARP after installing

lechacon:
This adds heuristics for matching ARP data with a package after installing it, for better correlation later.
* Defined a type for algorithms measuring how much an ARP entry matches with a package. We will ultimately use only one, but we keep several for comparison and experimenting.
* After installing, we will try to use the matching heuristic to correlate one of the ARP changes with the package. If we find a match, we will store the product code in the tracking catalog for future use.
* Added tests to evaluate each heuristic. The test would take some data and create a summary of how many (true|false)(positives|negatives) we got and report them for each heuristic. For the heuristic we ultimately use, we will want to also add a check for how good the results have to be.
* Added some test data with package names and publisher as coming from the manifest and as seen in ARP after installing. This was obtained by installing several Store apps and inspecting the ARP after the installation, with some manual editing of the data.

Note that this is still a work in progress. It doesn't work yet and it is not very clean..

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2044)


src/AppInstallerRepositoryCore/ARPCorrelation.cpp
```diff
@@ -0,0 +1,205 @@
...
+        };
+
+        double EditDistanceScore(std::string_view sv1, std::string_view sv2)
+        {
+            // Naive implementation of Levenshtein distance (scaled over the string size)
+            const double EditCost = 0.5;
+            const double AddCost = 1;
+
+            // Do it ignoring case
+            auto s1 = Utility::ToLower(sv1);
```
JohnMcPMS, line 41: Use `FoldCase` for user input strings like this.

src/AppInstallerCLITests/Correlation.cpp
```diff
@@ -0,0 +1,159 @@
...
+        entry.Publisher = testCase.ArpPublisher;
+        entry.IsNewOrUpdated = true;
+        arpEntries.push_back(entry);
+        // Add a couple of ARP entries as noise
+        for (size_t i = 0; i < std::min((size_t)0, allARPEntries.size()); ++i)
+        {
+            arpEntries.push_back(allARPEntries[i]);
+        }
+
+        auto match = measure.GetBestMatchForManifest(manifest, arpEntries);
```
JohnMcPMS, line 113: I think that the scoring algorithms should be only that; give the manifest and a single ARP entry and they compute a score in [0, 1.0].  Or to allow for optimization, construct them with a manifest so that they can pre-compute anything that is manifest specific, then compute a score per ARP entry.  A confidence threshold would be defined globally to standardize on the minimum for considering something a possible match.

Then the overall score would factor in other data, like new/updated entry, version matching, etc.  But that is agnostic of the name and publisher matching algorithm, and should mostly just be attempting to tiebreak or boost our confidence.

This test harness is about rating the name and publisher matching algorithm specifically, and would be a good place to also understand other metrics, like the gap between the first and second highest confidence intervals.  It having all of the scores enables that kind of investigation.

lechacon: This `GetBestMatchForManifest()` is defined in the base class. It loops over all the entries and finds the one with the highest score (that is also over the threshold). It's not defined by each algorithm; those only do what you say (assign a score).

The way I'm thinking about it, the scoring algorithms consider everything (name, version, if it is new), and the fact that we are currently only looking at name string matching is incidental. If we add scores for other things, the string matching can just be pushed one layer below these algorithm classes.

I'm making the threshold specific to each algorithm and not restricted to [0,1] because I think that is hard to normalize across multiple algorithms. Specially if we consider how to mix multiple scores into one. For example, I'm thinking that we could compare mixing the scores for name and publisher (or anything else) by a weighted sum, or by multiplying them, or by summing squares, and that would all affect the threshold and range.

JohnMcPMS: Ok, I didn't look closely at that, but I did find the base class header odd, so that is why.  I believe there may be value in being able to run multiple passes, such as:
1. Run a direct equality matcher looking for the easy cases where we actually have exact data
2. Run the existing normalized name and publisher matcher
3. Run these more complex ones we are working on now, which are slower, but better at matching

With this design, it is more complicated using a base class with implementation built in (not impossible, just a bit more convoluted).  If the matcher interface were separated out, then it would be easier to have a single orchestration class that can use one or more matchers.  Again, not impossible to do with the current base class, but I think you end up needing to wrap the functionality anyway.

The overall scoring algorithm considers everything, but I don't think that each name+publisher matching algorithm should separately consider other fields.  Not that they should be disallowed, as I could see potential value in versions occurring in the name, but generally those are separate algorithms.  If we want to consider different version matching heuristics, we should do so as its own thing.

So I guess where I'm headed is that a pure interface class makes generic matchers possible, and a single orchestration class/function allows us to combine them as we see fit.

> I'm making the threshold specific to each algorithm and not restricted to [0,1] because I think that is hard to normalize across multiple algorithms.

Unless the algorithm actually returns completely unbounded values, numbers in the range [0, 1] are always possible.  This confidence interval scoring is very common, and having a normalized range is what makes it possible to combine scores from different aspects.  Creating a standardize threshold for meaning also allows the different algorithms to be interchangeable to the code above.  If mine thinks that 0.5+ is a solid match, but yours thinks that 0.9+ is, then how can the code above combine them?  If the answer is "use the threshold", then the scoring algorithm could do the same to normalize to the standardized value.  And it can do it with knowledge of its statistics; a linear mapping may not be the best approach for all cases.

lechacon: I made some changes based on what you said (but after re-reading it, I think it doesn't exactly match what you wanted...).

I kept all of your original code for matching ARP using product code and normalized names, and then the heuristics only run if that doesn't return anything. That's intended to be part of the orchestrator you mentioned.

I added a base class for name+publisher matchers that take the strings and return a value in [0,1]. I kept the 3 sample matchers I had earlier. I think this is where you can plug in the matching algorithm you wanted to implement.

I kept the base class for the overall algorithm, and added an implementation that for now considers only the string matching. This was intended to be the orchestrator for mix-and-matching smaller matchers, but I just realized that I didn't do it the way you suggested because it doesn't allow for doing a full pass with a single matcher, then another one with other matcher if needed.

I also added implementations of that base class to the tests, which use only a single string matching algorithm. That way we can evaluate just the accuracy of those, without considering anything else we may end up adding on top in the orchestrator.

I still need to test this E2E and play with the matching algorithms.

JohnMcPMS: Yes, it is now worse for my purposes, as I need more ARP information than you are providing.  For my algorithm I want something more like:
```
struct IARPMatchConfidenceAlgorithm
{
    double Compute(const ARPEntry& entry) const;
};

struct WordsARPMatchConfidenceAlgorithm : public IARPMatchConfidenceAlgorithm
{
    WordsARPMatchConfidenceAlgorithm(const Manifest& manifest);
    double Compute(const ARPEntry& entry) const override;
};
```
Where the ARPEntry contains other information, like language, that I was going to leverage.

src/AppInstallerCLICore/ExecutionContextData.h
```diff
@@ -190,6 +191,12 @@ namespace AppInstaller::CLI::Execution
             using value_t = std::vector<std::tuple<Utility::LocIndString, Utility::LocIndString, Utility::LocIndString>>;
         };
 
+        template <>
+        struct DataMapping<Data::ProductCodeFromARP>
+        {
+            using value_t = std::vector<Utility::LocIndString>;
```
JohnMcPMS, line 197: This needs to be the entire `AppsAndFeaturesEntry`, not just the product code.

lechacon: Updated to include the whole AppAndFeaturesEntry, although I'm not entirely sure if I did it The Right Way®

src/AppInstallerCLITests/Correlation.cpp
lechacon: I'll leave those changes for the future as we update the matching algorithm.

src/AppInstallerRepositoryCore/ARPCorrelation.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+                auto itr = std::lower_bound(arpSnapshot.begin(), arpSnapshot.end(), entryKey);
+                if (itr == arpSnapshot.end() || *itr != entryKey)
+                {
+                    changes.emplace_back(std::move(entry));
+                }
+            }
+        }
+
+        // Also attempt to find the entry based on the manifest data
```
JohnMcPMS, line 274: I would like this older matching mechanism to fit into the same infrastructure so that the overall could be something like:
```
foreach (algorithm : { ... })
{
   algorithm.Init(manifest)
   if (Correlate(algorithm, ARP) > threshold)
       return that;
}
```
Thoughts for the future, doesn't need to change now.

src/AppInstallerCLITests/Correlation.cpp
```diff
@@ -0,0 +1,323 @@
...
+            {
+                ++result.TrueMismatches;
+            }
+        }
+    }
+
+    return result;
+}
+
+void ReportResults(ResultSummary results)
```
JohnMcPMS, line 145: I would include some details on the time that this is taking as well, specifically the average time to do a correlation.  Most useful in the "full" test case where there are multiple ARP entries to choose from, reporting the average time it took to check what would be a real world correlation would be good.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/ARPCorrelation.cpp
```diff
@@ -0,0 +1,442 @@
...
+        }
+        else
+        {
+            return 0;
+        }
+    }
+
+    double EditDistanceNormalizedNameAndPublisherCorrelationMeasure::GetMatchingScore(std::string_view packageName, std::string_view packagePublisher, std::string_view arpName, std::string_view arpPublisher) const
+    {
+        NameNormalizer normer(NormalizationVersion::Initial);
```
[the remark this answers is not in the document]
lechacon: I changed it to have an object initialized with the manifest. It did make a noticeable difference in the time it took. :)
Since I was doing that, I also changed it to use the `IARPMatchConfidenceAlgorithm` interface you had suggested earlier.

> On top of that, for now the name and publisher values coming out of the ARP snapshot are already normalized which can save some time.
I tried not normalizing them, but then the success rate dropped significantly. Seems like they are not normalized? So I'm keeping that normalization, although it did make a difference in the time removing that normalization :(

src/AppInstallerRepositoryCore/ARPCorrelation.cpp
```diff
@@ -0,0 +1,442 @@
...
+            else
+            {
+                return s_algorithm;
+            }
+        }
+
+        double EditDistanceScore(std::string_view sv1, std::string_view sv2)
+        {
+            // Naive implementation of edit distance (scaled over the string size)
+            // TODO: This implementation does not consider multi-byte symbols.
```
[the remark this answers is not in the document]
lechacon: Added a `ConvertToUTF32`. This function now takes `std::u32string`; the strings from the manifest are converted only once when initializing the algorithm object.

src/AppInstallerRepositoryCore/ARPCorrelation.cpp
```diff
@@ -0,0 +1,442 @@
...
+                // This considers only the matching between name/publisher.
+                // It ignores versions and whether the ARP entry is new.
+                const auto packageName = manifestLocalization.Get<Localization::PackageName>();
+                const auto packagePublisher = manifestLocalization.Get<Localization::Publisher>();
+
+                const auto arpNames = arpEntry.Entry->GetMultiProperty(PackageVersionMultiProperty::Name);
+                const auto arpPublishers = arpEntry.Entry->GetMultiProperty(PackageVersionMultiProperty::Publisher);
+
+                // TODO: Comments say that these should match, but it seems they don't always do
+                if (arpNames.size() != arpPublishers.size())
```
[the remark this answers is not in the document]
lechacon: I changed these to use the `PackageVersionProperty::Name` and `::Publisher`. I copied the code to get the publisher from your PR.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -506,172 +507,70 @@ namespace AppInstaller::CLI::Workflow
...
-            }
-
-            // We can only get the source identifier from an active source
-            std::string sourceIdentifier;
-            if (context.Contains(Execution::Data::PackageVersion))
-            {
-                sourceIdentifier = context.Get<Execution::Data::PackageVersion>()->GetProperty(PackageVersionProperty::SourceIdentifier);
-            }
-
-            Logging::Telemetry().LogSuccessfulInstallARPChange(
```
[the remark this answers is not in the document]
lechacon: I had moved it down to the function doing the correlation; but now it's back here. I changed the helper to return the count of changes/matches, although I'm keeping that count to only consider the exact matches from the source search as I couldn't figure out a good way to keep the count consistent across the multiple "passes".

Do you have any ideas how to count the matching manifests when sometimes we use the exact matching and sometimes the confidence measures?

src/AppInstallerCLITests/Correlation.cpp
lechacon: Scratch that last comment; I updated it to use the interface you suggested. Ended up cleaner with it :D

src/AppInstallerCLITests/Correlation.cpp
lechacon: Added total and average time to what is reported here.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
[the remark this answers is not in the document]
JohnMcPMS: As long as we can reason about the meaning, anything is fine.  It can stay using the exact same values as before, just with a better guess.  I thought we might use these values in some way to find things that weren't correlating, but it turned out to be very easy to find them 😉 

So basically, these numbers are probably not important.  Don't spend time trying to improve them, and if you think they are broken, we might consider just reporting 0 for all of them.

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -154,6 +154,43 @@ namespace AppInstaller::Utility
...
 
+    std::u32string ConvertToUTF32(std::string_view input)
+    {
+        if (input.empty())
+        {
+            return {};
+        }
+
+        UErrorCode errorCode = UErrorCode::U_ZERO_ERROR;
+        auto utf32ByteCount= ucnv_convert("UTF-32", "UTF-8", nullptr, 0, input.data(), static_cast<int32_t>(input.size()), &errorCode);
```
JohnMcPMS, line 165: If there is a Windows API to convert to UTF-32, they hid it well...

JohnMcPMS (approved):

src/AppInstallerCLITests/TestData/InputARPData.txt
```diff
@@ -82,13 +76,10 @@ XPDP2X1MMZ4KR8|Ashampoo Burning Studio 23|Ashampoo|Ashampoo Burning Studio 23|23
 XPFCFBB4FB3D6D|Emjysoft Cleaner|Emjysoft|Emjysoft Cleaner 2022 v4.1|4.1|Emjysoft|{167B1302-A739-42DE-BBD2-4C2F13D1FF51}_is1
 XPFCFKCNNTXGQD|Yandex Browser|Yandex|Yandex|21.9.1.686|ООО «ЯНДЕКС»|YandexBrowser
 XPFCFL5ZTNFGD7|Wondershare Anireel|WONDERSHARE GLOBAL LIMITED|Wondershare Anireel(Build 1.6.2)||Wondershare Software|Wondershare Anireel_is1
-XPFCFL5ZTNFGD7|Wondershare Anireel|WONDERSHARE GLOBAL LIMITED|Wondershare Helper Compact 2.6.0|2.6.0|Wondershare|{5363CE84-5F09-48A1-8B6C-6BB590FFEDF2}_is1
 XPFCG86X2PGLDJ|Christmas Elf by Pothos|Pothos|Christmas Elf|||ChristmasElf
 XPFCGHHXNH4WBW|Biblioteca e Prestiti Librari|Esposito Software di M. G. Caputo|Gestione Biblioteca e Prestiti Librari 3.0 Demo||Copyright Esposito Software|Gestione Biblioteca e Prestiti Librari 3.0 Demo_is1
 XPFCWP0SQWXM3V|CCleaner|Piriform Software Ltd|CCleaner|5.89|Piriform|CCleaner
 XPFCXFRDJ8VGPT|Домашняя Бухгалтерия|Keepsoft|Äîìàøíÿÿ áóõãàëòåðèÿ Lite|7.2|Keepsoft|Äîìàøíÿÿ áóõãàëòåðèÿ Lite
```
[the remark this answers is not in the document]
lechacon: It's like this in the original registry:
image

PR #2056, opened 2022-03-30
Make schemas more strict

Trenly:

-----

The current schemas allow for some weird things to pass validation. Some JSON examples are at the bottom [here](https://www.jsonschemavalidator.net/s/jbqOELwy)

This PR does a few things
* Makes any field which is a string not able to be null - Due to the way schema validation works, this should mean that if a string property is present, then it is required to be not null, and, if the property is not present, it doesn't get validated against anyways, so this should still pass validation
* Makes all array types have a minimum of 1 item - There should never be a case where an array would be empty. If such a case exists, it would be for an installer level node to override a root level node, in which case the root level node should be moved to the installer level. All other cases, the property would be excluded from the manifest
* Makes object types have a minimum of 1 property - There should never be a case where an object would be empty. If such a case exists, it would again be for the case of overriding a root level node. 
* Makes any field which is a boolean not able to be null - This should mean that if a boolean property is present then it is required to be either true or false, and if the property is not present, it doesn't get validated anyways and should still pass validation
* Makes `InstallerReturnCode` and `ReturnResponse` required for each expected return code
* Changes the logic of markets to be easier to read - It should be functionally the same, but is a bit simpler

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2056)

### DIFF
--- a/schemas/JSON/manifests/v1.0.0/manifest.defaultLocale.1.0.0.json
+++ b/schemas/JSON/manifests/v1.0.0/manifest.defaultLocale.1.0.0.json
@@ -4,13 +4,13 @@
   "description": "A representation of a multiple-file manifest representing a default app metadata in the OWC. v1.0.0",
   "definitions": {
     "Url": {
-      "type": [ "string", "null" ],
+      "type": "string",
       "pattern": "^([Hh][Tt][Tt][Pp][Ss]?)://.+$",
       "maxLength": 2048,
       "description": "Optional Url type"
     },
     "Tag": {
-      "type": [ "string", "null" ],
+      "type": "string",
       "minLength": 1,
       "maxLength": 40,
       "description": "Package moniker or tag"
@@ -56,7 +56,7 @@
       "description": "The publisher privacy page or the package privacy page"
     },
     "Author": {
-      "type": [ "string", "null" ],
+      "type": "string",
       "minLength": 2,
       "maxLength": 256,
       "description": "The package author"
@@ -82,7 +82,7 @@
       "description": "The license page"
     },
     "Copyright": {
-      "type": [ "string", "null" ],
+      "type": "string",
       "minLength": 3,
       "maxLength": 512,
       "description": "The package copyright"
@@ -98,7 +98,7 @@
       "description": "The short package description"
     },
     "Description": {
-      "type": [ "string", "null" ],
+      "type": "string",
       "minLength": 3,
       "maxLength": 10000,
       "description": "The full package description"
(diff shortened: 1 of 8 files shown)


yao-msft:
We intentionally make optional fields accept null because during our internal yaml to json conversion(for schema validation), we convert empty values to null. This change "may" break our validation in the service pipeline.

Trenly:
> We intentionally make optional fields accept null because during our internal yaml to json conversion(for schema validation), we convert empty values to null. This change "may" break our validation in the service pipeline.
> 

Are you saying that if a field does not exist - I'll use the example of `Description` - If a defaultLocale manifest does not have the `Description` key, it gets set to null? I can understand if it is set to null when it is an empty value, but it shouldn't be null if the key is not present. Examples -

I would expect this to FAIL validation because description is empty - it truly is null. This currently PASSES `winget validate`
```yaml
ShortDescription: Yes, this is good
Description:
ManifestType: defaultLocale
```
I would expect this to PASS validation because description is commented out - it is not null, rather it is undefined, not present.
```yaml
ShortDescription: Yes, this is good
# Description:
ManifestType: defaultLocale
```

> And any change that makes schema more strict is a potential breaking change, we need to run this through all existing manifests to make sure we will not break service.

I agree that it does need to be tested; The extent and how to test, I'm not certain;

yao-msft:
> ```yaml
> ShortDescription: Yes, this is good
> Description:
> ManifestType: defaultLocale
> ```

This was actually allowed when we had the first manifest implementation. At that time, we think of the yaml manifest as a form for users to fill out. For optional fields, they can leave blank.

Also in addition below would fail if null was not allowed
> ```yaml
> ShortDescription: Yes, this is good
> Description: ""
> ManifestType: defaultLocale
> ```

And yes, we definitely need to have a full test internally before accepting this change.

yao-msft:
> I would argue that a case like this should fail anyways due to string length being less than the minimum defined; I believe it should be handled as a string of length 0 and not null in this specific case, as the quotations indicate typing, and null is a different type than string

Yes, I agree null and empty string are different types. I'm referring to the actual schema validation code. Since we have [this](https://github.com/microsoft/winget-cli/blob/93f8690d6c529c465ab0d255d7ea1b221025a3cb/src/AppInstallerCommonCore/Public/winget/Yaml.h#L98), they are treated as null during schema validation after yaml is converted to json. Changing the IsNull() may involve some more effort in testing. So in the end, allowing null is what we chose for backward compatibility back then.

yao-msft:
I think a run through of winget-pkgs can give us great confidence. The most reliable test would be someone from winget team take in the change to service and do a rebuild.
Winget-create uses the schemas so a brief run-through over there would be helpful too.
Winget-restsource spec strictly matches the manifest schema here, not sure if any 3rd party rest source will return null fields, I think maybe it's ok to create a discrepancy between the two but not desired.

PR #2068, opened 2022-04-05
Add support for in-proc Com state separation

yao-msft:
## Change
Added a process global PackageManagerSettings Com object to support better state separation for in-proc Com invocations.
Also make the package applicability check take in InstallOptions to give better results.

## Validation
Added back unpackaged tests run under system context.
Manually tried calling the PackageManagerSettings and see logs, etc are saved under state specific folder.

Next is the Com e2e test.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2068)

### DIFF
--- a/src/AppInstallerCommonCore/UserSettings.cpp
+++ b/src/AppInstallerCommonCore/UserSettings.cpp
@@ -72,25 +72,31 @@ namespace AppInstaller::Settings
             return convertedValue;
         }
 
+        std::optional<Json::Value> ParseSettingsContent(const std::string& content, std::string_view settingName, std::vector<UserSettings::Warning>& warnings)
+        {
+            Json::Value root;
+            Json::CharReaderBuilder builder;
+            const std::unique_ptr<Json::CharReader> reader(builder.newCharReader());
+            std::string error;
+
+            if (reader->parse(content.c_str(), content.c_str() + content.size(), &root, &error))
+            {
+                return root;
+            }
+
+            AICLI_LOG(Core, Error, << "Error parsing " << settingName << ": " << error);
+            warnings.emplace_back(StringResource::String::SettingsWarningParseError, settingName, error, false);
+
+            return {};
+        }
+
         std::optional<Json::Value> ParseFile(const StreamDefinition& setting, std::vector<UserSettings::Warning>& warnings)
         {
             auto stream = Stream{ setting }.Get();
             if (stream)
             {
-                Json::Value root;
-                Json::CharReaderBuilder builder;
-                const std::unique_ptr<Json::CharReader> reader(builder.newCharReader());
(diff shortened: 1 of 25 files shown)


src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -423,40 +429,67 @@ namespace AppInstaller::Settings
         return UserSettings::Instance();
     }
 
-    UserSettings::UserSettings() : m_type(UserSettingsType::Default)
+    bool TryInitializeCustomUserSettings(std::string content)
+    {
+        if (s_userSettingsInitialized || s_userSettingsInInitialization)
```
JohnMcPMS, line 434: This is probably the right approach for now, but we should have a test to make sure that our suggested method for initialization is working so that we don't load the settings earlier than that.

src/WindowsPackageManager/main.cpp
```diff
@@ -14,13 +14,14 @@
...
 
-// CreatorMap for out-of-proc com registration
+// CreatorMap for out-of-proc com registration and direct in-proc com class construction
 CoCreatableClassWrlCreatorMapInclude(PackageManager);
 CoCreatableClassWrlCreatorMapInclude(FindPackagesOptions);
 CoCreatableClassWrlCreatorMapInclude(CreateCompositePackageCatalogOptions);
 CoCreatableClassWrlCreatorMapInclude(InstallOptions);
 CoCreatableClassWrlCreatorMapInclude(UninstallOptions);
 CoCreatableClassWrlCreatorMapInclude(PackageMatchFilter);
+CoCreatableClassWrlCreatorMapInclude(PackageManagerSettings);
```
JohnMcPMS, line 24: I feel like the only thing stopping this from being used OOP is the fact that it isn't registered in our `AppxManifest.xml`.  I don't know if that makes it impossible to create, or just harder.

Not that it would be the end of the world or anything, but I think we would prefer that they not create it OOP if that doesn't require a lot of work.

JohnMcPMS (approved):

src/AppInstallerCommonCore/UserSettings.cpp
yao-msft: yes, I verified manually it's working as expected, and it should and will be part of Com e2e test.

src/WindowsPackageManager/main.cpp
yao-msft: After several hours of exploring, I think it requires some amount of work to do custom registration and maintaining the registrations. And it needs more thorough testing if we choose that route. So I'll leave it for now.

PR #2071, opened 2022-04-06
Add a system for testing correlation E2E

JohnMcPMS:
This change adds a few scripts, a test project, and one small product change, all around enabling end-to-end validation of our correlation between system artifacts and packages in external sources.

## Change
The product change is to expose publisher on the `PackageVersion` COM object.  This enables the test system to collect that data for further use, as well as allowing human verification of the correlation (if one is made).

The test project uses the COM API to first install a package, then attempts to check for correlation in two ways (directions):
1. When the remote package identity is known, the caller will usually look it up via that remote identity.  This path can enable additional information to be retrieved to make the correlation with.
2. When listing all of the local packages, the remote identity must be determined for each one. We can use only data known locally to make the correlation.

The primary script, `Test-CorrelationInSandbox.ps1`, is based off of the sandbox test script in winget-pkgs (thanks to many people).  It sets up the sandbox and initial script to run there for each package to test, then waits for a sentinel file to be created in the output location.  The results of all of running the test project exe, as well as the ARP differences and winget logs are put in that output location, then the sandbox is destroyed for the next package to run.

```
Test-CorrelationInSandbox.ps1
[[-PackageIdentifiers] <string[]>] :: A set of package ids to test
[[-Source] <string>] :: The name of the source that the packages are from, ex. "winget"
[-ExePath <string>] :: Path to the test exe; defaults to the Release output location
[-UseDev] :: Switch to use the local dev winget build
[-DevPackagePath <string>] :: Path to the local dev *Release* winget build; defaults to the normal location
[-ResultsPath <string>] :: Path to output the results to; defaults to a temp directory
[-RegFileDirectory <string>] :: Path to a directory containing .reg files to insert before the test, creating noise for correlation
```

Once testing is done, `Process-CorrelationResults.ps1` will take all of the results JSON files and put them into `results.csv` in the directory.  If any tests failed to run, they will be in `failed.csv`.  There are correlation columns in the CSV that can be averaged in Excel to get the correlation percentage.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2071)


tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
```diff
@@ -0,0 +1,342 @@
...
+{
+  $ResultsPath = Join-Path ([System.IO.Path]::GetTempPath()) (New-Guid)
+}
+
+if (Test-Path $ResultsPath)
+{
+  Remove-Item -Recurse $ResultsPath -Force
+}
+
+New-Item -ItemType Directory $ResultsPath > $nul
```
lechacon, line 88: Nit 1: Did you mean `$null`?
Nit 2: You are using both `> $nul` and `| Out-Null`. Should it be consistent?

tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
```diff
@@ -0,0 +1,342 @@
...
+  $devVCLibsFileName = "Microsoft.VCLibs.x64.14.00.Desktop.appx"
+  $devVCLibsPath = Join-Path ${env:ProgramFiles(x86)} "Microsoft SDKs\Windows Kits\10\ExtensionSDKs\Microsoft.VCLibs.Desktop\14.0\Appx\Retail\x64"
+  $devVCLibsPath = Join-Path $devVCLibsPath $devVCLibsFileName
+
+  Copy-Item -Path $devVCLibsPath -Destination $tempFolder -Force
+
+  $devVCLibsPathInSandbox = Join-Path -Path $desktopInSandbox -ChildPath (Join-Path -Path $tempFolderName -ChildPath $devVCLibsFileName)
+}
+else
+{
```
lechacon, line 172: Is something missing here?

lechacon, line 172: Is something missing here?

tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
```diff
@@ -0,0 +1,342 @@
...
+
+    $sandboxTestWsbFileName = 'SandboxTest.wsb'
+    $sandboxTestWsbFile = Join-Path -Path $tempFolder -ChildPath $sandboxTestWsbFileName
+    $sandboxTestWsbContent | Out-File $sandboxTestWsbFile -Force
+
+    Write-Host @"
+--> Starting Windows Sandbox, and:
+    - Mounting the following directories:
+        - $tempFolder as read-only
+        - $outPath as read-write
```
lechacon, line 320: Nit: Missing Dev Package and Reg File folders

lechacon, line 320: Nit: Missing Dev Package and Reg File folders

tools/CorrelationTestbed/InstallAndCheckCorrelation/InstallAndCheckCorrelation/InstallAndCheckCorrelation.cpp
```diff
@@ -0,0 +1,552 @@
...
+        auto installResult = packageManager.InstallPackageAsync(package, installOptions).get();
+
+        if (installResult.Status() != InstallResultStatus::Ok)
+        {
+            hr = installResult.ExtendedErrorCode();
+            error = "Error installing package";
+            break;
+        }
+
+    } while(0); }
```
lechacon, line 337: I don't have any actual comment, but using the do-while + break like this just doesn't feel right to me..

lechacon, line 337: I don't have any actual suggestion, but using the do-while + break like this just doesn't feel right to me..

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -270,6 +270,8 @@ namespace Microsoft.Management.Deployment
         {
             /// Checks if this package version has at least one applicable installer.
             Boolean HasApplicableInstaller { get; };
+
+            String Publisher { get; };
```
lechacon, line 274: Nit: Add a comment. Seems like most other things here are documented.

lechacon, line 274: Nit: Add a comment. Seems like most other things here are documented.

lechacon, line 274: Do we need to do any review for these API changes?

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -134,6 +134,7 @@ namespace AppInstaller::Repository
         RelativePath,
         // Returned in hexadecimal format
         ManifestSHA256Hash,
+        Publisher,
```
lechacon, line 137: Looks like something I'll need to use in my change for matching heuristics. 🤔

lechacon, line 137: Looks like something I'll need to use in my change for matching heuristics. 🤔

lechacon (commented):
Can you add a README or something like that for how to use the scripts?

lechacon:
Why did all my comments show up twice...

tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
```diff
@@ -0,0 +1,342 @@
...
+    - Comparing ARP Entries
+"@
+
+    Write-Host
+
+    WindowsSandbox $SandboxTestWsbFile
+
+    $outputFileBlockerPath = Join-Path $outPath "done.txt"
+
+    while (-not (Test-Path $outputFileBlockerPath))
```
lechacon, line 332: I would add a timeout here. When I tried doing something similar, I saw many cases where the app would still require interaction and it would block the script until I went to click something.

I think that shouldn't happen, but it seems the msstore source doesn't always gives us the right arguments for the installers :(

tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
```diff
@@ -0,0 +1,342 @@
...
+
+    $sandboxTestWsbContent = @"
+<Configuration>
+    <MappedFolders>
+    <MappedFolder>
+        <HostFolder>$tempFolder</HostFolder>
+        <ReadOnly>true</ReadOnly>
+    </MappedFolder>
+    <MappedFolder>
+        <HostFolder>$ExePath</HostFolder>
```
lechacon, line 296: Seems like these need to be expanded to absolute paths for it to work.

tools/CorrelationTestbed/InSandboxScript.ps1
```diff
@@ -0,0 +1,111 @@
...
+
+--> Comparing ARP Entries
+"@
+
+$arpCompared = (Compare-Object (Get-ARPTable) $originalARP -Property DisplayName,DisplayVersion,Publisher,ProductCode)
+$arpCompared | Select-Object -Property * -ExcludeProperty SideIndicator | Format-Table
+
+$arpCompared | Select-Object -Property * -ExcludeProperty SideIndicator | Format-Table | Out-File (Join-Path $OutputPath "ARPCompare.txt")
+
+"Done" | Out-File (Join-Path $OutputPath "done.txt")
```
lechacon, line 111: This won't happen if there is any error during the script, so the main one will just hang waiting for this. I don't know if that is good or bad

tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
```diff
@@ -0,0 +1,342 @@
...
+      $sandbox | Wait-Process -Timeout 30
+
+      Start-Sleep 2
+
+      Write-Host
+    }
+    Remove-Variable sandbox
+}
+
+Close-WindowsSandbox
```
lechacon, line 107: Can you do this before line 85 where you remove the results folder?
If the script fails and the sandbox stays open, it will hold on to the results files and block removing the folder so we have to manually close the sandbox before running again.

tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
JohnMcPMS: No, I just didn't indent the entire existing block at the time.

tools/CorrelationTestbed/InstallAndCheckCorrelation/InstallAndCheckCorrelation/InstallAndCheckCorrelation.cpp
JohnMcPMS: No doubt, it is a lazy, monolithic function.  This or an immediately invoked lambda are the only ways to be that lazy 😄 

I've changed it to not use lazy control flow, but still be quite lazy on the values.

src/Microsoft.Management.Deployment/PackageManager.idl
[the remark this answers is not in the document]
JohnMcPMS: Not for these simple ones, no.

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
JohnMcPMS: Yes, although it just makes getting the value easier since you don't have to get all of the metadata, so it isn't really critical to move to it.

lechacon (approved):
:shipit:

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -16,7 +16,6 @@
   <Dependencies>
     <!-- Minimum supported version is 1809 (October 2018 Update, aka RS5) -->
     <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.17763.0" MaxVersionTested="10.0.19033.0" />
-    <PackageDependency Name="Microsoft.VCLibs.140.00.UWPDesktop" MinVersion="14.0.25426.0" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" />
```
yao-msft, line 19: ><PackageDependency Name="Microsoft.VCLibs.140.00.UWPDesktop" MinVersion="14.0.25426.0" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" /> [](http://example.com/codeflow?start=4&length=182)

Is this removal intentional?

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
```diff
@@ -48,7 +48,8 @@ namespace AppInstaller::Repository::Microsoft
                     return LocIndString{ GetReferenceSource()->GetDetails().Name };
                 default:
                     // Values coming from the index will always be localized/independent.
-                    return LocIndString{ GetReferenceSource()->GetIndex().GetPropertyByManifestId(m_manifestId, property).value() };
+                    std::optional<std::string> optValue = GetReferenceSource()->GetIndex().GetPropertyByManifestId(m_manifestId, property);
+                    return LocIndString{ optValue ? optValue.value() :std::string{} };
```
yao-msft, line 52: >:st [](http://example.com/codeflow?start=69&length=3)

nit: spacing

yao-msft (approved):
:shipit:

src/AppInstallerCLIPackage/Package.appxmanifest
JohnMcPMS: Yes, it is being injected by VS anyway.  This one was leading to debug needing the release package (as VS left this and injected the debug on as well).

PR #2087, opened 2022-04-12
Fix issue with correlation chaining from the SQLite index

JohnMcPMS:
## Change
The composite source correlation was using the name + publisher values in a very optimistic way, which was leading to the potential for some values to get lost.  This change creates a search on the cartesian product of the values, which is more expensive but also more correct.

Also includes another option for the correlation test exe to skip the install for testing/debugging purposes.  Also again, updates the winmd because I think that the IID for WinRT is based on a hash of the interface definition, and the change to the contract 4 interface broke the existing file.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2087)


src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -607,15 +607,21 @@ namespace AppInstaller::Repository
                 IPackageVersion* installedVersion,
                 PackageData& data)
             {
+                // Unfortunately the names and publishers are unique and not tied to each other strictly, so we need
+                // to go broad on the matches. Future work can hopefully make name and publisher work closer to the
```
yao-msft, line 611: >work closer to the [](http://example.com/codeflow?start=97&length=18)

typo?

yao-msft (approved):
:shipit:

PR #2107, opened 2022-04-20
Enable long aliases

Trenly:

This change allows for argument aliases to be more than 1 character

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2107)

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -28,7 +28,7 @@ namespace AppInstaller::CLI
         case Args::Type::Tag:
             return Argument{ "tag", NoAlias, Args::Type::Tag, Resource::String::TagArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         case Args::Type::Command:
-            return Argument{ "command", NoAlias, Args::Type::Command, Resource::String::CommandArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
+            return Argument{ "command", NoAlias, "cmd", Args::Type::Command, Resource::String::CommandArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         case Args::Type::Source:
             return Argument{ "source", 's', Args::Type::Source, Resource::String::SourceArgumentDescription, ArgumentType::Standard };
         case Args::Type::DependencySource:
@@ -54,9 +54,9 @@ namespace AppInstaller::CLI
         case Args::Type::InstallLocation:
             return Argument{ "location", 'l', Args::Type::InstallLocation, Resource::String::LocationArgumentDescription, ArgumentType::Standard };
         case Args::Type::HashOverride:
-            return Argument{ "force", Argument::NoAlias, Args::Type::HashOverride, Resource::String::InstallForceArgumentDescription, ArgumentType::Flag, Settings::TogglePolicy::Policy::HashOverride };
+            return Argument{ "force", NoAlias, Args::Type::HashOverride, Resource::String::InstallForceArgumentDescription, ArgumentType::Flag, Settings::TogglePolicy::Policy::HashOverride };
         case Args::Type::AcceptPackageAgreements:
-            return Argument{ "accept-package-agreements", Argument::NoAlias, Args::Type::AcceptPackageAgreements, Resource::String::AcceptPackageAgreementsArgumentDescription, ArgumentType::Flag };
+            return Argument{ "accept-package-agreements", NoAlias, Args::Type::AcceptPackageAgreements, Resource::String::AcceptPackageAgreementsArgumentDescription, ArgumentType::Flag };
         case Args::Type::HashFile:
             return Argument{ "file", 'f', Args::Type::HashFile, Resource::String::FileArgumentDescription, ArgumentType::Positional, true };
         case Args::Type::Msix:
@@ -80,7 +80,7 @@ namespace AppInstaller::CLI
         case Args::Type::RetroStyle:
             return Argument{ "retro", NoAlias, Args::Type::RetroStyle, Resource::String::RetroArgumentDescription, ArgumentType::Flag, Argument::Visibility::Hidden };
         case Args::Type::VerboseLogs:
-            return Argument{ "verbose-logs", NoAlias, Args::Type::VerboseLogs, Resource::String::VerboseLogsArgumentDescription, ArgumentType::Flag };
+            return Argument{ "verbose-logs", NoAlias, "verbose", Args::Type::VerboseLogs, Resource::String::VerboseLogsArgumentDescription, ArgumentType::Flag};
         case Args::Type::CustomHeader:
             return Argument{ "header", NoAlias, Args::Type::CustomHeader, Resource::String::HeaderArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         case Args::Type::AcceptSourceAgreements:
@@ -96,7 +96,7 @@ namespace AppInstaller::CLI
         case Args::Type::Wait:
             return Argument{ "wait", NoAlias, Args::Type::Wait, Resource::String::WaitArgumentDescription, ArgumentType::Flag, false };
         case Args::Type::ProductCode:
-            return Argument{ "product-code", NoAlias, Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
+            return Argument{ "product-code", NoAlias, "pid", Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
         default:
             THROW_HR(E_UNEXPECTED);
         }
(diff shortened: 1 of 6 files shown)


Trenly:
> Supporting multi-character aliases removes, or at least complicates greatly, this feature:

I did not realize that was a feature. You are correct that I did not consider it here.

> What is the goal in having longer aliases?

The goal here is to allow overall extended functionality. I had simply used verbose as an example. There are use cases where a single letter alias doesn’t make sense, or would conflict with an alias of an already existing argument. This would allow for a wider range of aliases to be available.

A secondary effect here is to make it easier to use in general. Looking from the perspective of a new user, and just as an example, I wouldn’t know that `--verbose-logs` was a thing, but I would expect (based on powershell and other conventions) that `-verbose` would be a standard behavior.

Trenly:
> But I wouldn't be opposed to an arcane alias for it seeing as it has none today. Something like `-+` makes some sense, and is not likely that anyone else would want to use it.

I've updated the PR so that any aliases longer than 1 character require the `-+` syntax

The other thought I had was that perhaps we could still use `--` for long aliases, since at that point it would just be a short argument rather than a long alias;

JohnMcPMS:
> I've updated the PR so that any aliases longer than 1 character require the `-+` syntax

I meant that the alias for the verbose logs option would be the character `+`, not that a new syntax should be created.

> The other thought I had was that perhaps we could still use `--` for long aliases, since at that point it would just be a short argument rather than a long alias;

I am not opposed to alternate names for arguments.  You don't even have to write new code for that necessarily (although it is probably the better solution).

src/AppInstallerCLICore/Argument.h
```diff
@@ -14,6 +14,8 @@
 
 #define APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR       '-'
 #define APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_STRING     "-"
+#define APPINSTALLER_CLI_LONG_ALIAS_IDENTIFIER_CHAR     '+'
```
JohnMcPMS, line 17: Not a fan of adding a new syntax here; would prefer the alternate name strategy.

JohnMcPMS (changes requested):
Sorry about the delay in reviewing.  I do think that the easiest way forward is allowing for alternate names for arguments while leaving the alias character stuff alone.

Trenly:
> I meant that the alias for the verbose logs option would be the character `+`, not that a new syntax should be created.
> 
Ahhh, I misunderstood. Removed.

> I am not opposed to alternate names for arguments. You don't even have to write new code for that necessarily (although it is probably the better solution).
>
Made this the functionality. Aliases longer than 1 character now are functionally equivalent to alternate names

src/AppInstallerCLICore/Argument.h
```diff
@@ -47,24 +47,24 @@ namespace AppInstaller::CLI
         };
 
         // Defines an argument with no alias.
-        constexpr static char NoAlias = '\0';
+        constexpr static char const* NoAlias = "\0";
```
JohnMcPMS, line 50:
```suggestion
        constexpr static std::string_view NoAlias;
```
Can just be empty string.

src/AppInstallerCLICore/Argument.h
```diff
@@ -47,24 +47,24 @@ namespace AppInstaller::CLI
         };
 
         // Defines an argument with no alias.
-        constexpr static char NoAlias = '\0';
+        constexpr static char const* NoAlias = "\0";
 
-        Argument(std::string_view name, char alias, Execution::Args::Type execArgType, Resource::StringId desc) :
+        Argument(std::string_view name, char const* alias, Execution::Args::Type execArgType, Resource::StringId desc) :
```
JohnMcPMS, line 52:
```suggestion
        Argument(std::string_view name, std::string_view alias, Execution::Args::Type execArgType, Resource::StringId desc) :
```
Although mildly more dangerous than `char*`, these should all be `std::string_view`, as should the storage and return value from the getter.

Also consider keeping `char Alias() const` and adding `std::string_view AlternateName() const`.  Then `Alias` returns `NoAlias` if the stored string view is not exactly 1 character, or the single character if it is 1.  `AlternateName` returns a non-empty view only if the stored view is > 1 character.  That way, the consuming code doesn't need to understand how we decide between the two options, it just needs to know how to tell if either of the values is valid or not.

src/AppInstallerCLICore/Command.cpp
```diff
@@ -226,7 +230,14 @@ namespace AppInstaller::CLI
                 std::ostringstream strstr;
                 if (arg.Alias() != Argument::NoAlias)
```
JohnMcPMS, line 231: Mildly dangerous check here if you don't change to `std::string_view`.  Comparing pointers only works for the literals.

src/AppInstallerCLICore/Command.cpp
```diff
@@ -492,7 +503,7 @@ namespace AppInstaller::CLI
         else if (currArg[1] != APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR)
         {
             // Parse the single character alias argument
-            char currChar = currArg[1];
+            std::string currChar(1, currArg.at(1));
```
JohnMcPMS, line 506: Is there any value to this change now?

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Argument.h
Trenly: I've added the alternate name as a member variable.

For the sake of not modifying every single Argument command again, I've created additional overloads for the constructor to handle these. If this is not the appropriate way to handle this, let me know so I can fix it accordingly

Thanks for your guidance and patience 😄

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -96,7 +96,7 @@ namespace AppInstaller::CLI
         case Args::Type::Wait:
             return Argument{ "wait", NoAlias, Args::Type::Wait, Resource::String::WaitArgumentDescription, ArgumentType::Flag, false };
         case Args::Type::ProductCode:
-            return Argument{ "product-code", NoAlias, Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
+            return Argument{ "product-code", NoAlias, "pid", Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
```
JohnMcPMS, line 99: `pid` usually means "Process ID".  I don't see how it relates to product code.

src/AppInstallerCLICore/Argument.h
```diff
@@ -49,23 +49,44 @@ namespace AppInstaller::CLI
         // Defines an argument with no alias.
         constexpr static char NoAlias = '\0';
 
+        // Defines an argument with no alternate name
+        constexpr static std::string_view NoAlternateName = "";
+
         Argument(std::string_view name, char alias, Execution::Args::Type execArgType, Resource::StringId desc) :
-            m_name(name), m_alias(alias), m_execArgType(execArgType), m_desc(std::move(desc)) {}
+            m_name(name), m_alias(alias), m_alternateName(NoAlternateName), m_execArgType(execArgType), m_desc(std::move(desc)) {}
```
JohnMcPMS, line 56: nit: The default constructed state of a `std::string_view` is already an empty string.

JohnMcPMS (changes requested):
The test that ensures that there are no commands with arguments that collide should be updated to check against `AlternateName` as well.

src/AppInstallerCLICore/Argument.cpp
Trenly: `pid` has multiple meanings. In the terms of a running service, it does typically mean "Process ID". I've seen it used in the application space to refer to "Product ID"

Is there a better alternate name to use here, or should I just remove this one?

JohnMcPMS: I find it to be confusing, I would remove it as I don't foresee this being a commonly used argument in need of shortening.

src/AppInstallerCLITests/Command.cpp
```diff
@@ -69,6 +81,7 @@ void EnsureCommandConsistency(const Command& command)
     auto args = command.GetArguments();
     Argument::GetCommon(args);
     EnsureStringsAreLowercaseAndNoCollisions(command.FullName() + " argument names", args, GetArgumentName);
+    EnsureStringsAreLowercaseAndNoCollisions(command.FullName() + " argument alternate name", args, GetArgumentAlternateName);
```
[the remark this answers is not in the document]
JohnMcPMS: Each one of the current calls to `EnsureStringsAreLowercaseAndNoCollisions` will detect collisions within their respective space, `Name` and `AlternateName`.  They do that by simply creating a set in memory and looking for duplicates as they go.  I'm suggesting that we can pass the set in and thus preserve it across the two calls:
```
std::unordered_set<std::string> allNames;
EnsureStringsAreLowercaseAndNoCollisions(command.FullName() + " argument names", args, GetArgumentName, allNames);
EnsureStringsAreLowercaseAndNoCollisions(command.FullName() + " argument alternate name", args, GetArgumentAlternateName, allNames);
```

JohnMcPMS (approved):

PR #2119, opened 2022-04-25
Remove ARP matching for single ARP change, and consider publisher+name for matching

lechacon:
* For ARP matching we had a rule that would consider something a match if it was the only ARP change, but that is likely to give us false matches. This change removes that rule.
* For the string matching confidence algorithm, we would only consider matching package name with ARP name, and package publisher with ARP publisher, but sometimes the publisher would be included in the name and that would decrease confidence. For example, for package name = "Microsoft PowerToys"  and ARP name = "PowerToys". This change makes the string matching also consider the string publisher+name.
* Also made the matching requirements for the unit tests stricter

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2119)

### DIFF
--- a/src/AppInstallerCLITests/Correlation.cpp
+++ b/src/AppInstallerCLITests/Correlation.cpp
@@ -225,10 +225,10 @@ DataSet GetDataSet_NoNoise()
     dataSet.TestCases = LoadTestData();
 
     // Arbitrary values. We should refine them as the algorithm gets better.
-    dataSet.RequiredTrueMatchRatio = 0.5;
-    dataSet.RequiredFalseMatchRatio = 0.1;
+    dataSet.RequiredTrueMatchRatio = 0.7;
+    dataSet.RequiredFalseMatchRatio = 0.05;
     dataSet.RequiredTrueMismatchRatio = 0; // There are no expected mismatches in this data set
-    dataSet.RequiredFalseMismatchRatio = 0.5;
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLITests/Correlation.cpp
```diff
@@ -242,10 +242,10 @@ DataSet GetDataSet_WithNoise()
     dataSet.TestCases = std::move(baseTestCases);
 
     // Arbitrary values. We should refine them as the algorithm gets better.
-    dataSet.RequiredTrueMatchRatio = 0.5;
-    dataSet.RequiredFalseMatchRatio = 0.1;
+    dataSet.RequiredTrueMatchRatio = 0.7;
+    dataSet.RequiredFalseMatchRatio = 0.05;
```
JohnMcPMS, line 246: I would really like false positives to be at 0.  If the current algorithm is not at 0 in these unit tests, we should reconsider.

JohnMcPMS (approved):

lechacon:
One of the tests for `winget list` was expecting the output to contain the product code, but now that we can match it the output has the Id :)

src/AppInstallerCLITests/Correlation.cpp
lechacon: I'll send another PR once I figure out how to reduce the false positives.

Issue #1249, opened 2021-07-01
Disallow empty query on search


Trenly:
# Description of the new feature/enhancement
When a user runs `winget search` with no parameters, no switches, or an empty search string, they receive a list of many, many packages. Running a search with no query is a very rare and infrequent use case for a package manager. This would help a user understand when the search is not going to give meaningful results

# Proposed technical implementation details (optional)
- If there are no switches used and no query string, show the command usage (alias of `winget search -?`)
- If the query string is `""` throw an invalid argument error and show the command usage
- Treat a combination of a switch and parameter as a valid query

Require searches to have a query that will return meaningful results
```powershell
# Expected Results:
winget search  # Returns the command usage
winget search ""  # Returns an invalid argument error and shows the command usage
winget search Test # Returns all packages matching the query "Test"
winget search --moniker Test # Returns all packages matching the moniker "Test"
winget search Test --tag system # Returns all packages matching the query "Test" with a tag of "system"
winget search --moniker "" # Returns an invalid argument error and shows the command usage
winget search --moniker # Returns argument value required error
winget search -n 5 # Returns an invalid argument error - query string is required
```

rdeforest:
I disagree. This proposal assumes it can read the user's mind and that they didn't do what they meant to do. Here's a use-case for an empty search:

```console
PS C:\Users\Robert de Forest> $i=0; winget search | foreach-object { $i++ }; echo $i
2699
PS C:\Users\Robert de Forest>
```

You could add a switch to search to request a count of all packages, but that feature already exists using the construction above. This proposal solves a problem which does not exist.

PR #2125, opened 2022-04-26
Require selection argument on install/show/search/uninstall

### LINKED ISSUE #1131: Winget install/show without arguments should not produce a list of all packages

JohnMcPMS:
Fixes #1131 

## Change
For `install`, `show`, `search`, and `uninstall` commands, require that one of the selection arguments be provided.  These are (where appropriate per command):
```
-q, --query [default positional argument]
-m, --manifest
--id
--name
--moniker
--tag
--command
```
I intentionally did not provide a way to override this behavior (like an `--all` for `search`), because I don't see a good use case for it.  If there is one, please explain it below.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2125)

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 #include "pch.h"
 #include "Argument.h"
+#include "Command.h"
 #include "Resources.h"
 #include <winget/UserSettings.h>
 
@@ -106,6 +107,19 @@ namespace AppInstaller::CLI
         args.push_back(ForType(Args::Type::RainbowStyle));
         args.push_back(ForType(Args::Type::RetroStyle));
         args.push_back(ForType(Args::Type::VerboseLogs));
+    }
+
+    void Argument::ValidatePackageSelectionArgumentSupplied(const Execution::Args& args)
+    {
+        for (Args::Type type : { Args::Type::Query, Args::Type::Manifest, Args::Type::Id, Args::Type::Name, Args::Type::Moniker, Args::Type::Tag, Args::Type::Command })
+        {
+            if (args.Contains(type))
+            {
+                return;
+            }
+        }
+
+        throw CommandException(Resource::String::NoPackageSelectionArgumentProvided);
     }
 
     Argument::Visibility Argument::GetVisibility() const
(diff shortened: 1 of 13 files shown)


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -108,6 +109,19 @@ namespace AppInstaller::CLI
         args.push_back(ForType(Args::Type::VerboseLogs));
     }
 
+    void Argument::ValidatePackageSelectionArgumentSupplied(const Execution::Args& args)
+    {
+        for (Args::Type type : { Args::Type::Query, Args::Type::Manifest, Args::Type::Id, Args::Type::Name, Args::Type::Moniker, Args::Type::Tag, Args::Type::Command })
+        {
+            if (args.Contains(type))
+            {
+                return;
```
Trenly, line 118: Would it be possible to ensure the query is also not an empty string?
https://github.com/microsoft/winget-cli/issues/1249

JohnMcPMS: It would be possible, but allowing an empty string does allow for an effective `--all` being `""`.  The current change prevents accidental "everything" searches, which I feel was the primary value.  Preventing intentional searching for everything seems like it would be due to the current performance limitations rather than due to protecting users.

If the user types in `winget search ""`, it could be an accident, but it is more likely that it is intentional than `winget search`.  I could see scripts constructing the command line resulting in more frequent empty strings as well, but if the script author wants to block that, they can.

Without any other reasons, I'm not inclined to block the empty strings.

Trenly: I think those are fair points, and are things I hadn't considered. After thinking more, I do agree that simply requiring an argument solves the potential issues here

yao-msft (approved):
:shipit:

Issue #152, opened 2020-05-15
Support installers that require elevated priveleges


denelon:
As a user I want graceful behavior when I try to install an application that requires admin privileges and I'm in user mode.

Notes:
Some applications don't gracefully behave when installation is initiated in user mode and admin mode is required.
We should consider an attribute in the Manifest (like SupportedMode: Admin, RequiresAdmin: true, or something like that)
When the value is specified, the client should not try to perform the install if it's not elevated and the user should be informed.

There should be some automation added when we validate the manifest after PR to check and see if the attribute should have been added.

The client should be able to validate a manifest that has the attribute.

# Proposed technical implementation details (optional)

megamorf:
See:
* #330
* #281
* #271
* #218

Trenly:
@denelon - Resolved by #2126 ?

Issue #1538, opened 2021-10-01
Winget not asking for privilege escalation for some packages


FurtadoPires:
# Brief description of your issue

Some applications like Libreoffice can't be installed with windows terminal in user mod, because the terminal doesn't ask to escalate privilege during the install process (crashing with the code 1603)

If it's executed using Powershell as admin, the install process silently finishes in the background but the terminal window remains stuck with the initiating install message.
# Steps to reproduce

`winget upgrade LibreOffice.LibreOffice`

# Environment

```none
[winget --info]
Windows: Windows.Desktop v10.0.19043.1237
Pacote: Microsoft.DesktopAppInstaller v1.16.12653.0

Registros: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

Links
-------------------------------------------------------------------------------
Política de Privacidade https://aka.ms/winget-privacy
Contrato de Licença     https://aka.ms/winget-license
Avisos de Terceiros     https://aka.ms/winget-3rdPartyNotice
Home page               https://aka.ms/winget
Termos da Windows Store https://www.microsoft.com/en-us/storedocs/terms-of-sale
Windows Package Manager version
Windows: Windows.Desktop version
Package: Microsoft.DesktopAppInstaller version

denelon:
#1243 includes #1284 intended to address this scenario. We still have some work to do in the client for this, but the schema work has been implemented, and the support for the v1.1 manifest schema is in progress on the validation pipelines.

denelon:
We've added the key to the schema to indicate when a package requires or prohibits elevation. This work item will be used to track the work to provide the information to the user. It will give an error for user mode if the package requires elevation, and it will give an error for admin mode if the package prohibits elevation.

Issue #271, opened 2020-05-21
UAC prompt when installing application


suhailmalik07:
It's ugly when you install a package using the command line and after that, a window prompt appears for a mouse click. Instead of this, something like `sudo` which ask for password before downloading the package will be beautiful and look consistent.

Yeah, you're right we can open the terminal `as an administrator` but when you working only on CLI it breaks the flow. It doesn't feel nice that you opened another terminal window just for installing a program.

TheRndCoder:
In Windows, the closest thing to 'sudo' is Runas: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc771525(v=ws.11)

But the only problem is that the command input is larger compared to just 'sudo winget install package', using runas you would have to do: 'runas /user:<localmachinename>\administrator winget install package' (haven't tested that yet)

byteduck:
Could winget elevate itself and then install the packages?

dmex:
> Could winget elevate itself and then install the packages?

You can run the winget command from an elevated command prompt and won't see any UAC prompts when installing packages.

jessehouwing:
start the command prompt elevated using:

    winkey, 'cmd', ctl-shift-enter'

suhailmalik07:
Yeah, you're right we can open the terminal `as an administrator` but when you working only on CLI it breaks the flow. It doesn't feel nice that you opened another terminal window just for installing a program.

suhailmalik07:
> In Windows, the closest thing to 'sudo' is Runas: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc771525(v=ws.11)
> 
> But the only problem is that the command input is larger compared to just 'sudo winget install package', using runas you would have to do: 'runas /user:\administrator winget install package' (haven't tested that yet)

This can do the work if implemented with windows 10 and winget. Never thought MSFT already had something similar to `sudo`.

dhenze-dwz:
> start the command prompt elevated using:
> 
> ```
> winkey, 'cmd', ctl-shift-enter'
> ```

For some reason I am getting UAC prompts despite doing this. Any ideas?

JoBrad:
Opening a separate elevated prompt not only breaks the user workflow, but also introduces other follow-up steps, too. The user potentially needs to refresh the environment in the other terminal instance, or even destroy it and start over, depending on the updates made.

[...]

PR #2126, opened 2022-04-26
Act on elevation requirements in majority cases

### LINKED ISSUE #1537: Add messaging to support package specific elevation requirements

When the client is not in the proper mode (user) vs. (administrator) and the package has conflicting requirements, or the installer performs it's own elevation, the user should be informed.

Related to:
* #1284
* microsoft/winget-pkgs/issues/29148

JohnMcPMS:
Fixes #1537, #1538

## Change
When `ElevationRequirement` is set to `elevationProhibited` is set and winget is run elevated, block the installation with an error.

When `elevatesSelf` or `elevationRequired` is set and winget is not run elevated, inform the user that an elevation prompt is expected.  Also in the case of `elevationRequired`, use the `runas` verb on ShellExecute to run the installer elevated.  This can be used on installers that will fail if they are not run elevated.

The message is a warning (yellow text) and is the last line below:
```
Starting package install...
The installer will request to run as administrator, expect a prompt.
```

## Validation
Manually validated each of the 3 elevation requirement options, as automated testing of elevation is not possible on the build server.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2126)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1287,4 +1287,10 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="WaitArgumentDescription" xml:space="preserve">
     <value>Prompts the user to press any key before exiting</value>
   </data>
+  <data name="InstallerElevationExpected" xml:space="preserve">
+    <value>The installer will request to run elevated, expect a prompt.</value>
```
yao-msft, line 1291: >elevated [](http://example.com/codeflow?start=45&length=8)

nit: "as administrator" since this is user facing string?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -116,6 +116,14 @@ namespace AppInstaller::CLI::Workflow
             context.Reporter.Error() << Resource::String::NoApplicableInstallers << std::endl;
             AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_NO_APPLICABLE_INSTALLER);
         }
+
+        // This installer cannot be run elevated, but we are running elevated.
+        // Implementation of de-elevation is complex; simply block for now.
+        if (installer->ElevationRequirement == ElevationRequirementEnum::ElevationProhibited && Runtime::IsRunningAsAdmin())
```
yao-msft, line 122: >if (installer->ElevationRequirement == ElevationRequirementEnum::ElevationProhibited && Runtime::IsRunningAsAdmin()) [](http://example.com/codeflow?start=8&length=116)

I'm sort of mixed. Did you think of moving to installer selection in case there're other applicable installers? Or it's preferred to tell the user to run non-elevated again since this is what we think is the best installer.

Trenly: I think it may be best to have it in the installer, especially given that there is a [feature request](https://github.com/microsoft/winget-cli/issues/1898) to filter by elevation requirement. However, if there are no applicable installers based solely on the fact that the user is running in an elevated terminal, then it still should say elevation prohibited rather than no applicable installer found

JohnMcPMS: I don't think that installer selection is appropriate, we don't even have user intent without #1898.  I don't consider "is winget running as admin or not" user intent here, because I want to be able to run my Terminal as user and elevate installers as needed.  #1898 should be implemented in installer selection (as well as search) though, because we would have user intent.

JohnMcPMS:
Side note: I broke VS Code on my machine due to forcing it to install elevated.  Be safe out there...

src/AppInstallerCLICore/Workflows/MsiInstallFlow.cpp
```diff
@@ -32,10 +32,17 @@ namespace AppInstaller::CLI::Workflow
...
         context.Reporter.Info() << Resource::String::InstallFlowStartingPackageInstall << std::endl;
 
+        const auto& installer = context.Get<Execution::Data::Installer>();
         const std::filesystem::path& installerPath = context.Get<Execution::Data::InstallerPath>();
 
         Msi::MsiParsedArguments parsedArgs = Msi::ParseMSIArguments(context.Get<Execution::Data::InstallerArgs>());
 
+        // Inform of elevation requirements
+        if (!Runtime::IsRunningAsAdmin() && installer->ElevationRequirement == Manifest::ElevationRequirementEnum::ElevatesSelf)
+         {
```
yao-msft, line 42: >  [](http://example.com/codeflow?start=8&length=1)

nit: alignment

yao-msft (approved):
:shipit:

Trenly:
Does this also resolve https://github.com/microsoft/winget-cli/issues/152?

JohnMcPMS:
> Does this also resolve #152 and #271 ?

@denelon to say if it resolves #152.  #271 is more of a general Windows issue; we can't magically bypass the UAC prompt.  I don't know how one could even secure a `sudo` command in Windows if it doesn't trigger the UAC prompt...

Trenly:
> > Does this also resolve #152 and #271 ?
> 
> @denelon to say if it resolves #152. #271 is more of a general Windows issue; we can't magically bypass the UAC prompt. I don't know how one could even secure a `sudo` command in Windows if it doesn't trigger the UAC prompt...

I thought that using runas would suppress the prompt, as suggested in the issue; But, perhaps I read it incorrectly or was mistaken

Masamune3210:
Nope, not having a way to bypass UAC is very much an intentional part of the design

PR #2137, opened 2022-05-03
Fix list/upgrade table for packages with multiple ARP entries.

jedieaston:

-----

Hooray! Packages with multiple ARP entries are being matched! However, there's a problem, which is that with the current way `ReportListResult()` works, the package's ARP name(s) are overwritten with the one from the manifest, which puts duplicate looking results in the list:

```
PS C:\Users\WDAGUtilityAccount> wingetdev list Mozilla
Name            Id              Version Available Source
--------------------------------------------------------
Mozilla Firefox Mozilla.Firefox 99.0.1  100.0     winget
Mozilla Firefox Mozilla.Firefox 99.0.1  100.0     winget
PS C:\Users\WDAGUtilityAccount> # Two Firefoxes at the same time? Highly dubious!
```

<s>This PR changes `ReportListResult()` to only output a single row for every package ID by default (which means we only show one entry in the list per "winget package", even if multiple pieces of software were installed via that package). It also adds a new argument to `winget list`, `--all`, which will show all of the software installed (but now using the original name in the ARP table instead of the name of the package in the repo, to cause less confusion). Here's what it looks like:

```
PS C:\Users\easton> wingetdev list Mozilla
Name            Id              Version Available Source
--------------------------------------------------------
Mozilla Firefox Mozilla.Firefox 99.0.1  100.0     winget
PS C:\Users\easton> wingetdev list Mozilla --all
Name                        Id              Version Available Source
--------------------------------------------------------------------
Mozilla Firefox (x64 en-US) Mozilla.Firefox 99.0.1  100.0     winget
Mozilla Maintenance Service Mozilla.Firefox 99.0.1  100.0     winget
PS C:\Users\easton>
```

The table printed in `winget upgrade` is always the simpler table (since we can only upgrade things we matched anyway), and the count of "upgrades available" is mapped to the number of packages instead of the number of ARP entries.

I know I did this without opening an issue to discuss it, so if there's a better solution I'm happy to fix this (or close it).</s>

Edit: If you're just joining us, @JohnMcPMS [brought up some great points with my original solution](https://github.com/microsoft/winget-cli/pull/2137#pullrequestreview-962735180), and as such this PR doesn't do as much as it did. Now, winget will always use the PackageName from the Add and Remove Programs table, instead of the one in the manifest:

```
PS C:\Users\easton> wingetdev list Mozilla
Name                        Id              Version Available Source
--------------------------------------------------------------------
Mozilla Firefox (x64 en-US) Mozilla.Firefox 99.0.1  100.0     winget
Mozilla Maintenance Service Mozilla.Firefox 99.0.1  100.0     winget
PS C:\Users\easton>
```

There is no argument to control this anymore, since if we tried to deduplicate the list with the currently available data we'd be hiding info (since winget isn't smart enough yet to uninstall all of the ARP entries associated with a certain package). Hopefully we'll be able to come back to that solution once the correlation logic is a bit more mature.

Tested: manually, although I'm happy to write tests if necessary.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2137)


src/AppInstallerCLICore/ExecutionArgs.h
```diff
@@ -81,6 +81,7 @@ namespace AppInstaller::CLI::Execution
 
             // Other
             All, // Used in Update command to update all installed packages to latest
+        	ListAll, // Used in List command to list all installed apps associated with a package
```
Trenly, line 84: Why do you need to add a new arg here? Couldn't you reuse `AppInstaller::CLI::Execution::Args::Type::All` ?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -731,26 +732,50 @@ namespace AppInstaller::CLI::Workflow
...
                     Utility::LocIndString availableVersion, sourceName;
+                    Utility::LocIndString packageId = match.Package->GetProperty(PackageProperty::Id);
 
                     if (latestVersion)
                     {
                         if (updateAvailable)
                         {
                             availableVersion = latestVersion->GetProperty(PackageVersionProperty::Version);
-                            availableUpgradesCount++;
+                            if (context.Args.Contains(Execution::Args::Type::ListAll) || !packageIdsPrinted.count(packageId))
```
Trenly, line 742: Is there any possibility that there could be two packages with the same ID from different sources?

src/AppInstallerCLICore/ExecutionArgs.h
jedieaston: `AppInstaller::CLI::Execution::Args::Type::All` is used in `winget upgrade --all`, where I wanted the simplified table to print out. Also, there's a different help string for this new argument, which would mean I'd have to have the Argument defined only in ListCommand instead of in the Arguments.cpp file with the rest of them...

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -731,26 +732,50 @@ namespace AppInstaller::CLI::Workflow
...
-                        match.Package->GetProperty(PackageProperty::Name),
-                        match.Package->GetProperty(PackageProperty::Id),
-                        installedVersion->GetProperty(PackageVersionProperty::Version),
-                        availableVersion,
-                        shouldShowSource ? sourceName : ""s
-                        });
+                    if (context.Args.Contains(Execution::Args::Type::ListAll))
+                    {
+                       table.OutputLine({
+							installedVersion->GetProperty(PackageVersionProperty::Name),
```
yao-msft, line 756: nit: a bunch of tabs should be replaced with spaces across the change

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
jedieaston: There is a possibility, yes. Should the set be a pair of the package ID and source?

Trenly: I think that might prevent false positives

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -63,6 +63,8 @@ namespace AppInstaller::CLI
             return Argument{ "msix", 'm', Args::Type::Msix, Resource::String::MsixArgumentDescription, ArgumentType::Flag };
         case Args::Type::ListVersions:
             return Argument{ "versions", NoAlias, Args::Type::ListVersions, Resource::String::VersionsArgumentDescription, ArgumentType::Flag };
+        case Args::Type::ListAll:
```
yao-msft, line 66: nit: since this is for list command specifically, this could be moved to ListCommand.cpp, just like the UpgradeCommand. This Argument::ForType was initially including every argument, then we switched to be "Argument::ForType is for common args, and command specific ones could just be defined in the specific command.cpp". Btw, upgrade all does not have an alias, should this be the same for consistency?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -710,6 +710,7 @@ namespace AppInstaller::CLI::Workflow
         int unknownPackagesCount = 0;
         auto &source = context.Get<Execution::Data::Source>();
         bool shouldShowSource = source.IsComposite() && source.GetAvailableSources().size() > 1;
+        std::set<Utility::LocIndString> packageIdsPrinted = std::set<Utility::LocIndString>();
```
yao-msft, line 713:
```suggestion
        std::set<Utility::LocIndString> packageIdsPrinted;

```

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -731,26 +732,50 @@ namespace AppInstaller::CLI::Workflow
...
+                           installedVersion->GetProperty(PackageVersionProperty::Name),
+                           match.Package->GetProperty(PackageProperty::Id),
+                           installedVersion->GetProperty(PackageVersionProperty::Version),
+                           availableVersion,
+                           shouldShowSource ? sourceName : ""s
+                       });
+                    }
+                    else
+                    {
+	                    // we need to only list once per package
```
yao-msft, line 765: nit: one more tab here :)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
yao-msft: yes, PackageId + SourceId will ensure a unique "winget package"

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1308,4 +1308,7 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="NoPackageSelectionArgumentProvided" xml:space="preserve">
     <value>No package selection argument was provided; see the help for details about finding a package.</value>
   </data>
+  <data name="ListAllArgumentDescription" xml:space="preserve">
+    <value>List all software installed by a package</value>
```
yao-msft, line 1312: nit: we try to make "package" the terminology for software across winget but I can see it's weird to say "all packages installed by a package"

Maybe "List all package components installed by a package" or "List all sub-packages installed by a package"

Open to any suggestions.

jedieaston: @denelon, thoughts? I think "List all components installed by a package makes the most sense".

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
jedieaston:
```suggestion
                            // we need to only list once per package
```

src/AppInstallerCLICore/Argument.cpp
jedieaston: Fixed, and I removed the alias for consistency :)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
jedieaston: Resolved!

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -731,26 +732,50 @@ namespace AppInstaller::CLI::Workflow
...
                     if (latestVersion)
                     {
+                        // Always show the source for correlated packages
+                        sourceName = latestVersion->GetProperty(PackageVersionProperty::SourceName);
+
                         if (updateAvailable)
                         {
                             availableVersion = latestVersion->GetProperty(PackageVersionProperty::Version);
-                            availableUpgradesCount++;
+                            if (context.Args.Contains(Execution::Args::Type::ListAll) || !packageIdsPrinted.count({ packageId, sourceName }))
```
yao-msft, line 745: Sorry, I may not make clear in previous comment. SourceName is for display. SourceIdentifier should be used to identify a source. We should use value from PackageVersionProperty::SourceIdentifier for the duplicate check.

jedieaston: Resolved in latest commit!

src/AppInstallerCLICore/Commands/ListCommand.cpp
```diff
@@ -24,6 +24,7 @@ namespace AppInstaller::CLI
             Argument::ForType(Execution::Args::Type::Exact),
             Argument::ForType(Execution::Args::Type::CustomHeader),
             Argument::ForType(Execution::Args::Type::AcceptSourceAgreements),
+            Argument{ "all", Argument::NoAlias, Execution::Args::Type::ListAll, Resource::String::ListAllArgumentDescription, ArgumentType::Flag},
```
JohnMcPMS, line 27: I agree with @Trenly that this could just use `Args::Type::All` (especially now that its not in `ForType` [which existed pre-localization, but now I don't think we should use]).

Ultimately it's just another internal enum value, so it is more about how its makes us feel than anything else.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: I refer to these as "system artifacts" internally, but that probably isn't a great name for them publicly.  I do like "components" or "items".

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -730,27 +731,52 @@ namespace AppInstaller::CLI::Workflow
                 // The only time we don't want to output a line is when filtering and no update is available.
                 if (updateAvailable || !m_onlyShowUpgrades)
                 {
-                    Utility::LocIndString availableVersion, sourceName;
+                    Utility::LocIndString availableVersion, sourceName, sourceIdentifier;
+                    Utility::LocIndString packageId = match.Package->GetProperty(PackageProperty::Id);
```
JohnMcPMS, line 735: I would suggest that rather than storing the `packageId` and `sourceIdentifier` separately, just store the `pair` directly.  That would remove 3 times where the strings will be copied to create a temporary pair.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -730,27 +731,52 @@ namespace AppInstaller::CLI::Workflow
...
+                           availableVersion,
+                           shouldShowSource ? sourceName : ""s
+                       });
+                    }
+                    else
+                    {
+                        // we need to only list once per package
+                        if (!packageIdsPrinted.count({ packageId, sourceIdentifier }))
+                        {
+                            packageIdsPrinted.insert({ packageId, sourceIdentifier });
```
JohnMcPMS, line 769:
```suggestion
                            packageIdsPrinted.emplace(std::move(thePair));
```
Assuming that you also took my other suggestion above about keeping these as a `pair` rather than separate.  And you named it `thePair`, which I'm sure you would come up with a better name than that 😉

JohnMcPMS (commented):
I gather from the example case that multiple `AppsAndFeaturesEntries` are being used to associate multiple components of some software together.  This change is then leveraging that to make it appear that we are being smarter than we are about handling that.  The problem that I see is that targeting these installed items becomes even more confusing, as the default `list` behavior will look like there is just one "Mozilla Firefox".

However, if I want to uninstall it, this falls into the "we don't handle side by side versions well".  I would expect that I cannot currently `uninstall Mozilla.Firefox` if both the maintenance service and primary application are correlated.

Ultimately, I think we need to solve the side by side and "multi-component package" issues by having the correlation engine actually output a single `Package` that has all of the details on its:
1. Multiple versions
2. Multiple components

Only then will the problem actually be solved.  But that might also require updates to the manifest to be able to mark one of the components as "primary" so that we know what to uninstall (assuming that it handles removing the other components).  And then of course all of the interaction models will need to be updated to handle targeting versions and components within as desired.

All of this isn't to say that your change is bad, just that I think it might end up hiding the problem that is created by the way the manifests are being constructed.  I think it points to a good way to deal with things at the surface, but we really need the internals to be in agreement or we will end up with a mess.

jedieaston:
> But that might also require updates to the manifest to be able to mark one of the components as "primary" so that we know what to uninstall (assuming that it handles removing the other components).

I completely spaced on this, as Firefox (the case I found the bug on) cleans up the maintenance service when you run the Firefox uninstaller. You're right though, winget needs to know if there's multiple uninstallers to run (and run them all). There's definitely other software (Autodesk stuff comes to mind) that make you uninstall each and every thing they add to the ARP table separately, so winget would need to know to execute all of the uninstallers associated with the ARP entries in the manifest. 

Should we table this for now then until the side-by-side/correlation work is done? I don't know if that's scheduled to happen in 1.3 or not, but if it's post 1.3 then I think there needs to be some fix for this table, since currently it's super confusing. (Maybe just no matter what print the ARP name for now, so the user can always know what name to tell `winget uninstall`?

JohnMcPMS:
Scheduled and "can fit in with all the random stuff going on" can be quite different 😄 I do like the idea of outputting a disambiguation and the name does seem like the best place for that.  The issue with removing the 2nd through nth items will be that maybe one of those was the "real" one that you should `uninstall`.  Until we know which one that is, I think we do want to just output all of them.

JohnMcPMS:
@jedieaston, are you planning to convert this PR based on our discussion?  Given the current situation I think we need to at least switch to using the installed version's name value for list.  If you aren't I can make a PR for that purpose.

jedieaston:
I am, let me try to get to it today. Sorry for the delay, I had to graduate from college over the weekend :)

JohnMcPMS:
> I am, let me try to get to it today. Sorry for the delay, I had to graduate from college over the weekend :)

🎊 Congratulations! 🎊

Oh no, I've dropped this https://careers.microsoft.com/us/en

Not trying to get you to do it, just didn't want to duplicate work if you were already doing it.

jedieaston:
@JohnMcPMS, I removed all of the code regarding the argument (wonderful thing about Git is that I can get that work back later), and now all this PR does is use the name in the ARP table in the list/upgrade table (the easiest solution to the current problem). Once this merges, I can write up an issue to talk about the deduplication stuff.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -731,26 +731,32 @@ namespace AppInstaller::CLI::Workflow
                 if (updateAvailable || !m_onlyShowUpgrades)
                 {
                     Utility::LocIndString availableVersion, sourceName;
+                    Utility::LocIndString packageId = match.Package->GetProperty(PackageProperty::Id);
```
JohnMcPMS, line 734: Getting value into `packageId` but not using it.

jedieaston: Resolved in latest commit. Oops!

JohnMcPMS (approved):

PR #2138, opened 2022-05-04
Fix packageId and sourceId match check for portable install

ryfu-msft:
Fixes a bug where the portable install flow should be checking whether ARP entry does NOT match the packageId or sourceId before blocking install.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2138)

### DIFF
--- a/src/AppInstallerCommonCore/PortableARPEntry.cpp
+++ b/src/AppInstallerCommonCore/PortableARPEntry.cpp
@@ -62,6 +62,7 @@ namespace AppInstaller::Registry::Portable
         }
         else
         {
+            m_exists = false;
             m_key = Key::Create(root, subKey);
         }
     }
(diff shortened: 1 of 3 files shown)


vedantmgoyal2009:
I just experienced this bug today and was about to open an issue. Thanks for the fix!

yao-msft (approved):
:shipit:

src/AppInstallerCommonCore/PortableARPEntry.cpp
```diff
@@ -62,6 +62,7 @@ namespace AppInstaller::Registry::Portable
         }
         else
         {
+            m_exists = false;
```
yao-msft, line 65: m_exists should be initialized to false in the header

JohnMcPMS (approved):

src/AppInstallerCommonCore/PortableARPEntry.cpp
ryfu-msft: add initialization to header

JohnMcPMS: nit: that would mean you wouldn't need to do it here.

PR #2139, opened 2022-05-05
Load index from validated msix for unpackaged context

yao-msft:
Validation: Added new tests and fixed existing tests. Also did manual verification that source opening works as expected.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2139)


src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.h
```diff
@@ -57,7 +58,7 @@ namespace AppInstaller::Repository::Microsoft
         };
 
         // Opens an existing index database.
-        static SQLiteIndex Open(const std::string& filePath, OpenDisposition disposition);
+        static SQLiteIndex Open(const std::string& filePath, OpenDisposition disposition, TempSQLiteIndexFile&& tempIndexFileHandle = {});
```
JohnMcPMS, line 61: I don't love this being passed in, but I understand the need.  Can we change the type to `WriteLockedFile` or something more generic?  That could be used for both the package and index files (with a `create` bool that controls both open and remove on destruction functionality).

Then also just move the "extract the index file" part out of the type as it isn't really necessary to be included.

JohnMcPMS (changes requested):

src/AppInstallerCLITests/TestCommon.cpp
```diff
@@ -214,4 +215,74 @@ namespace TestCommon
     {
         AppInstaller::Settings::SetUserSettingsOverride(nullptr);
     }
+
+    bool InstallCertFromSignedPackage(const std::filesystem::path& package)
```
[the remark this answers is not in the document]
yao-msft: I made the tests to skip if not running with admin. But I cannot find a way to "mark" the tests as skipped. So I just output an info message.

src/AppInstallerCLITests/MsixInfo.cpp
```diff
@@ -64,3 +55,37 @@ TEST_CASE("MsixInfo_WriteFile", "[msixinfo]")
 
     REQUIRE(1 == std::filesystem::file_size(file));
 }
+
+TEST_CASE("MsixInfo_ValidateMsixTrustInfo", "[msixinfo]")
+{
+    if (!Runtime::IsRunningAsAdmin())
+    {
+        INFO("Test requires admin privilege. Skipped.");
```
JohnMcPMS, line 63:
```suggestion
        WARN("Test requires admin privilege. Skipped.");
```
Not sure how this behaves instead, maybe it is more obvious?

src/AppInstallerCommonCore/Public/winget/ManagedFile.h
```diff
@@ -0,0 +1,36 @@
...
+    {
+        ManagedFile() = default;
+
+        ManagedFile(const ManagedFile&) = delete;
+        ManagedFile& operator=(const ManagedFile&) = delete;
+
+        ManagedFile(ManagedFile&&) = default;
+        ManagedFile& operator=(ManagedFile&&) = default;
+
+        HANDLE GetFileHandle() { return m_fileHandle.get(); }
```
JohnMcPMS, line 20:
```suggestion
        HANDLE GetFileHandle() const { return m_fileHandle.get(); }
```

src/AppInstallerCommonCore/Public/winget/ManagedFile.h
```diff
@@ -0,0 +1,36 @@
...
+        const std::filesystem::path& GetFilePath() { return m_filePath; }
```
JohnMcPMS, line 21:
```suggestion
        const std::filesystem::path& GetFilePath() const { return m_filePath; }
```

src/AppInstallerCommonCore/MsixInfo.cpp
```diff
@@ -117,6 +121,68 @@ namespace AppInstaller::Msix
...
+            {
+                ULONG bytesRead = 0;
+                HRESULT hr = stream->Read(buffer.get(), bufferSize, &bytesRead);
+
+                if (bytesRead)
+                {
+                    // If we got bytes, just accept them and keep going.
+                    LOG_IF_FAILED(hr);
+
+                    THROW_LAST_ERROR_IF(INVALID_SET_FILE_POINTER == SetFilePointer(target, 0, nullptr, FILE_END));
```
JohnMcPMS, line 143: Do we need to set this every time through the loop?

src/AppInstallerCommonCore/MsixInfo.cpp
```diff
@@ -117,6 +121,68 @@ namespace AppInstaller::Msix
...
+                    DWORD bytesWritten = 0;
+                    THROW_LAST_ERROR_IF(!WriteFile(target, buffer.get(), bytesRead, &bytesWritten, nullptr));
+                    THROW_HR_IF(E_UNEXPECTED, bytesRead != bytesWritten);
+                    totalBytesRead += bytesRead;
+                    progress.OnProgress(totalBytesRead, expectedSize, ProgressType::Bytes);
+                }
+                else
+                {
+                    // If given a size, and we have read it all, quit
+                    if (expectedSize && totalBytesRead == expectedSize)
```
JohnMcPMS, line 153: We don't appear to be doing anything to prevent writing more than `expectedSize`.  Is that intentional?

src/AppInstallerCommonCore/MsixInfo.cpp
```diff
@@ -335,6 +556,18 @@ namespace AppInstaller::Msix
         THROW_IF_FAILED(signatureStream->Stat(&stat, STATFLAG_NONAME));
         THROW_HR_IF(E_UNEXPECTED, stat.cbSize.HighPart != 0); // Signature size should be small
         signatureSize = stat.cbSize.LowPart;
+        THROW_HR_IF(E_UNEXPECTED, signatureSize <= P7xFileIdSize);
+
+        if (getRawSignature)
```
JohnMcPMS, line 561: `getRawSignature` being true, but the code then modifying the signature, doesn't make sense to me.  Maybe this was supposed to be `if (!getRawSignature)`?  Or should the name be changed to `skipFileID`?

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -361,15 +361,40 @@ namespace AppInstaller::Repository::Microsoft
...
+
+                // Validate index package trust info.
+                THROW_HR_IF(APPINSTALLER_CLI_ERROR_SOURCE_DATA_INTEGRITY_FAILURE, !Msix::ValidateMsixTrustInfo(packageLocation, WI_IsFlagSet(m_details.TrustLevel, SourceTrustLevel::StoreOrigin)));
+
+                // Create a temp lock exclusive index file.
+                GUID guid;
+                THROW_IF_FAILED(CoCreateGuid(&guid));
+                WCHAR tempFileName[256];
+                THROW_HR_IF(E_UNEXPECTED, StringFromGUID2(guid, tempFileName, ARRAYSIZE(tempFileName)) == 0);
+                auto tempIndexFilePath = Runtime::GetPathTo(Runtime::PathName::Temp);
```
JohnMcPMS, line 383: I would either create a new function `GetNewTempFilePath` or `GetPathTo(PathName::NewTempFile)`, but both would return the same type of path you are constructing here.  This is definitely reusable.

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -389,35 +414,65 @@ namespace AppInstaller::Repository::Microsoft
...
-                std::filesystem::path indexPath = packageState / s_PreIndexedPackageSourceFactory_IndexFileName;
+                std::filesystem::path packagePath = packageState / s_PreIndexedPackageSourceFactory_PackageFileName;
 
-                if (std::filesystem::exists(manifestPath) && std::filesystem::exists(indexPath))
+                if (std::filesystem::exists(packagePath))
                 {
-                    // If we already have a manifest, use it to determine if we need to update or not.
-                    if (!packageInfo.IsNewerThan(manifestPath))
+                    // If we already have a trusted index package, use it to determine if we need to update or not.
+                    if (Msix::ValidateMsixTrustInfo(packagePath, WI_IsFlagSet(details.TrustLevel, SourceTrustLevel::StoreOrigin)) &&
```
JohnMcPMS, line 428: You need another
```
auto indexPackageLock = Utility::ManagedFile::OpenWriteLockedFile(packagePath, 0);
```
to prevent a TOCTOU here as well (between validation and the `IsNewerThan`).

It might even be worth making a helper type like:
```
struct PersistedIndexPackage
{
PersistedIndexPackage(path);
bool ValidateMsixTrustInfo(bool checkMSOrigin);

private:
ManagedFile m_fileLock;
}
```
Then remove the free function `ValidateMsixTrustInfo`.  That way you *must* lock the file to validate it, and hopefully that translates into the appropriate lifetime of the lock in the calling code.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/MsixInfo.cpp
```diff
@@ -372,16 +589,16 @@ namespace AppInstaller::Msix
...
 
-    bool MsixInfo::IsNewerThan(const std::filesystem::path& otherManifest)
+    bool MsixInfo::IsNewerThan(const std::filesystem::path& otherPackage)
     {
         THROW_HR_IF(E_NOT_VALID_STATE, m_isBundle);
 
-        ComPtr<IStream> otherStream;
-        THROW_IF_FAILED(SHCreateStreamOnFileEx(otherManifest.c_str(), 
-            STGM_READ | STGM_SHARE_DENY_WRITE | STGM_FAILIFTHERE, 0, FALSE, nullptr, &otherStream));
+        MsixInfo other{ otherPackage.u8string() };
```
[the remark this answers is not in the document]
JohnMcPMS: I asked about this and found out that this is a reasonable way to do it:
```
    template<typename T, std::enable_if_t<std::is_same_v<T, std::filesystem::path>, int> = 0>
    MsixInfo(const T& path) { ConstructFromPath(path); }
```
Where `ConstructFromPath` can then be implemented in the `.cpp`.

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
yao-msft: by looking at GetPathTo() code, it's more of returning a directory caller can work on. I'll create GetNewTempFilePath() then

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -21,6 +21,22 @@ namespace AppInstaller::Repository::Microsoft
...
+        struct PersistedIndexPackage
+        {
+            PersistedIndexPackage(const std::filesystem::path& path)
+            {
+                m_file = Utility::ManagedFile::OpenWriteLockedFile(path, 0);
+            }
+
+            bool ValidateMsixTrustInfo(bool checkMicrosoftOrigin) const
+            {
+                return Msix::ValidateMsixTrustInfo(m_file.GetFilePath(), checkMicrosoftOrigin);
```
JohnMcPMS, line 33: The reason to move to this model was that the *only* way to call `ValidateMsixTrustInfo` would be to create a `PersistedIndexPackage`.  If you leave the free function `ValidateMsixTrustInfo` around, someone can still use it without the file lock.

JohnMcPMS (approved):

Issue #1315, opened 2021-07-20
Support for installers that install application as a Portable Application 


KaranKad:
# Description of the new feature/enhancement

There are some applications like [GNU/Emacs](https://github.com/microsoft/winget-pkgs/tree/master/manifests/g/GNU/Emacs) and [TorProject/TorBrowser](https://github.com/microsoft/winget-pkgs/tree/master/manifests/t/TorProject/TorBrowser) which come in a installer but install as a portable application. Because of that winget is not able to keep track of them.
# Proposed technical implementation details (optional)

denelon:
@KaranKad how is this different from:
* #182

denelon:
I wasn't aware of these. Thanks for the specific example. 

How would you like to see this happen?

What kinds of arguments might need to be used or added?
"--location" for example.

What kinds of settings make sense (if any)?

KaranKad:
I think winget needs to have a Program Files folder of its own where these types of packages should be installed. These installers do have location switch do they could be redirected there. Other thing that can be done is to treat installers as archive package similar to zip, 7z etc and just extract the application from it an place it in winget's Program Files folder.

PR #2140, opened 2022-05-05
Implementation for Portable Uninstall and Upgrade

ryfu-msft:
Related to: 
- #182 
- #1315 

This PR adds the implementation for supporting the uninstall and upgrade scenario for the portable installerType:

Changes:
- Rearranges the write to registry order so that if the upgrade scenario fails, the existing registry values are preserved.
- Adds the uninstall flow so that the uninstall flow removes the appropriate installed files related to the portable.
- Adds cleanup functionality to the install flow if install fails.
- Adds support for --purge and --preserve

Tests:
- Added unit tests to workflow for update and uninstall
- Added E2E tests for upgrade with the uninstallPrevious upgrade behavior
- Added E2E tests for uninstall as well as incorporating uninstall in the cleanup process for install tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2140)


src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -269,6 +269,13 @@ namespace AppInstaller::CLI::Workflow
                 (isUpdate ? MSStoreUpdate : MSStoreInstall);
             break;
         case InstallerTypeEnum::Portable:
+            if (isUpdate && installer.UpdateBehavior == UpdateBehaviorEnum::UninstallPrevious)
```
Trenly, line 272: I think this might be redundant? Isn’t the manifest validation that users can’t set upgradeBehavior for Portable type installers, and they will always be handled as uninstallPrevious

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
```diff
@@ -21,12 +21,15 @@ namespace AppInstaller::CLI
...
             Argument::ForType(Args::Type::Moniker),
+            Argument::ForType(Args::Type::ProductCode),
             Argument::ForType(Args::Type::Version),
             Argument::ForType(Args::Type::Channel),
             Argument::ForType(Args::Type::Source),
             Argument::ForType(Args::Type::Exact),
             Argument::ForType(Args::Type::Interactive),
             Argument::ForType(Args::Type::Silent),
+            Argument::ForType(Args::Type::Purge),
+            Argument::ForType(Args::Type::Preserve),
```
Trenly, line 32: Do these need to be added to the UpgradeCommand as well? I could see a case where a user would want to `--preserve` any user data files inside their folder when upgrading a portable app

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
```diff
@@ -90,13 +94,19 @@ namespace AppInstaller::CLI
...
              execArgs.Contains(Execution::Args::Type::Source) ||
              execArgs.Contains(Execution::Args::Type::Exact)))
         {
             throw CommandException(Resource::String::BothManifestAndSearchQueryProvided, "");
         }
+
+        if (execArgs.Contains(Execution::Args::Type::Purge) && execArgs.Contains(Execution::Args::Type::Preserve))
+        {
+            throw CommandException(Resource::String::BothPurgeAndPreserveFlagsProvided, "");
+        }
```
Trenly, line 109: This should probably be moved up a level into `Command::ValidateArguments` since it will apply to multiple commands, similar to how the Silent/Interactive flags are handled

src/AppInstallerCLIE2ETests/UpgradeCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace AppInstallerCLIE2ETests
+{
+    using NUnit.Framework;
+    using System.IO;
+
+    public class UpgradeCommand : BaseCommand
+    {
+        [Test]
+        public void UpgradePortable()
```
Trenly, line 12: Should there also be tests for when the ARP data is mismatched?
* PortableUpgradeARPMismatch
* PortableUpgradeForcedOverride

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -269,6 +269,13 @@ namespace AppInstaller::CLI::Workflow
                 (isUpdate ? MSStoreUpdate : MSStoreInstall);
             break;
         case InstallerTypeEnum::Portable:
+            if (isUpdate && installer.UpdateBehavior == UpdateBehaviorEnum::UninstallPrevious)
+            {
+                context <<
+                    GetUninstallInfo <<
+                    ExecuteUninstaller;
```
Trenly, line 276: I think the upgrade flow for portable apps might need to be a little bit more complex. Based on running an uninstall on upgrade, with the current logic, you're removing the executable, the install directory, and the symlink. If an upgrade is being performed, wouldn't it make more sense to leave the install directory and symlink, replacing just the binary itself?

denelon: One of the challenges is that the binary might still be running.

Trenly: > One of the challenges is that the binary might still be running.

Wouldn't that be a challenge either way? The executable wouldn't be removed if it were still running regardless of if it is just the executable being removed, or the entire structure of executable, directory, and symlink. I think failing to delete the executable while it is still running makes sense, and is what I would expect. 

To clarify, I'm not saying that the binary should be overwritten without an attempted deletion first. I'm just noting that in an upgrade scenario, it doesn't seem logical to me to remove the folder and the symlink just to create them again when the app is installed

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -280,16 +332,116 @@ namespace AppInstaller::CLI::Workflow
...
+            }
+            else
+            {
+                AICLI_LOG(CLI, Info, << "The registry value for [TargetFullPath] does not exist");
+            }
+        }
+
+        void RemoveInstallDirectory(Execution::Context& context)
+        {
+            PortableARPEntry uninstallEntry = GetPortableARPEntryForUninstall(context);
```
JohnMcPMS, line 395: Why not insert the `PortableARPEntry` into the context and get it out each time rather than reconstructing it repeatedly?

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -303,14 +455,66 @@ namespace AppInstaller::CLI::Workflow
...
+            {
+                installedScope = context.Get<Execution::Data::InstalledPackageVersion>()->GetMetadata()[Repository::PackageVersionMetadata::InstalledScope];
+            }
+
+            if (std::filesystem::is_empty(GetPortableLinksLocation(ConvertToScopeEnum(installedScope))))
+            {
+                RemoveFromPathRegistry(context);
+            }
+
+            uninstallEntry.Delete();
```
JohnMcPMS, line 500: Please intentionally remove the registry key on its own rather than hiding it inside the function for removing the symlink.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -341,17 +545,23 @@ namespace AppInstaller::CLI::Workflow
...
+    }
 
     void PortableInstallImpl(Execution::Context& context)
     {
         try
         {
             context.Reporter.Info() << Resource::String::InstallFlowStartingPackageInstall << std::endl;
 
+            // Add product code for uninstall subcontext prior to install in case of install failure.
+            const std::string& productCode = GetPortableProductCode(context);
```
JohnMcPMS, line 557: Why not construct this here before the `try` then have the uninstall invoked from this function as well?  I don't see any value in splitting it into two tasks.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -362,8 +572,26 @@ namespace AppInstaller::CLI::Workflow
...
+        try
+        {
+            context.Reporter.Info() << Resource::String::UninstallFlowStartingPackageUninstall << std::endl;
+
+            context <<
+                RemovePortableExe <<
+                RemoveInstallDirectory <<
+                RemovePortableSymlink;
+
+            context.Reporter.Info() << Resource::String::UninstallFlowUninstallSuccess << std::endl;
```
JohnMcPMS, line 588: You will print success even if a task has failed.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -126,6 +127,18 @@ namespace AppInstaller::CLI::Workflow
...
         }
+        case InstallerTypeEnum::Portable:
+        {
+            auto productCodes = installedPackageVersion->GetMultiProperty(PackageVersionMultiProperty::ProductCode);
+            if (productCodes.empty())
+            {
+                context.Reporter.Error() << Resource::String::NoUninstallInfoFound << std::endl;
+                AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_NO_UNINSTALL_INFO_FOUND);
+            }
+            context.Add<Execution::Data::ProductCodes>(std::move(productCodes));
```
JohnMcPMS, line 138: There is no need to reuse this value.  And I would suggest that creating a context data item for the ARP entry object would be better anyway.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -1063,6 +1067,11 @@ namespace AppInstaller::CLI::Workflow
             {
                 searchRequest.Inclusions.emplace_back(PackageMatchFilter(PackageMatchField::ProductCode, MatchType::Exact, installer.ProductCode));
             }
+            else if (installer.InstallerType == Manifest::InstallerTypeEnum::Portable)
+            {
+                const auto& productCode = Utility::MakeSuitablePathPart(manifest.Id + "_" + source.GetIdentifier());
+                searchRequest.Inclusions.emplace_back(PackageMatchFilter(PackageMatchField::ProductCode, MatchType::Exact, Utility::Normalize(productCode)));
```
JohnMcPMS, line 1073:
```suggestion
                searchRequest.Inclusions.emplace_back(PackageMatchFilter(PackageMatchField::ProductCode, MatchType::CaseInsensitive, Utility::Normalize(productCode)));
```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1314,4 +1314,24 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="NoPackageSelectionArgumentProvided" xml:space="preserve">
     <value>No package selection argument was provided; see the help for details about finding a package.</value>
   </data>
+  <data name="PortableInstallFailed" xml:space="preserve">
+    <value>Portable install failed; Cleaning up...</value>
+  </data>
+  <data name="BothPurgeAndPreserveFlagsProvided" xml:space="preserve">
+    <value>Both `purge` and `preserve` args are provided</value>
```
JohnMcPMS, line 1321:
```suggestion
    <value>Both `purge` and `preserve` arguments are provided</value>
```
Don't abbreviate as it can make it harder for non-English speakers to machine-translate for languages we don't have localizations for.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
Trenly: Even if the default is to preserve the directory, I could then see a case for a user to want to use —purge to perform a clean upgrade

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
Trenly: I was mistaken; I was thinking of the scope property, which isn’t able to be specified in the manifest for Portable installer type

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: After processing the change a little more, it might already be doing `upgrade` as "install over top of".  I think that the current pattern isn't quite forward looking for portable archives, and I'm not sure that upgrade will be as clean as install got to be for being able to remove at any point in the process.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -297,6 +304,22 @@ namespace AppInstaller::CLI::Workflow
         context <<
             PortableInstallImpl <<
             ReportInstallerResult("Portable"sv, APPINSTALLER_CLI_ERROR_PORTABLE_INSTALL_FAILED, true);
+
+        const auto& installReturnCode = context.Get<Execution::Data::OperationReturnCode>();
+
+        if (installReturnCode != 0)
```
JohnMcPMS, line 310: I don't think that an `upgrade` should uninstall if it fails.  It should probably just leave things how they are for the user to sort out.

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
ryfu-msft: I decided to only add `--purge` to the UpgradeCommand and not have the `UninstallBehavior.PurgePortablePackage` settings apply to the upgrade command as that would seem like an unexpected behavior. `--purge` will now remove all files and directories when performing an `uninstallPrevious`

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
ryfu-msft: I'm going to leave this here for now as I don't want to add this validation prematurely if its only currently being used in the uninstall command. (Upgrade command only supports --purge)

src/AppInstallerCLIE2ETests/UpgradeCommand.cs
ryfu-msft: Added suggested tests

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.h
```diff
@@ -51,6 +51,8 @@ namespace AppInstaller::Repository::Microsoft
         const std::wstring WindowsInstaller{ L"WindowsInstaller" };
         // REG_DWORD (bool)
         const std::wstring SystemComponent{ L"SystemComponent" };
+        // REG_SZ  (ex. "portable")
+        const std::wstring WinGetInstallerType{ L"WinGetInstallerType" };
```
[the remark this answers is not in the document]
ryfu-msft: Removed this defined string and replaced it's usage with `ToString(PortableValueName::WinGetInstallerType)`

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -162,6 +162,8 @@ namespace AppInstaller::Repository
         InstalledType,
         // The Scope of an installed package
         InstalledScope,
+        // The Architecture of an installed package
```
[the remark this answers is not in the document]
ryfu-msft: Moved InstalledArchitecture to the end and added suggested comment

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -208,11 +245,43 @@ namespace AppInstaller::CLI::Workflow
...
+            }
+
+            std::wstring pathName = std::wstring{ s_PathName };
+            std::string portableLinksDir = Normalize(linksDirectory.u8string());
+            std::string pathValue = Normalize(key[pathName]->GetValue<Value::Type::String>());
+
+            if (pathValue.find(portableLinksDir) != std::string::npos)
+            {
+                portableLinksDir += ";";
+                FindAndReplace(pathValue, portableLinksDir, "");
```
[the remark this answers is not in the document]
ryfu-msft: Changed to suggestion

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -280,16 +332,116 @@ namespace AppInstaller::CLI::Workflow
...
+                    {
+                        std::filesystem::remove(targetPathValue);
+                        AICLI_LOG(CLI, Info, << "Successfully deleted portable exe:" << targetPathValue);
+                    }
+                    else
+                    {
+                        if (context.Args.Contains(Execution::Args::Type::HashOverride))
+                        {
+                            context.Reporter.Warn() << Resource::String::PortableHashMismatchOverridden << std::endl;
+                            std::filesystem::remove(targetPathValue);
```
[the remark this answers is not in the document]
ryfu-msft: Fixed

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Created a separate task for removing the registry key.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -40,12 +40,47 @@ namespace AppInstaller::CLI::Workflow
...
+                installedScope = context.Get<Execution::Data::InstalledPackageVersion>()->GetMetadata()[Repository::PackageVersionMetadata::InstalledScope];
+            }
+
+            return Portable::PortableARPEntry(ConvertToScopeEnum(installedScope), installedArch, productCode);
+        }
+
+        PortableARPEntry GetPortableARPEntryForInstall(Execution::Context& context)
+        {
+            return Portable::PortableARPEntry(
+                ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope)),
```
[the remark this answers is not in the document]
ryfu-msft: I created a PortableARPEntry in context so that the entry is passed between tasks

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Added a PortableARPEntry data to context to be used between tasks

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Changed to constructing a PortableARPEntry before the `try` and adding that to context. Also changed to having the uninstall flow invoked from here

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Changed to only printing success of the operationReturnCode is 0

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
ryfu-msft: Created `PortableARPEntry` context data item

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
ryfu-msft: Changed to suggested

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: I actually followed the pattern of an existing string, so I updated both to use `arguments` instead of `args`

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Added a check that only performs a cleanup if install fails and the operation is not an upgrade.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: The uninstall flow is only invoked if the upgrade behavior is set to `uninstallPrevious`. If that is not present, then the upgrade will call the regular install flow. I have changed the install flow so that it replaces the files that needs to be updated and writes that change to registry. If the upgrade fails at any point, the existing version is preserved.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -281,15 +297,124 @@ namespace AppInstaller::CLI::Workflow
...
+                std::filesystem::remove(targetFullPath);
+                AICLI_LOG(CLI, Info, << "Removing existing portable exe at: " << targetFullPath);
+            }
 
             Filesystem::RenameFile(installerPath, targetFullPath);
             AICLI_LOG(CLI, Info, << "Portable exe moved to: " << targetFullPath);
 
+            if (!uninstallEntry[PortableValueName::InstallDirectoryCreated].has_value())
+            {
+                uninstallEntry.SetValue(PortableValueName::InstallDirectoryCreated, isDirectoryCreated);
```
JohnMcPMS, line 311: Why not just always set it?

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -281,15 +297,124 @@ namespace AppInstaller::CLI::Workflow
...
+            if (installDirectory.has_value())
+            {
+                const std::filesystem::path& installDirectoryValue = installDirectory.value().GetValue<Value::Type::UTF16String>();
+                bool isUpdate = WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerExecutionUseUpdate);
+
+                if (context.Args.Contains(Execution::Args::Type::Purge) ||
+                    (!isUpdate && Settings::User().Get<Settings::Setting::UninstallPurgePortablePackage>() && !context.Args.Contains(Execution::Args::Type::Preserve)))
+                {
+                    context.Reporter.Warn() << Resource::String::PurgeInstallDirectory << std::endl;
+                    const auto& removedFilesCount = std::filesystem::remove_all(installDirectoryValue);
```
JohnMcPMS, line 379: I'm not sure that we should support purging a directory that we did not create.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -363,7 +546,58 @@ namespace AppInstaller::CLI::Workflow
...
+                RemoveInstallDirectory <<
+                RemovePortableSymlink <<
+                RemovePortableARPEntry;
+
+            context.Add<Execution::Data::OperationReturnCode>(context.GetTerminationHR());
+        }
+        catch (...)
+        {
+            context.Add<Execution::Data::OperationReturnCode>(Workflow::HandleException(context, std::current_exception()));
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_PORTABLE_UNINSTALL_FAILED);
```
JohnMcPMS, line 582: This seems like it shouldn't terminate the context, but continue on like install does.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -363,7 +546,58 @@ namespace AppInstaller::CLI::Workflow
...
+        }
+
+        const auto& uninstallResult = context.Get<Execution::Data::OperationReturnCode>();
```
JohnMcPMS, line 585: We need a task to report uninstall result like we have for install.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Since upgrade and install use the same flow, if we always set it, then any upgrade after the first install will always overwrite this value to false because the directory now exists, even if we were the ones who created it during the first install. I think we should try to honor the initial set value because that should be set when the package is first installed.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Changed the flow around so that purging only works if we created the directory. Shows warning message to user if purge is provided and we did not create the directory.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Removed termination

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
ryfu-msft: Created a task similar to ReportInstallerResult called ReportUninstallerResult that processes the operation return code from the portable uninstall flow.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -231,4 +247,39 @@ namespace AppInstaller::CLI::Workflow
...
+            {
+                context.Reporter.Error() << Resource::String::UninstallFailedWithCode << ' ' << GetUserPresentableMessage(uninstallResult) << std::endl;
+            }
+            else
+            {
+                context.Reporter.Error() << Resource::String::UninstallFailedWithCode << ' ' << uninstallResult << std::endl;
+            }
+
+            // Show installer log path if exists
+            if (context.Contains(Execution::Data::LogPath) && std::filesystem::exists(context.Get<Execution::Data::LogPath>()))
```
JohnMcPMS, line 273: I don't think we have an uninstall log file, but if we ever happen to I suppose this is ready for it.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -201,14 +203,28 @@ namespace AppInstaller::CLI::Workflow
...
         }
 
         context.Reporter.Info() << Resource::String::UninstallFlowUninstallSuccess << std::endl;
     }
 
+    void ShellExecuteUninstall(Execution::Context& context)
+    {
+        context <<
+            Workflow::ShellExecuteUninstallImpl <<
+            ReportUninstallerResult("UninstallString", APPINSTALLER_CLI_ERROR_EXEC_UNINSTALL_COMMAND_FAILED);
```
JohnMcPMS, line 218: nit: I don't see the value in splitting this for two types when MSIX call its `ReportUninstallerResult` in `ExecuteUninstaller`.  Why not just put all of them in `ExecuteUninstaller`?

JohnMcPMS (approved):

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
ryfu-msft: Changed so that it directly calls from ExecuteUninstaller

PR #2144, opened 2022-05-06
Treat manifest localization validation error as warning for non full validation(manifest reading)

yao-msft:
Added tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2144)


JohnMcPMS (approved):
Do we also have tests in place that show that `Invalid` is still a "valid" option?  For instance, if there is only one installer set to `neutral`, would it be chosen or would it be inapplicable?

yao-msft:
> Do we also have tests in place that show that `Invalid` is still a "valid" option? For instance, if there is only one installer set to `neutral`, would it be chosen or would it be inapplicable?

Currently, for rest source, if the bad data is found in Installers, we'll treat the rest response as invalid. It won't go far enough to installer selection step.

Issue #980, opened 2021-05-21
Add alternative version number field to manifest to make correlation with entries in Add/Remove programs more reliable


chausner:
# Description of the new feature/enhancement

As reported [here](https://github.com/microsoft/winget-cli/issues/976), the `winget upgrade` command may incorrectly detect updates to be available even when the latest version is installed in case the version number in Add/Remove programs does not match the package version in the manifest.

One solution to this problem could be the addition of an additional version field to the manifest where the version number is specified in the format that is expected in Add/Remove programs. When present, the `winget upgrade` routine would use this version number instead of the package version to determine if an update is available or not.

This suggestion is closely related to microsoft/winget-pkgs#13620.

denelon:
This is part of the work we're doing with the new schema. We have added a section for data specifically matching what is reported via Apps & Features.

denelon:
Related to:
* #1952
* #1073

denelon:
We're going to use this Issue to represent the work to start looking at the version in the AppsAndFeatures section of the manifest (if it is poopulated) rather than the PackageVersion for version comparison.

```YAML
PackageVersion:         # Used for version comparison if no DisplayVersion key is present
- AppsAndFeatures
    DisplayVersion:     # Used for version comparison if the key is present
```

denelon:
When we have an installed package and the manifest in the repository uses the AppsAndFeatures version we need a rational way to display the "somewhat unknown" "displayVersion" to the user.

The example below is contrived, but is intended to illustrate the syntax we're considering using.

We're thinking of using "> " for the `winget list` scenario where the user has a newer version installed.

```text
Name           Id                   Version   Available Source
------------------------------------------------------------
Git            Git.Git              > 2.36.0            winget
GitHub Desktop GitHub.GitHubDesktop 2.9.15    3.0.0     winget
```

We're thinking of using "< " for the `winget upgrade` scenario where the users has an older version installed, but it doesn't match an existing manifest.

```text
Name           Id                   Version   Available Source
------------------------------------------------------------
Git            Git.Git              < 2.36.0  2.36.0    winget
GitHub Desktop GitHub.GitHubDesktop 2.9.15    3.0.0     winget
```

PR #2162, opened 2022-05-16
Doc for ARP version mapping change

yao-msft:
For #980

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2162)

### DIFF
--- a/doc/specs/#980 - Apps and Features entries version mapping.md
+++ b/doc/specs/#980 - Apps and Features entries version mapping.md
@@ -1,0 +1,60 @@
+---
+author: <first-name> <last-name> <github-id>/<email>
+created on: <yyyy-mm-dd>
+last updated: <yyyy-mm-dd>
+issue id: <github issue id>
+---
+
+# Spec Title
+
+[comment]: # Link to issue: "For [#1](https://github.com/microsoft/winget-cli/issues/1)"
+
+## Abstract
+
+[comment]: # Outline what this spec describes
+
+## Inspiration
+
+[comment]: # What were the drivers/inspiration behind the creation of this spec.
+
+## Solution Design
+
+[comment]: # Outline the design of the solution. Feel free to include ASCII-art diagrams, etc.
+
+## UI/UX Design
+
+[comment]: # What will this fix/feature look like? How will it affect the end user?
+
+## Capabilities
+
+[comment]: # Discuss how the proposed fixes/features impact the following key considerations:
+
+### Accessibility
+
+[comment]: # How will the proposed change impact accessibility for users of screen readers, assistive input devices, etc.
+
+### Security
+
+[comment]: # How will the proposed change impact security?
+
+### Reliability
+
+[comment]: # Will the proposed change improve reliability? If not, why make the change?
+
+### Compatibility
+
+[comment]: # Will the proposed change break existing code/behaviors? If so, how, and is the breaking change "worth it"?
+
+### Performance, Power, and Efficiency
+
+## Potential Issues
+
+[comment]: # What are some of the things that might cause problems with the fixes/features proposed? Consider how the user might be negatively impacted.
+
+## Future considerations
+
+[comment]: # What are some of the things that the fixes/features might unlock in the future? Does the implementation of this spec enable scenarios?
+
+## Resources
+
+[comment]: # Be sure to add links to references, resources, footnotes, etc.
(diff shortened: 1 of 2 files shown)


doc/specs/#980 - Apps and Features entries version mapping.md
```diff
@@ -0,0 +1,193 @@
...
+
+For example:  
+Version `1` is less than version `2`  
+Version `1.0.0` is less than version `2.0.0`  
+Version `0.0.1-alpha` is less than version `0.0.2-alpha`  
+Version `0.0.1-beta` is less than version `0.0.2-alpha`  
+Version `0.0.1-alpha` is less than version `0.0.1-beta`  
+Version `0.0.1-alpha` is less than version `0.0.1` 
+Version `13.9.8` is less than version `14.0`  
+Version `1.0` is equal to version `1.0.0`
```
JohnMcPMS, line 47: I don't know the answer, but based on "else if one side has no more parts, it is less ", `1.0` would be less than `1.0.0`.

JohnMcPMS (approved):

doc/specs/#980 - Apps and Features entries version mapping.md
```diff
@@ -0,0 +1,193 @@
...
+Version `0.0.1-alpha` is less than version `0.0.1-beta`  
+Version `0.0.1-alpha` is less than version `0.0.1` 
+Version `13.9.8` is less than version `14.0`  
+Version `1.0` is equal to version `1.0.0`  
+
+Both Winget version and ARP version will use the above parsing and comparison logic.
+
+### How Winget collects and stores ARP version
+
+The `DisplayVersion` field under `AppsAndFeaturesEntries` will be treated as ARP versions of the package.
```
Trenly, line 53: What will happen if a single app has multiple AppsAndFeaturesEntries for all of its components (e.g - Firefox) ? Is it expected that the first entry in AppsAndFeaturesEntries should be the "primary" component ARP entry?

doc/specs/#980 - Apps and Features entries version mapping.md
```diff
@@ -0,0 +1,193 @@
...
+For example:  
+Version `< 3.0` is less than version `3.0`  
+Version `< 3.0` is greater than version `2.9`(because `3.0` is greater than `2.9`)  
+Version `< 3.0` is less than version `4.0`  
+Version `< 3.0` is less than version `> 3.0`  
+Version `> 3.0` is greater than version `3.0`  
+Version `> 3.0` is less than version `3.1`  (because `3.0` is less than `3.1`)  
+Version `> 3.0` is greater than version `2.9`  
+
+**Note:** It is recommended for package authors to update ARP version info for all package versions of a package for better version mapping if this feature is to be used for a package.
```
Trenly, line 141: What if a publisher updates the ARP version but not the Marketing version? How should that be handled? I'm specifically thinking more from a manifest-creation and winget-pkgs standpoint here

doc/specs/#980 - Apps and Features entries version mapping.md
Trenly: I agree here, I think that the no-more-parts check should consider `0` the same as an undefined part in terms of equality

doc/specs/#980 - Apps and Features entries version mapping.md
JohnMcPMS: There can only be one package (Marketing) version manifest, so it would be an update to that, replacing it.  This would be the case for a bug fix only update that is intended to fully replace the existing version.

doc/specs/#980 - Apps and Features entries version mapping.md
jedieaston: I'm beginning to wonder if we need to have a `primary` boolean for this, as when I was trying to fix the table this came up too. I don't think we should assume the array is sorted properly, as the current tooling to get AppsAndFeaturesEntries just uses whatever order the table is in (so alphabetical, meaning things like "Microsoft VC Redist 2013-2015 x64" could come before "Mozilla Firefox 100.0").

JohnMcPMS: > I'm beginning to wonder if we need to have a `primary` boolean for this

We do, although I would expect that if there are multiple components listed, only the primary one needs to have the `DisplayVersion` set.  Certainly a potential pitfall though.  Having flags like primary (and others) would also enable the complex scenarios that your PR attempted to address (with the proper internal tracking of course).

doc/specs/#980 - Apps and Features entries version mapping.md
yao-msft: These examples are actually adapted from our unit tests, so these are what we are currently using for version comparison now. During parsing, version parts with 0 only get dropped. That's why 1.0.0 == 1.0. I'll update the doc to clarify that.

PR #2189, opened 2022-05-26
Addition of Initial 1.3 Schema

ryfu-msft:
Changes:

- Added Initial 1.3 schema and test files. There are no differences between 1.2 and 1.3 schema (except manifestVersion). 1.3 schema additions will be addressed in a separate PR.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2189)


src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -101,7 +101,30 @@ namespace AppInstaller::Manifest::YamlParser
...
+            case AppInstaller::Manifest::ManifestTypeEnum::DefaultLocale:
+                schemaStr = JsonSchema::LoadResourceAsString(MAKEINTRESOURCE(IDX_MANIFEST_SCHEMA_V1_3_DEFAULTLOCALE), MAKEINTRESOURCE(MANIFESTSCHEMA_RESOURCE_TYPE));
+                break;
+            case AppInstaller::Manifest::ManifestTypeEnum::Locale:
+                schemaStr = JsonSchema::LoadResourceAsString(MAKEINTRESOURCE(IDX_MANIFEST_SCHEMA_V1_3_LOCALE), MAKEINTRESOURCE(MANIFESTSCHEMA_RESOURCE_TYPE));
+                break;
+            default:
+                THROW_HR(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED));
+            }
+        }
```
lechacon, line 126: Isn't there a better way to do this with less repetition?

yao-msft (dismissed):

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
yao-msft: I guess it could be in a table if we want to reduce code repetition

yao-msft:
Do we need to hold off on merging this one until we have official v1.3 client release?

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
ryfu-msft: I refactored this method to reduce code repetition by using a table.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -99,82 +99,68 @@ namespace AppInstaller::Manifest::YamlParser
 
     Json::Value LoadSchemaDoc(const ManifestVer& manifestVersion, ManifestTypeEnum manifestType)
     {
-        std::string schemaStr;
+        int idx = 0;
```
lechacon, line 102: 1. Is there any guarantee that no resource has idx 0?
2. I couldn't find anything explicitly about the type, but I saw that it is integer and only uses 16 bytes. Maybe use `WORD` instead of `int`? In any case, I had to look up what was the type of these IDX_ macros to be sure an int was appropriate. Can you add a comment for the future?

ryfu-msft: Added comment for clarification

src/ManifestSchema/ManifestSchema.h
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 #pragma once
 
+// Do not use 0 as this will cause load schema to fail.
```
lechacon, line 5: Nit: I think it may be good to say why it would fail. Like "0 is used to represent when there is no resource". Maybe even have a `#define` for that special value.

lechacon (approved):

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -99,82 +99,68 @@ namespace AppInstaller::Manifest::YamlParser
...
+                { ManifestTypeEnum::Version, IDX_MANIFEST_SCHEMA_V1_VERSION },
+                { ManifestTypeEnum::Installer, IDX_MANIFEST_SCHEMA_V1_INSTALLER },
+                { ManifestTypeEnum::DefaultLocale, IDX_MANIFEST_SCHEMA_V1_DEFAULTLOCALE },
+                { ManifestTypeEnum::Locale, IDX_MANIFEST_SCHEMA_V1_LOCALE },
+            };
         }
         else
         {
-            schemaStr = JsonSchema::LoadResourceAsString(MAKEINTRESOURCE(IDX_MANIFEST_SCHEMA_PREVIEW), MAKEINTRESOURCE(MANIFESTSCHEMA_RESOURCE_TYPE));
+            idx = IDX_MANIFEST_SCHEMA_PREVIEW;
```
yao-msft, line 147: >idx = IDX_MANIFEST_SCHEMA_PREVIEW; [](http://example.com/codeflow?start=12&length=34)

this should be:
resourceMap = {
                { ManifestTypeEnum::Preview, IDX_MANIFEST_SCHEMA_PREVIEW },
            };

to force the manifestType to be correct from caller

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -99,82 +99,68 @@ namespace AppInstaller::Manifest::YamlParser
...
+        }
+
+        if (idx == MANIFESTSCHEMA_NO_RESOURCE)
```
yao-msft, line 150: >if (idx == MANIFESTSCHEMA_NO_RESOURCE) [](http://example.com/codeflow?start=8&length=38)

If you make the above change to make preview schema info a resource map, this check is not needed any more

Kartik29999:
Kon hai ya

PR #2211, opened 2022-06-01
Display InstallationNotes after a successful install

### LINKED ISSUE #607: Add a notes/info field to manifests.

An info/comment box to be added to manifests, that would be shown upon package installation.

I want a space in a manifest where I can put a note that will be printed to the terminal upon the completion of package installation (or before, I suppose it doesn't matter). It doesn't need to be more than 240 characters (maybe even less than that). Just enough space to say "Google Chrome manages its own updates." or "The Windows ADK is required for most of the functionality in MDT to work. Install via winget install adk". 

Chocolatey, npm, and scoop all have something similar. It may be good to have a flag in settings (or maybe just --skip-notes) to suppress it if people don't want to see them, but the character limit should be enough to make sure people don't put "jedieaston made this package!!1! check out my really cool medium blog here!" or something. And when the packages are checked into the repository, the maintainers could see them anyway and could filter.

ryfu-msft:
Resolves #607

- Adds support for displaying the installation notes after a successful installation
- Adds the `--display-notes` flag and the `--suppress-notes` flag to control the default behavior of displaying the install notes
- Adds the setting `suppressInstallNotes` which is set to false by default 
- Adds a unit test to verify that installationNotes are displayed when the `--display-notes` flag is present

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2211)

### DIFF
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -102,6 +102,11 @@
           "type": "boolean",
           "default": false
         },
+        "suppressInstallNotes": {
+          "description": "Controls whether installation notes are suppressed after a successful install",
+          "type": "boolean",
+          "default":  false
+        },
         "PortablePackageUserRoot": {
           "description": "The default root directory where packages are installed to under User scope. Applies to the portable installer type.",
           "type": "string",
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -97,6 +97,10 @@ namespace AppInstaller::CLI
             return Argument{ "wait", NoAlias, Args::Type::Wait, Resource::String::WaitArgumentDescription, ArgumentType::Flag, false };
         case Args::Type::ProductCode:
             return Argument{ "product-code", NoAlias, Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
+        case Args::Type::DisplayNotes:
+            return Argument{ "display-notes", NoAlias, Args::Type::DisplayNotes, Resource::String::DisplayNotesArgumentDescription, ArgumentType::Flag, false };
+        case Args::Type::SuppressNotes:
+            return Argument{ "suppress-notes", NoAlias, Args::Type::SuppressNotes, Resource::String::SuppressNotesArgumentDescription, ArgumentType::Flag, false };
         default:
             THROW_HR(E_UNEXPECTED);
         }
--- a/src/AppInstallerCLICore/Commands/InstallCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/InstallCommand.cpp
@@ -46,6 +46,8 @@ namespace AppInstaller::CLI
             Argument::ForType(Args::Type::CustomHeader),
             Argument::ForType(Args::Type::AcceptSourceAgreements),
             Argument::ForType(Args::Type::Rename),
+            Argument::ForType(Args::Type::DisplayNotes),
+            Argument::ForType(Args::Type::SuppressNotes),
         };
     }
 
@@ -136,6 +138,11 @@ namespace AppInstaller::CLI
             {
                 throw CommandException(Resource::String::InvalidArgumentValueErrorWithoutValidValues, Argument::ForType(Args::Type::Locale).Name(), {});
             }
+        }
+
+        if (execArgs.Contains(Execution::Args::Type::DisplayNotes) && execArgs.Contains(Execution::Args::Type::SuppressNotes))
+        {
+            throw CommandException(Resource::String::BothDisplayAndSuppressNotesFlagsProvided, "");
         }
     }
 
--- a/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
@@ -24,6 +24,7 @@ using namespace AppInstaller::Manifest;
 using namespace AppInstaller::Repository;
 using namespace AppInstaller::Settings;
 using namespace AppInstaller::Utility;
+using namespace AppInstaller::Utility::literals;
 
 
 namespace AppInstaller::CLI::Workflow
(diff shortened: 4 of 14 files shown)


schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -102,6 +102,11 @@
           "type": "boolean",
           "default": false
         },
+        "suppressInstallNotes": {
```
lechacon, line 105: Not sure it applies here, but I remember seeing recommendations of using only positive sentences to avoid confusion with double negatives. Like "do show notes" instead of "don't suppress notes".

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -97,6 +97,10 @@ namespace AppInstaller::CLI
             return Argument{ "wait", NoAlias, Args::Type::Wait, Resource::String::WaitArgumentDescription, ArgumentType::Flag, false };
         case Args::Type::ProductCode:
             return Argument{ "product-code", NoAlias, Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
+        case Args::Type::DisplayNotes:
+            return Argument{ "display-notes", NoAlias, Args::Type::DisplayNotes, Resource::String::DisplayNotesArgumentDescription, ArgumentType::Flag, false };
```
lechacon, line 101: Could we ever want to add another type of notes that may make this ambiguous? Say, if we wanted to show both installation notes and change notes.

src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -137,6 +139,11 @@ namespace AppInstaller::CLI
                 throw CommandException(Resource::String::InvalidArgumentValueErrorWithoutValidValues, Argument::ForType(Args::Type::Locale).Name(), {});
             }
         }
+
+        if (execArgs.Contains(Execution::Args::Type::DisplayNotes) && execArgs.Contains(Execution::Args::Type::SuppressNotes))
```
lechacon, line 143: Do we really need two arguments for this? I know for sure in PS we would do it as a single switch to which you could pass `$true` or `$false` to specify, but I'm not sure here.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -482,6 +499,7 @@ namespace AppInstaller::CLI::Workflow
                 }
                 installContext << Workflow::DownloadInstaller;
                 installContext << Workflow::InstallPackageInstaller;
```
lechacon, line 501: Nit: Why is this not all chained?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -423,7 +439,8 @@ namespace AppInstaller::CLI::Workflow
     {
         context <<
             Workflow::DownloadSinglePackage <<
-            Workflow::InstallPackageInstaller;
+            Workflow::InstallPackageInstaller <<
+            Workflow::DisplayInstallationNotes;
```
lechacon, line 443: Why not make it part of `InstallPackageInstaller`?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1337,4 +1337,16 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="UnableToPurgeInstallDirectory" xml:space="preserve">
     <value>Cannot purge install directory, as it was not created by WinGet</value>
   </data>
+  <data name="BothDisplayAndSuppressNotesFlagsProvided" xml:space="preserve">
+    <value>Both `display-notes` and `suppress-notes` arguments are provided</value>
```
lechacon, line 1341: Add Comment so that the args are not localized, like `{Locked="display-notes"}` (not sure how to lock two things)

src/AppInstallerCLICore/Argument.cpp
ryfu-msft: I would imagine any other type of notes that we want to optionally display could be grouped together and controlled by these flags.

JohnMcPMS: Do we even really need a command line argument for this?  I don't see it being desired enough to warrant adding it.  Maybe if someone complains we could do so later, but for now I would not.

src/AppInstallerCLICore/Commands/InstallCommand.cpp
JohnMcPMS: Repeating myself, but I don't think we need any arguments for this.

src/AppInstallerCLICore/Argument.cpp
ryfu-msft: Removed added command line arguments

src/AppInstallerCLICore/Commands/InstallCommand.cpp
ryfu-msft: Removed arguments

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Moved to `InstallPackageInstaller`

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Chained tasks

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: Removed localized string as its no longer needed.

ryfu-msft:
> We should consider if `show` should also display the notes. I feel like it probably should. Has this already been discussed @denelon?

Hasn't been discussed yet, but I agree with this. Added functionality to display installationNotes for show command

JohnMcPMS (approved):

PR #2213, opened 2022-06-02
Use AppsAndFeaturesEntries DisplayVersion info for installed package version mapping

yao-msft:
For #980 

## Changes
Described in [spec](https://github.com/microsoft/winget-cli/blob/master/doc/specs/%23980%20-%20Apps%20and%20Features%20entries%20version%20mapping.md)

## Validation
Added unit and e2e tests. Did a local manual validation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2213)


src/AppInstallerRepositoryCore/Microsoft/Schema/1_0/ManifestTable.cpp
```diff
@@ -83,17 +83,25 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_0
         SQLite::Statement ManifestTableGetValuesById_Statement(
             const SQLite::Connection& connection,
             SQLite::rowid_t id,
-            std::initializer_list<SQLite::Builder::QualifiedColumn> columns)
+            std::initializer_list<SQLite::Builder::QualifiedColumn> columns,
+            std::initializer_list<std::string_view> manifestColumnNames)
         {
+            THROW_HR_IF(E_UNEXPECTED, manifestColumnNames.size() != 0 && manifestColumnNames.size() != columns.size());
```
JohnMcPMS, line 89:
```suggestion
            THROW_HR_IF(E_UNEXPECTED, manifestColumnNames.size() != columns.size());
```
I don't think this code works if `manifestColumnNames` is not provided along with `columns`.

src/WinGetUtil/Exports.cpp
```diff
@@ -247,6 +255,106 @@ extern "C"
     }
     CATCH_RETURN()
 
+    DEFINE_ENUM_FLAG_OPERATORS(WinGetValidateManifestResult);
+
+    WINGET_UTIL_API WinGetValidateManifestV3(
```
JohnMcPMS, line 260: Add result of discussion `WinGetUtil.h`...

Should we make this a little more complicated from the number of APIs, but less complicated for each?

Imagine:
```
WinGetCreateManifest(path, message, merge path, option, manifest handle);
WinGetValidateManifestV3(manifest handle, index, result, message, operation, option);
WinGetCloseManifest(manifest handle);
```

The create is basically V2 validate, but we return a pointer to reuse for future validations.  It performs the direct manifest validations only.  The returned pointer saves continually parsing the manifest and enables efficient splitting of the various validations we perform.  Then the V3 validate can be "perform this one validation".

src/WinGetUtil/Exports.cpp
```diff
@@ -173,7 +174,14 @@ extern "C"
...
         THROW_HR_IF(E_INVALIDARG, !succeeded);
 
-        bool result = reinterpret_cast<SQLiteIndex*>(index)->CheckConsistency(true);
+        auto sqliteIndex = reinterpret_cast<SQLiteIndex*>(index);
+        bool result = sqliteIndex->CheckConsistency(true);
+
+        // Check arp version ranges do not overlap
+        if (result && sqliteIndex->GetVersion() >= Schema::Version{ 1, 5 })
+        {
+            result = ValidateArpVersionConsistency(sqliteIndex);
```
JohnMcPMS, line 183: Why is this not part of `CheckConsistency`?

src/WinGetUtil/WinGetUtil.h
```diff
@@ -24,6 +24,50 @@ extern "C"
...
+    enum WinGetValidateManifestOptionV2
+    {
+        // Below options define what validations to be performed
+ 
+        // No validation, caller will get E_INVALIDARG
+        None = 0,
+        // Only validate against json schema
+        SchemaValidation = 0x1,
+        // Validate against schema and also perform semantic validation
+        SchemaAndSemanticValidation = 0x2,
```
JohnMcPMS, line 36: Shouldn't this just be `SemanticValidation` and you pass both flags to get both?

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -51,6 +51,9 @@ namespace AppInstaller::Manifest
         const char* const ExceededAppsAndFeaturesEntryLimit = "Only zero or one entry for Apps and Features may be specified for InstallerType portable.";
         const char* const ExceededCommandsLimit = "Only zero or one value for Commands may be specified for InstallerType portable.";
         const char* const ScopeNotSupported = "Scope is not supported for InstallerType portable.";
+        const char* const ApproximateVersionNotAllowed = "Approximate version not allowed.";
+        const char* const ArpVersionOverlapWithIndex = "DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index: ";
```
Trenly, line 55: This got me thinking - Are version ranges architecture specific, and/or should they be? I could see a scenario where a publisher uses the following - 
```yaml
PackageVersion: 1.0-Alpha # This is their marketing version
Installers: 
  - Architecture: x86
    AppsAndFeaturesEntries:
      - DisplayVersion: 1.0.4.0
- Architecture: x64
  AppsAndFeaturesEntries:
    - DisplayVersion: 1.0.5.0
```
According to the spec, this would result in an Version Range of [1.0.4.0, 1.0.5.0]

Suppose they release a new update - 
```yaml
PackageVersion: 1.1-Alpha # This is their marketing version
Installers: 
  - Architecture: x86
    AppsAndFeaturesEntries:
      - DisplayVersion: 1.0.4.1
- Architecture: x64
  AppsAndFeaturesEntries:
    - DisplayVersion: 1.0.5.1
```
According to the logic, wouldn't 1.0.4.1 overlap with [1.0.4.0, 1.0.5.0]. since 1.0.4.0 =< 1.0.4.1 =< 1.0.5.0, thereby triggering a false overlap warning?

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -51,6 +51,9 @@ namespace AppInstaller::Manifest
         const char* const ExceededAppsAndFeaturesEntryLimit = "Only zero or one entry for Apps and Features may be specified for InstallerType portable.";
         const char* const ExceededCommandsLimit = "Only zero or one value for Commands may be specified for InstallerType portable.";
         const char* const ScopeNotSupported = "Scope is not supported for InstallerType portable.";
+        const char* const ApproximateVersionNotAllowed = "Approximate version not allowed.";
+        const char* const ArpVersionOverlapWithIndex = "DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index: ";
+        const char* const ArpVersionValidationInternalError = "Encountered internal error during DisplayVersion validation against index.";
```
Trenly, line 56:
```suggestion
        const char* const ArpVersionValidationInternalError = "Internal error while validating DisplayVersion against index.";
```
Nit: Passive voice

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
yao-msft: Correct, this will be treated as unsupported case according to the spec previously written. Since there's no restriction on what is written to DisplayVersion entry, an installer could also write different values for different locale. The above example can be extended to scope, installertype, etc. Essentially this will end up being each installer is in its own set.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -121,55 +121,77 @@ namespace AppInstaller::Repository
 
         std::string GetMappedInstalledVersion(const std::string& installedVersion, const std::shared_ptr<IPackage>& availablePackage)
         {
-            // Construct a map between manifest version and arp version range. The map is ordered in descending by package version.
-            std::vector<std::pair<std::string, Utility::VersionRange>> arpVersionMap;
+            // Stores raw versions value strings to run a preliminary check whether version mapping is needed.
+            std::vector<std::tuple<std::string, std::string, std::string>> rawVersionValues;
             auto versionKeys = availablePackage->GetAvailableVersionKeys();
```
JohnMcPMS, line 126: I've just realized that if the available package is a composite of multiple sources, then this is flawed.  Do we do that now or not?

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -153,7 +190,7 @@ namespace AppInstaller::Utility
 
     bool Version::IsUnknown() const
     {
-        return (m_parts.size() == 1 && m_parts[0].Integer == 0 && Utility::CaseInsensitiveEquals(m_parts[0].Other, s_Version_Part_Unknown));
+        return (m_approximateComparator == ApproximateComparator::None && IsBaseVersionUnknown());
```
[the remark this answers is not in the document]
yao-msft: Disabled approximate version for Unknown per discussion, I kept the IsBase* methods to keep parity

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.cpp
```diff
@@ -254,6 +255,12 @@ namespace AppInstaller::Repository::Microsoft
 
         bool result = m_interface->CheckConsistency(m_dbconn, log);
 
+        // Check arp version ranges do not overlap
+        if ((result || log) && GetVersion() >= Schema::Version{ 1, 5 })
```
JohnMcPMS, line 259: I think that this should be done in the versioned code.  Why is it not?

src/WinGetUtil/Exports.cpp
```diff
@@ -255,66 +249,102 @@ extern "C"
     }
     CATCH_RETURN()
 
+    WINGET_UTIL_API WinGetCreateManifest(
+        WINGET_STRING inputPath,
+        BOOL* succeeded,
+        WINGET_SQLITE_MANIFEST_HANDLE* manifest,
```
JohnMcPMS, line 255:
```suggestion
        WINGET_MANIFEST_HANDLE* manifest,
```

src/WinGetUtil/Exports.cpp
```diff
@@ -255,66 +249,102 @@ extern "C"
...
+        
         *result = WinGetValidateManifestResult::InternalError;
 
         std::string validationMessage;
         auto validationResult = WinGetValidateManifestResult::Success;
 
-        if (WI_IsFlagSet(option, WinGetValidateManifestOptionV2::SchemaValidation) || WI_IsFlagSet(option, WinGetValidateManifestOptionV2::SchemaAndSemanticValidation))
+        Manifest* manifestPtr = reinterpret_cast<Manifest*>(manifest);
+        SQLiteIndex* sqliteIndex = nullptr;
+        if (index)
```
JohnMcPMS, line 332: You can just `reinterpret_cast` without the null check.

src/WinGetUtil/WinGetUtil.h
```diff
@@ -150,6 +156,33 @@ extern "C"
...
+        WINGET_SQLITE_MANIFEST_HANDLE* manifest,
+        WINGET_STRING_OUT* message,
+        WINGET_STRING mergedManifestPath,
+        WinGetCreateManifestOption option);
+
+    // Validates a given manifest. Returns a bool for validation result and
+    // a string representing validation errors if validation failed.
+    // If mergedManifestPath is provided, this method will write a merged manifest
+    // to the location specified by mergedManifestPath
+    WINGET_UTIL_API WinGetValidateManifestV2(
```
JohnMcPMS, line 175:
```suggestion
    WINGET_UTIL_API WinGetValidateManifestV3(
```
And change the parameters to match.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -122,8 +130,19 @@ namespace AppInstaller::Utility
             // else parts are equal, so continue to next part
         }
 
-        // All parts tested were equal, so this is only less if there are more parts in other.
-        return m_parts.size() < other.m_parts.size();
+        // All parts tested were equal
```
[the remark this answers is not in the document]
yao-msft: yep, and unit tests failed for this... rushing to get a commit run in the pipeline before running tests locally

src/AppInstallerRepositoryCore/CompositeSource.cpp
yao-msft: I think composite available package may need more thinking, currently we stop when we find one from a source, I'll add a comment as a reminder when we enable composite available package one day

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -448,6 +448,18 @@ namespace AppInstaller::Manifest
...
+    bool DoesInstallerTypeSupportArpVersionRange(InstallerTypeEnum installerType)
+    {
+        return (
+            installerType == InstallerTypeEnum::Exe ||
+            installerType == InstallerTypeEnum::Inno ||
+            installerType == InstallerTypeEnum::Msi ||
+            installerType == InstallerTypeEnum::Nullsoft ||
+            installerType == InstallerTypeEnum::Wix ||
+            installerType == InstallerTypeEnum::Burn
+            );
```
Trenly, line 460: Is there any advantage to using `||`'s here, instead of a switch statement? I would think that when compiled, a switch statement would be slightly more efficient, given that it compiles as a jump table rather than a sequential evaluation

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -275,4 +355,77 @@ namespace AppInstaller::Utility
...
     }
+
+    VersionRange::VersionRange(Version minVersion, Version maxVersion)
+    {
+        THROW_HR_IF(E_INVALIDARG, minVersion > maxVersion);
+        m_minVersion = std::move(minVersion);
+        m_maxVersion = std::move(maxVersion);
+    }
+
+    bool VersionRange::HasOverlapWith(const VersionRange& other) const
```
Trenly, line 366:
```suggestion
    bool VersionRange::Overlaps(const VersionRange& other) const
```
Why use many word, when few word do trick?

src/AppInstallerCommonCore/Public/AppInstallerVersions.h
```diff
@@ -75,9 +91,48 @@ namespace AppInstaller::Utility
         // Gets the part breakdown for a given version; used for tests.
         const std::vector<Part>& GetParts() const { return m_parts; }
 
+        // Returns if the version is an approximate version.
+        bool IsApproximateVersion() const { return m_approximateComparator != ApproximateComparator::None; }
```
Trenly, line 95:
```suggestion
        bool IsApproximate() const { return m_approximateComparator != ApproximateComparator::None; }
```
Why use many word, when few word do trick?

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -44,25 +82,36 @@ namespace AppInstaller::Utility
...
+        {
+            // If only one is latest, this can only be less than if the other is and this is not.
             return (otherIsLatest && !thisIsLatest);
         }
 
         // Sort Unknown lower than any known values
-        bool thisIsUnknown = IsUnknown();
-        bool otherIsUnknown = other.IsUnknown();
+        bool thisIsUnknown = IsBaseVersionUnknown();
+        bool otherIsUnknown = other.IsBaseVersionUnknown();
```
Trenly, line 107: Is there a reason to change this, since `IsUnknown()` calls `IsBaseVersionUnkown()`

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -275,4 +355,77 @@ namespace AppInstaller::Utility
...
+        // No overlap if either is an empty range.
+        if (IsEmpty() || other.IsEmpty())
+        {
+            return false;
+        }
+
+        return m_minVersion <= other.m_maxVersion && m_maxVersion >= other.m_minVersion;
+    }
+
+    bool VersionRange::IsSameAsSingleVersion(const Version& version) const
```
Trenly, line 377: It seems like this could be an operator overload of `==` instead of a separate method?

yao-msft: I don't feel like doing `==` for different objects, which may cause confusion IMO.

src/AppInstallerCommonCore/Versions.cpp
yao-msft: In a previous commit, approximate to Unknown is actually supported. I disabled it after internal discussion about what we would do to Unknown. I kept it for parity. And in case we need it in the future.

JohnMcPMS (approved):

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
yao-msft: I just copied from above similar functions. I guess it's more readable. In the end, this function will be invoked at most once for each installed package so I think I'll leave it for consistency for now.

PR #2215, opened 2022-06-03
Validate manifest installers (MSIX)

AmelBawa-msft:
## What was done?
- Added msix and msixbundle manifest validation
  - Validate package family name.
  - Validate package version.
  - Validate Minimum OS Version.

## Test
- Added test installers (msix and msixbundle) to test the validation functionality
- Added manifest installer validation tests
- Added installer package fields reader tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2215)


src/AppInstallerCommonCore/Public/winget/MsixManifest.h
```diff
@@ -0,0 +1,60 @@
...
+#include <wrl/client.h>
+
+#include <string>
+
+namespace AppInstaller::Msix
+{
+    using namespace Microsoft::WRL;
+    using string_t = Utility::NormalizedString;
+
+    typedef Utility::FourPartsVersionNumber PackageVersion;
```
lechacon, line 17: Why not `using PackageVersion = Utility::FourPartsVersionNumber`?

AmelBawa-msft: Updated

src/AppInstallerCommonCore/Public/winget/MsixManifest.h
```diff
@@ -0,0 +1,60 @@
...
+#include <AppxPackaging.h>
+#include <AppInstallerStrings.h>
+#include <AppInstallerVersions.h>
+#include <wrl/client.h>
+
+#include <string>
+
+namespace AppInstaller::Msix
+{
+    using namespace Microsoft::WRL;
```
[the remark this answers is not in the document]
AmelBawa-msft: Oop,s moved to source

src/AppInstallerCommonCore/Public/winget/MsixManifest.h
```diff
@@ -0,0 +1,60 @@
...
+    using string_t = Utility::NormalizedString;
+
+    typedef Utility::FourPartsVersionNumber PackageVersion;
+    typedef Utility::FourPartsVersionNumber OSVersion;
+
+    // Package identity for a msix manifest
```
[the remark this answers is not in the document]
AmelBawa-msft: Updated comments

src/AppInstallerCommonCore/MsixManifest.cpp
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        return minOSVersion;
+    }
+
+    std::vector<MsixPackageManifestTargetDeviceFamily> MsixPackageManifest::GetTargetDeviceFamilies() const
+    {
+        std::vector<MsixPackageManifestTargetDeviceFamily> targetDeviceFamilies;
+        ComPtr<IAppxManifestReader3> manifest3;
+        if (SUCCEEDED(m_manifestReader->QueryInterface(IID_PPV_ARGS(&manifest3))))
```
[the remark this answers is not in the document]
AmelBawa-msft: Replaced with `THROW_IF_FAILED`

src/AppInstallerCommonCore/Public/winget/MsixManifest.h
```diff
@@ -0,0 +1,60 @@
...
+        MsixPackageManifest(ComPtr<IAppxManifestReader> manifestReader);
+
+        std::vector<MsixPackageManifestTargetDeviceFamily> GetTargetDeviceFamilies() const;
+        MsixPackageManifestIdentity GetIdentity() const;
+        std::optional<OSVersion> GetMinimumOSVersion() const;
+    private:
+        ComPtr<IAppxManifestReader> m_manifestReader;
+    };
+
+    class MsixPackageManifestManager
```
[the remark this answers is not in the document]
AmelBawa-msft: Added comment and updated class name to reflect its purpose more accurately

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -275,4 +275,35 @@ namespace AppInstaller::Utility
 
         return m_version < other.m_version;
     }
+
+    FourPartsVersionNumber::FourPartsVersionNumber(UINT64 version)
+    {
+        UINT64 mask16 = (1 << 16) - 1;
```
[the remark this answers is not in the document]
AmelBawa-msft: Updated

yao-msft:
To fix the spelling, you can add .msixbundle type to the excludes.txt under spelling.

src/WinGetUtil/WinGetUtil.h
```diff
@@ -22,6 +22,7 @@ extern "C"
         Default = 0,
         SchemaValidationOnly = 0x1,
         ErrorOnVerifiedPublisherFields = 0x2,
+        InstallerValidation = 0x4,
```
yao-msft, line 25: >InstallerValidation = 0x4, [](http://example.com/codeflow?start=8&length=26)

nit: if you want to keep this option under WinGetValidateManifestOption, can you name it InstallerValidations to avoid name confliction? Or if we're to remove this after V3 manifest validation integration, you can ignore this comment

src/AppInstallerCommonCore/Public/winget/MsixManifest.h
```diff
@@ -0,0 +1,61 @@
...
+
+        std::vector<MsixPackageManifestTargetDeviceFamily> GetTargetDeviceFamilies() const;
+        MsixPackageManifestIdentity GetIdentity() const;
+        std::optional<OSVersion> GetMinimumOSVersion() const;
+    private:
+        Microsoft::WRL::ComPtr<IAppxManifestReader> m_manifestReader;
+    };
+
+    // MSIX manifest cache
+    class MsixPackageManifestCache
```
yao-msft, line 53: >class [](http://example.com/codeflow?start=4&length=5)

nit: throughout winget code, we use struct and declare private fields

src/AppInstallerCLITests/TestData/Manifest-Missing-MsixInstallerFields.yaml
```diff
@@ -0,0 +1,13 @@
+PackageIdentifier: AppInstallerCliTest.GoodMsixInstaller
```
yao-msft, line 1: >GoodMsixInstaller [](http://example.com/codeflow?start=39&length=17)

I think this should be updated to avoid confusion. Also to keep file naming consistency, this file would be named Manifest-Bad-Missing*Fields.yaml

src/AppInstallerCLITests/TestCommon.h
```diff
@@ -135,4 +136,7 @@ namespace TestCommon
     // tests calling these functions should skip when not running with admin.
     bool InstallCertFromSignedPackage(const std::filesystem::path& package);
     bool UninstallCertFromSignedPackage(const std::filesystem::path& package);
+
+    // Get manifest reader from a msix file path
+    bool GetMsixPackageManifestReader(const std::string_view testFileName, IAppxManifestReader** manifestReader);
```
yao-msft, line 141: >const  [](http://example.com/codeflow?start=38&length=6)

nit: remove const

src/AppInstallerCommonCore/MsixManifest.cpp
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        return minOSVersion;
+    }
+
+    std::vector<MsixPackageManifestTargetDeviceFamily> MsixPackageManifest::GetTargetDeviceFamilies() const
+    {
+        std::vector<MsixPackageManifestTargetDeviceFamily> targetDeviceFamilies;
+        ComPtr<IAppxManifestReader3> manifest3;
+        THROW_IF_FAILED(m_manifestReader->QueryInterface(IID_PPV_ARGS(&manifest3)));
```
yao-msft, line 74:
```suggestion
        THROW_IF_FAILED(m_manifestReader.As(&manifestReader3));
```

src/AppInstallerCommonCore/MsixManifest.cpp
```diff
@@ -0,0 +1,123 @@
...
+        THROW_IF_FAILED(manifest3->GetTargetDeviceFamilies(&targetDeviceFamiliesIter));
+
+        BOOL hasCurrent = FALSE;
+        THROW_IF_FAILED(targetDeviceFamiliesIter->GetHasCurrent(&hasCurrent));
+        while (hasCurrent)
+        {
+            ComPtr<IAppxManifestTargetDeviceFamily> targetDeviceFamily;
+            THROW_IF_FAILED(targetDeviceFamiliesIter->GetCurrent(&targetDeviceFamily));
+
+            targetDeviceFamilies.emplace_back(targetDeviceFamily);
```
yao-msft, line 86:
```suggestion
            targetDeviceFamilies.emplace_back(std::move(targetDeviceFamily));
```

src/AppInstallerCommonCore/MsixManifest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    string_t MsixPackageManifestTargetDeviceFamily::GetName() const
+    {
+        wil::unique_cotaskmem_string name;
+        THROW_IF_FAILED(m_targetDeviceFamily->GetName(&name));
+        return Utility::ConvertToUTF8(name.get());
+    }
+
+    OSVersion MsixPackageManifestTargetDeviceFamily::GetMinVersion() const
+    {
+        UINT64 minVersion;
```
yao-msft, line 103:
```suggestion
        UINT64 minVersion = 0;
```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -234,4 +236,104 @@ namespace AppInstaller::Manifest
...
+        std::vector<Msix::MsixPackageManifest> msixManifests;
+        try {
+            msixManifests = msixManifestsCache.GetAppPackageManifests(installer.Url);
+        }
+        catch (...)
+        {
+            errors.emplace_back(ManifestError::InstallerFailedToProcess, "InstallerUrl", installer.Url);
+        }
+
+        for (auto msixManifest : msixManifests)
```
yao-msft, line 271: Actually, msixbundle has checks to make sure each payload msix within declare exactly the same core entries(including identity, package version, TDFs, etc) in the msix manifest

src/AppInstallerCommonCore/MsixManifest.cpp
```diff
@@ -0,0 +1,122 @@
...
+            auto targetDeviceFamilyMinOSVersion = targetDeviceFamily.GetMinVersion();
+            if (!minOSVersion.has_value())
+            {
+                minOSVersion = targetDeviceFamilyMinOSVersion;
+            }
+            else if(targetDeviceFamilyMinOSVersion < minOSVersion.value())
+            {
+                minOSVersion = targetDeviceFamilyMinOSVersion;
+            }
+        }
```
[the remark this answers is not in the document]
AmelBawa-msft: My only concern for changing this is that `wingetcreate` could potentially generate invalid yaml manifests as the current implementation from [Msix-packaging](https://github.com/microsoft/msix-packaging/blob/88a81569aaee456883e7233051b35524f9d5a11e/tools/utils/Utils/AppxPackaging/AppxMetadata.cs#L119) considers the min value among all the target device family records. Here's an example of a [package](https://github.com/microsoft/winget-pkgs/blob/2cb0674ddb123a289d46dd63bf7951dd72228cb0/manifests/b/bitmediae-SolutionsGmbH/ITSClient/3.0.0.0/bitmediae-SolutionsGmbH.ITSClient.installer.yaml#L9) where `wingetcreate` would set the `MinimumOSVersion` to `6.1.7601.0`, and the validate command (if updated) would expect `10.0.16299.0`.

Here are the dependencies from the package:
```xml
  <Dependencies>
    <TargetDeviceFamily Name="MSIXCore.Desktop" MinVersion="6.1.7601.0" MaxVersionTested="10.0.10240.0" />
    <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.16299.0" MaxVersionTested="10.0.18362.30" />
  </Dependencies>
```

Thoughts?

src/WinGetUtil/WinGetUtil.h
AmelBawa-msft: Updated to `InstallerValidations` to prevent conflicts 👍

src/AppInstallerCLITests/TestData/Manifest-Missing-MsixInstallerFields.yaml
AmelBawa-msft: Goot catch, updated and renamed

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -275,4 +275,35 @@ namespace AppInstaller::Utility
...
+
+    FourPartsVersionNumber::FourPartsVersionNumber(UINT64 version)
+    {
+        const UINT64 mask16 = (1 << 16) - 1;
+        UINT64 revision = version & mask16;
+        UINT64 build = (version >> 0x10) & mask16;
+        UINT64 minor = (version >> 0x20) & mask16;
+        UINT64 major = (version >> 0x30) & mask16;
+
+        std::stringstream ssVersion;
```
JohnMcPMS, line 287: Why build a string just so it can be converted back to numbers?

src/AppInstallerCLITests/Versions.cpp
```diff
@@ -191,3 +191,40 @@ TEST_CASE("VersionUnknownLessThanLatest", "[versions]")
...
+    REQUIRE(expectedVersion == versionNumberFromNumber);
+    REQUIRE(expectedVersion == versionNumberFromString);
+    REQUIRE(expectedVersion.ToString() == versionNumberFromNumber.ToString());
+    REQUIRE(expectedVersion.ToString() == versionNumberFromString.ToString());
+}
+
+TEST_CASE("FourPartsVersionNumber_Success_LessThanFourParts", "[versions]")
+{
+    FourPartsVersionNumber versionNumberFromNumber(0x0001000200030000);
+    FourPartsVersionNumber versionNumberFromString("1.2.3");
```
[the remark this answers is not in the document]
JohnMcPMS: `Version` has always removed trailing 0 parts, so I think this is in line with that (implied trailing 0s).

src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -365,4 +367,28 @@ namespace AppInstaller::Utility
...
+            ::IUnknown* rasAsIUnknown = (::IUnknown*)winrt::get_abi(randomAccessStream);
+            THROW_IF_FAILED(CreateStreamOverRandomAccessStream(
+                rasAsIUnknown,
+                IID_PPV_ARGS(inputStream.ReleaseAndGetAddressOf())));
+        }
+        else
+        {
+            std::filesystem::path path(Utility::ConvertToUTF16(uriStr));
+            THROW_IF_FAILED(SHCreateStreamOnFileEx(path.c_str(),
+                STGM_READ | STGM_SHARE_DENY_WRITE | STGM_FAILIFTHERE, 0, FALSE, nullptr, &inputStream));
```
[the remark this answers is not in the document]
JohnMcPMS: It is also really "URI [if http(s)] or file name" since `file://C:/foo.txt` won't actually work.

src/AppInstallerCommonCore/MsixManifest.cpp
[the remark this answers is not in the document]
JohnMcPMS: Update `wingetcreate`; that package won't install on Win10 pre 16299, but we currently think it will because of the bad data.

src/AppInstallerCommonCore/Public/AppInstallerDownloader.h
```diff
@@ -74,4 +74,7 @@ namespace AppInstaller::Utility
     // If IAttachmentExecute::Save is successfully invoked and the scan failed, the failure HRESULT is returned.
     // zoneIfScanFailure: URLZONE to apply if IAttachmentExecute::Save scan failed.
     HRESULT ApplyMotwUsingIAttachmentExecuteIfApplicable(const std::filesystem::path& filePath, const std::string& source, URLZONE zoneIfScanFailure);
+
+    // Function to create a stream from a uri string (url address or file system path)
+    Microsoft::WRL::ComPtr<IStream> GetStreamFromURI(std::string_view uriStr);
```
JohnMcPMS, line 79: Do we have to have `WRL::ComPtr` here?  I would prefer `wil` if possible...

src/AppInstallerCommonCore/Public/winget/MsixManifest.h
```diff
@@ -0,0 +1,61 @@
...
+    using string_t = Utility::NormalizedString;
+    using PackageVersion = Utility::FourPartsVersionNumber;
+    using OSVersion = Utility::FourPartsVersionNumber;
+
+    // Package identity for an MSIX manifest
+    struct MsixPackageManifestIdentity
+    {
+        MsixPackageManifestIdentity(Microsoft::WRL::ComPtr<IAppxManifestPackageId> packageId);
+
+        string_t GetPackageFamilyName() const;
```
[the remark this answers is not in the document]
JohnMcPMS: I don't know what characters are allowed in a PFN; if it is all of Unicode then we do.

src/AppInstallerCommonCore/Public/winget/MsixManifest.h
```diff
@@ -0,0 +1,63 @@
...
+#include <AppxPackaging.h>
+#include <AppInstallerStrings.h>
+#include <AppInstallerVersions.h>
+#include <wrl/client.h>
+
+#include <string>
+
+namespace AppInstaller::Msix
+{
+    using string_t = Utility::NormalizedString;
```
JohnMcPMS, line 14: nit: Putting these in the root of the namespace could lead to collisions if someone else `using`s this in.  Mostly a problem for something generic like `string_t`.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Versions.cpp
```diff
@@ -275,4 +275,35 @@ namespace AppInstaller::Utility
...
+        ssVersion << major
+            << Version::DefaultSplitChars << minor
+            << Version::DefaultSplitChars << build
+            << Version::DefaultSplitChars << revision;
+        Assign(ssVersion.str(), Version::DefaultSplitChars);
```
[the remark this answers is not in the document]
AmelBawa-msft: Not 100% sure how the compiler approaches this, but my guess is that we don't need to explicitly move the data here (in this case it's already an `rvalue` from the function call) and leave it for the compiler to optimize. Let me know if you have any suggestions or if you have some tests/experiments in mind you want me to execute.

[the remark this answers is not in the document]
JohnMcPMS: Looks like it isn't available until C++20:
https://en.cppreference.com/w/cpp/io/basic_stringstream/str

<html>
<body>


std::basic_string<CharT,Traits,Allocator> str() &&; | (3) | (since C++20)
-- | -- | --



</body>
</html>

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -836,4 +836,80 @@ TEST_CASE("ManifestLocalizationValidation", "[ManifestValidation]")
...
+}
+
+TEST_CASE("ReadManifestAndValidateMsixInstallers_Success", "[ManifestValidation]")
+{
+    auto testFile = TestDataFile("Manifest-Good-MsixInstaller.yaml");
+    Manifest manifest = YamlParser::CreateFromPath(testFile);
+
+    // Update the installer path for testing
+    REQUIRE(1 == manifest.Installers.size());
+    auto msixFile = TestDataFile(manifest.Installers[0].Url.c_str());
```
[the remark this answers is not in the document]
AmelBawa-msft: Compiler complains if I pass a Normalized string (`.Url` in this case).
```
no instance of constructor "TestCommon::TestDataFile::TestDataFile" matches the argument list
```

src/AppInstallerCommonCore/Public/AppInstallerVersions.h
```diff
@@ -80,6 +80,23 @@ namespace AppInstaller::Utility
         std::vector<Part> m_parts;
     };
 
+    // Four parts version number: 16-bits.16-bits.16-bits.16-bits
+    struct FourPartsVersionNumber : public Version
```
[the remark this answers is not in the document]
AmelBawa-msft: Renamed class to `UInt64Version`

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -234,4 +236,104 @@ namespace AppInstaller::Manifest
...
+            auto msixPackageFamilyName = msixManifestIdentity.GetPackageFamilyName();
+            if (!installer.PackageFamilyName.empty() && !msixPackageFamilyName.empty())
+            {
+                if (installer.PackageFamilyName != msixPackageFamilyName)
+                {
+                    errors.emplace_back(ManifestError::InstallerMsixInconsistencies, "PackageFamilyName", msixPackageFamilyName);
+                }
+            }
+            // Yaml manifest missing package family name
+            else if (!msixPackageFamilyName.empty())
```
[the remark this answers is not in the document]
JohnMcPMS: I don't think this was full addressed.  Package family name is not an optional value in the manifest, so it shouldn't be optional from `msixManifestIdentity.GetPackageFamilyName()` and thus should never be empty.  If you want to ensure that, throw in `GetPackageFamilyName` and don't have code here that makes it seem like it could be.

src/AppInstallerCommonCore/MsixManifest.cpp
```diff
@@ -42,55 +26,66 @@ namespace AppInstaller::Msix
...
+        // If target device family "Windows.Desktop" was not present,
+        // get minimum OS version from "Windows.Universal"
+        if (minOSVersion.has_value())
+        {
+            return minOSVersion.value();
+        }
+
+        // Can't get min OS version because none of the provided
+        // target device family names are supported
+        THROW_HR(E_UNEXPECTED);
```
JohnMcPMS, line 61: That isn't unexpected, there can be valid MSIX packages with this.  It should be better handled than this will result in.

The two comments on this weren't merged properly 😄 One comment was "this is not optional" and the other was "we should only care about the platforms we support".  Implementing the second comment made the value optional, so it should remain optional on the return type.  And I would change the function name to `GetMinimumOSVersionForSupportedPlatforms`.

Then the caller will need to fail with "No supported platforms" when this value is `null_opt`.

Finally, the function is implemented in a way that will make addition of another supported platform require basically rewriting it.  Why not simply find the minimum version of the supported platforms?

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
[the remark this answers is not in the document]
AmelBawa-msft: Pushed changes removing the condition checking for empty package family name from `msixManifestIdentity.GetPackageFamilyName()`

src/AppInstallerCommonCore/MsixManifest.cpp
AmelBawa-msft: Changed the function name, return value to optional, and updated the implementation to allow extending it in the future, but kept the logic to align with this [comment](https://github.com/microsoft/winget-cli/pull/2215#discussion_r890630289). Also added an additional test for "No supported platforms".

JohnMcPMS (approved):

PR #2216, opened 2022-06-03
Add support for UnsupportedArguments

### LINKED ISSUE #1277: Add the ability to identify unsupported arguments

Some installers don't support certain arguments like "--location".
See https://github.com/microsoft/winget-pkgs/issues/20228

Others may not support "--log"

Enhance the schema and the client to inform users when these are known conditions:
"--location"
"--log"

>Note: we may update the description of this issue to identify other scenarios

```YAML
Installers:
  - InstallerType: exe
    InstallerArchitecture: x64
    UnsupportedArguments:
      - Location
      - Log
```

ryfu-msft:
Resolves #1277

Changes:
- Fixes the schema as UnsupportedArguments was defined by not included in the installer object.
- Checks if arguments are provided that match the unsupportedArguments present in the manifest and shows an error message to the user.

Tests:
- Verifies that the correct error message is shown if an unsupported argument is provided

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2216)


src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -128,6 +128,35 @@ namespace AppInstaller::CLI::Workflow
...
+                {
+                    unsupportedArgFound = true;
+                    context.Reporter.Error() << Resource::String::LocationArgumentNotSupported << std::endl;
+                }
+            }
+        }
+
+        if (unsupportedArgFound)
+        {
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_UNSUPPORTED_ARGUMENT);
```
lechacon, line 156: Is there a way to make this a soft error?

As a made up scenario to justify it: I want to install 10 apps and I have a script that does `winget install $(App) --log logs\$(App).log` for each one. I don't want it to fail because I didn't know one of them didn't support it

src/AppInstallerCLICore/Workflows/InstallFlow.h
```diff
@@ -28,6 +28,12 @@ namespace AppInstaller::CLI::Workflow
     // Inputs: InstallationNotes
     // Outputs: None
     void DisplayInstallationNotes(Execution::Context& context);
+    
+    // Checks if there are any included arguments that are not supported for the package.
+    // Required Args: None
+    // Inputs: UnsupportedArguments
```
lechacon, line 34: Installer?

src/AppInstallerCLITests/TestData/InstallFlowTest_UnsupportedArguments.yaml
```diff
@@ -0,0 +1,18 @@
...
+Moniker: AICLITestExe
+License: Test
+UnsupportedArguments:
+  - log
+  - location
+Installers:
+    - Architecture: x86
+      InstallerUrl: https://ThisIsNotUsed
+      InstallerType: exe
+      InstallerSha256: 65DB2F2AC2686C7F2FD69D4A4C6683B888DC55BFA20A0E32CA9F838B51689A3B
```
lechacon, line 16: Nit: Inconsistent indentation

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -100,6 +100,7 @@
 #define APPINSTALLER_CLI_ERROR_PORTABLE_SYMLINK_PATH_IS_DIRECTORY ((HRESULT)0x8A150055)
 #define APPINSTALLER_CLI_ERROR_INSTALLER_PROHIBITS_ELEVATION    ((HRESULT)0x8A150056)
 #define APPINSTALLER_CLI_ERROR_PORTABLE_UNINSTALL_FAILED          ((HRESULT)0x8A150057)
```
lechacon, line 102: Nit: alignment

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -100,6 +100,7 @@
 #define APPINSTALLER_CLI_ERROR_PORTABLE_SYMLINK_PATH_IS_DIRECTORY ((HRESULT)0x8A150055)
 #define APPINSTALLER_CLI_ERROR_INSTALLER_PROHIBITS_ELEVATION    ((HRESULT)0x8A150056)
 #define APPINSTALLER_CLI_ERROR_PORTABLE_UNINSTALL_FAILED          ((HRESULT)0x8A150057)
+#define APPINSTALLER_CLI_ERROR_UNSUPPORTED_ARGUMENT             ((HRESULT)0x8A150058)
```
lechacon, line 103: Can you also add it to the test constants and add the mapping to string? I think the files are something like `TestConstants.cs` and `Errors.cpp`

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1349,4 +1349,12 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="ShowLabelInstallationNotes" xml:space="preserve">
     <value>InstallationNotes:</value>
   </data>
+  <data name="LocationArgumentNotSupported" xml:space="preserve">
+    <value>The argument --location is not supported for this package</value>
+    <comment>{Locked="--location"}</comment>
+  </data>
+  <data name="LogArgumentNotSupported" xml:space="preserve">
+    <value>The argument --log is not supported for this package</value>
+    <comment>{Locked="--log"}</comment>
```
lechacon, line 1358: Could this be a single string instead? Like "This argument is not supported for this package: {arg}".

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
ryfu-msft: No longer applicable because I removed the error

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: Changed to suggested

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
ryfu-msft: Fixed Alignment

src/AppInstallerCLITests/TestData/InstallFlowTest_UnsupportedArguments.yaml
ryfu-msft: Fixed indentation

src/AppInstallerCLICore/Workflows/InstallFlow.h
ryfu-msft: Added Installer to comments

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -128,6 +141,22 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
+    void CheckForUnsupportedArgs(Execution::Context& context)
+    {
+        const auto& unsupportedArgs = context.Get<Execution::Data::Installer>().value().UnsupportedArguments;
```
lechacon, line 146: Nit: Why `.value().` instead of `->`?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -128,6 +141,22 @@ namespace AppInstaller::CLI::Workflow
...
+    {
+        const auto& unsupportedArgs = context.Get<Execution::Data::Installer>().value().UnsupportedArguments;
+
+        if (!unsupportedArgs.empty())
+        {
+            for (auto unsupportedArg : unsupportedArgs)
+            {
+                if (context.Args.Contains(GetUnsupportedArgumentType(unsupportedArg)))
+                {
+                    context.Reporter.Error() << Resource::String::UnsupportedArgument << " --" << UnsupportedArgumentToString(unsupportedArg) << std::endl;
```
lechacon, line 154: Nit: I'd make this a warning instead of an error

```suggestion
                    context.Reporter.Warn() << Resource::String::UnsupportedArgument << " --"_liv << UnsupportedArgumentToString(unsupportedArg) << std::endl;
```

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -128,6 +141,22 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
+    void CheckForUnsupportedArgs(Execution::Context& context)
+    {
+        const auto& unsupportedArgs = context.Get<Execution::Data::Installer>().value().UnsupportedArguments;
+
+        if (!unsupportedArgs.empty())
```
lechacon, line 148: I'm pretty sure you don't need to check for empty and can just let the `for` do its thing.

src/AppInstallerCLICore/Workflows/InstallFlow.h
```diff
@@ -28,6 +28,12 @@ namespace AppInstaller::CLI::Workflow
     // Inputs: InstallationNotes
     // Outputs: None
     void DisplayInstallationNotes(Execution::Context& context);
+    
+    // Checks if there are any included arguments that are not supported for the package.
+    // Required Args: None
+    // Inputs: Installer, UnsupportedArguments
```
lechacon, line 34: I always took these comments as talking about what data is in the context. Like "this function will call `context.get<Input>()` and after it's done you can do `context.get<Output>()`". So I'd only list `Installer` and not `UnsupportedArguments` because there isn't an entry like that.

```suggestion
    // Inputs: Installer
```

src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -815,6 +815,29 @@ TEST_CASE("InstallFlow_InstallationNotes", "[InstallFlow][workflow]")
...
+    context.Args.AddArg(Execution::Args::Type::InstallLocation);
+
+    InstallCommand install({});
+    install.Execute(context);
+    INFO(installOutput.str());
+
+    // Verify Unsupported arguments error message is shown
+    REQUIRE(context.GetTerminationHR() == S_OK);
+    REQUIRE(std::filesystem::exists(installResultPath.GetPath()));
+    REQUIRE(installOutput.str().find(Resource::LocString(Resource::String::UnsupportedArgument).get() + " --log") != std::string::npos);
```
lechacon, line 838: Nit: I think a test to check that this is not printed when not needed would be good. E.g. if it is supported, or if it is unsupported but not used. But it's probably not needed.

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -417,6 +417,19 @@ namespace AppInstaller::Manifest
         return "Unknown"sv;
     }
 
+    std::string_view UnsupportedArgumentToString(UnsupportedArgumentEnum unsupportedArg)
+    {
+        switch (unsupportedArg)
+        {
+        case UnsupportedArgumentEnum::Log:
+            return "log"sv;
```
lechacon, line 425: Why not include the `--` so that you don't have to prepend it to print the message?

lechacon (approved):

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Changed to suggested

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Changed to suggested

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Removed empty check

src/AppInstallerCLICore/Workflows/InstallFlow.h
ryfu-msft: Changed to only listing `Installer`

src/AppInstallerCLITests/WorkFlow.cpp
ryfu-msft: Added another unit test to ensure error messages are not shown without the provided arguments

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
ryfu-msft: I decided not to include the `--` in case the actual name of the unsupported argument needed to be used elsewhere.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -128,6 +142,19 @@ namespace AppInstaller::CLI::Workflow
...
 
+    void CheckForUnsupportedArgs(Execution::Context& context)
+    {
+        const auto& unsupportedArgs = context.Get<Execution::Data::Installer>()->UnsupportedArguments;
+
+        for (auto unsupportedArg : unsupportedArgs)
+        {
+            if (context.Args.Contains(GetUnsupportedArgumentType(unsupportedArg)))
+            {
+                context.Reporter.Warn() << Resource::String::UnsupportedArgument << " --"_liv << UnsupportedArgumentToString(unsupportedArg) << std::endl;
```
Trenly, line 153: If `--log` is an unsupported argument, and a user uses `winget install <Package> -o <Path>` then the warning would be `This argument is not supported for this package: --log` which could be confusing to some users. I would suggest adding some logic to determine if its an unsupported argument or an unsupported alias which was used

JohnMcPMS, line 153: It feels like we should classify the unsupported arguments into two categories: warnings and errors.

For the two currently known values, logs would be warnings (either there is no log or we can't change where it logs).  Install location should be an error though.  If I ask for it to be installed in a specific location, but it won't be, I don't want installation to just proceed.

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -417,6 +417,19 @@ namespace AppInstaller::Manifest
         return "Unknown"sv;
     }
 
+    std::string_view UnsupportedArgumentToString(UnsupportedArgumentEnum unsupportedArg)
```
JohnMcPMS, line 420: I've been trying to call all of these `ToString` to make it easier to find them.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
Trenly: I think including the name and alias here would be just as effective of a solution

ryfu-msft: Changed the current code to find the actual `Argument` rather than just checking the `Args::Type`

ryfu-msft: Added a helper method called `ShouldErrorForUnsupportedArgument` that determines whether an error should be shown for the unsupportedArgument. If it returns false, a warning message is shown instead.

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
ryfu-msft: Removed this function as it is no longer needed.

src/AppInstallerCLICore/Argument.h
```diff
@@ -86,6 +86,9 @@ namespace AppInstaller::CLI
         // Requires that some form of package selection argument is present
         static void ValidatePackageSelectionArgumentSupplied(const Execution::Args& args);
 
+        // Gets the argument usage string in the format of "--name,-alias".
```
JohnMcPMS, line 89: Alias appears to be first in the code.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -57,6 +58,30 @@ namespace AppInstaller::CLI::Workflow
...
+            switch (arg)
+            {
+            case UnsupportedArgumentEnum::Location:
+                return true;
+            default:
+                return false;
+            }
+        }
+
+        AppInstaller::CLI::Argument GetUnsupportedArgument(UnsupportedArgumentEnum arg)
```
JohnMcPMS, line 72: I don't think this works; the command might use that `Args::Type` but have a different `Argument`.  You would need to search the commands arguments to find the one that matches.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Argument.h
ryfu-msft: Fixed order

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -252,7 +252,9 @@ namespace AppInstaller::CLI
                 context << SelectLatestApplicableUpdate(true);
             }
 
-            context << InstallSinglePackage;
+            context <<
+                CheckForUnsupportedArgs <<
```
JohnMcPMS, line 256: Why is this not just inside `InstallSinglePackage`?

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -158,18 +165,26 @@ namespace AppInstaller::CLI::Workflow
...
+            if (context.Args.Contains(unsupportedArgType))
             {
-                const auto& usageString = argument.GetUsageString();
-                if (ShouldErrorForUnsupportedArgument(unsupportedArg))
-                {
-                    context.Reporter.Error() << Resource::String::UnsupportedArgument << ' ' <<  usageString << std::endl;
-                    AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_UNSUPPORTED_ARGUMENT);
-                }
-                else
+                const auto& commandArguments = context.GetExecutingCommand()->GetArguments();
```
JohnMcPMS, line 171: You cannot assume that the executing command has been set.  I would make sure that `Resource::String::UnsupportedArgument` made sense if alone, then if the command is null just don't have a usage string after.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
ryfu-msft: Moved to `InstallSinglePackage`

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Changed `UnsupportedArgument` string so that it makes sense alone. Checks to make sure command is not null before attempting to print usage strings.

JohnMcPMS (approved):

Issue #2128, opened 2022-04-27
Breaking changes to properly justify a 2.0 release


denelon:
### Description of the new feature / enhancement

In the spirit of [semantic](https://semver.org) versioning, we want to identify issues that are potentially breaking changes to consider for a future 2.0 release. This list is **not** a commitment for what would be in a 2.0 release, but rather to identify candidates. This list may be amended over time, and any issues not taken into a 2.0 release should be moved to a similar Issue to identify a future 3.0 release.

* microsoft/winget-cli/discussions/2106
* microsoft/winget-cli/discussions/2112

Trenly:
I was looking through the list of issues and found this one which could be a consideration for a 2.0 release or to be dropped -

* https://github.com/microsoft/winget-cli/issues/429

Trenly:
[Policy] Breaking-Change

sredna:
If `DefaultInstallLocation` is ever going to be used by WinGet to find an existing install, its format should be restricted to sensible values.

My recommendation is that the pattern should be restricted to `"pattern": "^[%\\]"` which covers the following cases:
- `%ProgramFiles%\MyApp` and `%LocalAppData%\Programs\MyApp` (The recommended paths for 99% of apps)
- `\AppInRootOfDrive` (WinGet would check all local drives when a path starts with `\`)
- `\Program Files\MyApp` (Handles the case where an app hard-codes `c:\Program Files` instead of using the correct shell folder)
- `\\Computer\Share\MyApp` (This would not be allowed in the community repo. but might be useful in private sources. WinGet would treat something starting with `\\` as an absolute path)

Issue #715, opened 2021-01-23
Cannot override installer hash as admin. Even if EnableLua = 0 (UAC OFF)


MarcoPeraza:
# Cannot override installer hash when running as admin, even if EnableLua = 0 (UAC OFF)

I want to install the package. Please let me install the package.

# Steps to reproduce

Try to install a package with a mismatched hash while running as administrator.

# Expected behavior

It allows you to install.

# Actual behavior

It does not allow you to install.

jedieaston:
Did you run it with --force?

denelon:
@MarcoPeraza were you able to use --force?

MarcoPeraza:
Yes, when I use --force I get the following error still: "Installer hash does not match; this cannot be overridden when running as admin"

Note that when using an admin account and running Windows with EnableLua=0, all applications run with full privileges. This turns UAC off. The hash override should remain available to users running with this configuration. I have never encountered a feature in any Microsoft product that did not work with this configuration.

MarcoPeraza:
Hi, is there any reason why this feature is disabled at all for users running as admin? This seems to be upside-down. Usually, running as administrator gives you _more_ privileges, not _less_.

I also just want to reiterate that when you turn UAC off on a machine, it's impossible to _not_ run as admin, so the hash override feature becomes completely unavailable.

jedieaston:
If you are running as administrator and the installer on the other end is compromised, it can do whatever it wants to your system once winget runs it. If you aren't running as administrator and do --force, UAC should pop up and say "hey, this application wasn't signed by a vendor. run anyway?" 

It should be configurable via ` winget settings `, but by default winget shouldn't be allowing people to easily run unverified binaries (since if there is a hash mismatch, it means that the version that winget downloaded wasn't verified through the community repo and therefore could be a virus or something). 

> I also just want to reiterate that when you turn UAC off on a machine, it's impossible to not run as admin, so the hash override feature becomes completely unavailable.

This is good default behavior. The only reason hash overrides are really needed right now are because the manifest update time is so long, but once the bots get better and we have other repos to choose from this shouldn't be much of an issue.

denelon:
We're looking at several options to deal with this scenario. There have been several concerns about having command-line overrides that could potentially result in malicious code getting installed by the Windows Package Manager (eg. copy-paste this command from some "bad actor"). 

A few options have been suggested:

Have a setting requirement to enable the capability (enables a Group Policy to block behavior by IT administrators).
[...]

PR #2217, opened 2022-06-03
Add support for DisplayInstallWarnings

### LINKED ISSUE #1786: [New Feature]: Add ability to display warning to user before upgrade
### Description of the new feature/enhancement

While upgrading programs, all my explorer windows closed with no warning due to WinMerge closing it as part of its upgrade path. Would it be possible to add a warning to programs known to interfere with others as part of their upgrades or installs?

### Proposed technical implementation details (optional)

Add a upgrade after user issues command to the tone of "This upgrade might close other programs as part of its upgrade, make sure you save your work first!" with a prompt Y/N before continuing and a flag to the upgrade command to ignore possible upgrade closures to disable the prompt for scripts and people who don't care

ryfu-msft:
Resolves #1786

Changes:
- Adds support for prompting the user with a warning message if the DisplayInstallWarnings field is set to true. The message notifies the user that the installer may interfere with running applications
- Added `--ignore-warnings` flag to allow users to override the default behavior of warning prompts

Tests:
- Added unit tests to verify that prompts are called correctly with the right message. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2217)


yao-msft:
This change will probably block com callers with packages having warnings. We need to expose something in InstallOptions in Com interface

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -97,6 +97,8 @@ namespace AppInstaller::CLI
             return Argument{ "wait", NoAlias, Args::Type::Wait, Resource::String::WaitArgumentDescription, ArgumentType::Flag, false };
         case Args::Type::ProductCode:
             return Argument{ "product-code", NoAlias, Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
+        case Args::Type::IgnoreWarnings:
+            return Argument{ "ignore-warnings", NoAlias, Args::Type::IgnoreWarnings, Resource::String::IgnoreWarningsArgumentDescription, ArgumentType::Flag, false };
```
yao-msft, line 101: nit: I think "ignore-install-warnings" would be more accurate, "ignore-warnings" seems too generic. Also please update the variable name to leave room for future warnings

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -443,6 +458,7 @@ namespace AppInstaller::CLI::Workflow
     void InstallSinglePackage(Execution::Context& context)
     {
         context <<
+            Workflow::DisplayInstallWarnings <<
```
yao-msft, line 461: Need to update InstallMultiplePackages as well for import, dependencies, upgrade all etc

src/AppInstallerCLICore/Argument.cpp
yao-msft: > nit: I think "ignore-install-warnings" would be more accurate, "ignore-warnings" seems too generic. Also please update the variable name to leave room for future warnings

Also you'll need to expose them in each applicable command.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1349,4 +1349,10 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="ShowLabelInstallationNotes" xml:space="preserve">
     <value>InstallationNotes:</value>
   </data>
+  <data name="IgnoreWarningsArgumentDescription" xml:space="preserve">
+    <value>Disables warning prompts</value>
```
yao-msft, line 1353: > <value>Disables warning prompts</value> [](http://example.com/codeflow?start=3&length=40)

nit: Disables install warning prompts

yao-msft (changes requested):
:clock1:

yao-msft:
Since this field is per installer level, maybe for better user experience, shall we change the installer selection logic to pick installers without warnings over warnings?

src/AppInstallerCLICore/Argument.cpp
denelon: I've seen a couple of features reported with similar concerns like:
* #715

I just added it to:
* #2128

Trenly:
> Since this field is per installer level, maybe for better user experience, shall we change the installer selection logic to pick installers without warnings over warnings?

I would say no - I wouldn't want winget to select an x86 installer for a x64 system if the x64 installer had installer warnings. Similarly, I wouldn't want the warnings to change a user scoped installation to a machine scoped installation, just because the machine scope didn't have warnings. I could see adding a setting to always suppress the warnings, but I don't think that changing the selection behavior based on a warning is a good idea, since it isn't a guarantee that the application is running when the user attempts the upgrade

yao-msft:
> > Since this field is per installer level, maybe for better user experience, shall we change the installer selection logic to pick installers without warnings over warnings?
> 
> I would say no - I wouldn't want winget to select an x86 installer for a x64 system if the x64 installer had installer warnings. Similarly, I wouldn't want the warnings to change a user scoped installation to a machine scoped installation, just because the machine scope didn't have warnings. I could see adding a setting to always suppress the warnings, but I don't think that changing the selection behavior based on a warning is a good idea, since it isn't a guarantee that the application is running when the user attempts the upgrade

I agree, the installer selection logic actually has different levels of priority when determining which installer to pick, the priority of the InstallWarnings should always be in the lowest. Only when 2 installers have same scope, arch etc would the InstallWarnings be considered a factor. But given how rare this would be, I'm ok with not implementing anything. Just throwing some thought out.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -158,6 +158,21 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
+    void DisplayInstallWarnings(Execution::Context& context)
+    {
+        auto displayInstallWarnings = context.Get<Execution::Data::Installer>().value().DisplayInstallWarnings;
+
+        if (displayInstallWarnings && !context.Args.Contains(Execution::Args::Type::IgnoreWarnings) && !Settings::User().Get<Settings::Setting::InstallIgnoreWarnings>())
+        {
+            bool ignoreWarning = context.Reporter.PromptForBoolResponse(Resource::String::InstallWarning);
```
JohnMcPMS, line 167: I'm not sure what the use case is here, but I don't like prompting on a warning.  At a minimum I would think that the setting should be ignore/display/prompt with the default being display.

ignore = Don't show them at all
display = Display the warning once the installer is chosen (before download) which gives an interactive user a fair amount of time to react with CTRL+C
prompt = This behavior of prompting after displaying the warnings.

src/AppInstallerCLICore/Argument.cpp
JohnMcPMS: Breaking or not, we should change `--force` per this discussion: #2112

JohnMcPMS (changes requested):
I don't think that this feature has been though through enough.  I would expect that many upgrades affect their own package, which is in line with the current message of "interfering with running applications".  So the flag is added, then nearly every package prompts you?

I get that it is painful if running an installer messes with active processes, but I don't think this is the right way to deal with that.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
Trenly: I agree that not prompting is better, especially knowing that WinGet is used in many autonomous install/upgrade scenarios

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Added to `InstallMultiplePackages`

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: Changed to `Disables install warnings`

Issue #530, opened 2020-08-09
Ability to list tags of a specific Package


topherhoffman:
# Description of the new feature/enhancement

It would be great to list the tags being used on a package, it is always best to look up software on the website itself, however if just browsing the package, it would be great to get some more details on what it actually aims to accomplish.

Currently the only way to view the tags is via a search for the tag, 

example of current usage of tags

winget search android
```
Name                 Id                           Version    Match
-------------------------------------------------------------------------
PhotoSync            touchbyte.PhotoSync          4.0.4      Tag: android
Huawei Mobile Cloud  huawei.mobilecloud           10.0.0.300 Tag: android
Surface Duo Emulator Microsoft.SurfaceDuoEmulator 2020.624.2 Tag: android
Huawei Quick App IDE Huawei.QuickAppIde           2.5.2      Tag: android
Huawei Pcmanager     huawei.pcmanager             10.0.2.59  Tag: Android 
```

# Proposed technical implementation details (optional)
Add an additional Option to the show method/function (Not sure what to call it)

`winget show <Package Name> -tag`
`winget show <Package Name> -t`

that is the first thing that comes to mind.  It could list out the tags in a space separated list.  (kind of how ls in linux list files.)

PR #2235, opened 2022-06-10
Feat: Show package tags

Trenly:
  - #530

-----

```raw
PS C:\WINDOWS\system32> wingetdev show 7zip --tags --versions
Windows Package Manager (Preview) v1.3.2-preview

Both `tags` and `versions` arguments are provided

Shows information on a specific package. By default, the query must case-insensitively match the id, name, or moniker of the package. Other fields can be used by passing their appropriate option.

usage: winget show [[-q] <query>] [<options>]

  -m,--manifest               The path to the manifest of the package
  -v,--version                Use the specified version; default is the latest version
  --versions                  Show available versions of the package
  --tags                      Show the package tags

More help can be found at: https://aka.ms/winget-command-show
PS C:\WINDOWS\system32> wingetdev show 7zip.7zip --tags
Found 7-Zip [7zip.7zip]
Tags
----------------
archiver
compression
file-compression
foss
open-source
tool
utility
PS C:\WINDOWS\system32> wingetdev show 7zip.7zip --tags -v 16.04
Found 7-Zip [7zip.7zip]
Tags
----------------
7zip
compression
file-compression
tool
utility
zip
PS C:\WINDOWS\system32> wingetdev show 7zip --tags -s msstore
Found 9 ZIP - open rar, zip, 7zip, gzip [9NBLGGH3TDHG]
No tags were found.
```


lechacon:
Why not add the tags to the normal output of `show` instead of having a separate argument for it? It seems to me like the reasons to have the versions listed separately don't apply for the tags.

Trenly:
> Why not add the tags to the normal output of `show` instead of having a separate argument for it? It seems to me like the reasons to have the versions listed separately don't apply for the tags.

Because iirc the normal show will show information, and --tags will show the tags from all locales.

lechacon:
1. Why would I want to see tags from other locales? For example, if I do `winget show Mozilla.Firefox`, I probably don't want to see all tags "browser", "navegador", "браузер", "浏览器" that all mean just "browser".
2. Why can't that be part of all the other information?

But now that you bring up the locale in `show`, I think it could use the `--locale` argument to show the appropriate info.

Trenly:
Thats a fair point; I was going based just on the issue, but I can see how just showing the tags would be better

lechacon (approved):
It may be good to add this to the unit tests for `show`.

ppvnf:
I've tested v1.4.2161-preview and `winget show package.id ` indeed shows the package's tag, however it's not possible to use `winget show --tag` as proposed in #530. Is there any reason for that? Thanks.

Captura de tela 2022-10-10 182907

Trenly:
> I've tested v1.4.2161-preview and `winget show package.id ` indeed shows the package's tag, however it's not possible to use `winget show --tag` as proposed in #530. Is there any reason for that? Thanks.
> 

Tags are per version and per locale. To implement it as described in the original request would require either selection of a version and locale, or to aggregate all the tags from all locales and versions which would produce unwieldly data. 
> Why would I want to see tags from other locales? For example, if I do winget show Mozilla.Firefox, I probably don't want to see all tags "browser", "navegador", "браузер", "浏览器" that all mean just "browser".

The default behavior then would have to be to show the latest version and applicable locale (or defaultLocale if none were applicable), which is the same behavior as integrating it directly into the show command. If you read through the history above, I actually had implemented it as a [separate argument](https://github.com/microsoft/winget-cli/commit/f04ce766a2d44ddf08adf4654e61b7fecff4b227) initially, but then upon feedback it was changed since it would essentially be showing just the tags portion of the output and it doesn't make sense to require a separate argument for that.

Issue #2061, opened 2022-04-01
Add argument to pause after command execution


denelon:
### Description of the new feature / enhancement

Sometimes I'd like to be able to use PowerToys Run or Windows Run or some other mechanism to execute a Windows Package Manager command. The experience isn't great as the terminal is closed as soon as execution is completed.

### Proposed technical implementation details

I'd like to be able to add the "--wait" argument to any command to wait for user input.

```
PS C:\Users\denelon> winget search WindowsTerminal --wait
Name                     Id                                Version      Source
-------------------------------------------------------------------------------
Windows Terminal Preview Microsoft.WindowsTerminal.Preview 1.13.10733.0 winget
Windows Terminal         Microsoft.WindowsTerminal         1.12.10732.0 winget
Press any key to continue.
```

PR #2236, opened 2022-06-10
Feat: Wait upon command completion

Trenly:
  - #2061

-----

Seems the argument was already there, just never implemented
* Tested Manually


src/AppInstallerCLICore/Command.cpp
```diff
@@ -811,6 +811,10 @@ namespace AppInstaller::CLI
         {
             ExecuteInternal(context);
         }
+        if (context.Args.Contains(Execution::Args::Type::Wait))
+        {
+            system("pause");
```
jedieaston, line 816: Isn't `system` a (mild) crime against humanity?

I've done this before with `getchar()`, which just takes the next character in the console. How is the prompting done for agreements? Maybe that would work too.

Trenly: I wasn't aware that it was a crime against humanity

JohnMcPMS: I wouldn't call it a crime against humanity, but we certainly don't need to invoke another process to wait on.  I would also suggest using a similar pattern to `Reporter::PromptForBoolResponse` as @jedieaston suggested.  I would prefer though a `Reporter::PromptForAnyKey` that changes the semantics.

godgeema (approved):

src/AppInstallerCLICore/ExecutionReporter.cpp
```diff
@@ -155,6 +156,13 @@ namespace AppInstaller::CLI::Execution
         }
     }
 
+    int Reporter::PromptForAnyKey(Resource::LocString message, Level level)
+    {
+        auto out = GetOutputStream(level);
+        out << message << std::endl;
+        return _getch();
```
JohnMcPMS, line 163: The field `std::istream& m_in` has to be used for all input.  I think that despite the documentation, the buffer inside the standard input stream waits for characters to be present rather than returning errors.  From looking at the code, I don't see why `m_in.get()` wouldn't work here.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/ExecutionReporter.cpp
Trenly: It would work; I'll just have to change it to be `PromptForEnter` since buffered input streams won't return unless enter is pressed.

JohnMcPMS (approved):

JohnMcPMS:
Sorry that this slipped through for the last two weeks...

Trenly:
> Sorry that this slipped through for the last two weeks...

No worries; I figured someone would get to it eventually. Besides, you have more important stuff you're working on anyways

denelon:
I kinda' have a pet feature riding on this one elsewhere, so I was excited to see the PR 😊

denelon:
Time to cut 1.4-preview (1) I hope. 🤞

PR #2242, opened 2022-06-13
Check symlink target before removal

### LINKED ISSUE #2190: Portable uninstall should check target path of symlink before removal

Currently, Portable uninstall only checks that the portable symlink path matches what is recorded in ARP. However, the symlink could be modified to point to a different target exe. We should verify that the target of the symlink points to the correct portable exe that we placed down during installation.

### Steps to reproduce

Modify symlink to point to another exe. Uninstall will still remove the symlink even though it is modified.

### Expected behavior

Symlink should not be removed as it has been modified by the user.

### Actual behavior

If the symlink path matches the symlink path recorded in ARP, the symlink will always be removed regardless of whether it has been modified by the user.

### Environment

```shell
Windows Package Manager 1.3.1391-preview
```

ryfu-msft:
Resolves #2190 

Changes:
- Checks the target of the portable symlink to verify that it matches the target exe path that is stored in ARP.

Tests:
- Upgraded E2E tests to .NET6.0 to use File.CreateSymbolicLink()
- Added Uninstall Test to verify that the symlink will not be deleted if it does not match the target exe path.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2242)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/PortableFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/PortableFlow.cpp
@@ -451,14 +451,14 @@ namespace AppInstaller::CLI::Workflow
             const auto& symlinkPath = uninstallEntry[PortableValueName::PortableSymlinkFullPath];
             if (symlinkPath.has_value())
             {
-                const std::filesystem::path& symlinkPathValue = symlinkPath.value().GetValue<Value::Type::UTF16String>();
+                const std::filesystem::path& symlinkPathValue = symlinkPath->GetValue<Value::Type::UTF16String>();
                 if (std::filesystem::is_symlink(std::filesystem::symlink_status(symlinkPathValue)))
                 {
                     const auto& targetPath = uninstallEntry[PortableValueName::PortableTargetFullPath];
                     if (targetPath.has_value())
                     {
                         const std::filesystem::path& symlinkTargetPath = std::filesystem::read_symlink(symlinkPathValue);
-                        const std::filesystem::path& targetPathValue = targetPath.value().GetValue<Value::Type::UTF16String>();
+                        const std::filesystem::path& targetPathValue = targetPath->GetValue<Value::Type::UTF16String>();
                         if (symlinkTargetPath != targetPathValue)
                         {
                             AICLI_LOG(CLI, Warning, << "Portable symlink not deleted; Symlink points to a different target exe: " << symlinkTargetPath <<
--- a/src/AppInstallerCLICore/Workflows/PortableFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/PortableFlow.cpp
@@ -452,10 +452,28 @@ namespace AppInstaller::CLI::Workflow
             if (symlinkPath.has_value())
             {
                 const std::filesystem::path& symlinkPathValue = symlinkPath.value().GetValue<Value::Type::UTF16String>();
-
-                if (!std::filesystem::remove(symlinkPathValue))
+                if (std::filesystem::is_symlink(std::filesystem::symlink_status(symlinkPathValue)))
                 {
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -452,10 +452,28 @@ namespace AppInstaller::CLI::Workflow
...
                 {
-                    AICLI_LOG(CLI, Info, << "Portable symlink not found; Unable to delete portable symlink: " << symlinkPathValue);
+                    const auto& targetPath = uninstallEntry[PortableValueName::PortableTargetFullPath];
+                    if (targetPath.has_value())
+                    {
+                        const std::filesystem::path& symlinkTargetPath = std::filesystem::read_symlink(symlinkPathValue);
+                        const std::filesystem::path& targetPathValue = targetPath.value().GetValue<Value::Type::UTF16String>();
+                        if (symlinkTargetPath != targetPathValue)
+                        {
+                            AICLI_LOG(CLI, Info, << "Portable symlink not deleted; Symlink points to a different target exe: " << symlinkTargetPath <<
```
Trenly, line 464: Since the context reporter is reporting a warning, I would expect the log to be a warning also
```suggestion
                            AICLI_LOG(CLI, Warning, << "Portable symlink not deleted; Symlink points to a different target exe: " << symlinkTargetPath <<
```

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -452,10 +452,28 @@ namespace AppInstaller::CLI::Workflow
...
-
-                if (!std::filesystem::remove(symlinkPathValue))
+                if (std::filesystem::is_symlink(std::filesystem::symlink_status(symlinkPathValue)))
                 {
-                    AICLI_LOG(CLI, Info, << "Portable symlink not found; Unable to delete portable symlink: " << symlinkPathValue);
+                    const auto& targetPath = uninstallEntry[PortableValueName::PortableTargetFullPath];
+                    if (targetPath.has_value())
+                    {
+                        const std::filesystem::path& symlinkTargetPath = std::filesystem::read_symlink(symlinkPathValue);
+                        const std::filesystem::path& targetPathValue = targetPath.value().GetValue<Value::Type::UTF16String>();
```
JohnMcPMS, line 461:
```suggestion
                        const std::filesystem::path& targetPathValue = targetPath->GetValue<Value::Type::UTF16String>();
```
nit, but I don't know why you wouldn't want it to be easier to read...

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1349,4 +1349,7 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="ShowLabelInstallationNotes" xml:space="preserve">
     <value>InstallationNotes:</value>
   </data>
+  <data name="SymlinkModified" xml:space="preserve">
+    <value>Portable symlink not deleted as it was modified and points to a different target exe</value>
```
JohnMcPMS, line 1353: Making loc changes after we have begun the release process makes it take longer.  We should discuss whether this change makes it to 1.3 today with @denelon.

JohnMcPMS (approved):

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: Discussed with @denelon, this will be included in 1.4

PR #2268, opened 2022-06-22
Add alternate name to Usage String

Trenly:
When merge conflicts were resolved in #2216 it removed the alternate name from the usage string; This PR adds it back in

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2268)

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -119,6 +119,10 @@ namespace AppInstaller::CLI
             strstr << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << m_alias << ',';
         }
         strstr << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << m_name;
+        if (m_alternateName != Argument::NoAlternateName)
+        {
+            strstr << ',' << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << m_alternateName;
+        }
         return strstr.str();
     }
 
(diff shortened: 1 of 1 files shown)


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -119,6 +119,10 @@ namespace AppInstaller::CLI
             strstr << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << m_alias << ',';
         }
         strstr << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << m_name;
+        if (m_alternateName != Argument::NoAlternateName)
```
yao-msft, line 122: I saw the original code before merge in #2216 has this code block before line 121, did you intentionally make this change?

Trenly: I'm a bit confused what you mean; I see the code before merge in #2216 has the Alias before Line 121, but not the alternate name, which was added in #2107

yao-msft: Sorry, I meant the code before the refactor [here](https://github.com/microsoft/winget-cli/pull/2216/files#diff-4a382bca9249783510a05c7cafcc59198493470db58a22035dba5cc3c62c4f08) has this line 
`strstr << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << APPINSTALLER_CLI_ARGUMENT_IDENTIFIER_CHAR << arg.Name();`
after the if NoAlternateName block,
but in this change the line is before the if NoAlternateName, I think it's not an intentional change?

Trenly: I don't think it matters if it is before or after; I can update it to be before;

yao-msft (approved):

PR #2270, opened 2022-06-22
ZIP support spec 

ryfu-msft:
Related to:
#140 

Changes:
- Added a spec doc for detailing how ZIP installers will be supported.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2270)


doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,90 @@
...
+
+The check will be as follows:
+
+1. For any archive file, we will start by using Windows Shell APIs which treat ZIP files as a Folder object that can be utilized for viewing the contents (we are not extracting to a specified location, yet). We will iterate through each item and determine the cumulative sum of the sizes of all items contained in the zip and compare that value with the size of the ZIP file. **The check will verify that the compression ratio is no greater than 5:1 (20%)**. Any file that has a larger compression ratio (<20%) will be deemed as "suspicious" and a warning will be displayed to the user. Lossless compression of common data files tend to achieve compression ratios of ~2:1 (50%). Any extraction or copying of files will not take place until this check is passed.
+
+2. If a ZIP archive file contains nested archive files, we will support decompressing up to a maximum of 3 layers (2 additional layers including the initial ZIP file). 
+
+If either of these checks fail, the process will terminate and a warning will be displayed to the user. The user can include `--force` to bypass this check and continue installation at their own risk.
```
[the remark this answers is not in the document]
BEEDELLROKEJULIANLOCKHART: Why solely 3 layers? That decision appears arbitrary.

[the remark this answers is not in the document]
Trenly: > Why solely 3 layers? That decision appears arbitrary.

Although it is likely "arbitrary", it is actually an embodiment of good practices. While there are times that putting zipped files inside of zipped files does make sense, it is extremely rare that it is necessary to go more than one layer beyond that. In fact, in [this thread](https://stackoverflow.com/questions/1166385/how-many-times-can-a-file-be-compressed) Nosredna talks about how compressing a file too many times actually leads to the archive being bigger than it originally was.  

So a first layer of compression, which is putting a file inside of a zip, makes a lot of sense. In some cases, a second layer of compression, which is putting a zip inside of a zip, also makes sense, especially when files have a lot of duplicate data. From there, it doesn't really save much extra space to go to that third layer of compression. It can be useful in some situations, but doesn't generally do much. Fourth, fifth, and even more layers of compression are actually more likely to be counterproductive than anything.

Because one layer is common, two layers is not unheard of, and three layers is rare, it doesn't make sense for winget to support more than 3 layers. Additionally, supporting a maximum of 3 layers will force developers to not go overboard on the compression if they want their package to be available in winget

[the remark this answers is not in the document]
ryfu-msft: @BEEDELLROKEJULIANLOCKHART, it is relatively arbitrary. I decided to opt for a conservative number as it doesn't really make sense to support more than a 3rd layer as @Trenly has explained. If the community strongly feels that 3 is too restrictive, we can increase it but for now I feel that 3 is sufficient.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,90 @@
...
+
+>Compression Ratio is defined as the sum of the sizes of all the items contained within a ZIP file, divided by the size of the ZIP file.
+
+The check will be as follows:
+
+1. For any archive file, we will start by using Windows Shell APIs which treat ZIP files as a Folder object that can be utilized for viewing the contents (we are not extracting to a specified location, yet). We will iterate through each item and determine the cumulative sum of the sizes of all items contained in the zip and compare that value with the size of the ZIP file. **The check will verify that the compression ratio is no greater than 5:1 (20%)**. Any file that has a larger compression ratio (<20%) will be deemed as "suspicious" and a warning will be displayed to the user. Lossless compression of common data files tend to achieve compression ratios of ~2:1 (50%). Any extraction or copying of files will not take place until this check is passed.
+
+2. If a ZIP archive file contains nested archive files, we will support decompressing up to a maximum of 3 layers (2 additional layers including the initial ZIP file).
```
lechacon, line 71: Is this recursive decompressing talking only about the threat detection, or is it part of the installation?

If part of the installation, how do you intend to use decompressing multiple nested archives?
For example:
Would I be able to specify in the manifest that I want to run the installer `/archive.zip/setup.exe`? Where would you decompress the nested archives to? What if I gave you a zip containing `setup.exe` and `data.zip`, and `setup.exe` used the compressed `data.zip`; would you still decompress it?

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,90 @@
...
+## ZIP Threat Detection
+ZIP and other archive file types are known threat vectors for malware in the form of ZIP compression bombs. Two of the most common types of compression bombs are multi-layered (recursive) and single-layered (non-recursive). ZIP bombs rely on a repetition of identical files that have extremely large compression ratios. 
+
+1. **Multi-layered (recursive)**:
+A ZIP file containing multiple layers of nested ZIP files that achieve high compression ratios when decompressed recursively. This technique is often used to bypass compression ratio checks performed by ZIP parsers for a single layer.
+
+2. **Single-Layered (non-recursive)**
+A ZIP bomb that expands fully after a single round of decompression. The extremely high compression ratio of the ZIP bomb is achieved by overlapping files within the ZIP container. 
+
+In order to protect our users from these possible threats, we will need to implement several basic checks to verify whether it is safe to proceed with the contents of the ZIP archive file.
```
lechacon, line 63: Do we have a way to prevent these threats on other installer types? For example, imagine I create an EXE with a ZIP bomb embedded and have the EXE decompress it. Would we have the same issues we are trying to avoid here?

Trenly: > Do we have a way to prevent these threats on other installer types? For example, imagine I create an EXE with a ZIP bomb embedded and have the EXE decompress it. Would we have the same issues we are trying to avoid here?

Thats why we have the pipeline validation, isn’t it? If an exe were a zip bomb, it would hang the VM running the validation and trigger an Internal-Error, assuming it made it past the AV scan in the first place

denelon: That's part of it, but we would like to be able to offer some protection for users accessing third party REST sources as well.

doc/specs/#140 - ZIP Support.md
lechacon: Okay. Just to be annoying bringing up edge cases...
Imagine you have `foo.zip` contains `bar.zip` which contains `setup.exe`. But `foo.zip` also contains `bar/setup.exe` (`setup.exe` inside a directory called `bar`). Both would be decompressed to `bar/setup.exe`. What do we do then?

yao-msft: I think setup.exe with data.zip is a valid scenario. Maybe we should only unzip the top layer as default and add some manifest entries to let us know we should unzip additional layers?

doc/specs/#140 - ZIP Support.md
ryfu-msft: That brings up a good point, this scenario would need to additionally scan/verify whether the provided EXE has an embedded zip file, and once that is determined, we could apply the checks here. I do feel like scanning installers (as opposed to scanning a zip file before extracting to obtain an installer) seems to be out of the scope of winget.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,90 @@
+---
+author: Ryan Fu @ryfu-msft
+created on: 2022-05-24
+last updated: 2022-05-24
+issue id: 140
+---
+
+# Support ZIP/Archive installerTypes
```
yao-msft, line 8:
```suggestion
# Support ZIP/Archive InstallerTypes

```

doc/specs/#140 - ZIP Support.md
ryfu-msft: > Okay. Just to be annoying bringing up edge cases... Imagine you have `foo.zip` contains `bar.zip` which contains `setup.exe`. But `foo.zip` also contains `bar/setup.exe` (`setup.exe` inside a directory called `bar`). Both would be decompressed to `bar/setup.exe`. What do we do then?

Not annoying :) I like Yao's suggestion to include a manifest entry like `fullDecompression = true` to indicate whether we decompress additional layers. If there is already an existing folder with the same name, maybe the decompressed folder would append a string to the end of the name like `"bar (1)"` to signify that there is a duplicate folder. This might be slighly confusing but my hope is that there wouldn't be any nested zips that conflict with nested folders that contain the same files.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,90 @@
...
+Currently, information regarding a single installed portable is stored in the ARP entry. In order to support installing a single or multiple portables contained inside an archive file, an additional table of information will need to be appended to the tracking catalog. The table will contain a list of files that were created and placed down as well as various metadata to enable us to verify whether they have been modified. This table should replace most of the uninstall-related information that is stored in ARP for a given portable package. 
+
+The table will contain the following information for each item that we create or place down during installation.
+
+| Metadata    | Description |
+| ----------- | ----------- |
+| Path      | Path to the item |
+| Flag   | Flag to indicate what type of file was placed down (directory, symlink, exe) |
+| Hash   | Hash for files with contents (exe) |
+| SymlinkTarget | Target exe for the created symlink (only applies to symlink files) |
```
yao-msft, line 89: nit: are these metadata leftover from previous portable spec? They were not referenced in this spec.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,90 @@
...
+The initial implementation of this feature will only support the following installation scenarios:
+- Single base installer (exe, msi, msix) contained inside an archive file. 
+- Single or multiple portable exes bundled as a suite inside an archive file.
+
+## ZIP Extraction
+The extraction of ZIPs will be done using Windows Shell APIs. ZIP files can be represented as a [ShellFolder](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ishellfolder) object that can be used to manage the contents of the ZIP file. File contents are represented as [ShellItems](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ishellitem), which can be handled using the methods exposed by the [IFileOperation interface](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ifileoperation). If the `NestedInstallerType` is non-portable, the installer file will be copied to a temporary location and used during execution of the installation flow. If the `NestedInstallerType` is portable, all files will be moved to the appropriate install location and symlinks will be created for each `NestedInstallerFile` entry.
```
yao-msft, line 48: > If the `NestedInstallerType` is non-portable, the installer file will be copied to a temporary location and used during execution of the installation flow.

Are we going to support setup.exe with data cabs beside it? I thought this will be the most used cases for zip. In this case, only copying the installer file is not enough.

doc/specs/#140 - ZIP Support.md
JohnMcPMS: Do we have a use case for extracting nested archives?  It seems more likely to break things than help, in the event that a data.zip file exists.  I would say that until a concrete example exists, we should start with only extracting the top level archive.

doc/specs/#140 - ZIP Support.md
JohnMcPMS: Yes, it should be "the entire archive is extracted to a temporary location".

doc/specs/#140 - ZIP Support.md
BEEDELLROKEJULIANLOCKHART: @ryfu-msft, scanning what `winget`'s default repository provides is certainly not out of its scope, because every Linux distribution's repositories do that, and all of the software that is available via the `msstore` repository has been, too.

Trenly: > @ryfu-msft, scanning what `winget`'s default repository provides is certainly not out of its scope, because every Linux distribution's repositories do that, and all of the software that is available via the `msstore` repository has been, too.
> 
> Additionally, Microsoft's threat-detection services are so brilliant, not utilizing them would be really weird.

Thats not what Ryan was saying at all. The winget-pkgs repository does have all of the packages scanned before they can be added. What Ryan was saying is that private sources and third-party REST sources are outside of the scope of winget in terms of scanning the installer files they provide for security.

You’re also conflating “Package manager” with “Package source”. On Linux especially, apt and snap are package managers, but they do not scan the entire distribution network as you said. Rather, they only fetch the package from the source - and in order to get into the source the package has to be scanned. The difference with apt, snap, and msstore is that they also host the binary files. Winget-pkgs does not host the binaries, instead it hosts the manifests which point to where to download the binaries and how to install them. 

Apt and snap also allow users to host their own private repositories, and similar to winget, there is no guarantee that packages from those repositories are safe. This is because, like winget, apt and snap are just package managers - they get information about the package and have the ability to install, uninstall, and upgrade the package. It is, and almost always has been, the responsibility of the package source, or package repository, to maintain the integrity of their packages. The package managers do very little to ensure security outside of verifying that the installer is the correct installer and that the user has permission to perform the action which is being requested.

doc/specs/#140 - ZIP Support.md
ryfu-msft: No, this is metadata that we intend to capture in our tracking catalog instead of recording them in ARP to support the uninstall scenario for portable(s) contained inside a zip.

JohnMcPMS: Not tracking catalog, but a separate record of files stored with the extracted files.

doc/specs/#140 - ZIP Support.md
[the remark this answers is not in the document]
ryfu-msft: Updated suggested argument for overriding zip malware security checks. We also decided to leverage a zip file malware checker for scanning zip files rather than attempting to write our own.

doc/specs/#140 - ZIP Support.md
ryfu-msft: Changed to specify that the entire archive will be extracted to a temporary location

doc/specs/#140 - ZIP Support.md
ryfu-msft: Added explanation to specify that initial implementation will only extract the top-level archive to support the scenarios stated above

doc/specs/#140 - ZIP Support.md
ryfu-msft: We will now leverage a zip malware checking library to scan zip files rather than implementing our own malware detection.

doc/specs/#140 - ZIP Support.md
ryfu-msft: Updated description to clarify that this is not appended to the tracking catalog but instead captured as a separate record.

doc/specs/#140 - ZIP Support.md
```diff
@@ -0,0 +1,88 @@
...
+2. **Single-Layered (non-recursive)**
+A ZIP bomb that expands fully after a single round of decompression. The extremely high compression ratio of the ZIP bomb is achieved by overlapping files within the ZIP container. 
+
+In order to protect our users from these possible threats, we will need to scan the ZIP file for malware using [Pure, a static analysis file format checker](https://github.com/ronomon/pure). 
+
+> Pure is licensed under the [MIT license](https://github.com/ronomon/pure/blob/master/LICENSE).
+
+If the Pure library functions that are called on a given ZIP file detect malware, the process will terminate and a warning will be displayed to the user. The user can include `--override-archive-malware-checks` to bypass this check and continue installation at their own risk.
+
+These checks will only applied in the following scenarios to minimize unnecessary costs on performance during installation.
```
Trenly, line 73:
```suggestion
These checks will only applied when any of the following conditions are met to minimize unnecessary costs on performance during installation. 
```
Clarification that these are meant to be ‘or’ and not ‘and’

marypcbuk: only *be* applied?

ryfu-msft: Changed to suggested

JohnMcPMS (approved):

PR #2296, opened 2022-06-30
Fix WinGetUtil WinGetCreateManifest result

yao-msft:
Add succeeded result.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2296)


AmelBawa-msft (approved):

JohnMcPMS:
I was just saying that we should probably add tests in the E2E for WinGetUtil.dll as it is getting bigger and more use now.

PR #2311, opened 2022-07-06
Standardize 'Show' labels to manifest fields

### LINKED ISSUE #1690: Update "show" output

The show command should display the values in the manifest as the key / value pairs rather than the current names.

For example "Download URL" is not one of the manifest keys.

_No response_

Trenly:
    - Resolves #1690

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2311)

### DIFF
--- a/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
+++ b/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
@@ -1031,19 +1031,19 @@ Do you agree to the terms?</value>
     <value>Installer:</value>
   </data>
   <data name="ShowLabelInstallerLocale" xml:space="preserve">
-    <value>Locale:</value>
+    <value>Installer Locale:</value>
   </data>
   <data name="ShowLabelInstallerProductId" xml:space="preserve">
     <value>Store Product Id:</value>
   </data>
   <data name="ShowLabelInstallerSha256" xml:space="preserve">
-    <value>SHA256:</value>
+    <value>Installer SHA256:</value>
   </data>
   <data name="ShowLabelInstallerType" xml:space="preserve">
-    <value>Type:</value>
+    <value>Installer Type:</value>
   </data>
   <data name="ShowLabelInstallerUrl" xml:space="preserve">
-    <value>Download Url:</value>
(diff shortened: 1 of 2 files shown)


denelon:
@Trenly

We had a discussion internally a while back related to the labels. Part of the reason we were exposing the exact "YAML" key values is because we weren't localizing the titles. I'm not sure how well this plays from a localization perspective.

Thoughts?

Trenly:
> @Trenly
> 
> We had a discussion internally a while back related to the labels. Part of the reason we were exposing the exact "YAML" key values is because we weren't localizing the titles. I'm not sure how well this plays from a localization perspective.
> 
> Thoughts?

Hmmm. I'm not sure either. I think from a contributor perspective, I would rather see exact YAML keys (spaces added) as it provides simpler correlation between fields. From a user perspective, I would want the fields to be localized so I could understand what each of them were. 

However, I would say that there is a fairly clear link between the Yaml keys and the meaning of the key. From an english-First perspective, I don't see any reason not to update the keys. From a localized perspective, I can see how it could make the localization more difficult to understand.

yao-msft:
I think last time we discussed that we would de-localize the manifest keys for show, for scenarios Demitrius mentioned. Are we ok not doing that now?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1055,7 +1055,7 @@ Do you agree to the terms?</value>
     <value>Moniker:</value>
   </data>
   <data name="ShowLabelPackageUrl" xml:space="preserve">
-    <value>Homepage:</value>
+    <value>Package Url:</value>
```
JohnMcPMS, line 1058: I think that this one could be confusing, as some people may think that "package" means "installer".

Trenly: Apologies for the delay; Reverted

JohnMcPMS:
Not that a build is needed... just following protocol.

yao-msft (approved):

PR #2312, opened 2022-07-06
Move Release Notes to new line and indent

Trenly:

It bothered me that Multi-Line release notes had weird alignment. This places the release notes down one line and indents all lines.

image

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2312)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ShowFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShowFlow.cpp
@@ -8,6 +8,7 @@
 
 using namespace AppInstaller::Repository;
 using namespace AppInstaller::CLI;
+using namespace AppInstaller::Utility;
 using namespace AppInstaller::Utility::literals;
 
 namespace AppInstaller::CLI::Workflow
@@ -52,7 +53,16 @@ namespace AppInstaller::CLI::Workflow
         }
         if (!description.empty())
         {
-            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelDescription << ' ' << description << std::endl;
+            bool isMultiLine = RegexFindAndReplace(description, std::regex("\n"), "\n  ");
+            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelDescription;
+            if (isMultiLine)
+            {
+                info << std::endl << "  " << description << std::endl;
+            }
+            else
+            {
+                info << ' ' << description << std::endl;
+            }
         }
         auto homepage = manifest.CurrentLocalization.Get<Manifest::Localization::PackageUrl>();
         if (!homepage.empty())
@@ -83,8 +93,16 @@ namespace AppInstaller::CLI::Workflow
         auto releaseNotes = manifest.CurrentLocalization.Get<Manifest::Localization::ReleaseNotes>();
         if (!releaseNotes.empty())
         {
-            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelReleaseNotes << std::endl;
-            info << "  " << std::regex_replace(releaseNotes, std::regex("\n"), "\n  ") << std::endl;
+            bool isMultiLine = RegexFindAndReplace(releaseNotes, std::regex("\n"), "\n  ");
+            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelReleaseNotes;
+            if (isMultiLine)
+            {
+                info << std::endl << "  " << releaseNotes << std::endl;
+            }
+            else
+            {
+                info << ' ' << releaseNotes << std::endl;
+            }
         }
         auto releaseNotesUrl = manifest.CurrentLocalization.Get<Manifest::Localization::ReleaseNotesUrl>();
         if (!releaseNotesUrl.empty())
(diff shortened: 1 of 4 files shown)


src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -83,7 +83,8 @@ namespace AppInstaller::CLI::Workflow
         auto releaseNotes = manifest.CurrentLocalization.Get<Manifest::Localization::ReleaseNotes>();
         if (!releaseNotes.empty())
         {
-            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelReleaseNotes << ' ' << releaseNotes << std::endl;
+            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelReleaseNotes << std::endl;
+            info << "  " << std::regex_replace(releaseNotes, std::regex("\n"), "\n  ") << std::endl;
```
lechacon, line 87: 1. I like the idea, but if we're doing it maybe we should do the same for any multi-line field? For example, description, agreements, installation notes.
2. I'm a little concerned about how it will look if it has only one line, in the context of everything else being in the same line. Like
```
Label1: Something in one line
Release Notes:
  This is on another line
Label2: In the same line again
```
3. Are we sure this will play nicely in all locales? I'm not an expert in Unicode so I'll trust you if you say it works. I'm thinking something like a multi-byte character where the second byte is the same as `\n` or some other weirdness, but maybe that's not how UTF-8 works.

lechacon: 4. I'm pretty sure it wouldn't be officially supported but do we have anybody parsing the output that we could be breaking?

Trenly: I think doing it for all multi-line would be a good idea too; Guess I hadnt noticed it on other fields

I could implement a check to see if it is single line or multi-line, as your representation is accurate that this currently would put single lines onto the next row and indented also

I haven’t tested with other locales, but my understanding is that it explicitly matches the newline character, which is a standard bytecode in UTF-8 and shouldn’t impact multi-byte characters.

Trenly: > 4\. I'm pretty sure it wouldn't be officially supported but do we have anybody parsing the output that we could be breaking?

I think no, because any multi-line release notes would already be breaking output parsing. If anything, this would make it easier to parse release notes using regex

Trenly: @lechacon - I implemented a new function to perform the replacement as well as to return whether a replacement was made. Used that function to determine if a field was single-line or multi-line and handle appropriately for each case, then extended the functionality to description and installation notes

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -442,6 +444,13 @@ namespace AppInstaller::Utility
         return result;
     }
 
+    bool RegexFindAndReplace(std::string& inputStr, std::regex pattern, const char* value)
```
lechacon, line 447: For what you are using it (replace `\n` by `\n  `), I believe you can use `FindAndReplace()` above without resorting to using regex.

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -52,7 +53,16 @@ namespace AppInstaller::CLI::Workflow
...
+            bool isMultiLine = RegexFindAndReplace(description, std::regex("\n"), "\n  ");
+            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelDescription;
+            if (isMultiLine)
+            {
+                info << std::endl << "  " << description << std::endl;
+            }
+            else
+            {
+                info << ' ' << description << std::endl;
+            }
```
lechacon, line 65: I think this could be abstracted to a function like `ShowMultiLineField` to avoid repeating it for each one.

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -52,7 +53,16 @@ namespace AppInstaller::CLI::Workflow
         }
         if (!description.empty())
         {
-            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelDescription << ' ' << description << std::endl;
+            bool isMultiLine = RegexFindAndReplace(description, std::regex("\n"), "\n  ");
+            info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelDescription;
+            if (isMultiLine)
+            {
+                info << std::endl << "  " << description << std::endl;
```
lechacon, line 60: Nit: `"  "_liv`

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -6,6 +6,8 @@
 #include "Public/AppInstallerLogging.h"
 #include "Public/AppInstallerSHA256.h"
 
+#include <regex>
```
lechacon, line 9: Nit: You already added this in the header

lechacon (approved):
I like it, but I'd wait to see if @denelon or @JohnMcPMS  like the new behavior.

denelon:
I think it looks good.

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -245,4 +246,18 @@ namespace AppInstaller::CLI::Workflow
         }
         table.Complete();
     }
+
+    void ShowMultiLineField(Execution::OutputStream& outputStream, StringResource::StringId label, std::string& value)
```
lechacon, line 250: I'd put this function in an anonymous namespace at the top of this file, instead of declaring it in the header. I believe the *Flow.h headers contain just WorkflowTasks that can be called with `context << task`.

Trenly: Ohhhhh, that makes sense. I think I see now how they did it

lechacon (approved):

JohnMcPMS (approved):

Issue #221, opened 2020-05-19
Add Native PowerShell Support


ThomasNieto:
# Description of the new feature/enhancement

A vast majority of Windows automation is done with PowerShell. To support those users create and ship a winget PowerShell module on [PowerShell Gallery](https://www.powershellgallery.com/).

aglasson:
A colleague of mine sent me a link to the Windows Package Manager Preview just now. The first thing I said to myself was "Why is this a command line application and not a native Powershell cmdlet?". I'm surprised any new command only software from Microsoft isn't Powershell (dotnet/core based) let alone not with Powershell out of the box.
This might just be my naivety showing through.
I'm genuinely curious as to the reasons for that, if any of the devs can provide insight?
Is it perhaps related to this being delivered via the Microsoft Store?

ethanbergstrom:
I've started actively working on a new PackageManagement provider that can lay over the top of WinGet. Based a quick review the CLI doesn't look too different from Chocolatey, so it shouldn't be difficult to adapt to it as a gap-filler until Microsoft puts out something official.

AurimasNav:
The one thing I dislike about chocolatey is having to deal with those string outputs of commands in automation tasks, please make it so we can work with powershell objects.

doctordns:
+ 1 on this.

WInget is a great concept, badly let down by its UI. Here are my problems:

1. No Get-Help - existing help is thin to say the least
2. inconsistent command structure
3. No objects emitted or taken as a parameter
4. No Tab Completion making it harder to use
5. Output is string so no Get-Member to discover output - less discoverable
6. All in all harder to automate programmatically without having to go back to the stone age with grep/sed/.awk
7. Probably not based on .NET so less portable and less easy to leverage.

My view: Winget is a great product, let down by an underimagined UI. The team needs to re-read the Monad Manifesto, then come back with a better implementation. 

Please add real cmdlets. Please make winget fully automatabile. Please bring back the consistency inherent with PowerShell.

JasonFossen:
How could a project like this even get started without planning for PowerShell integration?  The word "PowerShell" does not appear once on the Roadmap page for this project.  Is this a turf war issue between different managers or technical fellows at Microsoft?  

[...]

PR #2314, opened 2022-07-07
Add Microsoft.WinGet.Client PowerShell Module files

jontab:
    - Progress towards #221 
-----
- Manual WinMD collocation with PowerShell executable is still required for PowerShell Desktop
- Help files ([PlatyPS](https://docs.microsoft.com/en-us/powershell/utility-modules/platyps/create-help-using-platyps?view=ps-modules)) are unpopulated

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2314)


src/PowerShell/Microsoft.WinGet.Client/README.md
```diff
@@ -0,0 +1,25 @@
...
+     actual build files. That is, the build script will look for a **bin**
+     directory in this folder and expects a directory structure like:
+     - `bin\x64\Release\net6.0-windows10.0.22000.0\`
+     - `bin\x86\Release\net461\`
+
+### Running the Script
+Once you've done all of the pre-requisite steps, you can build the PowerShell
+module folder by running the PowerShell build script. For example,
+
+- `.\BuildModule.ps1 -In bin -Out build\bin -Configuration Release`
```
lechacon, line 25: Can this be added to the build pipeline?

src/PowerShell/Microsoft.WinGet.Client/Crescendo.json
```diff
@@ -0,0 +1,127 @@
...
+                "Windows"
+            ],
+            "OriginalName": "winget.exe",
+            "OriginalCommandElements": [
+                "--version"
+            ]
+        },
+        {
+            "Verb": "Enable",
+            "Noun": "WinGetLocalManifest",
```
JohnMcPMS, line 17: Maybe the problem is that authoring it in Crescendo is difficult, but I know I wanted this to be something more like `Enable-WinGetSetting LocalManifestFiles` rather than having a cmdlet per setting.

I think I would rather not have the cmdlet at all than have this if we have to choose.

src/PowerShell/Microsoft.WinGet.Client/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Building
+### Pre-Requisite Steps
+
+A **PowerShell module** is merely a folder containing a **.psd1** file or a
+**.psm1** file. In this scenario, we will be building a folder with a **.psd1**
+file and all of the other necessary bits inside. But, before we do that, we need
+to make sure of the following steps.
+
+1. The project has been built for the **x64**, **x86**, **ARM64**, and **ARM**
```
JohnMcPMS, line 13: ARM should not be necessary.

src/PowerShell/Microsoft.WinGet.Client/README.md
JohnMcPMS: Agreed, it would be great if that were done and the directory published as an artifact so that it was fairly easy to test from any given build.  I think that might require custom work though given that the current pipeline involves splitting the jobs by architecture (?).

JohnMcPMS (changes requested):

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseFinderCommand.cs
```diff
@@ -0,0 +1,207 @@
...
+                    : string.Join(" ", this.Query);
+            }
+        }
+
+        private PackageFieldMatchOption ExactAsMatchOption
+        {
+            get
+            {
+                return this.Exact.ToBool()
+                    ? PackageFieldMatchOption.EqualsCaseInsensitive
```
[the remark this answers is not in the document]
jontab: Do you happen to know what the `--exact` flag does for the `winget install` command? If `winget install` is by default `EqualsCaseInsensitive`, does `--exact` just cause it to be `Exact` instead?

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseClientCommand.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        public BaseClientCommand()
+            : base()
+        {
+            var identity = WindowsIdentity.GetCurrent();
+            var principal = new WindowsPrincipal(identity);
+            var admin = principal.IsInRole(WindowsBuiltInRole.Administrator);
+            if (admin)
+            {
+                throw new Exception(@"This cmdlet is currently not available in an administrator context.");
```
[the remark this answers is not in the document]
JohnMcPMS: I think we need to include unique errors in the translation workflow separately.  For cherry picking some strings out from the main loc data, we probably have to write a script.  I don't know how the help gets localized though.

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseFinderCommand.cs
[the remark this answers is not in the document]
JohnMcPMS: `--exact` is the same as `PackageFieldMatchOption.Equals` for all cases.

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseInstallCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+            operation.Progress = (context, progress) =>
+            {
+                var record = new ProgressRecord(1, activity, progress.State.ToString())
+                {
+                    RecordType = ProgressRecordType.Processing,
+                };
+
+                if (progress.State == PackageInstallProgressState.Downloading)
+                {
+                    record.StatusDescription = $"{progress.BytesDownloaded / 1000000.0f:0.0} MB / {progress.BytesRequired / 1000000.0f:0.0} MB";
```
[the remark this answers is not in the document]
jontab: Okay, I've added a `BytesRequired != 0` check in a newer commit. That way the status should be preserved as `Downloading` but no progress bar will be shown.

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseFinderCommand.cs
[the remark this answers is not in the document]
jontab: Got it, should be fixed in a newer commit.

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseLifecycleCommand.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    return results[0].CatalogPackage;
+                }
+                else if (results.Count == 0)
+                {
+                    throw new RuntimeException(@"No package found matching input criteria.");
+                }
+                else
+                {
+                    throw new Errors.VagueCriteriaException(results[0].CatalogPackage, results[1].CatalogPackage);
```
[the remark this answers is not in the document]
jontab: There was no reason, I've updated it to attach all results to the exception and to preview the first two in the exception message.

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseLifecycleCommand.cs
```diff
@@ -0,0 +1,133 @@
...
+        private PackageVersionId GetPackageVersionId(CatalogPackage package)
+        {
+            if (this.Version != null)
+            {
+                var versions = package.AvailableVersions;
+
+                for (var i = 0; i < versions.Count; i++)
+                {
+                    // PackageVersionInfo.DisplayName (?)
+                    bool match = this.Version.Equals(
```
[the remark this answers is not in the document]
jontab: Got it, should be fixed.

src/PowerShell/Microsoft.WinGet.Client/Commands/BaseLifecycleCommand.cs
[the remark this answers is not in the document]
JohnMcPMS: Just showing two is the issue.  Maybe:
```
Package 1, Package 2, and 17 others.
```

src/PowerShell/Microsoft.WinGet.Client/Crescendo.json
jontab: A possible solution could be giving `Enable-WinGetSetting` a `Parameter(Mandatory, Position = 0)` parameter of something like `-Name` and then invoking `winget settings --enable $Name` which is possible in Crescendo.

src/PowerShell/Microsoft.WinGet.Client/BuildModule.ps1
```diff
@@ -0,0 +1,38 @@
...
+    [Parameter()]
+    [string]
+    $Out = "$PSScriptRoot\build\bin",
+
+    [Parameter()]
+    [string]
+    $Configuration = 'Release'
+)
+
+$CoreFramework = 'net6.0-windows10.0.22000.0'
```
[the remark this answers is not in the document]
jontab: Although I'm targeting `net5.0-windows10.0.22000.0` now, I've added a `TargetPlatformMinVersion` tag for `10.0.17763.0` as per [https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-enhance#supporting-multiple-windows-os-versions](this link) which should correspond with the minimum version for the CLI. I believe you have to target some sort of Windows TFM for CsWinRT and I had some weird issues targeting `net5.0-windows10.0.17763.0` outright. Theoretically, setting the `TargetPlatformMinVersion` should be enough.

src/PowerShell/Microsoft.WinGet.Client/README.md
jontab: I've made a first pass at adding it to the build pipeline.

src/PowerShell/Microsoft.WinGet.Client/Commands/UpdatePackageCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+            options.AllowUpgradeToUnknownVersion = this.IncludeUnknown.ToBool();
+            return options;
+        }
+
+        private InstallResult UpgradePackage(
+            CatalogPackage package,
+            InstallOptions options)
+        {
+            var operation = PackageManager.Value.UpgradePackageAsync(package, options);
+            return this.RegisterCallbacksAndWait(operation, $"Updating '{package.Name}'");
```
JohnMcPMS, line 57: Is this string shown to the user?

JohnMcPMS (approved):

src/PowerShell/Microsoft.WinGet.Client/Commands/UpdatePackageCommand.cs
jontab: Yes, I've just realized I should've put those into the .RESX. I've just done that.

src/PowerShell/Microsoft.WinGet.Client/Microsoft.WinGet.Client.csproj
```diff
@@ -67,5 +69,13 @@
     <!-- Ensure Support for Windows 10, Version 1809 -->
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
 	</PropertyGroup>
+
+	<Target Name="CopyWinRTAct" AfterTargets="Build" Condition="'$(TargetFramework)' != ''">
+		<ItemGroup>
+			<WinRTActDll Include="$(SolutionDir)_build\$(Platform)\Debug\winrtact.dll" />
```
JohnMcPMS, line 75: Did you also add a build dependency on that project at some point?

jontab: Yes, `winrtact.dll` comes from the `UndockedRegFreeWinRT` project (added as a `ProjectReference`) and across `Debug` and `Release`, it should only build in `Debug`. That should be set in the configuration manager. The `winrtact.dll` will end up being somewhere in the source directory with `$(Platform)` in its path, so I've just added an event to copy `**\$(Platform)\**\winrtact.dll` to the output directory (`net5.0-windows10.0.22000.0` and `net461`). It will probably find multiple `winrtact.dll` files. But since they should all be built for `$(Platform)` in `Debug`, they will just overwrite each other & it should be fine.

PR #2320, opened 2022-07-08
Implementation for Zip Install (Non-Portable)

ryfu-msft:
Related to:
#140 

Changes:
- Adds the initial implementation for supporting installing from a zip file that contains a nested installer 
- Does not add support for nested Portable packages which will be addressed in a separate PR due to complexity

Tests:
- Added E2E tests
- Added Unit tests for verifying zip workflow and archive extraction behavior

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2320)


src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -0,0 +1,50 @@
...
+    void ExtractInstallerFromArchive(Execution::Context& context)
+    {
+        const auto& installerPath = context.Get<Execution::Data::InstallerPath>();
+        const auto& installerParentPath = installerPath.parent_path();
+
+        HRESULT hr = AppInstaller::Archive::ExtractArchive(installerPath, installerParentPath);
+
+        if (SUCCEEDED(hr))
+        {
+            AICLI_LOG(CLI, Info, << "Successfully extracted archive");
```
Trenly, line 18: Can we log the path the archive is extracted to?

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -0,0 +1,50 @@
...
+            context.SetFlags(Execution::ContextFlag::InstallerExtractedFromArchive);
+        }
+        else
+        {
+            AICLI_LOG(CLI, Info, << "Failed to extract archive");
```
Trenly, line 23: Can we also log the path where it failed to extract to?

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
ryfu-msft: added path to log

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
ryfu-msft: added path to log

src/AppInstallerCLICore/Workflows/ArchiveFlow.h
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "ExecutionContext.h"
+
+namespace AppInstaller::CLI::Workflow
+{
+    void ExtractInstallerFromArchive(Execution::Context& context);
```
JohnMcPMS, line 8: See other workflow tasks for the comments to place above these.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -364,11 +377,22 @@ namespace AppInstaller::CLI::Workflow
...
             THROW_HR(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED));
         }
     }
 
+    void ArchiveInstall(Execution::Context& context)
+    {
+        context <<
+            ExtractInstallerFromArchive <<
+            VerifyAndSetNestedInstaller <<
+            ExecuteInstaller;
```
JohnMcPMS, line 393: This flow will be less efficient for portables; we should just extract directly to the final location.  Not sure if that works great with the database tracking files though.  We might need to be less efficient just so any files we could lose track of are in `%TEMP%` and will be more readily cleaned up.

JohnMcPMS, line 393: My gut is to create `ExecuteInstallerForType(InstallerType)` and then have `ExecuteInstaller` call it with `installer.InstallerType` and this call it with `installer.NestedInstallerType` rather than the extra state management.

src/AppInstallerCommonCore/Archive.cpp
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+
+namespace AppInstaller::Archive
+{
+    HRESULT ExtractArchive(const std::filesystem::path& archivePath, const std::filesystem::path& destPath)
+    {
+        IFileOperation* pFileOperation;
```
JohnMcPMS, line 9: Every one of these COM pointers should be `wil::com_ptr<Interface>`.

src/AppInstallerCommonCore/Archive.cpp
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+
+namespace AppInstaller::Archive
+{
+    HRESULT ExtractArchive(const std::filesystem::path& archivePath, const std::filesystem::path& destPath)
+    {
+        IFileOperation* pFileOperation;
+        HRESULT hr = CoCreateInstance(CLSID_FileOperation, NULL, CLSCTX_ALL, IID_PPV_ARGS(&pFileOperation));
```
JohnMcPMS, line 10:
```suggestion
        RETURN_IF_FAILED(CoCreateInstance(CLSID_FileOperation, NULL, CLSCTX_ALL, IID_PPV_ARGS(&pFileOperation)));
```
Same everywhere else.

src/AppInstallerCommonCore/Archive.cpp
```diff
@@ -0,0 +1,68 @@
...
+                                        (StrRetToBuf(&strFolderName, pidlChild, szFolderName, MAX_PATH) == S_OK))
+                                    {
+                                        hr = SHCreateItemWithParent(pidlParent, pArchiveShellFolder, pidlChild, IID_PPV_ARGS(&pShellItemFrom));
+                                        if (SUCCEEDED(hr))
+                                        {
+                                            hr = pFileOperation->CopyItem(pShellItemFrom, pShellItemTo, NULL, NULL);
+                                        }
+                                        pShellItemFrom->Release();
+                                    }
+                                    ILFree(pidlChild);
```
JohnMcPMS, line 50: See if there are already `wil` resource definitions for these types, and if not create a `using` definition for a `wil::unique_any`.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: I am thinking that portables will require more tasks than the one I have currently specified. 
Here is what I am proposing when we start the implementation for portable(s) in an archive. 

- `ExtractInstallerFromArchive` (now called `ExtractFilesFromArchive`) will have a separate flow for portables that will copy the extracted files to the final install location
- `TryExtractArchive` will have an additional parameter to indicate that will append to the database to keep track of the files that are extracted and copied. 
- The install location will need to have the directory created prior to extraction, so this will likely need to be its own step that creates the InstallDirectory and the local db file if it does not exist already. So something like `SetupPortableInstallDirectory` that we call prior to all of these steps if the `NestedInstallerType` is `Portable`

Once we enter the portable install flow, we can then use `IsArchiveType` to check whether we should install multiple portables (I plan to use subcontexts to handle multiple installs) for a given archive package.

src/AppInstallerCLICore/ExecutionContext.h
```diff
@@ -64,6 +64,7 @@ namespace AppInstaller::CLI::Execution
         // TODO: Remove when the source interface is refactored.
         TreatSourceFailuresAsWarning = 0x10,
         ShowSearchResultsOnPartialFailure = 0x20,
+        InstallerExtractedFromArchive = 0x40,
```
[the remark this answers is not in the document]
ryfu-msft: Removed `InstallerExtractedFromArchive` context flag. 

Added `ExecuteInstallerForInstallerType` which selects the appropriate install flow based on the installerType to handle whether to use `InstallerType` or `NestedInstallerType`

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "winget/Archive.h"
+
+namespace AppInstaller::CLI::Workflow
+{
+    void ExtractInstallerFromArchive(Execution::Context& context)
+    {
+        const auto& installerPath = context.Get<Execution::Data::InstallerPath>();
+        const auto& installerParentPath = installerPath.parent_path();
+
+        HRESULT hr = AppInstaller::Archive::ExtractArchive(installerPath, installerParentPath);
```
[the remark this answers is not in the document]
ryfu-msft: Changed to suggested

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+        if (SUCCEEDED(hr))
+        {
+            AICLI_LOG(CLI, Info, << "Successfully extracted archive to: " << installerParentPath );
+            context.SetFlags(Execution::ContextFlag::InstallerExtractedFromArchive);
+        }
+        else
+        {
+            AICLI_LOG(CLI, Info, << "Failed to extract archive to: " << installerParentPath);
```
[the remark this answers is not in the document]
ryfu-msft: Added logging for `hr`

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -0,0 +1,50 @@
...
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_EXTRACT_ARCHIVE_FAILED);
+        }
+    }
+
+    void VerifyAndSetNestedInstaller(Execution::Context& context)
+    {
+        const auto& installer = context.Get<Execution::Data::Installer>().value();
+        const auto& installerPath = context.Get<Execution::Data::InstallerPath>();
+        const auto& installerParentPath = installerPath.parent_path();
+        const auto& relativeFilePath = ConvertToUTF16(installer.NestedInstallerFiles[0].RelativeFilePath);
```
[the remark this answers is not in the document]
ryfu-msft: Added a check to verify that `NestedInstallerFiles` is not empty

src/AppInstallerCLICore/Workflows/ArchiveFlow.h
ryfu-msft: Added comments to workflow task

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Added `ExecuteInstallerForType(Context, InstallerType)` to handle the install flow selection logic based on installerType

src/AppInstallerCommonCore/Archive.cpp
ryfu-msft: Changed to using `wil::com_ptr<Interface>`

src/AppInstallerCommonCore/Archive.cpp
ryfu-msft: Changed to suggested

src/AppInstallerCommonCore/Archive.cpp
ryfu-msft: Created a `using` definition for `PIDLIST_ABSOLUTE` and `LPITEMIDLIST` and removed free statements

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -318,22 +318,10 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
-    void ExecuteInstaller(Execution::Context& context)
+    void ExecuteInstallerForType(Execution::Context& context, InstallerTypeEnum installerType)
```
JohnMcPMS, line 321: The intention was to create a `WorkflowTask` derivative.  Example:
https://github.com/microsoft/winget-cli/blob/18ee6999f442d1b9f6015cefc4ad22804e744978/src/AppInstallerCLICore/Workflows/InstallFlow.h#L42

That allows the continued use of the stream operators rather than checking `IsTerminated`.

src/AppInstallerCommonCore/Archive.cpp
```diff
@@ -4,65 +4,43 @@
...
+
+        unique_lpitemidlist pidlChild;
+        ULONG nFetched;
+        while (pEnumIdList->Next(1, wil::out_param_ptr<LPITEMIDLIST*>(pidlChild), &nFetched) == S_OK && nFetched == 1)
         {
-            hr = pFileOperation->SetOperationFlags(FOF_NO_UI);
-            if (SUCCEEDED(hr))
+            wil::com_ptr<IShellItem> pShellItemFrom;
+            STRRET strFolderName;
+            WCHAR szFolderName[MAX_PATH];
```
JohnMcPMS, line 34: Are the paths actually limited to `MAX_PATH`?  If not, I would prefer if we could do a variable length path so that we don't hit a length issue here.

JohnMcPMS (changes requested):

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1370,4 +1370,8 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="NestedInstallerNotFound" xml:space="preserve">
     <value>Nested installer file does not exist. Ensure the specified relative path of the nested installer matches: </value>
   </data>
+  <data name="NestedInstallerFilesNotSpecified" xml:space="preserve">
+    <value>No nested installer files specified; ensure that your manifest includes an entry for NestedInstallerFiles</value>
```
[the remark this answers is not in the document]
ryfu-msft: Added a comment clarifying that this check should be prevented by manifest validation and removed localized string.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Created a `WorkflowTask` derivative for ExecuteInstallerForType

src/AppInstallerCommonCore/Archive.cpp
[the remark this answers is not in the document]
ryfu-msft: STRRET is has a max buffer length of MAX_PATH but this should be okay as it is only the name of the file

JohnMcPMS (approved):

PR #2321, opened 2022-07-09
Add URL Handler to pass winget commands from a browser

Trenly:

This change would allow publishers to provide direct links to install with winget in the form `winget:install Package.ID`
This also allows users to run any winget commands from their browser using `winget:<command>`

I'm almost certain I've done quite a few things wrong here, but just trying to use it as a learning opportunity for myself. If this isn't something which the team wants to support/implement yet, let me know and I'll close the PR.

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2321)

### DIFF
--- a/src/AppInstallerCLICore/Core.cpp
+++ b/src/AppInstallerCLICore/Core.cpp
@@ -80,12 +80,30 @@ namespace AppInstaller::CLI
         Logging::BeginLogFileCleanup();
 
         context << Workflow::ReportExecutionStage(Workflow::ExecutionStage::ParseArgs);
-
-        // Convert incoming wide char args to UTF8
         std::vector<std::string> utf8Args;
-        for (int i = 1; i < argc; ++i)
+
+        if (wcsncmp(argv[1], L"winget:", 7) == 0) {
+            // Called by URI
+            std::wstring argString = argv[1];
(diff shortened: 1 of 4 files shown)


lechacon:
I like the idea of a URI scheme, the MS Store does something similar and so does the App Installer. But we need to think this very thoroughly. We probably need to consider the security implications of allowing people to create links that install stuff in the computers of whoever clicks them, we'd at least need to add a confirmation prompt.

I think having a link in the web open an app is something that would be more at home in an app with a GUI than in a CLI because anybody could open it and I imagine most people would not know what is happening if a command prompt shows up.

The scheme should support only specific commands, not any arbitrary one. For example, I don't see the point of `winget:source list`.

When we introduce new URI schemes, we also have to do an internal review. The scheme should be prefixed with `ms-`, like `ms-appinstaller` or `ms-edge`. We probably would have to use a different syntax than with the CLI. For example, `winget:install My.App --locale es-ES` doesn't really look like an URI, and we'd have to make up a new syntax, something like `winget:install/My.App?locale=es-ES`.

I think this could be done with a separate GUI that used the COM API, instead of adding it to the CLI.

Trenly:
All great points which I hadnt considered; I think that right now, this is something which needs more refinement

PR #2324, opened 2022-07-11
Explicit ACLs

JohnMcPMS:
## Change
Explicitly apply ACLs to some locations.  All are consistent with the default owners of these locations (on the current versions of Windows supported).

The locations being affected are:
- Temp :: Giving the current user full control (this has affected users in the past if their temp location did not allow Execute)
- SecureSettings :: Admins own and user can read.  This is how it is now, but being explicit.
- LocalState/StandardSettings/UserFileSettings :: Current user owns.  Again, not changed but being explicit.

## Validation
Tried scenarios manually to ensure functionality.
Added a few tests to ensure that things beyond the current scenarios work as well.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2324)


src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -269,217 +329,282 @@ namespace AppInstaller::Runtime
...
-        bool create = true;
+        // Could be expanded to actually check the current owner/ACL on the path, but isn't worth it currently
+        return (CurrentUser != ACEPermissions::None || Admins != ACEPermissions::None);
+    }
 
-#ifndef WINGET_DISABLE_FOR_FUZZING
-        if (IsRunningInPackagedContext())
+    void PathDetails::ApplyACL() const
+    {
+        ULONG entries = 0;
```
yao-msft, line 340: >entries [](http://example.com/codeflow?start=14&length=7)

nit: entriesCount ?

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -269,217 +329,282 @@ namespace AppInstaller::Runtime
...
+            entry.Trustee.ptstrName = reinterpret_cast<LPWCH>(userToken->User.Sid);
+        }
+
+        if (Admins != ACEPermissions::None)
+        {
+            if (Admins == ACEPermissions::Owner)
             {
-                result = GetPathToUserTemp();
-                result /= s_DefaultTempDirectory;
+                ownerSID = adminSID.get();
```
yao-msft, line 374: >ownerSID = adminSID.get(); [](http://example.com/codeflow?start=16&length=26)

Do we allow both User and Admins to declare owner? Looks like if both are declared owner, the Admin one will override CurrentUser one. Is this intentional or we should return e_invalidarg for such input?

yao-msft (approved):
:shipit:

PR #2330, opened 2022-07-12
Add UndockedRegFreeWinRT as subtree

jontab:
This adds the `src/UndockedRegFreeWinRT` directory from the `master` branch of https://github.com/microsoft/xlang as a subtree in the `src` directory.


JohnMcPMS:
> Out of curiosity, is there a reason for including the entire source tree instead of just using the NuGet package?

Because we are going to make changes to it as it doesn't actually work for the scenario.  This is a stopgap for either coming to a design that is supported more broadly (and then we can use the NuGet) or implementing one of the other options for getting the COM proxies working.

JohnMcPMS:
> Out of curiosity, is there a reason for including the entire source tree instead of just using the NuGet package?

Because we are going to make changes to it as it doesn't actually work for the scenario.  This is a stopgap for either coming to a design that is supported more broadly (and then we can use the NuGet) or implementing one of the other options for getting the COM proxies working.

JohnMcPMS:
@jontab We typically subtree these into a subdirectory under the `src/FOO` location so that we can put our files (like the project) in that root.  I think given that we fully intend to make changes to this subtree it is ok to do it the way you have it now, just letting you know.

JohnMcPMS:
Done by #2333

PR #2342, opened 2022-07-15
Prevent access to parent directories from relativeFilePath for archive install

ryfu-msft:
Fixes a security issue where the relativeFilePath could be exploited to execute files outside of the directory containing the installer. 

Examples: (../../relativeFilePath)

Changes:
- Adds checks prior to executing the install to verify that the metadata for NestedInstallerfiles is valid
- Shows appropriate error message to user if checks fail.
Test:
- Unit test for verifying relativefilePath with escape string
- Unit test for verifying if no NestedInstallers are specified.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2342)


src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -4,8 +4,18 @@
...
 
+using namespace AppInstaller::Manifest;
+
 namespace AppInstaller::CLI::Workflow
 {
+    namespace
+    {
+        bool RelativePathContainsEscapeString(const std::string& relativePath)
+        {
+            return relativePath.find("..") != std::string::npos;
```
lechacon, line 15: Is there no API for this kind of check? It would be good to not have to think of every possible case ourselves.

Maybe instead of checking that the relative path doesn't contain anything like this, you could check that the final path looks like you expect. Like `(basePath + relativePath).resolve().starts_with(basePath)` (made up syntax).

ryfu-msft: I didn't find any API that could check for this specifically. From what I was able to find, it seems like ".." was the only string that could be used to access parent directories.  

Regarding the check, the reason why I chose to validate just the RelativeFilePath is because this is the only input the user could have to affect the installerPath since the basePath is created by us. This also allows us to run checks without having to start the download/install flow. However, if you feel strongly about this, I could also add this check to the fullPath once available in the install flow to be extra safe.

yao-msft: [std::filesystem:canonical](https://en.cppreference.com/w/cpp/filesystem/canonical) will help resolve . and .. and even symbolic link for us.
It may be also helpful to move the related "check if a relative path is still under expected folder" code to filesystem.cpp under CommonCore for future "launch" related work(which also needs this check I think)

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -56,4 +66,31 @@ namespace AppInstaller::CLI::Workflow
             context.Add<Execution::Data::InstallerPath>(nestedInstallerPath);
         }
     }
+
+    void EnsureValidNestedInstallerMetadataForArchiveInstall(Execution::Context& context)
+    {
+        auto installer = context.Get<Execution::Data::Installer>().value();
+
+        if (installer.InstallerType == InstallerTypeEnum::Zip)
```
yao-msft, line 74: >installer.InstallerType == InstallerTypeEnum::Zip [](http://example.com/codeflow?start=12&length=49)

use IsArchiveType()

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -105,6 +105,7 @@
 #define APPINSTALLER_CLI_ERROR_BIND_WITH_EMBEDDED_NULL          ((HRESULT)0x8A15005A)
 #define APPINSTALLER_CLI_ERROR_NESTEDINSTALLER_NOT_FOUND        ((HRESULT)0x8A15005B)
 #define APPINSTALLER_CLI_ERROR_EXTRACT_ARCHIVE_FAILED           ((HRESULT)0x8A15005C)
+#define APPINSTALLER_CLI_ERROR_NESTEDINSTALLER_INVALID_PATH     ((HRESULT)0x8A15005D)
```
yao-msft, line 108: >APPINSTALLER_CLI_ERROR_NESTEDINSTALLER_INVALID_PATH [](http://example.com/codeflow?start=8&length=51)

nit: please add missing error codes description to errors.cpp

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
Trenly: I also feel like a std::find may cause a very odd edge case where a publisher or an error in their build pipeline may create a filename like `File.Major.Minor..Build`

It may be a good use case not only for canonical as Yao suggested, but perhaps even [std::filesystem:relative](https://en.cppreference.com/w/cpp/filesystem/relative)

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -56,4 +66,31 @@ namespace AppInstaller::CLI::Workflow
...
+        {
+            auto const& nestedInstallerFiles = installer.NestedInstallerFiles;
+            if (nestedInstallerFiles.empty())
+            {
+                AICLI_LOG(CLI, Error, << "No entries specified for NestedInstallerFiles");
```
Trenly, line 79: Do we need to ensure length == 1 for NestedInstallerType != portable here?

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
ryfu-msft: Created a method called `PathEscapesBaseDirectory` in `Filesystem.cpp` that checks if the canonical form of a path fails to contain the base path. If true, then we know that the path points to a location outside of the provided base directory

src/AppInstallerCommonCore/Filesystem.cpp
```diff
@@ -62,6 +62,14 @@ namespace AppInstaller::Filesystem
         return (GetVolumeInformationFlags(path) & FILE_SUPPORTS_REPARSE_POINTS) != 0;
     }
 
+    bool PathEscapesBaseDirectory(const std::filesystem::path& path, const std::filesystem::path& base)
+    {
+        // Input for path must exist or this will throw a filesystem error.
+        const auto& pathString = std::filesystem::canonical(path).u8string();
```
JohnMcPMS, line 68:
```suggestion
        const auto& pathString = std::filesystem::weakly_canonical(path).u8string();
```
Using `weakly_canonical` would make this more usable in the abstract case.  It could also be `lexically_normal`, which doesn't even care about whether any files exist or not.  But that doesn't prevent the potential for symlinks to enable escaping the location (although I don't know if it is possible to have a symlink in an archive file...).

src/AppInstallerCommonCore/Filesystem.cpp
```diff
@@ -62,6 +62,14 @@ namespace AppInstaller::Filesystem
         return (GetVolumeInformationFlags(path) & FILE_SUPPORTS_REPARSE_POINTS) != 0;
     }
 
+    bool PathEscapesBaseDirectory(const std::filesystem::path& path, const std::filesystem::path& base)
+    {
+        // Input for path must exist or this will throw a filesystem error.
+        const auto& pathString = std::filesystem::canonical(path).u8string();
+        const auto& baseString = base.lexically_normal().u8string();
+        return pathString.find(baseString) == std::string::npos;
```
JohnMcPMS, line 70: From the description on how `lexically_relative` is implemented, we should use `mismatch` here instead:
```
auto [a, b] = mismatch(path.begin(), path.end(), base.begin(), base.end());
return (b != base.end());
```
Which I believe ensures that the paths match at least until the end of `base`.

src/AppInstallerCommonCore/Filesystem.cpp
```diff
@@ -62,6 +62,14 @@ namespace AppInstaller::Filesystem
         return (GetVolumeInformationFlags(path) & FILE_SUPPORTS_REPARSE_POINTS) != 0;
     }
 
+    bool PathEscapesBaseDirectory(const std::filesystem::path& path, const std::filesystem::path& base)
```
JohnMcPMS, line 65: Explicit unit tests for this function please.

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -32,28 +35,56 @@ namespace AppInstaller::CLI::Workflow
...
-        std::filesystem::path nestedInstallerPath = installerParentPath / relativeFilePath;
+        const std::filesystem::path& nestedInstallerPath = installerParentPath / relativeFilePath;
 
         if (!std::filesystem::exists(nestedInstallerPath))
         {
             AICLI_LOG(CLI, Error, << "Unable to locate nested installer at: " << nestedInstallerPath);
             context.Reporter.Error() << Resource::String::NestedInstallerNotFound << ' ' << nestedInstallerPath << std::endl;
             AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_NESTEDINSTALLER_NOT_FOUND);
         }
+        else if (Filesystem::PathEscapesBaseDirectory(nestedInstallerPath, installerParentPath))
```
JohnMcPMS, line 55: The manifest validation should also do this kind of check to doubly prevent it from happening (and to prevent a manifest with such a thing from getting into `winget-pkgs`).

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Filesystem.cpp
ryfu-msft: Added unit test for this function

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
ryfu-msft: Added a manifest validation check for checking the relative path using the std::current_path as the base path. I also rearranged the order of the checks above so that verifying the relativeFilePath comes before checking if the nestedInstaller exists as that logically makes more sense.

src/AppInstallerCommonCore/Filesystem.cpp
ryfu-msft: Didn't know about mismatch... learned something new :) Changed to suggested.

src/AppInstallerCLITests/Filesystem.cpp
```diff
@@ -0,0 +1,30 @@
...
+{
+    TestCommon::TempDirectory tempDirectory("TempDirectory");
+    const std::filesystem::path& basePath = tempDirectory.GetPath();
+
+    std::string badRelativePath = "../../target.exe";
+    std::string badRelativePath2 = "test/../../target.exe";
+    std::string goodRelativePath = "target.exe";
+    std::string goodRelativePath2 = "test/../test1/target.exe";
+
+    std::filesystem::path badPath = basePath / ConvertToUTF16(badRelativePath);
```
JohnMcPMS, line 22: Converting to UTF-16 is necessary here because you know that the input is ASCII.

src/AppInstallerCommonCore/Filesystem.cpp
```diff
@@ -64,10 +64,9 @@ namespace AppInstaller::Filesystem
 
     bool PathEscapesBaseDirectory(const std::filesystem::path& path, const std::filesystem::path& base)
     {
-        // Input for path must exist or this will throw a filesystem error.
-        const auto& pathString = std::filesystem::canonical(path).u8string();
-        const auto& baseString = std::filesystem::absolute(base).u8string();
-        return pathString.find(baseString) == std::string::npos;
+        const auto& canonicalPath = std::filesystem::weakly_canonical(path);
+        auto [a, b] = std::mismatch(canonicalPath.begin(), canonicalPath.end(), base.begin(), base.end());
```
JohnMcPMS, line 68: You do need to use `weakly_canonical` on the base path as well, or a symlink in the base path could result in a false negative result.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Filesystem.cpp
ryfu-msft: Converted base path to weakly_canonical prior to comparison

JohnMcPMS (approved):

PR #2347, opened 2022-07-19
Server certificate pinning for Store source

JohnMcPMS:
## Change
This change adds a generic certificate chain verification infrastructure for pinning certificate chains. It is specifically used to pin the Microsoft Store source by default. More sources may be pinned later, but currently the packaged index is less in need of it because it is already signed.

The pinning configuration consists of 1 or more chains, only one of which needs to successfully validate the incoming certificate. This allows for rolling to a new certificate when needed. Each chain consists of a fixed set of certificates, which can each be configured to validate any or all of the following properties:

- Public Key
- Subject
- Issuer

If the certificate is configured to validate none of the values, it will allow any certificate through.

An admin setting is added to disable pinning, both as an emergency measure in the event that there is a bug or rolled certificate that was not communicated, but also because there are test scenarios where the user actively wants to disable it (HTTPS redirection via something like Fiddler).

The configuration can be loaded from JSON for future dynamic configuration, but it is currently only as a test hook to enable configuration via Group Policy.

In order to better secure the source by default, reconfiguring (remove then add) the Store source manually will convert it back to the built-in values. This includes the pinning configuration.

It was necessary to modify the cpprestsdk subtree to add a new callback. This enables the request handle to be passed back to our code when the server certificate is first available. We can then check the server certificate against the configured pinning chain, making a decision to terminate the request before it is sent.

## Validation
Unit tests are added for the various utilities and the primary pinning configuration/validation.
An E2E test is added for a negative case with the Store, but more could be done to test REST in general in our CI pipeline.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2347)


src/CertificateResources/resource.h
```diff
@@ -0,0 +1,14 @@
+//{{NO_DEPENDENCIES}}
+// Microsoft Visual C++ generated include file.
+// Used by WinGetSchemas.rc
```
yao-msft, line 3: >WinGetSchemas.rc [](http://example.com/codeflow?start=11&length=16)

nit: CertificateResources.rc

src/cpprestsdk/cpprestsdk/Release/src/http/client/http_client_winhttp.cpp
```diff
@@ -2039,6 +2039,18 @@ class winhttp_client final : public _http_client_communicator
             case WINHTTP_CALLBACK_STATUS_SENDING_REQUEST:
             {
                 p_request_context->on_send_request_validate_cn();
+
+                try
```
yao-msft, line 2043: >try [](http://example.com/codeflow?start=16&length=3)

Are we directly modifying subtree modules?

JohnMcPMS: The upstream codebase is in maintenance mode.  We can try to get this change in there but it seems unlikely, and very unlikely to be able to do so in the time frame we need.

src/AppInstallerCommonCore/AdminSettings.cpp
```diff
@@ -180,11 +195,14 @@ namespace AppInstaller::Settings
...
-        if (setting == AdminSetting::LocalManifestFiles &&
-            GroupPolicies().GetState(TogglePolicy::Policy::LocalManifestFiles) == PolicyState::Enabled)
+        // Check for a policy that overrides this setting.
+        if (setting == AdminSetting::LocalManifestFiles)
         {
-            return true;
+            PolicyState localManifestFilesPolicy = GroupPolicies().GetState(TogglePolicy::Policy::LocalManifestFiles);
+            if (localManifestFilesPolicy != PolicyState::NotConfigured)
+            {
+                return (localManifestFilesPolicy == PolicyState::Enabled ? true : false);
```
yao-msft, line 204: >return (localManifestFilesPolicy == PolicyState::Enabled ? true : false); [](http://example.com/codeflow?start=16&length=73)

nit: return localManifestFilesPolicy == PolicyState::Enabled

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -196,6 +196,19 @@ namespace AppInstaller::Settings
                 return std::nullopt;
             }
 
+#ifndef AICLI_DISABLE_TEST_HOOKS
+            // Enable certificate pinning configuration through GP sources for testing
+            std::string pinningConfigurationName = "CertificatePinning";
```
yao-msft, line 201: >std::string [](http://example.com/codeflow?start=12&length=11)

nit const

src/AppInstallerCommonCore/GroupPolicy.cpp
```diff
@@ -196,6 +196,19 @@ namespace AppInstaller::Settings
                 return std::nullopt;
             }
 
+#ifndef AICLI_DISABLE_TEST_HOOKS
```
yao-msft, line 199: >AICLI_DISABLE_TEST_HOOKS [](http://example.com/codeflow?start=8&length=24)

Are we planning to expose this in the future?

src/AppInstallerCommonCore/Certificates.cpp
```diff
@@ -0,0 +1,511 @@
...
+        return *this;
+    }
+
+    PinningDetails& PinningDetails::SetPinning(PinningVerificationType type)
+    {
+        m_pinning = type;
+        return *this;
+    }
+
+    bool PinningDetails::LoadFrom(const Json::Value& configuration)
```
yao-msft, line 106: >PinningDetails::LoadFrom [](http://example.com/codeflow?start=9&length=24)

A comment with sample configuration json would better help with reading the parsing logic

src/AppInstallerCommonCore/Certificates.cpp
```diff
@@ -0,0 +1,511 @@
...
+    {
+        std::string validationName = "Validation";
```
yao-msft, line 108: >std::string [](http://example.com/codeflow?start=8&length=11)

nit: const or declare above for all the json node names across this function?

src/AppInstallerCommonCore/Certificates.cpp
```diff
@@ -0,0 +1,511 @@
...
+    bool PinningDetails::Validate(PCCERT_CONTEXT certContext) const
+    {
+        if (WI_IsFlagSet(m_pinning, PinningVerificationType::PublicKey))
+        {
+            if (!CertComparePublicKeyInfo(
+                X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
+                &m_certificateContext.get()->pCertInfo->SubjectPublicKeyInfo,
+                &certContext->pCertInfo->SubjectPublicKeyInfo))
+            {
+                AICLI_LOG(Core, Verbose, << "Public key mismatch: Expected certificate [" << GetSimpleDisplayName(m_certificateContext.get()) << "], Actual certificate [" << GetSimpleDisplayName(certContext) << "]");
```
yao-msft, line 172: >Verbose [](http://example.com/codeflow?start=32&length=7)

Should these logging levels at least warning?

src/AppInstallerCommonCore/Certificates.cpp
```diff
@@ -0,0 +1,511 @@
...
+        THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), m_chain.empty());
+        return { const_cast<std::vector<PinningDetails>&>(m_chain), 0 };
+    }
+
+    bool PinningChain::Validate(PCCERT_CHAIN_CONTEXT chainContext) const
+    {
+        if (m_chain.empty())
+        {
+            // An empty chain rejects all inputs.
+            AICLI_LOG(Core, Verbose, << "Empty pinning chain blindly rejecting chain context");
```
yao-msft, line 254: >Verbose [](http://example.com/codeflow?start=28&length=7)

I think logging level of this function should be revisited?

src/AppInstallerCLITests/GroupPolicy.cpp
```diff
@@ -266,6 +273,54 @@ TEST_CASE("GroupPolicy_Sources", "[groupPolicy]")
...
+        auto getBytesString = [](const PinningDetails& details)
+        {
+            std::vector<BYTE> bytes;
+            bytes.assign(details.GetCertificate()->pbCertEncoded, details.GetCertificate()->pbCertEncoded + details.GetCertificate()->cbCertEncoded);
+            return AppInstaller::Utility::ConvertToUTF16(AppInstaller::Utility::ConvertToHexString(bytes));
+        };
+
+        std::wostringstream pinningConfig;
+        pinningConfig << LR"({
+"Chains": [{
```
yao-msft, line 298: nit: seems bad indentation

yao-msft (approved):
:shipit:

src/AppInstallerCommonCore/GroupPolicy.cpp
JohnMcPMS: Possibly, but I want that to be an explicit decision, not part of this change.

src/AppInstallerCommonCore/Certificates.cpp
JohnMcPMS: I chose not to log them by default for the case of multiple chains.  It is expected that only one of the chains would be valid for any given connection, so logging "Hey there is seemingly a security issue over here!" is not warranted when it turns out that it is actually just the one not being used.

`PinningConfiguration::Validate`, which is the expected entry point for any non-test use case, does log at `Error` on a failure to find any good chain.  If that is happening, then the user can manually gather a verbose log.

src/AppInstallerCommonCore/Certificates.cpp
JohnMcPMS: I changed this line specifically to `Warning` as an empty chain is not an expected use case.

src/AppInstallerCLITests/GroupPolicy.cpp
JohnMcPMS: It is a raw string, but I can bring the beginning down to make it look less awkward.

PR #2352, opened 2022-07-21
Move to using sqlite3_errmsg to extract a contextual error for SQLite failures

JohnMcPMS:
## Change
Use `sqlite3_errmsg` to get an error message with actual context as opposed to the string for the error code.  We can always look up the error code string if needed, and we will get the string for the HRESULT version in the logs as well.

For instance, the message for the new test is now:
```
Msg:[no such table: simpletest2]
```
Rather than the prior "Bad SQL" type default value.

## Validation
All unit tests pass, manually confirmed more useful error messages in at least some cases.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2352)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -117,6 +117,7 @@ enr
 enums
 EQU
 ERANGE
+errmsg
 errno
 ESRB
 etest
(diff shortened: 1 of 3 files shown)


yao-msft (approved):
:shipit:

PR #2378, opened 2022-07-26
Logging improvements

JohnMcPMS:
## Change
Add a connection identifier and log it when opening the connection and with each statement.  This makes it somewhat easier to follow the (extremely) verbose SQLite logging.

Also changes the file logger to append by default, enabling logging to the same file repeatedly a possibility (a util scenario mostly).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2378)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -291,6 +291,7 @@ NX
 objbase
 objidl
 ofile
+openmode
 Outptr
 packageinuse
 PACL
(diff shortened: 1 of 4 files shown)


Trenly:
> Also changes the file logger to append by default, enabling logging to the same file repeatedly a possibility (a util scenario mostly).

Just clarifying - This will still create a new log file for each command, but just enables future possibility to log to the same file, correct?

JohnMcPMS:
Correct, there should be no change at all to the behavior of `winget.exe` here.  This is only to make some scenarios work better for other places were we use the code.

yao-msft (approved):

PR #2383, opened 2022-07-27
Remove scope filter from being applied to portables

ryfu-msft:
If the --scope argument is present when installing a portable package, the command will fail if the incorrect scope is not present for a portable installer. This is the wrong behavior as the scope field in the manifest should not apply to portable packages. 

To resolve this issue, I have removed the applied scope filter if the installerType is portable (and also for msix). 

Changes:
- Bypass scope filter if the installertype is portable (or msix) using a semantic function
- Added an admin requirement to portable install if the scope is Machine
- Fixed several discrepancies in the portableAppUserRoot and portableAppMachineRoot naming/documentation
- Added unit and E2E tests to verify that portable install works as expected when scope arguments are provided.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2383)


src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -481,6 +481,14 @@ namespace AppInstaller::Manifest
             );
     }
 
+    bool DoesInstallerTypeIgnoreScopeFromManifest(InstallerTypeEnum installerType)
+    {
+        return (
```
Trenly, line 486: Should this also check nested installer type if a package is an archive, or does that not matter here?

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -453,16 +453,16 @@ namespace AppInstaller::Runtime
                 result.Path /= s_PortablePackagesDirectory;
             }
             break;
-        case PathName::PortablePackageMachineRootX64:
+        case PathName::PortableAppMachineRootX64:
             result.Path = Settings::User().Get<Setting::PortableAppMachineRoot>();
             if (result.Path.empty())
             {
-                result.Path = GetKnownFolderPath(FOLDERID_ProgramFilesX64);
+                result.Path = GetKnownFolderPath(FOLDERID_ProgramFiles);
```
ryfu-msft, line 460: `FOLDERID_ProgramFilesX64` does not work for x86 version. We will have to resort to using `FOLDERID_ProgramFiles`.

doc/Settings.md
```diff
@@ -58,25 +58,25 @@ The `disableInstallNotes` behavior affects whether installation notes are shown
     },
 ```
 
-### Portable Package User Root
-The `PortablePackageUserRoot` setting affects the default root directory where packages are installed to under `User` scope. This setting only applies to packages with the `portable` installer type. Defaults to `%LOCALAPPDATA%/Microsoft/WinGet/Packages/` if value is not set or is invalid.
+### Portable App User Root
+The `portableAppUserRoot` setting affects the default root directory where packages are installed to under `User` scope. This setting only applies to packages with the `portable` installer type. Defaults to `%LOCALAPPDATA%/Microsoft/WinGet/Packages/` if value is not set or is invalid.
```
JohnMcPMS, line 62: I explicitly asked for this to be package last time, why the change?

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
JohnMcPMS: Indeed, it might be productive to make all of these semantic functions take in an `Installer` rather than a type so that we force it the check to be done properly.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -342,7 +342,7 @@ namespace AppInstaller::CLI::Workflow
 
             InapplicabilityFlags IsApplicable(const Manifest::ManifestInstaller& installer) override
             {
-                if (m_requirement == Manifest::ScopeEnum::Unknown || installer.Scope == m_requirement)
+                if (m_requirement == Manifest::ScopeEnum::Unknown ||installer.Scope == m_requirement || DoesInstallerTypeIgnoreScopeFromManifest(installer.InstallerType))
```
JohnMcPMS, line 345: Also need to change `InstalledScopeFilter` similarly, or upgrade won't work.  Given that you have an upgrade test, it might mean instead that you need to look into why it is working when it shouldn't be...

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -481,6 +481,14 @@ namespace AppInstaller::Manifest
             );
     }
 
+    bool DoesInstallerTypeIgnoreScopeFromManifest(InstallerTypeEnum installerType)
+    {
+        return (
+            installerType == InstallerTypeEnum::Msix ||
```
JohnMcPMS, line 487: I don't think `MSIX` is ready to be there yet, I was just saying that I want to include it eventually (when the work to actually support machine install is done).

src/AppInstallerCommonCore/Runtime.cpp
JohnMcPMS: When would you be installing an x64 portable on an x86 machine?

JohnMcPMS (changes requested):

JohnMcPMS (approved):

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -634,6 +659,7 @@ namespace AppInstaller::CLI::Workflow
         {
             context <<
                 EnsureSymlinkCreationPrivilege <<
+                EnsureRunningAsAdminForMachineScopeInstall <<
```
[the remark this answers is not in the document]
JohnMcPMS: It is less efficient, but `EnsureSymlinkCreationPrivilege` is planned to be temporary, so it is easier to keep it separated.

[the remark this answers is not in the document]
ryfu-msft: `EnsureSymlinkCreationPrivilege` is meant to be a temporary check to handle the special case where a user may have developer mode off.  I expect that this check will be removed once we fully address this issue in 1.4, so I decided to keep these as separate checks for better clarity and cleanup in the near future.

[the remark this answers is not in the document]
ryfu-msft: Guess I needed to refresh the page :)

SBVels (approved):

PR #2384, opened 2022-07-27
Validate SignatureSha256 for MSIX packages during the manifest validation

AmelBawa-msft:
## What was done?
- Added validation for SignatureSha256 in the manifest validation code.
- Promoted all MSIX manifest validation code to a separate class.
- Updated existing MSIX manifest validation unit test code.

## Test
- Ensured existing AppInstallerCLITests are successful.
- Covered MSIX manifest validation for all combination of signed/unsigned msix/msixbundle packages.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2384)


src/AppInstallerCommonCore/Public/winget/MsixManifestValidation.h
```diff
@@ -0,0 +1,59 @@
...
+        // parse input.
+        std::optional<Msix::OSVersion> GetManifestInstallerMinOSVersion(
+            std::string minOSVersion,
+            std::vector<ValidationError>& errors);
+
+        // Validate Msix package family name.
+        void ValidateMsixManifestPackageFamilyName(
+            Utility::NormalizedString msixPackageFamilyName,
+            std::string manifestPackageFamilyName,
+            std::vector<ValidationError>& errors);
```
AmelBawa-msft, line 38: Should both input here be of the same type: `std::string` or `NormalizedString`?
- First input `msixPackageFamilyName` is the value from the MSIX package.
- Second input `manifestPackageFamilyName` is the value from the YAML manifest.

src/AppInstallerCommonCore/Public/winget/MsixManifestValidation.h
```diff
@@ -0,0 +1,59 @@
...
+#include "winget/Manifest.h"
+#include "winget/ManifestValidation.h"
+
+namespace AppInstaller::Manifest
+{
+    struct MsixManifestValidation
+    {
+        MsixManifestValidation(ValidationError::Level validationErrorLevel) : m_validationErrorLevel(validationErrorLevel) {}
+
+        std::vector<ValidationError> Validate(
```
AmelBawa-msft, line 15: Add comment here

src/AppInstallerCommonCore/Public/winget/MsixManifestValidation.h
JohnMcPMS: `NormalizedString`, which is what the manifest already contains I believe.

src/AppInstallerCommonCore/Manifest/MsixManifestValidation.cpp
```diff
@@ -0,0 +1,159 @@
...
+    void MsixManifestValidation::ValidateMsixManifestSignatureHash(
+        const std::shared_ptr<Msix::MsixInfo> msixInfo,
+        const Utility::SHA256::HashBuffer& manifestSignatureHash,
+        std::vector<ValidationError>& errors)
+    {
+        try
+        {
+            if (!manifestSignatureHash.empty())
+            {
+                auto msixSignature = msixInfo->GetSignature();
```
JohnMcPMS, line 145: I would consider adding a `GetSignatureHash` method since we are now doing it in two places.  It can even be optimized to not actually store the whole stream in memory, not that it is actually large enough to really matter.

JohnMcPMS (approved):

src/AppInstallerCommonCore/Public/winget/MsixManifestValidation.h
AmelBawa-msft: You're right, I misread `string_t` (which is `= Utility::NormalizedString`) as `std::string`

PR #2386, opened 2022-07-28
Enable MSI testing in CI/CD pipeline

JohnMcPMS:
## Change
Build the test MSI locally and include the binary to enable the CI/CD pipeline to test installing it.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2386)


JohnMcPMS:
> * emplates

Not every string is a literal, so seeing `\t` does not always mean a tab...

lechacon (approved):
Would it be possible to add instructions or a script to update the .msi when needed?

JohnMcPMS:
> Would it be possible to add instructions or a script to update the .msi when needed?

It turned out to be "Build for Release/x86".  The cryptic error message you get when you don't choose that is the only thing making it non-obvious.  Where would you suggest putting such instructions?

lechacon:
>  Where would you suggest putting such instructions?

Maybe near to where it is used in the e2etest template, or in a `TestData\README`?

Issue #2302, opened 2022-07-02
Add short aliases for installations for the user scope.


wmstack:
### Description of the new feature / enhancement

Add a short command line flag for installing an app with the scope of `user`. While that is the default scope, it will be implicitly changed if the package requires an admin. Having such a flag makes an early exit if the package in question requires administrator privileges. Otherwise winget runs, and then the user *discovers* that the package requires admin.

If we have the alias for the flag that explicitly forces the package to require user privileges, we guarantee that frustration won't happen later when it is discovered that a package doesn't support admin version. 

### Proposed technical implementation details

Have flags, e.g. `-g` or -u` for limiting the scope of the installation of the package to the current user or globally. Those can be documented along with the `--scope` flag:

```
 --scope, -u, -g                      Select install scope (user or machine)
```

The `-m` flag is already taken, plus the `-g` flag would have the benefit of being analogous to e.g. NPM.

PR #2390, opened 2022-07-28
Add command aliases

### LINKED ISSUE #380: Command aliases

Inspired by #334 in which the author tried to use `find` to search for packages, I think some common command aliases could be added to WinGet so that the user would not get confused when using synonyms of a command.

Provide common aliases such as "find ==> search" and "get ==> install", etc.

```
> winget find git
Name                   Id                    Version                    Matched
-------------------------------------------------------------------------------------
Git                    Git.Git               2.26.2                     Moniker: git
......
```

### LINKED ISSUE #1026: Make "Update" an alias for "Upgrade"
Greetings,

Is it possible to make "update" an alias for the "upgrade" command? 

I keep thinking that the command is: WinGet update SomeApp

I doubt I am the only one making this mistake.

Appreciatively,
-Bruce

### LINKED ISSUE #1299: `winget find` should be a synonym of `winget search`

The more intuitive (and grammatically correct) command for finding packages should be `winget find` rather than `winget search`. Therefore, I suggest that `winget find` should be a synonym. Alternatively, `winget find` should do what `winget search` currently does, and `winget search` could be a more in-depth listing that also shows the versions of every package found.

To explain what I mean by `winget search xxxx` being grammatically incorrect: As an _intransitive_ verb, in English, you must "search *for*" something if you mean "find something". Used _transitively_ ("search something"), it has a different meaning, like "Search that man!" (it doesn't mean "find that man", it means "go through his pockets/clothes and see what he's carrying"). Consider the difference between "you must search this house" and "you must find this house". This is what I mean when I say that the more intuitive verb for finding an app should be `winget find xxxx`. 

* I suggest `winget find` should do what `winget search` currently does. This could be a simple synonym, so both commands do the same thing.
* Alternatively, `winget search` could be more in-depth (closer to the meaning of "search"), something like a combination of current `winget search` and `winget show xxxx --versions`.

### LINKED ISSUE #1978: Add `winget remove` as an alias command for `winget uninstall`

Most popular package managers (`apt`, `dnf`) have `remove` as the argument to uninstall/remove the packages from the system. I feel `winget` should also provide this functionality. This seems very easy to implement, and shouldn't take much time.

Cheers!

When the argument parser reads `remove`, call the method to `uninstall`! As simple as that!

### LINKED ISSUE #2303: Add aliases for installation and uninstallation

Make `add` and `i` aliases for `install` in winget. `add` is used by NPM, by the yarn package manager, and by cargo-edit. Different names improve the experience for newcomers, which, if they come from different package managers, might be used to different names for installing packages, and it would be delightful if they just work.

Also add `r` or `rm` and `remove` as aliases to `uninstall`. `remove` is used by the apt-package manager, so it would be helpful if users of apt can have the same feel with Winget. One letter subcommand aliases might feel wrong, however, NPM has one for installation. Another option is to have `u` as an alias for uninstall, though I don't recommend it.

Could also add `ls` as an alias for `list`. However, it depends on how much the team focuses on UX or DX. I personally strongly suggest those aliases, which need to be strongly tied to historical command names. For example, the apt package manager for linux was perhaps the oldest and most popular, and it has the `remove` keyword for removing a package. So if we reuse its subcommand names, we reduce the friction.

### Notes
One downside is this would give us fewer names for new commands. However, I think that if the package manager makes two similar names for different commands, then it is not a good package manager, solely for that reason. My experience with the rustup package manager in particular strengthens that feeling about similar names, because it has both a target subcommand and a toolchain subcommand and I always mess them up.

Edit: I want to show you the NPM package manager `npm-install` synopsis:
https://docs.npmjs.com/cli/v8/commands/npm-install
```
npm install [<package-spec> ...]

aliases: add, i, in, ins, inst, insta, instal, isnt, isnta, isntal, isntall
```
What do you think?

Make `i`, `add` aliases for `install`, and document them, including in the help page. Make `rm` and `remove` aliases for `uninstall` and document them in the help page as well. Make `ls` an alias for `list`.

Trenly:
  * Resolves #1299
  * Resolves #2303 
  * Resolves #1978
  * Resolves #1026
  * Resolves #380

I have tested that this is functional, and that alias collisions do not occur when commands are at different levels. E.g - `winget add <params>` does not conflict with `winget source add <params>`.

I need some guidance on how to implement the unit tests for this - mainly just how to modify the `EnsureCommandConsistency` test. I presume that the command consistency will need to check against command full names, and that command aliases will have to have a method added to get the alias full name, but I'd like some help and guidance before making those changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2390)

### DIFF
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -44,6 +44,21 @@ namespace AppInstaller::CLI
         {
             m_fullName = name;
         }
+
+        if (!aliases.empty())
+        {
+            for (auto& alias : aliases)
+            {
+                std::string _aliasFullName;
+                if (!parent.empty())
+                {
+                    _aliasFullName = parent;
+                    _aliasFullName += ParentSplitChar;
+                }
+                _aliasFullName += alias;
+                m_aliasFullNames.emplace_back(_aliasFullName);
+            }
+        }
     }
 
     void Command::OutputIntroHeader(Execution::Reporter& reporter) const
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -26,11 +26,12 @@ namespace AppInstaller::CLI
 
     Command::Command(
         std::string_view name,
+        std::vector<std::string_view> aliases,
         std::string_view parent,
         Command::Visibility visibility,
         Settings::ExperimentalFeature::Feature feature,
         Settings::TogglePolicy::Policy groupPolicy) :
-        m_name(name), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
+        m_name(name), m_aliases(aliases), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
     {
         if (!parent.empty())
         {
@@ -291,7 +292,10 @@ namespace AppInstaller::CLI
 
         for (auto& command : commands)
         {
-            if (Utility::CaseInsensitiveEquals(*itr, command->Name()))
+            if (
+                Utility::CaseInsensitiveEquals(*itr, command->Name()) ||
+                Utility::CaseInsensitiveContains(command->Aliases(), *itr)
+            )
             {
                 if (!ExperimentalFeature::IsEnabled(command->Feature()))
                 {
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -760,6 +760,17 @@ namespace AppInstaller::CLI
                 {
                     context.Reporter.Completion() << command->Name() << std::endl;
                 }
+                // Allow for command aliases to be auto-completed
+                if (!(command->Aliases()).empty())
+                {
+                    for (const auto& commandAlias : command->Aliases())
+                    {
+                        if (Utility::CaseInsensitiveStartsWith(commandAlias, word))
+                        {
+                            context.Reporter.Completion() << commandAlias << std::endl;
+                        }
+                    }
+                }
             }
         }
 
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -131,6 +131,7 @@ namespace AppInstaller::CLI
         // Output the command preamble and command chain
         infoOut << Resource::String::Usage << ": winget"_liv << Utility::LocIndView{ commandChain };
 
+        auto commandAliases = Aliases();
         auto commands = GetVisibleCommands();
         auto arguments = GetVisibleArguments();
 
@@ -198,6 +199,30 @@ namespace AppInstaller::CLI
         infoOut <<
             std::endl <<
             std::endl;
+
+        if (!commandAliases.empty())
+        {
+            infoOut << Resource::String::AvailableCommandAliases << std::endl << "  "_liv;
+            
+            size_t i = 1;
+            for (const auto& commandAlias : commandAliases)
+            {
+                infoOut << Execution::HelpCommandEmphasis << commandAlias;
+                //infoOut << "  "_liv << Execution::HelpCommandEmphasis << commandAlias;
+                if (i%7 != 0 && i!= commandAliases.size()) {
+                    infoOut << ", "_liv;
+                }
+                else if (i != commandAliases.size()) {
+                    infoOut << std::endl << "  "_liv;
+                }
+                else
+                {
+                    infoOut << std::endl;
+                }
+                i++;
+            }
+            infoOut << std::endl;
+        }
 
         if (!commands.empty())
         {
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -208,15 +208,22 @@ namespace AppInstaller::CLI
             for (const auto& commandAlias : commandAliases)
             {
                 infoOut << Execution::HelpCommandEmphasis << commandAlias;
-                //infoOut << "  "_liv << Execution::HelpCommandEmphasis << commandAlias;
+                /*
+                *  Print up to 7 command aliases per line, comma separated for readability
+                *  This is intended to be future-looking so that in case common typos are
+                *  added to command aliases as scoop/chocolatey do, the list does not become
+                *  extremely long to the point where it makes the help page less usable.
+                */
                 if (i%7 != 0 && i!= commandAliases.size()) {
                     infoOut << ", "_liv;
                 }
                 else if (i != commandAliases.size()) {
+                    // Command aliases which are on new lines should be indented
                     infoOut << std::endl << "  "_liv;
                 }
                 else
                 {
+                    // Always add a newline after the last command alias
                     infoOut << std::endl;
                 }
                 i++;
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -49,14 +49,14 @@ namespace AppInstaller::CLI
         {
             for (auto& alias : aliases)
             {
-                std::string _aliasFullName;
+                std::string aliasFullName;
                 if (!parent.empty())
                 {
-                    _aliasFullName = parent;
-                    _aliasFullName += ParentSplitChar;
+                    aliasFullName = parent;
+                    aliasFullName += ParentSplitChar;
                 }
-                _aliasFullName += alias;
-                m_aliasFullNames.emplace_back(_aliasFullName);
+                aliasFullName += alias;
+                m_aliasFullNames.emplace_back(std::move(aliasFullName));
             }
         }
     }
@@ -214,10 +214,12 @@ namespace AppInstaller::CLI
                 *  added to command aliases as scoop/chocolatey do, the list does not become
                 *  extremely long to the point where it makes the help page less usable.
                 */
-                if (i%7 != 0 && i!= commandAliases.size()) {
+                if (i%7 != 0 && i!= commandAliases.size())
+                {
                     infoOut << ", "_liv;
                 }
-                else if (i != commandAliases.size()) {
+                else if (i != commandAliases.size())
+                {
                     // Command aliases which are on new lines should be indented
                     infoOut << std::endl << "  "_liv;
                 }
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -31,7 +31,7 @@ namespace AppInstaller::CLI
         Command::Visibility visibility,
         Settings::ExperimentalFeature::Feature feature,
         Settings::TogglePolicy::Policy groupPolicy) :
-        m_name(name), m_aliases(aliases), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
+        m_name(name), m_aliases(std::move(aliases)), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
     {
         if (!parent.empty())
         {
@@ -43,21 +43,6 @@ namespace AppInstaller::CLI
         else
         {
             m_fullName = name;
-        }
-
-        if (!aliases.empty())
-        {
-            for (auto& alias : aliases)
-            {
-                std::string aliasFullName;
-                if (!parent.empty())
-                {
-                    aliasFullName = parent;
-                    aliasFullName += ParentSplitChar;
-                }
-                aliasFullName += alias;
-                m_aliasFullNames.emplace_back(std::move(aliasFullName));
-            }
         }
     }
 
@@ -202,33 +187,11 @@ namespace AppInstaller::CLI
 
         if (!commandAliases.empty())
         {
-            infoOut << Resource::String::AvailableCommandAliases << std::endl << "  "_liv;
+            infoOut << Resource::String::AvailableCommandAliases << std::endl;
             
-            size_t i = 1;
             for (const auto& commandAlias : commandAliases)
             {
-                infoOut << Execution::HelpCommandEmphasis << commandAlias;
-                /*
-                *  Print up to 7 command aliases per line, comma separated for readability
-                *  This is intended to be future-looking so that in case common typos are
-                *  added to command aliases as scoop/chocolatey do, the list does not become
-                *  extremely long to the point where it makes the help page less usable.
-                */
-                if (i%7 != 0 && i!= commandAliases.size())
-                {
-                    infoOut << ", "_liv;
-                }
-                else if (i != commandAliases.size())
-                {
-                    // Command aliases which are on new lines should be indented
-                    infoOut << std::endl << "  "_liv;
-                }
-                else
-                {
-                    // Always add a newline after the last command alias
-                    infoOut << std::endl;
-                }
-                i++;
+                infoOut << "  "_liv << Execution::HelpCommandEmphasis << commandAlias << std::endl;
             }
             infoOut << std::endl;
         }
@@ -763,7 +726,7 @@ namespace AppInstaller::CLI
                     context.Reporter.Completion() << command->Name() << std::endl;
                 }
                 // Allow for command aliases to be auto-completed
-                if (!(command->Aliases()).empty())
+                if (!(command->Aliases()).empty() && !word.empty())
                 {
                     for (const auto& commandAlias : command->Aliases())
                     {
--- a/src/AppInstallerCLICore/Command.h
+++ b/src/AppInstallerCLICore/Command.h
@@ -82,6 +82,7 @@ namespace AppInstaller::CLI
         std::string_view Name() const { return m_name; }
         std::vector<std::string_view> Aliases() const { return m_aliases; }
         const std::string& FullName() const { return m_fullName; }
+        const std::vector<std::string> AliasFullNames() const { return m_aliasFullNames; }
         Command::Visibility GetVisibility() const;
         Settings::ExperimentalFeature::Feature Feature() const { return m_feature; }
         Settings::TogglePolicy::Policy GroupPolicy() const { return m_groupPolicy; }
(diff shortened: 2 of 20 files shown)


Trenly:
@ryfu-msft / @JohnMcPMS - Do you have any recommendations on how I should modify the `EnsureCommandConsistency` test?

JohnMcPMS:
@denelon for this potential change to the user interface.  Just want to make sure that he is aware and agreeable to the aliases being added.

I assume that many of these are less about the amount of typing and more about offering more verbs for a more intuitive experience.

There has been a lack of rigor (mostly from me) about completion, with many new features not being added.  I'm not going to hold you to it, but now that I've started thinking about it I see it more.

I would think that the tests should be able to steal from the way that the aliases for arguments are verified:
https://github.com/microsoft/winget-cli/blob/888b4ed8f4f7d25cb05a47210e083fe29348163b/src/AppInstallerCLITests/Command.cpp#L91-L93

Trenly:
> I would think that the tests should be able to steal from the way that the aliases for arguments are verified

I was thinking this, but I wasn’t sure quite how to implement it given that arg aliases are all strings, and I chose to implement this as a vector

> Let me be more explicit, by adding a way to do a similar thing but op returns a vector<string>, or by changing all of them to use vector<string> and just having the current ones wrap the single string in a vector.

I can fairly easily add a method to get all of the aliasFullNames as a vector, and create a method for `EnsureVectorStringsAreLowercaseAndNoCollisions` . Thank you for the advice!

denelon:
I've seen plenty of feedback in this area with the list of Issues. I think it will help in terms of muscle memory for users coming from other package managers.

Can we get a table of the original command and the aliases in the description? That would help with consistency when release notes come out.

How are aliases to be discovered?

Maybe: 
```text
winget search -?                                                                in pwsh at 16:56:34
Windows Package Manager v1.3.2091

Searches for packages from configured sources.

usage: winget search [[-q] <query>] [<options>]

The following command aliases are available:
  find

  --tag                       Filter results by tag
  --command                   Filter results by command
  -n,--count                  Show no more than specified number of results (between 1 and 1000)

More help can be found at: https://aka.ms/winget-command-search
```

I expect these would only impact the CLI and not the PowerShell cmdlets as those have a discrete set of approved nouns and verbs.

JohnMcPMS:
Agreed, no change to PowerShell.

Trenly:
Agreed with all points

src/AppInstallerCLICore/Command.cpp
```diff
@@ -183,6 +200,37 @@ namespace AppInstaller::CLI
...
+            size_t i = 1;
+            for (const auto& commandAlias : commandAliases)
+            {
+                infoOut << Execution::HelpCommandEmphasis << commandAlias;
+                /*
+                *  Print up to 7 command aliases per line, comma separated for readability
+                *  This is intended to be future-looking so that in case common typos are
+                *  added to command aliases as scoop/chocolatey do, the list does not become
+                *  extremely long to the point where it makes the help page less usable.
+                */
```
Trenly, line 216: For your reference @denelon -
image

Trenly:
> There has been a lack of rigor (mostly from me) about completion, with many new features not being added. I'm not going to hold you to it, but now that I've started thinking about it I see it more. - @JohnMcPMS 

I added command aliases to completion

> How are aliases to be discovered? - @denelon 

Added to the command help per the proposal

> Can we get a table of the original command and the aliases in the description? That would help with consistency when release notes come out. - @denelon 

| Original Command | New Aliases |
| ------------- | ------------- |
| search  | find  |
| install  | add |
| upgrade | update |
| uninstall | remove, rm |
| source update | source refresh |
| source remove | source rm |

lechacon:
Pie in the sky idea: What if instead of adding the aliases, we add a feature where users can create their own aliases, like git does?

denelon:
I could see something like that as a part of the settings.json file. It's an interesting idea. I'd expect we would need to be fairly strict when defining those aliases, but I like it.

Trenly:
> Pie in the sky idea: What if instead of adding the aliases, we add a feature where users can create their own aliases, like git does?

I think that having that option as a user setting would be good; I think that a combination here would be best, where default aliases are specified as they are here, but if a user specified one that conflicted with a pre-defined alias, the user-defined alias would take precedence. But, I think overall it's an interesting idea and could be a good feature in addition to the standard aliases

vedantmgoyal2009:
|Command|Alias|
|-|-|
|install|i|
|list|ls|
|show|view|
|settings|config|

src/AppInstallerCLICore/Command.cpp
```diff
@@ -291,7 +339,10 @@ namespace AppInstaller::CLI
 
         for (auto& command : commands)
         {
-            if (Utility::CaseInsensitiveEquals(*itr, command->Name()))
+            if (
+                Utility::CaseInsensitiveEquals(*itr, command->Name()) ||
+                Utility::CaseInsensitiveContains(command->Aliases(), *itr)
```
lechacon, line 344: I think at some point we send telemetry about what command was executed with its full name. I think it could be helpful to report the specific alias. If most people use the alias than the original command, maybe we could rename the command in a v2 to what people prefer.

src/AppInstallerCLICore/Command.cpp
```diff
@@ -43,6 +44,21 @@ namespace AppInstaller::CLI
         {
             m_fullName = name;
         }
+
+        if (!aliases.empty())
+        {
+            for (auto& alias : aliases)
+            {
+                std::string _aliasFullName;
```
lechacon, line 52: Nit: I think the leading underscore is not consistent with the style in the rest of the project

src/AppInstallerCLICore/Command.cpp
```diff
@@ -43,6 +44,21 @@ namespace AppInstaller::CLI
...
         }
+
+        if (!aliases.empty())
+        {
+            for (auto& alias : aliases)
+            {
+                std::string _aliasFullName;
+                if (!parent.empty())
+                {
+                    _aliasFullName = parent;
```
lechacon, line 55: What if the parent is also aliased?

src/AppInstallerCLICore/Command.cpp
```diff
@@ -43,6 +44,21 @@ namespace AppInstaller::CLI
...
+                    _aliasFullName += ParentSplitChar;
+                }
+                _aliasFullName += alias;
+                m_aliasFullNames.emplace_back(_aliasFullName);
```
lechacon, line 59: `std::move` here

src/AppInstallerCLICore/Command.cpp
Trenly: Assuming you're referring to this - https://github.com/microsoft/winget-cli/blob/a56372ba05d8f4423cb7cf5be0b5c0bacb445ca3/src/AppInstallerCLICore/Command.cpp#L888

Just clarifying - All that would need to be done to enable telemetry around specific aliases would be to pass the aliasFullName to the telemetry logger instead of the command full name?

src/AppInstallerCLICore/Command.cpp
```diff
@@ -26,11 +26,12 @@ namespace AppInstaller::CLI
 
     Command::Command(
         std::string_view name,
+        std::vector<std::string_view> aliases,
```
lechacon, line 29: Maybe this should be `const &`?
Edit: Or `&&`

src/AppInstallerCLICore/Command.cpp
```diff
@@ -183,6 +200,37 @@ namespace AppInstaller::CLI
...
+            for (const auto& commandAlias : commandAliases)
+            {
+                infoOut << Execution::HelpCommandEmphasis << commandAlias;
+                /*
+                *  Print up to 7 command aliases per line, comma separated for readability
+                *  This is intended to be future-looking so that in case common typos are
+                *  added to command aliases as scoop/chocolatey do, the list does not become
+                *  extremely long to the point where it makes the help page less usable.
+                */
+                if (i%7 != 0 && i!= commandAliases.size()) {
```
lechacon, line 217: Nit: Braces in new line for consistency

src/AppInstallerCLICore/Command.h
```diff
@@ -78,7 +80,9 @@ namespace AppInstaller::CLI
         constexpr static char ParentSplitChar = ':';
 
         std::string_view Name() const { return m_name; }
+        std::vector<std::string_view> Aliases() const { return m_aliases; }
```
lechacon, line 83: Return a `const &` to avoid creating a copy.

src/AppInstallerCLICore/Commands/SearchCommand.h
```diff
@@ -5,9 +5,10 @@
 
 namespace AppInstaller::CLI
 {
+
     struct SearchCommand final : public Command
     {
-        SearchCommand(std::string_view parent) : Command("search", parent) {}
+        SearchCommand(std::string_view parent) : Command("search", {"find"}, parent) {}
```
lechacon, line 11: Nit: Add spaces around `"find"` for consistency

src/AppInstallerCLICore/Command.h
```diff
@@ -56,16 +56,18 @@ namespace AppInstaller::CLI
...
+            Command(name, aliases, parent, visibility, Settings::ExperimentalFeature::Feature::None) {}
+        Command(std::string_view name,std::vector<std::string_view> aliases, std::string_view parent, Settings::ExperimentalFeature::Feature feature) :
+            Command(name, aliases, parent, Command::Visibility::Show, feature) {}
+        Command(std::string_view name,std::vector<std::string_view> aliases, std::string_view parent, Settings::TogglePolicy::Policy groupPolicy) :
+            Command(name, aliases, parent, Command::Visibility::Show, Settings::ExperimentalFeature::Feature::None, groupPolicy) {}
+        Command(std::string_view name,std::vector<std::string_view> aliases, std::string_view parent, Command::Visibility visibility, Settings::ExperimentalFeature::Feature feature) :
+            Command(name, aliases, parent, visibility, feature, Settings::TogglePolicy::Policy::None) {}
+        Command(std::string_view name,std::vector<std::string_view> aliases, std::string_view parent, Command::Visibility visibility, Settings::ExperimentalFeature::Feature feature, Settings::TogglePolicy::Policy groupPolicy);
```
lechacon, line 70: Why not keep the existing constructors so that we don't need to pass the empty list for aliases when there aren't any?

src/AppInstallerCLITests/Command.cpp
```diff
@@ -47,6 +57,11 @@ std::string GetArgumentAlias(const Argument& arg)
     }
 }
 
+bool StringIsLowercase(std::string& s)
+{
+    return { Utility::ToLower(s) == s };
```
lechacon, line 62: Nit: Are the braces needed?

src/AppInstallerCLITests/Command.cpp
```diff
@@ -47,6 +57,11 @@ std::string GetArgumentAlias(const Argument& arg)
     }
 }
 
+bool StringIsLowercase(std::string& s)
```
lechacon, line 60: `const std::string&`?

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -105,6 +105,18 @@ namespace AppInstaller::Utility
...
 
+    bool CaseInsensitiveContains(std::vector<std::string_view> a, std::string_view b)
+    {
+        for (std::string_view c : a)
+        {
+            if (CaseInsensitiveEquals(c, b))
+            {
+                return true;
+            }
+        }
```
lechacon, line 116: Not saying that it would be better, but FYI you can do this with `std::any_of`. Something like
```c++
return std::any_of(a.begin(), a.end(), [=](const std::string_view s) { return CaseInsensitiveEquals(s, b); });
```

src/AppInstallerCLICore/Command.cpp
lechacon: I think it would be more complicated. We'd probably need a new event, like `Logging::Telemetry().LogCommandAlias(...)` to avoid tripping any data processing we currently have that looks at the full name. Also, we'd need to have `CaseInsensitiveContains` tell us which alias matched. Now that I'm saying this, it seems too complicated and I'm not sure if it is needed..

src/AppInstallerCLICore/Command.h
Trenly: I chose not to keep the existing constructors to force consideration of aliases and to reduce the amount of overloads. If the old constructors are still desired, I can add them back, but I thought it made more sense to just add aliases, or the empty vector, to everything.

src/AppInstallerCLICore/Command.cpp
Trenly: I'm not sure I quite follow what you mean

src/AppInstallerCLICore/Command.cpp
Trenly: Are you suggesting that the entire tree needs to be expanded to find all potential alias full names?

src/AppInstallerCLICore/Command.cpp
lechacon: I think the type should be `const std::vector<std::string_view>&` or `std::vector<std::string_view>&&` to avoid creating a copy of the vector when the constructor is called.

src/AppInstallerCLICore/Command.cpp
lechacon: I'm suggesting that we need to give some thought to that case. Depending on how we use the full name, we could need to generate all possible full names, or just one, or maybe we don't need the alias full names at all.

I *think* we only use the full name for logging? If that's the case we probably don't need to compute the alias full names, but we'd need to verify.

Trenly: I did have to use the full name to ensure that there weren't collisions between aliases and commands, given that aliases can be the same but not collide if they have different parents. Since the collision check is recursive, I don't think the full expansion is needed. By checking that `root/search` and `root/find` don't collide with any other commands, then any sub-commands underneath them will also not collide with commands in other branches. So, as long as there are no aliases that collide at the same level, checking only one branch of the sub-trees suffices as a check for all other branches within that path.

wmstack:
I suggest that https://github.com/microsoft/winget-cli/issues/2302 is considered as well. I feel like user scope and machine scope could also be aliases when installing a package

wmstack:
If you want to find aliases, I think NPM does a great job with their aliases. They make it so that if you make a typo like `instal` or `istnall` it would still work. These could be *hidden* aliases:
```
npm install [<package-spec> ...]

aliases: add, i, in, ins, inst, insta, instal, isnt, isnta, isntal, isntall
```
> https://docs.npmjs.com/cli/v8/commands/npm-install


That might be considered butchering the aliases, lol. Otherwise we could have this mechanism that says "did you mean ...", however, I think most of the time the author really means it.

Another type of alias that I want to mention is adding `s` to signify listing:
```
# equivalent to `winget list`
winget packages
```

Trenly:
> I suggest that #2302 is considered as well. I feel like user scope and machine scope could also be aliases when installing a package (with `-g` like NPM and `-u` to force a user scope package).

Those are flags, and not commands. Therefore, they fall outside the scope of a command alias and cannot be considered as part of this PR

src/AppInstallerCLICore/Command.cpp
```diff
@@ -26,11 +26,12 @@ namespace AppInstaller::CLI
 
     Command::Command(
         std::string_view name,
+        std::vector<std::string_view> aliases,
         std::string_view parent,
         Command::Visibility visibility,
         Settings::ExperimentalFeature::Feature feature,
         Settings::TogglePolicy::Policy groupPolicy) :
-        m_name(name), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
+        m_name(name), m_aliases(aliases), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
```
JohnMcPMS, line 34:
```suggestion
        m_name(name), m_aliases(std::move(aliases)), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
```
This solves @lechacon 's comment above.

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -105,6 +105,11 @@ namespace AppInstaller::Utility
         return ToLower(a) == ToLower(b);
     }
 
+    bool CaseInsensitiveContains(std::vector<std::string_view> a, std::string_view b)
+    {
+        return std::any_of(a.begin(), a.end(), [=](const std::string_view s) { return CaseInsensitiveEquals(s, b); });
```
JohnMcPMS, line 110:
```suggestion
        auto B = ToLower(b);
        return std::any_of(a.begin(), a.end(), [&](const std::string_view& s) { return ToLower(s) == B; });
```

src/AppInstallerCommonCore/AppInstallerStrings.cpp
```diff
@@ -105,6 +105,11 @@ namespace AppInstaller::Utility
         return ToLower(a) == ToLower(b);
     }
 
+    bool CaseInsensitiveContains(std::vector<std::string_view> a, std::string_view b)
```
JohnMcPMS, line 108:
```suggestion
    bool CaseInsensitiveContains(const std::vector<std::string_view>& a, std::string_view b)
```

JohnMcPMS (changes requested):

src/AppInstallerCLITests/Command.cpp
```diff
@@ -80,9 +94,41 @@ void EnsureStringsAreLowercaseAndNoCollisions(const std::string& info, const Enu
...
+            values.emplace(std::move(valString));
+        }
+    }
+}
+
 void EnsureCommandConsistency(const Command& command)
 {
-    EnsureStringsAreLowercaseAndNoCollisions(command.FullName() + " commands", command.GetCommands(), GetCommandName);
+    // Command names and aliases exist in the same space, so both need to be checked as a set
+    // However, collisions do not occur between levels, so the full name must be used to check for collision
```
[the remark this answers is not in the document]
Trenly: Ahh, I see. I was misunderstanding

Trenly:
I think I missed a type conversion in the test; I don't think std::string_view can be implicitly cast to std::string; If the build fails, I'll push a fix

JohnMcPMS (approved):

PR #2401, opened 2022-08-01
Add support for installing portables without developer mode and running as admin

### LINKED ISSUE #2368: Support portable packages in user mode without Developer Mode enabled

During testing for installing portable packages in user mode with Developer Mode disabled, we identified an error:

* #2364 

Users will be prompted for administrator mode to install portable packages if Developer Mode is disabled.

Reference: https://docs.microsoft.com/windows/win32/api/winbase/nf-winbase-createsymboliclinkw

A solution for installing portable packages in user mode without requiring Developer Mode to be enabled.

ryfu-msft:
Resolves #2368 

Changes: 
- Handles the scenario for installing a portable where a user does not have developer mode enabled or is not running as admin by skipping the symlink creation step and adding the exe directly. 

Tests:
- Adds E2E tests to verify that this scenario is supported as expected.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2401)


src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -435,14 +367,74 @@ namespace AppInstaller::CLI::Workflow
                 context.Reporter.Warn() << Resource::String::OverwritingExistingFileAtMessage << ' ' << symlinkFullPath.u8string() << std::endl;
             }
 
-            std::filesystem::create_symlink(targetFullPath, symlinkFullPath);
-            AICLI_LOG(CLI, Info, << "Symlink created at: " << symlinkFullPath);
-            uninstallEntry.SetValue(PortableValueName::PortableSymlinkFullPath, symlinkFullPath.wstring());
+            if (Filesystem::CreateSymlink(targetFullPath, symlinkFullPath))
+            {
+                AICLI_LOG(CLI, Info, << "Symlink created at: " << symlinkFullPath);
+                uninstallEntry.SetValue(PortableValueName::PortableSymlinkFullPath, symlinkFullPath.wstring());
```
JohnMcPMS, line 373: The general flow for these should be:
1. Write artifact stating that you added a file
2. Attempt to add the file
3. <Time passes, user uninstalls...>
4. Attempt to delete the file due to presence of artifact (if it points to something valid)
5. Remove artifact

If you wait to write the artifact until after you add the file, you could abandon it due to failing to write the artifact or a crash or a system power loss.  If you write the artifact and it points to nothing due to one of the error states, then step 4 is a check only and nothing is left behind.

I agree that it makes things more complicated, but the goal is to create a robust system, and that robustness is less efficient in both code and runtime.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -435,14 +367,74 @@ namespace AppInstaller::CLI::Workflow
...
+            }
+        }
 
-            Manifest::ScopeEnum scope = ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope));
-            if (AddToPathRegistry(scope))
+        void AddToPathVariable(Execution::Context& context)
+        {
+            PortableARPEntry& uninstallEntry = context.Get<Execution::Data::PortableARPEntry>();
+            Manifest::ScopeEnum scope = uninstallEntry.GetScope();
+            const auto& installDirectoryAddedToPath = uninstallEntry[PortableValueName::InstallDirectoryAddedToPath];
```
JohnMcPMS, line 389: You shouldn't be passing information from one task to another through the registry.

Suppose:
1. Install 1.0 and it can't create a symlink, so it adds to path
2. Upgrade to 2.0 as admin so it can create a symlink
3. This code sees that it should add to path again (although I think you do check for the path being there already)
4. Oops, now both a symlink and a path?

This is one reason not to pass the information using the registry, but it also brings up an interesting question of what to do on upgrades around this behavior.  I would think my case above should just preserve the path and not create a symlink.  I would think that the inverse (admin then user) should attempt to preserve the symlink, as if it doesn't need to be changed (same link and target paths) then there is no need.

Basically, we should determine if any changes are actually needed, then only attempt to make them if necessary.

My intuition says that this would be easier to understand if everything was collected in an object structure rather than having every task handling many things repeatedly.

JohnMcPMS (approved):

ToothyDev:
Apparently, this misbehaviour still exists currently and it will refuse to install portables without admin/developer mode

Trenly:
> Apparently, this misbehaviour still exists currently and it will refuse to install portables without admin/developer mode

Which version are you using? This fix first appears in v1.4.3132-preview

ToothyDev:
Ah, I'm on 1.3-something. Thanks for clearing it up!

PR #2410, opened 2022-08-03
Return error codes for dependencies failure

hackean-msft:
The goal of these change is to allow the WinGetValidateManifestV3 to return dependencies failure as error code for programmatic usage. This prevents the need to parse strings to decode the different failure types associated with dependences.

Changes:
- Use localized messages for validation errors.
- Parse exceptions to decode failure types and surface that to the caller. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2410)


src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
 
 namespace AppInstaller::Manifest
 {
+    namespace
+    {
+        std::map<AppInstaller::StringResource::StringId, std::string_view> ErrorIdToMessageMap = {
+            { AppInstaller::Manifest::ManifestError::InvalidRootNode, "Encountered unexpected root node."},
```
lechacon, line 16: Maybe mark all those literals as `sv` so they are `string_view`? (TBH, I'm not sure what difference it makes, I just know we do that in other places.)

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -398,4 +447,9 @@ namespace AppInstaller::Manifest
 
         return errors;
     }
+
+    std::string_view ValidationError::GetErrorMessage() const
+    {
+        return ErrorIdToMessageMap[Message];
```
lechacon, line 453: Maybe if the StringID is not in the map it would be better to return the ID instead of the empty string.

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
```diff
@@ -24,7 +24,7 @@ namespace AppInstaller::Manifest::YamlParser
             if (!entry.Root["PackageIdentifier"])
             {
                 errors.emplace_back(ValidationError::MessageFieldWithFile(
-                    ManifestError::RequiredFieldMissing, "PackageIdentifier", entry.FileName));
+                    (ManifestError::RequiredFieldMissing), "PackageIdentifier", entry.FileName));
```
lechacon, line 27: Is this change needed?

src/AppInstallerRepositoryCore/ArpVersionValidation.cpp
```diff
@@ -67,11 +67,10 @@ namespace AppInstaller::Repository
             {
                 if (manifestArpVersionRange.Overlaps(arpInIndex))
                 {
-                    std::string errorMsg = Manifest::ManifestError::ArpVersionOverlapWithIndex;
-                    errorMsg.append("[" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
-                    AICLI_LOG(Repo, Error, << errorMsg);
+                    std::string errorMsg = ("[" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
```
lechacon, line 70: Rename `errorMsg` to something else as this is not the full message anymore.

src/AppInstallerRepositoryCore/ArpVersionValidation.cpp
```diff
@@ -67,11 +67,10 @@ namespace AppInstaller::Repository
             {
                 if (manifestArpVersionRange.Overlaps(arpInIndex))
                 {
-                    std::string errorMsg = Manifest::ManifestError::ArpVersionOverlapWithIndex;
-                    errorMsg.append("[" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
-                    AICLI_LOG(Repo, Error, << errorMsg);
+                    std::string errorMsg = ("[" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
+                    AICLI_LOG(Repo, Error, << Utility::ConvertToUTF8(Manifest::ManifestError::ArpVersionOverlapWithIndex) << errorMsg);
```
lechacon, line 71: Add a space in the output between `ArpVersionOverlapWithIndex` and `errorMsg`.

lechacon, line 71: Is there a way to output the actual string instead of the ID?

src/AppInstallerRepositoryCore/ArpVersionValidation.cpp
```diff
@@ -67,11 +67,10 @@ namespace AppInstaller::Repository
...
                 if (manifestArpVersionRange.Overlaps(arpInIndex))
                 {
-                    std::string errorMsg = Manifest::ManifestError::ArpVersionOverlapWithIndex;
-                    errorMsg.append("[" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
-                    AICLI_LOG(Repo, Error, << errorMsg);
+                    std::string errorMsg = ("[" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
+                    AICLI_LOG(Repo, Error, << Utility::ConvertToUTF8(Manifest::ManifestError::ArpVersionOverlapWithIndex) << errorMsg);
                     THROW_EXCEPTION(Manifest::ManifestException(
-                        { Manifest::ValidationError(errorMsg) },
+                        { Manifest::ValidationError(Manifest::ManifestError::ArpVersionOverlapWithIndex, errorMsg) },
```
lechacon, line 73: This is calling the constructor `ValidationError(AppInstaller::StringResource::StringId message, std::string field)`. That doesn't seem right to me as this is not a manifest field (unless I'm misunderstanding what `field` means here).

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.h
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "pch.h"
```
lechacon, line 4: Don't include `pch.h` in a header.

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.h
```diff
@@ -0,0 +1,23 @@
...
+    enum WinGetManifestDependenciesErrorResult
+    {
+        None = 0x0,
+
+        // Each validation step should have an enum for corresponding failure.
+        SingleManifestPackageHasDependencies = 0x10000,
+        MultiManifestPackageHasDependencies = 0x20000,
+        MissingManifestDependenciesNode = 0x40000,
+        NoSuitableMinVersion = 0x80000,
+        FoundLoop = 0x1000000,
```
lechacon, line 19: Are these errors specific to dependencies? If so, could you make the names more descriptive?

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,34 @@
...
+                {AppInstaller::Manifest::ManifestError::MultiManifestPackageHasDependencies, WinGetManifestDependenciesErrorResult::MultiManifestPackageHasDependencies },
+                {AppInstaller::Manifest::ManifestError::MissingManifestDependenciesNode, WinGetManifestDependenciesErrorResult::MissingManifestDependenciesNode },
+                {AppInstaller::Manifest::ManifestError::NoSuitableMinVersion, WinGetManifestDependenciesErrorResult::NoSuitableMinVersion },
+                {AppInstaller::Manifest::ManifestError::FoundLoop, WinGetManifestDependenciesErrorResult::FoundLoop}
+            };
+
+            if (dependenciesErrorMessageMap.find(message) != dependenciesErrorMessageMap.end())
+            {
+                return dependenciesErrorMessageMap[message];
+            }
```
lechacon, line 29: This is making the same lookup in the map twice. You could do something like
```C++
auto itr = dependenciesErrorMessageMap.find(message);
if (itr != dependenciesErrorMessageMap.end())
{
    return itr->second;
}
```

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,34 @@
...
+    WinGetManifestDependenciesErrorResult GetValidationResultFromException(const AppInstaller::Manifest::ManifestException& manifestException)
+	{
+        auto validationErrors = manifestException.Errors();
+        for (auto validationError : validationErrors)
+        {
+            auto message = validationError.Message;
+
+            std::map<AppInstaller::StringResource::StringId, WinGetManifestDependenciesErrorResult> dependenciesErrorMessageMap =
+            {
+                {AppInstaller::Manifest::ManifestError::SingleManifestPackageHasDependencies, WinGetManifestDependenciesErrorResult::SingleManifestPackageHasDependencies},
```
lechacon, line 19: Why not add a `using namespace` to make these shorter?

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
...
+            { AppInstaller::Manifest::ManifestError::InconsistentMultiFileManifestDefaultLocale, "DefaultLocale value in version manifest does not match PackageLocale value in defaultLocale manifest." },
+            { AppInstaller::Manifest::ManifestError::FieldFailedToProcess, "Failed to process field." },
+            { AppInstaller::Manifest::ManifestError::InvalidBcp47Value, "The locale value is not a well formed bcp47 language tag." },
+            { AppInstaller::Manifest::ManifestError::BothAllowedAndExcludedMarketsDefined, "Both AllowedMarkets and ExcludedMarkets defined." },
+            { AppInstaller::Manifest::ManifestError::DuplicateReturnCodeEntry, "Duplicate installer return code found." },
+            { AppInstaller::Manifest::ManifestError::FieldRequireVerifiedPublisher, "Field usage requires verified publishers." },
+            { AppInstaller::Manifest::ManifestError::SingleManifestPackageHasDependencies, "Package has a single manifest and is a dependency of other manifests." },
+            { AppInstaller::Manifest::ManifestError::MultiManifestPackageHasDependencies, "Deleting the manifest will be break the following dependencies." },
```
Trenly, line 42:
```suggestion
            { AppInstaller::Manifest::ManifestError::MultiManifestPackageHasDependencies, "Deleting the manifest will break the following dependencies." },

```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
...
+    {
+        std::map<AppInstaller::StringResource::StringId, std::string_view> ErrorIdToMessageMap = {
+            { AppInstaller::Manifest::ManifestError::InvalidRootNode, "Encountered unexpected root node."},
+            { AppInstaller::Manifest::ManifestError::FieldUnknown, "Unknown field."},
+            { AppInstaller::Manifest::ManifestError::FieldIsNotPascalCase, "All field names should be PascalCased."},
+            { AppInstaller::Manifest::ManifestError::FieldDuplicate, "Duplicate field found in the manifest." },
+            { AppInstaller::Manifest::ManifestError::RequiredFieldEmpty, "Required field with empty value." },
+            { AppInstaller::Manifest::ManifestError::RequiredFieldMissing,  "Required field missing." },
+            { AppInstaller::Manifest::ManifestError::InvalidFieldValue, "Invalid field value." },
+            { AppInstaller::Manifest::ManifestError::ExeInstallerMissingSilentSwitches, "Silent and SilentWithProgress switches are not specified for InstallerType exe.Please make sure the installer can run unattended." },
```
Trenly, line 23:
```suggestion
            { AppInstaller::Manifest::ManifestError::ExeInstallerMissingSilentSwitches, "Silent and SilentWithProgress switches are not specified for InstallerType 'exe'. Please make sure the installer can run unattended." },
```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
...
+            { AppInstaller::Manifest::ManifestError::FieldNotSupported, "Field is not supported." },
+            { AppInstaller::Manifest::ManifestError::FieldValueNotSupported, "Field value is not supported." },
+            { AppInstaller::Manifest::ManifestError::DuplicateInstallerEntry, "Duplicate installer entry found." },
+            { AppInstaller::Manifest::ManifestError::InstallerTypeDoesNotSupportPackageFamilyName, "The specified installer type does not support PackageFamilyName." },
+            { AppInstaller::Manifest::ManifestError::InstallerTypeDoesNotSupportProductCode, "The specified installer type does not support ProductCode." },
+            { AppInstaller::Manifest::ManifestError::InstallerTypeDoesNotWriteAppsAndFeaturesEntry, "The specified installer type does not write to Apps and Features entry." },
```
Trenly, line 29:
```suggestion
            { AppInstaller::Manifest::ManifestError::InstallerTypeDoesNotWriteAppsAndFeaturesEntry, "The specified installer type does not write an entry to Apps and Features." },
```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
...
+            { AppInstaller::Manifest::ManifestError::IncompleteMultiFileManifest, "The multi file manifest is incomplete.A multi file manifest must contain at least version, installer and defaultLocale manifest."},
```
Trenly, line 30:
```suggestion
            { AppInstaller::Manifest::ManifestError::IncompleteMultiFileManifest, "The multi file manifest is incomplete. A multi file manifest must contain at least a version, installer, and defaultLocale manifest file."},

```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
...
+            { AppInstaller::Manifest::ManifestError::InconsistentMultiFileManifestFieldValue, "The multi file manifest has inconsistent field values." },
+            { AppInstaller::Manifest::ManifestError::DuplicateMultiFileManifestType, "The multi file manifest should contain only one file with the particular ManifestType." },
+            { AppInstaller::Manifest::ManifestError::DuplicateMultiFileManifestLocale, "The multi file manifest contains duplicate PackageLocale." },
+            { AppInstaller::Manifest::ManifestError::UnsupportedMultiFileManifestType, "The multi file manifest should not contain file with the particular ManifestType." },
```
Trenly, line 34:
```suggestion
            { AppInstaller::Manifest::ManifestError::UnsupportedMultiFileManifestType, "The multi file manifest should not contain a file with the particular ManifestType." },

```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
...
+            { AppInstaller::Manifest::ManifestError::ExceededCommandsLimit, "Only zero or one value for Commands may be specified for InstallerType portable." },
+            { AppInstaller::Manifest::ManifestError::ScopeNotSupported, "Scope is not supported for InstallerType portable." },
+            { AppInstaller::Manifest::ManifestError::InstallerMsixInconsistencies, "Inconsistent value in the manifest." },
+            { AppInstaller::Manifest::ManifestError::OptionalFieldMissing, "Optional field missing." },
+            { AppInstaller::Manifest::ManifestError::InstallerFailedToProcess, "Failed to process installer." },
+            { AppInstaller::Manifest::ManifestError::NoSupportedPlatforms, "No supported platforms." },
+            { AppInstaller::Manifest::ManifestError::ApproximateVersionNotAllowed, "Approximate version not allowed." },
+            { AppInstaller::Manifest::ManifestError::ArpVersionOverlapWithIndex, "DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index.Existing DisplayVersion range in index : " },
```
Trenly, line 54:
```suggestion
            { AppInstaller::Manifest::ManifestError::ArpVersionOverlapWithIndex, "DisplayVersion declared in the manifest has overlap with existing DisplayVersion range in the index. Existing DisplayVersion range in index : " },
```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -10,6 +10,55 @@
...
+            { AppInstaller::Manifest::ManifestError::ArpVersionValidationInternalError, "Internal error while validating DisplayVersion against index." },
+            { AppInstaller::Manifest::ManifestError::ExceededNestedInstallerFilesLimit, "Only one entry for NestedInstallerFiles can be specified for non-portable InstallerTypes." },
+            { AppInstaller::Manifest::ManifestError::RelativeFilePathEscapesDirectory, "Relative file path must not point to a location outside of archive directory" },
+            { AppInstaller::Manifest::ManifestError::ArpValidationError, "Arp Validation Error" },
```
Trenly, line 58:
```suggestion
            { AppInstaller::Manifest::ManifestError::ArpValidationError, "ARP Validation Error" },

```

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
JohnMcPMS: Depending on the context it might mean a `constexpr` construction of a view at compile time (with `sv`), or a runtime construction that needs to look for the null char.

But some contexts it doesn't matter.

Better to put it by default than to need to understand exactly when it will be one or the other.

src/AppInstallerRepositoryCore/ArpVersionValidation.cpp
JohnMcPMS: I suggested that it was now "context" rather than "field".  Would be good to change the name to reflect that.

JohnMcPMS:
Note to other reviewers: The main goal of the change was to get the specifics of the errors out of the function, but I suggested that it be done this way in order to move 1 step closer to actually localizing these messages (and any/all others in common).

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
hackean-msft: Ditto @JohnMcPMS's comment https://github.com/microsoft/winget-cli/pull/2410#issuecomment-1204531101. This change is simply about surfacing specific errors. A look at some of the comments indicates that some more changes could be made in the unit tests, way beyond the scope of the change. 

This might have the ripple of effect of demanding more changes to tests/code that rely on this externally in turn blowing up the scope of externally changes. I'm a bit hesitant to change these messages.

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,38 @@
...
+                {ManifestError::MissingManifestDependenciesNode, WinGetManifestDependenciesErrorResult::MissingManifestDependenciesNode },
+                {ManifestError::NoSuitableMinVersionDependency, WinGetManifestDependenciesErrorResult::NoSuitableMinVersionDependency },
+                {ManifestError::FoundDependencyLoop, WinGetManifestDependenciesErrorResult::FoundDependencyLoop}
+            };
+
+            auto itr = dependenciesErrorMessageMap.find(message);
+
+            if (itr != dependenciesErrorMessageMap.end())
+            {
+                return itr->second;
```
JohnMcPMS, line 32: Shouldn't we `||` together all of the errors that we find?

JohnMcPMS (changes requested):

JohnMcPMS (commented):
Removing my blocking status for now.

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
hackean-msft: Dependencies error are mutually exclusive, Renamed the method name to properly convey its functionality.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -11,6 +11,61 @@
 
 namespace AppInstaller::Manifest
 {
+    namespace
+    {
+        auto GetErrorIdToMessageMap()
```
lechacon, line 16: `const auto& GetErrorIdToMessageMap()` to avoid creating a copy each time.

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -71,59 +73,61 @@ namespace AppInstaller::Manifest
...
+            ValidationError error{ message, context };
             error.FileName = file;
             return error;
         }
 
-        static ValidationError MessageFieldValueWithFile(std::string message, std::string field, std::string value, std::string file)
+        static ValidationError MessageFieldValueWithFile(AppInstaller::StringResource::StringId message, std::string context, std::string value, std::string file)
         {
-            ValidationError error{ message, field, value };
+            ValidationError error{ message, context, value };
```
lechacon, line 130: We should rename these since it is no longer `Field`.

src/AppInstallerRepositoryCore/AppInstallerRepositoryCore.vcxproj.filters
```diff
@@ -464,12 +467,6 @@
     <ClCompile Include="ArpVersionValidation.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="Microsoft\Schema\1_6\Interface_1_6.cpp">
-      <Filter>Microsoft\Schema\1_6</Filter>
-    </ClCompile>
-    <ClCompile Include="Microsoft\Schema\1_6\SearchResultsTable_1_6.cpp">
-      <Filter>Microsoft\Schema\1_6</Filter>
-    </ClCompile>
```
lechacon, line 472: Why is this deleted?

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -82,23 +82,23 @@ namespace AppInstaller::Repository
...
+            std::vector<Manifest::ValidationError> validationErrors;
+            validationErrors.emplace_back(Manifest::ValidationError{error, "PackageIdentifier.PackageVersion", itrStart->first.Id + "." + itrStart->first.Version });
 
             std::for_each(
                 itrStart + 1,
                 failedManifests.end(),
                 [&](std::pair<DependentManifestInfo, Utility::Version> current)
                 {
-                    dependentPackages.append(", " + current.first.Id + "." + current.first.Version);
+                    validationErrors.emplace_back(Manifest::ValidationError{ error, "PackageIdentifier.PackageVersion", current.first.Id + "." + current.first.Version });
```
lechacon, line 96: I think if you are calling `emplace_back` you can omit the constructor and just pass the params. Like
`validationErrors.emplace_back(error, "PackageIdentifier.PackageVersion", current.first.Id + "." + current.first.Version);`

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
```
lechacon, line 3: `#pragma once` in a `.cpp` file?

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,38 @@
...
+            auto itr = dependenciesErrorMessageMap.find(message);
+
+            if (itr != dependenciesErrorMessageMap.end())
+            {
+                return itr->second;
+            }
+        }
+
+        return WinGetManifestDependenciesErrorResult::None;
+	}
```
lechacon, line 37: Nit: Indentation

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "pch.h"
+#include "PackageDependenciesValidationUtil.h"
+#include <winget/ManifestValidation.h>
+
+namespace AppInstaller::Repository::Util
+{
+    using namespace AppInstaller::Manifest;
+
+    WinGetManifestDependenciesErrorResult GetDependenciesValidationResultFromException(const AppInstaller::Manifest::ManifestException& manifestException)
+	{
```
lechacon, line 13: Nit: Indentation

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,38 @@
...
+        {
+            auto message = validationError.Message;
+
+            std::map<AppInstaller::StringResource::StringId, WinGetManifestDependenciesErrorResult> dependenciesErrorMessageMap =
+            {
+                {ManifestError::SingleManifestPackageHasDependencies, WinGetManifestDependenciesErrorResult::SingleManifestPackageHasDependencies},
+                {ManifestError::MultiManifestPackageHasDependencies, WinGetManifestDependenciesErrorResult::MultiManifestPackageHasDependencies },
+                {ManifestError::MissingManifestDependenciesNode, WinGetManifestDependenciesErrorResult::MissingManifestDependenciesNode },
+                {ManifestError::NoSuitableMinVersionDependency, WinGetManifestDependenciesErrorResult::NoSuitableMinVersionDependency },
+                {ManifestError::FoundDependencyLoop, WinGetManifestDependenciesErrorResult::FoundDependencyLoop}
```
lechacon, line 25: Nit: Inconsistent spaces before closing brackets

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <winget/ManifestValidation.h>
+
+namespace AppInstaller::Repository::Util
+{
+    using namespace AppInstaller::Manifest;
+
+    WinGetManifestDependenciesErrorResult GetDependenciesValidationResultFromException(const AppInstaller::Manifest::ManifestException& manifestException)
+	{
+        auto validationErrors = manifestException.Errors();
+        for (auto validationError : validationErrors)
```
lechacon, line 15: Could these be `const auto&` to avoid copies?

src/WinGetUtil/Exports.cpp
```diff
@@ -356,7 +357,9 @@ extern "C"
             }
             catch (const ManifestException& e)
             {
-                WI_SetFlagIf(validationResult, WinGetValidateManifestResult::DependenciesValidationFailure, !e.IsWarningOnly());
+                validationResult |= WinGetValidateManifestResult::DependenciesValidationFailure;
```
lechacon, line 360: We are converting these from warnings to actual errors. Is there anybody that could break due to this change?

lechacon (approved):

src/WinGetUtil/Exports.cpp
```diff
@@ -356,7 +357,9 @@ extern "C"
             }
             catch (const ManifestException& e)
             {
-                WI_SetFlagIf(validationResult, WinGetValidateManifestResult::DependenciesValidationFailure, !e.IsWarningOnly());
+                validationResult |= WinGetValidateManifestResult::DependenciesValidationFailure;
+                validationResult |= static_cast<WinGetValidateManifestResult>( AppInstaller::Repository::Util::GetDependenciesValidationResultFromException(e));
```
yao-msft, line 361: >  [](http://example.com/codeflow?start=78&length=1)

nit: extra space

src/WinGetUtil/Exports.cpp
yao-msft: Even if we don't have warning level dependencies now, we should still only set the error when ther level is error, in case we have warning level in the future

src/AppInstallerRepositoryCore/Rest/Schema/1_0/RestInterface_1_0.cpp
```diff
@@ -241,7 +241,7 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
             {
                 if (error.ErrorLevel == Manifest::ValidationError::Level::Error)
                 {
-                    AICLI_LOG(Repo, Error, << "Received manifest contains validation error: " << error.Message);
+                    AICLI_LOG(Repo, Error, << "Received manifest contains validation error: " << Utility::ConvertToUTF8(error.Message));
```
yao-msft, line 244: >Utility::ConvertToUTF8(error.Message) [](http://example.com/codeflow?start=97&length=37)

error.GetErrorMessage()

we want to log the actual error message

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -71,59 +73,61 @@ namespace AppInstaller::Manifest
...
-        static ValidationError MessageWithFile(std::string message, std::string file)
+        static ValidationError MessageWithFile(AppInstaller::StringResource::StringId message, std::string file)
         {
             ValidationError error{ message };
             error.FileName = file;
             return error;
         }
 
-        static ValidationError MessageFieldWithFile(std::string message, std::string field, std::string file)
+        static ValidationError MessageFieldWithFile(AppInstaller::StringResource::StringId message, std::string context, std::string file)
```
yao-msft, line 121: >MessageField [](http://example.com/codeflow?start=31&length=12)

I guess this should be called MessageWithContextFile now?

And same for below
MessageWithContextValueFile

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -318,4 +373,17 @@ namespace AppInstaller::Manifest
 
         return errors;
     }
+
+    std::string ValidationError::GetErrorMessage() const
+    {
+        auto ErrorIdToMessageMap = GetErrorIdToMessageMap();
```
yao-msft, line 379: >auto [](http://example.com/codeflow?start=8&length=4)

const auto& here to avoid another copy

src/AppInstallerCLITests/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,23 @@
+#include "pch.h"
```
yao-msft, line 1: nit: include copyright headers

src/AppInstallerCLITests/PackageDependenciesValidationUtil.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "TestSource.h"
+#include "PackageDependenciesValidationUtil.h"
+
+using namespace AppInstaller::Repository::Util;
+
+TEST_CASE("GetValidationResultFromException", "[PackageDependenciesValidationUtil][dependencies]")
+{
+    std::map<WinGetManifestDependenciesErrorResult, const AppInstaller::Manifest::ManifestException> dependenciesErrorMessageMap =
+    {
+        {WinGetManifestDependenciesErrorResult::SingleManifestPackageHasDependencies, AppInstaller::Manifest::ManifestException {  { AppInstaller::Manifest::ManifestError::SingleManifestPackageHasDependencies }, APPINSTALLER_CLI_ERROR_MANIFEST_FAILED }},
```
yao-msft, line 12: >{WinGetManifestDependenciesErrorResult::SingleManifestPackageHasDependencies, AppInstaller::Manifest::ManifestException {  { AppInstaller::Manifest::ManifestError::SingleManifestPackageHasDependencies }, APPINSTALLER_CLI_ERROR_MANIFEST_FAILED }}, [](http://example.com/codeflow?start=8&length=246)

nit: the spacing for these lines looks off

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
```
yao-msft, line 1: This header is used outside of the RepositoryCore project, so it needs to be put under Public/winget and exposed there

And imo, this is easily expandable to other ManifestException types in the future, we could probably put it in CommonCore beside ManifestValidation.h, or just declare the function in  ManifestValidation.h

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.cpp
yao-msft: I think || is more future proof

src/AppInstallerRepositoryCore/PackageDependenciesValidation.cpp
```diff
@@ -82,23 +82,23 @@ namespace AppInstaller::Repository
         }
     
         void ThrowOnManifestValidationFailed(
-            std::vector<std::pair<DependentManifestInfo, Utility::Version>> failedManifests, std::string error)
+            std::vector<std::pair<DependentManifestInfo, Utility::Version>> failedManifests, AppInstaller::StringResource::StringId error)
         {
             auto itrStart = failedManifests.begin();
-            std::string dependentPackages{ itrStart->first.Id + "." + itrStart->first.Version };
+            std::vector<Manifest::ValidationError> validationErrors;
+            validationErrors.emplace_back(Manifest::ValidationError{error, "PackageIdentifier.PackageVersion", itrStart->first.Id + "." + itrStart->first.Version });
```
yao-msft, line 89: >"." [](http://example.com/codeflow?start=132&length=3)

nit: '.'

src/AppInstallerRepositoryCore/ArpVersionValidation.cpp
```diff
@@ -67,11 +67,12 @@ namespace AppInstaller::Repository
             {
                 if (manifestArpVersionRange.Overlaps(arpInIndex))
                 {
-                    std::string errorMsg = Manifest::ManifestError::ArpVersionOverlapWithIndex;
-                    errorMsg.append("[" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
-                    AICLI_LOG(Repo, Error, << errorMsg);
+                    std::string context = (" [" + arpInIndex.GetMinVersion().ToString() + ", " + arpInIndex.GetMaxVersion().ToString() + "]");
+                    auto validationError = Manifest::ValidationError(Manifest::ManifestError::ArpVersionOverlapWithIndex, context);
+                    
+                    AICLI_LOG(Repo, Error, << validationError.GetErrorMessage());
```
yao-msft, line 73: >AICLI_LOG(Repo, Error, << validationError.GetErrorMessage()); [](http://example.com/codeflow?start=20&length=61)

Please also log the context, this is the important part for log viewer

yao-msft (dismissed):
:clock1:

src/AppInstallerRepositoryCore/ArpVersionValidation.cpp
hackean-msft: The context is part of the error message, previous implementation does log the message with the assumption, i don't see why it should be any different here.

src/AppInstallerRepositoryCore/PackageDependenciesValidationUtil.h
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+#pragma once
+#include <winget/ManifestValidation.h>
+
+namespace AppInstaller::Repository::Util
+{
+    enum WinGetManifestDependenciesErrorResult
```
yao-msft, line 8: >enum WinGetManifestDependenciesErrorResult [](http://example.com/codeflow?start=4&length=42)

enum class WinGetManifestDependenciesErrorResult: uint32_t
    {
        .....
    };

    DEFINE_ENUM_FLAG_OPERATORS(WinGetManifestDependenciesErrorResult);

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -71,59 +73,61 @@ namespace AppInstaller::Manifest
...
+        static ValidationError MessageContextWithFile(AppInstaller::StringResource::StringId message, std::string context, std::string file)
         {
-            ValidationError error{ message, field };
+            ValidationError error{ message, context };
             error.FileName = file;
             return error;
         }
 
-        static ValidationError MessageFieldValueWithFile(std::string message, std::string field, std::string value, std::string file)
+        static ValidationError MessageFieldValueWithFile(AppInstaller::StringResource::StringId message, std::string context, std::string value, std::string file)
```
yao-msft, line 128: >MessageFieldValueWithFile [](http://example.com/codeflow?start=31&length=25)

MessageContextValueWithFile

src/AppInstallerRepositoryCore/AppInstallerRepositoryCore.vcxproj.filters
```diff
@@ -489,16 +489,16 @@
...
-      <Filter>Microsoft\Schema\Portable_1_0</Filter>
+      <Filter>Source Files</Filter>
     </ClCompile>
     <ClCompile Include="Microsoft\Schema\Portable_1_0\PortableTable.cpp">
-      <Filter>Microsoft\Schema\Portable_1_0</Filter>
+      <Filter>Source Files</Filter>
     </ClCompile>
     <ClCompile Include="Microsoft\SQLiteStorageBase.cpp">
-      <Filter>Microsoft</Filter>
+      <Filter>Source Files</Filter>
```
yao-msft, line 501: > <Filter>Source Files</Filter> [](http://example.com/codeflow?start=5&length=30)

The changes in this file is incorrect(merge error)

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -11,6 +11,61 @@
 
 namespace AppInstaller::Manifest
 {
+    namespace
+    {
+        auto& GetErrorIdToMessageMap()
```
yao-msft, line 16: >auto& [](http://example.com/codeflow?start=8&length=5)

const auto&

src/AppInstallerRepositoryCore/ArpVersionValidation.cpp
yao-msft: Previous logged message is "Arp Version Overlap [1.0, 2.0]", the range is part of the message
With the new code, the failing range is in the context part of the ValidationError, new code using GetErrorMessage() only logs "Arp Version Overlap" without the violating version range, which is less helpful.

yao-msft (approved):

PR #2413, opened 2022-08-03
Ensure default values are passed to nested installers within zip

### LINKED ISSUE #2392: Zip installers not getting nested installer type defaults

Installers within zip files are not getting the default values for their nested installer, such as switches and error codes.

### Steps to reproduce

Use zip.

### Expected behavior

Works as if not zipped.

### Actual behavior

Has no command line arguments passed.

### Environment

```shell
CI/CD
```

ryfu-msft:
Fixes #2392 

Changes:
- Changes installer.InstallerType -> installer.BaseInstallerType.
- Created a method called EffectiveInstallerType() that returns the actual installerType that will be executed (i.e. the NestedInstallerType if it is a zip installer)

Tests:
- Reenabled ZipWithMsi test
- All existing tests should pass

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2413)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
@@ -33,7 +33,7 @@ namespace AppInstaller::CLI::Workflow
         std::wstring_view GetInstallerFileExtension(Execution::Context& context)
         {
             const auto& installer = context.Get<Execution::Data::Installer>();
-            switch (installer->InstallerType)
+            switch (installer->BaseInstallerType)
             {
             case InstallerTypeEnum::Burn:
             case InstallerTypeEnum::Exe:
@@ -143,7 +143,7 @@ namespace AppInstaller::CLI::Workflow
         if (!context.Contains(Execution::Data::InstallerPath))
         {
             const auto& installer = context.Get<Execution::Data::Installer>().value();
-            switch (installer.InstallerType)
+            switch (installer.BaseInstallerType)
             {
             case InstallerTypeEnum::Exe:
             case InstallerTypeEnum::Burn:
@@ -183,7 +183,7 @@ namespace AppInstaller::CLI::Workflow
     void CheckForExistingInstaller(Execution::Context& context)
     {
         const auto& installer = context.Get<Execution::Data::Installer>().value();
-        if (installer.InstallerType == InstallerTypeEnum::MSStore)
+        if (installer.EffectiveInstallerType() == InstallerTypeEnum::MSStore)
         {
             // No installer is downloaded in this case
             return;
@@ -415,12 +415,12 @@ namespace AppInstaller::CLI::Workflow
             auto existingFileHash = SHA256::ComputeHash(inStream);
             context.Add<Execution::Data::HashPair>(std::make_pair(installer.Sha256, existingFileHash));
         }
-        else if (installer.InstallerType == InstallerTypeEnum::MSStore)
+        else if (installer.EffectiveInstallerType() == InstallerTypeEnum::MSStore)
         {
             // No installer file in this case
             return;
         }
-        else if (installer.InstallerType == InstallerTypeEnum::Msix && !installer.SignatureSha256.empty())
+        else if (installer.EffectiveInstallerType() == InstallerTypeEnum::Msix && !installer.SignatureSha256.empty())
         {
             // We didn't download the installer file before. Just verify the signature hash again.
             context << GetMsixSignatureHash;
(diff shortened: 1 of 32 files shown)


src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
```diff
@@ -183,7 +183,7 @@ namespace AppInstaller::CLI::Workflow
     void CheckForExistingInstaller(Execution::Context& context)
     {
         const auto& installer = context.Get<Execution::Data::Installer>().value();
-        if (installer.InstallerType == InstallerTypeEnum::MSStore)
+        if (installer.EffectiveInstallerType() == InstallerTypeEnum::MSStore)
```
JohnMcPMS, line 186: [This is the correct usage here just calling this out if it isn't already done...]
As `MSStore` is not an installer type that actually has an installer file (to us), it should not be a valid nested type.  Is that already handled elsewhere?

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -173,7 +173,7 @@ namespace AppInstaller::CLI::Workflow
         info << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelInstaller << std::endl;
         if (installer)
         {
-            info << "  "_liv << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelInstallerType << ' ' << Manifest::InstallerTypeToString(installer->InstallerType) << std::endl;
+            info << "  "_liv << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelInstallerType << ' ' << Manifest::InstallerTypeToString(installer->BaseInstallerType) << std::endl;
```
JohnMcPMS, line 176: I think that this should be something like:
```
Effective (Base)
```
When effective and base are not the same.  The important part is the effective type, but also showing the base is useful.

src/AppInstallerRepositoryCore/Rest/Schema/1_1/Json/ManifestDeserializer_1_1.cpp
```diff
@@ -158,7 +158,7 @@ namespace AppInstaller::Repository::Rest::Schema::V1_1::Json
             }
 
             // Populate installer default return codes if not present in ExpectedReturnCodes and InstallerSuccessCodes
-            auto defaultReturnCodes = GetDefaultKnownReturnCodes(installer.InstallerType);
+            auto defaultReturnCodes = GetDefaultKnownReturnCodes(installer.BaseInstallerType);
```
JohnMcPMS, line 161: Effective, unless it is not yet know, then some more effort will be required.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/ManifestDeserializer_1_0.cpp
```diff
@@ -366,7 +366,7 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0::Json
         }
 
         // Installer Switches
-        installer.Switches = Manifest::GetDefaultKnownSwitches(installer.InstallerType);
+        installer.Switches = Manifest::GetDefaultKnownSwitches(installer.BaseInstallerType);
```
JohnMcPMS, line 369: Effective, unless it is not yet know, then some more effort will be required.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/ManifestDeserializer_1_0.cpp
ryfu-msft: Changed to effective

src/AppInstallerRepositoryCore/Rest/Schema/1_1/Json/ManifestDeserializer_1_1.cpp
ryfu-msft: Changed to effective

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
ryfu-msft: Changed to suggested. 
 
If effective and base are not the same, will show both "Effective (Base)". Added tests to verify output

src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
ryfu-msft: The only check I had was from manifest validation. I added another check prior to the install flow to verify that if the NestedInstallertype is valid if the BaseInstallertype is an archive. Outputs the appropriate error message if NestedInstallertype is not supported.

JohnMcPMS (approved):

PR #2415, opened 2022-08-03
Collect scope in installer metadata

JohnMcPMS:
## Change
Include the installed scope during metadata collection.  This uses the current detection method, which is "HKLM or HKCU?"

When there us a mismatch, the value will be explicitly set to `Unknown`.

## Validation
New unit tests added, mostly for the mismatch cases.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2415)


yao-msft (approved):
:shipit:

Issue #1345, opened 2021-08-04
Add Purchase URL to schema


denelon:
# Description of the new feature/enhancement

Some packages require some form of "entitlement" to install a particular package. When a user is interacting with a source, some apps may require purchase or parental consent. The canonical example is the Microsoft Store. If the user is not authenticated or does not posses an entitlement for the App, the user should be provided a Purchase URL so they can acquire the entitlement. In the case of the Microsoft Store this could be a deep link to the store where the user can authenticate, purchase, or request parental consent.

# Proposed technical implementation details (optional)

REST API Issue - https://github.com/microsoft/winget-cli-restsource/issues/54

vedantmgoyal2009:
* @denelon the issue can be closed here while the issue at the https://github.com/microsoft/winget-cli-restsource/issues/54 can left opened, if not yet implemented.

PR #2416, opened 2022-08-04
Show PurchaseUrl if present in manifest

Trenly:

This does not take into account any of the entitlement checking for the MSStore source; The only change this should have is to Pre-Indexed and Rest sources where, if the PurchaseUrl is present in the manifest, the URL will be displayed.

As part of this I have also taken the many if statements for checking if fields are empty and moved them into a function for ease of readability.

Loosely related to #1345

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2416)


src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -12,8 +12,28 @@ using namespace AppInstaller::Utility;
 using namespace AppInstaller::Utility::literals;
 
 namespace {
-    void ShowMultiLineField(Execution::OutputStream& outputStream, AppInstaller::StringResource::StringId label, std::string& value)
+
+    template <typename CharT, typename String>
+    void ShowSingleLineField(Execution::OutputStream& outputStream, CharT label, String value, bool indent = false)
```
JohnMcPMS, line 17: I didn't see anything but `StringID` values for label (and I wouldn't expect anything else).  I don't think it needs to be a templated type.  Applies to `ShowMultiLineField` as well.

And it's best to use `const String& value` to show the intent (we do not want to modify or own this parameter).

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -38,111 +72,49 @@ namespace AppInstaller::CLI::Workflow
...
-            {
-                info << "  "_liv << tag << std::endl;
-            }
-        }
+        ShowMultiValueField(info, Resource::String::ShowLabelTags, manifest.CurrentLocalization.Get<Manifest::Localization::Tags>());
         const auto& agreements = manifest.CurrentLocalization.Get<Manifest::Localization::Agreements>();
         if (!agreements.empty())
         {
-            context.Reporter.Info() << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelAgreements << std::endl;
+            ShowSingleLineField(info, Resource::String::ShowLabelAgreements, " "_liv);
```
JohnMcPMS, line 117: Not a fan of reusing the function for something that is not semantically correct.

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -170,30 +142,15 @@ namespace AppInstaller::CLI::Workflow
...
-                info << "  "_liv << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelInstallerProductId << ' ' << installer->ProductId << std::endl;
-            }
-            if (!installer->ReleaseDate.empty())
-            {
-                info << "  "_liv << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelInstallerReleaseDate << ' ' << installer->ReleaseDate << std::endl;
-            }
+            ShowSingleLineField(info, Resource::String::ShowLabelInstallerType, Manifest::InstallerTypeToString(installer->InstallerType), true);
+            ShowSingleLineField(info, Resource::String::ShowLabelInstallerLocale, installer->Locale, true);
+            ShowSingleLineField(info, Resource::String::ShowLabelInstallerUrl, installer->Url, true);
+            ShowSingleLineField(info, Resource::String::ShowLabelInstallerSha256, (installer->InstallerType == Manifest::InstallerTypeEnum::MSStore) ? "" : Utility::SHA256::ConvertToString(installer->Sha256), true);
```
JohnMcPMS, line 151: The check should be based on the value at hand, not inferred from the installer type.

ryfu-msft (approved):

ryfu-msft:
Looks good to me, will defer to @yao-msft for sign-off and merge

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -12,17 +12,57 @@ using namespace AppInstaller::Utility;
 using namespace AppInstaller::Utility::literals;
 
 namespace {
-    void ShowMultiLineField(Execution::OutputStream& outputStream, AppInstaller::StringResource::StringId label, std::string& value)
+
+    template <typename String>
+    void ShowSingleLineField(Execution::OutputStream outputStream, const AppInstaller::StringResource::StringId& label, const String& value, bool indent = false)
```
yao-msft, line 17: nit: StringId is actually a string_view, they'd better be passed by value.

Same for the MultiLine and multivalue one

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -12,17 +12,57 @@ using namespace AppInstaller::Utility;
...
+        if (value.empty())
+        {
+            return;
+        }
+        /*
+            We need to be able to find and replace within the string.However, we don't want to own the original string
+            Therefore, a copy is created here so we can manipulate it. The memory should be freed again once this method
+            returns and the string is no longer in scope.
+        */
+        std::string m_value = value;
```
yao-msft, line 42: nit: m_value is not a class member, maybe name it shownValue or something better?

src/AppInstallerCLICore/Workflows/ShowFlow.cpp
```diff
@@ -175,36 +154,18 @@ namespace AppInstaller::CLI::Workflow
...
-            if (!installer->Sha256.empty())
-            {
-                info << "  "_liv << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelInstallerSha256 << ' ' << Utility::SHA256::ConvertToString(installer->Sha256) << std::endl;
-            }
-            if (!installer->ProductId.empty())
-            {
-                info << "  "_liv << Execution::ManifestInfoEmphasis << Resource::String::ShowLabelInstallerProductId << ' ' << installer->ProductId << std::endl;
-            }
-            if (!installer->ReleaseDate.empty())
+            std::stringstream shownInstallerType;
```
yao-msft, line 157: I'm not sure if a stringstream is necessary here, can it be just the std::string

PR #2418, opened 2022-08-04
Use `UpgradeCode` for matching MSI apps

### LINKED ISSUE #1145: Add MSI UpgradeCode to manifest
# MSI Installer UpgradeCode

MSI Installers have an UpgradeCode to help with package upgrades:
[Using an UpgradeCode]https://docs.microsoft.com/en-us/windows/win32/msi/using-an-upgradecode)

lechacon:
This change adds support for using an installer's `UpgradeCode` from the `AppsAndFeaturesEntries` for matching against the installed apps.

* Added a new version of the index interface to create a new table for holding the UpgradeCode and a new version of the SearchResultsTable
* Added UpgradeCode as a `PackageVersionMultiProperty` and `SystemReferenceString`
* Updated ARP correlation to also consider UpgradeCode
* Updated the ProductCode and PackageFamilyName tables to also include the data from `AppsAndFeaturesEntries`
* Added extraction of the UpgradeCode for MSI apps when reading the ARP

Closes #1145

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2418)

### DIFF
--- a/src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
@@ -106,18 +106,21 @@ namespace AppInstaller::Repository::Microsoft
             std::map<std::string, std::string> upgradeCodes;
 
             Registry::Key upgradeCodesKey = Registry::Key::OpenIfExists(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Installer\\UpgradeCodes");
-            for (const auto& upgradeCodeKeyRef : upgradeCodesKey)
+            if (upgradeCodesKey)
             {
-                auto upgradeCode = TryUnpackUpgradeCodeGuid(upgradeCodeKeyRef.Name());
-                if (upgradeCode)
+                for (const auto& upgradeCodeKeyRef : upgradeCodesKey)
                 {
-                    auto upgradeCodeKey = upgradeCodeKeyRef.Open();
-                    for (const auto& productCodeValue : upgradeCodeKey.Values())
+                    auto upgradeCode = TryUnpackUpgradeCodeGuid(upgradeCodeKeyRef.Name());
+                    if (upgradeCode)
                     {
-                        auto productCode = TryUnpackUpgradeCodeGuid(productCodeValue.Name());
-                        if (productCode)
+                        auto upgradeCodeKey = upgradeCodeKeyRef.Open();
+                        for (const auto& productCodeValue : upgradeCodeKey.Values())
                         {
-                            upgradeCodes[*productCode] = *upgradeCode;
+                            auto productCode = TryUnpackUpgradeCodeGuid(productCodeValue.Name());
+                            if (productCode)
+                            {
+                                upgradeCodes[*productCode] = *upgradeCode;
+                            }
                         }
                     }
                 }
--- a/src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
@@ -340,6 +340,23 @@ namespace AppInstaller::Repository::Microsoft
                     //manifest.Id = normalizedName.Publisher() + '.' + normalizedName.Name();
                 }
 
+                // Pick up WindowsInstaller to determine if this is an MSI install.
+                // TODO: Could also determine Inno (and maybe other types) through detecting other keys here.
+                auto installedType = Manifest::InstallerTypeEnum::Exe;
+
+                if (GetBoolValue(arpKey, WindowsInstaller))
+                {
+                    installedType = Manifest::InstallerTypeEnum::Msi;
+
+                    // If this is an MSI, look up the UpgradeCode
+                    auto upgradeCode = GetUpgradeCodeByProductCode(productCode);
+                    if (upgradeCode)
+                    {
+                        manifest.Installers[0].AppsAndFeaturesEntries.emplace_back();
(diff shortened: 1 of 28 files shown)


src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
```diff
@@ -3,11 +3,57 @@
...
+        {
+            // The UpgradeCode is not stored in the ARP registry keys, so we use the MSI API to query it.
+            //
+            // The UpgradeCode is apparently also stored in the registry under
+            //   HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UpgradeCodes\<UpgradeCode>
+            // but:
+            //   * This key does not seem to be documented
+            //   * The format used to store the GUIDs is harder to parse
+            //   * We cannot look up by ProductCode; we need to take an UpgradeCode and then find the ProductCode
+            // Using the registry could potentially be faster, so we may consider using it if this proves to be too slow.
```
lechacon, line 25: This turned out to be quite slow once I actually tested it, so I'll look into other ways of getting the UpgradeCode

src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/Interface_1_1.cpp
```diff
@@ -29,25 +29,36 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_1
     {
         std::vector<Utility::NormalizedString> GetSystemReferenceStrings(
             const Manifest::Manifest& manifest,
-            std::function<const Utility::NormalizedString&(const Manifest::ManifestInstaller&)> func)
+            std::function<const Utility::NormalizedString&(const Manifest::ManifestInstaller&)> extractStringFromInstaller,
+            std::optional<std::function<const Utility::NormalizedString& (const Manifest::AppsAndFeaturesEntry&)>> extractStringFromAppsAndFeaturesEntry = {})
```
JohnMcPMS, line 33: A `std::function` can already be empty; no need for a `std::optional` wrapper.

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
```diff
@@ -3,11 +3,133 @@
...
 
 namespace AppInstaller::Repository::Microsoft
 {
     using namespace AppInstaller::Registry::Portable;
 
+    namespace
+    {
+        // "Unpacks" a GUID in the format used by the UpgradesCode registry key into the usual format.
+        // Returns empty if it is not a valid GUID
+        std::optional<std::string> TryUnpackUpgradeCodeGuid(std::string_view packed)
```
JohnMcPMS, line 16: As discussed, there is a simpler way to do this just using a transformation map from the given 32 character input into the desired GUID-like output.  Up to you on whether to change to that.

I do want to confirm that the output format here does match the value that would be returned by `MsiGetProperty`, as wingetcreate et al. will likely use that function to read the property out of the MSI files themselves when creating manifests.

JohnMcPMS (approved):

src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/Interface_1_1.cpp
lechacon: Thanks. Didn't know that

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
lechacon: Changed to use the transformation map.

Yes, it does match.

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
```diff
@@ -3,11 +3,133 @@
...
+            std::string_view part2 = packed.substr(Part2Start, Part2Length);
+            std::string_view part3 = packed.substr(Part3Start, Part3Length);
+
+            std::string_view part4[Part4Count];
+            for (size_t i = 0; i < Part4Count; ++i)
+            {
+                part4[i] = packed.substr(Part4Start + i * Part4BlockLength, Part4BlockLength);
+            }
+
+            std::string unpacked;
```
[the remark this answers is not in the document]
lechacon: I'm now initializing to a GUID with just 0s to set the characters that are not mapped.

src/AppInstallerRepositoryCore/Microsoft/ARPHelper.cpp
```diff
@@ -8,6 +8,102 @@ namespace AppInstaller::Repository::Microsoft
...
+            //
+            // Under 'UpgradeCodes' there is one key for each upgrade code, and each upgrade code key
+            // contains the product code as a value. All the upgrade codes and product codes are GUIDs,
+            // but represented in an unusual way - see TryUnpackUpgradeCodeGuid()
+
+            AICLI_LOG(Repo, Info, << "Reading MSI UpgradeCodes");
+            std::map<std::string, std::string> upgradeCodes;
+
+            // There is no UpgradeCodes key on the x86 view of the registry
+            Registry::Key upgradeCodesKey = Registry::Key::OpenIfExists(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Installer\\UpgradeCodes", 0, KEY_READ | KEY_WOW64_64KEY);
```
lechacon, line 81: Small important change since the review to fix test failure.

JohnMcPMS: It does make sense that these would only be stored in the OS architecture.

JohnMcPMS (approved):

PR #2420, opened 2022-08-05
Expose deeper installation detection through Com 

yao-msft:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2420)


src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -247,6 +248,233 @@ namespace AppInstaller::Repository
...
+
+            return installLocationStatus;
+        }
+
+        // Map to cache already calculated file hashes.
+        struct FilePathComparator
+        {
+            bool operator()(const std::filesystem::path& a, const std::filesystem::path& b) const
+            {
+                if (std::filesystem::equivalent(a, b))
```
JohnMcPMS, line 272: This part seems unnecessary if the goal is producing `a < b`.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -247,6 +248,233 @@ namespace AppInstaller::Repository
...
+        {
+            HRESULT fileStatus = WINGET_INSTALLED_STATUS_FILE_NOT_FOUND;
+            try
+            {
+                if (std::filesystem::exists(filePath) && std::filesystem::is_regular_file(filePath))
+                {
+                    fileStatus = WINGET_INSTALLED_STATUS_FILE_FOUND_WITHOUT_HASH_CHECK;
+                    if (!expectedHash.empty())
+                    {
+                        if (fileHashes.find(filePath) == fileHashes.end())
```
JohnMcPMS, line 295: If we are going to have a map for performance, then you should:
```
auto itr = find();
if (itr != end)
{
itr = emplace().first;
}

AreEqual(expected, itr->second);
```
To prevent any more map lookups than necessary.

JohnMcPMS (approved):
Ensure that consumers are going to be happy with the API design before merging.

PR #2431, opened 2022-08-09
Update Labels from Hash Command

### LINKED ISSUE #938: Update output text for hash values

The Windows Package Manager manifest v.1.0 uses:
`InstallerSha256` and `SignatureSha256` as key names for the installer hash and MSIX signature hash respectively.

The winget.exe client should use the same labels in the command output.

Trenly:
  - Resolves #938
-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2431)

### DIFF
--- a/src/AppInstallerCLICore/Commands/HashCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/HashCommand.cpp
@@ -44,7 +44,7 @@ namespace AppInstaller::CLI
             auto inputFile = context.Args.GetArg(Execution::Args::Type::HashFile);
             std::ifstream inStream{ Utility::ConvertToUTF16(inputFile), std::ifstream::binary };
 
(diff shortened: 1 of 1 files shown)


lechacon (approved):
I'm assuming these are supposed to be the literal manifest field so that one can just copy-and-paste it into a yaml file. I think it would be good to add a comment about that so that we don't mistakenly localize the strings later.

Trenly:
> I think it would be good to add a comment about that so that we don't mistakenly localize the strings later.

I thought that's what the Localization Independent View was for?
image

lechacon:
`_liv` does mean that it is supposed to be loc independent, but it doesn't say why. My first thought when I saw the change was to wonder why it wasn't localized like in the `show` command, if it was a leftover from when we didn't have localization, and if we needed to update.

The comment is not needed, I just prefer to err on the side of too many comments than too few.

JohnMcPMS (approved):

PR #2461, opened 2022-08-19
Increase PackageIdentifier segments to max 8

yao-msft:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2461)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,6 +11,7 @@ pr:
     include:
     - azure-pipelines.yml
     - src/*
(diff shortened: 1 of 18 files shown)


lechacon (approved):

JohnMcPMS:
Maybe we need to review the PR build rules and include a build when these schema files change.  While they don't affect the client in most cases, there are some and there are tests that will cover those.

JohnMcPMS (approved):

PR #2492, opened 2022-09-02
Add ExpectedReturnCode::ReturnResponse for PackageInUseByApplication

### LINKED ISSUE #2491: Alternative expected return code for `packageInUse`

So, today I tried to upgrade OBS Studio, and it failed with error code 6. After running the installer manually, the error code was actually because "Logi Tune" and "NVIDIA Broadcast" was somehow interfering with the installation, and the installer ask to close them.

So I created:

- https://github.com/microsoft/winget-pkgs/pull/77794

Which is better than nothing.

However, `packageInUse` gives the following description on CLI:

> Application is currently running. Exit the application then try again.

Which is not quite the case (i.e. OBS Studio was not open, so I could not close it).

I think we could introduce another variant: `packageInUseByAnotherApplication` that would translate to a more helpful message, like:

> Application is currently in use by another application.

image

_No response_

Trenly:
  * Resolves #2491
-----

Added to Schema 1.4 only, as to not be confused whether or not the 1.3 client supports this return response.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2492)


denelon:
We will also need to have the same feature supported in the REST source. 

CC: @zachcarp

yao-msft (approved):

PR #2500, opened 2022-09-06
Add support for installing portables from a zip 

### LINKED ISSUE #140: Support installing .zip files

ryfu-msft:
Resolves #140 

This PR adds functionality for supporting the installation of portable(s) from a zip archive.


src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -0,0 +1,496 @@
...
+using namespace AppInstaller::Repository::Microsoft;
+using namespace AppInstaller::Repository::Microsoft::Schema;
+
+namespace AppInstaller::CLI::Portable
+{
+    namespace
+    {
+        constexpr std::wstring_view c_PortableIndexFileName = L"portable.db";
+
+        bool VerifyPortableExeHash(const std::filesystem::path& targetPath, const std::string& hashValue)
```
Trenly, line 29: It seems like there should already be a VerifyHash function somewhere for validating other downloaded installers. Is this duplicate code? If not, perhaps this should be moved to Common as `ValidateFileHash` or something of that nature, since I doubt it applies just to portable EXEs

src/AppInstallerCLICore/PortableInstaller.h
```diff
@@ -0,0 +1,113 @@
...
+            {
+                return UninstallFromIndex(purge);
+            }
+            else
+            {
+                return UninstallSingle(purge);
+            }
+        }
+
+        template<typename T>
```
Trenly, line 57: This name isn’t very descriptive to me. What is being committed and where. Perhaps adding something tying it to ARP Values would help

src/AppInstallerRepositoryCore/Microsoft/PortableIndex.cpp
```diff
@@ -26,6 +26,33 @@ namespace AppInstaller::Repository::Microsoft
...
+        else if (std::filesystem::is_symlink(path))
+        {
+            portableFile.FileType = Schema::IPortableIndex::PortableFileType::Symlink;
+            portableFile.SymlinkTarget = std::filesystem::read_symlink(path).u8string();
+        }
+        else
+        {
+            portableFile.FileType = Schema::IPortableIndex::PortableFileType::File;
+
+            std::ifstream inStream{ path, std::ifstream::binary };
```
Trenly, line 47: This seems to be the same code for getti g the File hash as the VerifyPortableExe. Perhaps there should be a SHA256::GetFileHash method added to Utility that returns the hash given a path

src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -0,0 +1,489 @@
...
+        CommitToARPEntry(PortableValueName::WinGetSourceIdentifier, WinGetSourceIdentifier);
+        CommitToARPEntry(PortableValueName::UninstallString, "winget uninstall --product-code " + GetProductCode());
+        CommitToARPEntry(PortableValueName::WinGetInstallerType, InstallerTypeToString(Manifest::InstallerTypeEnum::Portable));
+        CommitToARPEntry(PortableValueName::SHA256, SHA256);
+        CommitToARPEntry(PortableValueName::InstallLocation, InstallLocation);
+    }
+
+    void PortableInstaller::FinalizeRegistryEntry()
+    {
+        CommitToARPEntry(PortableValueName::DisplayName, DisplayName);
```
JohnMcPMS, line 442: Why are these split from the "initialize" set of values?

src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -0,0 +1,489 @@
...
+                return false;
+            }
+            else
+            {
+                return true;
+            }
+        }
+    }
+
+    HRESULT PortableInstaller::InstallSingle(const std::filesystem::path& installerPath)
```
JohnMcPMS, line 141: I don't think splitting single/multiple is necessarily a good idea from a maintainability standpoint.  It isn't really necessary either given what I'm going to suggest in this rather long comment.

Any action (install/upgrade/uninstall) taken on the portable file(s) can be described by 3 states: { Desired, Expected, Actual }
- Desired contains the incoming file(s)
- Expected contains the description of the file(s) that we maintain (registry or index)
- Actual is the state of the filesystem

Every filesystem entry that we want can be described in the desired state, and each one has a related state in the expected and actual sets (even if it isn't in the set in memory).  This includes the root install directory for the package, which should not be treated as a special case.  Each filesystem entry can also be acted on individually, so long as they are properly ordered.

If you use this model, there is no install/upgrade/uninstall distinction in the core code.  Those APIs can still be presented, but they are used to create the desired state, which is then passed to the resolution engine.  That code can be clear and consistent because it needs only handle one filesystem entry at a time.  Each entry can be properly recorded before it committed to disk, or properly removed before removing the recorded entry.

Additionally, creating an abstraction over the single/multiple file state storage seems like a better way to keep the core code clean and in one place.  The desired state can easily be inspected to determine if it meets the criteria for storing in the registry (contains exactly one directory, file, and link) or not.  The correct interface implementation can be created and not need to have multiple code paths.

src/AppInstallerCLIE2ETests/TestData/Manifests/TestZipInstaller_Portable.2.0.0.0.yaml
```diff
@@ -0,0 +1,18 @@
...
+License: Test
+ShortDescription: E2E test for installing a zip with portable.
+Installers:
+  - Architecture: x64
+    InstallerUrl: https://localhost:5001/TestKit/AppInstallerTestZipInstaller/AppInstallerTestZipInstaller.zip
+    InstallerType: zip
+    InstallerSha256: <ZIPHASH>
+    NestedInstallerType: portable
+    NestedInstallerFiles:
+      - RelativeFilePath: AppInstallerTestExeInstaller.exe
```
Trenly, line 15: There should be some tests which consider two entries with the same `RelativeFilePath`, Two entries with the same `PortableCommandAlias`, and possibly Exact duplicate entries?

I was doing some testing and I don't think that these cases are being validated against at all. (#2523) 
Perhaps @denelon can provide his thoughts on if these validations would be required

Trenly:
Is the output of the files removed intentional?
```
PS D:\Git\winget-pkgs> wingetdev install -m D:\Git\winget-pkgs\manifests\c\Codex\ffmpeg\essentials\5.1.1
Downloading https://www.gyan.dev/ffmpeg/builds/packages/ffmpeg-5.1.1-essentials_build.zip
  ██████████████████████████████  78.9 MB / 78.9 MB
Starting package install...
Successfully installed
PS D:\Git\winget-pkgs> winget list ffmpeg
Name              Id                                     Version
ffmpeg essentials Codex.ffmpeg.essentials__DefaultSource 5.1.1
PS D:\Git\winget-pkgs> wingetdev uninstall ffmpeg  
No installed package found matching input criteria.
PS D:\Git\winget-pkgs> wingetdev uninstall 'ffmpeg essentials'
Found ffmpeg essentials [Codex.ffmpeg.essentials__DefaultSource]
Starting package uninstall...
"C:\\Users\\Trenly\\AppData\\Local\\Microsoft\\WinGet\\Packages\\Codex.ffmpeg.essentials__DefaultSource\\ffmpeg-5.1.1-essentials_build"
"C:\\Users\\Trenly\\AppData\\Local\\Microsoft\\WinGet\\Links\\ffmpeg.exe"
"C:\\Users\\Trenly\\AppData\\Local\\Microsoft\\WinGet\\Links\\ffplay.exe"
"C:\\Users\\Trenly\\AppData\\Local\\Microsoft\\WinGet\\Links\\ffprobe.exe"
Successfully uninstalled
```

vedantmgoyal2009:
> Is the output of the files removed intentional?

I don't know if it is there for debugging purposes, but it should write them to log too.

[WinGet-2022-09-17-21-50-32.732.log](https://github.com/microsoft/winget-cli/files/9592130/WinGet-2022-09-17-21-50-32.732.log)

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -4,27 +4,33 @@
...
     {
         const auto& installerPath = context.Get<Execution::Data::InstallerPath>();
-        const auto& installerParentPath = installerPath.parent_path();
+        std::filesystem::path destinationFolder = installerPath.parent_path() / s_Extracted;
+        std::filesystem::create_directory(destinationFolder);
 
-        // TODO: For portables, extract portables to final install location and log to local database.
-        HRESULT hr = AppInstaller::Archive::TryExtractArchive(installerPath, installerParentPath);
-        AICLI_LOG(CLI, Info, << "Extracting archive to: " << installerParentPath);
+        AICLI_LOG(CLI, Info, << "Extracting archive to: " << destinationFolder);
```
Trenly, line 24: In testing this, it feels like perhaps there should be output written to inform the user the archive is being extracted. I noticed with particularly large archives there appeared to be significant delay between `Successfully verified installer hash` and `Starting package install. . .`

JohnMcPMS: Agreed; the extraction should be part of the install phase, not the download.

JohnMcPMS: We may also be able to have progress callbacks for the extraction, and certainly can for the progress of moving files to their target location.

src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -0,0 +1,460 @@
...
+            std::filesystem::remove(filePath);
+        }
+        else if (fileType == PortableFileType::Directory && std::filesystem::exists(filePath))
+        {
+            AICLI_LOG(CLI, Info, << "Removing directory at " << filePath);
+            std::filesystem::remove_all(filePath);
+        }
+    }
+
+    void PortableInstaller::ApplyDesiredState()
```
JohnMcPMS, line 163: So while this has the structure of what I was pushing for (a single function that applies the state), this is a very inefficient implementation.  I was hoping that we would be able to apply the differences rather than removing everything and putting all the files back.  That isn't strictly necessary; we could have a follow-up PR at some point to handle that.  It is possible that we will find it necessary to do before shipping for real depending on the performance/size of the archives that people are using.

Additionally, while it also isn't necessary to do it this way, it would be cleaner if the index/no index was decided once by creating an object that implemented the necessary interface.  That way you just throw values at it and it decides where to put them, rather than having if blocks sprinkled throughout.

src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -0,0 +1,460 @@
...
+        DisplayVersion = entry.DisplayVersion;
+        Publisher = entry.Publisher;
+        InstallDate = Utility::GetCurrentDateForARP();
+        URLInfoAbout = manifest.CurrentLocalization.Get<Manifest::Localization::PackageUrl>();
+        HelpLink = manifest.CurrentLocalization.Get<Manifest::Localization::PublisherSupportUrl>();
+    }
+
+    void PortableInstaller::SetExpectedState()
+    {
+        const auto& indexPath = InstallLocation / GetPortableIndexFileName();
```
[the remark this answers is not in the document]
ryfu-msft: Yes, GetPortableIndexFileName returns the package specific filename which is now the productCode + ".db" extension

src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -0,0 +1,460 @@
...
+        AddToPathVariable();
+    }
+
+    void PortableInstaller::Uninstall()
+    {
+        ApplyDesiredState();
+
+        RemoveInstallDirectory();
+
+        RemoveFromPathVariable();
```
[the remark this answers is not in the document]
ryfu-msft: The path should only be removed if the directory added to path is empty. That can change depending on whether the user wants to purge the install directory as checked in `RemoveInstallDirectory` or if there are leftover files after applying the desired state. Even though the flow is not fully in the opposite order, I think this flow ensures that we don't accidentally remove from path if other portables depend on it.

JohnMcPMS (approved):

SeanFeldman:
@ryfu-msft, is there documentation on how to prepare WinGet zip-based manifest? Thank you.

Trenly:
> @ryfu-msft, is there documentation on how to prepare WinGet zip-based manifest? Thank you.

I have a few examples on my fork - https://github.com/Trenly/winget-pkgs/branches/all?query=zips

[YamlCreate.ps1] also supports Zip, but you have to enable developer options and override the manifest version to 1.4.0 using the YamlCreate settings

denelon:
We're going to update wingetcreate as well so it can also create manifests for .zip based packages.

AdrianoCahete:
> We're going to update wingetcreate as well so it can also create manifests for .zip based packages.

It's working  @denelon ? Where I can find the issue so I can keep tracking it?

mdanish-kh:
@AdrianoCahete The latest version of winget-create is able to create .zip based manifests. You can download it from here:

https://github.com/microsoft/winget-create/releases/

Issue #2355, opened 2022-07-21
Open log files or provide path as output


KevinLaMS:
### Description of the new feature / enhancement

As you know the Log path is a less than intuitive location.  ;) 
%LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir

I had to create a CMD file in order to easily get at the LOG file when opening issues.

I would like a way to quickly and easily discover the LOG file I just created.  So that I can attach in bugs and in email.  Yeah, yeah, I know I can do it in my CMD file using Batch, but I cannot be alone in this desire.  

### Proposed technical implementation details

I propose 3 things that would make my life easier.
1) if you pass in the --verbose-logs option, the output includes path to the file:
**Example:**
```
winget list zoom --verbose-logs
Name Id      Version
--------------------------
Zoom ZoomUMX 5.10.4 (5035)

Log: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir\<logfile>
```
We can debate if the path should be before or after the results.  If we do it before, the benefit is any crazy script looking at the output may still succeed.  

2) Same as above, but you need to ask for the path
**Example:**
```
winget list zoom --verbose-logs --logfile-path
Name Id      Version
--------------------------
Zoom ZoomUMX 5.10.4 (5035)

Log: %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir\<logfile>
```
We can debate if the path should be before or after the results.  If we do it before, the benefit is any crazy script looking at the output may still succeed.  

3) We provide an --open option.
**Example:**
```
winget list zoom --verbose-logs  --open
```
In this case the command is executed and afterwards, the log file is launched in the default windows editor.

I am leaning towards option 1.  What do you think?

jedieaston:
I have long thought a `winget logs` command that just opened the folder in Explorer would be helpful. If we just print the log file path then you'd have to copy paste it into a explorer window, which is an extra step. 

(So I like option 3 the best, but I'd rather it be a command than an option so I can use it regardless of whether I remembered to put the argument on the command I'm looking for logs for.)

Trenly:
[...]

PR #2510, opened 2022-09-12
Add experimental argument to open logs folder

Trenly:
  - https://github.com/microsoft/winget-cli/issues/2355#issuecomment-1191937273
-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2510)


denelon:
Does this argument work with any command? Does it display the help across all commands? Maybe a quick spec would make sense for this one. I'm expecting we need to think about how it would interact with the PowerShell cmdlets as well.

Trenly:
> Does this argument work with any command? 

Yes

> Does it display the help across all commands? 

Yes

> Maybe a quick spec would make sense for this one. I'm expecting we need to think about how it would interact with the PowerShell cmdlets as well.

I think it would be a good consideration, but I would suggest instead speccing out all common arguments; How does verbose logs interact with the cmdlets? How does wait interact? 

I also think this could be refined more in the future (hence only mentioning the linked issue, not marking it for closure) as how users interact with logs is a complex subject. I was thinking build in some minimal functionality first, then spec out a larger feature around it?

denelon:
Maybe it should be an experimental argument initially then. It allows us to get feedback before releasing it with the GA client if we need to iterate.

denelon:
Yes, I agree we should cover the other types of arguments as well as they relate to PowerShell.

Trenly:
> Maybe it should be an experimental argument initially then. It allows us to get feedback before releasing it with the GA client if we need to iterate.

Made it experimental

yao-msft (approved):

PR #2518, opened 2022-09-15
Add Aliases to allow for quick upgrade of all and unknown

Trenly:
  - Nope, just felt right

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2518)

### DIFF
--- a/src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
@@ -83,27 +83,27 @@ namespace AppInstaller::CLI
     std::vector<Argument> UpgradeCommand::GetArguments() const
     {
         return {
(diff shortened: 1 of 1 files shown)


denelon:
Is this going to cause the need to update Microsoft Docs when we GA 1.4?

Trenly:
> Is this going to cause the need to update Microsoft Docs when we GA 1.4?

Depends. If every single piece of information needs to be there, then yes. Practically, the information on Microsoft docs would still be accurate, it just wouldn't show these two aliases. Since the docs contain the full arguments anyways, it could be argued that the aliases are for power users, and since they would be shown with the help text anyways, don't need to be added to docs.

yao-msft (approved):

PR #2542, opened 2022-09-22
Added WinGetUtil E2E tests

AmelBawa-msft:
- Added E2E tests for WinGetUtil project library:
  - CompareVersions
  - Download
  - Log
  - Manifest
  - SQLiteIndex
- Added reference from E2E test project to WinGetUtil project
  - Copy `dll` to output directory

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2542)


src/IndexCreationTool/Program.cs
```diff
@@ -19,12 +21,19 @@ static void Main(string[] args)
             string appxManifestPath = string.Empty;
             string certPath = string.Empty;
 
+            // List of directories to include. By default, include all directories.
+            List<string> includeDirList = new() { "" };
```
yao-msft, line 25: >"" [](http://example.com/codeflow?start=50&length=2)

nit: String.Empty to be more c# coding style?

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilWrapper.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <summary>
+    /// Wrapper class for WinGetUtil exports.
+    /// For more details about methods in this class visit WinGetUtil.h file.
+    /// </summary>
+    internal class WinGetUtilWrapper
+    {
+        private const string DllName = @"WinGetUtil.dll";
+
+        [Flags]
+        public enum CreateManifestOption
```
yao-msft, line 18: >CreateManifestOption [](http://example.com/codeflow?start=20&length=20)

I think at least 1 failure case or SchemaAndSemanticValidation should be tested to trigger the manifest validation code path.

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilManifest.cs
```diff
@@ -0,0 +1,70 @@
...
+            uint minorVersion = 2;
+            WinGetUtilWrapper.WinGetSQLiteIndexCreate(sqliteFile, majorVersion, minorVersion, out this.indexHandle); ;
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            WinGetUtilWrapper.WinGetSQLiteIndexClose(this.indexHandle);
+        }
+
```
yao-msft, line 28: could be removed

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilWrapper.cs
yao-msft: Also, even if we are not using them yet, we should have the complete copy of the enum values for completeness

yao-msft:
The InstallerMetadataCollection tests are missing, are they to be sent in a separate pr?

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilSQLiteIndex.cs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        [Test]
+        public void WinGetUtil_SQLiteIndex_AddManifest()
+        {
+            SQLiteIndex((indexHandle) =>
+            {
+                // Add manifest
+                IntPtr hresult = WinGetUtilWrapper.WinGetSQLiteIndexAddManifest(indexHandle, addManifestsFile_1, relativePath_1);
+                Assert.AreEqual(IntPtr.Zero, hresult);
```
yao-msft, line 35: >Assert.AreEqual(IntPtr.Zero, hresult); [](http://example.com/codeflow?start=16&length=38)

hresult should be compared to ErrorCode.S_OK(defined in Constants.cs) across all tests (in other files too)

Intptr valus are specifically for native handles

yao-msft: I just saw all dllimport functions has preserveSig = false, so we probably don't have to do these checks since if the call failed, they are automatically converted to exceptions.

And the return type in dllimport could be void. I saw definitions in IndexCreationTool can also be changed (not your fault)

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilManifest.cs
```diff
@@ -14,57 +14,68 @@ public class WinGetUtilManifest
...
         public void SetUp()
         {
+            this.indexHandle = IntPtr.Zero;
             var sqliteFile = TestCommon.GetRandomTestFile(".db");
             uint majorVersion = 1;
             uint minorVersion = 2;
             WinGetUtilWrapper.WinGetSQLiteIndexCreate(sqliteFile, majorVersion, minorVersion, out this.indexHandle); ;
         }
 
-        [TearDown]
```
yao-msft: My bad, I thought the index handle is never used. Please add back the TearDown to free the handle

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilManifest.cs
```diff
@@ -14,57 +14,68 @@ public class WinGetUtilManifest
...
-        public void TearDown()
+        [Test]
+        public void WinGetUtil_ValidateManifest_Fail_SchemaAndSemanticValidation()
         {
-            WinGetUtilWrapper.WinGetSQLiteIndexClose(this.indexHandle);
+            string manifestsDir = TestCommon.GetTestDataFile(@"WinGetUtil\Manifests\Unmerged\ValidateManifest");
```
yao-msft, line 27: I tried and saw the manifest is failing because ShortDescription is missing(error) and a warning message about missing silent switch for exe installer type(which you can change to nullsoft to avoid for example), can we add that to the test manifest and make it a passing case?

yao-msft (approved):

PR #2577, opened 2022-10-05
Enable mechanism for some control over correlation

JohnMcPMS:
## Change
The primary change is to enable some control over the correlation functionality so that different situations can behave differently.  Specifically, the utility functionality of metadata collection, which is intended to be run in a "clean room" environment, can assume that a single change to ARP represents the target entry.

In addition, this change enables diagnostic information to be output by the metadata collection.  This includes the reason that an entry was chosen, and the confidence value for various possible entries if the heuristic matching is attempted.

Finally, the correlation testbed is enhanced to run the metadata collection, allowing it to be tested alongside the inline correlation.  The scripts have various improvements, such as a `-Wait` to leave the sandbox active and forcing the CSV to be output with a UTF8-BOM (Excel requires the BOM in order to open the CSV as UTF8).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2577)

### DIFF
--- a/src/AppInstallerRepositoryCore/Public/winget/ARPCorrelation.h
+++ b/src/AppInstallerRepositoryCore/Public/winget/ARPCorrelation.h
@@ -42,16 +42,65 @@ namespace AppInstaller::Repository::Correlation
         bool IsNewOrUpdated;
     };
 
-    struct ARPCorrelationResult
+    // One of the possible options that could be chosen for correlation.
+    struct CorrelationMeasure
+    {
+        // The value that the correlation algorithm assigned to the match with the package.
+        double Measure{};
+
+        // The package that was measured.
+        std::shared_ptr<AppInstaller::Repository::IPackageVersion> Package{};
+    };
+
+    // The result of a heuristics correlation attempt.
+    struct ARPHeuristicsCorrelationResult
     {
         // Correlated package from ARP
         std::shared_ptr<AppInstaller::Repository::IPackageVersion> Package{};
+
+        // The reason for the correlation (for diagnostics).
+        std::string Reason;
+
+        // The correlation metrics and their associated ARP package information (for diagnostics).
+        std::vector<CorrelationMeasure> Measures;
+    };
+
+    // The result of a correlation attempt.
+    struct ARPCorrelationResult : public ARPHeuristicsCorrelationResult
+    {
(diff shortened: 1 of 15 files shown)


tools/CorrelationTestbed/InSandboxScript.ps1
```diff
@@ -81,6 +83,12 @@ if ($UseDev)
   $installAndCorrelationExpression = -join($installAndCorrelationExpression, ' -dev')
 }
 
+if ($MetadataCollection)
+{
+  $wingetUtilPath = Join-Path $PSScriptRoot "WinGetUtil.dll"
+  $installAndCorrelationExpression = -join($installAndCorrelationExpression, ' -meta "', $wingetUtilPath, '" -sys32 "', $System32Path,'"')
```
yao-msft, line 89: >' [](http://example.com/codeflow?start=134&length=1)

nit: space

src/AppInstallerRepositoryCore/Public/winget/ARPCorrelation.h
```diff
@@ -42,16 +42,65 @@ namespace AppInstaller::Repository::Correlation
...
         // Number of ARP entries that are new or updated
         size_t ChangesToARP{};
+
         // Number of ARP entries that match with the installed package
         size_t MatchesInARP{};
+
         // Number of changed ARP entries that match the installed package
         size_t CountOfIntersectionOfChangesAndMatches{};
+
+        ARPCorrelationResult& operator=(ARPHeuristicsCorrelationResult&& other)
```
yao-msft, line 80: >ARPCorrelationResult& operator=(ARPHeuristicsCorrelationResult&& other) [](http://example.com/codeflow?start=8&length=71)

will this be misused?

yao-msft (approved):
:shipit:

src/AppInstallerRepositoryCore/Public/winget/ARPCorrelation.h
JohnMcPMS: Honestly, I was surprised it didn't work out of the box.  The goal is to be able to assign just the base portion of the object, which seems very intuitive to me.

PR #2581, opened 2022-10-09
Update NuGet dependencies

AtariDreams:
- [X] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
-----
Some of these dependencies are years out of date

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2581)


src/AppInstallerTestExeInstaller/main.cpp
```diff
@@ -39,8 +39,8 @@ path GenerateUninstaller(std::wostream& out, const path& installDirectory, const
 
     std::wofstream uninstallerScript(uninstallerPath);
     uninstallerScript << "@echo off\n";
-    uninstallerScript << "ECHO. >" << uninstallerOutputTextFilePath << "\n";
-    uninstallerScript << "ECHO AppInstallerTestExeInstaller.exe uninstalled successfully.\n";
+    uninstallerScript << "ECHO. >" << uninstallerOutputTextFilePath;
```
yao-msft, line 42: nit: just curious is there a specific reason to move the \n to next line? For code style consistency within this file, we put \n to the line end.

src/AppInstallerCLI/AppInstallerCLI.vcxproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\packages\Microsoft.Windows.CppWinRT.2.0.220929.3\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('..\packages\Microsoft.Windows.CppWinRT.2.0.220929.3\build\native\Microsoft.Windows.CppWinRT.props')" />
```
yao-msft, line 3: This should be replacing the line below instead of adding a new line and also this should use $(SolutionDir). This applies to other places across this pr.

yao-msft:
For updating all the package dependencies, we may need some internal reviews to determine which versions to use. For example, we may prefer versions with long term support instead of the latest. For Target Device Family, we choose 22000 to align with other  tools winget shipped with, etc.

Trenly:
Thank you for the contribution @AtariDreams;

This pull request has unfortunately become quite outdated and has several merge conflicts. I'm confident that the team will continue evaluating the NuGet dependencies as new versions become available or moved into Long Term Support. As @yao-msft mentioned, there are several considerations around which version to use. If you have a specific package you believe needs to be updated outside of the normal cadence, I would encourage you to open an issue so that the consideration by the team can be tracked and documented.

-----
##### Close with reason: PR is outdated by more than a year;

PR #2591, opened 2022-10-12
Remove MSIX installers after validation is done

AmelBawa-msft:
Remove MSIX installers _after validation is done_ in order to ensure that the information used during validation is available when needed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2591)

### DIFF
--- a/src/AppInstallerCommonCore/Manifest/MsixManifestValidation.cpp
+++ b/src/AppInstallerCommonCore/Manifest/MsixManifestValidation.cpp
@@ -36,6 +36,23 @@ namespace AppInstaller::Manifest
         return errors;
     }
 
+    void MsixManifestValidation::Cleanup()
+    {
+        // Remove all downloaded files (if any)
+        AICLI_LOG(Core, Info, << "Removing downloaded installers");
+        for (const auto& installerPath : m_downloadedInstallers)
+        {
+            try
+            {
+                std::filesystem::remove(installerPath);
+            }
+            catch (...)
+            {
+                AICLI_LOG(Core, Warning, << "Failed to remove downloaded installer");
+            }
+        }
+    }
+
     std::optional<std::filesystem::path> MsixManifestValidation::DownloadInstaller(std::string installerUrl, int retryCount)
     {
         while (retryCount-- > 0)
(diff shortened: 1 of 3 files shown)


src/AppInstallerCommonCore/Manifest/MsixManifestValidation.cpp
```diff
@@ -36,6 +36,27 @@ namespace AppInstaller::Manifest
         return errors;
     }
 
+    void MsixManifestValidation::Cleanup()
+    {
+        // Clean cache
+        AICLI_LOG(Core, Info, << "Clearing Msix info cache");
+        m_msixInfoCache.clear();
```
yao-msft, line 43: >m_msixInfoCache.clear(); [](http://example.com/codeflow?start=8&length=24)

This is probably not needed, they are cleared during destruction.

And shall we make this cleanup() method just the destructor, I don't see it being reused. This way caller don't have to remember to call cleanup

src/AppInstallerCommonCore/Manifest/MsixManifestValidation.cpp
```diff
@@ -36,6 +36,27 @@ namespace AppInstaller::Manifest
...
+        AICLI_LOG(Core, Info, << "Removing downloaded installers");
+        for (const auto& installerPath : m_downloadedInstallers)
+        {
+            try
+            {
+                std::filesystem::remove(installerPath);
+            }
+            catch (...)
+            {
+                AICLI_LOG(Core, Warning, << "Failed to remove downloaded installer");
```
yao-msft, line 55: nit: log the failed path too since we have the info

src/AppInstallerCommonCore/Manifest/MsixManifestValidation.cpp
```diff
@@ -36,6 +36,22 @@ namespace AppInstaller::Manifest
...
+        AICLI_LOG(Core, Info, << "Removing downloaded installers");
+        for (const auto& installerPath : m_downloadedInstallers)
+        {
+            try
+            {
+                std::filesystem::remove(installerPath);
+            }
+            catch (...)
+            {
+                AICLI_LOG(Core, Warning, << "Failed to remove downloaded installer: " << installerPath.u8string());
```
yao-msft, line 50:
```suggestion
                AICLI_LOG(Core, Warning, << "Failed to remove downloaded installer: " << installerPath);
```

We have overload for std::filesystem::path already

PR #2610, opened 2022-10-20
Fix internal build

yao-msft:
Fix internal build by renaming package.json. Internal build has strict feed policy while winget directly subtree the purelib and built itself. The package.json is not used in winget build.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2610)


JohnMcPMS (approved):
I would just delete the file, but this works.

PR #2617, opened 2022-10-21
Add AdvancedInstaller (exe) type and Default Switches

Trenly:
  - No. I noticed that several (~100+) packages use exe's built with Advanced Installer and that they all had the same switches. I felt that having a dedicated installer type for this would save contributors from having to add installer switches and expected return codes for this type of executable. Additionally, it would show good faith and reciprocity towards Advanced Installer as they mention winget on their page about [Finding Silent Install Switches](https://www.advancedinstaller.com/find-exe-silent-install-switches.html)

* This PR does not consider all changes required for REST support
* This PR does not consider any learn.microsoft.com documentation that would need to be added. For example - https://www.advancedinstaller.com/silent-install-exe-msi-applications.html being linked in [Installer Switches](https://www.advancedinstaller.com/silent-install-exe-msi-applications.html)
* This PR does not consider wingetcreate support for the installer type
-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2617)

### DIFF
--- a/schemas/JSON/manifests/v1.4.0/manifest.installer.1.4.0.json
+++ b/schemas/JSON/manifests/v1.4.0/manifest.installer.1.4.0.json
@@ -65,7 +65,8 @@
         "wix",
         "burn",
         "pwa",
-        "portable"
+        "portable",
+        "advancedinstaller"
       ],
       "description": "Enumeration of supported installer types. InstallerType is required in either root level or individual Installer level"
     },
@@ -80,7 +81,8 @@
         "nullsoft",
         "wix",
         "burn",
-        "portable"
+        "portable",
+        "advancedinstaller"
       ],
       "description": "Enumeration of supported nested installer types contained inside an archive file"
     },
--- a/schemas/JSON/manifests/v1.4.0/manifest.installer.1.4.0.json
+++ b/schemas/JSON/manifests/v1.4.0/manifest.installer.1.4.0.json
@@ -66,7 +66,7 @@
         "burn",
         "pwa",
         "portable",
-        "advancedinstaller"
+        "advancedInstaller"
       ],
       "description": "Enumeration of supported installer types. InstallerType is required in either root level or individual Installer level"
     },
@@ -82,7 +82,7 @@
         "wix",
         "burn",
         "portable",
-        "advancedinstaller"
+        "advancedInstaller"
       ],
       "description": "Enumeration of supported nested installer types contained inside an archive file"
     },
--- a/src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
+++ b/src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
@@ -608,7 +608,6 @@ namespace AppInstaller::Manifest
         switch (installerType)
         {
         case InstallerTypeEnum::Burn:
-        case InstallerTypeEnum::AdvancedInstaller:
         case InstallerTypeEnum::Wix:
         case InstallerTypeEnum::Msi:
             // See https://docs.microsoft.com/windows/win32/msi/error-codes
@@ -620,6 +619,9 @@ namespace AppInstaller::Manifest
                 { ERROR_SUCCESS_REBOOT_REQUIRED, ExpectedReturnCodeEnum::RebootRequiredToFinish },
                 { ERROR_SUCCESS_REBOOT_INITIATED, ExpectedReturnCodeEnum::RebootInitiated },
                 { ERROR_INSTALL_USEREXIT, ExpectedReturnCodeEnum::CancelledByUser },
+                { ERROR_BAD_CONFIGURATION, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_INSTALL_UI_FAILURE, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_CREATE_FAILED, ExpectedReturnCodeEnum::ContactSupport },
                 { ERROR_PRODUCT_VERSION, ExpectedReturnCodeEnum::AlreadyInstalled },
                 { ERROR_INSTALL_REJECTED, ExpectedReturnCodeEnum::SystemNotSupported },
                 { ERROR_INSTALL_PACKAGE_REJECTED, ExpectedReturnCodeEnum::BlockedByPolicy },
@@ -633,6 +635,24 @@ namespace AppInstaller::Manifest
                 { ERROR_INVALID_PATCH_XML, ExpectedReturnCodeEnum::InvalidParameter },
                 { ERROR_INSTALL_LANGUAGE_UNSUPPORTED, ExpectedReturnCodeEnum::SystemNotSupported },
                 { ERROR_INSTALL_PLATFORM_UNSUPPORTED, ExpectedReturnCodeEnum::SystemNotSupported },
+            };
+        case InstallerTypeEnum::AdvancedInstaller:
+            // See https://www.advancedinstaller.com/user-guide/exe-setup-file.html
+            return
+            {
+                { ERROR_INSTALL_ALREADY_RUNNING, ExpectedReturnCodeEnum::InstallInProgress },
+                { ERROR_DISK_FULL, ExpectedReturnCodeEnum::DiskFull },
+                { ERROR_INSTALL_SERVICE_FAILURE, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_SUCCESS_REBOOT_REQUIRED, ExpectedReturnCodeEnum::RebootRequiredToFinish },
+                { ERROR_SUCCESS_REBOOT_INITIATED, ExpectedReturnCodeEnum::RebootInitiated },
+                { ERROR_INSTALL_USEREXIT, ExpectedReturnCodeEnum::CancelledByUser },
+                { ERROR_BAD_CONFIGURATION, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_INSTALL_UI_FAILURE, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_CREATE_FAILED, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_PRODUCT_VERSION, ExpectedReturnCodeEnum::AlreadyInstalled },
+                { ERROR_INSTALL_REJECTED, ExpectedReturnCodeEnum::BlockedByPolicy },
+                { -1, ExpectedReturnCodeEnum::CancelledByUser },
+                // 1 when EXE bootstrapper is launched with wrong value for /aespassword parameter
             };
         case InstallerTypeEnum::Inno:
             // See https://jrsoftware.org/ishelp/index.php?topic=setupexitcodes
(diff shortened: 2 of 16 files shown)


vedantmgoyal2009:
* There's an issue: https://github.com/microsoft/winget-pkgs/issues/58886

yao-msft:
@denelon, do we want to take this in as part of v1.4 or we'll include it in v1.5? Seems like it'd be better to do some e2e validations before we take in a new InstallerType support.

schemas/JSON/manifests/v1.4.0/manifest.installer.1.4.0.json
```diff
@@ -65,7 +65,8 @@
         "wix",
         "burn",
         "pwa",
-        "portable"
+        "portable",
+        "advancedinstaller"
```
yao-msft, line 69: nit: enum values use camelCase, if advanced installer is treated as 2 words, this should be "advancedInstaller"

yao-msft:
For adding support of a new type, it'd be better to add the basic install, upgrade, uninstall workflow tests in Workflow.cpp in unit tests to exercise the whole flow, and 1 basic install, uninstall tests in e2e tests.

Trenly:
> For adding support of a new type, it'd be better to add the basic install, upgrade, uninstall workflow tests in Workflow.cpp in unit tests to exercise the whole flow, and 1 basic install, uninstall tests in e2e tests.

I completely agree; However - I have no clue what I'm doing when it comes to writing the tests. I took what I saw from other unit tests and think I kind of understood them, but I really have no idea. Any guidance that you can provide here would be appreciated.

Regarding the e2e tests - I think I understand the code portion of it. I see where the manifests are written, I see how the test cases are running the install/uninstall,  What I'm not exactly sure on is how to generate / build / include the actual exe file for testing. I presume this would have to be included as either a) A separate project with a build output, or b) Taking the Taking the AppInstallerTestExeInstaller.exe and re-packaging it with AdvancedInstaller, then placing it in the TestData folder. Again, this is something I don't really know how to do myself, but if there's someone who can guide me through it, I'd be happy to learn.

yao-msft:
Sorry I should provide more details in my previous comments.

Regarding unit tests, they should be added to WorkFlow.cpp under AppInstallerCliTests peoject, the project builds as an exe, you can just run the exe to run the tests. For Workflow tests specifically, we create a fake source with predefined query like [here](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/AppInstallerCLITests/WorkFlow.cpp#L144) , then we create the command directly and override some workflows as necessary like [here](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/AppInstallerCLITests/WorkFlow.cpp#L1623), finally we verify the installer is successfully invoked with expected args. You can maybe write something similar to [this test](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/AppInstallerCLITests/WorkFlow.cpp#L1616)
Same for upgrade and uninstall flow

Regarding e2e tests, it would be p2 as I believe most advancedInstaller should execute like an exe And e2e tests is quite complex to setup. But if there's time to add one , please just use the existing AppInstallerTestExeInstaller.exe, winget wants to test the installer is invoked as expected(not the installer itself is working as expected or not). If you add a new manifest to the Manifests folder where other test manifests sit, it should be picked up and included in a test source automatically.

By the way, I just saw Com api needs to be updated and hooked up accordingly [here](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/Microsoft.Management.Deployment/PackageManager.idl#L300) too.

Trenly:
> Regarding unit tests, they should be added to WorkFlow.cpp under AppInstallerCliTests peoject, the project builds as an exe, you can just run the exe to run the tests. For Workflow tests specifically, we create a fake source with predefined query like [here](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/AppInstallerCLITests/WorkFlow.cpp#L144) , then we create the command directly and override some workflows as necessary like [here](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/AppInstallerCLITests/WorkFlow.cpp#L1623), finally we verify the installer is successfully invoked with expected args. You can maybe write something similar to [this test](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/AppInstallerCLITests/WorkFlow.cpp#L1616) Same for upgrade and uninstall flow
>
Got it; I think my understanding was correct, I'll wait for the code review to see how badly I messed up.
 
> Regarding e2e tests, it would be p2 as I believe most advancedInstaller should execute like an exe And e2e tests is quite complex to setup. But if there's time to add one , please just use the existing AppInstallerTestExeInstaller.exe, winget wants to test the installer is invoked as expected(not the installer itself is working as expected or not). If you add a new manifest to the Manifests folder where other test manifests sit, it should be picked up and included in a test source automatically.
> 
Got it. I'll take a look and see what I can do

> By the way, I just saw Com api needs to be updated and hooked up accordingly [here](https://github.com/microsoft/winget-cli/blob/5cd56162ecd70d84f19c01f11320afb9a1995ea2/src/Microsoft.Management.Deployment/PackageManager.idl#L300) too.
>
Ah, good catch. Thank you. Will update

yao-msft:
Regarding 


> I think I got those as part of my initial commit?

That should be good I think, maybe I overlooked this one..

SpecterShell:
Advanced Installer returns the same exit code as MSI's except -1 and 1, according to <https://www.advancedinstaller.com/user-guide/exe-setup-file.html>.

Trenly:
> Advanced Installer returns the same exit code as MSI's except -1 and 1, according to https://www.advancedinstaller.com/user-guide/exe-setup-file.html.

I must have missed that; Will update

src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -592,8 +611,29 @@ namespace AppInstaller::Manifest
...
+                { ERROR_INSTALL_SERVICE_FAILURE, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_SUCCESS_REBOOT_REQUIRED, ExpectedReturnCodeEnum::RebootRequiredToFinish },
+                { ERROR_SUCCESS_REBOOT_INITIATED, ExpectedReturnCodeEnum::RebootInitiated },
+                { ERROR_INSTALL_USEREXIT, ExpectedReturnCodeEnum::CancelledByUser },
+                { ERROR_BAD_CONFIGURATION, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_INSTALL_UI_FAILURE, ExpectedReturnCodeEnum::ContactSupport },
+                { ERROR_CREATE_FAILED, ExpectedReturnCodeEnum::ContactSupport },
                 { ERROR_PRODUCT_VERSION, ExpectedReturnCodeEnum::AlreadyInstalled },
                 { ERROR_INSTALL_REJECTED, ExpectedReturnCodeEnum::BlockedByPolicy },
+                { 1, ExpectedReturnCodeEnum::CancelledByUser },
```
SpecterShell, line 635:
```suggestion
                { -1, ExpectedReturnCodeEnum::CancelledByUser },
```
And 1 is "launched with wrong value".

Trenly:
> Azure Pipelines successfully started running 1 pipeline(s).

I cri. For some reason it won't let me run the tests locally, seems like its related to my setup though. It's complaining the VS2019 build tools aren't installed when they are.

Would you be able to help me identify where the error is?

yao-msft:
> 
> Would you be able to help me identify where the error is?

The pipeline shows 10 tests failed, some seems not directly related to this change. Do you mind if I clone your branch and directly push the fix while testing?

Trenly:
> > Would you be able to help me identify where the error is?
> 
> The pipeline shows 10 tests failed, some seems not directly related to this change. Do you mind if I clone your branch and directly push the fix while testing?

I don't mind at all; If you have any suggestions on how I can improve too, I'm happy to learn

yao-msft:
> Edit: I'd like to rebase this on master first to ensure the additional expected return codes can be used

Sure, thanks

Trenly:
> > Edit: I'd like to rebase this on master first to ensure the additional expected return codes can be used
> 
> Sure, thanks

Rebased and pushed

Trenly:
> /azp run

Thank you, I think I see why it was erroring now with the commits you added

yao-msft:
Though the tests were fixed, I just realized AdvancedInstaller is an installer authoring tool that supports exe, msi, msix, appv, etc. We may need to reevaluate how we approach this installer type, if some advanced installer exe was actually just a bootstrapper for some msi(I may just be making things up based on my initial very basic exploration), the upgrade will be broken. Since we are going to have a release candidate for v1.4 soon, we'll have to revisit and take this change in v1.5 or future after we have time to think this through.

Trenly:
> Though the tests were fixed, I just realized AdvancedInstaller is an installer authoring tool that supports exe, msi, msix, appv, etc. We may need to reevaluate how we approach this installer type, if some advanced installer exe was actually just a bootstrapper for some msi(I may just be making things up based on my initial very basic exploration), the upgrade will be broken. Since we are going to have a release candidate for v1.4 soon, we'll have to revisit and take this change in v1.5 or future after we have time to think this through.

Fair enough; I'll convert to draft for now then

SpecterShell:
While some AI installers such as `Nutstore.Nutstore` write `WindowsInstaller = 1` to their ARP entries and would be identified as msi installers by WinGet, there are also some AI installers such as `Cocos.CocosDashboard` doesn't do that and would be identified as normal exe installers.

SpecterShell:
MSI installers built by Advanced Installer uses `APPDIR` as installation directory instead of `TARGETDIR` AND `INSTALLDIR`, so it may be necessary to add advanced installer (msi) to installer types.

Trenly:
In that case, would changing this to advInstallerExe suffice, which would allow for advInstallerMsi as well?

SpecterShell:
Yes.

Trenly:
> In that case, would changing this to advInstallerExe suffice, which would allow for advInstallerMsi as well?

@yao-msft Thoughts on this approach?

yao-msft:
> > In that case, would changing this to advInstallerExe suffice, which would allow for advInstallerMsi as well?
> 
> @yao-msft Thoughts on this approach?

More granular definitely works. If winget is to support AdvancedInstaller, we need some investigation and testing to be done internally to see how much we need to support them. From the discussion above, I think we need at least advInstallerExe and advInstalleMsi, it does not look very good if winget only supports "partial" of an installer technology. Mainly, the team may need a bit more time to try and understand what this installer technology is doing and decide how much we will support.

Trenly:
Going to close this due to it being extremely far behind on commits, no immediate need for it, and it would be easier to redo the changes than to try and fix the merge conflicts.

Issue #2516, opened 2022-09-14
Feature Request: Print command added to path for portable installs


Trenly:
> I've just installed Process Explorer
> 
> `winget install Microsoft.Sysinternals.ProcessExplorer`
> 
> but can't find out how to start it :) Start menu show me nothing on "Process Explorer" or "procexp" search requests

_Originally posted by @energister in https://github.com/microsoft/winget-cli/issues/377#issuecomment-1235685206_

I propose that when a portable package is installed, the alias used for calling the program be printed to the command output
```
PS C:\Windows\system32> winget install microsoft.sysinternals.processexplorer
Found Process Explorer [Microsoft.Sysinternals.ProcessExplorer] Version 16.43
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://live.sysinternals.com/procexp.exe
  ██████████████████████████████  2.70 MB / 2.70 MB
Successfully verified installer hash
Starting package install...
Path environment variable modified; restart your shell to use the new value.
Command line alias: 'process-explorer'
Successfully installed

Trenly:
@denelon - I may be able to do this one; Any issues you foresee with the proposed implementation?

denelon:
@ryfu-msft take a look :)

PR #2621, opened 2022-10-22
Print command aliases added on portable install

Trenly:
	- #2516

There is a slight technicality with this in that it prints the alias has been added when the symlink/file is created before the path has been modified. This is because the alias should be printed for each portable package installed. From my perspective, I don't believe this matters since there isn't any case in the code where adding to path would fail. 

image

image

-----

### DIFF
--- a/src/AppInstallerCLICore/PortableInstaller.cpp
+++ b/src/AppInstallerCLICore/PortableInstaller.cpp
@@ -141,7 +141,7 @@ namespace AppInstaller::CLI::Portable
                 AICLI_LOG(Core, Info, << "Portable install executed in user mode. Adding package directory to PATH.");
                 CommitToARPEntry(PortableValueName::InstallDirectoryAddedToPath, InstallDirectoryAddedToPath = true);
             }
-            m_stream << Resource::String::PortableAliasAdded << filePath.stem() << std::endl;
+            m_stream << Resource::String::PortableAliasAdded << ' ' << filePath.stem() << std::endl;
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1341,6 +1341,9 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="NoPackageSelectionArgumentProvided" xml:space="preserve">
     <value>No package selection argument was provided; see the help for details about finding a package.</value>
   </data>
+  <data name="PortableAliasAdded" xml:space="preserve">
+    <value>Command line alias added: </value>
```
yao-msft, line 1345: nit:
```suggestion
    <value>Command line alias added:</value>

```

And update the above m_stream << code to print the ' '

src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -141,6 +141,7 @@ namespace AppInstaller::CLI::Portable
                 AICLI_LOG(Core, Info, << "Portable install executed in user mode. Adding package directory to PATH.");
                 CommitToARPEntry(PortableValueName::InstallDirectoryAddedToPath, InstallDirectoryAddedToPath = true);
             }
+            m_stream << Resource::String::PortableAliasAdded << filePath.stem() << std::endl;
```
ryfu-msft: So if you choose to print it here, the message will only be displayed if the symlink was successfully created. If it fails, `InstallDirectoryAddedToPath` will be set to true and this block of code will never be hit again, thus hiding your command alias message.  Here is what I would do: 

```
else if (entry.Filetype == PortableFileType.Symlink)
{
   if (!InstallDirectoryAddedToPath) { ... }

  m_stream << Resource::String::PortableAliasAdded << filePath.stem() << std::endl;
}
```

That way you can guarantee the message is always displayed to the user for each symlink that was intended to be created.

Trenly: Ahh, I see. I thought placing it outside the if-else was enough, but I can see how it could fail now

ryfu-msft (approved):

Issue #2625, opened 2022-10-22
Add ExpectedReturnCodeEnum value for `Invalid Parameter`


Trenly:
### Description of the new feature / enhancement

MSI and Advanced Installer both support this, as do many custom installers

PR #2635, opened 2022-10-24
Add Expected Return Code for InvalidParameter

Trenly:
  - #2625 
-----

This PR adds two additional expected return codes:
1) InvalidParameter
2) SystemNotSupported

It also incorporates these into the default expected return codes where applicable.
This will likely cause a merge conflict with #2617, but I would prefer this PR to be merged first, if accepted.

Other issues fixed:
- Fixes 'packageInUseByApplication' to use `else if` instead of `if`
- Fixes the expected response of 'ERROR_INSTALL_REJECTED', as that error 1654 is not `BlockedByPolicy`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2635)


src/AppInstallerCommonCore/Manifest/ManifestCommon.cpp
```diff
@@ -614,6 +633,28 @@ namespace AppInstaller::Manifest
                 { HRESULT_FROM_WIN32(ERROR_INSTALL_CANCEL), ExpectedReturnCodeEnum::CancelledByUser },
                 { HRESULT_FROM_WIN32(ERROR_PACKAGE_ALREADY_EXISTS), ExpectedReturnCodeEnum::AlreadyInstalled },
                 { HRESULT_FROM_WIN32(ERROR_INSTALL_PACKAGE_DOWNGRADE), ExpectedReturnCodeEnum::Downgrade },
+                { HRESULT_FROM_WIN32(ERROR_DEPLOYMENT_BLOCKED_BY_POLICY), ExpectedReturnCodeEnum::BlockedByPolicy},
+                { HRESULT_FROM_WIN32(ERROR_INSTALL_POLICY_FAILURE), ExpectedReturnCodeEnum::BlockedByPolicy},
+                { HRESULT_FROM_WIN32(ERROR_PACKAGES_IN_USE), ExpectedReturnCodeEnum::PackageInUse },
+                /*
+                * For the below return codes, 'argument': conversion from 'HRESULT' to 'unsigned long', signed/unsigned mismatch
+                * 
+                { HRESULT_FROM_WIN32(APPX_E_INVALID_PACKAGING_LAYOUT), ExpectedReturnCodeEnum::InvalidParameter },
```
yao-msft, line 642: These appx errors are already HRESULT, so no need to use HRESULT_FROM_WIN32.

But I feel these appx errors do not belong to InvalidParameter category perfectly, as they are corrupted appx/msix package.

For example, I doubt APPX_E_INVALID_PACKAGESIGNCONFIG will ever happen during deployment because this is a package signing time config file, we should probably remove these error code mapping for all the commented out ones

yao-msft (approved):

Issue #1620, opened 2021-10-21
Ability to select installer when multiple matches are returned.


denelon:
I ran into the same problem today. I would expect it to be interactive like:

"Multiple packages matched your query. Please select one of the following:
[1] Microsoft.Powershell
[2] Microsoft.PowerShell-Preview"

Anyway thanks for the awesome work 💪

_Originally posted by @marvinpuethe in https://github.com/microsoft/winget-cli/issues/282#issuecomment-632585106_

denelon:
That's a great question, I was thinking about that, and haven't come up with the "best" solution yet. With the import command, the Id and source are specified.

SirusCodes:
> Another example for this in an unrelated issue: [#301 (comment)](https://github.com/microsoft/winget-cli/issues/301#issuecomment-940172239)

This kind of solution is fine I feel. Are there any plans to integrate it? maybe I can work on it

denelon:
@SiriusCodes we would love to have a community contributed solution. The first step would be to draft a spec so we can take feedback on the proposed implementation. There is a template and there are a few examples in the [specs folder](https://github.com/microsoft/winget-cli/tree/master/doc/specs). Please create a draft PR when you're ready for review and feedback 😊

denelon:
I think we would want a feature toggle and some discussion about how an interactive command might work. Some users may also prefer the current experience so they can continue having scripted solutions which don't support interactive behaviors. I'd expect that to be controlled by some kind of setting with a rational default behavior/setting.

SirusCodes:
If they have scripted solutions, they would be using IDs and won't ever see this interactive option while using scripts. I don't see a need for a feature toggle 🤔

danny007in:
any progress

SirusCodes:
> any progress

https://github.com/microsoft/winget-cli/pull/2650 you can see the spec here and further updates will be on this PR I guess

theking2:
> How would this work if winget was running non-interactively?

It would just fail in the same way it does now. Is there an exit code now? I don't know but it should be the same as to not break and be backwards compatible.

denelon:
Thanks for filing this. Closing as a duplicate of #301, which now captures this same request — an interactive numbered selection prompt when a query matches multiple packages — as the canonical issue. Consolidating the discussion there keeps the signal in one place.

If this is important to you, please add a 👍 to #301 so we can gauge interest.

*Drafted with the help of GitHub Copilot.*

Issue #2627, opened 2022-10-23
"winget upgrade -q Microsoft" doesn't find any package to upgrade


birbilis:
### Brief description of your issue

I do:

`winget list -q Microsoft`

and get back:

> Name                                              Id                                                    Version             Source
> -----------------------------------------------------------------------------------------------------------------------------------
> Microsoft Azure Compute Emulator - v2.9.7         Microsoft Azure Compute Emulator - v2.9.7             2.9.8999.43
> Microsoft Edge                                    Microsoft.Edge                                        106.0.1370.52       winget
> Microsoft Edge Update                             Microsoft Edge Update                                 1.3.169.31
> Microsoft Edge WebView2 Runtime                   Microsoft.EdgeWebView2Runtime                         106.0.1370.52       winget
> Microsoft Help Viewer 2.3                         Microsoft Help Viewer 2.3                             2.3.32406
> Microsoft Windows 10 SDK Installer                Microsoft Windows 10 SDK Installer                    21.0
> Xbox Game Speech Window                           Microsoft.XboxSpeechToTextOverlay_8wekyb3d8bbwe       1.21.13002.0
> Xbox TCUI                                         Microsoft.Xbox.TCUI_8wekyb3d8bbwe                     1.24.10001.0
> Microsoft Pay                                     Microsoft.Wallet_8wekyb3d8bbwe                        2.4.18324.0
> XAML Studio                                       Microsoft.XAMLStudio_8wekyb3d8bbwe                    1.1.0.0
> Fluent XAML Theme Editor                          Microsoft.FluentXAMLThemeEditor_8wekyb3d8bbwe         1.5.15.0
> Microsoft Emulator                                Microsoft.MicrosoftEmulator_8wekyb3d8bbwe             1.1.1081.0
> Xbox Game Bar Plugin                              Microsoft.XboxGameOverlay_8wekyb3d8bbwe               1.54.4001.0
> Snip & Sketch                                     Microsoft.ScreenSketch_8wekyb3d8bbwe                  10.2008.2277.0
> Page Analyzer                                     Microsoft.PageAnalyzer_8wekyb3d8bbwe                  1.5.0.0
> Windows Camera                                    Microsoft.WindowsCamera_8wekyb3d8bbwe                 2021.105.10.0
> Microsoft Remote Desktop Classic                  Microsoft.MicrosoftRemoteDesktopPreview_8wekyb3d8bbwe 10.1.1226.0
[...]

Issue #2013, opened 2022-03-10
Git Bash unresponsive after  Do you agree to all the source agreements terms? [Y] Yes  [N] No: 


codyempress:
Hello,
I am starting out in software development - please don't crucify me if I don't know the basics! 

I have been trying to cache my GitHub credentials in Git. I am trying to install Git CLI with Bash 

`winget install --id GitHub.cli`

I'm getting the following prompt: 
Do you agree to all the source agreements terms?
[Y] Yes  [N] No:

The problem is that I can't find a way to answer this! I have tried Yes, Y, y and even no but each time the cursor just jumps to a new line when I press enter. How do I accept this and get over this part?

denelon:
Which Terminal are you using? Is it CMD or Windows Terminal?

For a "work around", you can append "--acceptSourceAgreements" to accept the agreement from the "msstore" source. It's checked the first time you connect to get the list of packages from a source with agreements. You can also append "--acceptPackageAgreements" to accept any package level agreements. Those also come from the "msstore" source. In this case, the GitHub.cli package is not coming from the "msstore" source so:

`winget install GitHub.cli --acceptSourceAgreements` should work.

The prompt is looking for "Y" so when that is prompted, you should be able to type "Y" and then press [Enter].

mburde7:
I believe you have to run the install command in your command prompt, not your git bash

AndrewB-S:
> 

This solved my issue.

Trenly:
This is actually a [Known Issue](https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.md#known-issues) in Git-for-windows
> Some console programs, most notably non-MSYS2 Python, PHP, Node and OpenSSL, interact correctly with MinTTY only when called through winpty (e.g. the Python console needs to be started as winpty python instead of just python).

Calling `winget --wait` will cause the same unresponsiveness, but calling `winpty winget --wait` works as expected. In fact, the default aliases file for Git Bash contains a section on apps that require a win32 console -
```bash
case "$TERM" in
xterm*)
        # The following programs are known to require a Win32 Console
        # for interactive usage, therefore let's launch them through winpty
        # when run inside `mintty`.
        for name in node ipython php php5 python2.7
        do
                case "$(type -p "$name".exe 2>/dev/null)" in
                ''|/usr/bin/*) continue;;
                esac
                alias $name="winpty $name.exe"
        done
        ;;
esac

[...]

PR #2650, opened 2022-10-29
add spec for #1620

SirusCodes:

-----

Create a spec as requested by @denelon at https://github.com/microsoft/winget-cli/issues/1620#issuecomment-1206565768 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2650)


SirusCodes:

SirusCodes:
@denelon can you review it?

denelon:
My "meta" feedback is this seems to work well for installing a single package. 

We shouldn't have ambiguity with import since the source and the ID need to be specified in the packages.json file.

I wonder if we should have a "default" behavior and settings to change the default behavior. It may not be necessary, but this is a departure from the current behavior. 

I'm also not sure what this is going to "sound" like with narrator running.

Our telemetry needs to be reasoned about as well. If a user runs `winget install <something>` and we're not able to narrow it down to a single package, I don't want to have some "false" increased concept of "success".

denelon:
@yao-msft what are your thoughts here?

yao-msft:
This package selection mechanism could be extended to `winget upgrade <single package>`, `winget uninstall <single package>` and `winget show <single package>` as well, since these use the same worflow to pick a package internally.

Regarding default behavior, I think it'd be better to make this one opt-in from a specific argument since most automated systems, script writers will not be compatible with this change. Or we should announce this "breaking" default behavior well ahead before we make this the default behavior. In the early winget design principal, winget tries not to prompt for user input unless it has to(i.e. accepting agreements, missing arguments, etc), and the prompt can be disabled in many ways.

Regarding installing multiple packages in one command, it requires further thinking and framework changes are likely required to support this. Command line syntax may also need to be revisited.

denelon:
@yao-msft 
So maybe something like adding "--prompt" could prompt if the results are ambiguous, and a setting could be added to make the setting default and "--no-prompt" could override the setting via the CLI?

yao-msft:
> @yao-msft So maybe something like adding "--prompt" could prompt if the results are ambiguous, and a setting could be added to make the setting default and "--no-prompt" could override the setting via the CLI?

We already have a cli arg and user setting both called "DisableInteractivity" to indicate "no prompt", it's opted in for com callers(can not be overriden) but not from cli. So I guess the current default in cli is already "allow prompt". But currently only very few packages will trigger a prompt. With this feature, we'll see prompts a lot more, so I'm wondering if we should have a specific opt in for this(from settings), we may change the default behavior in future. In another thought, if existing scripts work, they should already be able to point to a specific package, so it might not be that "breaking" for existing scripts and automated systems. I think a setting for opt in is enough. And I'm open to other suggestions as well.

SirusCodes:
> This package selection mechanism could be extended to `winget upgrade <single package>`, `winget uninstall <single package>` and `winget show <single package>` as well, since these use the same worflow to pick a package internally.

Yes this can/should be extended to those as well. Will make that change in the spec. 

> Regarding default behavior, I think it'd be better to make this one opt-in from a specific argument (or settings?) since most automated systems, script writers will not be compatible with this change. Or we should announce this "breaking" default behavior well ahead before we make this the default behavior. In the early winget design principal, winget tries not to prompt for user input unless it has to(i.e. accepting agreements, missing arguments, etc), and the prompt can be disabled in many ways.

Can you please elaborate on this more? As I feel it will only come in if there are multiple matches and we get a table of packages in current implementation.
Everyone would write a script using Id's which are a unique to each package and if they get the selector they will get the current table in current implementation.
Only difference would be that in current implementation the command would return and not wait for an user input.

> Regarding installing multiple packages in one command, it requires further thinking and framework changes are likely required to support this. Command line syntax may also need to be revisited. 

Yeah, that is not the goal of this spec. I added that because if in future we plan to add that this is how selector should work.

SirusCodes:
> With this feature, we'll see prompts a lot more

How? This will only come into picture if there are multiple matches in current implementation.

For example.

```pwsh
winget install Powershell
```

It currently gives you 2 packages and after this it will also give you 2 packages just will a easy selector.

SirusCodes:
> I'm also not sure what this is going to "sound" like with narrator running.

Yeah I'm also not sure about that but if it reads it line by line then there shouldn't be a big problem.
Maybe I can try implementing a dummy one and see.

> Our telemetry needs to be reasoned about as well. If a user runs `winget install <something>` and we're not able to narrow it down to a single package, I don't want to have some "false" increased concept of "success".

I haven't thought about this. Can you help me out with it?

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,93 @@
...
+
+For Example -
+
+```pswh
+winget install Powershell Git
+```
+
+The above command will not install anything.
+
+If will ever support this then we might consider having _multiple_, _range_ and _not_ selectors.
```
Trenly, line 85:
```suggestion
If this is ever supported, potential considerations include having _multiple_, _range_ and _not_ selectors.
```
It is best practice to not reference a specific viewpoint such as `we`

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,93 @@
...
+Installing <user-input>...
+```
+
+> `<user-input>` - is the input by user
+
+After the user will give the input the installation should start.
+
+## UI/UX Design
+
+The idea is explained in [Solution Design](#solution-design).
```
Trenly, line 47: Will the user be required to press enter to confirm their input?

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,93 @@
...
+This will help the users to have a better experience and would decrease the chance of human error while copy-pasting the Id.
+
+### Compatibility
+
+It will not break anything.
+
+### Performance, Power, and Efficiency
+
+## Potential Issues
+
```
Trenly, line 70: There is a potential issue where non-interactivity would need to be preserved. Please add in specifications for how the client should behave when interactivity is disabled through the COM context, through the `--disable-interactivity` argument, or through user settings.

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly, line 47: What is the UI/UX if the user provides invalid input?

Trenly, line 47: What rules constitute valid input?

Trenly, line 47: How does the user exit the selector if they don't want to choose any of the available options?

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,93 @@
...
+After the user will give the input the installation should start.
+
+## UI/UX Design
+
+The idea is explained in [Solution Design](#solution-design).
+
+## Capabilities
+
+### Accessibility
+
```
Trenly, line 52: New text in the menu (e.g `Please select the packages`) will need to be localized. There may need to be considerations made for if the indices used for selecting packages also need to be localized

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly, line 47: Is there a limit to how many items can be in the selector?

Trenly, line 47: Is tab completion valid for the selector?

Trenly, line 47: Are users required to select an index, or could they enter a package ID if they wanted to?

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,93 @@
...
+It should not have any direct impact on screen readers.
+
+### Security
+
+It should not have any security issues.
+
+### Reliability
+
+This will help the users to have a better experience and would decrease the chance of human error while copy-pasting the Id.
```
Trenly, line 61: Will there need to be any measures taken against mistyping the selection? It is a lot easier to accidentally type `2` instead of `3`  than it is to enter an incorrect package ID.

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Abstract
+
+The winget.exe client simply lists all the matches and if the user wants to install the app they need to re-run the command with Id.
+
+The UX can be improved here by giving users an option to select the package instead of asking them to re-run the command with Id.
+
+## Inspiration
+
+It is not a good experience to copy Id and then run the command again.
```
Trenly, line 20: Some users (me especially) will not want to use this feature. There should be a way to toggle this feature off in settings, outside of the setting for disabling interactivity.

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly, line 70: There is potential confusion when using the `--silent` argument. As a user, I would expect that `--silent` would not lead to an interactive menu. Having reviewed the spec and being familiar with how Winget works, I understand that the selector is not part of the package install, but this could be a sticking point for many new users and should be preempted.

Trenly (changes requested):

Trenly:
> I wonder if we should have a "default" behavior and settings to change the default behavior. It may not be necessary, but this is a departure from the current behavior.

I think this is a necessity, especially for this feature. As was mentioned in the original issue and I touched upon in my review, there are users who will not want this feature at all - even outside of a scripting perspective. I would be a strong advocate for including user settings.

> Our telemetry needs to be reasoned about as well. If a user runs winget install <something> and we're not able to narrow it down to a single package, I don't want to have some "false" increased concept of "success".

I think this is as "simple" as adding a context flag that gets set whenever the interactive selector gets called, and sending that flag along with the telemetry. I would still count it as a success if the user was able to manually disambiguate, but I can understand if the internal needs differ here.

> This package selection mechanism could be extended to winget upgrade <single package>, winget uninstall <single package> and winget show <single package> as well, since these use the same workflow to pick a package internally.

I think implementing this for upgrade would conflict with [#2627](https://github.com/microsoft/winget-cli/issues/2627). It may be a good feature in the meantime though. I agree with adding it for uninstall and show.

> With this feature, we'll see prompts a lot more, so I'm wondering if we should have a specific opt in for this(from settings), we may change the default behavior in future. In another thought, if existing scripts work, they should already be able to point to a specific package, so it might not be that "breaking" for existing scripts and automated systems. I think a setting for opt in is enough. And I'm open to other suggestions as well.

I go both ways on this. On one hand you have people like me who are technically inclined that would prefer to see and work with CLI output directly rather than have any form of interactivity in a CLI application, except that which is strictly necessary. On the other hand you have people who are using winget to install their favorite application and want it to be as quick and seamless as possible, and having interactivity would help. I think opt-out is easiest to implement from the `DisableInteractivity` perspective, and keeps the same current default behavior as other interactivity, but I prefer the original design principal of not asking for input unless we have to. I say that `opt-in` would be my final preference. (Apologies for long section, had to write my thoughts down for me to understand them myself)

> Regarding default behavior, I think it'd be better to make this one opt-in from a specific argument (or settings?) since most automated systems, script writers will not be compatible with this change

I agree. Not only because of the points mentioned above, but I know that some scripts like [marinet101/WingetUI](https://github.com/martinet101/WingetUI/blob/53222c6baeafd9a1669c8fbe089120d78f3be24b/wingetui/wingetHelpers.py#L174) actually use the table output and parse the text. Adding in an additional column here or making it so that the feature is opt-out would be a breaking change for them. 

> Everyone would write a script using Id's which are a unique to each package and if they get the selector they will get the current table in current implementation.

Id's aren't always guaranteed to be unique to a package. The item which is guaranteed to be unique is the Source-ID pair. It is entirely possible that two sources have the same identifier for different packages. 

> In another thought, if existing scripts work, they should already be able to point to a specific package, so it might not be that "breaking" for existing scripts and automated systems.

The more I think about this, I believe that the dynamic nature of sources would make this a breaking change. It is entirely possible that a script which is working today could be broken tomorrow by someone adding a new package to the repository, assuming the script was pointing at a volatile field like a package name; Or even in cases where a sub-package is added and there is potential for a moniker conflict that we aren't aware of. I would suspect that a sysadmin would rather have 1 package that failed to install with a very clear log about why as opposed to coming back several hours later just to find that the script hung because they weren't aware of the `--disable-interactivity` flag, and thought there would be no issues since the script was working fine before.

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: Yes

SirusCodes: We can tell them this is an invalid input. Please give the correct input.

SirusCodes: The packages will be placed in the listed in a table. User will type the row number

SirusCodes: We can have something like "0 to exit"

Trenly: That error message should be included in the spec then

Trenly: That should be included in the spec then

SirusCodes: Haven't thought about it. But my initial thought is to show all the matches which we get🤔

SirusCodes: No

SirusCodes: Both seems fine to me, but index is easier in my opinion. What do you feel?

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: Do windows localize numerical forms automatically?

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: So if a user entered `1` it would be valid, but what if they entered `1.` or `1withSomeText` or `99999`. These need to be included in the spec. Something like `Only entries which exactly match a row number are considered valid input` with an example table output and then bullet point examples of what would be considered valid or invalid

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: I'm not sure, but it would need to be a consideration that needs to be made as to whether they will or won't be

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: That should be included in the spec then

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: Yeah, it's something which is something that can happen. Adding another prompt to take the confirmation saying "Do you want to install XYZ package? (Y/n): ". But again, it's an extra step.

I would add this in the in doc. Don't have any idea how we can handle it

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,93 @@
...
+
+It should not have any security issues.
+
+### Reliability
+
+This will help the users to have a better experience and would decrease the chance of human error while copy-pasting the Id.
+
+### Compatibility
+
+It will not break anything.
```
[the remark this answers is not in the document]
SirusCodes: Oh, never knew this. Thanks for highlighting this.

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: Yeah, we should not show selector when `--silent` argument is passed.

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: If I run `winget install -q " "` there would be 1899 entries for the user to select from, and that's just using the winget source, not including the msstore source or private sources. I'm not saying that it's necessarily a bad idea to show them all, since the current behavior of `winget search` is to show them all.

I believe that once you exceed 10-20 options, the user input wasn't good enough to begin with and it should revert to the list behavior without the selector, with a message like 'please refine your query`

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: Thanks, will keep this in mind.

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: Why wouldn't it be? It could very easily cycle through the row indices. I'd rather hit <kbd>Tab</kbd> <kbd>Enter</kbd> than <kbd>1</kbd> <kbd>Enter</kbd> or <kbd>Tab</kbd> <kbd>Tab</kbd> <kbd>Enter</kbd>than <kbd>2</kbd> <kbd>Enter</kbd>

Either way, whether it is supported or not supported should be in the spec

SirusCodes: Yeah will add that in docs

Trenly: I'd be more inclined to flexibility. As @denelon likes to say - meet the users where they are. As a user, I would feel more confident entering a full package ID just to have the peace of mind that I didn't make a typo, but I can also see where others would want to use the index

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: Perhaps it's a future consideration

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: > Yeah, we should not show selector when `--silent` argument is passed.

I think more opinions would be needed here. IMO `--silent` shouldn't control how winget behaves as a CLI, that's what the `--disable-interactivity` prompt is for. My point was more that some mention should be made around this, or it should be documented that `--silent` doesn't suppress this and that `--disable-interactivity` would. @yao-msft and @ryfu-msft would be the experts here though

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: I guess it's a personal choice. I prefer `1` `Enter`.
Surely, it can support both, but should it?

What are your opinions @denelon @yao-msft?

SirusCodes: Yeah, supporting both seems like a better idea. Will add it in the doc.

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: I will verify it and add in the spec

doc/specs/#1620 - Installer selector on multiple matches.md
denelon: I'm curious as to whether I would prefer to enter the "number" or the "ID" to disambiguate. My initial thought is that I would type the number.

The number is certainly less typing, but when we think about the tab completion it would likely be less error prone in either case.

If we support both a number or a package identifier, I think I would want the tab to prioritize the package ID first just for the sake of clarity.

In that case, the tab auto complete would be pretty fantastic.

denelon:
I think the default behavior should not change. This would certainly be breaking. I would expect to have a setting for "disambiguation". maybe "prompt" and "no-prompt". Then if one had the default they could override with "--prompt" or "--no-prompt".

The default setting would be "no-prompt"

Overriding the default setting would look like:

`winget install PowerToys --prompt`

Changing the default setting would look like:
```JSON
"packageDisambiguation": {"prompt"}
```
Overriding the "prompt" setting would look like:

`winget install PowerToys --no-prompt`

denelon:
Changing this behavior to be the default would require:
* #2128

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: Interesting thoughts;  Although that does bring up the edge case where there may be two packages with the same Identifier from different sources, in which case the number would have to be used instead of the package identifier.

Perhaps that calls for some sort of logic like -
> If user enters a row number, treat it as a selection. 
> If the user has entered text, treat it as a new query using only the table results as the source

That would mean that the behavior would be as follows, assuming a table like this:
```pwsh
> winget install Example
Multiple packages found matching.
No. Name            Id              Source
-------------------------------------------
 1  Example         Example.Package custom
 2  Example Beta    Example.Package winget
 3  Example Browser Example.Browser winget

Please enter a row number or refine your query.
> _
```

User enters `2` -
```pwsh
> 2
Found Example Beta [Example.Package] *****
Downloading ******
```
User enters `1` -
```pwsh
> 1
Found Example [Example.Package] *****
Downloading ******
```
User enters `Example.Browser` -
```pwsh
> Example.Browser
Found Example Browser [Example.Browser] *****
Downloading ******
```
User enters `--source custom`
```pwsh
> --source custom
Found Example [Example.Package] *****
Downloading ******
```
User enters `--source winget` -
```pwsh
> --source winget
Multiple packages found matching.
No. Name            Id              Source
-------------------------------------------
 1  Example Beta    Example.Package winget
 2  Example Browser Example.Browser winget

Please enter a row number or refine your query.
> _
```
User enters `Example.Package` -
```pwsh
> Example.Package
Multiple packages found matching.
No. Name            Id              Source
-------------------------------------------
 1  Example         Example.Package custom
 2  Example Beta    Example.Package winget

Please enter a row number or refine your query.
> _
```
User enters `--source msstore` -
```pwsh
> --source msstore
No package found matching selection criteria.
```

SirusCodes: Thanks @Trenly this looks pretty powerful

SirusCodes:
Hey @denelon, @yao-msft and @Trenly for your suggestions. I have updated the specs from suggestions you have given.

- How it will handle `--disable-interactivity`
- How it will handle invalid inputs
- How user can refine their search results
- Added how the settings should look like
- Added a compatibility issue with MinTTY (#2013) 
- Added a limit on the number of packages shown
- Made it an opt-in feature



Tho there are a few things where I don't have much knowledge about and would be great if you can help me out in those parts.

> I think implementing this for upgrade would conflict with https://github.com/microsoft/winget-cli/issues/2627. It may be a good feature in the meantime though. I agree with adding it for uninstall and show.

@Trenly I'm not sure how this will conflict it 

---

> I think this is as "simple" as adding a context flag that gets set whenever the interactive selector gets called, and sending that flag along with the telemetry. I would still count it as a success if the user was able to manually disambiguate, but I can understand if the internal needs differ here.

I don't have any knowledge how telemetry and localization work and hence couldn't add anything about it in spec.

It would be great if you can point be where I can get to know about those.

---

I tried by printing an example on terminal with screen reader on. It read me the whole thing correctly but read `No.` (abbreviation of number) as `No` (as in answer to question).

Is there a way where we can get to know that the screen reader is on and print `Number` instead of `No.` or we should stick with `Number`?

--- 

I need your suggestions on how refine search should work. Should it refine results on the basis of exact match or match pattern?

I'm inclined towards matching patterns, but it may result into extra steps for users. Like if there are 2 packages with name Example and a package with name Example2 pattern matching will result in to retaining all three if user types Example to refine them. But at the same time, it will help users to type less to refine results.

Again, I would like it to be as pattern matching thing instead of exact match.

---

As previously discussed in [review comment](https://github.com/microsoft/winget-cli/pull/2650#discussion_r1010012165) to have tab selection.

I'm not sure if it will be intuitive for users to use `Tab` to cycle through number. But if we are going with this, I want your opinion on if it should only cycle through number or package names as well? It doesn't make much sense to me to have both, cycling through numbers is more than enough in my opinion.

Also, if we are having `Tab` to cycle forward then we should have `Tab` + `Shift` to cycle backwards.

Trenly:
> @Trenly I'm not sure how this will conflict it

If a user were to type `winget upgrade Microsoft` - there are multiple packages that match. The conflict would be whether the selection menu is shown, or if all packages matching `Microsoft` would be upgraded

> Is there a way where we can get to know that the screen reader is on

I'm certain this is possible, it would just be a matter of how big the change would be to support this. It's probably better to use the full text `Number` all the time, or an alternate name like `Index` 

> I need your suggestions on how refine search should work. Should it refine results on the basis of exact match or match pattern?

I would suggest this be implemented as a temporary source. Essentially, the results in the table (including those that are truncated), would become a local source, and the default query behavior would be used. I believe this is a substring match, but I'd leave it up to whatever the current default search behavior is

>  I want your opinion on if it should only cycle through number or package names as well

My interpretation of denelon's comments is that tab completion would be most beneficial for cycling through Package Identifiers, and wouldn't be as helpful / necessary for numbers.

> Also, if we are having Tab to cycle forward then we should have Tab + Shift to cycle backwards.

I agree. This is the current implementation of the tab completion used in the client, and the behavior should be carried forward.

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,235 @@
...
+
+```pwsh
+"packageDisambiguation": "no-prompt"
+```
+
+### Valid inputs
+
+- Number - It should match with a _row number_ (0 will exit the selector)
+- Text - It should match either the package _Name_ or _Id_
+- `--source` flag - It refines the search with a specific package source. It can only be used if a user enters text or standalone.
```
Trenly, line 52: I would extend this to be any of the default query flags. Essentially, I think of the table results as a new source that can be queried against, and whatever is input is the [[-q] <query>] for a search against the new source.

Probably don't need things like `-n` or `--header` or `--accept-source-agreements`, but I would imagine that a majority of the options would still be available
```
Searches for packages from configured sources.

usage: winget search [[-q] <query>] [<options>]

The following command aliases are available:
  find

  --tag                       Filter results by tag
  --cmd,--command             Filter results by command
  -n,--count                  Show no more than specified number of results (between 1 and 1000)
```

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,235 @@
...
+
+Example: `23` in the list of `20`` entries
+
+```pwsh
+Please enter row number which are present in entries above.
+```
+
+### Number of results
+
+We should add a `--list-count` flag to customize the number of packages listed in the prompt. By default, the user will see a list of 20 packages.
```
Trenly, line 74: Users already have the `-n,--number` flag they can pass; Perhaps this would be better as a user setting?

```json
 "Interactivity": {
      "description": "Interactivity settings",
      "type": "object",
      "properties": {
        "disable": {
          "description": "Controls whether interactive prompts are shown by the Windows Package Manager client",
          "type": "boolean",
          "default": false
        },
        "listItemLimit":{
	  "description": "Controls how many items are shown in interactive selection lists",
	  "type": "integer",
          "default": 20,
          "minimum": 0
        }
      }
    },

File: doc/specs/#1620 - Installer selector on multiple matches.md [hunk 10]
```diff
@@ -0,0 +1,235 @@
...
+
+If the number of packages is more than allowed it should show a text saying how many more packages are available
```
Trenly, line 76: How would users see the additional packages if they don't want any of the first 20?

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,235 @@
...
+ 1  Example    Example.Example1     msstore
+ 2  Example    Example.Example2     winget
+ 3  Example    Example.Example22    msstore
+
+Please select a row number, or enter package name or ID to refine search...
+> --source winget
+Downloading Example (Example.Example2)...
+```
+
+If the `--source` query matches nothing then ask the user to retry.
```
Trenly, line 172: I think this gets back to the idea of preference vs requirement. If the user specifies the source, it would be a requirement and not a preference, so if no packages are found using the source, then I would expect the result to be `No package found matching input criteria` and the menu to exit.

> A short note on preferences vs requirements. The original concept has always been that there are user preferences, and user requirements. User requirements must be met in order to proceed; example - a user sets a requirement of x64 packages only, an x86 package can never be installed. If a manifest only contains x86 versions, it won't be available to the user. User preferences change the prioritization, but do not need to be met to proceed; example - a user sets a preference of en-GB for Installer Locale. If a manifest contains the en-GB locale, it will be used, regardless of what the system locale is, but if the en-GB locale is not present, the installation will continue with the most applicable locale.
>  
> User requirements and preferences can be set through the settings file. However, any command line switches override the settings and are treated as a requirement, not as preference. So if the user has a requirement for x64 architecture, but specifically uses `--architecture x86`, then the requirement becomes x86 just for that run of the install command.

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: I never thought it as CLI inside a CLI. Wouldn't that be an overkill for a simple thing? 👀

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: Oh, we can use that... we don't need to add more I guess

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: I thought about it. I felt it was not necessary but more I think about it I feel it's good to have.

`-m` and `--more` flag is ok I feel.

Will add it in the next iteration.

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: Oh, that makes so much sense. Thanks for that. I always that it as a preference.

SirusCodes:
> If a user were to type `winget upgrade Microsoft` - there are multiple packages that match. The conflict would be whether the selection menu is shown, or if all packages matching `Microsoft` would be upgraded

So that should be added in _potential issues_ section, right?

> I'm certain this is possible, it would just be a matter of how big the change would be to support this. It's probably better to use the full text `Number` all the time, or an alternate name like `Index`

`Number` feels a better user-friendly word. 
 
> I would suggest this be implemented as a temporary source. Essentially, the results in the table (including those that are truncated), would become a local source, and the default query behavior would be used. I believe this is a substring match, but I'd leave it up to whatever the current default search behavior is

I feel default one is pattern matching by looking at the `winget upgrade Microsoft` example.
 
> My interpretation of denelon's comments is that tab completion would be most beneficial for cycling through Package Identifiers, and wouldn't be as helpful / necessary for numbers.

---

Perfect, will make the necessary changes in next iteration.

doc/specs/#1620 - Installer selector on multiple matches.md
Trenly: > I never thought it as CLI inside a CLI. Wouldn't that be an overkill for a simple thing? 👀

It isn’t a CLI in a CLI though.

I'll use the example of the default `winget` source. That source is a pre-indexed source, which is built from the community repository. At its core, the `winget` source is just a database file, called `index.db`. When a user runs `winget search`, the CLI runs several queries against the index (index.db) to come up with a list of the packages matching the search criteria. 

My suggestion is, take all of the results that are returned and store them in a temporary database file. Then, a user can use *any* of the default search parameters when refining  the query.

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,270 @@
...
+
+If we find multiple packages then the user should get a selector.
+
+> It should not be shown if `--diable-interactivity` is set.
+
+### Settings
+
+The user needs to **opt-in** to this feature.
+
+Winget can use `--prompt` and `--no-prompt` flags to enable and disable it respectively.
```
Trenly, line 34:
```suggestion
Winget can use `--prompt` and `--disable-interactivity` flags to enable and disable it respectively.
```

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,270 @@
...
+
+The user can change the default setting to always show the prompt by -
+
+```pwsh
+"packageDisambiguation": "prompt"
```
Trenly, line 39:
```suggestion
```json
"interactivity" " {
  "packageDisambiguationPrompt": true
}
```
It would be best to use the existing settings structure

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+and disable it by -
+
+```pwsh
+"packageDisambiguation": "no-prompt"
```
Trenly, line 45:
```suggestion
```json
"interactivity" " {
  "packageDisambiguationPrompt": false
}
```

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,270 @@
...
+
+The text input is said to be invalid only if it doesn't match any of the names or IDs. The user will see an error for invalid input and will be asked to retry.
+
+If the user enters the name/ID and source or only source and there isn't any source then the user will see an error message which says that cannot find the package with the source.
+
+The first query will fetch package details from the indexed database and further refinements will be made to the result from the initial query.
+
+### Tab completion
+
+The Tab completion would complete the package name/ID and cycle through the package ID on `Tab` press and reverse cycle through it on `Tab` + `Shift`.
```
Trenly, line 131:
```suggestion
The Tab completion would complete the package name/ID and cycle through the package ID on <kbd>Tab</kbd> press and reverse cycle through it on <kbd>Tab</kbd>+<kbd>Shift</kbd>.
```

doc/specs/#1620 - Installer selector on multiple matches.md
```diff
@@ -0,0 +1,270 @@
...
+- `--name` - Filter results by name
+- `--moniker` - Filter results by the moniker
+- `--tag` - Filter results by tag
+- `--cmd` or `--command` - Filter results by command
+- `-s` or `--source` - Find the package using the specified source
+- `-e` or `--exact` - Find the package using exact match
+
+### Invalid inputs
+
+If a user enters an invalid value it should show an error asking the user for input again,
```
Trenly, line 66: I think this could be a point of contention. If using the value as additional refinement, I would prefer it to exit with the message `No package found matching input criteria`. However, this would be a UX choice that denelon or someone else at MSFT would need to make

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: I used `--no-prompt` because @denelon suggested it in https://github.com/microsoft/winget-cli/pull/2650#issuecomment-1297777064

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: @denelon or @yao-msft can you please suggest something here?

doc/specs/#1620 - Installer selector on multiple matches.md
SirusCodes: This was something from @denelon's https://github.com/microsoft/winget-cli/pull/2650#issuecomment-1302541619

Trenly: It isn’t very clear what this is for then, as the way it is described here doesn’t match denelons comment.

I'll take a try at making some smaller tweaks and will create a PR into your branch

SirusCodes: Merged

SirusCodes:
@Trenly should we also rename of file to _Query refinement.md_?

Trenly:
> @Trenly should we also rename of file to _Query refinement.md_?

Probably a good idea

doc/specs/#1620 - Installer selector on multiple matches.md
denelon: Are you looking for a suggestion on the error or something related to the heuristic for determining error cases?

Trenly: @denelon - Already resolved it with my recent PR into the forked branch

SirusCodes:
@denelon, @yao-msft can you have a look?

yao-msft:
> @denelon, @yao-msft can you have a look?

Hi @SirusCodes, 

Thanks for your contribution and continued refinement of the spec. We are currently in the process of getting a release candidate of winget v1.4.0. We will get back to the spec when this feature work is scheduled.

Usually the spec will be checked in around the time implementation is done. And the spec might need to be modified and extended to match the actual implementation. I.e. features actually done, features moved to be considered in the future, a little description on the technical implementation like data schema, and expected behaviors and known issues, etc.

Thanks,
Yao

bidaumm:
It appears this has been stagnated since Dec '22
Is this still on the cards for future implementation?

denelon:
Yes, we still want to offer an "interactive" mode for WinGet. We've been discussing several scenarios internally. This is one of the topics most likely to lead towards a WinGet 2.0.

Ideally, the default behavior for that version of WinGet will support better interactive experiences for users. We're also very aware of the desire to have non-interactive scripted experiences. We've been discussing the impact for automated scenarios and how we would have the client respond so scripted behaviors could more easily isolate the failures encountered by interruptions caused by interactive prompts expecting a user to respond.

If we offer the interactive experience as an option a user could "opt-in" we would need to figure out the best way to help them discover this new mode. We would certainly need settings and command-line arguments to toggle the behavior as a default mode as well as an override option.

mominshaikhdevs:
what's the latest update on this?

doc/specs/#1620 - Query Refinement for Disambiguation.md
```diff
@@ -0,0 +1,220 @@
...
+## Potential Issues
+* There should be no effect of the `--silent` flag on the selector.
+* Having the `upgrade` command included in this would conflict with [#2627](https://github.com/microsoft/winget-cli/issues/2627)
+  
+  For example - `winget upgrade Microsoft` might have a match with hundreds of packages and use might not want to see this prompt. Although this is an opt-in feature, there might be a clash if both features are enabled.
+
+## Future considerations
+* Keyboard interactivity using <kbd>↑</kbd>, <kbd>↓</kbd>, and <kbd>Ener</kbd>
+
+* Currently, Winget does not support installing multiple packages in a single command.
```
Trenly, line 201: This has changed since the last iteration of the spec;

It should probably be updated so that each package which needs disambiguation goes through the disambiguation prompt at the start of the install flow.

Trenly (changes requested):

SirusCodes:
Hi all, I'm a bit busy. Will try to look at the PR this weekend.

doc/specs/#1620 - Query Refinement for Disambiguation.md
SirusCodes: @Trenly are you talking about https://github.com/microsoft/winget-cli/pull/4282 this? If yes, then what needs to be considered?

Trenly: > @Trenly are you talking about #4282 this? If yes, then what needs to be considered?

No, I'm talking about #2861

SirusCodes:
@Trenly @denelon I have updated the spec.

I thought disambiguating each package separately would be a better idea because it will leave out edge cases where a user might filter out the package they wanted to install and go back through this whole process.

denelon:
@SiriusCodes, I've done a readthrough. The spec looks fairly thorough to me. At this point I'd like to get one of the engineers to perform a bit of a sanity check 😊

I'm also thinking we need to reason about what this would imply with the PowerShell way of doing things. I'm not sure it makes sense to have this kind of treatment in PowerShell though, so I'll see if I can get a few folks to look at it and give feedback.

@JohnMcPMS

doc/specs/#1620 - Query Refinement for Disambiguation.md
```diff
@@ -0,0 +1,275 @@
...
+}
+```
+In order to disable the query refinement feature -
+```json
+"interactivity" : {
+   "refineAmbiguousQueries": false // This is the default behavior
+}
+```
+
+From the command line, users should be able to use `--refine-query` to override the user setting and show the query refinement, and `--no-refinement` to override the user setting and not show the query refinement.
```
JohnMcPMS, line 47: Not a big fan of the command line parameters.  These just don't seem like something that would need to be frequently specified.

doc/specs/#1620 - Query Refinement for Disambiguation.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Refinement Source
+The results from the query needing refinement should be placed into a temporary database to be used for further refinement. This database should be treated as a new temporary source to be queried. Upon completion of the command, this temporary database shall be removed. If the database is already existing and a new command is run, the database shall be overwritten.
```
JohnMcPMS, line 50: I don't see any need for a file for this purpose.  Even if SQLite is used, it can be in memory.

doc/specs/#1620 - Query Refinement for Disambiguation.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Refinement queries
+Input to the package refinement should be treated as a query by default. This should match the behavior of the parent command. Each entry should also be assigned an index number which can be used to reference the package. The index number shall be treated as a fully qualified reference, and entry of an exact match to the number as a query shall take precedent.
```
JohnMcPMS, line 53: I would rather we define a smaller scope to the subsequent searches so that it can be implemented via property comparisons rather than needing to use SQLite for full fidelity searching.

doc/specs/#1620 - Query Refinement for Disambiguation.md
```diff
@@ -0,0 +1,275 @@
...
+
+> Example: A user runs `winget install Example` and receives the result:
```
JohnMcPMS, line 55: There is an implication that enabling this feature will make the initial query use the same semantics as `search`.  I don't think that you want that.

doc/specs/#1620 - Query Refinement for Disambiguation.md
```diff
@@ -0,0 +1,275 @@
...
+> ```
+> Multiple packages found matching input criteria.
+> Number Name       Id                   Source
+> -----------------------------------------------
+>   1    Example 3  Example.Example3     msstore
+>   2    Example 2  Example.Example2     winget
+> ```
+> If the user were to input `'Example 2'`, it would be treated as `winget install -q 'Example 2'` using only the results in the table above as the source
+ 
+Users should be able to refine their query using any of the standard or shared flags.
```
JohnMcPMS, line 65: Again, I would prefer that it be more of a straightforward property search without any need to evaluate complex inputs like this.

doc/specs/#1620 - Query Refinement for Disambiguation.md
```diff
@@ -0,0 +1,275 @@
...
+- `--help` or `?` - To show help for all the flags for disambiguation 
+
+### Number of results
+
+The `-n` or `--number` flag should be available for the `install`, `upgrade`, `show`, and `uninstall` commands to limit the number of packages returned for refinement.
+
+If the number of packages is more than allowed it should show the message that additional entries have been truncated due to the result limit. Packages truncated from the results should be included in the database for refinement queries.
+
+### Tab Completion
+Tab completion shall complete the package identifier just as it would for the parent command of the refinement. The behavior of forward cycling with <kbd>Tab</kbd> and reverse cycling with <kbd>Shift</kbd>+<kbd>Tab</kbd> shall be retained.
```
JohnMcPMS, line 85: This is probably quite complicated to implement properly; I would not suggest supporting it.

JohnMcPMS (commented):
I would suggest that this feature should be targeted at minor disambiguation, not as a full iterative search mechanism.  The number of results that one can reasonably expect with a single package targeted search is small enough that I think the limit on package count and some of the flows should be reconsidered.

SirusCodes:
> I would suggest that this feature should be targeted at minor disambiguation, not as a full iterative search mechanism. The number of results that one can reasonably expect with a single package targeted search is small enough that I think the limit on package count and some of the flows should be reconsidered.

Just went through the spec again, as I had forgotten about it 😅. I vaguely remember that I wanted to have a way where I can type `winget install Node` and it would give me a list of available packages with a hit with Node, and I can select the package I want to install.

Slowly it became a full-on iterative search. I agree it would be better if we could keep it simple as a first iteration and only target minor disambiguations, but I don't get the limit on package count?
If we have 20 hits for Node, we should show them all, which could be displayed on the terminal, and let the user type something to filter it out and finally select what they want.

I'm not using Windows (and winget); hence, I might not be up-to-date on what the current implementation looks like vs. when I first wrote the spec, but I would love to see this go live. I would see when I can find some time to update the spec and clean it up.

Issue #929, opened 2021-05-05
winget should not install an already installed package


besmirzanaj:
# Description of the new feature/enhancement
winget should check if an application is already installed before it runs the installer

# Proposed technical implementation details (optional)
check for installed software (could be a registry key) before you run the install command. 
example: if the `winget install <software>` is run once, then the next time it runs, it should prompt that the `<software>` is already installed and not proceed to reinstall it again.

Current implementation will reinstall the software over and over: 

```
PS C:\Windows\system32> winget install "Visual Studio Code"
Found Visual Studio Code [Microsoft.VisualStudioCode]
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://az764295.vo.msecnd.net/stable/3c4e3df9e89829dce27b7b5c24508306b151f30d/VSCodeUserSetup-x64-1.55.2.exe
  ██████████████████████████████  68.9 MB / 68.9 MB
Successfully verified installer hash
Starting package install...
Successfully installed

PS C:\Windows\system32> winget install "Visual Studio Code"
Found Visual Studio Code [Microsoft.VisualStudioCode]
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://az764295.vo.msecnd.net/stable/3c4e3df9e89829dce27b7b5c24508306b151f30d/VSCodeUserSetup-x64-1.55.2.exe
  ██████████████████████████████  68.9 MB / 68.9 MB
Successfully verified installer hash
Starting package install...
Successfully installed
```

denelon:
There may be cases where a subsequent install is intentional. A user might be attempting a downgrade scenario.

We might also prompt "Did you want to upgrade?" or something like that as well.

besmirzanaj:
Can we add option reinstall or flag --force?

denelon:
We should notify the user that the package is already installed and inform them it's possible to use "upgrade". We will still likely keep the `--force` option for install, and may add this to upgrade. There are scenarios like an unknown version that might be helped out in this case. It might seem a bit strange in the case it's actually a downgrade. That might be a case where we can't detect when the version is unknown, but if the version is known, we might recommend using force as well.

denelon:
How about:

The current version is installed on the system:
```
>winget install "visual studio code"
Warning: Microsoft Visual Studio Code [Microsoft.VisualStudioCode] is already installed.
No upgrade is available.
>
[...]

PR #2655, opened 2022-11-01
Support --no-upgrade option for install flow

yao-msft:
Related to #929 

There are community asks to allow skipping package upgrade if any version is already installed.

Unit tests added.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2655)


src/AppInstallerCommonCore/Errors.cpp
```diff
@@ -188,10 +190,28 @@ namespace AppInstaller
...
+                return "Archive malware scan failed.";
+            case APPINSTALLER_CLI_ERROR_PACKAGE_ALREADY_INSTALLED:
+                return "Found at least one version of the package installed.";
+
+            // Install errors
             case APPINSTALLER_CLI_ERROR_INSTALL_PACKAGE_IN_USE:
-                return "Application is currently running.Exit the application then try again.";
+                return "Application is currently running. Exit the application then try again.";
             case APPINSTALLER_CLI_ERROR_INSTALL_INSTALL_IN_PROGRESS:
                 return "Another installation is already in progress.Try again later.";
```
Trenly, line 216: Nit: Since you fixed the space above, might want to fix this one too

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -309,7 +309,8 @@ namespace AppInstaller::CLI::Workflow
                         context.Reporter.Info() << Resource::String::Cancelled << std::endl;
                         return;
                     }
-                    else if (searchContext.GetTerminationHR() == APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE)
```
ryfu-msft, line 312: >searchContext.GetTerminationHR() [](http://example.com/codeflow?start=29&length=32)

Nit: should this be assigned to a variable so that it is not called 3 times within this if-else block.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1467,4 +1466,10 @@ Please specify one of them using the `--source` option to proceed.</value>
     <value>Archive scan detected malware; proceeding due to --force</value>
     <comment>{Locked="--force"}</comment>
   </data>
+  <data name="NoUpgradeArgumentDescription" xml:space="preserve">
+    <value>Do not try to upgrade any installed version during install operation</value>
```
ryfu-msft, line 1470: >>Do not try to upgrade any installed version during install operation [](http://example.com/codeflow?start=10&length=69)

Nit: This sounds like a warning message rather than an argument description.

How about: "Skips upgrade if an installed version already exists"

ryfu-msft (approved):

PR #2662, opened 2022-11-03
Remove redundant GetInstallerArgs call

yao-msft:
Also added test to verify default args be passed to nested installer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2662)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
@@ -319,7 +319,6 @@ namespace AppInstaller::CLI::Workflow
             ScanArchiveFromLocalManifest <<
             ExtractFilesFromArchive <<
             VerifyAndSetNestedInstaller <<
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -319,7 +319,6 @@ namespace AppInstaller::CLI::Workflow
             ScanArchiveFromLocalManifest <<
             ExtractFilesFromArchive <<
             VerifyAndSetNestedInstaller <<
-            GetInstallerArgs <<
```
yao-msft, line 322: GetInstallerArgs is removed because it will be invoked accordingly in the ExecuteInstallerForType below.

yao-msft:
Will use #2663

Issue #2545, opened 2022-09-23
Add and Validate NestedInstallerSha256


Trenly:
### Description of the new feature / enhancement

Some applications distributed through zip files or other archive types are no longer in active development. In other cases, these installers may not be available from the original publisher, but may still be safe. Although hash collisions are rare in any scenario, a malicious actor with sufficient knowledge of compression algorithms can force a hash collision much easier on an archive type installer than on other types of installers. This creates a scenario where it may be beneficial in some cases, but certainly not all cases, that having and verifying the hash of the nested installer would provide an additional layer of security. 

One great example of this is [Universal Silent Switch Finder (USSF)](https://github.com/microsoft/winget-pkgs/pull/80300). The software is no longer under active development and is only available from insecure sites. However, the application is known to be safe with the current hash. Being able to add the nested installer hash for an additional verification would make me as a user feel more secure knowing that both the file being downloaded and the file being installed are being verified.

I am aware of the `InstallationMetadata` which allows for an optional hash to be supplied for ensuring the install is detected correctly, but this doesn't seem to be related to validation before install.

### Proposed technical implementation details

Add an optional `NestedInstallerSha256` key, valid when installer type is Zip

Trenly:
@denelon @ryfu-msft - I understand that I'm asking for a bit of scope creep here, but is it possible this could be pulled into the initial 1.4 release? I'm no expert on implementation details, but given that infrastructure is already in place for validating hashes, I'm assuming this would be as "simple" as adding a WorkflowTask to validate files between extraction and install, and a small bit of additional logic in manifest validation.

Pipeline validation of nested hashes could be left as a separate enhancement request since they aren't currently being validated.

denelon:
Related to:
* #140

Trenly:
[Policy] Area-Manifest

PR #2675, opened 2022-11-07
Add and Validate NestedInstaller FileSha256

Trenly:
  - #2545

Still working on adding unit and E2E tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2675)


yao-msft:
Just my 2 cents before we get too far into implementation. I'm not sure if this is of high importance since we already validated the zip hash before extracting. If we are concerned that extracted files may be tampered after extraction and before execution, we could improve the code to put a write exclusive handle on to be extracted files and keep the handle after execution completes.

Trenly:
> Just my 2 cents before we get too far into implementation. I'm not sure if this is of high importance since we already validated the zip hash before extracting. If we are concerned that extracted files may be tampered after extraction and before execution, we could improve the code to put a write exclusive handle on to be extracted files and keep the handle after execution completes.

I think that is certainly one concern, but the other concern I have is that it is much easier to cause a hash collision on Zip files than other file types. Presume for a moment that someone had a malicious application, put it inside a zip, and included in the zip a second binary file that had been carefully crafted to make the hash of the overall zip file the same. That would pass hash validation still. Ignoring the fact that malware scans would probably catch it anyways, its just an optional added layer of protection to validate the hash of the file that was actually extracted

Trenly:
Closed as this is not a high-priority item, I don't feel like resolving merge conflicts, and there are probably better implementations that could be planned for the future

PR #2683, opened 2022-11-08
Don't enable/disable admin settings if group policy prevents it

AmelBawa-msft:
## What was done?
- If an admin setting was disabled by a group policy, and user attempts to enable it using `winget settings --enable`, then report an error message.
- Similarly, if an admin setting was enabled by group policy, and user attempts to disable it using `winget settings --disable`, then report an error message.

## Test?
- Added tests covering the various state combinations.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2683)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/SettingsFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/SettingsFlow.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 #include "pch.h"
+#include "AppInstallerStrings.h"
 #include "SettingsFlow.h"
 #include <winget/UserSettings.h>
 #include <winget/AdminSettings.h>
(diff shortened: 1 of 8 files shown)


src/AppInstallerCLICore/Workflows/SettingsFlow.cpp
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 #include "pch.h"
+#include "AppInstallerStrings.h"
```
yao-msft, line 4: >"AppInstallerStrings.h" [](http://example.com/codeflow?start=9&length=23)

nit: <AppInstallerStrings.h> for headers outside of current project

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1475,4 +1475,12 @@ Please specify one of them using the `--source` option to proceed.</value>
   <data name="PackageAlreadyInstalled" xml:space="preserve">
     <value>A package version is already installed. Installation cancelled.</value>
   </data>
+  <data name="EnableAdminSettingFailed" xml:space="preserve">
+    <value>Cannot enable %1. This setting is controlled by policy. For more information contact your system administrator.</value>
```
yao-msft, line 1479: >This setting is controlled by policy. [](http://example.com/codeflow?start=29&length=37)

nit: do you foressee other cases admin setting will fail to anable/disable? If yes, the wording may need to be revisited.

yao-msft (approved):
:shipit:

PR #2690, opened 2022-11-11
Update NuGet Packages

Trenly:
As far as I could tell, these changes don't cause any build issues. But, knowing my luck, I'm probably missing something. Lets just see where this lands. 

This doesn't update all NuGet packages, just the ones that seemed like they were close in version and were unlikely to have breaking changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2690)


yao-msft:
Usually, we update package reference versions only when we need to (i.e. vulnerability detected, compliance issue, need new version for new feature, out of support, etc). And we "try" to make all projects point to same package version to minimize potential issues.

Since this one mostly only affects tests and projects that are not used in other places(except the wingetutilinterop which is used in winget service), I think it's probably ok if the wingetutilinterop change is reverted.

yao-msft (approved):

Issue #2697, opened 2022-11-14
PowerShell DSC Resources for WinGet


denelon:
### Description of the new feature / enhancement

I would like to be able to use PowerShell DSC to install and configure WinGet.

Several issues have made mention to challenges associated with getting WinGet installed (or in a valid state) on different Windows SKUs or environments:

* https://github.com/microsoft/winget-cli/issues/2040
* https://github.com/microsoft/winget-cli/issues/2081
* https://github.com/microsoft/winget-cli/issues/2222

@cdhunt also made a PR with a PowerShell script to help with this challenge:
* https://github.com/microsoft/winget-cli/pull/2474
* https://github.com/microsoft/winget-cli/issues/2230
* https://github.com/microsoft/winget-cli/issues/2434

A couple of mentions have also suggested PowerShell DSC:
* https://github.com/microsoft/winget-cli/issues/698#issuecomment-1070044758
* https://github.com/microsoft/winget-cli/issues/291#issuecomment-632922571

### Proposed technical implementation details

As we are nearing the first release of WinGet PowerShell modules and cmdlets, this would be aligned with that same effort.

* #221

denelon:
Naming things is hard:
* https://github.com/microsoft/winget-cli/issues/2697

denelon:
We should consider the options for making this work in the System context in terms of getting "WinGet" on the system, and then when the user logs in, the actual "registration" for the user is "faster".

PR #2701, opened 2022-11-16
Spec PowerShell DSC Resources for WinGet

msftrubengu:
Spec for 

* #2697

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2701)

### DIFF
--- a/doc/specs/#2697 - PowerShell DSC resources.md
+++ b/doc/specs/#2697 - PowerShell DSC resources.md
@@ -26,8 +26,10 @@ The proposed DSC resources are (names TBD):
 2. WinGetAdminSettings
 3. WinGetSources
 4. WinGetIntegrity
+5. WinGetPackages
+6. WinGetPackage
 
-### WinGetUserSettings
+### 1. WinGetUserSettings
 This resource will be in charge of editing the settings.json with the specified settings.
 
 Enable-WinGetSetting and Disable-WinGetSetting functions already exists in Microsoft.WinGet.Client, but those in reality refer to administrator settings. User settings are not necessarily something that you can just enable or disable. Because of that we need new cmdlets that are specific to user settings. We should also take in consideration that some settings are temporal, for example experimental features. We cannot add a new/remove new parameters every time one of them is implemented. Also, there won't be any validation other than converting the Hashtable into a JSON. This is similar on how `winget settings` will just open the preferred text editor to let the user modify the file and will only inform of malformations on the next winget command.
@@ -40,11 +42,11 @@ Returns the Hashtable representation of winget's settings.json file.
 
 *Set-WinGetUserSettings*
 
-Writes the input settings into winget's settings.json file. Default behavior must be additive and there must be an option to overwrite content.
+Writes the input settings into winget's settings.json file. Default behavior is overwrite. Optionally do an additive set.
 
 *Test-WinGetUserSettings*
 
-Returns true is the inputs settings values match the current settings. By default, compares only the settings specified, optionally make a full comparison.
+Returns true is the inputs settings values match the current settings. By default, makes a full comparison, optionally compares only the settings specified.
 This could live inside the DSC resource, but in order to keep the code contained and easy to manage and test is better to expose it.
 
 #### DSC Resource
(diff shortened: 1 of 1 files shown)


doc/specs/#2697 - PowerShell DSC resources.md
```diff
@@ -0,0 +1,256 @@
...
+    Settings = [string[]]
+    SID = [string]
+    [DependsOn = [string[]]]
+    [PsDscRunAsCredential = [PSCredential]]
+}
+```
+
+*Settings*
+
+A Hashtable with the admin settings to be set.
```
JohnMcPMS, line 130: Is this to future proof against non-boolean admin settings?

doc/specs/#2697 - PowerShell DSC resources.md
```diff
@@ -0,0 +1,256 @@
...
+    [PsDscRunAsCredential = [PSCredential]]
+}
+```
+
+*SID (Key)*
+
+See SID property of UserSettings
+
+*Sources*
+This is an array of a new PowerShell WinGetSource class. It contains the name, type and argument of the source. Type is optional as it default as "Microsoft.PreIndexed.Package". Think about it as the deserialized form of the result of `winget sources export` without the Identifier and Data properties.
```
JohnMcPMS, line 176: Could use a Hashtable that matches our group policy schema; keeps the resource from changing but its likely that the implementation will have to account for it either way.  I think worth doing.

doc/specs/#2697 - PowerShell DSC resources.md
```diff
@@ -0,0 +1,256 @@
...
+
+It might seem natural to use the type "Microsoft.Management.Deployment.PackageCatalogReference" returned by Get-WinGetSource, but PackageCatalogReference is not what we really want.
+
+Existing PowerShell DSC property Ensure will tell if the sources need to be added or deleted. Test will verify that all the sources are present.
+
+*Ensure*
+
+This is a enum with two values: Absent and Present. Defaults to present. For Set, Absent means call Remove-WinGetSource and Ensure means if not there call Add-WinGetSource otherwise call Reset-WinGetSource to repair it.
```
JohnMcPMS, line 184: I think it might be worth having a separate property for reset rather than always defaulting to doing it.

doc/specs/#2697 - PowerShell DSC resources.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Solution Design
+
+A new module Microsoft.WinGet.DSC is going to be created where all the WinGet related Powershell DSC resources are going to live. The resources are simple wrappers around functionality that Microsoft.WinGet.Client currently provides, with some additions. We are going to use PSDesiredStateConfiguration 3.0.0 for development, so the new module is going to initially be an alpha release.
+
+The proposed DSC resources are (names TBD):
+1. WinGetUserSettings
+2. WinGetAdminSettings
+3. WinGetSources
+4. WinGetIntegrity
```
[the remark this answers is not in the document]
msftrubengu: To me it feels better to have one resource per package, but I get that we want to have one that takes the output of `winget export`. Maybe we can have both? Implementation-wise it will be mostly similar except parsing the json/hashtable.

doc/specs/#2697 - PowerShell DSC resources.md
msftrubengu: I was thinking it as more of a "don't update the module every time there's a new admin setting"

doc/specs/#2697 - PowerShell DSC resources.md
```diff
@@ -0,0 +1,256 @@
...
+
+A Hashtable with the user settings to be set.
+
+*SID (Key)*
+
+DSC resources require a key that is a string, a signed/unsigned integer, or Enum types. The key must be unique per resource, but we only expect to have one winget user settings resource in a session. To resolve this, we can use the SID of the current user as the key. This will also allow a future implementation where an administrator sets different user settings for different users. For now, we won't expect people to set this, we will do it internally.
+
+*Action*
+
+Action is an enum with two values: Partial, Full.
```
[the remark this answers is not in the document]
msftrubengu: Maybe we could use Purge. True means remove any unmanaged user settings basically doing an overwrite, false means merge.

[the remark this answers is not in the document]
msftrubengu: Or just Merge as a bool

PR #2702, opened 2022-11-16
Add trusted remote settings and initial code for MSStore source cert rotation

yao-msft:
Code is behind Experimental Feature and intentionally be hidden from other Experimental Features as server side work is to be done.
Added unit tests to validate the cert rotation flow.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2702)


src/AppInstallerCommonCore/Public/winget/Settings.h
```diff
@@ -76,6 +78,8 @@ namespace AppInstaller::Settings
         constexpr static StreamDefinition BackupUserSettings{ Type::UserFile, "settings.json.backup"sv };
         // The admin settings.
         constexpr static StreamDefinition AdminSettings{ Type::Secure, "admin_settings"sv };
+        // Trusted remote settings.
+        constexpr static StreamDefinition TrustedRemoteSettings{ Type::TrustedRemote, "trusted_remote_settings.msix"sv };
```
JohnMcPMS, line 82: I was imagining that a `StreamDefinition` for this would be more like:
```
constexpr static StreamDefinition Name { Type::TrustedRemote, "package relative file path"sv };
```
Thus removing the need for `TrustedRemoteSettings.h|cpp`.

The well known `StreamDefinition` for `MSStoreSoureCertificatePinningConfiguration.json` (not sure of actual file name) could be here, and the code loading it would construct new `StreamDefinition`s for each file within that.  So:
```
"RemoteSettingFileCertificate":"cert.cer"
```
   becomes
```
std::string certFilePath = remoteSettingsFileCertificateNode.value();
StreamDefinition certStreamDef{ Type::TrustedRemote, certFilePath };
```
With `certFilePath == "cert.cer"`

src/AppInstallerCommonCore/Certificates.cpp
```diff
@@ -546,4 +575,27 @@ namespace AppInstaller::Certificates
 
         return true;
     }
+
+    bool PinningConfiguration::LoadFromTrustedRemoteSettings()
```
JohnMcPMS, line 579: All of this JSON certificate pinning configuration load was previously test code.  This is making it product code; it probably should have tests added (for the JSON parsing part, not strictly targeting this that reads from the trusted settings).

src/AppInstallerCommonCore/Settings.cpp
```diff
@@ -347,7 +350,132 @@ namespace AppInstaller::Settings
...
+                    AICLI_LOG(Core, Info, << "Cached remote settings file is not up to date.");
+                    return false;
+                }
+
+                return true;
+            }
+
+            bool LockSettingsFile(const std::filesystem::path& settingsFile, std::unique_ptr<Msix::WriteLockedMsixFile>& lock)
+            {
+                lock = std::make_unique<Msix::WriteLockedMsixFile>(settingsFile);
```
JohnMcPMS, line 457: Is there a built-in wait for this lock creation, or could we be creating spurious failures due to contention over locking this file to read it?  Given that you are already planning on reading the entire thing in to memory, why not just read the file into memory without a lock.

src/AppInstallerCommonCore/Settings.cpp
```diff
@@ -347,7 +350,132 @@ namespace AppInstaller::Settings
...
+                    return false;
+                }
+
+                if (!LockSettingsFile(m_settingsFile, m_fileLock))
+                {
+                    return false;
+                }
+
+                Msix::MsixInfo remoteInfo{ s_RemoteSettings_Url };
+                if (remoteInfo.IsNewerThan(m_settingsFile))
```
JohnMcPMS, line 446: One thing this does not account for is an attempt at a replay attack where we never try to move forward (the attacker makes us see the same old file forever).

To remain trusted, we would need to have some way to time out an older file.  Encoding the creation date in the version is already being done, and this could be used to ensure that the current (and remote) packages are up to date.

Additionally, for performance we probably want to not even check for a remote update unless there is a failure (or maybe based on time).

JohnMcPMS (commented):
I think some changes are needed, but not blocking as I won't be able to review again for a while.

PR #2719, opened 2022-11-23
Export command for winget settings.

msftrubengu:
Add new command for `winget settings export`. This command will report the value of the winget admin settings formatted as json, as well as the path of the user settings file. 
```
C:\Windows\System32>wingetdev settings export
{"adminSettings":{"BypassCertificatePinningForMicrosoftStore":false,"LocalManifestFiles":true},"userSettingsFile":"%LOCALAPPDATA%\\Packages\\WinGetDevCLI_8wekyb3d8bbwe\\LocalState\\settings.json"}

C:\Windows\System32>winget.exe settings export
{"adminSettings":{"BypassCertificatePinningForMicrosoftStore":false,"LocalManifestFiles":true},"userSettingsFile":%LOCALAPPDATA%\\Microsoft\\WinGet\\Settings\\defaultState\\settings.json"}
```

Also adding the user settings json path in `winget --info` for completeness.

In preparation of #2697

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2719)


slumericanB (approved):

src/AppInstallerCLICore/Workflows/SettingsFlow.cpp
```diff
@@ -99,4 +126,22 @@ namespace AppInstaller::CLI::Workflow
             ShellExecuteW(nullptr, nullptr, L"notepad", filePathUTF16.c_str(), nullptr, SW_SHOW);
         }
     }
+
+    void ExportAdminSettings(Execution::Context& context)
+    {
+        AdminSetting settings[] =
```
JohnMcPMS, line 132: This would be better if there were a `Max` enum value and we could just enumerate `for (i = firstValidSetting; i < AdminSetting::Max; ++i)` so that any future settings are automatically included.

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
```diff
@@ -79,4 +87,26 @@ namespace AppInstaller::CLI
...
+
+    std::string SettingsExportCommand::HelpLink() const
+    {
+        return std::string{ s_SettingsCommand_HelpLink };
+    }
+
+    void SettingsExportCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        context <<
+            Workflow::EnsureRunningAsAdmin <<
```
yao-msft, line 109: >Workflow::EnsureRunningAsAdmin  [](http://example.com/codeflow?start=12&length=31)

Just curious is there a particular reason to require admin for export? Exporting seems harmless to me.

src/AppInstallerCLICore/Workflows/SettingsFlow.cpp
```diff
@@ -12,6 +12,33 @@ namespace AppInstaller::CLI::Workflow
     using namespace AppInstaller::Settings;
     using namespace AppInstaller::Utility;
 
+    namespace
+    {
+        struct AdminSettings
+        {
+            AdminSettings()
+            {
+                root["adminSettings"] = Json::ValueType::objectValue;
```
yao-msft, line 21: > root["adminSettings"] = Json::ValueType::objectValue; [](http://example.com/codeflow?start=15&length=54)

Is the exported settings to be imported in Powershell DSC? Do we need a schema version or it's unlikely to change in the future?

src/AppInstallerCLICore/Commands/SettingsCommand.cpp
msftrubengu: There's no real reason. Removing.

src/AppInstallerCLICore/Workflows/SettingsFlow.cpp
msftrubengu: For DSC... not entirely. The admin settings resource will take a hashtable with something really similar to value of the adminSettings. We just needed a way to expose it without reading registry keys.

Also, because reasons (mostly testing and not having to special case powershell) I'm adding the path of the local user settings file to `winget settings export` . Also `winget --info` for completeness. 

But yes, I'm going to add a new schema.

src/AppInstallerCLICore/Commands/RootCommand.cpp
```diff
@@ -193,6 +193,7 @@ namespace AppInstaller::CLI
             };
 
             info << std::endl << Resource::String::Logs << ": "_liv << Runtime::GetPathTo(Runtime::PathName::DefaultLogLocationForDisplay).u8string() << std::endl;
+            info << std::endl << Resource::String::UserSettings << ": "_liv << Runtime::GetPathTo(Runtime::PathName::UserSettingsFileLocation).u8string() << std::endl;
```
yao-msft, line 196: >info << std::endl << Resource::String::UserSettings << ": "_liv << Runtime::GetPathTo(Runtime::PathName::UserSettingsFileLocation).u8string() << std::endl; [](http://example.com/codeflow?start=12&length=155)

We may need to remember to update this one when [https://github.com/microsoft/winget-cli/pull/2454](https://github.com/microsoft/winget-cli/pull/2454) is merged

yao-msft (approved):
:shipit:

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1526,4 +1526,13 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="PinAddBlockingArgumentDescription" xml:space="preserve">
     <value>Block from updating until the pin is removed, preventing override arguments</value>
   </data>
+  <data name="SettingsExportCommandLongDescription" xml:space="preserve">
+    <value>Export administrator settings as JSON</value>
```
yao-msft, line 1530: >Export administrator settings as JSON [](http://example.com/codeflow?start=11&length=37)

nit: update descriptions and below one

src/AppInstallerCLICore/Workflows/SettingsFlow.cpp
```diff
@@ -14,14 +14,16 @@ namespace AppInstaller::CLI::Workflow
...
     {
-        struct AdminSettings
+        struct ExportSettingsJson
         {
-            AdminSettings()
+            ExportSettingsJson()
             {
+                root["$schema"] = "https://aka.ms/winget-settings-export.schema.json";
                 root["adminSettings"] = Json::ValueType::objectValue;
+                root["userSettingsFile"] = Runtime::GetPathTo(Runtime::PathName::UserSettingsFileLocation).u8string();
```
JohnMcPMS, line 23: And this one should be the full path without the environment variable (thus not named `ForDisplay`).

JohnMcPMS (approved):

PR #2733, opened 2022-12-02
Add empty `pin` command

florelis:
This change adds an empty `pin` command. Actual implementation of the command will come as a separate PR. This change includes:
* Definition of the `pin` command and its sub-commands `add`, `remove`, `list`, `reset`. These definitions contain only the definition of arguments but are missing the actual execution code.
* A new experimental feature flag to hide this command.
* All the help resource strings for the command. Still pending all the resources that will be shown during actual execution.
* Adds a `source ls` alias for the `source list` command.
* Changes an existing string to make it consistent with the rest.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2733)

### DIFF
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -266,7 +266,7 @@ Experimental feature with the aim of managing dependencies, as of now it only sh
 
 ### pinning
 
-This feature enables the ability to pin packages to a specific version to prevent the Windows Package Manager from updating them.
+This feature enables the ability to pin packages to prevent the Windows Package Manager from updating them.
 You can enable the feature as shown below.
 
 ```json
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -253,12 +253,24 @@ You can enable the feature as shown below.
        "openLogsArgument": true
    },
 ```
-### Dependencies
+
+### dependencies
 
 Experimental feature with the aim of managing dependencies, as of now it only shows package dependency information. You can enable the feature as shown below.
 
 ```json
    "experimentalFeatures": {
        "dependencies": true
+   },
+```
+
+### pinning
+
+This feature enables the ability to pin packages to a specific version to prevent the Windows Package Manager from updating them.
+You can enable the feature as shown below.
+
+```json
+   "experimentalFeatures": {
+       "openLogsArgument": true
    },
 ```
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -271,6 +271,6 @@ You can enable the feature as shown below.
 
 ```json
    "experimentalFeatures": {
-       "openLogsArgument": true
+       "pinning": true
    },
 ```
--- a/src/AppInstallerCLICore/Commands/PinCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/PinCommand.cpp
@@ -13,7 +13,6 @@ namespace AppInstaller::CLI
     using namespace AppInstaller::Utility::literals;
     using namespace std::string_view_literals;
 
-    // TODO: Create this link with the docs!
     static constexpr std::string_view s_PinCommand_HelpLink = "https://aka.ms/winget-command-pin"sv;
 
     std::vector<std::unique_ptr<Command>> PinCommand::GetCommands() const
@@ -123,7 +122,6 @@ namespace AppInstaller::CLI
             Argument::ForType(Args::Type::Exact),
             Argument::ForType(Args::Type::CustomHeader),
             Argument::ForType(Args::Type::AcceptSourceAgreements),
-            Argument::ForType(Args::Type::Force),
         };
     }
 
--- a/src/AppInstallerCLICore/Commands/PinCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/PinCommand.cpp
@@ -1,0 +1,166 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "PinCommand.h"
+#include "Workflows/CompletionFlow.h"
+#include "Workflows/PinFlow.h"
+#include "Workflows/WorkflowBase.h"
+#include "Resources.h"
+
+namespace AppInstaller::CLI
+{
+    using namespace AppInstaller::CLI::Execution;
+    using namespace std::string_view_literals;
+
+    // TODO: Create this link with the docs!
+    static constexpr std::string_view s_PinCommand_HelpLink = "https://aka.ms/winget-command-pin"sv;
+
+    std::vector<std::unique_ptr<Command>> PinCommand::GetCommands() const
+    {
+        return InitializeFromMoveOnly<std::vector<std::unique_ptr<Command>>>({
+            std::make_unique<PinAddCommand>(FullName()),
+            std::make_unique<PinRemoveCommand>(FullName()),
+            std::make_unique<PinListCommand>(FullName()),
+        });
+    }
+
+    Resource::LocString PinCommand::ShortDescription() const
+    {
+        return { Resource::String::PinCommandShortDescription };
+    }
+
+    Resource::LocString PinCommand::LongDescription() const
+    {
+        return { Resource::String::PinCommandLongDescription };
+    }
+
+    std::string PinCommand::HelpLink() const
+    {
+        return std::string{ s_PinCommand_HelpLink };
+    }
+
+    void PinCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        OutputHelp(context.Reporter);
+    }
+
+    std::vector<Argument> PinAddCommand::GetArguments() const
+    {
+        // TODO
+    }
+
+    Resource::LocString PinAddCommand::ShortDescription() const
+    {
+        return { Resource::String::PinAddCommandShortDescription };
+    }
+
+    Resource::LocString PinAddCommand::LongDescription() const
+    {
+        return { Resource::String::PinAddCommandLongDescription };
+    }
+
+    void PinAddCommand::Complete(Execution::Context& context, Execution::Args::Type valueType) const
+    {
+        // TODO
+    }
+
+    std::string PinAddCommand::HelpLink() const
+    {
+        return std::string{ s_PinCommand_HelpLink };
+    }
+
+    void PinAddCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        // TODO
+    }
+
+    std::vector<Argument> PinRemoveCommand::GetArguments() const
+    {
+        // TODO
+    }
+
+    Resource::LocString PinRemoveCommand::ShortDescription() const
+    {
+        return { Resource::String::PinRemoveCommandShortDescription };
+    }
+
+    Resource::LocString PinRemoveCommand::LongDescription() const
+    {
+        return { Resource::String::PinRemoveCommandLongDescription };
+    }
+
+    void PinRemoveCommand::Complete(Execution::Context& context, Execution::Args::Type valueType) const
+    {
+        // TODO
+    }
+
+    std::string PinRemoveCommand::HelpLink() const
+    {
+        return std::string{ s_PinCommand_HelpLink };
+    }
+
+    void PinRemoveCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        // TODO
+    }
+
+    std::vector<Argument> PinListCommand::GetArguments() const
+    {
+        // TODO
+    }
+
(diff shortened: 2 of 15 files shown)


florelis:
Related to #476 and #2611 but does not resolve them yet

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -0,0 +1,244 @@
...
+#include "Resources.h"
+
+namespace AppInstaller::CLI
+{
+    using namespace AppInstaller::CLI::Execution;
+    using namespace AppInstaller::Utility::literals;
+    using namespace std::string_view_literals;
+
+    // TODO: Create this link with the docs!
+    static constexpr std::string_view s_PinCommand_HelpLink = "https://aka.ms/winget-command-pin"sv;
```
florelis, line 17: @denelon Can you create this even if it points to just the general documentation (or add a note for you to do that)? I remember with a previous command I added it we completely forgot about it...

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1357,7 +1357,8 @@ Please specify one of them using the `--source` option to proceed.</value>
     <value>Portable install failed; Cleaning up...</value>
```
florelis, line 1357: I really want to sort this file alphabetically but I know VS is just going to mess it up again...

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1483,4 +1484,47 @@ Please specify one of them using the `--source` option to proceed.</value>
...
+    <comment>{Locked{"'upgrade'"} Locked{"--include-pinned"} Locked{"winget upgrade --all"}</comment>
+  </data>
+  <data name="PinAddCommandShortDescription" xml:space="preserve">
+    <value>Add a new pin</value>
+  </data>
+  <data name="PinCommandLongDescription" xml:space="preserve">
+    <value>Manage package pins with the sub-commands. A pin can limit the Windows Package Manager from updating a package to specific ranges of versions, or it can prevent it from updating the package altogether. A pinned package may still update on its own and be updated from outside the Windows Package Manager.</value>
+  </data>
+  <data name="PinCommandShortDescription" xml:space="preserve">
+    <!-- TODO: We need to add "package pin[ning]" to the dictionary of terms for the loc team to know before sending this for localization -->
```
florelis, line 1498: Have we already done this for the other terms we use, like "source" or "package"?

src/AppInstallerCommonCore/Public/winget/ExperimentalFeature.h
```diff
@@ -25,6 +25,7 @@ namespace AppInstaller::Settings
             DirectMSI = 0x2,
             ZipInstall = 0x4,
             OpenLogsArgument = 0x8,
+            Pinning = 0x10,
```
florelis, line 28: It took me longer than I want to admit to remember what 2 x 8 is in hex 😅

src/AppInstallerCLICore/Commands/SourceCommand.h
```diff
@@ -37,7 +37,7 @@ namespace AppInstaller::CLI
 
     struct SourceListCommand final : public Command
     {
-        SourceListCommand(std::string_view parent) : Command("list", parent) {}
+        SourceListCommand(std::string_view parent) : Command("list", { "ls" }, parent) {}
```
florelis, line 40: This came to mind while I was adding the alias for the `pin` sub-commands. I don't know if we want to make this a separate PR so it goes out sooner than pinning.

src/AppInstallerCLICore/Commands/PinCommand.cpp
denelon: If I remember correctly, aka.ms will just default somewhere else if the URL isn't registered.
Yes, we can create the aka.ms links ahead of time with a temporary/placeholder URL and modify it at any point later.

denelon:
This should also include the update to the settings page as an experimental feature.

doc/Settings.md
```diff
@@ -262,3 +263,14 @@ Experimental feature with the aim of managing dependencies, as of now it only sh
...
 ```
+
+### pinning
+
+This feature enables the ability to pin packages to a specific version to prevent the Windows Package Manager from updating them.
+You can enable the feature as shown below.
+
+```json
+   "experimentalFeatures": {
+       "openLogsArgument": true
```
Trenly, line 274:
```suggestion
       "pinning": true
```

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+    std::vector<std::unique_ptr<Command>> PinCommand::GetCommands() const
+    {
+        return InitializeFromMoveOnly<std::vector<std::unique_ptr<Command>>>({
+            std::make_unique<PinAddCommand>(FullName()),
+            std::make_unique<PinRemoveCommand>(FullName()),
+            std::make_unique<PinListCommand>(FullName()),
```
Trenly, line 24: Does `reset` need to be included here?

src/AppInstallerCLICore/Commands/PinCommand.cpp
Trenly: Aka.ms links that don't exist just redirect to the Bing homepage iirc

src/AppInstallerCLICore/Commands/SourceCommand.h
Trenly: I would suggest a separate PR just to keep the git history cleaner

src/AppInstallerCLICore/Commands/PinCommand.cpp
florelis: Yeah, they redirect to the Bing homepage. I was mentioning creating it now pointing to the general documentation, because if we do forget about creating the right link, in my mind it's better if the link at least points to somewhere close instead of going to Bing.

src/AppInstallerCLICore/Commands/SourceCommand.h
florelis: Yeah, you're right... I have the bad habit of seeing things that I want to change but then being too lazy to do it in a separate PR...

Anyways, here it is #2736

doc/Settings.md
florelis: Right. That's what I get for just copying the one above 😅

src/AppInstallerCLICore/Commands/PinCommand.cpp
florelis: Added!

Trenly:
@jsoref

jsoref:
@Trenly: interesting, when I tested it, I got a very different result: https://github.com/check-spelling/winget-cli/pull/2#issuecomment-1336272299 (this comment has since been collapsed, but you can expand it).

From looking at the logs ([here](https://github.com/microsoft/winget-cli/actions/runs/3605526073/jobs/6076011084#step:2:508) and [there](https://github.com/check-spelling/winget-cli/actions/runs/3610555666/jobs/6084422806#step:2:508)), I suspect that [checkout-merge step](https://github.com/check-spelling/check-spelling/blob/d7cd2973c513e84354f9d6cf50a6417a628a78ce/action.yml#L324-L326) is not doing the right thing tm.

Here'a a rebased run:

https://github.com/check-spelling/winget-cli/pull/2/files#file-src-powershell-microsoft-winget-client-examples-sample_findpackage-ps1-L4

And here's me _trying_ to reproduce this PR as it was:
https://github.com/check-spelling/winget-cli/pull/3

I'm going to have to investigate the checkout-merge logic, it feels like it isn't quite behaving.

Can I ask that this be rebased after #2739 is merged? (Preferably squashing 74a97e6 into 1ede614f7b8a168ddbc30e7454aed24be5898a21 -- when I did a rebase, I tripped on a conflict there...)

src/AppInstallerCLICore/Commands/PinCommand.cpp
denelon: I added https://aka.ms/winget-command-pin

It currently defaults to the main winget page.

florelis:
Thanks for looking into this @jsoref !

> Can I ask that this be rebased after #2739 is merged? 
Rebased. Looks like it found one unrecognized word but I'm not planning on merging it with that line so I'll leave it like this for now.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
florelis: Looks like not. I added pinning to the internal work item tracking these.

doc/Settings.md
```diff
@@ -262,3 +263,14 @@ Experimental feature with the aim of managing dependencies, as of now it only sh
        "dependencies": true
    },
 ```
+
+### pinning
+
+This feature enables the ability to pin packages to a specific version to prevent the Windows Package Manager from updating them.
```
yao-msft, line 269: nit: it does not have to be a specific version

src/AppInstallerCLICore/Commands/SourceCommand.h
```diff
@@ -37,7 +37,7 @@ namespace AppInstaller::CLI
 
     struct SourceListCommand final : public Command
     {
-        SourceListCommand(std::string_view parent) : Command("list", { "ls" }, parent) {}
+        SourceListCommand(std::string_view parent) : Command("list", parent) {}
```
yao-msft, line 40: Will this actually remove the "ls" change done in previous pr during merging?

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -0,0 +1,245 @@
...
+            Argument::ForType(Args::Type::Name),
+            Argument::ForType(Args::Type::Moniker),
+            Argument::ForType(Args::Type::Tag),
+            Argument::ForType(Args::Type::Command),
+            Argument::ForType(Args::Type::Exact),
+            Argument{ "version"_liv, 'v', Args::Type::GatedVersion, Resource::String::GatedVersionArgumentDescription, ArgumentType::Standard },
+            Argument::ForType(Args::Type::Source),
+            Argument::ForType(Args::Type::CustomHeader),
+            Argument::ForType(Args::Type::AcceptSourceAgreements),
+            Argument::ForType(Args::Type::Force),
```
yao-msft, line 63: Curious what is the scenario for --force in pin add?

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -0,0 +1,245 @@
...
+            Argument::ForType(Args::Type::Id),
+            Argument::ForType(Args::Type::Name),
+            Argument::ForType(Args::Type::Moniker),
+            Argument::ForType(Args::Type::Source),
+            Argument::ForType(Args::Type::Tag),
+            Argument::ForType(Args::Type::Command),
+            Argument::ForType(Args::Type::Exact),
+            Argument::ForType(Args::Type::CustomHeader),
+            Argument::ForType(Args::Type::AcceptSourceAgreements),
+            Argument::ForType(Args::Type::Force),
```
yao-msft, line 126: Also curious what is the scenario for --force in pin remove here?

yao-msft (approved):

src/AppInstallerCLICore/Commands/PinCommand.cpp
florelis: The scenario I was thinking of is overwriting the pin. Like `winget pin add Pkg` adds a pin, then `winget pin add Pkg --version 1.*` would fail because it exists, but if you add `--force`, it overwrites it. But honestly, I just added it because it was in the spec and I was hoping you had a scenario in mind :D

yao-msft: That makes sense. I guess I thought about it some time ago and now my memory does not serve me well.

Trenly: Even more curious - should `--force` be added to the list of generic arguments (like `--wait` and `--verbose-logs`)? In the worst case, it doesn't change the behavior, and in most cases it will remove a few lines of code.

PR #2737, opened 2022-12-02
Standardize how arguments are used in resource strings

florelis:
This standardizes how we use arguments in resource strings and adds comments so that the localization team does not change them. We had different styles, like `"--argument"`, `argument`, `--argument`. This standardizes all to `--argument` for consistency.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2737)


yao-msft (approved):
:shipit:

Issue #2727, opened 2022-11-29
Switch for --uninstallPrevious for upgrade


Trenly:
### Description of the new feature / enhancement

The upgrade behavior can be set to uninstall the previous version of the application by default in the manifest. Usually, upgrade behavior is set to install unless uninstallPrevious is needed for a specific reason. 

As a power user there are times when it would be nice to override the behavior and uninstall the previous package anyways. One example is when a user has a local build of wingetcreate, the application will show for upgrade but cannot be upgraded unless it is uninstalled and re-installed.

This would also allow for the bypassing/overriding of the `Install Technology Differs` error by forcing a clean install

Trenly:
@denelon - Experimental

PR #2755, opened 2022-12-07
Add Experimental Argument for UninstallPrevious in Upgrade Flow

Trenly:
  - #2727

Left this just for the upgrade flow (and Installs that switched to using the upgrade flow). May potentially want to consider enabling for default install flow (for when --force) is used. Depends on user feedback and use cases

This will cause a merge conflict with microsoft/winget-cli#2733 due to the Experimental Features Enum; If pinning is merged first, I will be happy to rebase this

-----

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2755)


Trenly:
I need to start checking more thoroughly for typos before I push . . .

florelis:
> This will cause a merge conflict with https://github.com/microsoft/winget-cli/pull/2733 due to the Experimental Features Enum; If pinning is merged first, I will be happy to rebase this

Looks like I beat you to it, sorry!

Trenly:
> > This will cause a merge conflict with #2733 due to the Experimental Features Enum; If pinning is merged first, I will be happy to rebase this
> 
> Looks like I beat you to it, sorry!

No apology needed :) 
I was hoping pinning would be first anyways

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -257,6 +257,7 @@ namespace AppInstaller::CLI::Workflow
     {
         bool isUpdate = WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerExecutionUseUpdate);
         UpdateBehaviorEnum updateBehavior = context.Get<Execution::Data::Installer>().value().UpdateBehavior;
+        bool doUninstallPrevious = isUpdate && (updateBehavior == UpdateBehaviorEnum::UninstallPrevious || context.Args.Contains(Execution::Args::Type::UninstallPrevious));
```
yao-msft, line 260: This may have impact when dependencies are enabled, but I'm not blocking it since it's in experimental

yao-msft (commented):
Sorry we reviewed this late as we just had our v1.4 release candidate created last week.

The change looks good. Please merge to the latest when time available. Thanks.

And an open question, since import command also converts to upgrade when needed, does power user want to remove previous for them as well? I guess not, but I'll just leave the question here.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
Trenly: I'm not sure I understand how it would have an impact when dependencies are enabled, but since both are experimental I agree that it's not a blocker. I'll be happy to revisit the argument implementation if necessary

Trenly:
> And an open question, since import command also converts to upgrade when needed, does power user want to remove previous for them as well? I guess not, but I'll just leave the question here.

Possibly; I think it may be best to wait and see if we get feedback on it since I'm not sure how useful it will be in the import scenario

PR #2761, opened 2022-12-08
Fix issue that only 1 dependency package is installed

yao-msft:
Added tests. Also verified locally.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2761)

### DIFF
--- a/src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
+++ b/src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
@@ -202,14 +202,15 @@ namespace AppInstaller::Manifest
     struct Dependency
     {
         DependencyType Type;
-        string_t Id;
+        string_t Id() const { return m_id; };
         std::optional<Utility::Version> MinVersion;
 
-        Dependency(DependencyType type, string_t id, string_t minVersion) : Type(type), Id(std::move(id)), MinVersion(Utility::Version(minVersion)), m_foldedId(FoldCase(Id)) {}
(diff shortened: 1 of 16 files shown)


src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -223,7 +224,15 @@ namespace AppInstaller::Manifest
             return MinVersion <= Utility::Version(version);
         }
 
+        // m_foldedId should be set whenever Id is set
```
florelis, line 227: So, the issue was that in some cases it wasn't set?

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -202,14 +202,15 @@ namespace AppInstaller::Manifest
     struct Dependency
     {
         DependencyType Type;
-        string_t Id;
+        string_t Id() const { return m_id; };
         std::optional<Utility::Version> MinVersion;
```
florelis, line 207: Not `const string_t&`?

florelis (approved):

ryfu-msft (approved):

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
yao-msft: Correct. Before the fix, manifest parsing directly sets Id without anyway to set m_foldedId resulting in all dependency packages having same folded id and get treated as same package.

PR #2766, opened 2022-12-14
Add scope support to more commands and add provisioning support for msix and msstore types

yao-msft:
- Add scope parameter support to show, list, upgrade and uninstall commands
   Scope parameter in list, upgrade and uninstall commands works slightly differently from install and show commands. 
   Show, install commands: scope parameter is to filter the installer selection
   List, upgrade and uninstall: scope parameter is to filter the installed package selection, for upgrade, that means the package to upgrade is also filtered to the scope due to the InstalledScopeFilter in ManifestComparator
- Rewrite GetPackageFullNameFromFamilyName to use PackageManager methods to support system callers
- Add provisioning support for msix and msstore types when scope is machine
- Added e2e tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2766)


src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -121,16 +133,29 @@ namespace AppInstaller::Repository::Microsoft
                 SQLiteIndex index = SQLiteIndex::CreateNew(SQLITE_MEMORY_DB_CONNECTION_TARGET, Schema::Version::Latest());
 
                 // Put installed packages into the index
-                if (filter == PredefinedInstalledSourceFactory::Filter::None || filter == PredefinedInstalledSourceFactory::Filter::ARP)
+                if (filter == PredefinedInstalledSourceFactory::Filter::None || filter == PredefinedInstalledSourceFactory::Filter::ARP ||
+                    filter == PredefinedInstalledSourceFactory::Filter::User || filter == PredefinedInstalledSourceFactory::Filter::Machine)
                 {
```
ryfu-msft, line 138: Is this done on purpose for readability? Could be simplified to: 

if (filter != PredefinedInstalledSourceFactory::Filter::Msix)

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -121,16 +133,29 @@ namespace AppInstaller::Repository::Microsoft
...
+                    filter == PredefinedInstalledSourceFactory::Filter::MSIX ||
+                    filter == PredefinedInstalledSourceFactory::Filter::User)
                 {
-                    PopulateIndexFromMSIX(index);
+                    PopulateIndexFromMSIX(index, Manifest::ScopeEnum::User);
+                }
+                else if (filter == PredefinedInstalledSourceFactory::Filter::Machine)
+                {
+                    PopulateIndexFromMSIX(index, Manifest::ScopeEnum::Machine);
                 }
```
ryfu-msft, line 159:
```
if (filter != PredefinedInstalledSourceFactory::Filter::Arp)
{
    if (filter == PredefinedInstalledSourceFactory::Filter::Machine)
    {
        PopulateIndexFromMSIX(index, Manifest::ScopeEnum::Machine);
    }
    else
    {
        PopulateIndexFromMSIX(index, Manifest::ScopeEnum::User);
    }
}
```

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -78,7 +78,21 @@ namespace winrt::Microsoft::Management::Deployment::implementation
                 // Check if search behavior indicates that the caller does not want to do local correlation.
                 if (m_compositePackageCatalogOptions.CompositeSearchBehavior() != Microsoft::Management::Deployment::CompositeSearchBehavior::RemotePackagesFromRemoteCatalogs)
                 {
-                    ::AppInstaller::Repository::Source installedSource = ::AppInstaller::Repository::Source{ ::AppInstaller::Repository::PredefinedSource::Installed };
+                    ::AppInstaller::Repository::Source installedSource;
+                    auto manifestInstalledScope = GetManifestScope(m_compositePackageCatalogOptions.InstalledScope());
+                    if (manifestInstalledScope.first == ::AppInstaller::Manifest::ScopeEnum::User)
```
ryfu-msft, line 83: seems like you can just assign manifestInstalledScope.first directly to the local variable.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -308,6 +309,23 @@ namespace AppInstaller::CLI::Workflow
...
 
+    void EnsureRunningAsAdminForMachineScopeInstall(Execution::Context& context)
+    {
+        // Admin is required for machine scope install for installer types like portable, msix and msstore.
+        auto installerType = context.Get<Execution::Data::Installer>().value().EffectiveInstallerType();
+
+        if (installerType == InstallerTypeEnum::Portable ||
+            installerType == InstallerTypeEnum::MSStore ||
+            installerType == InstallerTypeEnum::Msix)
+        {
```
ryfu-msft, line 320: Use semantic check like IsAdminRequiredForMachineScopeInstall(installerType)

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -133,6 +134,11 @@ namespace AppInstaller::CLI::Workflow
             installOptions.CompletedInstallToastNotificationMode(AppInstallationToastNotificationMode::NoToast);
         }
 
+        if (Manifest::ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope)) == Manifest::ScopeEnum::Machine)
```
ryfu-msft, line 137: >if (Manifest::ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope)) == Manifest::ScopeEnum::Machine) [](http://example.com/codeflow?start=8&length=123)

Do we need to check if the InstallScope arg exists before calling this to prevent an exception from being thrown?

ryfu-msft (approved):

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
yao-msft: If arg does not exist, the GetArg returns empty, no exception is thrown. I copied this snippet from existing usage in other places.

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
yao-msft: I think it'd be better to check expected values instead of "!=" so future additions of the enum will not accidentally change previous behavior

PR #2767, opened 2022-12-14
Fix winget after a call to winget settings export

msftrubengu:
Fix bug introduced by #2719

Using `Runtime::GetPathTo` uses `PathDetails` which creates a directory when PathDetails.Create is true (which is the default). When `winget settings export` is called, that makes the code create a directory named the same as the settings file. Now, next winget run will try to open the file but it will find out it is a path and give up on life.

Since `Runtime::GetPathTo` is more related to path, getting the user settings json file path is not the best option. Move logic to UserSettings.cpp. Also catch the exception when trying to parse the fail to not make winget fail but use the default settings is something happens.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2767)

### DIFF
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -30,8 +30,9 @@ namespace AppInstaller::CLI
         std::string_view parent,
         Command::Visibility visibility,
         Settings::ExperimentalFeature::Feature feature,
-        Settings::TogglePolicy::Policy groupPolicy) :
-        m_name(name), m_aliases(std::move(aliases)), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy)
+        Settings::TogglePolicy::Policy groupPolicy,
+        CommandOutputFlags outputFlags) :
+        m_name(name), m_aliases(std::move(aliases)), m_visibility(visibility), m_feature(feature), m_groupPolicy(groupPolicy), m_outputFlags(outputFlags)
     {
         if (!parent.empty())
         {
@@ -933,7 +934,8 @@ namespace AppInstaller::CLI
     {
         try
         {
-            if (!Settings::User().GetWarnings().empty())
+            if (!Settings::User().GetWarnings().empty() &&
+                !WI_IsFlagSet(command->GetOutputFlags(), CommandOutputFlags::IgnoreSettingsWarnings))
             {
                 context.Reporter.Warn() << Resource::String::SettingsWarnings << std::endl;
             }
--- a/src/AppInstallerCommonCore/UserSettings.cpp
+++ b/src/AppInstallerCommonCore/UserSettings.cpp
@@ -95,11 +95,22 @@ namespace AppInstaller::Settings
 
         std::optional<Json::Value> ParseFile(const StreamDefinition& setting, std::vector<UserSettings::Warning>& warnings)
         {
-            auto stream = Stream{ setting }.Get();
-            if (stream)
+            try
             {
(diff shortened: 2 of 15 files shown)


yao-msft (approved):
:shipit:

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -93,11 +95,22 @@ namespace AppInstaller::Settings
...
+                    return ParseSettingsContent(settingsContentStr, setting.Name, warnings);
+                }
+            }
+            catch (const std::exception& e)
+            {
+                AICLI_LOG(Core, Error, << "Failed to read settings.json file. Reason: " << e.what());
+            }
+            catch (...)
+            {
+                AICLI_LOG(Core, Error, << "Failed to read settings.json file.. Reason unknown.");
```
JohnMcPMS, line 113:
```suggestion
                AICLI_LOG(Core, Error, << "Failed to read settings.json file. Reason unknown.");
```

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -93,11 +95,22 @@ namespace AppInstaller::Settings
...
-                std::string settingsContentStr = Utility::ReadEntireStream(*stream);
-                return ParseSettingsContent(settingsContentStr, setting.Name, warnings);
+                auto stream = Stream{ setting }.Get();
+                if (stream)
+                {
+                    std::string settingsContentStr = Utility::ReadEntireStream(*stream);
+                    return ParseSettingsContent(settingsContentStr, setting.Name, warnings);
+                }
+            }
+            catch (const std::exception& e)
```
JohnMcPMS, line 107: This exposes the scenario where both the normal and backup settings fail, which can result in no output to the user about settings load failure?  I think it just means an else for when the backup settings fails to load (~line 520 now) where we put in a different warning about not being able to load any settings file.

JohnMcPMS (approved):

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -518,6 +518,10 @@ namespace AppInstaller::Settings
                     m_type = UserSettingsType::Backup;
                     settingsRoot = settingsBackupJson.value();
                 }
+                else
+                {
+                    AICLI_LOG(Core, Warning, << "Failed loading settings files.");
```
JohnMcPMS, line 523: I was thinking more along the lines of the user visible portion (without opening the log file).  So a new:
```
m_warnings.emplace_back(StringResource::String::SettingsWarningWhoopsSettingsDidntLoad);
```
That would then at least warn the user "Hey, we couldn't load *any* settings, so you are getting the default behavior".

src/AppInstallerCLICore/Command.cpp
```diff
@@ -933,7 +934,8 @@ namespace AppInstaller::CLI
     {
         try
         {
-            if (!Settings::User().GetWarnings().empty())
+            if (!Settings::User().GetWarnings().empty() &&
+                !WI_IsFlagSet(command->GetOutputFlags(), CommandOutputFlags::IgnoreSettingsWarnings))
```
JohnMcPMS, line 938: As discussed, I would think this would be better done using the output channels mechanism, but this is acceptable.

The nice thing about that mechanism being built out is that any error/exception could be output in a JSON format as well.  But we have only so much time.

PR #2769, opened 2022-12-15
Add database for tracking pins and base implementation for pin commands

florelis:
This adds a database for tracking pins and a basic implementation of the `pin` commands to add, remove and list them. The implementation for the database is roughly copied from the database for portable packages.
* Added a new index schema `IPinningIndex` and a v1.0 implementation for it. The database contains a single table, where each pin is a separate row. The fields for each pin are
  + package ID, source ID - These two form the key used to look up pins
  + pin type - pinning/blocking/gating
  + version - only for gating pins
  A package available from multiple sources could have multiple independent pins. The operations allowed are adding a pin (`pin add`), updating a pin (`pin add --force`), removing a pin (`pin remove`), getting all pins (`pin list`), resetting all pins (`pin reset --force`), and getting a single pin (to be used on update/install).
* Added workflow tasks to implement the `pin` commands with the basic functionality of adding, removing and querying.

What is missing from this PR:
* All the logic for version gating
* Respecting the pins in the install/update/import commands
* Make the pin commands apply to all sources for a package, instead of a single one
* Some rules to auto-remove pins (e.g. on uninstall, and maybe at some points if the packages are no longer available/installed).

Related to #476 and #2611 but does not resolve them yet

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2769)


florelis:
This thing's full of TODOs, many of them intentionally left for the future, but do tell me if there's any I should resolve for this PR.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,223 @@
...
+                return Pinning::Pin::CreatePinningPin({ packageId, sourceId });
+            }
+        }
+    }
+
+    void OpenPinningIndex(Execution::Context& context)
+    {
+        auto indexPath = Runtime::GetPathTo(Runtime::PathName::LocalState) / "pinning.db";
+
+        AICLI_LOG(CLI, Info, << "Openning pinning index");
```
Trenly, line 40:
```suggestion
        AICLI_LOG(CLI, Info, << "Opening pinning index");

```

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,223 @@
...
+        std::vector<Pinning::Pin> pins;
+
+        // TODO: We should support querying the multiple sources for a package, instead of just one
+        auto availableVersion = package->GetLatestAvailableVersion();
+
+        Pinning::PinKey pinKey{
+            availableVersion->GetProperty(PackageVersionProperty::Id).get(),
+            availableVersion->GetProperty(PackageVersionProperty::SourceIdentifier).get() };
+        auto pin = CreatePin(context, pinKey.PackageId, pinKey.SourceId);
+        AICLI_LOG(CLI, Info, << "Adding pin with type " << ToString(pin.GetType()) << " for package[" << pin.GetPackageId() << "] from source [" << pin.GetSourceId() << "]");
```
Trenly, line 92:
```suggestion
        AICLI_LOG(CLI, Info, << "Adding pin with type " << ToString(pin.GetType()) << " for package [" << pin.GetPackageId() << "] from source [" << pin.GetSourceId() << "]");

```

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "pch.h"
+#include "PinningIndex.h"
+#include "SQLiteStorageBase.h"
+#include "Schema/Pinning_1_0/PinningIndexInterface.h"
+
+namespace AppInstaller::Repository::Microsoft
+{
+    PinningIndex PinningIndex::CreateNew(const std::string& filePath, Schema::Version version)
+    {
+        AICLI_LOG(Repo, Info, << "Creating new Pinning Index [" << version << "] at '" << filePath << "'");
```
Trenly, line 12: Are the logs for the pinning index meant to be Repo, or should they be SQL?

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,223 @@
...
+namespace AppInstaller::CLI::Workflow
+{
+    namespace
+    {
+        // Creates a Pin appropriate for the context based on the arguments provided
+        Pinning::Pin CreatePin(Execution::Context& context, std::string_view packageId, std::string_view sourceId)
+        {
+            if (context.Args.Contains(Execution::Args::Type::BlockingPin))
+            {
+                return Pinning::Pin::CreateBlockingPin({ packageId, sourceId });
```
Trenly, line 23: Is there a reason we wouldn't want to record the currently installed version for a blocking pin or a pinning pin? Having the version would allow for heuristics on if pins are violated from installs outside of winget, and in the case of pinning pins could be updated when the user runs an upgrade with the proper argument. 

Especially as users may have side by side installs for a few packages, they may also want to have one of the versions pinned and another unpinned. It may be easier to record the version for now, even if nothing is done with it, to enable future functionality

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
florelis: I have to admit I didn't think of that, I just copied what the Portable index did. It looks like SQL is used exclusively at the level of the SQLiteWrapper, so I think Repo is fine here.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: There was no specific reason for me to not include it, it was just that I didn't need it at the moment.

I see that in the spec the sample tables for listing pinning and blocking pins don't show versions, @yao-msft do you remember if there was a specific reason for that?

I think if we needed the version, it would be a gating pin, wouldn't it? The other two aren't saying "keep it in this version", but rather "don't touch them". What future functionality are you picturing that would benefit from having the version? The only I can think of is showing some message when we detect a version change outside of winget, but I don't know what we would like to do with that.

Trenly: > I think if we needed the version, it would be a gating pin, wouldn't it?

The way I see it, a pinning pin is a gating pin that is set to a specific version instead of a version range, which would exclude it from `upgrade --all`, and `upgrade <package>` would update the package and the version of the pin. Which would make a blocking pin a gating pin that is set to a specific version instead of a version range, and is blocked from upgrade entirely. 

> What future functionality are you picturing that would benefit from having the version?

You got the main one that I am thinking of, which would be to show a message when there is a version change outside of winget.

Another one that I see would be around import/export - Presume for a moment that a user has `Example.Package` with a blocking pin. On their computer, they have version 1.0.0 installed. The user then exports their package list, which should include the information that there is a blocking pin on the package.  They go to a new computer, and manually install `Example.Package` version 1.2.0. When running the package import, the package install would be skipped due to it already existing, but the blocking pin would still be added (unless I'm mistaken here and the pin wouldn't be added). Having the version associated with the blocking pin would allow for clearer messaging during import as to why a package was skipped, or why a pin wasn't added.

Another potential use case - I have `Example.Package` installed on my machine, I have version 1.0.0 installed and I have version 1.1.0 installed. For some workflows, I specifically need version 1.0.0, but for other workflows I need the latest version. It would be beneficial to be able to add a blocking pin on version 1.0.0 to ensure that it is never upgraded, while still allowing version 1.1.0 to be upgraded to 1.1.1 or 1.2.0. 

Another use case could be a future pinning type - Locking. This future type might prevent both upgrade and uninstall. 

Directly mentioned in the spec is a future consideration for integration with the Dependencies flow. It could make it easier for dependencies to check the version of the blocking pin, rather than having to check the installed version if there is a blocking pin.

florelis: Updated to keep the version in all cases. For now I'm just keeping it as a string since I'm not using it yet.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    void OpenPinningIndex(Execution::Context& context)
+    {
+        auto indexPath = Runtime::GetPathTo(Runtime::PathName::LocalState) / "pinning.db";
+
+        AICLI_LOG(CLI, Info, << "Opening pinning index");
+
+        auto pinningIndex = std::filesystem::exists(indexPath) ?
+            PinningIndex::Open(indexPath.u8string(), SQLiteStorageBase::OpenDisposition::ReadWrite) :
+            PinningIndex::CreateNew(indexPath.u8string());
```
florelis, line 44: This needs to be in the RepositoryCore project so that we can search the pins when readding a package. I'll move it in a future PR when I add that.

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+#include "SQLiteWrapper.h"
+#include "Microsoft/Schema/IPinningIndex.h"
+#include "Microsoft/SQLiteStorageBase.h"
+#include "winget/Pin.h"
+#include <winget/ManagedFile.h>
+
+namespace AppInstaller::Repository::Microsoft
+{
+    struct PinningIndex : SQLiteStorageBase
```
florelis, line 10: I didn't remember I had already added pins from manifest to the tracking index for `RequireExplicitUpgrade`... I need to figure out how to deal with those two at the same time or how to merge them.

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -0,0 +1,120 @@
...
+        PinningIndex result{ filePath, version };
+
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(result.m_dbconn, "pinningIndex_createNew");
```
yao-msft, line 15:
```suggestion
        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(result.m_dbconn, "pinningindex_createnew");

```

nit: update all strings used by sqlite to lower case

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -0,0 +1,120 @@
...
+            SetLastWriteTime();
+            savepoint.Commit();
+        }
+
+        return result;
+    }
+
+    void PinningIndex::RemovePin(const Pinning::PinKey& pinKey)
+    {
+        AICLI_LOG(Repo, Verbose, << "Removing Pin for package [" << pinKey.PackageId << "] from source [" << pinKey.SourceId << "]");
```
yao-msft, line 65: nit: should this logging happen after we acquire the interface lock?

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    void OpenPinningIndex(Execution::Context& context)
+    {
+        auto indexPath = Runtime::GetPathTo(Runtime::PathName::LocalState) / "pinning.db";
+
+        AICLI_LOG(CLI, Info, << "Opening pinning index");
+
+        auto pinningIndex = std::filesystem::exists(indexPath) ?
```
yao-msft, line 42: We have issues before with settings where a folder with same name exists and code will crash when opening them. We should have better handling for corner cases here.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+            PinningIndex::Open(indexPath.u8string(), SQLiteStorageBase::OpenDisposition::ReadWrite) :
```
yao-msft, line 43: Do we need to open as read only for winget pin list case?

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+    // Required Args: None
+    // Inputs: PinningIndex, Package
+    // Outputs: None
+    void AddPin(Execution::Context& context)
+    {
+        auto package = context.Get<Execution::Data::Package>();
+        auto installedVersion = context.Get<Execution::Data::InstalledPackageVersion>();
+        if (!installedVersion)
+        {
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_NO_APPLICATIONS_FOUND);
```
yao-msft, line 86: Is there an error code for no "installed" package found. We may want to differentiate that user is trying to pin something not installed vs. package is not found at all.

Or maybe just limit the search behavior to SearchInstalled only so we don't even need this check.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+        {
+            AICLI_LOG(CLI, Warning, << "Pin does not exist");
+            context.Reporter.Warn() << Resource::String::PinDoesNotExist(pinKey.PackageId) << std::endl;
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_PIN_DOES_NOT_EXIST);
+        }
+
+        pinningIndex->RemovePin(pinKey);
+    }
+
+    // Report the pins in a table.
```
yao-msft, line 160: nit: Move these comments to header

and below 2 methods

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+                Resource::String::PinType,
+            });
+
+        for (const auto& pin : pins)
+        {
+            // TODO: Avoid these conversions to string
+            table.OutputLine({
+                pin.GetPackageId(),
+                std::string{ pin.GetSourceId() },
+                std::string{ pin.GetVersionString() },
```
yao-msft, line 188: This may be another reason to not mix installed version and gated version, as this will be confusing in report table.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+    // Outputs: None
+    void CrossReferencePinsWithSource(Execution::Context& context)
+    {
+        const auto& pins = context.Get<Execution::Data::Pins>();
+        const auto& source = context.Get<Execution::Data::Source>();
+
+        std::vector<Pinning::Pin> matchingPins;
+        std::copy_if(pins.begin(), pins.end(), std::back_inserter(matchingPins), [&](Pinning::Pin pin) {
+            // TODO: Filter to source
+            SearchRequest searchRequest;
```
yao-msft, line 237: That seems expensive to do a search for each pin, shall we do this when actually needed during upgrade?
meaning do we need to support listing pins filtered by source?

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+            context.Reporter.Info() << Resource::String::PinResettingAll << std::endl;
+
+            if (context.Get<Execution::Data::Pins>().empty())
+            {
+                context.Reporter.Info() << Resource::String::PinNoPinsExist << std::endl;
+            }
+            else
+            {
+                auto pinningIndex = context.Get<Execution::Data::PinningIndex>();
+                pinningIndex->ResetAllPins();
```
yao-msft, line 214: From looking at PinCommand implementation, seems like we intend to support resetting filtered by source, but this ResetAllPins() just resets everything.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,247 @@
...
+            }
```
yao-msft, line 215: I think we should also report to user "success" after resetting completed.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1605,4 +1605,35 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Select installed package scope filter (user or machine)</value>
     <comment>This argument allows the user to select installed packages for just the user or for the entire machine.</comment>
   </data>
-</root>
+  <data name="PinAdded" xml:space="preserve">
+    <value>Pin added successfully</value>
```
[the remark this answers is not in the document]
florelis: We do need to do that. I know they have a term dictionary for that kind of things that as I understand @AmelBawa-msft was going to add our terms to. Should I add something here in the meantime?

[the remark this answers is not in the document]
yao-msft: It's not urgent so if we can add the terms to term dictionary any time before release I think it's good.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
Trenly: > That's also the reason "pinning from installed source" is put in the future consideration section since this depends on underlying Composite package improvement.
> 
I see. It wasn't clear what exactly 'installed source' meant, I think I understand now

> In the spec time, I was thinking rather than recording something unknown or uncertain, we would leave it until we have better underlying framework code support.
>
If multiple results are returned for an installed version, the field could be left empty; If a single version is installed, then it could be populated? Unless I'm misunderstanding how the composite packages work

src/AppInstallerCLICore/Workflows/PinFlow.cpp
Trenly: If the search behavior is limited, then a user wouldn't be able to add a pin without having the package installed.  The only use case I can think of for pinning a package that isn't installed would be a gating pin so that `winget install` doesn't exceed the gate. But, this can be worked around using `-v` and then adding the pin after install. 

Limiting to search only installed would be good

src/AppInstallerCLICore/Workflows/PinFlow.cpp
yao-msft: > If multiple results are returned for an installed version, the field could be left empty; If a single version is installed, then it could be populated? Unless I'm misunderstanding how the composite packages work

That's exactly the issue we have with current CompositePackage implementation. The CompositePackage returned by the search result always return at most 1 installed version, and we don't know if there's only 1 installed version vs. there are multiple installed versions. And I think we'd better deal this after the underlying CompositePackage is improved.

src/AppInstallerCommonCore/Public/AppInstallerRuntime.h
```diff
@@ -62,6 +62,11 @@ namespace AppInstaller::Runtime
         PortableLinksUserLocation,
         // The location where symlinks to portable packages are stored under machine scope.
         PortableLinksMachineLocation,
+        // The location of the user settings json file.
+        UserSettingsFileLocation,
+        // The location of the user settings json file, anonymized using environment variables.
+        UserSettingsFileLocationForDisplay,
+        PinningIndex,
```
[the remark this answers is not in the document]
florelis: Huh, I just realized that I never ended up using this. I probably should use it to resolve the other issue you mentioned with edge cases with the path already existing.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: Good catch :)
Added filtering by source

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: Changed to remove the version from non-gating pins

src/AppInstallerCommonCore/Public/AppInstallerVersions.h
```diff
@@ -169,6 +169,21 @@ namespace AppInstaller::Utility
...
+        // TODO
+        // For now, using dummy implementation that just holds a string
+        GatedVersion() {}
+        GatedVersion(std::string_view s) : m_tmp(s) {}
+        std::string ToString() const { return m_tmp; }
+
+        bool operator==(const GatedVersion& other) const { return m_tmp == other.m_tmp; }
+
+    private:
+        std::string m_tmp;
```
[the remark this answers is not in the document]
florelis: It's not. In the follow-up PR I have already replaced it by something else.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_0/PinTable.cpp
```diff
@@ -0,0 +1,183 @@
...
+        SQLite::Builder::StatementBuilder builder;
+        builder.Update(s_PinTable_Table_Name).Set()
+            .Column(s_PinTable_PackageId_Column).Equals((std::string_view)pin.GetPackageId())
+            .Column(s_PinTable_SourceId_Column).Equals(pin.GetSourceId())
+            .Column(s_PinTable_Type_Column).Equals(pin.GetType())
+            .Column(s_PinTable_Version_Column).Equals(pin.GetVersionString())
+            .Where(SQLite::RowIDName).Equals(pinId);
+
+        builder.Execute(connection);
+        return connection.GetChanges() != 0;
```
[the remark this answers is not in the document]
florelis: I'm not entirely sure, but it at least returns not 0 as I use that in a test. GetChanges() is supposed to return the number of rows affected, so it should be 1.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: I didn't remember that it is already filtered to only installed as the default search behavior for composite source. Removed this check.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: I removed the searches from the case when we are actually resetting the pins (--force). But when we don't have --force, I'm listing the pins and I want to not show pins that are no longer applicable (e.g. because the app was uninstalled or because the source no longer provides the package). I'm assuming there won't be that many pins, but I certainly didn't consider that it would be expensive for REST sources.

src/AppInstallerCommonCore/Public/AppInstallerVersions.h
```diff
@@ -169,6 +169,21 @@ namespace AppInstaller::Utility
         bool m_isEmpty = false;
     };
 
+    // A range of versions indicated by a version and optionally a wildcard at the end.
+    struct GatedVersion
```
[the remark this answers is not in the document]
florelis: In the actual implementation what I'm doing is just storing the version as a member and having a function to check if a version matches a gated version by iterating their parts. I thought of making the gating version a version range, but that has an exclusive upper bound unlike the gating version.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: Changed to read only for list and reset without --force

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: Do you have a pointer to how we handle that for other files?

yao-msft: https://github.com/microsoft/winget-cli/pull/2767 is where Ruben fixes issues when settings.json is a directory, but mainly we try to catch exceptions, remove bad one and recrate an empty new one.

florelis: I changed it to have a softer error if we cannot open the database instead of crashing. I'll also update the follow-up PR so that if we cannot open it, we just don't consider pins during upgrade/other commands.

I don't think this will cause the situation of making a directory in the location of the file because I'm not getting the full path with Runtime::GetPathTo, rather, I'm getting the path to LocalState and adding to it.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,231 @@
...
+            context.Reporter.Info() << Resource::String::PinNoPinsExist << std::endl;
+            return;
+        }
+
+        // TODO: Use package and source names
+        Execution::TableOutput<4> table(context.Reporter,
+            {
+                Resource::String::SearchId,
+                Resource::String::SearchSource,
+                Resource::String::SearchVersion,
```
yao-msft, line 165: >Resource::String::SearchVersion, [](http://example.com/codeflow?start=16&length=32)

Do you think the column name "Version" should be more specific to indicate it's a gated version?

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    std::shared_ptr<PinningIndex> PinningIndex::OpenOrCreateDefault(OpenDisposition openDisposition)
+    {
+        auto indexPath = Runtime::GetPathTo(Runtime::PathName::LocalState) / "pinning.db";
+        AICLI_LOG(CLI, Info, << "Opening pinning index");
+
+        try
+        {
+            if (std::filesystem::exists(indexPath))
```
yao-msft, line 36: >if (std::filesystem::exists(indexPath)) [](http://example.com/codeflow?start=12&length=39)

I think it may be better if we do 
if (exists)
{
   if (isfile)
   {
       try open and return
   }
   
    delete and create new
}
else
{
     create new
}

yao-msft (approved):
:shipit:

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: I don't think we need it. I believe a user would easily understand what we mean if the header is "Version" and the value is something like "1.2.*". I also didn't actually use "gated version" in any of the resource strings, just in logs and names in the code, so it's not something they'd have seen elsewhere.

PR #2772, opened 2022-12-15
Add additional admin setting to match security related group policies

yao-msft:
Added corresponding admin settings for InstallerHashOverride and LocalArchivemalwareScanOverride
Also added tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2772)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
@@ -5,6 +5,7 @@
 #include "winget/Archive.h"
 #include "winget/Filesystem.h"
 #include "PortableFlow.h"
+#include <winget/AdminSettings.h>
 
 using namespace AppInstaller::Manifest;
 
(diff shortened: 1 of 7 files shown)


src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -5,6 +5,7 @@
 #include "winget/Archive.h"
 #include "winget/Filesystem.h"
```
ryfu-msft, line 6: Nit: Should these also use <>? I can fix when I create the PR for removing zip experimental feature

yao-msft: yes

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -27,7 +28,8 @@ namespace AppInstaller::CLI::Workflow
             }
             else
             {
-                if (context.Args.Contains(Execution::Args::Type::IgnoreLocalArchiveMalwareScan))
+                if (context.Args.Contains(Execution::Args::Type::IgnoreLocalArchiveMalwareScan) &&
+                    Settings::IsAdminSettingEnabled(Settings::AdminSetting::LocalArchiveMalwareScanOverride))
```
ryfu-msft, line 32: >         Settings::IsAdminSettingEnabled(Settings::AdminSetting::LocalArchiveMalwareScanOverride)) [](http://example.com/codeflow?start=11&length=98)

Should there be a separate elseif block for when someone includes --ignore-local-archive-malware-scan but does not have the admin setting for LocalArchiveMalwareScanOverride enabled? If so, should there be a different error message specific to admin settings?

ryfu-msft (approved):

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
yao-msft: This message is handled in command invocation.

PR #2773, opened 2022-12-15
Split tests in `Workflow.cpp` into multiple files

florelis:
I've always disliked how big this file is, having tests for all flows. I didn't want to add even more to it for pinning, so I'm splitting it up :)

* Created multiple files for each workflow and moved the appropriate tests there based on the tags they had. For example, moved tests tagged `[ExportFlow]` to `ExportFlow.cpp`. Tests that were the only ones with that tag, I left in `Workflow.cpp`
* Created `WorkflowCommon.h` and `WorkflowCommon.cpp` with just the shared test source/context definition, and common task overrides.
* Cleaned up the `#include`s and `using`s

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2773)


florelis:
It would be so nice if git could see that I'm just moving code around and it could apply the changes automatically instead of having to do it by hand :(

If we want to take this, can we merge it before I get more conflicts?

The test failures from last run seem to have been a consequence of not having #2767, which only hit the tests once they were executed in a different order.

src/AppInstallerCLITests/WorkflowCommon.cpp
```diff
@@ -0,0 +1,627 @@
...
+            result.Matches.emplace_back(
+                ResultMatch(
+                    TestPackage::Make(std::vector<Manifest>{ manifest2 }, shared_from_this()),
+                    PackageMatchFilter(PackageMatchField::Id, MatchType::Exact, "TestQueryReturnTwo")));
+        }
+
+        return result;
+    }
+
+    SearchResult WorkflowTestCompositeSource::Search(const SearchRequest& request) const
```
florelis, line 62: I also want to split this function, I may do it later in another PR.

yao-msft (approved):

florelis:
The actual reason for the test failures was fun to find. There were some test hook overrides that were being set and never cleared when the test using them was done. So subsequent tests would still get the behavior from the override. Moving the tests around changed the order and revealed that. Since it was about a test hook override set at runtime, it would only show if executing the two relevant tests in the same invocation and in a specific order. Not having the fix for export settings was just a red herring. (I'll be mad if after this the tests still fail...)

PR #2774, opened 2022-12-16
Make zip feature stable

ryfu-msft:
Removes all code related to the zip install feature as experimental in preparation for a stable release.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2774)


src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -1139,7 +1139,6 @@ TEST_CASE("InstallFlow_Zip_Exe", "[InstallFlow][workflow]")
 {
     TestCommon::TempFile installResultPath("TestExeInstalled.txt");
     TestCommon::TestUserSettings testSettings;
```
yao-msft, line 1141: This testUserSettings will not be used after removing, so please clean up across this file

yao-msft (approved):

PR #2776, opened 2022-12-16
Implement WinGetUserSettings cmdlets

msftrubengu:
This PR implements 3 cmdlets:
- Get-WinGetUserSettings - returns the contents of the settings.json file. If the file is not valid json, throws UserSettingsReadException.
- Set-WinGetUserSettings - writes the serialized input hashtable into the settings.json. If the Merge option is set, merges the current settings.json with the input given, otherwise overwrites the current settings with input settings. If Merge is used and settings.json is not valid json throws UserSettingsReadException. 
- Test-WinGetUserSettings - compares the given input settings with the current settings.json. If the IgnoreNotSet option is set, makes a partial comparison where it ignores properties not set in the input that are present in settings.json, otherwise make a full comparison. 

The settings file is read via `winget settings export`, so these commands require a newer version of winget. In the near future, I'll add a new attribute in each cmdlet that specify the minimum winget version that is required.

To be testable, the cmdlets follow the similar patter to pick which version of winget to use via USE_PROD_CLSIDS. I'll change the name to something more general in a future PR. We also need to modify the crescendo generated functions to point to wingetdev without it being that intrusive.

I also tried to move the PowerShell E2E to use PowerShell host to make better tests, but I got some issues and I kind of gave up for now to keep the code flowing. The user settings cmdlets tests use PowerShell host though. However, the Microsoft.WinGet.Client project cannot be references by AppInstallerCLIE2ETests because PowerShell host will complain while trying to load the module. My guess is that there's some conflict loading the Microsoft.WinGet.Client.dll

Also, made more general improvements to Microsoft.WinGet.Client.

Related to #2697


src/PowerShell/Microsoft.WinGet.Client/Commands/GetUserSettingsCommand.cs
```diff
@@ -0,0 +1,29 @@
...
+    [Cmdlet(VerbsCommon.Get, Constants.WinGetNouns.UserSettings)]
+    [OutputType(typeof(string))]
+    public sealed class GetUserSettingsCommand : BaseUserSettingsCommand
+    {
+        /// <summary>
+        /// Writes the settings file contents.
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            this.WriteObject(this.GetLocalSettingsFileContents());
```
JohnMcPMS, line 26: I would strongly prefer that the output type here match the input type for `Set`.  And `HashTable` is preferable over `string`.  Can we use [JsonObject.ConvertFromJson](https://learn.microsoft.com/en-us/dotnet/api/microsoft.powershell.commands.jsonobject.convertfromjson?view=powershellsdk-7.3.0) which is what `ConvertFrom-Json` uses to do it?  I looked at its implementation and it does the looping and converting necessary.

JohnMcPMS (approved):

PR #2782, opened 2022-12-19
Refactor test workflow source search

florelis:
Another refactor that nobody asked for, for something that has been bothering me for a long time and I couldn't ignore while working on something else... (I swear I am also working on the stuff I'm supposed to!)

This change breaks up the huge `WorkflowTestCompositeSource::Search()` function, which was a long series of "if query is 'package' then add package to search results" (and which had a fun compilation warning about using too much space in the stack). I'm doing this by creating objects containing the matching query string and the modifications made to the search results in case of a match. Then, the search function can just iterate over these objects and modify the results when appropriate.

On each test, we now declare which results we want to get and we can use the query strings from these objects instead of duplicating the strings on each place they are needed.

I couldn't come up with a good name for these objects, so I'm expecting feedback on that if this change is something we want.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2782)


yao-msft (approved):
:shipit:

PR #2784, opened 2022-12-19
Suppress CA1416 warning #2

ryfu-msft:
Suppresses a build error relating to a windows-only api.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2784)


florelis (approved):
I wonder if it is possible to set some property project-wide to mark everything as Windows-only instead of suppressing these warnings as they pop up.

PR #2786, opened 2022-12-20
Return localization independent string for help links

AmelBawa-msft:
- Updated `HelpLink` return type to a localization independent string.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2786)


florelis (approved):
Is it possible to use `LocIndView` instead of `LocIndString`?

florelis (approved):

PR #2792, opened 2022-12-21
Don't treat the COM subcommands as distinct for telemetry

JohnMcPMS:
## Issue
Calls to COM install/upgrade were logging the distinct steps (download/install) as separate commands under the same activity.  This was leading to an incorrect success metric due to the method of processing.

## Change
This change treats the `OrchestratorQueueItem` as the command for telemetry purposes.  This aligns with the `winget.exe` method of counting success/failure.  It also moves to use the same command names as `winget.exe`, making it easier to join/slice the data.

## Validation
Using an ETW event tracer, manually verified the current bad state.  Then verified that the change produces a single start/stop pair under the same activity for an install.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2792)

### DIFF
--- a/src/AppInstallerCLICore/ContextOrchestrator.cpp
+++ b/src/AppInstallerCLICore/ContextOrchestrator.cpp
@@ -91,6 +91,9 @@ namespace AppInstaller::CLI::Execution
         if (item->IsOnFirstCommand())
         {
             THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INSTALL_ALREADY_RUNNING), FindById(item->GetId()));
+
+            // Log the beginning of the item
+            item->GetContext().GetThreadGlobals().GetTelemetryLogger().LogCommand(item->GetItemCommandName());
         }
 
         std::string commandQueueName{ GetCommandQueueName(item->GetNextCommand().Name()) };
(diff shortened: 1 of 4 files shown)


src/AppInstallerCLICore/ContextOrchestrator.cpp
```diff
@@ -253,12 +256,11 @@ namespace AppInstaller::CLI::Execution
...
 
-                item->GetContext().GetThreadGlobals().GetTelemetryLogger().LogCommand(command->FullName());
                 command->ValidateArguments(item->GetContext().Args);
 
                 item->GetContext().EnableCtrlHandler();
 
-                terminationHR = ::AppInstaller::CLI::Execute(item->GetContext(), command);
+                ::AppInstaller::CLI::ExecuteWithoutLoggingSuccess(item->GetContext(), command.get());
             }
             WINGET_CATCH_STORE(terminationHR, APPINSTALLER_CLI_ERROR_COMMAND_FAILED);
```
florelis, line 265: It took me a while to understand why you were removing the assignment to `terminationHR` a couple of lines above, why that doesn't affect the `if (FAILED(terminationHR))` below, and why the next conditional uses `item->GetContext().GetTerminationHR()` instead of `terminationHR`.
Can we add a comment, or rename `terminationHR` to something like `caughtHResult` so that it's clearer in the future?

florelis (approved):
Why is the branch called `actwovity`? I can't help but add owo in my mind when I read it...

JohnMcPMS:
> Why is the branch called `actwovity`? I can't help but add owo in my mind when I read it...

There were two activities, and I didn't need a particularly helpful name since this was a short fix.

florelis (approved):

Issue #2793, opened 2022-12-21
Winget v1.4.3531 passes `/SILENT` to Inno installers when using `--silent`, should use `/VERYSILENT`


o-l-a-v:
### Brief description of your issue

Winget [v1.4.3531](https://github.com/microsoft/winget-cli/releases/tag/v1.4.3531) passes `/SILENT` to Inno installers instead of `/VERYSILENT` when using `--silent`.

This behavior is not present in the latest stable from Microsoft Store, v1.3.2691, which passes `/VERYSILENT` as it should.

#### Related issues that got me here
* https://github.com/JanDeDobbeleer/oh-my-posh/issues/3233
* https://github.com/microsoft/winget-cli/issues/2008

#### Resources
* https://github.com/microsoft/winget-pkgs/tree/master/manifests/j/JanDeDobbeleer/OhMyPosh
* https://jrsoftware.org/ishelp/index.php?topic=setupcmdline

### Steps to reproduce

Command:

* `winget install --exact --silent --scope user --id JanDeDobbeleer.OhMyPosh`

v1.3.2691 / latest in App Installer from Microsoft Store:

* No progress is shown, winget logs says `/VERYSILENT` was passed.

[v1.4.3531](https://github.com/microsoft/winget-cli/releases/tag/v1.4.3531)

* Progress is shown, logs says `/SILENT` was passed.

### Expected behavior

`/VERYSILENT` should be used to be `--silent`.

### Actual behavior

`/SILENT` was used, which is not `--silent`.

#### v1.3.2691 installed with Microsoft Store app App Installer

```shell
Windows Package Manager v1.3.2691
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19045.2364
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.18.2691.0
```

#### v1.4.3531 extracted from msix with 7-zip

```shell
Windows Package Manager v1.4.0
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.19045.2364
System Architecture: X64
```

florelis:
I just tried doing that with the same v1.4.3531 and it did use `/VERYSILENT`.

Log snippet:
```
2022-12-21 15:10:43.160 [CORE] WinGet, version [1.4.3531], activity [{175C57E7-D866-4BAF-83D3-80629FD20E3A}]
...
[...]

Issue #2680, opened 2022-11-08
upgrade '--all' and '--accept-package-agreements' flags should work together


MaZe3D:
### Description of the new feature / enhancement

I tried to use the command `winget upgrade --all` command but after the package repos were successfully downloaded, it aborted because the package-agreements werent accepted. But I was not asked either if I tried to upgrade all. So I tried to append the `--accept-package-agreements` flag, but assumingly this won't work in combination with `--all`.

The annoying part is that it abborts the whole proccess of upgrading all packages and wont continue. So I have to manually upgrade all packages with included package-agreements in order to accept it, and then use --all to upgrade all others.

Without using both flags in combination it makes the whole purpurse of --all obsolete because it wont run no matter if you want or wont want to accept the package agreements

### Proposed technical implementation details

I want the command `winget upgrade --all --accept-package-agreements` to work and upgrade all packages and automatically accept the agreemets, if nessecary.

awecomp:
Just had the same issue, one program wouldn't upgrade with "Package agreements were not agreed to. Operation cancelled."
"The arguments provided can only be used with a query." with "winget upgrade --accept-package-agreements --all"
Ended up just uninstalling the software that was causing issues and upgraded the rest.

eabase:
Yeah, this is not good for automation. Also make sure that `--accept-source-agreements` is also compatible.

joshuatrevissmith:
I have this problem when I install using Windows Store source, because I am prompted to accept the agreement upon install and every update. 

`winget --all` will fail with:
```
Package agreements were not agreed to. Operation cancelled.
```

The workaround has been to first use winget or Store app to update the store sourced package(s), then `winget --all` will update the remaining winget sourced packages that I have.

gellenburg:
Just adding that this definitely needs to be fixed:

`winget upgrade --all --accept-package-agreements --accept-source-agreements`

What's the point in having an `upgrade --all` flag if the flag won't even work?

If I can do an `rpm -Uvh` or `apt-get -y upgrade` on Linux, I should be able to do an `winget upgrade --all --accept-package-agreements --accept-source-agreements` as well.

soulhax:
I will bump this as well. This must be fixed ASAP.

eabase:
This is a rather serious UX issue, with what appears to be an easy fix. 
Would be great to get some ETA update/feedback from the MS Team.

rootomatic:
[...]

PR #2794, opened 2022-12-22
Copy install behavior flags on upgrade --all

### LINKED ISSUE #2008: winget (v1.2) upgrade --all --silent only installs the first package silently

```winget upgrade --all --silent``` only installs the first package silently. The rest shows progress.

I had following upgrades available when running ```winget upgrade```.

```
Name                                               Id                                Version       Available     Source
Microsoft Teams                                    Microsoft.Teams                   1.5.00.2164   1.5.00.4689   winget
Microsoft Visual C++ 2015-2022 Redistributable (x… Microsoft.VC++2015-2022Redist-x86 14.31.30919.0 14.31.31103.0 winget
Microsoft Teams                                    Microsoft.Teams                   1.4.0.19572   1.5.00.4689   winget
Microsoft Visual C++ 2015-2022 Redistributable (x… Microsoft.VC++2015-2022Redist-x64 14.31.30919.0 14.31.31103.0 winget
Eclipse Temurin JDK with Hotspot 11                EclipseAdoptium.Temurin.11        11.0.12.7     11.0.14.101   winget
5 upgrades available.
```

First Teams installed silently, but neither of the VCRedist or JDK did.

### Steps to reproduce

Have more than one package to upgrade, then:

* ```winget upgrade --all --silent```

### Expected behavior

All packages should install silently.

### Actual behavior

Only the first package installs silently.

### Environment

```shell
Windows Package Manager v1.2.10271

Windows: Windows.Desktop v10.0.19044.1586
Package: Microsoft.DesktopAppInstaller v1.17.10271.0
```

florelis:
*Issue:*
When we install multiple packages in the same execution (`upgrade --all` and `import`), we create a sub-context to install each individual package but this sub-context does not respect the parent's install behavior flags, like `--silent`.

*Change:*
Added a step to the creation of a sub-context to copy the relevant flags from the parent context, so that the same behavior is used for all packages. Added a test to verify the new behavior.

---

This came from a user running `winget upgrade --all --silent` and expecting all the upgrades to be done silently. We need to either respect those flags (this PR) or change the argument validation to forbid that combination.

I figure the most contentious issue with this PR will be which args we should forward. I added basically everything that made sense to me, but we can trim the list as much as needed.

Fixes #2008 and #2793

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2794)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionContext.cpp
+++ b/src/AppInstallerCLICore/ExecutionContext.cpp
@@ -135,12 +135,13 @@ namespace AppInstaller::CLI::Execution
     void Context::CopyArgsToSubContext(Context* subContext)
     {
         // Copy over install behavior flags from the parent context.
-        // Arguments not copied: Override, Log
         for (auto flag : {
             Args::Type::Interactive,
             Args::Type::Silent,
             Args::Type::HashOverride,
             Args::Type::IgnoreLocalArchiveMalwareScan,
+            Args::Type::NoUpgrade,
+            Args::Type::Force,
             })
         {
             if (Args.Contains(flag))
@@ -151,7 +152,6 @@ namespace AppInstaller::CLI::Execution
 
         for (auto arg : {
             Args::Type::Locale,
-            Args::Type::InstallLocation,
             Args::Type::InstallScope,
             Args::Type::InstallArchitecture,
             })
--- a/src/AppInstallerCLICore/ExecutionContext.cpp
+++ b/src/AppInstallerCLICore/ExecutionContext.cpp
@@ -128,7 +128,39 @@ namespace AppInstaller::CLI::Execution
         {
             clone->EnableCtrlHandler();
         }
+        CopyArgsToSubContext(clone.get());
         return clone;
+    }
+
+    void Context::CopyArgsToSubContext(Context* subContext)
+    {
+        // Copy over install behavior flags from the parent context.
+        // Arguments not copied: Override, Log
+        for (auto flag : {
+            Args::Type::Interactive,
+            Args::Type::Silent,
+            Args::Type::HashOverride,
+            Args::Type::IgnoreLocalArchiveMalwareScan,
+            })
+        {
+            if (Args.Contains(flag))
+            {
+                subContext->Args.AddArg(flag);
+            }
+        }
+
+        for (auto arg : {
+            Args::Type::Locale,
+            Args::Type::InstallLocation,
+            Args::Type::InstallScope,
+            Args::Type::InstallArchitecture,
+            })
+        {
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -128,9 +128,41 @@ namespace AppInstaller::CLI::Execution
...
         }
+        CopyArgsToSubContext(clone.get());
         return clone;
     }
 
+    void Context::CopyArgsToSubContext(Context* subContext)
+    {
+        // Copy over install behavior flags from the parent context.
+        // Arguments not copied: Override, Log
+        for (auto flag : {
```
yao-msft, line 139: Do we need to copy over some other behavior args like DisableInteractivity, Force, AcceptPackageAgreements, etc? I feel this needs a more thorough revist.

I also feel this selective copying is hard to maintain but I cannot find a better way at the moment.

florelis: DisableInteractivity and AcceptPackageAgreements don't need to be copied over because those are handled at the top level. Unless we have nested contexts...

I've been thinking we can add some tags/flags about what an argument is for. Like, IsInstallBehavior, IsGlobalBehavior, IsQueryArgument. Then we could do this copying based on that, and we could also use it for argument validation. For example, Manifest cannot be used with query/source arguments, and installer specific args (like override) cannot be used when dealing with multiple packages (update --all)

yao-msft: I just went over all the args, I suggest to copy over Interactive, Silent, Locale, InstallScope, InstallArchitecture, HashOverride, IgnoreLocalArchiveMalwareScan, NoUpgrade,  Force

Regarding InstallLocation, I saw in PromptForMultiplePackages we set them all to same value already, but I feel it will be more likely to cause conflicts for packages if installerss write to the InstallLocation root. Perhaps we should add a todo somewhere to revisit this. And this copy is probably not needed.

yao-msft (approved):
Please fix the tests before merge

src/AppInstallerCLICore/ExecutionContext.cpp
Trenly: @florelis / @yao-msft; Regarding this comment -
> DisableInteractivity and AcceptPackageAgreements don't need to be copied over because those are handled at the top level. Unless we have nested contexts...

I just stumbled across this issue:
- #2680

Perhaps the "top-level" flags *do* need to be copied over?

Issue #2559, opened 2022-09-29
RequireExplicitUpgrade not requiring explicit upgrade


Trenly:
### Brief description of your issue

When installing Microsoft Teams, it should be pinned through `RequireExplicitUpgrade` as implemented in #1795 and released in [v1.4.2161-preview](https://github.com/microsoft/winget-cli/releases/tag/v1.4.2161-preview)

Installing teams and changing the version to trigger an upgrade reveals that the package is not requiring explicit upgrade.

https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/Microsoft/Teams/1.5.00.21668/Microsoft.Teams.installer.yaml#L8

### Steps to reproduce

* Uninstall Microsoft Teams
* Reinstall Microsoft Teams using `winget install Microsoft.Teams --scope=machine --source=winget`
* Modify the ARP entry of `Teams Machine-Wide Installer` to have DisplayVersion be a lower value, so that an upgrade would be triggered. Be sure to only change the DisplayVersion
* Run `winget upgrade --all`

### Expected behavior

Teams to be pinned by default and not upgrade without explicit upgrade - as it was installed with a manifest that stated explicit upgrade was required

### Actual behavior

```
PS D:\Git\winget-pkgs> winget uninstall Microsoft.Teams
Found Microsoft Teams [Microsoft.Teams]
Starting package uninstall...
Successfully uninstalled
PS D:\Git\winget-pkgs> winget install Microsoft.Teams --scope=machine -s=winget
Found Microsoft Teams [Microsoft.Teams] Version 1.5.00.21668
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://statics.teams.cdn.office.net/production-windows-x64/1.5.00.21668/Teams_windows_x64.msi
  ██████████████████████████████   124 MB /  124 MB
Successfully verified installer hash
Starting package install...
Successfully installed
PS D:\Git\winget-pkgs> winget list teams
Name                         Id              Version
--------------------------------------------------------
Teams Machine-Wide Installer Microsoft.Teams 1.5.0.21668

// Registry Modified Manually

PS D:\Git\winget-pkgs> winget list teams
Name                         Id              Version     Available
---------------------------------------------------------------------
Teams Machine-Wide Installer Microsoft.Teams 1.5.0.21667 1.5.00.21668
PS D:\Git\winget-pkgs> winget upgrade --all
[...]

PR #2813, opened 2023-01-03
Implement package pinning

### LINKED ISSUE #476: Pin a package

florelis:
This PR builds on top of #2769. I can close that PR in favor of this one, or we can merge that one first to make the changes smaller and easier to review :)

Changes in this PR (not in #2769):
* Modified the `CompositePackage` to be able to hold more than one available package, each from a different source. The `GetAvailableVersionKeys()` function now can return versions from multiple sources, so the source field of `PackageVersionKey`  is more relevant now. If the same version is available from multiple sources, they are returned in the order the available packages were added, which matches the previous behavior of just adding the first available package found.
* Added a `PinBehavior` enum which is passed as an argument to an `IPackage`'s functions for getting available versions. Only the `CompositePackage` uses it (as it needs both an installed and an available package) and it is used to determine whether we ignore pinned versions when listing the available versions or not. For example, for `list` we don't care about pins, for `upgrade` we do consider them, and for `upgrade --include-pinned` we care about them unless they have Pinning type.
  + Added this new argument all through the codebase. I'm open to making it have a default value, but I wanted to highlight all the places it is used to be sure I made the appropriate choice on each place.
* Updated existing tests to new behaviors. For example, `CompositeSource` tests now need to ensure that the test packages have a reference to their source as it is used to determine if they are pinned.
* There was already some work for pinning done for the `RequiresExplicitUpgrade` manifest field. This change is mostly independent of that one, as in my mind the pinning from `RequiresExplicitUpgrade` is a property of the installed package, whereas user-defined pins are properties of the available packages from each source, so they are handled at different places.

Closes #476
Related to #2611

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2813)


florelis:
Here's a little screenshot of this thing in action:

image

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
```diff
@@ -190,19 +196,18 @@ namespace AppInstaller::CLI::Workflow
...
-            if (pinnedState != PackagePinnedState::NotPinned)
+            auto pinnedState = ConvertToPinTypeEnum(installedMetadata[PackageVersionMetadata::PinnedState]);
+            if (pinnedState == PinType::PinnedByManifest)
             {
+                // Note that for packages pinned by the manifest
+                // this does not consider whether the update to be installed has
+                // RequireExplicitUpgrade. While this has the downside of not working with
+                // packages installed from another source, it ensures consistency with the
+                // list of available updates (there we don't have the selected installer)
+                // and at most we will update each package like this once.
```
Trenly, line 210: Does this PR also resolve https://github.com/microsoft/winget-cli/issues/2559#issuecomment-1262712788 ?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1605,4 +1597,39 @@ Please specify one of them using the --source option to proceed.</value>
...
+  </data>
+  <data name="PinAlreadyExists" xml:space="preserve">
+    <value>The pin already exists</value>
+  </data>
+  <data name="PinExistsOverwriting" xml:space="preserve">
+    <value>A pin already exists for this package. Overwriting due to the --force argument.</value>
+    <comment>{Locked="--force"}</comment>
+  </data>
+  <data name="PinExistsUseForceArg" xml:space="preserve">
+    <value>A pin already exists for this package. Use the --force argument to overwrite it.</value>
```
Trenly, line 1611: Is there a reason not to use the `{0}` and the package name here? Just curious, since the `PinDoesNotExist` message does use it

src/AppInstallerCommonCore/Errors.cpp
```diff
@@ -208,6 +208,12 @@ namespace AppInstaller
                 return "Archive malware scan failed.";
             case APPINSTALLER_CLI_ERROR_PACKAGE_ALREADY_INSTALLED:
                 return "Found at least one version of the package installed.";
+            case APPINSTALLER_CLI_ERROR_PIN_ALREADY_EXISTS:
+                return "A pin already exists for the package.";
+            case APPINSTALLER_CLI_ERROR_PIN_DOES_NOT_EXIST:
+                return "There is no pin for the package.";
+            case APPINSTALLER_CLI_ERROR_PACKAGE_HAS_BLOCKING_PIN:
+                return "The package has a blocking pin that prevents from updating.";
```
Trenly, line 216:
```suggestion
                return "The package has a blocking pin that prevents upgrade.";
```

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -0,0 +1,267 @@
...
+        context.Add<Execution::Data::PinningIndex>(std::make_shared<PinningIndex>(std::move(pinningIndex)));
+    }
+
+    void GetAllPins(Execution::Context& context)
+    {
+        AICLI_LOG(CLI, Info, << "Getting all existing pins");
+        context.Add<Execution::Data::Pins>(context.Get<Execution::Data::PinningIndex>()->GetAllPins());
+    }
+
+    void SearchPin(Execution::Context& context)
```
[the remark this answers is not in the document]
florelis: I think I'll remove the comments from the other workflow tasks instead. From what I see, we generally keep the comments in the headers. The others have comments only because I copied the definitions from the header and I forgot to remove them.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
florelis: No reason. I'll change that. It's part of what I meant with "Make this pretty. Lots of places could use better or more messages." in the PR description :)

src/AppInstallerCLICore/Workflows/PinFlow.cpp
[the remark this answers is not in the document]
Trenly: Works for me!

src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
florelis: No, it does not. If we make any changes for that, I'd rather have them in a separate PR as this one is already huge.

Trenly: I understand :)
Just not sure where the original bug actually came in, so was hoping it might have been cauggt in the mix. Thank you for the clarification!

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -243,8 +233,16 @@ namespace AppInstaller::Repository
 
         // The channel.
         Utility::NormalizedString Channel;
+
+        bool operator<(const PackageVersionKey& other) const
+        {
+            // Sort using only the version and channel.
+            // The order for the sources depends on the context.
+            return Utility::VersionAndChannel({ Version }, { Channel }) < Utility::VersionAndChannel({ other.Version }, { other.Channel });
```
yao-msft, line 241: This code is assuming versions across sources use same semantic. I remember seeing packages that use different versions for Store source and other sources. Should we not allow version comparison if it's from different sources?

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -26,9 +26,24 @@ namespace AppInstaller::Repository::Microsoft
...
     {
-        auto indexPath = Runtime::GetPathTo(Runtime::PathName::LocalState) / "pinning.db";
+        const auto DefaultIndexPath = Runtime::GetPathTo(Runtime::PathName::LocalState) / "pinning.db";
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        const auto indexPath = s_PinningIndexOverride.has_value() ? s_PinningIndexOverride.value() : DefaultIndexPath;
+#else
+        const auto indexPath = DefaultIndexPath;
+#endif
+
+        AICLI_LOG(CLI, Info, << "Opening pinning index");
```
yao-msft, line 46:
```suggestion
        AICLI_LOG(Repo, Info, << "Opening pinning index");

```

yao-msft:
>                         // TODO: Needs a whole separate change to fix the fact that we don't support multiple available packages and what the different search behaviors mean

Is this still relevant?

---
Refers to: src/AppInstallerRepositoryCore/CompositeSource.cpp:1293 in 395a9f3. [](commit_id = 395a9f39c1337469ceea6e9bba41833f3805cc11, deletion_comment = False)

yao-msft (commented):
I think we can add some e2e tests later since after this change, the feature is considiered primarily complete.

yao-msft:
By the way, I did not see changes related to --force override behavior for pinning, I thought with user passing in --force, PinBehavior will be IgnorePins, is that not needed in the whole flow?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -708,14 +709,15 @@ namespace AppInstaller::CLI::Workflow
...
         for (const auto& match : searchResult.Matches)
         {
             auto installedVersion = match.Package->GetInstalledVersion();
 
             if (installedVersion)
             {
-                auto latestVersion = match.Package->GetLatestAvailableVersion();
-                bool updateAvailable = match.Package->IsUpdateAvailable();
+                auto latestVersion = match.Package->GetLatestAvailableVersion(pinBehavior);
+                bool updateAvailable = match.Package->IsUpdateAvailable(pinBehavior);
```
yao-msft, line 720: I think for `winget list`, we are hiding latest available versions from users forever if a pin is set, maybe we should just apply the pin behavior for upgrade related scenarios

src/AppInstallerCLICore/Workflows/DependencyNodeProcessor.cpp
```diff
@@ -41,7 +41,7 @@ namespace AppInstaller::CLI::Workflow
         const auto& package = match.Package;
         auto packageId = package->GetProperty(PackageProperty::Id);
         m_nodePackageInstalledVersion = package->GetInstalledVersion();
-        m_nodePackageLatestVersion = package->GetLatestAvailableVersion();
+        m_nodePackageLatestVersion = package->GetLatestAvailableVersion(PinBehavior::ConsiderPins);
```
yao-msft, line 44: Does --include-pinned and --force apply here too?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -861,7 +862,7 @@ namespace AppInstaller::CLI::Workflow
     void GetManifestWithVersionFromPackage::operator()(Execution::Context& context) const
     {
         PackageVersionKey key("", m_version, m_channel);
-        auto requestedVersion = context.Get<Execution::Data::Package>()->GetAvailableVersion(key);
+        auto requestedVersion = context.Get<Execution::Data::Package>()->GetAvailableVersion(key, PinBehavior::ConsiderPins);
```
yao-msft, line 865:
```suggestion
        auto requestedVersion = context.Get<Execution::Data::Package>()->GetAvailableVersion(key, PinBehavior::IgnorePins);

```

This looks to be only used by show command

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
```diff
@@ -113,6 +113,7 @@
 #define APPINSTALLER_CLI_ERROR_PIN_ALREADY_EXISTS               ((HRESULT)0x8A150062)
 #define APPINSTALLER_CLI_ERROR_PIN_DOES_NOT_EXIST               ((HRESULT)0x8A150063)
 #define APPINSTALLER_CLI_ERROR_CANNOT_OPEN_PINNING_INDEX        ((HRESULT)0x8A150064)
+#define APPINSTALLER_CLI_ERROR_PACKAGE_HAS_BLOCKING_PIN         ((HRESULT)0x8A150065)
```
yao-msft, line 116: This error code is not used.

I think this is a general issue with the current implementation of hiding available versions by pin behavior. User will not be notified why a package upgrade is not applicable (they'll see a generic "no applicable upgrade found" for all pin related failures)

We may need more work to support messages like:

Package upgrade skipped due to this blocking pin.. Use --force to override
Package upgrade skipped due to this pinning pin.. Upgrade specifically or use --force
Package upgrade skipped due to gated pin.. Use --force to override

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
denelon: I think we need to have a mechanism to show if a package is pinned in list so a user could see when new versions are available. The pinning might be temporary to avoid a known bug, and the user might still be looking for any new versions.

I believe the same is true for winget upgrade. We might want a separate section for pinned packages so we can still show if new versions are available for the same reason.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -57,17 +57,26 @@ namespace AppInstaller::CLI::Workflow
...
-        auto availableVersion = package->GetLatestAvailableVersion();
+        std::set<std::string> sources;
 
         auto pinningIndex = context.Get<Execution::Data::PinningIndex>();
-        auto pin = pinningIndex->GetPin({
-            availableVersion->GetProperty(PackageVersionProperty::Id).get(),
-            availableVersion->GetProperty(PackageVersionProperty::SourceIdentifier).get() });
-        if (pin)
+
+        auto packageVersionKeys = package->GetAvailableVersionKeys(PinBehavior::IgnorePins);
```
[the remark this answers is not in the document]
florelis: This function doesn't return a reference. It returns a new vector, so we take ownership of it.

src/AppInstallerCommonCore/Public/AppInstallerVersions.h
```diff
@@ -172,16 +172,23 @@ namespace AppInstaller::Utility
     // A range of versions indicated by a version and optionally a wildcard at the end.
     struct GatedVersion
```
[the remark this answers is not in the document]
florelis: I don't know. I don't want to have to reason around, for example, order relative to other versions or gated versions, which I'd probably have to if it is a Version. And I don't think we need to.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
florelis: The upgrade command also uses this. I added an argument to the workflow task to consider or not the pins so that `show` can ignore them.

src/AppInstallerCommonCore/Public/AppInstallerErrors.h
florelis: Removed the error code and replaced it by a generic error for when a package is pinned (not specific to blocking)

src/AppInstallerCommonCore/Public/AppInstallerVersions.h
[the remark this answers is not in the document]
yao-msft: Ah I see. Got it.

florelis:
I moved the logic to filter out pinned versions out of CompositePackage's GetAvailableVersionKeys() and GetAvailableVersion(). Instead, now the version keys returned from GetAvailableVersionKeys() include the pinned state for that version. Versions from different sources may have different pin types, and versions that match a gating pin are not marked as pinned. Callers can the use that info to decide what to do.

I added another function like GetAvailableVersion() that also returns the pinned state.

I kept the PinBehavior argument for GetLatestAvailableVersion() because the logic there is more complicated and so it is harder to have each caller do it.

I'm still missing using the --force argument, and the messages that I'm exposing could probably be more specific.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
florelis: I added a count of user-pinned packages at the end of the update list, and I'm now showing the latest available version on list even if it is pinned.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -467,30 +603,53 @@ namespace AppInstaller::Repository
...
+                    auto pinKey = GetPinKey(availablePackage.AvailablePackage.get());
+                    if (m_installedPackage)
+                    {
+                        auto pin = pinningIndex.GetPin(pinKey);
+                        if (pin.has_value())
+                        {
+                            availablePackage.Pin = std::move(pin.value());
+                        }
+                    }
+                    else if (pinningIndex.GetPin(pinKey))
```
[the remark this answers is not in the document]
florelis: This is only reachable from `SearchInstalled()`. Added a comment about it.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -430,34 +539,61 @@ namespace AppInstaller::Repository
...
+                return m_installedPackage;
             }
 
             const std::shared_ptr<IPackage>& GetTrackingPackage() const
             {
                 return m_trackingPackage;
             }
 
-            void SetAvailablePackage(std::shared_ptr<IPackage> availablePackage)
+            void AddAvailablePackage(std::shared_ptr<IPackage> availablePackage)
```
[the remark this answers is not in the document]
florelis: I made using multiple available packages dependent on the pinning experimental feature. A pin may make us need multiple sources in almost all commands, for example if a gated version hides all the available versions from one source but not from other sources.

I also added a "default" available source that is used for things like TrySetOverrideInstalledVersion, and that is used first as the source of truth in GetProperty.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -320,11 +333,88 @@ namespace AppInstaller::Repository
...
         };
 
+        struct CompositeAvailablePackage
+        {
+            CompositeAvailablePackage() {}
+            CompositeAvailablePackage(std::shared_ptr<IPackage> availablePackage, std::optional<Pinning::Pin> pin = {})
+                : AvailablePackage(availablePackage), Pin(pin)
+            {
+                auto lastAvailable = AvailablePackage->GetLatestAvailableVersion(PinBehavior::IgnorePins);
+                if (lastAvailable)
```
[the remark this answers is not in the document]
florelis: We actually have a test to check that a null available version is handled properly. I'm not sure when that would happen, but I trust there was a reason for it.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -320,11 +369,113 @@ namespace AppInstaller::Repository
...
+                : AvailablePackage(availablePackage), Pin(pin)
+            {
+                auto latestAvailable = AvailablePackage->GetLatestAvailableVersion(PinBehavior::IgnorePins);
+                if (latestAvailable)
+                {
+                    SourceId = latestAvailable->GetSource().GetIdentifier();
+                }
+            }
+
+            std::string SourceId;
```
yao-msft, line 387: Should the 3 member variables be private? and named m_*?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -719,17 +722,30 @@ namespace AppInstaller::CLI::Workflow
 
         int availableUpgradesCount = 0;
         int packagesWithUnknownVersionSkipped = 0;
+        int packagesWithUserPinsSkipped = 0;
         auto &source = context.Get<Execution::Data::Source>();
         bool shouldShowSource = source.IsComposite() && source.GetAvailableSources().size() > 1;
 
+        PinBehavior pinBehavior;
```
yao-msft, line 729: nit: initialize value

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -875,7 +906,71 @@ namespace AppInstaller::CLI::Workflow
...
+        std::shared_ptr<IPackageVersion> requestedVersion;
+
+        if (m_considerPins && ExperimentalFeature::IsEnabled(ExperimentalFeature::Feature::Pinning))
+        {
+            bool isPinned = false;
+
+            // TODO: The logic here will probably have to get more difficult once we support channels
+            if (Utility::IsEmptyOrWhitespace(m_version) && Utility::IsEmptyOrWhitespace(m_channel))
+            {
+                PinBehavior pinBehavior;
```
yao-msft, line 920: nit: initialize value

src/AppInstallerCLICore/Workflows/DependencyNodeProcessor.cpp
```diff
@@ -41,7 +41,18 @@ namespace AppInstaller::CLI::Workflow
         const auto& package = match.Package;
         auto packageId = package->GetProperty(PackageProperty::Id);
         m_nodePackageInstalledVersion = package->GetInstalledVersion();
-        m_nodePackageLatestVersion = package->GetLatestAvailableVersion();
+
+        PinBehavior pinBehavior;
```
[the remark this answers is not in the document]
florelis: The compiler doesn't complain because it can tell that it is set in both branches of the conditional. In my mind it is clearer this way. If I initialize it, in my mind it opens the questions of "when does it stay with this value?", "why is this the default?", neither of which really matters.

I just checked and if it can stay uninitialized, the compiler errors out, so there's no risk of using it uninitialized even if I don't do it here.

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -232,8 +223,9 @@ namespace AppInstaller::Repository
     {
         PackageVersionKey() = default;
 
-        PackageVersionKey(Utility::NormalizedString sourceId, Utility::NormalizedString version, Utility::NormalizedString channel) :
-            SourceId(std::move(sourceId)), Version(std::move(version)), Channel(std::move(channel)) {}
+        // TODO #219 check all uses
```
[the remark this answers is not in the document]
florelis: Oh, I got the issue numbers confused and then I forgot to remove this todo. It was about checking all the uses of the constructor to make sure that the default value for pin type/state was appropriate. I just did that, so I'll remove it.

src/AppInstallerRepositoryCore/CompositeSource.cpp
florelis: Do you feel strongly about this?
It was easier for me to keep them public as I use those fields in the composite package, and the Pin one I also set from there.

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -149,6 +149,8 @@ namespace AppInstaller::CLI
             return { type, "all"_liv, 'r', "recurse"_liv, ArgTypeCategory::MultiplePackages };
         case Execution::Args::Type::IncludeUnknown:
             return { type, "include-unknown"_liv, 'u', "unknown"_liv };
+        case Execution::Args::Type::IncludePinned:
+            return { type, "include-pinned"_liv, "pinned"_liv, ArgTypeCategory::CopyFlagToSubContext };
         case Execution::Args::Type::UninstallPrevious:
             return { type, "uninstall-previous"_liv, ArgTypeCategory::InstallerBehavior };
```
yao-msft, line 155: >uninstall-previous [](http://example.com/codeflow?start=28&length=18)

I guess this should be copied to subcontext as well since we support multiple installs using subcontexts?

src/AppInstallerRepositoryCore/CompositeSource.cpp
yao-msft: If it's not a lot of code changes, I think it'd be better to be private. Because this struct is more of a class and less of a data structure now.

But if it's too many changes, I'm good with leaving it as is. Your call.

yao-msft (approved):
:shipit:

felipecrs:
That's great news! I can't wait to test it :)

denelon:
We're checking to see if any other PRs are super close to done, and we'll get a preview build out quickly™️.

PR #2820, opened 2023-01-05
Stop using `m_summary` in the global globals for telemetry

florelis:
We were observing crashes in `TelemetryTraceLogger::LogFailure()` due to heap corruption. The root cause of this is that for many types of COM calls we don't have a thread local telemetry trace logger, but instead use a process global. If there are multiple simultaneous calls, there is a possibility for a race condition, specifically around writing to the `TelemetryTraceLogger`'s `m_summary`.

The long term solution would be to stop using a process globals there, but that is tricky. As a band-aid to prevent these crashes, disable the use of `m_summary` in the process globals.

Tested by creating a local repro of the crash and verifying it didn't crash after the change. Repro'd by adding an artificial `THROW_HR_MSG` to force going to `LogFailure()` and then using the sample COM caller project to generate many simultaneous calls that would hit it.

Credit to @JohnMcPMS for figuring out the issue and the fix :)

Related to #1880

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2820)

### DIFF
--- a/src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
+++ b/src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
@@ -100,7 +100,7 @@ namespace AppInstaller::Logging
         this->IsCOMCall = other.IsCOMCall;
     }
 
-    TelemetryTraceLogger::TelemetryTraceLogger()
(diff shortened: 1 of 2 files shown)


yao-msft (approved):
:shipit:

Issue #1403, opened 2021-08-26
possibility to add parameters for installation of package


herrxyz:
# Description of the new feature/enhancement

Hey there, I would like to add parameters to the installation of a package, for example like `"sql-server-express  /SECURITYMODE=SQL /SAPWD=localadministrator"` ~~(I know that it's possible to change password afterwards)~~

some (a lot?) installation programms features installation using commandline with parameters for settings like folderpath, setting password, changing language etc
this request is inspired by chocolatey, in example `choco install sql-server-express -ia "/SECURITYMODE=SQL /SAPWD=localadministrator" -y`

vedantmgoyal2009:
@denelon Needs author feedback or closed?

denelon:
@vedantmgoyal2009 I think this is the "add switches" rather than "--override". I've been thinking about something like "--augment" to make the switches in addition to those in the manifest. That way users would still get silent and maybe scope in addition to what they add rather than having to use "--override" and know all the switches.

vedantmgoyal2009:
Do you mean what git does?

`git <command> <more-cmd> -- <additional-blah-blah>`

denelon:
Imagine you were using winget to install Visual Studio and you wanted to add some workloads.

You wouldn't necessarily want to have to look up the arguments for silent and scope, but you just wanted to add additional arguments for the installer.

winget would call the installer and have the switches for silent and scope, or during the "upgrade" flow the upgrade argument. You could then just pass the workloads with the "--augment" (or whatever we call it).

ppvnf:
I think this is an important feature for Winget to have. In the issue https://github.com/microsoft/winget-cli/issues/2553 I compared how Chocolatey and Winget work:

> I think it's easier to explain what I mean by giving an example (VirtualBox). In Chocolatey if I want to install VirtualBox without adding a desktop icon, but enabling the extension pack, I would use `choco install virtualbox --params "/NoDesktopShortcut /ExtensionPack"`. It's also possible to use `choco info virtualbox` to find out about the parameters of certain package. I am not very experienced with Winget, so I would like to know if it has a similar feature (if it doesn't I would like to know the reason and suggest it to be implemented). I tried `winget show virtualbox` but couldn't find any package parameter. In the [docs ](https://learn.microsoft.com/en-us/windows/package-manager/winget/install) the only similar package parameters I saw were `--locale`, `-l` and `-a`. Thank you in advance.

Trenly:
[...]

Issue #239, opened 2020-05-20
Support additional arguments as well as overridden arguments


gmckeown:
# Description of the new feature/enhancement

Some installers, for example WinSCP, pop up a dialog asking whether I wish to install for current user or all users (note: #149 might be relevant for this _specific_ case).

WinSCP supports the /ALLUSERS  and /CURRENTUSER options to cater for this; however, if I use ``--override`` it clears all options specified in the manifest and replaces them. So the following ends up performing an interactive install:

```winget install --override "/ALLUSERS" winscp```

So I locate the manifest, identify that it has the /VERYSILENT option and change my command to:

```winget install --override "/VERYSILENT /ALLUSERS" winscp```

This achieves what I want -- a completely silent install with my chosen option.

# Proposed technical implementation details (optional)

A new switch that *adds* the listed arguments to the installer command-line rather than replaces them. This would supplement the ``--override`` option rather than replacing it.

Something along the lines of the following would then allow me to retain the manifest-specified default options, whilst still including my choices:

```winget install --pass-args "/ALLUSERS" winscp```

They key benefit here is that if the authors of the manifest add new default options, I don't have to manually modify my install command to match.

**Comment by RehanSaeed** [comment]
I would like to see custom install arguments for each app when I do:

```
> winget install Microsoft.VisualStudio --help
> --python - Installs python
> --Node - Installs Node
> etc....
```

I would also like a **guided installation** where the installer UI shows up and I can personally select options and manually hit next, next, next, finish.

```
> winget install Microsoft.VisualStudio --guided
```

**Comment by iDolmatov** [comment]
I found an opportunity to specify additional attributes in the manifest and in the source code, but this is not fully implemented.
Cannot be started with the described command. You cannot get data from manifest when getting information about a package (_winget show_) and when polling a command (example _winget install TestApp -custom -?_).
https://github.com/microsoft/winget-cli/discussions/1067

**Comment by denelon** [comment]
@RehanSaeed `winget install <package> --interactive` should do what you're looking for.

```
 winget install -?
Windows Package Manager v1.0.11451
Copyright (c) Microsoft Corporation. All rights reserved.

Installs the selected package, either found by searching a configured source or directly from a manifest. By default, the query must case-insensitively match the id, name, or moniker of the package. Other fields can be used by passing their appropriate option.

usage: winget install [[-q] <query>] [<options>]

The following arguments are available:
  -q,--query        The query used to search for a package

The following options are available:
[...]

PR #2832, opened 2023-01-10
Add --custom argument for passing additional installer arguments

Trenly:
  - #239
  - #1403
-----

- I chose the argument based on what made sense to me and what was in the referenced issues. If a different name is desired, I'm open to suggestions. 
- Decided to name the arg type CustomSwitches to avoid future confusion; again, if there's a better name, feel free to suggest.

```raw
// Installation failure was expected as the '/?' argument brings up the MSI help menu

PS D:\Git\winget-pkgs> wingetdev install --manifest D:\Git\winget-pkgs\manifests\7\7zip\7zip\22.01 --custom /? --verbose --open-logs
Found 7-Zip [7zip.7zip] Version 22.01
Downloading https://www.7-zip.org/a/7z2201-x64.msi
  ██████████████████████████████  1.82 MB / 1.82 MB
Starting package install...
Installer failed with exit code: 1639
One of the installation parameters is invalid. The package installation log may have additional details.
```

```
2023-01-09 22:29:04.465 [CLI ] Installer args: /passive /norestart /log "C:\Users\Trenly\AppData\Local\Packages\WinGetDevCLI_8wekyb3d8bbwe\LocalState\DiagOutputDir\WinGet-7zip.7zip.22.01-2023-01-09-22-29-04.465.log" /?
2023-01-09 22:29:04.466 [CLI ] Starting: 'C:\Users\Trenly\AppData\Local\Temp\WinGet\7zip.7zip.22.01\7z2201-x64.msi' with arguments '/passive /norestart /log "C:\Users\Trenly\AppData\Local\Packages\WinGetDevCLI_8wekyb3d8bbwe\LocalState\DiagOutputDir\WinGet-7zip.7zip.22.01-2023-01-09-22-29-04.465.log" /?'
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2832)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
@@ -117,9 +117,16 @@ namespace AppInstaller::CLI::Workflow
             {
                 installerArgs += ' ' + installerSwitches.at(InstallerSwitchType::Custom);
             }
+
+            // Construct custom arg passed in by cli arg
             if (context.Args.Contains(Execution::Args::Type::CustomSwitches))
             {
-                installerArgs += ' ' + Utility::Normalize(context.Args.GetArg(Execution::Args::Type::CustomSwitches));
+                std::string_view customSwitches = context.Args.GetArg(Execution::Args::Type::CustomSwitches);
+                // Since these arguments are appended to the installer at runtime, it doesn't make sense to append them if empty or whitespace
+                if (!customSwitches.empty() && !std::all_of(customSwitches.begin(), customSwitches.end(), isspace))
+                {
+                    installerArgs += ' ' + Utility::Normalize(customSwitches);
+                }
             }
 
             // Construct update arg if applicable
--- a/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
@@ -123,9 +123,9 @@ namespace AppInstaller::CLI::Workflow
             {
                 std::string_view customSwitches = context.Args.GetArg(Execution::Args::Type::CustomSwitches);
                 // Since these arguments are appended to the installer at runtime, it doesn't make sense to append them if empty or whitespace
-                if (!customSwitches.empty() && !std::all_of(customSwitches.begin(), customSwitches.end(), isspace))
+                if (!Utility::IsEmptyOrWhitespace(customSwitches))
                 {
-                    installerArgs += ' ' + Utility::Normalize(customSwitches);
+                    installerArgs += ' ' + std::string{ customSwitches };
                 }
             }
 
--- a/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
@@ -117,6 +117,10 @@ namespace AppInstaller::CLI::Workflow
             {
                 installerArgs += ' ' + installerSwitches.at(InstallerSwitchType::Custom);
             }
+            if (context.Args.Contains(Execution::Args::Type::CustomSwitches))
+            {
+                installerArgs += ' ' + Utility::Normalize(context.Args.GetArg(Execution::Args::Type::CustomSwitches));
+            }
 
             // Construct update arg if applicable
             if (isUpdate && installerSwitches.find(InstallerSwitchType::Update) != installerSwitches.end())
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -53,6 +53,8 @@ namespace AppInstaller::CLI
             return Argument{ "architecture"_liv, 'a', Args::Type::InstallArchitecture, Resource::String::InstallArchitectureArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         case Args::Type::Log:
             return Argument{ "log"_liv, 'o', Args::Type::Log, Resource::String::LogArgumentDescription, ArgumentType::Standard };
+        case Args::Type::CustomSwitches:
+            return Argument{ "custom"_liv, NoAlias, Args::Type::CustomSwitches, Resource::String::CustomSwitchesArgumentDescription, ArgumentType::Standard};
         case Args::Type::Override:
             return Argument{ "override"_liv, NoAlias, Args::Type::Override, Resource::String::OverrideArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         case Args::Type::InstallLocation:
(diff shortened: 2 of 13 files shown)


yao-msft:
Thanks for the pr. The change looks great. But as this will be a supported feature, for feature parity, we'd better add it to Com api around [here](https://github.com/microsoft/winget-cli/blob/939e4ddb09654fe725bdc0a73bee85fffe348c00/src/Microsoft.Management.Deployment/PackageManager.idl#L766) with related implementation, also this should be added as contract version 6 since winget v1.4 just got finalized. And this is also a good one to update Powershell module around [here](https://github.com/microsoft/winget-cli/blob/939e4ddb09654fe725bdc0a73bee85fffe348c00/src/PowerShell/Microsoft.WinGet.Client/Commands/Common/BaseInstallCommand.cs#L32) with related tests.

Or alternatively, if you feel not comfortable adding above changes at the moment, we can add it behind an experimental feature, and we'll implement the Com api and PS module during winget v1.5 milestone and remove experimental feature later.

Trenly:
> Thanks for the pr. The change looks great. But as this will be a supported feature, for feature parity, we'd better add it to Com api around [here](https://github.com/microsoft/winget-cli/blob/939e4ddb09654fe725bdc0a73bee85fffe348c00/src/Microsoft.Management.Deployment/PackageManager.idl#L766) with related implementation, also this should be added as contract version 6 since winget v1.4 just got finalized. And this is also a good one to update Powershell module around [here](https://github.com/microsoft/winget-cli/blob/939e4ddb09654fe725bdc0a73bee85fffe348c00/src/PowerShell/Microsoft.WinGet.Client/Commands/Common/BaseInstallCommand.cs#L32) with related tests.
> 
> Or alternatively, if you feel not comfortable adding above changes at the moment, we can add it behind an experimental feature, and we'll implement the Com api and PS module during winget v1.5 milestone and remove experimental feature later.

Thanks for the feedback! I think I got it implemented, but I couldn't find where the tests were for the PowerShell module. Any guidance you can provide on where I should be adding additional tests would be appreciated

Trenly:
I also decided to add some whitespace handling just in case. I didn't think it made sense to error on whitespace, but if there's some other way you'd like it handled (or if you'd like the whitespace handling reverted) just let me know

yao-msft:
> > Thanks for the pr. The change looks great. But as this will be a supported feature, for feature parity, we'd better add it to Com api around [here](https://github.com/microsoft/winget-cli/blob/939e4ddb09654fe725bdc0a73bee85fffe348c00/src/Microsoft.Management.Deployment/PackageManager.idl#L766) with related implementation, also this should be added as contract version 6 since winget v1.4 just got finalized. And this is also a good one to update Powershell module around [here](https://github.com/microsoft/winget-cli/blob/939e4ddb09654fe725bdc0a73bee85fffe348c00/src/PowerShell/Microsoft.WinGet.Client/Commands/Common/BaseInstallCommand.cs#L32) with related tests.
> > Or alternatively, if you feel not comfortable adding above changes at the moment, we can add it behind an experimental feature, and we'll implement the Com api and PS module during winget v1.5 milestone and remove experimental feature later.
> 
> Thanks for the feedback! I think I got it implemented, but I couldn't find where the tests were for the PowerShell module. Any guidance you can provide on where I should be adding additional tests would be appreciated

Thank you. The Powershell tests are [here](https://github.com/microsoft/winget-cli/blob/939e4ddb09654fe725bdc0a73bee85fffe348c00/src/AppInstallerCLIE2ETests/PowerShell/WinGetClientModule.cs#L136). But after a closer look, they are just some basic smoke tests. So it's probably fine with the current unit tests only.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -118,6 +118,17 @@ namespace AppInstaller::CLI::Workflow
                 installerArgs += ' ' + installerSwitches.at(InstallerSwitchType::Custom);
             }
 
+            // Construct custom arg passed in by cli arg
+            if (context.Args.Contains(Execution::Args::Type::CustomSwitches))
+            {
+                std::string_view customSwitches = context.Args.GetArg(Execution::Args::Type::CustomSwitches);
+                // Since these arguments are appended to the installer at runtime, it doesn't make sense to append them if empty or whitespace
+                if (!customSwitches.empty() && !std::all_of(customSwitches.begin(), customSwitches.end(), isspace))
```
yao-msft, line 126: There is an existing Utility::IsEmptyOrWhitespace() implementation in AppInstallerStrings.h, please use that.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -118,6 +118,17 @@ namespace AppInstaller::CLI::Workflow
...
             }
 
+            // Construct custom arg passed in by cli arg
+            if (context.Args.Contains(Execution::Args::Type::CustomSwitches))
+            {
+                std::string_view customSwitches = context.Args.GetArg(Execution::Args::Type::CustomSwitches);
+                // Since these arguments are appended to the installer at runtime, it doesn't make sense to append them if empty or whitespace
+                if (!customSwitches.empty() && !std::all_of(customSwitches.begin(), customSwitches.end(), isspace))
+                {
+                    installerArgs += ' ' + Utility::Normalize(customSwitches);
```
yao-msft, line 128:
```suggestion
                    installerArgs += ' ' + customSwitches;

```

If it's from direct user input, I think we'd better just pass on without modification.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
Trenly: I knew I was missing something!

yao-msft (approved):
:shipit:

denelon:
We're running out of letters :)

Naming things is hard.

--augment
--custom
--switches

What happens if we have both "--override" and this new argument?

I had a discussion with the Visual Studio team about this feature a few days ago. It's common for users to want to add switches to their Visual Studio install so they can include workloads. We might also have a situation where the default install from WinGet is "Silent with progress" and the user attempts to add another switch that conflicts with the ones WinGet already has for install mode or scope. I don't know if installers would take the first argument, the last argument, or throw an exception.

Trenly:
> What happens if we have both "--override" and this new argument?
> 
Override takes precedence and only the value for override is used

> I had a discussion with the Visual Studio team about this feature a few days ago. It's common for users to want to add switches to their Visual Studio install so they can include workloads. We might also have a situation where the default install from WinGet is "Silent with progress" and the user attempts to add another switch that conflicts with the ones WinGet already has for install mode or scope. I don't know if installers would take the first argument, the last argument, or throw an exception.
>
Most installers would throw exception (from what I've seen). However, the idea of "adding your own switches" to the default switches winget uses is almost certainly a power user feature, or at least one that most users should understand to check their input before blaming the cli

akash07k:
Hey @Trenly,
With this `--custom` flag, how can we identify the supported parameters by a certain package/installer?

Trenly:
> Hey @Trenly, With this `--custom` flag, how can we identify the supported parameters by a certain package/installer?

You would have to look at the installer documentation released by the publisher of whatever application you’re trying to install.

akash07k:
I see, ok.

> > Hey @Trenly, With this `--custom` flag, how can we identify the supported parameters by a certain package/installer?
> 
> You would have to look at the installer documentation released by the publisher of whatever application you’re trying to install.

PR #2838, opened 2023-01-11
Implement PowerShell cmdlets

msftrubengu:
This PR implements two new PowerShelll cmdlets `Assert-WinGetPackageManager` and `Repair-WinGetPackageManager` as well as the WinGetPackageManager DSC Resource.

`Assert-WinGetPackageManager` verifies the integrity of winget for the current user. If winget is not set up correctly, it will throw an exception with the type of failure if detected.
It can detect:
- Failed call into `winget --version`.
- winget is not found because is not in the PATH environment variable.
- winget is not found because the App execution alias is disabled.
- Windows OS is not supported.
- The AppInstaller package is not installed in the machine.
- The AppInstaller package is not registered for the user.
- The AppInstaller package is old and doesn't contain winget.

`Assert-WinGetPackageManager -Version foo` also verifies the installed version is the same as the expected.
`Assert-WinGetPackageManager -Latest` also verifies the installed version is the latest release version.
`Assert-WinGetPackageManager -Latest -IncludePrerelease` also verifies the installed version is the latest prerelease version.

`Repair-WinGetPackageManager` does the same as `Assert-WinGetIntegrity` but attempts to fix winget depending on the failure. Returns 0 is succeeded.

`Repair-WinGetPackageManager` repairs current winget installed and performs no updates. If winget is not installed or there's catastrophic failure then installs the latest released winget.
`Repair-WinGetPackageManager -Version foo` repairs winget and makes sure that the current installed version is the expected one.
`Repair-WinGetPackageManager -Latest` repairs winget and makes sure winget is the latest released version.
`Repair-WinGetPackageManager -Latest -IncludePreRelease` repairs winget and makes sure winget is the latest prereleased version.

It sadly can't repair if AppExecution alias is disabled because there's no programmatic way to do it (requires internal APIs).

It also doesn't install Microsoft.UI.Xaml.2.7 as a dependency because the current solution to download the nuget package, extract it and install the appx within is not the right approach we want to follow. We will follow up internally with the owners of the package to do something similar as the VCLibs packages.

There's still a lot of opportunity for `Repair-WinGetPackageManager` to make it more robust, but this PR is just the starting phase.

The `WinGetPackageManager` DSC resource is just a simple wrapper around those two cmdlets. There's also a sample on how to use it. 

For now, all versions need to be the tag name of the releases in GitHub.

Also, moved Get-WinGetVersion from Crescendo to a binary cmdlet.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2838)


src/PowerShell/Microsoft.WinGet.Client/Commands/RepairCommand.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            RepairResult repairResult = RepairResult.Failure;
+
+            if (string.IsNullOrEmpty(toInstallVersion))
+            {
+                var gitHubRelease = new GitHubRelease();
+                toInstallVersion = gitHubRelease.GetLatestVersionTagName(preRelease);
+            }
+
+            if (toInstallVersion != WinGetVersionHelper.InstalledWinGetVersion)
```
Trenly, line 101: I would love to see implied wildcarding for version matching. For example - If the user specified `-Version 1.3` or `-Version 1.3.2` it would map to the version tag [1.3.2691](https://github.com/microsoft/winget-cli/releases/tag/v1.3.2691). But if they used `-Version 1.3.1` it would map to the version tag [1.3.1872](https://github.com/microsoft/winget-cli/releases/tag/v1.3.1872) and require the `-Prerelease` switch. Whereas `-Version 1.3.20` would map to [1.3.2091](https://github.com/microsoft/winget-cli/releases/tag/v1.3.2091). 

This follows the idea that the cli version is often referenced as the "1.3 Client" or the "1.4 Client", so it's very common to only use the prefix of the version number.

(Yes, I know, I'm asking for scope creep here; Sorry @denelon)

src/PowerShell/Microsoft.WinGet.Client/Commands/RepairCommand.cs
```diff
@@ -0,0 +1,165 @@
...
+                     integrityCategory == IntegrityCategory.Failure)
+            {
+                if (this.DownloadAndInstall(preRelease, this.Version, false))
+                {
+                    repairResult = RepairResult.Installed;
+                }
+            }
+            else if (integrityCategory == IntegrityCategory.AppExecutionAliasDisabled)
+            {
+                // Sorry, but the user has to manually enabled it.
```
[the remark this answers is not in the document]
msftrubengu: Sadly no, adding/deleting the registry keys doesn't imply enabling/disabling app execution alias. There are things that are executed when the toggle happens in the settings app that can only be done with internal APIs.

src/PowerShell/Microsoft.WinGet.Client/Helpers/WinGetVersionHelper.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            if (string.IsNullOrEmpty(version))
+            {
+                throw new ArgumentNullException();
+            }
+
+            // WinGet version starts with v
+            if (version[0] != 'v')
+            {
+                throw new ArgumentException();
```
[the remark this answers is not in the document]
msftrubengu: I see your point, but here we want to get rid of that `v`, but yeah throwing seems excessive.

src/PowerShell/Microsoft.WinGet.Client/Helpers/AppxModuleHelper.cs
```diff
@@ -0,0 +1,181 @@
...
+                string url;
+                if (arch == Architecture.X64)
+                {
+                    url = "https://aka.ms/Microsoft.VCLibs.x64.14.00.Desktop.appx";
+                }
+                else if (arch == Architecture.X86)
+                {
+                    url = "https://aka.ms/Microsoft.VCLibs.x86.14.00.Desktop.appx";
+                }
+                else
```
JohnMcPMS, line 159: else if ARM64 then get the ARM64 package...

src/PowerShell/Microsoft.WinGet.Client/Helpers/AppxModuleHelper.cs
```diff
@@ -0,0 +1,186 @@
...
+
+                this.psCmdlet.WriteDebug($"Installing VCLibs {url}");
+                this.psCmdlet.InvokeCommand.InvokeScript(string.Format(AddAppxPackageFormat, tmpFile));
+            }
+        }
+
+        private void InstallUiXaml()
+        {
+            // TODO: We need to follow up for Microsoft.UI.Xaml.2.7
+            // downloading the nuget and extracting it doesn't sound like the right thing to do.
```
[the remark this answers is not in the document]
JohnMcPMS: I agree that we should implement that before fully releasing this, but I don't think it should necessarily block this PR.

src/PowerShell/Microsoft.WinGet.Client/Helpers/WinGetVersionHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+            // WinGet version starts with v
+            if (version[0] == 'v')
+            {
+                version = version.Substring(1);
+            }
+
+            // WinGet version might end with -preview
+            if (version.EndsWith("-preview"))
+            {
+                version = version.Substring(0, version.IndexOf('-'));
```
JohnMcPMS, line 50: Shouldn't that be part of the version though?  Even if we only use it to distinguish between the github release/prerelease versions?

src/PowerShell/Microsoft.WinGet.Client/Properties/Resources.resx
```diff
@@ -119,48 +119,88 @@
...
+    <value>The App Installer does not contain the Windows Package Manager.</value>
+  </data>
+  <data name="IntegrityFailureMessage" xml:space="preserve">
+    <value>Windows Package Manager returned an unexcepted result.</value>
+  </data>
+  <data name="IntegrityNotInPathMessage" xml:space="preserve">
+    <value>The App Installer did not automatically add the PATH environment variable.</value>
+  </data>
+  <data name="IntegrityOsNotSupportedMessage" xml:space="preserve">
+    <value>The Windows Package Manager requires at least Version 1809 (October 2018 Update).</value>
```
JohnMcPMS, line 183:
```suggestion
    <value>The Windows Package Manager requires Windows Version 1809 (October 2018 Update) or later.</value>
```

JohnMcPMS (changes requested):
At a minimum the ARM64 change and the string change are needed.  If we could do the UI XAML code easily, then I suspect you would have done it.  But maybe you didn't investigate how difficult it would be to implement?

src/PowerShell/Microsoft.WinGet.Client/Common/IntegrityCategory.cs
```diff
@@ -0,0 +1,64 @@
...
+//     Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+// </copyright>
+// -----------------------------------------------------------------------------
+
+namespace Microsoft.WinGet.Client.Common
+{
+    /// <summary>
+    /// The type of the integrity check failure.
+    /// </summary>
+    public enum IntegrityCategory
```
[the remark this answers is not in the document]
msftrubengu: I'll fix this in another iteration.

src/PowerShell/Microsoft.WinGet.Client/Helpers/WinGetVersionHelper.cs
msftrubengu: Created WinGetVersion that deals with all that

src/PowerShell/Microsoft.WinGet.Client/Helpers/AppxModuleHelper.cs
```diff
@@ -139,31 +151,79 @@ private PSObject GetAppxObject(string packageName)
...
                 vcLibsPackageObjs.Count == 0)
             {
+                var packages = new List<string>();
+
                 var arch = RuntimeInformation.OSArchitecture;
-                string url;
                 if (arch == Architecture.X64)
                 {
-                    url = "https://aka.ms/Microsoft.VCLibs.x64.14.00.Desktop.appx";
+                    packages.Add(VCLibsUWPDesktopX64);
```
JohnMcPMS, line 200: Shouldn't this just be implemented as iterating over whatever you get back from `GetVCLibsDependencies`?

JohnMcPMS (approved):

Issue #520, opened 2020-07-30
tedious to list available versions of a package..


darbotron:
It would be nice if getting hold of the available versions of a package was less involved.
If you don't know the package path in advance, you have to search for the package with `winget search xxx` then again with `winget show xxx.yyy.zzz -versions`, then `winget install xxx.yyy.zzz -v a.b.c`.
I'd like to be able to use `-versions` with `search` as this would cut out an unnecessary step,

megamorf:
I think the separation of concern between searching across all latest versions of packages (`winget search`) and listing available versions of a specific packages (`winget search`) is good as it is. Mixing both just makes the code base needlessly complex and harder to maintain.

denelon:
I think I have some good news coming. @JohnMcPMS did a hackathon project and has a PR out #521. If you're using PowerShell, you will have the ability to tab through the versions.

vedantmgoyal2009:
* https://github.com/microsoft/winget-cli/pull/521

Auto-completion is now available. This can be marked as completed.

denelon:
@vedantmgoyal2009 I was looking at a YouTube video someone created for the Windows Package Manager. I noticed that they used "winget search" to try and find versions. I think this is a feature we want to implement. It's essentially adding "--versions" to the search command, and would produce the same output as "winget show --versions".

denelon:
This is resolved with:
* #2847

The feature is available in WinGet [v1.5.441-preview](https://github.com/microsoft/winget-cli/releases/tag/v1.5.441-preview).

PR #2847, opened 2023-01-13
Allow Version Listing through 'Winget Search'

Trenly:
  - #520
-----

This PR moves the workflow for ShowAppVersions into the workflow base and makes it available to the search command.
Intentionally left the search command as `SearchSourceForMany` to retain the same searching behavior. 

```
PS C:\WINDOWS\system32> wingetdev search 7zip.7zip --versions
Name              Id                  Source
--------------------------------------------
7-Zip             7zip.7zip           winget
7-Zip Alpha (msi) 7zip.7zip.Alpha.msi winget
7-Zip Alpha (exe) 7zip.7zip.Alpha.exe winget
PS C:\WINDOWS\system32> wingetdev search -e 7zip.7zip --versions
Found 7-Zip [7zip.7zip]
Version
----------
22.01
22.00
21.07
21.06
19.00.00.0
16.04
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2847)

### DIFF
--- a/src/AppInstallerCLICore/Commands/SearchCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/SearchCommand.cpp
@@ -76,26 +76,22 @@ namespace AppInstaller::CLI
     {
         context.SetFlags(Execution::ContextFlag::TreatSourceFailuresAsWarning);
 
+        context <<
+            Workflow::OpenSource() <<
+            Workflow::SearchSourceForMany <<
+            Workflow::HandleSearchResultFailures <<
+            Workflow::EnsureOneMatchFromSearchResult(false);
 
-        if (context.Args.Contains(Execution::Args::Type::ListVersions))
-        {
-            context <<
-                Workflow::OpenSource() <<
-                Workflow::SearchSourceForMany <<
-                Workflow::HandleSearchResultFailures <<
-                Workflow::EnsureOneMatchFromSearchResult(false) <<
+            if (context.Args.Contains(Execution::Args::Type::ListVersions))
+            {
+                context <<
                 Workflow::ReportPackageIdentity <<
                 Workflow::ShowAppVersions;
-        }
-        else
-        {
-            context <<
-                Workflow::OpenSource() <<
-                Workflow::SearchSourceForMany <<
-                Workflow::HandleSearchResultFailures <<
-                Workflow::EnsureMatchesFromSearchResult(false) <<
-                Workflow::ReportSearchResult;
-        }
+            }
+            else
+            {
+                context << Workflow::ReportSearchResult;
+            }
         
     }
 }
--- a/src/AppInstallerCLICore/Commands/SearchCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/SearchCommand.cpp
@@ -79,18 +79,20 @@ namespace AppInstaller::CLI
         context <<
             Workflow::OpenSource() <<
             Workflow::SearchSourceForMany <<
-            Workflow::HandleSearchResultFailures <<
-            Workflow::EnsureOneMatchFromSearchResult(false);
+            Workflow::HandleSearchResultFailures;
 
             if (context.Args.Contains(Execution::Args::Type::ListVersions))
             {
                 context <<
+                Workflow::EnsureOneMatchFromSearchResult(false) <<
                 Workflow::ReportPackageIdentity <<
                 Workflow::ShowAppVersions;
             }
             else
             {
-                context << Workflow::ReportSearchResult;
+                context << 
+                    Workflow::EnsureMatchesFromSearchResult(false) <<
+                    Workflow::ReportSearchResult;
             }
         
     }
--- a/src/AppInstallerCLICore/Commands/SearchCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/SearchCommand.cpp
@@ -25,6 +25,7 @@ namespace AppInstaller::CLI
             Argument::ForType(Execution::Args::Type::Exact),
             Argument::ForType(Execution::Args::Type::CustomHeader),
             Argument::ForType(Execution::Args::Type::AcceptSourceAgreements),
+            Argument::ForType(Execution::Args::Type::ListVersions),
         };
     }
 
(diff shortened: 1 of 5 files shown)


src/AppInstallerCLICore/Commands/SearchCommand.cpp
```diff
@@ -75,11 +76,26 @@ namespace AppInstaller::CLI
...
-            Workflow::SearchSourceForMany <<
-            Workflow::HandleSearchResultFailures <<
-            Workflow::EnsureMatchesFromSearchResult(false) <<
-            Workflow::ReportSearchResult;
+
+        if (context.Args.Contains(Execution::Args::Type::ListVersions))
+        {
+            context <<
+                Workflow::OpenSource() <<
+                Workflow::SearchSourceForMany <<
```
yao-msft, line 84: Should this be SearchSourceForSingle?

Now I'm a little torn.

If we use `SearchSourceForMany`, it's highly likely search will return more than 1 result, and this will print "too many packages matched, please refine input" most of the times unless user inputs filters like --id --exact, etc.

If we use `SearchSourceForSingle`, the search will try to lock down on 1 result though other search candidates may be relevant. Sort of against the purpose of Search command.

Previously, Search command intends to be searching the broad and Show command intends to find 1 result. But with this feature, Search will behave differently based on whether --ListVersions are passed in. I somewhat feel --ListVersions does not relate well with purpose of Search.

@denelon, any thoughts?

Trenly: My thought was that it would be better to use `SearchSourceForMany` here, since as you said, filtering down is sort of against the purpose of the search command. I would expect that most users would be able to understand how to refine a search query down to a single package just by looking at the help page, especially when in many cases it will be as simple as adding `-e` or `-n 1`

As a user, I find the difference between `SearchSourceForMany` and `SearchSourceForSingle` quite confusing. Take a look at this for example -
```
PS C:\WINDOWS\system32> winget search resourcehacker
Name            Id                          Version
---------------------------------------------------
Resource Hacker AngusJohnson.ResourceHacker 5.1.8
PS C:\WINDOWS\system32> winget show resourcehacker
No package found matching input criteria.
```

Search clearly returned a single result, so why wouldn't show also find that single result? Is this actually a bug in `SearchSourceForSingle`? 

As to how this relates to the intent behind search being broad and show being for a single package, that is certainly a consideration and is actually the primary reason I left it as the `SearchSourceForMany`. More precisely, it was so that the search behavior was always consistent within the search command. The intent of this change is to reduce friction for users. As @denelon said in [#520](https://github.com/microsoft/winget-cli/issues/520#issuecomment-1251242279) -
>  I was looking at a YouTube video someone created for the Windows Package Manager. I noticed that they used "winget search" to try and find versions. I think this is a feature we want to implement. It's essentially adding "--versions" to the search command, and would produce the same output as "winget show --versions".

-----
The other option I thought of was, if there are multiple packages returned, to add a column to the table listing the latest few versions of each package. However, I decided against this due to scalability if more than just a few results are returned. Of course thresholds could be set, but that seemed excessive

yao-msft: Regarding the ResourceHacker question, SearchForMany searches more fields with `substring` search, SearchForSingle on the other hand, searches only PackageName, Id, Moniker, PackageFamilyName and ProductCode with `CaseInsensitive` search, which is more strict and less likely to return multiple results.

Trenly: > Regarding the ResourceHacker question, SearchForMany searches more fields with `substring` search, SearchForSingle on the other hand, searches only PackageName, Id, Moniker, PackageFamilyName and ProductCode with `CaseInsensitive` search, which is more strict and less likely to return multiple results.

I see; Thank you for that clarification, that makes more sense.

denelon:
We could add the optional argument as a hint to the output when we get ambiguous results.

Trenly:
> What if we do the same thing we're doing for `winget install` with the "interactive / prompt" mode?
> 
> I've been looking at several issues this morning and thinking more about the other "mode" of behavior for WinGet.

UX is certainly the hard part here. I think that the interactivity / other "mode" is a bit of a tangent that's getting intermixed. At the root of the question in this PR is really more "how important is it to keep search behavior the same within a command"

I see it as this -

1) Do we truly want users to be able to list versions of a package through `winget search`?

If yes:

2) Do we want the search results when using the `--versions` argument to be consistent with `winget show <query>`  (search source for multiple) or  with `winget search <query>`  (search source for single) or, does this feature need to wait for the alternate "mode" of UX?
   * If  consistent with `winget search` -
     + Pro: Search results will be the same when using --versions
     + Pro: Partial package names / ids can be used
     + Pro: Packages can be matched by tag or command
     -  Con: There is a higher likelihood that users will see `Please refine your input` message
     - Con: It may be hard to get to the single package a user wants a version for
   * If consistent with `winget show` -
     + Pro: It will be easier to get to a single package
     - Con: Full ID's, monikers, or names must be used
     - Con: Could lead to confusion as to why the same base command produces different search results
     + Pro: It will help the user refine their input before they go to use `winget install`

denelon:
Let's get some user feedback: https://twitter.com/DenelonMs/status/1613963532873388032

denelon:
poll

yao-msft:
So when search returns multiple results, do we want to do:
Option 1: "Multiple matches found, please refine input"
Or
Option 2: "List all versions as a new column for all matched ones" like
```
Id        Name       Versions
---------------------------------------
Foo1   Bar1         1.0, 2.0, 3.0
Foo2   Bar2         1.0, 2.0, 3.0
```

I guess 2 if user is expecting "consistency with search behavior"?

denelon:
I think it's pretty much what was put in the original description. The user will need to refine the input to get versions with "search".

yao-msft:
> I think it's pretty much what was put in the original description. The user will need to refine the input to get versions with "search".

Ok, then this change is good at current state.

src/AppInstallerCLICore/Commands/SearchCommand.cpp
```diff
@@ -75,11 +76,26 @@ namespace AppInstaller::CLI
...
+                Workflow::HandleSearchResultFailures <<
```
yao-msft, line 85: nit: we should move common workflows above the if else block

          context <<
                Workflow::OpenSource() <<
                Workflow::SearchSourceForMany <<
                Workflow::HandleSearchResultFailures <<

Trenly: Moved

src/AppInstallerCLICore/Commands/SearchCommand.cpp
```diff
@@ -79,7 +80,18 @@ namespace AppInstaller::CLI
             Workflow::OpenSource() <<
             Workflow::SearchSourceForMany <<
             Workflow::HandleSearchResultFailures <<
-            Workflow::EnsureMatchesFromSearchResult(false) <<
-            Workflow::ReportSearchResult;
+            Workflow::EnsureOneMatchFromSearchResult(false);
```
yao-msft, line 83: Moved 1 more workflow up than needed
For list versions, we should use Workflow::EnsureOneMatchFromSearchResult(false)
else, Workflow::EnsureMatchesFromSearchResult(false)

yao-msft (approved):

yao-msft:
Somehow the license/cla task is stuck. And I cannot override it without changing project settings. Try close and reopen.

Issue #2845, opened 2023-01-12
WinGet configure 


crutkas:
### Description of the new feature / enhancement

As I developer I want an automated configuration for my project so it's easier for developers to contribute. 

[PowerToys](https://github.com/microsoft/PowerToys) has a nuanced configuration required to help developers succeed when contributing new features or bug fixes. WinGet can install the tools we need, but there are some configuration parts that are missing and we need a smidge more.  

I want to be able to `winget configure <configuration file>` or `winget configure <URL to configuration file>`so I can build an automated setup process and make anyone that wants to build / contribute back to PowerToys the easiest process possible.

After the process has completed, users can simply open the solution and [F5] in Visual Studio to build the solution / installer / installer installs PowerToys 

### Proposed technical implementation details

1. Read a configuration file. 
2. Apply the configuration to my local environment.

denelon:
@crutkas

Thank you for creating the Issue. We've been discussing some things internally as a part of our findings related to:
* #2697

We've also been working on:
* #221

As we've been working with the [PowerShell](https://github.com/powershell/PowerShell) team during the implementation of the WinGet DSC resources to provide a way to get WinGet delivered as a part of "App Installer" without the dependency of the Microsoft Store, we've also been thinking about configuration management. Some of our engineers have been doing further exploration. 

*Note: The DSC work may also be a way to begin support for Windows Server and Windows LTCS editions. We still have work to enable WinGet to "upgrade" itself with the absence of the Microsoft Store. We still have work to understand the implications of Windows Server without the Desktop experience.*

We have been working on a declarative configuration file and ways to apply a configuration to the local system. It would help address the lack of pre-install and post-install actions in WinGet. As a consistent part of our approach to package management, we would want the same kind of behavior for configuration management:

1. Read a "declarative" configuration file
2. Validate the configuration file syntax
3. Build the dependency graph
4. Load the necessary PowerShell DSC resources
5. Apply the configuration (Invoke-DscResource)

denelon:
@SteveL-MSFT,

I shared the work we have in progress with the PowerShell DSC Community:
* #2697 

We're planning to present a demo at their next [community call](https://dsccommunity.org/community_calls/).

For the work on this Issue, we would love to continue collaborating with you and the PowerShell team.

MorNando:
> @crutkas
> 
> 
> 
> Thank you for creating the Issue. We've been discussing some things internally as a part of our findings related to:
> 
> * #2697
> 
> 
> 
[...]

PR #2851, opened 2023-01-13
Schema draft for WinGet configure

ryfu-msft:
Related to: #2845 

This is a rough draft of the proposed schema that outlines the fields that could be utilized to provide the information necessary for WinGet configure.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2851)


denelon:
@SteveL-MSFT based on some of our internal proposals, @ryfu-msft drafted this JSON schema. We believe this will ultimately best live in a PowerShell DSC GitHub repository in the future, but we would like to iterate here. We will create an https://aka.ms link for the schema so preview users can also try this out, and when the time is right to move this to a PowerShell repository, we can create the PR and adjust the https://aka.ms link.

I've added you as a reviewer since it should follow any prior art with respect to PowerShell naming and idioms.

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+        "resources": {
+            "type": "array",
+            "items": { "$ref": "#/defs/resource" }
+        },
+        "parameters": {
+            "type": "array",
+            "items": { "$ref": "#/defs/resource" },
+            "description": "Resources that retrieve information via a 'get' operation."
+        },
+        "configurationType": {
```
JohnMcPMS, line 20: I don't understand what other values are expected here in the future.  Unless we very clearly have other schemas that will use other values and we use this to differentiate them, *and* those files are expected to be confusable with our own, we should hold off on adding it.

Additionally, `dsc` seems a bit too generic.

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+            "type": "string",
+            "default": "dsc",
+            "const": "dsc",
+            "description": "The configuration type."
+        },
+        "configurationVersion": {
+            "type": "string",
+            "default": "0.1",
+            "pattern": "^(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])(\\.(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])){2}$",
```
JohnMcPMS, line 29: This looks like neither of the regex options offered at https://semver.org/.  Is there a reason for us not to use semantic versioning?

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+            "description": "The configuration syntax version"
+        }
+    },
+    "$defs": {
+        "resource": {
+            "type": "object",
+            "properties": {
+                "resourceName": {
```
JohnMcPMS, line 37: The proposed property here was `resource`.  Why was it changed?

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+                },
+                "id": {
+                    "$ref": "#/defs/id"
+                },
+                "dependsOn": {
+                    "type": [ "string", "null" ],
+                    "description": "The list of resource ids identifying dependencies.",
+                    "uniqueItems": true
+                },
+                "attribute": {
```
JohnMcPMS, line 50: The proposed property was `directives`; these values all fall under that one.

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+                            "$ref": "#/defs/module"
+                        },
+                        "description": {
+                            "$ref": "#/defs/description"
+                        }
+                    }
+                },
+                "settings": {
+                    "type": "object",
+                    "properties": {
```
JohnMcPMS, line 63: None of the properties here should be defined; they are all defined by the individual resource and thus can change wildly in value.

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+                        },
+                        "windowsUpdate": {
+                            "type": [ "boolean", "null" ],
+                            "description": "Indicates whether to allow Windows to update."
+                        }
+                    }
+                },
+                "directives": {
+                    "type": "object", 
+                    "properties": {
```
JohnMcPMS, line 131: Is the plan to be strict on directives?  That does provide the possibility of the guarantee that "if you can provide it, it will be implemented".  I think that if we want to be strict here, it might be good to provide a defined extensibility point for users to attach arbitrary data to a resource.

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+        },
+        "module": {
+            "type": "string",
+            "maxLength": 128,
+            "description": "The name of the module."
+        },
+        "description": {
+            "type": "string",
+            "maxLength": 512,
+            "description": "The description of the resource."
```
JohnMcPMS, line 196:
```suggestion
            "description": "The description of the desired state."
```
The point is to describe this invocation of the resource, not the resource itself.

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+        },
+        "version": {
+            "type": "string",
+            "pattern": "^[^\\\\/:\\*\\?\"<>\\|\\x01-\\x1f]+$",
```
JohnMcPMS, line 200: Why not semver?

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+            },
+            "required": [
+                "resourceName",
+                "id"
+            ]
+        },
+        "id": {
+            "type": "string",
+            "maxLength": 128,
+            "description": "The identifier of the resource."
```
JohnMcPMS, line 186:
```suggestion
            "description": "The identifier of this item."
```
Important to distinguish between a particular instance of a resource, and the resource itself.

JohnMcPMS (changes requested):

schemas/JSON/configuration/configuration.schema.0.1.json
```diff
@@ -0,0 +1,204 @@
...
+                "resourceName": {
+                    "type": "string",
+                    "maxLength": 128,
+                    "description": "The name of the resource."
+                },
+                "id": {
+                    "$ref": "#/defs/id"
+                },
+                "dependsOn": {
+                    "type": [ "string", "null" ],
```
JohnMcPMS, line 46: This should be an array of strings.  Maybe I don't really understand the JSON schema well enough but it doesn't appear to be here.

schemas/JSON/configuration/configuration.schema.0.1.json
ryfu-msft: Makes sense, removed "configurationType"

schemas/JSON/configuration/configuration.schema.0.1.json
ryfu-msft: Changed to semantic versioning regex

schemas/JSON/configuration/configuration.schema.0.1.json
ryfu-msft: Changed to "resource". I thought having the same property name as the object could cause a naming conflict but I guess not.

schemas/JSON/configuration/configuration.schema.0.1.json
ryfu-msft: Removed all properties in settings and defined it as a generic object.

schemas/JSON/configuration/configuration.schema.0.1.json
ryfu-msft: I set the ```"additionalProperties": true``` so that arbitrary data can be included in the ```directives``` object. I left the properties as is for now since I'm not sure which ones are confirmed yet.

schemas/JSON/configuration/configuration.schema.0.1.json
ryfu-msft: changed to semver

schemas/JSON/configuration/configuration.schema.0.1.json
ryfu-msft: changed to an array object.

schemas/JSON/configuration/configuration.schema.0.1.json
denelon: @SteveL-MSFT is this the same regular expression PowerShell uses in general?

ryfu-msft:
This schema is not final and we will continue to iterate on this version for now. Planning to check this in so that people can use it as a reference for building their own configuration yamls.

denelon (approved):

PR #2855, opened 2023-01-14
Block msix provisioning api calls where known OS bugs exist

yao-msft:
- msix provision and deprovision apis under packaged context
- msstore provision apis under admin context

Related OS bugs have been filed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2855)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
+++ b/src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
@@ -85,12 +85,22 @@ namespace AppInstaller::CLI::Workflow
             // Verifying/Acquiring product ownership
             context.Reporter.Info() << Resource::String::MSStoreInstallTryGetEntitlement << std::endl;
 
-            AICLI_LOG(CLI, Info, << "Get user entitlement.");
-            GetEntitlementResult result = installManager.GetFreeUserEntitlementAsync(productId, winrt::hstring(), winrt::hstring()).get();
-            if (result.Status() == GetEntitlementStatus::NoStoreAccount)
+            GetEntitlementResult result{ nullptr };
+
+            if (Manifest::ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope)) == Manifest::ScopeEnum::Machine)
             {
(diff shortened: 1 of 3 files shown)


src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -143,6 +153,17 @@ namespace AppInstaller::CLI::Workflow
 
         if (Manifest::ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope)) == Manifest::ScopeEnum::Machine)
         {
+            // TODO: There was a bug in InstallService where admin user is incorrectly identified as not admin,
+            // causing false access denied on many OS versions.
+            // Remove this check when the OS bug is fixed and back ported.
+            if (!Runtime::IsRunningAsSystem())
+            {
+                context.Reporter.Error() << Resource::String::InstallFlowReturnCodeSystemNotSupported << std::endl;
```
florelis, line 161: That string is "The current system configuration does not support the installation of this package." If I saw that I would think the package is not supported at all, but I would like to know that I may at least be able to install it without admin privileges.

florelis (approved):

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
yao-msft: The only applicable case where this will succeed is the InProcCom with system context. And for InProcCom, reporter strings are not reported. This is here mainly for consistency and for local system context testing. This will be cherry picked as hot fix for our second release candidate and we'd prefer not have a loc update to get it delayed, so if we are not happy with this string, I can just remove this line.

Trenly: How about `PendingWorkError`?

florelis: I'm okay with keeping it as-is

PR #2861, opened 2023-01-17
Allow multiple apps in a single command

### LINKED ISSUE #219: Allow multiple apps to be installed using a single command

Let us call `winget install` with a list of apps, so it'll install them at the same time OR in sequence.

Proposal for command: `winget install app-1 app-2 app-3`

florelis:
This change adds support specifying multiple packages in a single command for install, update and uninstall. For example, `winget install app1 app2`.

* Added a new argument type called MultiQuery, which is the same as Query but has a limit for values larger than 1. Thankfully the arg parsing was already implemented :)
* Used this new argument in the install, upgrade and uninstall commands. The first step in these commands is now to check if MultiQuery has more than 1 value, and if it doesn't it is replaced by a single Query argument. This way, the change is not visible to existing code.
* If there is more than one value for the multi-query argument, we proceed to search for each query in a separate sub-query, and then we install them all one at a time. The code for this is refactored from the import flow which does pretty much the same thing.
* Added unit tests for these 3 commands.

I did not extend argument validation in this PR. Ideally we would block arguments that don't make sense for more than one package when doing multi-queries, like --override. I will send a separate PR to refactor the argument validation to make this easier without repeating it in the three commands.

Closes #219

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2861)


src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -137,7 +138,20 @@ namespace AppInstaller::CLI
                     Workflow::OpenCompositeSource(Repository::PredefinedSource::Installed, false, Repository::CompositeSearchBehavior::AvailablePackages);
             }
 
-            context << Workflow::InstallOrUpgradeSinglePackage(false);
+            if (!context.Args.Contains(Execution::Args::Type::MultiQuery))
```
Trenly, line 141: Nit: Seems a bit odd to be using a `not` here. Would be more readable if you switched the if-else and removed the not

src/AppInstallerCLICore/Workflows/MultiQueryFlow.cpp
```diff
@@ -0,0 +1,148 @@
...
+                }
+            }
+
+            packageSubContexts.emplace_back(std::move(searchContextPtr));
+        }
+
+        if (!foundAll)
+        {
+            AICLI_LOG(CLI, Info, << "Could not find one or more packages");
+            if (context.Args.Contains(Execution::Args::Type::IgnoreUnavailable))
```
Trenly, line 135: Does this arg need to be added to install / upgrade in case one of the multiqueries returns no packages? What happens if the user runs `winget install 7zip chrome ThisExampleIsNotPresent`? Would all packages fail, or would 7zip and chrome still be installed?

florelis: It does not *need* to be added, but we could add it. As it is, it would fail with a message that not all packages were found before installing anything.

src/AppInstallerCLICore/Workflows/MultiQueryFlow.cpp
```diff
@@ -0,0 +1,153 @@
...
+                    if (searchTerminationHR == APPINSTALLER_CLI_ERROR_UPDATE_NOT_APPLICABLE ||
+                        searchTerminationHR == APPINSTALLER_CLI_ERROR_PACKAGE_ALREADY_INSTALLED)
+                    {
+                        AICLI_LOG(CLI, Info, << "Package is already installed: [" << packageString << "]");
+                        context.Reporter.Info() << Resource::String::MultiInstallPackageAlreadyInstalled(packageString) << std::endl;
+                        continue;
+                    }
+                    else
+                    {
+                        AICLI_LOG(CLI, Info, << "Package not found: [" << packageString << "]");
```
yao-msft, line 124: Looks like it could also be "multiple package found" since we EnsureOneMatch earlier

src/AppInstallerCLICore/Workflows/MultiQueryFlow.cpp
```diff
@@ -0,0 +1,153 @@
...
+        if (!foundAll)
+        {
+            AICLI_LOG(CLI, Info, << "Could not find one or more packages");
+            if (context.Args.Contains(Execution::Args::Type::IgnoreUnavailable))
+            {
+                AICLI_LOG(CLI, Info, << "Ignoring unavailable packages due to command line argument");
+            }
+            else
+            {
+                AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_NOT_ALL_PACKAGES_FOUND);
```
yao-msft, line 146: This may be not all found for single match

src/AppInstallerCLICore/Workflows/MultiQueryFlow.cpp
```diff
@@ -0,0 +1,153 @@
...
+                searchContext << Workflow::SelectSinglePackageVersionForInstallOrUpgrade(/* isUpgrade */ false);
+                break;
+            case SearchPurpose::Upgrade:
+                searchContext << Workflow::SelectSinglePackageVersionForInstallOrUpgrade(/* isUpgrade */ true);
+                break;
+            case SearchPurpose::Uninstall:
+                searchContext <<
+                    Workflow::HandleSearchResultFailures <<
+                    Workflow::EnsureOneMatchFromSearchResult(true) <<
+                    Workflow::ReportPackageIdentity;
```
yao-msft, line 97: Remove this one, looks like we are double reporting since it's reported in line 91 in UninstallMultiplePackages( and that is the better place to report since we started to do uninstall)

src/AppInstallerCLICore/Command.cpp
```diff
@@ -405,14 +410,10 @@ namespace AppInstaller::CLI
     const CLI::Argument* ParseArgumentsStateMachine::NextPositional()
     {
         // Find the next appropriate positional arg if the current itr isn't one or has hit its limit.
-        if (m_positionalSearchItr != m_arguments.end() &&
+        while (m_positionalSearchItr != m_arguments.end() &&
```
yao-msft, line 413: This seems like behavior changing, is it intentional? I do not see how multiquery would need to change this behavior.

Previous:
If itr is not positional or it's positional reaching limit, find next positional
Now:
If itr is not positional or it's positional reaching limit, find next positional not reaching limit

florelis: Yes, this is intentional. As it was before we would always move to the next argument, even if it had not reached its limit, effectively making the limit 1 for all

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -443,10 +452,19 @@ namespace AppInstaller::CLI::Workflow
...
         if (args.Contains(Execution::Args::Type::Query))
         {
             searchRequest.Query.emplace(RequestMatch(matchType, args.GetArg(Execution::Args::Type::Query)));
         }
+        if (args.Contains(Execution::Args::Type::MultiQuery))
+        {
+            auto queries = args.GetArgs(Execution::Args::Type::MultiQuery);
+            for (const auto& query : *queries)
+            {
+                searchRequest.Query.emplace(RequestMatch(matchType, query));
```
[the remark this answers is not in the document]
florelis: I don't know why I assumed it was a vector... This was only for the list command. I don't think it is that important to have multi query in that one, so I just reverted that one back to single query instead of reworking it..

florelis:
Only thing pending now should be arg validation, which once I merge #2862 will be only tagging the arg with the right categories. Although it will have to be a special case since it depends on whether it has one or multiple values.

src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -128,6 +128,7 @@ namespace AppInstaller::CLI
         else
```
Trenly, line 128: **This is outside the scope of the PR**, but could manifest also be a multi-parameter field? It's possible that a user would want to install an entire set of manifests

Should I create a separate issue for that?

src/AppInstallerCLICore/Commands/UninstallCommand.cpp
```diff
@@ -129,19 +130,29 @@ namespace AppInstaller::CLI
...
+                    Workflow::HandleSearchResultFailures <<
+                    Workflow::EnsureOneMatchFromSearchResult(true) <<
+                    Workflow::ReportPackageIdentity <<
+                    Workflow::UninstallSinglePackage;
+            }
+            else
+            {
+                context <<
+                    Workflow::GetMultiSearchRequests <<
+                    Workflow::SearchSubContextsForSingle(Workflow::SearchSubContextsForSingle::SearchPurpose::Uninstall) <<
```
Trenly, line 153: After reading through this again, it seems like the implementation in #2877 of SearchResultType and the SearchPurpose here are similar; Would it make sense to align them both to `SearchPurpose` and use it more globally rather than inside the namespace of SearchSubContextsForSingle?

If so, I'll update that PR

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -432,6 +432,15 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
+    void CheckForMultiQuery(Execution::Context& context)
+    {
+        if (context.Args.GetCount(Execution::Args::Type::MultiQuery) == 1)
```
Trenly, line 437: Is there ever a time where the MultiQuery count could be 0? Or where MultiQuery could be null? I'm not sure how all the other parameters and filters like `--id` and `--tag` interact with query.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -831,9 +831,9 @@ They can be configured through the settings file 'winget settings'.</value>
   <data name="ImportInstallFailed" xml:space="preserve">
     <value>One or more imported packages failed to install</value>
   </data>
-  <data name="ImportSearchFailed" xml:space="preserve">
-    <value>Package not found for import: {0}</value>
-    <comment>{Locked="{0}"} Error message displayed when the user attempts to import an application package that was not found. {0} is a placeholder replaced by the import package name .</comment>
+  <data name="MultiInstallSearchFailed" xml:space="preserve">
+    <value>Package not found or multiple packages found matching query: {0}</value>
```
Trenly, line 835: I'm not a big fan of ambiguous error messages. It would be better UX if we can tell the user which of the two it was specifically. 

I believe in standard queries this is done through the use of separate workflows -
```c++
<< Workflow::EnsureMatchesFromSearchResult /* determines if package was not found */
<< Workflow::EnsureOneMatchFromSearchResult /* determines if multiple packages were found */
```

src/AppInstallerCLITests/Command.cpp
```diff
@@ -554,3 +566,51 @@ TEST_CASE("ParseArguments_UnknownName", "[command]")
 
     REQUIRE_COMMAND_EXCEPTION(command.ParseArguments(inv, args), CLI::Resource::String::InvalidNameError(Utility::LocIndView{ values[1] }));
 }
+
+TEST_CASE("ParseArguments_PositionalWithMultipleValues", "[command]")
+{
+    Args args;
+    TestCommand command({
+            Argument{ "multi", 'm', Args::Type::Query, DefaultDesc, ArgumentType::Positional }.SetCountLimit(5),
```
Trenly, line 574: For clarity and consistency, might be better to use MultiQuery type in the tests. I got a bit confused when I saw just `Query` as I thought it was supposed to be for just a single package

src/AppInstallerCLITests/Command.cpp
```diff
@@ -554,3 +566,51 @@ TEST_CASE("ParseArguments_UnknownName", "[command]")
...
+
+TEST_CASE("ParseArguments_PositionalWithMultipleValuesAndName", "[command]")
+{
+    Args args;
+    TestCommand command({
+            Argument{ "multi", 'm', Args::Type::Query, DefaultDesc, ArgumentType::Positional }.SetCountLimit(5),
+        });
+
+    std::vector<std::string> values{ "--multi", "one", "two", "three" };
+    Invocation inv{ std::vector<std::string>(values) };
```
Trenly, line 612: Is this a cast to the type that values already is?

src/AppInstallerCLITests/Command.cpp
```diff
@@ -554,3 +566,51 @@ TEST_CASE("ParseArguments_UnknownName", "[command]")
...
+
+    command.ParseArguments(inv, args);
+    RequireValuesParsedToArg({ values[1], values[2], values[3] }, command.m_args[0], args);
+}
```
Trenly, line 616: These all seem like happy-path tests; Should some be added for negative scenarios?
`ParseArguments_PositionalWithTooManyValues`
`ParseArguments_PositionalWithUndefinedCountLimit`

Trenly, line 616: I didn't see anywhere that there was a test for `Args::Type::MultiQuery` being converted to `Args::Type::Query` when there is only a single query in the multi-query. Is that something that should be tested?

src/AppInstallerCLICore/Commands/InstallCommand.cpp
florelis: I hadn't thought about that. I think that'd be a bit harder to implement because manifest is not a positional argument so we'd need to reason about the arg parsing a bit more.

I'd create a separate issue and mention it in the original issue for installing multiple apps in a single command as related.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
florelis: From looking at the implementation of `Args`, it is possible for count to be 0, in that case we also have that `GetArgs(MultiQuery)` is null and `Contains(MultiQuery)` is false.

Other args are just that, other args. You can include only those (like `id`), or you can use only (multi-)query, or you can use both. The interaction happens at the arg validation, or we just pass everything to the search query and let it figure it out.

Trenly: Thanks for the explanation!

src/AppInstallerCLICore/Workflows/WorkflowBase.h
```diff
@@ -126,12 +126,21 @@ namespace AppInstaller::CLI::Workflow
...
+    // Outputs: None
+    void CheckForMultiQuery(Execution::Context& context);
+
     // Performs a search on the source.
     // Required Args: None
     // Inputs: Source
     // Outputs: SearchResult
     void SearchSourceForMany(Execution::Context& context);
 
+    void GetSearchRequestForSingle(Execution::Context& context);
```
yao-msft, line 142: nit: missing comments

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -432,6 +432,15 @@ namespace AppInstaller::CLI::Workflow
         }
     }
 
+    void CheckForMultiQuery(Execution::Context& context)
+    {
+        if (context.Args.GetCount(Execution::Args::Type::MultiQuery) == 1)
+        {
+            context.Args.AddArg(Execution::Args::Type::Query, context.Args.GetArg(Execution::Args::Type::MultiQuery));
+            context.Args.RemoveMultiQueryArg();
+        }
```
yao-msft, line 441: Do you think the whole logic should be in ContextArg class to ensure  args integrity that whenever we remove MultiQuery we add the Query, and do that in arg parsing phase?
Or you think ContextArg should be simple and not handle  complex logic?

I personally think handling it in arg parsing phase may be easier in workflow and less error prone.

yao-msft (approved):
:shipit:

src/AppInstallerCLICore/Commands/InstallCommand.cpp
yao-msft: And since installing from manifest is more of a winget dev only feature, it's probably ok to leave this feature out and not make the workflow more complex. It's also harder for debugging if multiple manifests are being installed.

src/AppInstallerCLITests/Command.cpp
florelis: Oh, this was interesting once I figured it out. I originally just did it because it was done in the rest of the file and it didn't compile without it.

The constructor for `Invocation` takes a `std::vector<std::string>&&`, so we cannot pass `values` directly. Instead of `std::move`-ing it, we create a copy to pass to it, because we later want to use the values for the assertions.

Trenly: That is . . . interesting. Thanks!

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
florelis: I hadn't even thought about doing it in the arg parsing phase instead of the execution. That does make it a lot cleaner.

I'm moving the logic to a single method in ExecutionArgs, and calling it at the end of `Command::ParseArguments()`. It ends up being simple and in a single place, it also helps with the arg validation because I can set the categories for MultiQuery assuming that it always has more than one value.

src/AppInstallerCLITests/Command.cpp
florelis: Added a test for `ParseArguments_PositionalWithTooManyValues`. I didn't add `ParseArguments_PositionalWithUndefinedCountLimit` as the undefined count limit is the usual scenario (limit = 1).

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
florelis: This message doesn't need to be unambiguous. I'm showing it after performing the search in the sub-context, which would show the specific message like "package not found" or "multiple matches found" using the tasks you mentioned. The main purpose of this message was to show the which query had failed, as the specific messages don't show it. It's hard to show the specific query for single packages because it can be a combination of multiple things, but at least for now, with multiple queries we always have exactly one query string per package.

Anyway, I made more specific messages. It looks like this:

```
> wingetdev install doesntexist terminal
No package found matching input criteria.
Package not found: doesntexist <highlighted as warning>
Name             Id                        Source
Windows Terminal 9N0DX20HK701              msstore
Windows Terminal Microsoft.WindowsTerminal winget
Multiple packages found for: terminal <highlighted as warning>
```

emoxam:
When and how ill we get the new release ?

yao-msft:
If you are looking for stable releases, this will be included in v1.5 client release tracked by milestone here https://github.com/microsoft/winget-cli/milestone/39
It is planned to be released around May

If you are looking to test it early, you can join the Windows Insider program to get preview builds, or watch our release page here https://github.com/microsoft/winget-cli/releases

emoxam:
Where can i get Windows Insider program to get preview builds ?

denelon:
https://insider.windows.com/ 😊

Trenly:
> I mean - on github there are words Development Releases - "Join the Windows Package Manager Insider program by [signing up](http://aka.ms/winget-InsiderProgram)." and "If you decide to install the latest release from GitHub, and you have successfully joined the insider program, you will receive updates when the next development release has been published in the Microsoft Store." I've joined, and i've got mail a few days ago, but MS store has nothing to give me, i have winget -v v1.4.10173 and no update so far. What is going on ? Where is a new version? For example this one [`v1.5.101-preview` (release)](https://github.com/microsoft/winget-cli/releases/tag/v1.5.101-preview)

Preview releases aren't cut after every PR. There is internal validation and testing that has to happen before a release is cut

emoxam:
What we gave at store is 1.19.10173 - how does this relate to github versions ? There is no 1.19 at github!

Trenly:
> What we gave at store is 1.19.10173 - how does this relate to github versions ? There is no 1.19 at github!

AppInstaller contains winget, but it is not winget itself; The version marked `Latest` in Github Releases is what is bundled with the `AppInstaller` from MSStore (unless you're on an insider build)

Trenly:
> What we gave at store is 1.19.10173 - how does this relate to github versions ? There is no 1.19 at github!

AppInstaller contains winget, but it is not winget itself; The version marked `Latest` in Github Releases is what is bundled with the `AppInstaller` from MSStore (unless you're on an insider build)

emoxam:
So no hope to wait new appinstaller at store ?

florelis:
> What we gave at store is 1.19.10173 - how does this relate to github versions ? There is no 1.19 at github!

I know it doesn't seem to make sense, but there is a difference of 15 between the minor version of the App Installer package and the minor version of the winget cli. App Installer v1.19.10173 includes winget v1.4.10173.

This change will come in the preview for winget v1.5, so it will be a App Installer v1.20.

I think it takes some time to start getting preview builds after joining the insider program. Once you see you're on a v1.5 build you'll know you are receiving preview builds, so you should get the one with this change once it's out.

emoxam:
So i should look at store to see if there an update to App Installer v1.20 ?
P.S. Maybe it could be useful to write somewhere at package description what version of wiget it includes ?

src/AppInstallerCLICore/Workflows/MultiQueryFlow.cpp
g0t4: Failing is desirable so as to avoid confusion about what installed vs failed!

PR #2862, opened 2023-01-17
Refactor arg validation 

florelis:
This change adds a set of argument "categories" that tell us what an argument is used for and makes it easier to reason around which arguments can be used at the same time. For example, a manifest arg cannot be used at the same time as an arg related to sources.

The main change is adding a function to identify the categories of arguments used in an invocation, and a function to use those categories for validation.  Then, I used these new functions to replace the arg validation for commands that had similar rules.

This is what I wanted to do for #1788, and should make arg validation for #2861 easier to get right.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2862)

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -13,6 +13,14 @@ namespace AppInstaller::CLI
     using namespace Settings;
     using namespace AppInstaller::Utility::literals;
 
+    namespace
+    {
+        bool ContainsArgumentFromList(const Execution::Args& args, const std::vector<Execution::Args::Type>& argTypes)
+        {
+            return std::any_of(argTypes.begin(), argTypes.end(), [&](Execution::Args::Type arg) { return args.Contains(arg); });
+        }
+    }
+
     Argument Argument::ForType(Execution::Args::Type type)
     {
         switch (type)
@@ -54,7 +62,7 @@ namespace AppInstaller::CLI
         case Args::Type::Log:
             return Argument{ "log"_liv, 'o', Args::Type::Log, Resource::String::LogArgumentDescription, ArgumentType::Standard };
         case Args::Type::CustomSwitches:
-            return Argument{ "custom"_liv, NoAlias, Args::Type::CustomSwitches, Resource::String::CustomSwitchesArgumentDescription, ArgumentType::Standard};
+            return Argument{ "custom"_liv, NoAlias, Args::Type::CustomSwitches, Resource::String::CustomSwitchesArgumentDescription, ArgumentType::Standard };
         case Args::Type::Override:
             return Argument{ "override"_liv, NoAlias, Args::Type::Override, Resource::String::OverrideArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         case Args::Type::InstallLocation:
@@ -90,7 +98,7 @@ namespace AppInstaller::CLI
         case Args::Type::RetroStyle:
             return Argument{ "retro"_liv, NoAlias, Args::Type::RetroStyle, Resource::String::RetroArgumentDescription, ArgumentType::Flag, Argument::Visibility::Hidden };
         case Args::Type::VerboseLogs:
-            return Argument{ "verbose-logs"_liv, NoAlias, "verbose"_liv, Args::Type::VerboseLogs, Resource::String::VerboseLogsArgumentDescription, ArgumentType::Flag};
+            return Argument{ "verbose-logs"_liv, NoAlias, "verbose"_liv, Args::Type::VerboseLogs, Resource::String::VerboseLogsArgumentDescription, ArgumentType::Flag };
         case Args::Type::CustomHeader:
             return Argument{ "header"_liv, NoAlias, Args::Type::CustomHeader, Resource::String::HeaderArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         case Args::Type::AcceptSourceAgreements:
@@ -108,7 +116,7 @@ namespace AppInstaller::CLI
         case Args::Type::ProductCode:
             return Argument{ "product-code"_liv, NoAlias, Args::Type::ProductCode, Resource::String::ProductCodeArgumentDescription, ArgumentType::Standard, false };
         case Args::Type::OpenLogs:
-            return Argument{ "open-logs"_liv, NoAlias, "logs"_liv, Args::Type::OpenLogs, Resource::String::OpenLogsArgumentDescription, ArgumentType::Flag, Argument::Visibility::Help};
+            return Argument{ "open-logs"_liv, NoAlias, "logs"_liv, Args::Type::OpenLogs, Resource::String::OpenLogsArgumentDescription, ArgumentType::Flag, Argument::Visibility::Help };
         case Args::Type::UninstallPrevious:
             return Argument{ "uninstall-previous"_liv, NoAlias, Args::Type::UninstallPrevious, Resource::String::UninstallPreviousArgumentDescription, ArgumentType::Flag, ExperimentalFeature::Feature::UninstallPreviousArgument };
         case Args::Type::Force:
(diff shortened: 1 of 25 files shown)


src/AppInstallerCLICore/Argument.h
```diff
@@ -108,8 +140,11 @@ namespace AppInstaller::CLI
 
         // Static argument validation helpers; throw CommandException when validation fails.
 
-        // Requires that some form of package selection argument is present
-        static void ValidatePackageSelectionArgumentSupplied(const Execution::Args& args);
+        static ArgTypeCategory GetCategoriesPresent(const Execution::Args& arg);
+
+        // Requires that arguments meet common requirements
+        static ArgTypeCategory GetCategoriesAndValidateCommonArguments(const Execution::Args& args, bool requirePackageSelectionArg = true);
+        static void ValidateCommonArguments(const Execution::Args& args, bool requirePackageSelectionArg = true) { std::ignore = GetCategoriesAndValidateCommonArguments(args, requirePackageSelectionArg); }
```
Trenly, line 147: One thing I don't see as part of this PR is the concept of mutually exclusive sets (and there are multiple). I see mutually exclusive categories implemented here, which is great, but there's also things like `--rainbow`, `--retro`, and `--no-vt` which probably need to be mutually exclusive of one another. Or `--silent` and `--interactive`, `--version` and `--versions`, `--custom` and `--override`, `--purge` and `--preserve`, `--enable` and `--disable`. 

I do understand that these are probably more on a per-command basis, so it would be good to see a function that can be used in the command files where we can pass in a set of arg types, as well as the context and a potential error message to validate that only one argument within the set is present. Something like `ValidateExclusiveArguments`

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -145,17 +153,97 @@ namespace AppInstaller::CLI
...
-    void Argument::ValidatePackageSelectionArgumentSupplied(const Execution::Args& args)
+    ArgTypeCategory Argument::GetCategoriesPresent(const Execution::Args& args)
     {
-        for (Args::Type type : { Args::Type::Query, Args::Type::Manifest, Args::Type::Id, Args::Type::Name, Args::Type::Moniker, Args::Type::ProductCode, Args::Type::Tag, Args::Type::Command })
+        ArgTypeCategory result = ArgTypeCategory::None;
+
+        WI_SetFlagIf(result, ArgTypeCategory::Manifest, args.Contains(Args::Type::Manifest));
+
+
+        if (ContainsArgumentFromList(args, {
```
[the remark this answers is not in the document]
florelis: I added it to the category for multiple packages.

src/AppInstallerCLICore/Argument.h
florelis: Added `ValidateExclusiveArguments()` and a single use of it in the root `Command` class for `--rainbow` and `--retro`.

I didn't look closely at all the arguments as my main motivation for this change was limiting arguments that don't make sense for multiple installs in an easy way. But I tried to make it so that it would be easy to add any rules we may be missing in the future.

Trenly: Thank you!

src/AppInstallerCLICore/Argument.h
```diff
@@ -36,6 +36,38 @@ namespace AppInstaller::CLI
         Flag,
     };
 
+    // Categories an arg type can belong to.
+    // Used to reason about the arguments present without having to repeat the same
+    // lists every time.
+    enum class ArgTypeCategory
```
yao-msft, line 42: >ArgTypeCategory [](http://example.com/codeflow?start=15&length=15)

Do we want to add the "CopyInSubContext" to the category here too? The idea you previously had for copying args in subcontext.

yao-msft (approved):

src/AppInstallerCLICore/Commands/UpgradeCommand.cpp
```diff
@@ -175,30 +128,14 @@ namespace AppInstaller::CLI
...
     void UpgradeCommand::ValidateArgumentsInternal(Execution::Args& execArgs) const
     {
-        if (execArgs.Contains(Execution::Args::Type::Manifest) && 
-            (HasSearchQueryArguments(execArgs) ||
-             HasArgumentsForMultiplePackages(execArgs) ||
-             HasArgumentsForSource(execArgs)))
-        {
-            throw CommandException(Resource::String::BothManifestAndSearchQueryProvided);
-        }
+        const auto ArgCategories = Argument::GetCategoriesAndValidateCommonArguments(execArgs, /* requirePackageSelectionArg */ false);
```
yao-msft, line 131:
```suggestion
        const auto argCategories = Argument::GetCategoriesAndValidateCommonArguments(execArgs, /* requirePackageSelectionArg */ false);

```

yao-msft:
I was thinking to tag each arg with ArgTypeCategory so it's more clear and whenever we add new arg, we are more likely to think about it.

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -145,17 +168,121 @@ namespace AppInstaller::CLI
...
+        WI_SetFlagIf(result, ArgTypeCategory::SinglePackageQuery, ContainsArgumentFromList(args, {
+            Args::Type::Query,
+            Args::Type::Id,
+            Args::Type::Name,
+            Args::Type::Moniker,
+            Args::Type::ProductCode,
+            Args::Type::Tag,
+            Args::Type::Command,
+            Args::Type::Version,
+            Args::Type::Channel,
```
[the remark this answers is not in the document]
Trenly: Technically couldn't `Tag` or `Command` also?
```
PS C:\WINDOWS\system32> winget search --tag chrome
Name                       Id                         Version      Match
----------------------------------------------------------------------------------------
Chrome Remote Desktop Host Google.ChromeRemoteDesktop 110.0.5481.7 Tag: chrome
Inssist                    SlashedIo.Inssist          16.1.0       Tag: chrome-extension

PS C:\WINDOWS\system32> winget search --command pdf
Name         Id                    Version  Match
wkhtmltox    wkhtmltopdf.wkhtmltox 0.12.6-1 Command: wkhtmltopdf
SumatraPDF   SumatraPDF.SumatraPDF 3.4.6    Command: sumatrapdf
PDFtk Server PDFLabs.PDFtk.Server  2.02     Command: pdftk
PDFtk Free   PDFLabs.PDFtk.Free    2.02     Command: PdftkXp
```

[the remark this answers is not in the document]
florelis: I didn't think of that. Now I can see how channels could be somewhat standardized and not completely tied to each package, like "release" or "beta".

I can also see it for the tags, multiple packages sharing the same one, but not so much for the command. Command seems like you would only want one instance of each command, even though for search it looks for substring instead of full match.

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -107,11 +107,12 @@ namespace AppInstaller::CLI
 
     void PinAddCommand::ValidateArgumentsInternal(Execution::Args& execArgs) const
     {
+        Argument::ValidateCommonArguments(execArgs);
+
         if (execArgs.Contains(Execution::Args::Type::GatedVersion) && execArgs.Contains(Execution::Args::Type::BlockingPin))
         {
             throw CommandException(Resource::String::BothGatedVersionAndBlockingFlagProvided);
```
Trenly, line 114: I absolutely love what you did with the string builder for the exclusive arguments, it looks great and once this PR gets merged I can take a look at what other exclusive arguments need to be validated. 

I know that this specific line isn't part of the PR, but these scenarios could also be a good fit for validating exclusive arguments. I don't think it needs to be added in this PR, I can try and do it on my own, but I was wondering how you would recommend going about it. Would it be best to create a function overload that accepted the string ID as a third parameter, or would it be better to implement it as an optional defaulting to nullopt?

src/AppInstallerCLICore/Argument.cpp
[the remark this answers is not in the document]
Trenly: > Command seems like you would only want one instance of each command, even though for search it looks for substring instead of full match.

Even with `-e`, there are still multiple packages that use the same command. Visual studio is a great example. Even if channels are implemented to allow for the abstraction of Professional/Community/Enterprise, the different versions such as 2019/2022 use the same command -
```
PS C:\WINDOWS\system32> winget search --command devenv -e
Name                                    Id                                               Version             Match
----------------------------------------------------------------------------------------------------------------------------
Visual Studio Professional 2022 Preview Microsoft.VisualStudio.2022.Professional-Preview 17.0.0 Preview 2.1  Command: devenv
Visual Studio Professional 2022         Microsoft.VisualStudio.2022.Professional         17.4.4              Command: devenv
Visual Studio Enterprise 2022 Preview   Microsoft.VisualStudio.2022.Enterprise-Preview   17.0.0 Preview 2.1  Command: devenv
Visual Studio Community 2022 Preview    Microsoft.VisualStudio.2022.Community-Preview    17.0.0 Preview 2.1  Command: devenv
Visual Studio Professional 2019 Preview Microsoft.VisualStudio.2019.Professional-Preview 16.11.0 Preview 3.0 Command: devenv
Visual Studio Professional 2019         Microsoft.VisualStudio.2019.Professional         16.11.23            Command: devenv
Visual Studio Enterprise 2019 Preview   Microsoft.VisualStudio.2019.Enterprise-Preview   16.11.0 Preview 3.0 Command: devenv
Visual Studio Community 2019 Preview    Microsoft.VisualStudio.2019.Community-Preview    16.11.0 Preview 3.0 Command: devenv
Visual Studio Community 2022            Microsoft.VisualStudio.2022.Community            17.4.4              Command: devenv
Visual Studio Team Explorer 2022        Microsoft.VisualStudio.2022.TeamExplorer         17.4.4              Command: devenv
Visual Studio Load Test Controller 2022 Microsoft.VisualStudio.2022.TestController       17.4.4              Command: devenv
Visual Studio Community 2019            Microsoft.VisualStudio.2019.Community            16.11.23            Command: devenv
Visual Studio Team Explorer 2019        Microsoft.VisualStudio.2019.TeamExplorer         16.11.23            Command: devenv
Visual Studio Load Test Controller 2019 Microsoft.VisualStudio.2019.TestController       16.11.23            Command: devenv
```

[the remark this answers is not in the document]
yao-msft: I think we would categorize them based on user intent (Command seems like it's for single package). Or just put it in PackageQuery if we are not sure. The previous code just tries to validate that --manifest and PackageQuery should not exist together. Either way I'm not too worried.

florelis:
I saw the comments about exclusive sets and using the categories in the context cloning, but I missed the part where it said it wasn't needed in this PR :D
So maybe I spent more in this than I should have but anyways...

I added a new type for the couple of fields in an argument that I'm using here: the name (for error messages), the categories and a new bitmask for exclusive sets. Since I added the name, I also put the alternate name and alias in here.
Then I added a `ForType()` function for this class, like the one that exists for `Argument`, but I added a test to enforce that all types are listed in this function. Using this function we can get all the information we needed to drive checking incompatible categories, exclusive sets, and we are now able to list the arg name in error messages for any arg.
The more complete `Argument` type now wraps this new smaller type to provide the name/alias, but we can still use a different help string or setting or policy on each command that uses the argument. Maybe it should have extended instead of wrapping, and maybe some other fields like the argument type (positional/flag) should have also been moved to the new smaller type.
My change adds the limitation that if two commands use the same arg type, it has to have the same name, but I don't think that will be an issue.

Trenly:
> I saw the comments about exclusive sets and using the categories in the context cloning, but I missed the part where it said it wasn't needed in this PR :D So maybe I spent more in this than I should have but anyways...
> 
> I added a new type for the couple of fields in an argument that I'm using here: the name (for error messages), the categories and a new bitmask for exclusive sets. Since I added the name, I also put the alternate name and alias in here. Then I added a `ForType()` function for this class, like the one that exists for `Argument`, but I added a test to enforce that all types are listed in this function. Using this function we can get all the information we needed to drive checking incompatible categories, exclusive sets, and we are now able to list the arg name in error messages for any arg. The more complete `Argument` type now wraps this new smaller type to provide the name/alias, but we can still use a different help string or setting or policy on each command that uses the argument. Maybe it should have extended instead of wrapping, and maybe some other fields like the argument type (positional/flag) should have also been moved to the new smaller type. My change adds the limitation that if two commands use the same arg type, it has to have the same name, but I don't think that will be an issue.

Wow, that's way more than I would have ever expected! A cursory glance looks amazing, but I'll look closer once I've got some time

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1653,4 +1641,11 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Unable to open pin database.</value>
     <comment>Error message for when we cannot open the database containing package pins.</comment>
   </data>
+  <data name="ArgumentForSinglePackageProvidedWithMultipleQueries" xml:space="preserve">
+    <value>An argument was provided that can only be used for single packages</value>
```
yao-msft, line 1645:
```suggestion
    <value>An argument was provided that can only be used for single package</value>

```

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -127,35 +323,98 @@ namespace AppInstaller::CLI
...
+
+            if (argsFromSet.size() > 1)
+            {
+                // Create a string showing the exclusive args.
+                std::string argsString;
+                for (const auto& arg : argsFromSet)
+                {
+                    if (!argsString.empty())
+                    {
+                        argsString += "|";
```
yao-msft, line 366: nit:
```suggestion
                        argsString += '|';

```

yao-msft (approved):

PR #2863, opened 2023-01-17
WinGetPackage DSC Resource

msftrubengu:
This is a preview implementation of a DSC resource that installs packages on winget. 

For now, ensures a package is installed or not. Its install behavior depends if the package is already installed for the user or not.

This resource is not final and it just provides a simple implementation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2863)


src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -304,4 +310,175 @@ class WinGetSourcesResource
...
+            throw "Id is required"
+        }
+
+        $wingetPackageResource = [WinGetPackage]::new()
+        $wingetPackageResource.Id = $this.Id
+
+        # We can't have CatalogPackage as member of this class because it doesn't have a default constructor.
+        # A call to Invoke-DscResource will fail with
+        # 'ImportClassResourcesFromModule: Exception calling "ImportClassResourcesFromModule" with "4" argument(s): "The DSC resource 'CatalogPackage' has no default constructor."'
+        $catalogPackage = Get-WinGetPackage -Id $this.Id -Exact
```
JohnMcPMS, line 348: `Exact` is not the correct matching strategy; it requires the case to match.  What we want is `EqualsCaseInsensitive`, but that isn't exposed via PS currently.  We alternate between that and `ContainsCaseInsensitive` in `winget` commands based on whether the command is targeting a single package or multiple, but the PS commands don't carry the actions that are intended during the search phase.

The matching strategy parameter needs to be changed to expose all of the options, probably as a parameter that takes an enumeration value.

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -304,4 +310,175 @@ class WinGetSourcesResource
...
+            if ($this.Ensure -eq [Ensure]::Present)
+            {
+                if (-not([string]::IsNullOrWhiteSpace($this.Source)))
+                {
+                    $hashArgs.Add("Source", $this.Source)
+                }
+
+                if ($this.Installed)
+                {
+                    $catalogPackage = Get-WinGetPackage -Id $this.Id -Exact
```
JohnMcPMS, line 436: This seems extremely inefficient.  Couldn't you make an internal type/hashtable that contains all of the relevant information and get it just once?

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -304,4 +310,175 @@ class WinGetSourcesResource
...
+
+                    if ([string]::IsNullOrWhiteSpace($this.Version))
+                    {
+                        Update-WinGetPackage @hashArgs
```
JohnMcPMS, line 440: Again, I don't see how not specifying a version means "the latest version" rather than "I don't care what the version is".

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -304,4 +310,175 @@ class WinGetSourcesResource
...
+                            {
+                                # Installed package has a lower version. Update.
+                                # TODO: if update fails we should uninstall and install.
+                                Update-WinGetPackage @hashArgs
+                                break
+                            }
+                            {'Greater' -or 'Unknown'}
+                            {
+                                # The installed package has a greated version or unknown. Uninstall and install.
+                                Uninstall-WinGetPackage -Id $this.Id
```
JohnMcPMS, line 459: We don't have a pipelineable parameter to just take in the package object?  This call is not safe as written (ignores source).

JohnMcPMS (changes requested):

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -398,4 +419,220 @@ class WinGetPackageManager
...
+    [bool]$IsUpdateAvailable = $false
+
+    [PSObject] hidden $CatalogPackage = $null
+
+    hidden Initialize()
+    {
+        # DSC only validates keys and mandatories in a Set call.
+        if ([string]::IsNullOrWhiteSpace($this.Id))
+        {
+            throw "WinGetPackage: Id is required"
```
JohnMcPMS, line 462: //TODO: Localize this and other error messages

JohnMcPMS (approved):

PR #2897, opened 2023-01-31
Expose Agreements and Locale Fields in COM

ryfu-msft:
The changes in this PR expose all of the locale fields including agreements. Also provides a way for the COM caller to accept or reject package/source agreements. To avoid breaking existing usage, the default behavior for accepting source/package agreements has been set to true.

Changes:
- Modifies InstallOptions to expose ```AcceptPackageAgreements(bool)```
- Modifies PackageCatalogReference to expose ````AcceptSourceAgreements(bool)````
- Modifies PackageVersionInfo to expose two methods for obtaining a CatalogPackageMetadata object which contains the locale fields (including agreements). ```GetCatalogPackageMetadata()``` with no parameters uses the default localization, while ```GetCatalogPackageMetadata(string locale)``` applies the specified locale and retrieves those locale fields. 
Tests:

Added tests to verify that all locale fields and agreements are outputted correctly. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2897)


src/Microsoft.Management.Deployment/CatalogPackageMetadata.h
```diff
@@ -0,0 +1,65 @@
...
+        hstring InstallationNotes();
+
+#if !defined(INCLUDE_ONLY_INTERFACE_METHODS)
+    private:
+        ::AppInstaller::Manifest::ManifestLocalization m_manifestLocalization;
+        Windows::Foundation::Collections::IVector<winrt::Microsoft::Management::Deployment::PackageAgreement> m_packageAgreements{ nullptr };
+        Windows::Foundation::Collections::IVector<winrt::Microsoft::Management::Deployment::Documentation> m_documentations{ nullptr };
+        Windows::Foundation::Collections::IVector<hstring> m_tags{ nullptr };
+#endif
+    };
```
Trenly, line 64: PurchaseURL?
Moniker?

JohnMcPMS (changes requested):

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -643,6 +650,15 @@ namespace Microsoft.Management.Deployment
             /// A string that will be passed to the source server if using a REST source
             String AdditionalPackageCatalogArguments;
         }
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 6)]
+        {
+            /// Gets the required agreements for connecting to the package catalog (source).
+            Windows.Foundation.Collections.IVectorView<SourceAgreement> SourceAgreements{ get; };
+
+            Windows.Foundation.IAsyncOperation<ConnectResult> ConnectAsync(PackageCatalogConnectOptions packageCatalogConnectOptions);
```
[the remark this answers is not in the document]
ryfu-msft: Removed new connect methods and moved AcceptSourceAgreements() to PackageCatalogReference

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -459,6 +459,13 @@ namespace Microsoft.Management.Deployment
         /// IsSame from IPackage in winget/RepositorySearch is not implemented in V1.
         /// Determines if the given IPackage refers to the same package as this one.
         /// virtual bool IsSame(const IPackage*) const = 0;
+        
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 6)]
+        {
+            CatalogPackageMetadata GetCatalogPackageMetadata(PackageVersionId versionKey, String locale);
```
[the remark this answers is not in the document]
ryfu-msft: Moved method to PackageVersionInfo class and added an additional method that takes in no arguments which uses the default localization.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -459,6 +459,13 @@ namespace Microsoft.Management.Deployment
         /// IsSame from IPackage in winget/RepositorySearch is not implemented in V1.
         /// Determines if the given IPackage refers to the same package as this one.
         /// virtual bool IsSame(const IPackage*) const = 0;
+        
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 6)]
+        {
+            CatalogPackageMetadata GetCatalogPackageMetadata(PackageVersionId versionKey, String locale);
+
+            Windows.Foundation.Collections.IVectorView<CatalogPackageMetadata> GetAllCatalogPackageMetadata(PackageVersionId versionKey);
```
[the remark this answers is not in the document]
ryfu-msft: Removed as GetCatalogPackageMetadata should now cover this scenario.

src/Microsoft.Management.Deployment/PackageVersionInfo.cpp
```diff
@@ -138,4 +139,33 @@ namespace winrt::Microsoft::Management::Deployment::implementation
     { 
         return m_packageVersion; 
     }
+    Microsoft::Management::Deployment::CatalogPackageMetadata PackageVersionInfo::GetCatalogPackageMetadata()
+    {
+        auto catalogPackageMetadata = winrt::make_self<wil::details::module_count_wrapper<winrt::Microsoft::Management::Deployment::implementation::CatalogPackageMetadata>>();
+        if (m_packageVersion)
+        {
+            auto manifest = m_packageVersion->GetManifest();
+            catalogPackageMetadata->Initialize(manifest.DefaultLocalization);
```
JohnMcPMS, line 148: As you have implemented this, there is no way to get the values for the locale as configured for the user without the caller having to determine what that locale is.  Was that intentional?  I don't think we should require the caller to do that work unless there is a real reason to force it (and remember that they won't know the settings values).

This function returning the default locale is not useful; I think you misinterpreted my previous comment.  I was referring to the default localization as a concept, not `DefaultLocalization`.  The one that would be used by default within winget, based on the user/system configuration.

I'm fine with not allowing an empty/null string into the other method, but then this one should be doing the proper determination of what the default locale is and then use it.  That is done with `manifest.ApplyLocale()`.

JohnMcPMS (changes requested):
I haven't looked at the implementation really, was just reviewing the interface changes.

src/Microsoft.Management.Deployment/PackageVersionInfo.cpp
ryfu-msft: I definitely misunderstood what you meant by `DefaultLocalization` 😅 

I decided to keep both methods but changed the one without a parameter to point to the CurrentLocalization after calling `manifest.ApplyLocale()` as suggested. That way the caller can get the default locale values without needing to determine what that locale is and still have the option to get other locale values if needed.

src/AppInstallerCLIE2ETests/Interop/InstallInterop.cs
```diff
@@ -516,5 +517,51 @@ public async Task InstallRequireUserScopeAndUnknown()
...
+            var installOptions = this.TestFactory.CreateInstallOptions();
+            installOptions.PackageInstallMode = PackageInstallMode.Silent;
+            installOptions.PreferredInstallLocation = this.installDir;
+            installOptions.AcceptPackageAgreements = false;
+
+            // Install
+            var installResult = await this.packageManager.InstallPackageAsync(searchResult.CatalogPackage, installOptions);
+
+            // Assert
+            Assert.AreEqual(InstallResultStatus.CatalogError, installResult.Status);
```
yao-msft, line 564: >Assert.AreEqual(InstallResultStatus.CatalogError, installResult.Status); [](http://example.com/codeflow?start=12&length=72)

I don't think CatalogError should be returned if agreements were not accepted

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -562,6 +567,10 @@ namespace winrt::Microsoft::Management::Deployment::implementation
     {
         // Add installed version
         AddInstalledVersionToContext(package.InstalledVersion(), comContext.get());
+
+        // Uninstall should not require agreements acceptance.
```
yao-msft, line 571: Uninstall should never check for agreements, so this code is not needed

src/AppInstallerCLICore/COMContext.h
```diff
@@ -40,14 +40,12 @@ namespace AppInstaller::CLI::Execution
         {
             Reporter.SetChannel(Reporter::Channel::Disabled);
             Reporter.SetProgressSink(this);
-            SetFlags(CLI::Execution::ContextFlag::AgreementsAcceptedByCaller);
```
yao-msft, line 43: ContextFlag::AgreementsAcceptedByCaller is specifically created for ComContext, since ComContext is no longer using it, this falg definition should be removed from Context

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -401,6 +401,11 @@ namespace winrt::Microsoft::Management::Deployment::implementation
             }
 
             // Note: AdditionalPackageCatalogArguments is not needed during install since the manifest is already known so no additional calls to the source are needed. The property is deprecated.
+
+            if (options.AcceptPackageAgreements())
```
yao-msft, line 405: If InstallOptions is null, we are breaking previous behavior

src/Microsoft.Management.Deployment/PackageVersionInfo.h
```diff
@@ -24,8 +24,12 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         winrt::Windows::Foundation::Collections::IVectorView<hstring> ProductCodes();
         winrt::Microsoft::Management::Deployment::PackageCatalog PackageCatalog();
         winrt::Microsoft::Management::Deployment::CompareResult CompareToVersion(hstring versionString);
```
yao-msft, line 26: nit: should this be `const hstring&` too?

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -826,6 +846,84 @@ namespace Microsoft.Management.Deployment
         }
     }
 
+    /// IMPLEMENTATION NOTE: SourceOrigin from AppInstaller::Repository::Documentation
```
yao-msft, line 849: Update the comments here and below

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -54,6 +55,11 @@ namespace winrt::Microsoft::Management::Deployment::implementation
                 return GetConnectCatalogErrorResult();
             }
 
+            if (!m_acceptSourceAgreements)
+            {
+                return GetConnectCatalogErrorResult();
```
yao-msft, line 60: Should we fail only if !AcceptAgreements && Agreements exist?

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -124,7 +130,17 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         }
         return GetConnectCatalogErrorResult();
     }
+    winrt::Windows::Foundation::Collections::IVectorView<winrt::Microsoft::Management::Deployment::SourceAgreement> PackageCatalogReference::SourceAgreements()
+    {
+        for (auto const& agreement : m_sourceReference.GetInformation().SourceAgreements)
```
yao-msft, line 135: This logic should be guarded by once_flag, otherwise if caller calls this api multiple times, the m_sourceagreements keeps adding duplicate content, and it's not thread safe.

src/Microsoft.Management.Deployment/PackageAgreement.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "pch.h"
```
yao-msft, line 4: Don't include pch.h in header

src/Microsoft.Management.Deployment/PackageAgreement.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "pch.h"
+#include "PackageAgreement.h"
+#include "PackageAgreement.g.cpp"
+#include <wil\cppwinrt_wrl.h>
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    void PackageAgreement::Initialize(::AppInstaller::Manifest::Agreement packageAgreement)
+    {
+        m_packageAgreement = packageAgreement;
```
yao-msft, line 12:
```suggestion
        m_packageAgreement = std::move(packageAgreement);
```

src/Microsoft.Management.Deployment/Documentation.h
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "pch.h"
```
yao-msft, line 4: remove pch.h

src/Microsoft.Management.Deployment/Documentation.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "pch.h"
+#include "Documentation.g.cpp"
+#include "Documentation.h"
+#include <wil\cppwinrt_wrl.h>
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    void Documentation::Initialize(::AppInstaller::Manifest::Documentation documentation)
+    {
+        m_documentation = documentation;
```
yao-msft, line 12: std::move

src/Microsoft.Management.Deployment/CatalogPackageMetadata.cpp
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include <winget/RepositorySource.h>
+#include "CatalogPackageMetadata.h"
+#include "CatalogPackageMetadata.g.cpp"
+#include "Documentation.h"
```
yao-msft, line 7: nit: Documentation.h already included in header

src/Microsoft.Management.Deployment/CatalogPackageMetadata.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <wil\cppwinrt_wrl.h>
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    using Localization = ::AppInstaller::Manifest::Localization;
+
+    void CatalogPackageMetadata::Initialize(::AppInstaller::Manifest::ManifestLocalization manifestLocalization)
+    {
+        m_manifestLocalization = manifestLocalization;
```
yao-msft, line 16:
```suggestion
        m_manifestLocalization = std::move(manifestLocalization);
```

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -280,6 +280,15 @@ namespace Microsoft.Management.Deployment
         /// a lot of fields and no one requesting it.
         /// Gets the manifest of this package version.
         /// virtual Manifest::Manifest GetManifest() = 0;
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 6)]
+        {
+            /// Gets the package catalog metadata of this package version with the default localization based on user settings.
+            CatalogPackageMetadata GetCatalogPackageMetadata();
+
+            /// Gets the package catalog metadata of this package version with the specified locale.
```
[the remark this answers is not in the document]
ryfu-msft: Changed to preferredLocale and updated comments

src/Microsoft.Management.Deployment/PackageManager.cpp
ryfu-msft: Removed

src/AppInstallerCLICore/COMContext.h
ryfu-msft: Removed Agreementsacceptedbycaller flag entirely.

src/Microsoft.Management.Deployment/PackageManager.cpp
ryfu-msft: Modified this so that if installoptions is null we assume the package agreements are accepted as to not break previous behavior

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
ryfu-msft: Changed condition to suggested

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
ryfu-msft: added once-flag to guard this logic.

src/AppInstallerCLIE2ETests/Interop/InstallInterop.cs
ryfu-msft: Changed to new error - 'PackageAgreementsNotAccepted'

src/Microsoft.Management.Deployment/CatalogPackage.cpp
```diff
@@ -5,13 +5,15 @@
 #include <winget/RepositorySource.h>
 #include "CatalogPackage.h"
 #include "CatalogPackage.g.cpp"
+#include "CatalogPackageMetadata.h"
```
yao-msft, line 8: nit: does not seem used

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -631,6 +641,10 @@ namespace Microsoft.Management.Deployment
     {
         Ok,
         CatalogError,
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 6)]
+        {
+            SourceAgreementsNotAccepted,
+    }
```
yao-msft, line 647: nit: align?

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -663,6 +678,14 @@ namespace Microsoft.Management.Deployment
             /// A string that will be passed to the source server if using a REST source
             String AdditionalPackageCatalogArguments;
         }
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 6)]
+        {
+            /// Gets the required agreements for connecting to the package catalog (source).
+            Windows.Foundation.Collections.IVectorView<SourceAgreement> SourceAgreements { get; };
+
+            Boolean AcceptSourceAgreements{ get; set; };
```
yao-msft, line 687:
```suggestion
            Boolean AcceptSourceAgreements;

```

src/Microsoft.Management.Deployment/Converters.h
```diff
@@ -82,6 +82,9 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         case APPINSTALLER_CLI_ERROR_INTERNAL_ERROR:
             resultStatus = TStatus::InternalError;
             break;
+        case APPINSTALLER_CLI_ERROR_PACKAGE_AGREEMENTS_NOT_ACCEPTED:
```
yao-msft, line 85: nit: move above InternalError section

yao-msft (approved):

PR #2923, opened 2023-02-04
Update zlib library in Pure project

ryfu-msft:
The zlib library included from the Pure library is outdated and triggering a component governance issue. Since the library has not been updated in a while, I think it is best to manually update the library using the `git subtree` command and having the subtree point to the latest zlib version (1.2.13)

Git command executed: `git subtree add --prefix=src/PureLib/pure/zlib https://github.com/madler/zlib.git 04f42ceca40f73e2978b50e93806c2a18c1281fc --squash`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/winget-cli/pulls/2923)


florelis (approved):
Looks like the build is failing when trying to restore one of the projects in the updated lib. If we don't need to build it, we can make the dotnet restore step more specific to only target what we need, or we can remove the csproj so we don't even see it. Otherwise, we may need to add a different restore task for this, or change the csproj.

ryfu-msft:
> Looks like the build is failing when trying to restore one of the projects in the updated lib. If we don't need to build it, we can make the dotnet restore step more specific to only target what we need, or we can remove the csproj so we don't even see it. Otherwise, we may need to add a different restore task for this, or change the csproj.

This issue has been resolved. Just needed to open the project files in the latest Visual Studio and have it update.

PR #2925, opened 2023-02-06
Add how to install from MS Store using PowerShell

JuryA:

-----
Adding new installation option to `README.md`: **PowerShell script for installation from Microsoft Store** utilizing **CIM MDM**.

# Rationale
Manual installation of **winget-cli** from Microsoft Store is easy, fast, and straightforward, but automated installation (unattended) - although possible with a lot of pain, manual dependency hell solving and finally deployed without automatic updates, etc. - is (as described in the documentation) unusable and with a lot of problems.

I propose a solution that de facto mimics manual GUI attended installation from Microsoft Store leveraging CIM MDM + PowerShell. This solution is clean, quick, reusable, and can be used as part of deployment scripts/solutions including Microsoft DSC with fully functional automatic updates, and automatic dependency resolving (by MS Store) and should work no matter what changes (major/minor) will be in next versions.

_(Inspired by an awesome script [Install-CompanyPortal.ps1](https://gist.github.com/adotcoop/0241d371684c3771000385dd93da77e4#file-install-companyportal-ps1) from @adotcoop! Thank you!)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,50 @@ If you are new to the Windows Package Manager, you might want to [Explore the Wi
 
 ### Microsoft Store [Recommended]
 
-The client is distributed within the [App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) package. 
+The client is distributed within the [App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) package.
+
+### Microsoft Store using PowerShell
+
+```PowerShell
+$applicationId = "9NBLGGH4NNS1" # Microsoft.DesktopAppInstaller_8wekyb3d8bbwe
+$skuId = 0016
+
+# Obtain packageFamilyName from applicationId
+$webpage = Invoke-WebRequest -UseBasicParsing -Uri "https://bspmts.mp.microsoft.com/v1/public/catalog/Retail/Products/$applicationId/applockerdata"
+$packageFamilyName = ($webpage | ConvertFrom-JSON).packageFamilyName
+
+# Prepare CIM session object
+$namespaceName = "root\cimv2\mdm\dmmap"
+$session = New-CimSession
+$omaUri = "./Vendor/MSFT/EnterpriseModernAppManagement/AppInstallation"
+$newInstance = New-Object Microsoft.Management.Infrastructure.CimInstance "MDM_EnterpriseModernAppManagement_AppInstallation01_01", $namespaceName
+$property = [Microsoft.Management.Infrastructure.CimProperty]::Create("ParentID", $omaUri, "string", "Key")
+$newInstance.CimInstanceProperties.Add($property)
+$property = [Microsoft.Management.Infrastructure.CimProperty]::Create("InstanceID", $packageFamilyName, "String", "Key")
+$newInstance.CimInstanceProperties.Add($property)
+
+# Set CIM session parameters
+$flags = 0
+$paramValue = [Security.SecurityElement]::Escape($('<Application id="{0}" flags="{1}" skuid="{2}"/>' -f $applicationId, $flags, $skuId))
+$params = New-Object Microsoft.Management.Infrastructure.CimMethodParametersCollection
+$param = [Microsoft.Management.Infrastructure.CimMethodParameter]::Create("param", $paramValue, "String", "In")
+$params.Add($param)
+
+try {
+    # we create the MDM instance and trigger the StoreInstallMethod
+    $instance = $session.CreateInstance($namespaceName, $newInstance)
+    $result = $session.InvokeMethod($namespaceName, $instance, "StoreInstallMethod", $params)
+}
+catch [Exception] {
+    write-host $_ | out-string
+}
+
+# Dispose CIM session object
+Remove-CimSession -CimSession $session
+
+# Wait for a while - application is installed from Microsoft Store on the background
+```
+*Credit: [@JuryA](https://github.com/JuryA), inspired by [@adotcoop](https://gist.github.com/adotcoop) and his [Install-CompanyPortal.ps1](https://gist.github.com/adotcoop/0241d371684c3771000385dd93da77e4#file-install-companyportal-ps1)*
 
 ### Development Releases
 


JuryA:
> 

Trenly:
I think this would probably be better as a separate file, so that it can be downloaded and run directly or called using `Invoke-WebRequest`. Including large blocks of script in README is often a deterrent for non-technical people to get the information they are looking for. 

Another point to consider is that this relies on an API, which could be deprecated or updated at any point in time. Since documentation in the readme is often one of the lower priority items, it could be a while before anyone notices the install script is broken; If the script is in a separate file then it would be easier for users to create an issue against, since it would be clearer where the exact issue is.

README.md
```diff
@@ -12,7 +12,50 @@ If you are new to the Windows Package Manager, you might want to [Explore the Wi
...
-The client is distributed within the [App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) package. 
+The client is distributed within the [App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) package.
+
+### Microsoft Store using PowerShell
+
+```PowerShell
+$applicationId = "9NBLGGH4NNS1" # Microsoft.DesktopAppInstaller_8wekyb3d8bbwe
+$skuId = 0016
+
+# Obtain packageFamilyName from applicationId
```
Trenly, line 23: Why do we need to obtain the packageFamilyName if it is already known?

README.md
```diff
@@ -12,7 +12,50 @@ If you are new to the Windows Package Manager, you might want to [Explore the Wi
...
+$webpage = Invoke-WebRequest -UseBasicParsing -Uri "https://bspmts.mp.microsoft.com/v1/public/catalog/Retail/Products/$applicationId/applockerdata"
+$packageFamilyName = ($webpage | ConvertFrom-JSON).packageFamilyName
```
Trenly, line 25: `Invoke-RestMethod` will automatically parse the JSON rather than requiring an explicit call to `ConvertFrom-JSON`

README.md
```diff
@@ -12,7 +12,50 @@ If you are new to the Windows Package Manager, you might want to [Explore the Wi
...
+$session = New-CimSession
+$omaUri = "./Vendor/MSFT/EnterpriseModernAppManagement/AppInstallation"
+$newInstance = New-Object Microsoft.Management.Infrastructure.CimInstance "MDM_EnterpriseModernAppManagement_AppInstallation01_01", $namespaceName
+$property = [Microsoft.Management.Infrastructure.CimProperty]::Create("ParentID", $omaUri, "string", "Key")
+$newInstance.CimInstanceProperties.Add($property)
+$property = [Microsoft.Management.Infrastructure.CimProperty]::Create("InstanceID", $packageFamilyName, "String", "Key")
+$newInstance.CimInstanceProperties.Add($property)
+
+# Set CIM session parameters
+$flags = 0
```
Trenly, line 38: Why create a variable here when it's only used once later?

README.md
```diff
@@ -12,7 +12,50 @@ If you are new to the Windows Package Manager, you might want to [Explore the Wi
...
 ### Microsoft Store [Recommended]
 
-The client is distributed within the [App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) package. 
+The client is distributed within the [App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) package.
+
+### Microsoft Store using PowerShell
+
+```PowerShell
+$applicationId = "9NBLGGH4NNS1" # Microsoft.DesktopAppInstaller_8wekyb3d8bbwe
+$skuId = 0016
```
Trenly, line 21: Why create a variable here when it's only used once later?

README.md
```diff
@@ -12,7 +12,50 @@ If you are new to the Windows Package Manager, you might want to [Explore the Wi
...
+
+# Obtain packageFamilyName from applicationId
+$webpage = Invoke-WebRequest -UseBasicParsing -Uri "https://bspmts.mp.microsoft.com/v1/public/catalog/Retail/Products/$applicationId/applockerdata"
+$packageFamilyName = ($webpage | ConvertFrom-JSON).packageFamilyName
+
+# Prepare CIM session object
+$namespaceName = "root\cimv2\mdm\dmmap"
+$session = New-CimSession
+$omaUri = "./Vendor/MSFT/EnterpriseModernAppManagement/AppInstallation"
```
Trenly, line 30: Why create a variable here when it's only used once later?

Trenly (changes requested):

README.md
JuryA: > I think this would probably be better as a separate file, so that it can be downloaded and run directly or called using `Invoke-WebRequest`. Including large blocks of script in README is often a deterrent for non-technical people to get the information they are looking for.

I agree - let me prepare a separate installation script. I suggest using simple `install.ps1` - is that ok?
 
> Another point to consider is that this relies on an API, which could be deprecated or updated at any point in time. Since documentation in the readme is often one of the lower priority items, it could be a while before anyone notices the install script is broken; If the script is in a separate file then it would be easier for users to create an issue against, since it would be clearer where the exact issue is.

Good point! Installation script in the dedicated file will have better maintainability.

JuryA:
> I think this would probably be better as a separate file, so that it can be downloaded and run directly or called using `Invoke-WebRequest`. Including large blocks of script in README is often a deterrent for non-technical people to get the information they are looking for.

I agree - let me prepare a separate installation script. I suggest using simple `install.ps1` - is that ok?
 
> Another point to consider is that this relies on an API, which could be deprecated or updated at any point in time. Since documentation in the readme is often one of the lower priority items, it could be a while before anyone notices the install script is broken; If the script is in a separate file then it would be easier for users to create an issue against, since it would be clearer where the exact issue is.

Good point! Installation script in the dedicated file will have better maintainability.

README.md
```diff
@@ -12,7 +12,50 @@ If you are new to the Windows Package Manager, you might want to [Explore the Wi
...
+The client is distributed within the [App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) package.
+
+### Microsoft Store using PowerShell
+
+```PowerShell
+$applicationId = "9NBLGGH4NNS1" # Microsoft.DesktopAppInstaller_8wekyb3d8bbwe
+$skuId = 0016
+
+# Obtain packageFamilyName from applicationId
+$webpage = Invoke-WebRequest -UseBasicParsing -Uri "https://bspmts.mp.microsoft.com/v1/public/catalog/Retail/Products/$applicationId/applockerdata"
```
yao-msft, line 24: I'm not sure if talking directly to this endpoint would be advertised by us.

yao-msft (commented):
Winget team is working on mechanisms to make sure AppInstaller/Winget is installed on a target machine as part of some future features. I'm not sure if using MDM would be our advertised way of doing that. We'll post back after the team has reached a conclusion. Thanks.

JuryA:
> Winget team is working on mechanisms to make sure AppInstaller/Winget is installed on a target machine as part of some future features. I'm not sure if using MDM would be our advertised way of doing that. We'll post back after the team has reached a conclusion. Thanks.

I understand your point but - for now - it's the cleanest way how to deploy AppInstaller/Winget correctly, with all dependencies and without any complicated hacks. This is a really critical thing. Maybe better - highly critical. I don't see anything bad to use MDM for this purpose - it's the turn of Microsoft to prepare a better way to do that. Users don't have time to sit and wait for a "better endpoint" - despite that this one is very well documented and used (in advanced Intune tasks). I'm still wondering why Microsoft Store doesn't have API for application deployment... Is that intended to avoid something? This problem is here for a long time and - to be honest - installation of Winget if - for correct installation - we need Winget ... this is the classic egg-chicken problem... But thank you very much for opening the discussion internally - really appreciate it. Meanwhile, I prepare changes proposed by @Trenly and wait for your statement.

README.md
JuryA: This is not needed... We can hardcode `packageFamilyName` directly as:

```PowerShell
$packageFamilyName = 'Microsoft.Microsoft.DesktopAppInstaller_8wekyb3d8bbwe'
```

yao-msft:
Thanks for clarification. @denelon, would we add "installing through MDM from MSStore" script and referenced in Readme.md as temporary solution before our "make sure winget is installed" work is done?

denelon:
@yao-msft,
Is the "Repair-WinGetPackageManager" cmdlet in the "Microsoft.WinGet.Client" PowerShell module fully functioning with the new changes to the UI.Xaml package?

I know we needed to add some clarity about what happens when a user installs the PowerShell module and WinGet isn't present in release notes here at GitHub. I was wondering if the other cmdlets could "check" to see if WinGet is installed and output some kind of error message with instructions to run the repair cmdlet.

denelon:
@JuryA,

It's been quite a while since this PR was created.

We've made substantial progress with the `Repair-WinGetPackageManager` cmdlet inside the "Microsoft.WinGet.Client" PowerShell module. This still installs via GitHub though.

I need to get a couple of internal folks to review this to ensure the URL is publicly and officially documented before I could include instructions as "official" for performing an install leveraging the Microsoft Store. As @Trenly pointed out, we would likely want to move this to a separate markdown file.

Issue #658, opened 2020-12-02
`winget download`


Rakib:
# Description of the new feature/enhancement

I want to install notepad++ but maybe not trough winget but just want to download the package, and also choose the package format, msi or msix.

This to have one place to download my apps. 

# Proposed technical implementation details (optional)

denelon:
Are you asking for some kind of "download only" function?

The manifests are essentially pointing to the source for the installers. We wouldn't have the ability to provide installer types that aren't available.

Rakib:
Both basically, since MS is pushing MSIX in the enterprise. It would be nice to be able to download .msix files from winget since most apps are not available in that format from their respective homepage.

thbch:
An enhancement of having a "download" only functionality would make sense for us as well. This would allow for a separation of downloading and installation.

JoergWu:
Would it not even be better simply to integrate the download / caching in "Delivery Optimization" and/or "BranchCache"?

denelon:
command:
`winget download <package>`

The typical options should be included here relative to search and install when specifying the package.

options:
`--path <path>`
`--installer-architecture <architecture>`
`--installer-locale <locale>`
`--installer-platform <platform>`
`--installer-scope <scope>`
`--installer-type <type>`

If these options are not specified, the installer that would have been selected by the default Windows Package Manager settings for `winget install <package>` would apply.

Typical output:

`Downloading <package name> version <version> from <source> source via <installer URL>.`

It's possible the user settings (requirements) or specified options cannot be met when attempting to locate an installer.
`ERROR: Unable to locate <package> installer with specified options or required settings.`

Some of the packages (non-Win32) from the Microsoft Store are prohibited from direct download due to licensing or agreements.
`ERROR: Unable to download Microsoft Store package.`

MarcoPeraza:
Great idea. Would it be possible to then point winget at the downloaded files when you do want to install? I guess that would be similar to a cache feature.

denelon:
[Discussion](https://github.com/microsoft/winget-cli/discussions/2911#discussioncomment-4857962) on cache management and reference to how this may be incorporated with a "winget download" capability.

JoergWu:
[...]

PR #2953, opened 2023-02-11
Create #658 - WinGet Download.md

RDMacLachlan:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2953)


doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+* `-e, --exact`
+* `-o, --output <path>`
+* `-t, --type <installer type>`
+* `--ignore-security-hash`
+* `--accept-package-agreements`
+* `--help`
+* `--wait`
+* `--verbose, --verbose-logs`
+
+An update to the existing schema is required for the inclusion of the `<boolean> isDownloadable` property. If this property is set to `True` or is not specified, then the package's installer will be considered downloadable. If the property is set to `False`, then the request to download the package's installer will be refused.
```
Trenly, line 42: Why is the `isDownloadable` needed? Users need to be able to download packages in the install process, so to specify that they can download it and run it, but can’t download it and not run it doesn’t make a lot of sense to me, especially since users could just run the install flow and terminate the thread between download and install to bypass this

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+    --installer-type                Specifies the filetype for the installer
+    --ignore-security-hash          Ignore the installer hash check failure
+    --accept-package-agreements     Accept all license agreements for packages
+    --help                          Shows help about the selected command
+    --wait                          |Prompts the user to press any key before existing
+    --verbose, --verbose-logs       Enables verbose logging for WinGet
+
+```
+
+WinGet search will be used to enlighten customers of the package's installers downloadable status. the new column will only be displayed if the application has an installer can be downloaded.
```
Trenly, line 102: Again, I don’t understand the downloadable status requirement

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+Microsoft Visual Studio Code          Microsoft.VisualStudioCode          1.75.0  winget True
+Microsoft Visual Studio Code Insiders Microsoft.VisualStudioCode.Insiders 1.76.0  winget
+```
+
+### WinGet Setting - Default Download Output
+
+A new setting will be included as the default output location for the user when package's installer is downloaded.
+
+```PS
+PS C:\> winget settings export
```
Trenly, line 117: Default should be a user setting, not an admin setting, and so wouldn’t be included in export

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+```PS
+"network: {
+        "userDefaultAppOutput":"C:\\Users\\Roy\\Downloads\\"
+        "useMeteredNetwork":"False"
+}"
+```
+
+The "userDefaultAppOutput" setting will be used as the default folder used for outputting the package's installer to.
+
+The "useMeteredNetwork" setting will accept boolean inputs, controlling the use of metered networks when downloading applications to the device.
```
Trenly, line 132: I don’t understand the requirement around metered networks here. Its not like windows update where files are usually so large as to cause issues, and downloads wouldn’t be background downloads, users have to explicitly request the download to begin

Trenly (commented):
I would think that the ability to download only would make more sense in the context of cache management. [This discussion](https://github.com/microsoft/winget-cli/discussions/2911) around cache management even includes the Download only request. I think we should assess if cache management makes more sense here to decrease code complexity

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+```PS
+PS C:\> winget settings export
+{"$schema":"https://aka.ms/winget-settings-export.schema.json","adminSettings":{"BypassCertificatePinningForMicrosoftStore":false,"InstallerHashOverride":false,"LocalArchiveMalwareScanOverride":false,"LocalManifestFiles":false},"userSettingsFile":"C:\\Users\\Roy\\AppData\\Local\\Packages\\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\\LocalState\\settings.json","userDefaultAppOutput:"C:\\Users\\Roy\\Downloads\\"}
+```
+
+The following items will be included in the WinGet Settings Export Schema 
+
+```PS
+"network: {
+        "userDefaultAppOutput":"C:\\Users\\Roy\\Downloads\\"
```
yao-msft, line 125: I think it should be in its own separate settings, not under "network" settings

useMeteredNetwork: DoDownloader honors the metered network settings, we probably don't want a separate settings for winget only, or at least it will be p2.

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+* `--moniker <moniker>`
+* `--scope <scope>`
+* `--locale <locale>`
+* `--fileName <name>`
+* `-v, --version <version>`
+* `-s, --source <Source>`
+* `-a, --architecture <architecture>`
+* `-e, --exact`
+* `-o, --output <path>`
+* `-t, --type <installer type>`
```
yao-msft, line 35:
```suggestion
* `-t, --installer-type <installer type>`
```

We probably also need nested installer type to differentiate between zips containing different contained installer types

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Solution Design
+
+The Windows Package Manager will implement a new command `WinGet Download` that will offer customers with the ability to download a single package's installers. The new command will make use of the following optional parameter values:
+* `--id <Identifier>`
+* `--name <name>`
+* `--moniker <moniker>`
+* `--scope <scope>`
+* `--locale <locale>`
+* `--fileName <name>`
```
yao-msft, line 29: Some description on how we'll use this would be great (i.e. we'll append file extension)

If file name is not provided, we'll use PackageIdentifier

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+issue id: 658
+---
+
+# Spec Title
+
+"For [#658](https://github.com/microsoft/winget-cli/issues/658)"
+
+## Abstract
+
+This spec describes the functionality and high-level implementation design for the downloading of applications using the Windows Package Manager.
```
yao-msft, line 14: Maybe a few words about "we are targeting downloading 1 installer at a time", downloading multiple may need some more thoughts in future

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+* `-t, --type <installer type>`
+* `--ignore-security-hash`
+* `--accept-package-agreements`
+* `--help`
+* `--wait`
+* `--verbose, --verbose-logs`
+
+An update to the existing schema is required for the inclusion of the `<boolean> isDownloadable` property. If this property is set to `True` or is not specified, then the package's installer will be considered downloadable. If the property is set to `False`, then the request to download the package's installer will be refused.
+
+Downloading the package's installer will still require that the package's installer hash be verified before becoming available to the user to interact with. The installer will be downloaded to the device with a `*.tmp` file extension that will be removed after the installer's hash has been validated. If the hash validation fails, the package's installer will be removed from the device.
```
yao-msft, line 44: nit: we already have temp file mechanisms, it's not "*.tmp", it's a guid at the moment

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+
+When downloading the package's installer, if a file with the same name exists the new download will overwrite the existing file.
```
yao-msft, line 46: I think it'll be better if file exists, we'll exit, override with --force

yao-msft:
> I would think that the ability to download only would make more sense in the context of cache management. [This discussion](https://github.com/microsoft/winget-cli/discussions/2911) around cache management even includes the Download only request. I think we should assess if cache management makes more sense here to decrease code complexity

We had some internal discussions about this. Since we already moved installer download to temp, there's not a lot more benefit of doing cache management. For customers wanting to do offline installation, we think explicit winget download is the way to go. We had customer asks that they need to manage a group of devices with limited network access, so winget download would be an option for this.

doc/specs/#658 - WinGet Download.md
yao-msft: > Why is the `isDownloadable` needed? Users need to be able to download packages in the install process, so to specify that they can download it and run it, but can’t download it and not run it doesn’t make a lot of sense to me, especially since users could just run the install flow and terminate the thread between download and install to bypass this

This is originally from msstore type where Store needs to limit what apps could be downloadable, apps marked as downloadable needs app developer approval.

Trenly: Thanks for the explanation!

doc/specs/#658 - WinGet Download.md
Trenly: Maybe use the `--rename` arg that already exists for portable packages?

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+* `--name <name>`
+* `--moniker <moniker>`
+* `--scope <scope>`
+* `--locale <locale>`
+* `--fileName <name>`
+* `-v, --version <version>`
+* `-s, --source <Source>`
+* `-a, --architecture <architecture>`
+* `-e, --exact`
+* `-o, --output <path>`
```
florelis, line 34: Seems redundant to have both `--output` and `--fileName`

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+* `-t, --type <installer type>`
+* `--ignore-security-hash`
+* `--accept-package-agreements`
```
florelis, line 37: Do we need to accept package agreements just to download? I would have thought these were for installing.

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+
+The following options are available
+    --id                            Filter results by id 
+    --name                          Filter results by name
+    --moniker                       Filter results by moniker
+    --scope                         Select install scope (user or machine)
+    --locale                        Locale to use (BCP47 format)
+    --filename                      The value to rename the downloaded executable file to
+    -v, --version                   Use the specified version; default is the latest version
+    -s, --source                    Find package using the specified source
```
florelis, line 89: If we have `--source` we will also need `--accept-source-agreements` and the custom header for REST.

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,209 @@
...
+"For [#658](https://github.com/microsoft/winget-cli/issues/658)"
+
+## Abstract
+
+This spec describes the functionality and high-level implementation design for the downloading of applications using the Windows Package Manager.
+
+## Inspiration
+
+This is inspired by customer feedback, and a need for broader application deployments:
+* Customers want to share the installer with an offline device.
```
florelis, line 19: I'd think for this to be worth it, we'd also have to have a way to tell the install command to use already downloaded installers. Probably something like `--local-installer-location` that defaults to the `userDefaultAppOutput` directory.

doc/specs/#658 - WinGet Download.md
RDMacLachlan: Thank you @Trenly you raise a valid point on this. If the customer is intentionally requesting an application download we should respect their request. I will remove this from the Spec :-).

doc/specs/#658 - WinGet Download.md
RDMacLachlan: Will add that to the Spec.

doc/specs/#658 - WinGet Download.md
RDMacLachlan: I am ok with that. Will update the spec.

yao-msft:
By the way, I just thought of something, do we deal with dependency packages for download? If yes, what will be the behavior and layout (folder structure, filename, etc)

doc/specs/#658 - WinGet Download.md
RDMacLachlan: In enterprise scenarios, if an IT Admin downloads or installs an application they accept the EULA on-behalf of their customer. If an application has a EULA, they should accept this EULA when retrieving the applications. At time of Install the customer may not see it as the install could occur outside of the Windows Package Manager.

doc/specs/#658 - WinGet Download.md
RDMacLachlan: Agree. Will add this to the spec.

doc/specs/#658 - WinGet Download.md
RDMacLachlan: I'm not sure I understand? Why wouldn't they use the local manifest for local installations?

RDMacLachlan:
> By the way, I just thought of something, do we deal with dependency packages for download? If yes, what will be the behavior and layout (folder structure, filename, etc)

We will need to handle dependency packages.

RDMacLachlan:
Need to include an option for only downloading the apps and not the license file.

Does having a similar flow for License file only make sense?

yao-msft:
> Does having a similar flow for License file only make sense?

No, licensing requires a specific package (content id), licensing call must come after download call.

RDMacLachlan:
> 
> > Does having a similar flow for License file only make sense?
> 
> No, licensing requires a specific package (content id), licensing call must come after download call.

I was thinking about the scenario of when the app downloads but the license file fails to download. Should the administrator then re-download the app again to get the license file.

RDMacLachlan:
Thinking about the use of --download-directory, would it make more sense for us to use "--Location", or "--Path"?

Trenly:
> Another thought, what if we include a new parameter "--continueOnError" that when applied would continue the download of the main and dependencies even after a single dependency fails.

I'd propose a more generic `-ErrorAction` like PowerShell's implementation where WinGet would respect `stop`, `continue`, `silentlyContinue`, `inquire`, `ignore` and maybe even `suspend`. This would probably make the PowerShell cmdlet implementation and could also be useful in other areas too. It could replace the IgnoreUnavailable argument for the import flow, could be used to set the behavior of upgrade all, could be used to continue when depenency installation fails, and would be a great option as a generic parameter for scripting

RDMacLachlan:
> > Another thought, what if we include a new parameter "--continueOnError" that when applied would continue the download of the main and dependencies even after a single dependency fails.
> 
> I'd propose a more generic `-ErrorAction` like PowerShell's implementation where WinGet would respect `stop`, `continue`, `silentlyContinue`, `inquire`, `ignore` and maybe even `suspend`. This would probably make the PowerShell cmdlet implementation and could also be useful in other areas too. It could replace the IgnoreUnavailable argument for the import flow, could be used to set the behavior of upgrade all, could be used to continue when depenency installation fails, and would be a great option as a generic parameter for scripting

I like this idea of having a parameter that can provide a deterministic experience for the user, that allows the user to determine what they want the experience to be.

Trenly:
> > I like this idea of having a parameter that can provide a deterministic experience for the user, that allows the user to determine what they want the experience to be.
> 
> If we are to replace existing parameters with new ErrorAction parameter, it will be a breaking change. And needs to go through all existing workflows to ensure the experience is accurate. As well as all Com and Powershell work. I would suggest making this a different work item not attached to this download work. For download, I think using --force or a specific flag is good for the purpose.
> 
> We also need to think about each different error case by case, some errors are hard stop (installer corrupt, etc), some errors (like dependency download failure for winget download) may let the workflow continue.
> 
> By the way, the reason I would like dependency download to fail the workflow is for consistency. Looks like in winget install flow, a dependency download failure will cause main package to fail to install. In winget download flow, a dependency download failure may be ok if user desires.

I'll make a separate issue for the proposed ErrorAction

yao-msft:
> I guess the download command got implemented, and the spec never made to the repo.

Yep, we just noticed yesterday and Ryan is working on updating it to match the final implementation. Thanks.

yao-msft (dismissed):

doc/specs/#658 - WinGet Download.md
```diff
@@ -0,0 +1,156 @@
...
+  --locale                     Locale to use (BCP47 format)
+  --ignore-security-hash       Ignore the installer hash check failure
+  --skip-dependencies          Skips processing package dependencies and Windows features
+  --accept-package-agreements  Accept all license agreements for packages
+  --accept-source-agreements   Accept all source agreements during source operations
+  -?,--help                    Shows help about the selected command
+  --wait                       Prompts the user to press any key before exiting
+  --logs,--open-logs           Open the default logs location
+  --verbose,--verbose-logs     Enables verbose logging for winget
+  --disable-interactivity      Disable interactive prompts
```
yao-msft, line 46: Does this list match the latest? I added custom-header for rest source recently.

ryfu-msft: Added missing arguments.

yao-msft (approved):

PR #2961, opened 2023-02-14
Update settings.schema.0.2.json to include pinning feature

florelis:
Add the "pinning" experimental feature to the settings file JSON schema to make it discoverable.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2961)

### DIFF
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -208,6 +208,11 @@
           "description": "Enable argument to override upgrade behavior to UninstallPrevious",
           "type": "boolean",
           "default": false
+        },
+        "pinning": {
+          "description": "Enable support for package pinning",
+          "type": "boolean",
+          "default": false
         }
       }
     }


ryfu-msft (approved):

Trenly (approved):

denelon (approved):

Trenly:
Unrelated to this PR - Is there a reason that the settings schema is still at 0.2? I'm just surprised that it doesn't have supported versions of the schema per client like the manifests do

denelon:
We need to add that to the checklist so we can support versioned schemas. Great call out @Trenly!

yao-msft:
> Unrelated to this PR - Is there a reason that the settings schema is still at 0.2? I'm just surprised that it doesn't have supported versions of the schema per client like the manifests do

Currently the schema is more of informational only and not enforced by client code. Client will read settings as much as it understands. That's probably why we did not bother updating the schema version in the past.

PR #2970, opened 2023-02-15
Add alternate url support for some predefined sources

JohnMcPMS:
Sometimes bad things happen; even when you thought you had everything set up properly it might turn out that you didn't.  The best laid plans, etc.  If there is a way to continue functioning, even if it is less efficient, shouldn't we use that?

## Change
Adds an alternate base URL for the sources behind `cdn.winget.microsoft.com`.  In the event that the files hosted there are inaccessible or corrupted, the alternate URL will be used instead.  This mechanism does rely on the fact that these are basically two names for the exact same set of data, which is true as the alternate is the upstream store for the CDN itself.

Additionally, we now require preindexed packaged sources to contain the manifest hash, rather than just using it if present.

Finally, this change adds a setting that can be used to disable the alternate URL for the `winget` source.  I intentionally did not add it to the schema or documentation, as I don't think it is something that should actually be set by anyone unless there is an explicit reason to do so (such as for testing purposes).

## Validation
Manually validated by changing the primary URL to a non-existent one.  The error was present in the logs, but the user experience was not impacted.  Tested the setting in this case to ensure that it would cause a failure.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2970)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/2970)

### DIFF
--- a/src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
@@ -21,19 +21,82 @@ namespace AppInstaller::Repository::Microsoft
         // TODO: This being hard coded to force using the Public directory name is not ideal.
         static constexpr std::string_view s_PreIndexedPackageSourceFactory_IndexFilePath = "Public\\index.db"sv;
 
-        // Construct the package location from the given details.
-        // Currently expects that the arg is an https uri pointing to the root of the data.
-        std::string GetPackageLocation(const SourceDetails& details)
+        struct PreIndexedPackageInfo
         {
-            THROW_HR_IF(E_INVALIDARG, details.Arg.empty());
-            std::string result = details.Arg;
-            if (result.back() != '/')
+            template <typename LocationCheck>
+            PreIndexedPackageInfo(const SourceDetails& details, LocationCheck&& locationCheck)
             {
-                result += '/';
+                // Get both locations to force the alternate location check
+                m_packageLocation = GetPrimaryPackageLocation(details);
+                locationCheck(m_packageLocation);
+
+                std::string alternateLocation = GetAlternatePackageLocation(details);
+                if (!alternateLocation.empty())
+                {
+                    locationCheck(alternateLocation);
+                }
+
+                // Try getting the primary location's info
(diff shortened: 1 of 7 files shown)


src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -325,6 +331,7 @@ namespace AppInstaller::Repository
             details.Name = s_Source_DesktopFrameworks_Name;
             details.Type = Microsoft::PreIndexedPackageSourceFactory::Type();
             details.Arg = s_Source_DesktopFrameworks_Arg;
+            details.AlternateArg = s_Source_DesktopFrameworks_AlternateArg;
```
yao-msft, line 334: Is it intentional to not check settings for this one?

yao-msft (approved):

src/AppInstallerRepositoryCore/SourceList.cpp
JohnMcPMS: Yes, I was intentionally not affecting the platform one because it is used in an automated fashion.  If we had a GP, then I would have it affected then.

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -21,19 +21,82 @@ namespace AppInstaller::Repository::Microsoft
...
+                CATCH_LOG_MSG("PreIndexedPackageInfo failed on alternate location");
+
+                THROW_HR(primaryHR);
             }
-            result += s_PreIndexedPackageSourceFactory_PackageFileName;
-            return result;
-        }
+
+            const std::string& PackageLocation() const { return m_packageLocation; }
+            Msix::MsixInfo& MsixInfo() { return *m_msixInfo; }
```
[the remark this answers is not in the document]
JohnMcPMS: The functions that it is passed to take `MsixInfo&`, so I am not returning it as const.  Likely `MsixInfo` just isn't const correct. 
 Probably my fault.  But not trying to correct that with this change.

Issue #2949, opened 2023-02-10
[Package Issue]: Gyan.FFmpeg does not uninstall correctly


mstevens83:
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

Other

### Brief description of your issue

Package Gyan.FFmpeg (v5.1.2) does not properly uninstall:
![image](https://user-images.githubusercontent.com/2787305/218076903-e1054d9e-6444-459d-870b-622a346a765f.png)

I'm `using --force` because otherwise it refuses to uninstall due to the package being supposedly modified, which is not the case. I even get this straight away after installing it fresh.

Notice I'm running this as Administrator because that also how I install it (due to microsoft/winget-pkgs#95349 / https://github.com/microsoft/winget-cli/issues/2909).

### Steps to reproduce

(As administrator:)
`winget install gyan.ffmpeg`
`winget uninstall gyan.ffmpeg` (complains about the package being modified, which is not the case)
`winget uninstall gyan.ffmpeg --force`

### Actual behavior

`%LOCALAPPDATA%\Microsoft\WinGet\Packages\Gyan.FFmpeg_Microsoft.Winget.Source_8wekyb3d8bbwe` and subfolder/files remain behind.

### Expected behavior

All files/folders related to package are gone.

### Environment

```shell
Windows Package Manager v1.4.10173
Windows: Windows.Desktop v10.0.22621.1105
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.19.10173.0


Gyan.FFmpeg v5.1.2
```

mdanish-kh:
This has been fixed in WinGet [v1.5.1081-preview release](https://github.com/microsoft/winget-cli/releases/tag/v1.5.1081-preview) through PR https://github.com/microsoft/winget-cli/pull/3002

cc @denelon

alexyo02:
Hi, still have the same problem on the latest version of Winget (1.6.2291-preview), on Windows 11 22h2. No matter what command I use, the process is stuck

mdanish-kh:
@alexyo02 That is probably related to https://github.com/microsoft/winget-cli/issues/3279. There are a couple workarounds to fix the uninstall mentioned in that issue

Issue #2909, opened 2023-02-02
Root directory of portable packages is always added to the PATH regardless of `RelativeFilePath`


sitiom:
### Brief description of your issue

In a machine without developer mode and in a non-admin shell, Portable packages with subdirectories under `RelativeFilePath` are not properly added to PATH. This is because the root directory of the package is **always** added to the PATH regardless.

### Steps to reproduce

- Ensure developer mode is turned off and not running under an admin shell (symlinking would happen otherwise)
- Install a package whose `RelativeFilePath` contains subdirectories. Let us install ffmpeg for this example: `winget install ffmpeg`
- The root of the package is added to the path (`%localappdata%\Microsoft\WinGet\Packages\Gyan.FFmpeg_Microsoft.Winget.Source_8wekyb3d8bbwe\`) instead of the actual file path (`%localappdata%\Microsoft\WinGet\Packages\Gyan.FFmpeg_Microsoft.Winget.Source_8wekyb3d8bbwe\ffmpeg-5.1.2-full_build\bin`)

### Expected behavior

The root directory should not be added to the path but directories relative to the `RelativeFilePath` (`%localappdata%\Microsoft\WinGet\Packages\Gyan.FFmpeg_Microsoft.Winget.Source_8wekyb3d8bbwe\ffmpeg-5.1.2-full_build\bin`).

### Actual behavior

The root directory of the package is **always** added to the PATH regardless of the `RelativeFilePath` (`%localappdata%\Microsoft\WinGet\Packages\Gyan.FFmpeg_Microsoft.Winget.Source_8wekyb3d8bbwe\`) .

### Environment

```shell
Windows Package Manager v1.4.10173
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22621.1194
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.19.10173.0
```
---
Another question, how are command aliases handled in non-developer mode? What if there are multiple executables in one directory, and you only want to add one of them to the path?

DanielHabenicht:
I can confirm this. So this basically effects all users who are not developing winget? (Best real life issue for "Works on ~my~our machines" 😆)

On a side note: 
In issue https://github.com/microsoft/winget-cli/issues/2711 they discuss about how it is currently impossible to execute executables linked to binary (`.dll`s) one way to solve it is by adding the installed folder directly to the `PATH`. 

[...]

PR #3002, opened 2023-02-22
Fix PATH behavior of non-symlink installations for Portables/Zip

ryfu-msft:
Fixes:
- #2909 
- #2949 

Description:
When failing to create a symlink (non-developer mode or non-admin), we should be adding the package install directory directly to PATH. However, the default root directory for portable packages was always being added instead. This has been causing issues when users attempt to execute the command alias, which fails to recognize the portable executable/symlink. This is especially problematic for portable(s) in zips as their `RelativeFilePath` was being ignored. 

Changes:
- If we fail to create a symlink, the InstallDirectoryAddedToPath key value is set to true and the portable target directory is added directly to PATH for all portables. 
- Modified the AddToPathVariable method to take in a value to add to PATH instead of predetermining the value to add to the PATH registry. This allows for multiple values to be appended to PATH to support multiple portables in an archive scenario. 
- Installing portable from zip always write to index. Previously, there was a check to see if there was greater than 1 item in the archive before writing to the index, but this is actually a bug because it could point to a single directory. The index is also a better way to keep track of all the files that were added.

Tests:
- Modified the SymlinkCreationFail unit tests to point to a temporary install location and verified that the temp install location is properly appended to the PATH registry during install and removed during uninstall.
- Verified that the changes could successfully install and uninstall `ffmpeg` which was the example identified in the issues above.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3002)

### DIFF
--- a/src/AppInstallerCLICore/PortableInstaller.cpp
+++ b/src/AppInstallerCLICore/PortableInstaller.cpp
@@ -138,9 +138,9 @@ namespace AppInstaller::CLI::Portable
                 }
                 else
                 {
-                    // Symlink creation should only fail if the user executes without admin rights or developer mode.
-                    // Resort to adding install directory to PATH directly.
-                    AICLI_LOG(Core, Info, << "Portable install executed in user mode. Adding package directory to PATH.");
+                    // If symlink creation fails, resort to adding the package directory to PATH.
+                    AICLI_LOG(Core, Info, << "Failed to create symlink at: " << filePath);
+                    AddToPathVariable(std::filesystem::path(entry.SymlinkTarget).parent_path());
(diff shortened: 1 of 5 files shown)


yao-msft (approved):
:shipit:

candrapersada:
how to restart your shell to use the new value?

yao-msft:
> how to restart your shell to use the new value?

Basically, to make the console (you are using) recognize newly added env variables, you'll have to restart it. For example, if you are running winget in Powershell, just close and start a new Powershell window. Only new instances of Powershell will recognize new portable added by winget. Same for CMD, WindowsTerminal, etc.

PR #3003, opened 2023-02-22
Use copy instead of rename for moving extracted archive directories

### LINKED ISSUE #2931: Cannot determine location to exract a portable software

I want to determine another folder to extract a portable software, but no matter how I use `--location``, it fails.

Screenshot 2023-02-07 010440

### Steps to reproduce

```powershell
winget install --id=TeamSophia.SophiApp --exact --accept-source-agreements --location "D:\test"
```

### Expected behavior

The ZIP archive should be extracted without errors.

### Actual behavior

I receive `An unexpected error occurred while executing the command:
copy_file: �������� � �������.: "C:\Users\SANCTU~1\AppData\Local\Temp\WinGet\TeamSophia.SophiApp.1.0.94\extracted\SophiApp", "D:\test\SophiApp"`

### Environment

```shell
Windows Package Manager v1.4.10173

Windows: Windows.Desktop v10.0.22621.1194
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.19.10173.0
```

ryfu-msft:
Fixes #2931 

Special thanks to @farag2 for the extremely thorough investigation on this one. 😄 

Issue:
When specifying a specific `--location` that points to a different drive, the installation fails with an access denied error. The issue is caused when trying to rename a directory across a different drive which isn't supported as `rename` is not a "move/copy" operation. The correct way to handle moving directories is to explicitly copy the directory recursively over to the intended install location. 

Changes:
Replaced `Rename` with `Copy` operation for directory file types during archiveinstallation. 

Tests:
Verified that the following command identified from the issue bug successfully installed. All existing tests should continue to pass. 
```winget install --id=TeamSophia.SophiApp --exact --accept-source-agreements --location "D:\test"```

### DIFF
--- a/src/AppInstallerCLICore/PortableInstaller.cpp
+++ b/src/AppInstallerCLICore/PortableInstaller.cpp
@@ -107,7 +107,7 @@ namespace AppInstaller::CLI::Portable
         }
         else if (fileType == PortableFileType::Directory)
         {
-            // Copy directory instead of renaming as there is a known issue with renaming across filesystems.
+            // Copy directory instead of renaming as there is a known issue with renaming across drives.
             AICLI_LOG(Core, Info, << "Copying directory to: " << filePath);
             std::filesystem::copy(entry.CurrentPath, filePath, std::filesystem::copy_options::overwrite_existing | std::filesystem::copy_options::recursive);
         }
--- a/src/AppInstallerCLICore/PortableInstaller.cpp
+++ b/src/AppInstallerCLICore/PortableInstaller.cpp
@@ -107,8 +107,9 @@ namespace AppInstaller::CLI::Portable
         }
         else if (fileType == PortableFileType::Directory)
         {
-            AICLI_LOG(Core, Info, << "Moving directory to: " << filePath);
-            Filesystem::RenameFile(entry.CurrentPath, filePath);
+            // Copy directory instead of renaming as there is a known issue with renaming across filesystems.
+            AICLI_LOG(Core, Info, << "Copying directory to: " << filePath);
+            std::filesystem::copy(entry.CurrentPath, filePath, std::filesystem::copy_options::overwrite_existing | std::filesystem::copy_options::recursive);
         }
         else if (entry.FileType == PortableFileType::Symlink)
         {

### CODE REVIEW

File: src/AppInstallerCLICore/PortableInstaller.cpp [hunk 1]
```diff
@@ -107,8 +107,9 @@ namespace AppInstaller::CLI::Portable
         }
         else if (fileType == PortableFileType::Directory)
         {
-            AICLI_LOG(Core, Info, << "Moving directory to: " << filePath);
-            Filesystem::RenameFile(entry.CurrentPath, filePath);
+            // Copy directory instead of renaming as there is a known issue with renaming across drives.
+            AICLI_LOG(Core, Info, << "Copying directory to: " << filePath);
+            std::filesystem::copy(entry.CurrentPath, filePath, std::filesystem::copy_options::overwrite_existing | std::filesystem::copy_options::recursive);
```
JohnMcPMS, line 112: I would really rather that we do something like:
```
if (IsSameVolume(current, target)
  Rename
else
  Copy
```
Renaming is just *so* much more efficient when it can be done that I think it is worth the extra effort.

ryfu-msft: Added Filesystem::IsSameVolume to check for this and only copies when false.

src/AppInstallerCommonCore/Public/winget/Filesystem.h
```diff
@@ -42,4 +42,7 @@ namespace AppInstaller::Filesystem
 
     // Gets the path of a known folder.
     std::filesystem::path GetKnownFolderPath(const KNOWNFOLDERID& id);
+
+    // Verifies that the target path is on the same volume as the current path.
+    bool IsSameVolume(const std::filesystem::path& current, const std::filesystem::path& target);
```
yao-msft, line 47: nit: maybe the concept of "current" and "target" does not apply to IsSameVolume, path1 and path2?

src/AppInstallerCommonCore/Filesystem.cpp
```diff
@@ -238,4 +238,18 @@ namespace AppInstaller::Filesystem
         THROW_IF_FAILED(SHGetKnownFolderPath(id, KF_FLAG_NO_ALIAS | KF_FLAG_DONT_VERIFY | KF_FLAG_NO_PACKAGE_REDIRECTION, NULL, &knownFolder));
         return knownFolder.get();
     }
+
+    bool IsSameVolume(const std::filesystem::path& current, const std::filesystem::path& target)
+    {
+        LPWSTR currentVolumeBuffer = new WCHAR[MAX_PATH];
```
yao-msft, line 244: new without delete is problematic

Use std::wstring as buffer like what we did in other places, and set buffer length to input length

src/AppInstallerCLICore/PortableInstaller.cpp
```diff
@@ -107,8 +107,16 @@ namespace AppInstaller::CLI::Portable
         }
         else if (fileType == PortableFileType::Directory)
         {
-            AICLI_LOG(Core, Info, << "Moving directory to: " << filePath);
-            Filesystem::RenameFile(entry.CurrentPath, filePath);
+            if (Filesystem::IsSameVolume(entry.CurrentPath, filePath))
+            {
```
yao-msft, line 111: nit: AICLI_LOG is missing in this path

ryfu-msft: adding AICLI_LOG to this if block

src/AppInstallerCommonCore/Public/winget/Filesystem.h
ryfu-msft: changed to 'path1' and 'path2'

src/AppInstallerCommonCore/Filesystem.cpp
```diff
@@ -238,4 +238,18 @@ namespace AppInstaller::Filesystem
         THROW_IF_FAILED(SHGetKnownFolderPath(id, KF_FLAG_NO_ALIAS | KF_FLAG_DONT_VERIFY | KF_FLAG_NO_PACKAGE_REDIRECTION, NULL, &knownFolder));
         return knownFolder.get();
     }
+
+    bool IsSameVolume(const std::filesystem::path& path1, const std::filesystem::path& path2)
+    {
+        std::wstring path1Volume;
+        std::wstring path2Volume;
+        DWORD path1Length = static_cast<DWORD>(path1.u16string().length());
```
yao-msft, line 246: >DWORD path1Length = static_cast<DWORD>(path1.u16string().length()); [](http://example.com/codeflow?start=8&length=67)

DWORD path1VolumeLength = static_cast<DWORD>(path1.wstring().length() + 1);

Same for below

yao-msft (approved):

Issue #556, opened 2020-09-04
Turn Windows features on and off


blogcraft:
Part of my setup is using Docker for Windows, obviously it works best with WSL2.

It would be nice to be able to turn things on in windows for my setup to be complete.

megamorf:
This can be done easily with PowerShell or the Deployment Image Servicing and Management (DISM) tool. I don't see how winget would help with that...

maciejpankanin:
It would be nice if  `winget install telnet` turned on telnet in Windows features.

megamorf:
Sorry but that's really just `Install-WindowsFeature Telnet-Client` in PowerShell with less ambiguity about what's going to happen. Or alternatively `dism /online /Enable-Feature /FeatureName:TelnetClient` with DISM.

jedieaston:
It would help for installing the RSAT, since that's been split up into 20-someodd packages in optional features. Chocolatey has a `choco install rsat` that installs all of them via a PowerShell script. 

Maybe PowerShell script support as an installer type would be the real feature request? Although that may encourage bad habits, since people could bypass the manifest creation step and put all of the logic inside of a non-standard script which then has to be debugged in a different way.

megamorf:
> If all you know is a hammer everything looks like a nail.

I'm a strong proponent of using the right tool for the job and the RSAT installation in PowerShell really is just a single line:

```powershell
# List RSAT components that are not installed
Get-WindowsCapability -Name RSAT* -Online | Where-Object { $_.State -eq "NotPresent" }

# Install all RSAT components
Get-WindowsCapability -Name RSAT* -Online | Add-WindowsCapability –Online
```

denelon:
This could be part of #163 "Support for package dependencies". If you `winget install ubuntu` we should be able to see if WSL is enabled or not, and take care of that for you.

doctordns:
If it were me, I;d keep the focus on WInget being a tool to lay down applications on developer workstations. 
If you need to configure a system after installation, you just use the existing PowerShell tool. THat has considerable advantages for not causing the winget guys to have to reinvent more wheels and would certainly end up with less confusing commands.

CatTheHacker:
What could be done is to allow packages to define post installation configuration (like `apt` has `debconf`) but I'm strongly against adding option such as enabling features via `winget`.

quangkieu:
I would suggest that **when** we have dependencies in `winget`, and the app manifest define that it depend on some windows features, winget would check if that exist and installed
[...]

PR #3005, opened 2023-02-23
Add experimental feature support for enabling Windows Feature dependencies

ryfu-msft:
Related to: #556 

Description:
This PR adds experimental feature support for enabling Windows Features prior to installation. There are known packages such as Wix toolkit that require certain optional Windows Features to be enabled before installation can run to completion. 

Changes:
- Added new `windowsFeature` experimental feature to the settings schema.
- Enabling windows feature is done by loading the DismApi.dll and calling those dll functions. 

Implementation flow:
- Checks if experimental feature is enabled
- Checks if admin mode is enabled in order to support calling DismApis
- Retrieves the WindowsFeatures dependencies and verifies that they all exist, if they do not, winget exits with a friendly error message. A user can add the `--ignore-missing-dependencies` flag to bypass this check and continue with the installation.
- Each valid Windows Feature dependency is enabled if not already enabled. If any of the dependencies fail to be enabled, an error message with the corresponding HRESULT is displayed to the user. The user can add `--force` to bypass this failure and continue with installation. Often times, a reboot will be required to fully enable a feature, in this scenario we warn the user but continue with the installation as it is possible that the installation may still succeed.

Tests:
- Added unit tests for verifying the behavior of the CLI output based on the exit codes from enabling/disabling Windows Features using TestHooks
- Added tests enabling `netfx3` and disabling `netfx3` and verifying the output is successful.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3005)


src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,109 @@ namespace AppInstaller::CLI::Workflow
...
+    void EnableWindowsFeaturesDependencies(Execution::Context& context)
+    {
+        if (!Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::WindowsFeature))
+        {
+            return;
+        }
+
+        const auto& rootDependencies = context.Get<Execution::Data::Installer>()->Dependencies;
+
+        if (rootDependencies.Empty())
```
Trenly, line 147: Is this already handled by `HasAnyOf` ? If not, should it be?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1677,4 +1677,36 @@ Please specify one of them using the --source option to proceed.</value>
...
+    <comment>"Windows Feature(s)" is the name of the options Windows features setting.
+{Locked="--force"}
+{Locked="--ignore-missing-dependencies"}</comment>
+  </data>
+  <data name="IgnoreMissingDependenciesArgumentDescription" xml:space="preserve">
+    <value>Ignores missing Windows Feature dependencies</value>
+    <comment>"Windows Feature" is the name of the options Windows features setting.</comment>
+  </data>
+  <data name="EnablingWindowsFeatures" xml:space="preserve">
+    <value>Enabling Windows Feature dependencies...</value>
```
Trenly, line 1697: As a user, I would like to know specifically which features are being enabled
```
    <value>Enabling Windows Feature {0}.</value>
```

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,109 @@ namespace AppInstaller::CLI::Workflow
...
+        {
+            return;
+        }
+
+        if (rootDependencies.HasAnyOf(DependencyType::WindowsFeature))
+        {
+            context << Workflow::EnsureRunningAsAdmin;
```
Trenly, line 154: Installer dependencies are known before this workflow is ever executed; Specific dependency types are also checked for in the `ReportDependencies` workflow.

I'm wondering, now that multiple dependency types will be supported, if it makes more sense to set an execution flag for "Contains Package Dependencies" or "Contains Windows Features Dependencies" in the ReportDependencies flow.

With the flags, then would be possible to have a separate workflow around `CheckSupportForWindowsFeatures` that checks the flag ensures the experimental feature is enabled and user is running as admin, to avoid creating copies of the dependencies data in memory unnecessarily

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,109 @@ namespace AppInstaller::CLI::Workflow
...
+                {
+                    AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_INSTALL_MISSING_DEPENDENCY);
+                }
+            }
+
+            HRESULT hr = S_OK;
+            bool continueOnFailure = context.Args.Contains(Execution::Args::Type::Force);
+            bool rebootRequired = false;
+
+            context.Reporter.Info() << Resource::String::EnablingWindowsFeatures << std::endl;
```
Trenly, line 203: As a user, I would prefer this be shown per-feature as they are enabled; I don't need to be informed that features dependencies are being enabled in the case that all the features are already enabled

doc/Settings.md
```diff
@@ -263,3 +263,14 @@ You can enable the feature as shown below.
...
 ```
+
+### windowsFeature
+
+This feature enables the ability to enable Windows Feature dependencies during installation.
+You can enable the feature as shown below.
+
+```json
+   "experimentalFeatures": {
+       "windowsFeature": true
```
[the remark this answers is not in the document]
JohnMcPMS: Features are a code construct whose state is driven by settings.  Settings can theoretically be arbitrarily complex, but I don't think we have any that are not simple types currently.  The settings code could even handle both bool and object values for `dependencies` if we were inclined.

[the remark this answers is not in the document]
Trenly: > The settings code could even handle both bool and object values for dependencies if we were inclined.

So you're saying it would be possible to have it where
```json
"experimentalFeatures": {
  "dependencies": true
}
```
Would either enable all dependencies, or keep the current behavior of only enabling package dependencies

OR, users could specify individual dependencies features?
```json
"experimentalFeatures": {
  "dependencies": {
    "packageDependencies" : true
    "windowsFeatures": false
  }
}
```

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,109 @@ namespace AppInstaller::CLI::Workflow
...
+
+            std::vector<std::string> invalidFeatures;
+
+            rootDependencies.ApplyToType(DependencyType::WindowsFeature, [&invalidFeatures](Dependency dependency)
+                {
+                    const auto& name = dependency.Id();
+                    WindowsFeature::WindowsFeature windowsFeature{ name };
+                    if (!windowsFeature.DoesExist())
+                    {
+                        invalidFeatures.emplace_back(name);
```
Trenly, line 168: Instead of using memory to create a vector, why not have a string like: `<value>Windows Feature not found: {0}</value>` and list the features out as they are found to be invalid, and make invalidFeatures a bool; Then, later you can show the message `<value>One or more dependencies were not found; proceeding due to --ignore-missing-dependencies.</value>`  (or the equivalent for not proceeding) which could be used for any dependency type.

Then you wouldn't need the bool for whether or not to terminate, and avoid the iterator for invalidFeatures

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,109 @@ namespace AppInstaller::CLI::Workflow
...
+                            }
+                        }
+                    }
+            });
+
+            if (FAILED(hr))
+            {
+                if (continueOnFailure)
+                {
+                    context.Reporter.Warn() << Resource::String::FailedToEnableWindowsFeatureOverridden << std::endl;
```
Trenly, line 226: Other than the log, there isn't any indication of *which* feature(s) failed to enable; Users may want this information to attempt manually enabling the feature without having to dig into the log.

src/AppInstallerCLITests/TestData/InstallFlowTest_WindowsFeatures.yaml
```diff
@@ -0,0 +1,23 @@
...
+      InstallerType: exe
+      InstallerSha256: 65DB2F2AC2686C7F2FD69D4A4C6683B888DC55BFA20A0E32CA9F838B51689A3B
+      InstallerSwitches:
+        Custom: /custom /scope=machine
+        SilentWithProgress: /silentwithprogress
+        Silent: /silence
+        Update: /update
+      Dependencies:
+        WindowsFeatures:
+        - netfx3
```
Trenly, line 21: Does casing matter for windows features? If it does, it probably wouldn't be a bad idea to have a test that validates incorrect casing fails.

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,109 @@ namespace AppInstaller::CLI::Workflow
...
+                }
+                else
+                {
+                    context.Reporter.Error() << Resource::String::FailedToEnableWindowsFeatureOverrideRequired << std::endl;
```
Trenly, line 230: @denelon - I may be asking the wrong (scope creep) questions here, so forgive me, but would we want any sort of telemetry around these?
1) Which features are requested to be enabled
2) The success / fail rates of enabling features

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: There is a symbolic name and a friendly (localized) name.  I don't know how easy it is to get the friendly name from the symbolic name.  Which do you expect to see (or both)?

src/AppInstallerCLICore/ExecutionArgs.h
```diff
@@ -46,6 +46,7 @@ namespace AppInstaller::CLI::Execution
             AcceptPackageAgreements, // Accept all license agreements for packages
             Rename, // Renames the file of the executable. Only applies to the portable installerType
             NoUpgrade, // Install flow should not try to convert to upgrade flow upon finding existing installed version
+            IgnoreMissingDependencies, // Ignores missing Windows Feature dependencies.
```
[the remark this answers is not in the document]
JohnMcPMS: Also, should `--force` apply here as well?

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
JohnMcPMS: I think we should be showing each feature as its own item and not aggregating them as well.

doc/Settings.md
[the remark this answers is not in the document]
JohnMcPMS: Yes, it would be possible to write the code the enable this.  I'm not signing @ryfu-msft up for that though 😄

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -495,6 +495,7 @@ namespace AppInstaller::CLI::Workflow
             Workflow::GetDependenciesFromInstaller <<
             Workflow::ReportDependencies(Resource::String::InstallAndUpgradeCommandsReportDependencies) <<
             Workflow::ManagePackageDependencies(Resource::String::InstallAndUpgradeCommandsReportDependencies) <<
+            Workflow::EnableWindowsFeaturesDependencies <<
```
JohnMcPMS, line 498: This whole thing doesn't seem like the right way to handle dependencies.  At least, "download" is not consistent with "and also, handle getting all dependencies squared away".  So at least maybe change the name of this task?

src/AppInstallerCLICore/ExecutionArgs.h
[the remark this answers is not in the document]
JohnMcPMS: Since you are using `--force` for overriding a failed install, I think that equally applies to one that doesn't exist.  I would remove this option.

Along those lines, I wouldn't treat an unknown feature as a special failure checked up front, but rather as a special failure when it is being satisfied in the normal flow.

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
```diff
@@ -0,0 +1,159 @@
...
+    {
+        WindowsFeature(const std::string& name) : m_featureName{ name } {};
+
+        bool DoesExist();
+        bool IsEnabled();
+        HRESULT EnableFeature();
+        HRESULT DisableFeature();
+
+    private:
+        DismApiHelper m_dismApiHelper;
```
JohnMcPMS, line 156: There is a lot of repeated loading and probing for the functions using this pattern.  It would be much better to just have this created one time.

If you want dism to be able to unload (which seems reasonable), then I would design this differently as well.  I would then create a single `DismHelper` object in the workflow and have it create `WindowsFeature` objects for you.

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
```diff
@@ -0,0 +1,159 @@
...
+        using DismGetFeatureInfoPtr = HRESULT(WINAPI*)(UINT, PCWSTR, PCWSTR, DismPackageIdentifier, DismFeatureInfo**);
+        using DismEnableFeaturePtr = HRESULT(WINAPI*)(UINT, PCWSTR, PCWSTR, DismPackageIdentifier, BOOL, PCWSTR*, UINT, BOOL, HANDLE, DISM_PROGRESS_CALLBACK, PVOID);
+        using DismDisableFeaturePtr = HRESULT(WINAPI*)(UINT, PCWSTR, PCWSTR, BOOL, HANDLE, DISM_PROGRESS_CALLBACK, PVOID);
+        using DismDeletePtr = HRESULT(WINAPI*)(VOID*);
+
+        typedef UINT DismSession;
+
+        wil::unique_hmodule m_module;
+        DismSession m_session = DISM_SESSION_DEFAULT;
+        DismFeatureInfo* m_featureInfo = nullptr;
```
JohnMcPMS, line 130: This shouldn't be holding state that is feature specific.
You shouldn't be using a raw pointer if you can help it, use a smart pointer type instead.  Both `std::unique_ptr` and `wil::unique_any` can handle this.

src/AppInstallerCommonCore/WindowsFeature.cpp
```diff
@@ -0,0 +1,207 @@
...
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        if (s_IsWindowsFeatureEnabledResult_TestHook_Override)
+        {
+            return *s_IsWindowsFeatureEnabledResult_TestHook_Override;
+        }
+#endif
+        DismFeatureInfo* featureInfo = m_dismApiHelper.GetFeatureInfo(m_featureName);
+        DismPackageFeatureState featureState = featureInfo->FeatureState;
+        AICLI_LOG(Core, Verbose, << "Feature state of " << m_featureName << " is " << featureState);
+        return (featureState == DismStateInstalled || featureState == DismStateInstallPending);
```
JohnMcPMS, line 205: Do you have more information on why `DismStateInstallPending` should be included but `DismStatePartiallyInstalled` is not? My gut would be that only `DismStateInstalled` would be installed and everything else should require us to do something.

src/AppInstallerCommonCore/FileLogger.cpp
```diff
@@ -24,6 +25,14 @@ namespace AppInstaller::Logging
         m_name = GetNameForPath(filePath);
         m_filePath = filePath;
 
+        // Prevent inheritance to ensure log file handle is not opened by other processes.
+        FILE* filePtr;
+        if (!fopen_s(&filePtr, m_filePath.string().c_str(), "w") && filePtr != NULL)
+        {
+            SetHandleInformation((HANDLE)_get_osfhandle(_fileno(filePtr)), HANDLE_FLAG_INHERIT, 0);
+            std::ofstream outFile(filePtr);
```
JohnMcPMS, line 33: There are two functions that open the stream; a helper should be used to update both.

This code is creating `outfile` and then immediately throwing it away.  You would want:
```
m_stream = std::ofstream{ filePtr };
```
And then to not call `m_stream.open`.

src/AppInstallerCommonCore/FileLogger.cpp
```diff
@@ -24,6 +25,14 @@ namespace AppInstaller::Logging
         m_name = GetNameForPath(filePath);
         m_filePath = filePath;
 
+        // Prevent inheritance to ensure log file handle is not opened by other processes.
+        FILE* filePtr;
+        if (!fopen_s(&filePtr, m_filePath.string().c_str(), "w") && filePtr != NULL)
```
JohnMcPMS, line 30: This needs proper error handling.

src/AppInstallerCommonCore/FileLogger.cpp
```diff
@@ -134,4 +124,22 @@ namespace AppInstaller::Logging
                 catch (...) {}
             }).detach();
     }
+
+    void FileLogger::OpenFileLoggerStream() 
+    {
+        // Prevent inheritance to ensure log file handle is not opened by other processes.
+        FILE* filePtr;
+        if (!fopen_s(&filePtr, m_filePath.string().c_str(), "w") && filePtr != NULL)
```
JohnMcPMS, line 132:
```suggestion
        errno_t fopenError = _wfopen_s(&filePtr, m_filePath.wstring().c_str(), L"w");
```

Use wide strings and capture the return value from `_wfopen_s` rather than using the global `errno`.  If you want to guard against a null pointer even when the return value indicates success, do that inside the block.

src/AppInstallerCommonCore/FileLogger.cpp
```diff
@@ -134,4 +124,22 @@ namespace AppInstaller::Logging
...
             }).detach();
     }
+
+    void FileLogger::OpenFileLoggerStream() 
+    {
+        // Prevent inheritance to ensure log file handle is not opened by other processes.
+        FILE* filePtr;
+        if (!fopen_s(&filePtr, m_filePath.string().c_str(), "w") && filePtr != NULL)
+        {
+            if (!SetHandleInformation((HANDLE)_get_osfhandle(_fileno(filePtr)), HANDLE_FLAG_INHERIT, 0))
```
JohnMcPMS, line 134:
```suggestion
            THROW_IF_WIN32_BOOL_FALSE(SetHandleInformation(static_cast<HANDLE>(_get_osfhandle(_fileno(filePtr))), HANDLE_FLAG_INHERIT, 0));
```

src/AppInstallerCommonCore/FileLogger.cpp
```diff
@@ -24,25 +24,15 @@ namespace AppInstaller::Logging
...
+        OpenFileLoggerStream();
     }
 
     FileLogger::FileLogger(const std::string_view fileNamePrefix)
     {
         m_name = "file";
         m_filePath = Runtime::GetPathTo(Runtime::PathName::DefaultLogLocation);
         m_filePath /= fileNamePrefix.data() + ('-' + Utility::GetCurrentTimeForFilename() + s_fileLoggerDefaultFileExt.data());
-
-        m_stream.open(m_filePath, s_fileLoggerDefaultOpenMode);
```
JohnMcPMS, line 36: I can't comment on that line, but `s_fileLoggerDefaultOpenMode` is no longer used with these changes.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: Changed to showing friendly name as well as the symbolic name such as `[netfx3]`

src/AppInstallerCLITests/TestData/InstallFlowTest_WindowsFeatures.yaml
ryfu-msft: casing doesn't matter, added tests to verify that

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
ryfu-msft: Changed to report each feature that failed to enable

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,104 @@ namespace AppInstaller::CLI::Workflow
...
+                    AICLI_TERMINATE_CONTEXT(hr);
+                }
+            }
+            else if (rebootRequired)
```
Trenly, line 221: I think this neglects the case where one feature fails to enable, and another has a reboot required. I don't know if we care about that case or not, or if that's even a possibility with the way `hr` gets handled.

If `--force` isn't present, I don't think it matters, since both the reboot and the fail would cause the context to terminate, and the termination reason doesn't make much of a difference. If `--force` is present, the user would only be informed they are bypassing the failed install, and they wouldn't be informed they are also bypassing the reboot requirement.

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,104 @@ namespace AppInstaller::CLI::Workflow
...
+            rootDependencies.ApplyToType(DependencyType::WindowsFeature, [&hr, &dismHelper, &info, &warn, &force, &rebootRequired](Dependency dependency)
+                {
+                    if (SUCCEEDED(hr) || force)
+                    {
+                        auto featureName = dependency.Id();
+                        DismHelper::WindowsFeature windowsFeature = dismHelper.CreateWindowsFeature(featureName);
+
+                        if (windowsFeature.DoesExist())
+                        {
+                            if (!windowsFeature.IsEnabled())
```
Trenly, line 178: I think there's a missing check for a pending reboot if the feature is already enabled?

Thinking of the case where a user attempts an install and gets a failure that a reboot is required. If they attempt to run the install command again without rebooting, the Windows Feature will be shown as enabled(?) and therefore won't trigger the rebootRequired, effectively negating the need to use `--force` by running the command twice.

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
ryfu-msft: no longer create vector to aggregate missing dependencies, they will be reported at the end after each windows feature is processed

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Moved this out of "download" to the step right before `InstallPackageInstaller`.

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
ryfu-msft: I tried using a unique_ptr but I don't think it's possible to access the memory address of a unique_ptr which is what I need for the **FeatureInfo parameter in [DismFeatureInfo](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/dism/dismgetfeatureinfo-function?view=windows-11). I believe this is okay though because DismDelete should handle the cleanup of this pointer.

src/AppInstallerCommonCore/WindowsFeature.cpp
ryfu-msft: I changed the condition so that only `DismStateInstalled` is considered as enabled.

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
ryfu-msft: Changed so that each windows feature is represented as its own item.

Trenly: Sounds good; I'm presuming any telemetry needs can be added once support is no longer experimental

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/WindowsFeature.cpp
```diff
@@ -0,0 +1,288 @@
...
+        }
+#endif
+        return WindowsFeature(name, m_dismGetFeatureInfo, m_dismEnableFeature, m_dismDisableFeature, m_dismDelete, m_dismSession);
+    }
+
+    void DismHelper::Initialize()
+    {
+        if (m_dismInitialize)
+        {
+            LOG_IF_FAILED(m_dismInitialize(2, NULL, NULL));
```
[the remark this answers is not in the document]
ryfu-msft: Changed to using enum value `DismLogErrorsWarningsInfo`

src/AppInstallerCommonCore/WindowsFeature.cpp
```diff
@@ -0,0 +1,288 @@
...
+            AICLI_LOG(Core, Error, << "Could not load dismapi.dll");
+            return;
+        }
+
+        m_dismInitialize =
+            reinterpret_cast<DismInitializePtr>(GetProcAddress(m_module.get(), "DismInitialize"));
+        if (!m_dismInitialize)
+        {
+            AICLI_LOG(Core, Error, << "Could not get proc address of DismInitialize");
+            return;
```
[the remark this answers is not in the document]
ryfu-msft: Changed to using suggested helper function. I chose to use `THROW_LAST_ERROR` since that returns the error code returned from `GetProcAddress`

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
```diff
@@ -0,0 +1,198 @@
...
+                DismEnableFeaturePtr enableFeaturePtr,
+                DismDisableFeaturePtr disableFeaturePtr,
+                DismDeletePtr deletePtr,
+                DismSession session);
+
+            void GetFeatureInfo();
+
+            std::string m_featureName;
+            DismFeatureInfo* m_featureInfo = nullptr;
+            DismGetFeatureInfoPtr m_getFeatureInfoPtr = nullptr;
```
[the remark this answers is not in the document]
ryfu-msft: I moved WindowsFeature out of DismHelper and created a `shared_ptr<DismHelper>` member. WindowsFeature now calls the functions through the dismhelper shared pointer.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1782,4 +1782,33 @@ Please specify one of them using the --source option to proceed.</value>
...
+    <value>Failed to enable Windows Feature dependencies; proceeding due to --force</value>
+    <comment>"Windows Feature(s)" is the name of the options Windows features setting.
+{Locked="--force"}</comment>
+  </data>
+  <data name="EnablingWindowsFeature" xml:space="preserve">
+    <value>Enabling {0} [{1}]...</value>
+    <comment>{Locked="{0}","{1}"} Message displayed to the user regarding which Windows Feature is being enabled.</comment>
+  </data>
+  <data name="FailedToEnableWindowsFeature" xml:space="preserve">
+    <value>Failed to enable [{0}] feature.</value>
```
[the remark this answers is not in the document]
ryfu-msft: Added hresult error code to output

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1782,4 +1782,33 @@ Please specify one of them using the --source option to proceed.</value>
...
+  <data name="FailedToEnableWindowsFeatureOverrideRequired" xml:space="preserve">
+    <value>Failed to enable Windows Feature dependencies. To proceed with installation, use --force</value>
+    <comment>{Locked="--force"}</comment>
+  </data>
+  <data name="RebootRequiredToEnableWindowsFeatureOverrideRequired" xml:space="preserve">
+    <value>Reboot required to fully enable the Windows Feature(s); to override this check use --force</value>
+    <comment>"Windows Feature(s)" is the name of the options Windows features setting.</comment>
+  </data>
+  <data name="FailedToEnableWindowsFeatureOverridden" xml:space="preserve">
+    <value>Failed to enable Windows Feature dependencies; proceeding due to --force</value>
```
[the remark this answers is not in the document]
ryfu-msft: I left this as is because there are many other examples where `proceeding due to --arg` is used such as:
InstallerHashMismatchOverridden
PortableHashMismatchOverridden
ArchiveFailedMalwareScanOverridden

I think this is fine because it tells the user what arg they included to override the check. I'm also not sure what else to put so if you feel strongly about this, I can change this (and the others as well) to follow a different pattern.

src/AppInstallerCLICore/Workflows/DependenciesFlow.h
```diff
@@ -59,4 +59,6 @@ namespace AppInstaller::CLI::Workflow
     // Inputs: PackageVersion, Manifest
     // Outputs: DependencySource
     void OpenDependencySource(Execution::Context& context);
+
+    void EnableWindowsFeaturesDependencies(Execution::Context& context);
```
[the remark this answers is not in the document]
ryfu-msft: Added workflow comments

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -133,6 +135,104 @@ namespace AppInstaller::CLI::Workflow
...
+                        {
+                            // Note: If a feature is not found, continue enabling the rest of the dependencies but block immediately after unless force arg is present.
+                            AICLI_LOG(Core, Info, << "Windows Feature [" << featureName << "] does not exist");
+                            hr = APPINSTALLER_CLI_ERROR_INSTALL_MISSING_DEPENDENCY;
+                            warn << Resource::String::WindowsFeatureNotFound(Utility::LocIndView{ featureName }) << std::endl;
+                        }
+                    }
+            });
+
+            if (FAILED(hr) || hr == APPINSTALLER_CLI_ERROR_INSTALL_MISSING_DEPENDENCY)
```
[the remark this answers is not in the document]
ryfu-msft: Removed APPINSTALLER_CLI_ERROR_INSTALL_MISSING_DEPENDENCY  since you are right and it is covered by `FAILED(hr)`

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
```diff
@@ -0,0 +1,198 @@
...
+        /// <summary>
+        /// Struct representation of a single Windows Feature.
+        /// </summary>
+        struct WindowsFeature
+        {
+            HRESULT Enable();
+            HRESULT Disable();
+            bool DoesExist();
+            bool IsEnabled();
+            std::wstring GetDisplayName();
```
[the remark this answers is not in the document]
ryfu-msft: Changed to returning a LocIndString directly.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: Thanks for the explanation, I didn't know that COM allowed for multiple downloads. I moved the task back to the suggested location and added a TODO comment explaining the issue that needs to be addressed.

JohnMcPMS (changes requested):

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1804,8 +1804,9 @@ Please specify one of them using the --source option to proceed.</value>
     <comment>{Locked="{0}","{1}"} Message displayed to the user regarding which Windows Feature is being enabled.</comment>
   </data>
   <data name="FailedToEnableWindowsFeature" xml:space="preserve">
-    <value>Failed to enable [{0}] feature.</value>
-    <comment>{Locked="{0}"} Error message displayed when a failure was encountered when enabling a Windows Feature.</comment>
+    <value>Failed to enable [{0}] feature with exit code: {1}</value>
```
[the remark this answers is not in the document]
ryfu-msft: Changed to suggested, I used GetPresentableUserMessage to handle `{2}: {3}`

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
```diff
@@ -113,69 +115,17 @@ namespace AppInstaller::WindowsFeature
...
-        {
-            CloseSession();
-            Shutdown();
-        };
-
-        WindowsFeature CreateWindowsFeature(const std::string& name);
+        ~DismHelper();
 
     private:
+        friend WindowsFeature;
```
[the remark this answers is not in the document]
ryfu-msft: removed and exposed public functions

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
```diff
@@ -194,5 +144,46 @@ namespace AppInstaller::WindowsFeature
...
+    };
+
+    /// <summary>
+    /// Struct representation of a single Windows Feature.
+    /// </summary>
+    struct WindowsFeature
+    {
+        // This default constructor is only used for mocking unit tests.
+        WindowsFeature() = default;
+        WindowsFeature(std::shared_ptr<DismHelper> dismHelper, const std::string& name);
```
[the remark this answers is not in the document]
ryfu-msft: I left the constructor as is to be able to get a shared_ptr to DismHelper. I changed DismHelper to derive from `enable_shared_from_this` so that we can safely get another shared_ptr as suggested.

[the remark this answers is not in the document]
JohnMcPMS: The awkward part is the call site; the need to pass the `shared_ptr` yourself.  (I should have put the comment there I suppose...)  I wasn't suggesting that you should take the constructor away (although you could make it protected and friend the other way).  I *was* suggesting that you add a function like:
```
WindowsFeature DismHelper::GetFeature(std::string featureName);
```
So that the call site was less awkward, like:
```
WindowsFeature::WindowsFeature windowsFeature = dismHelper->GetFeature(featureName};
```
If you did make the constructor protected, you then don't have to worry about the `shared_ptr` being null.

JohnMcPMS (approved):

src/AppInstallerCommonCore/Public/winget/WindowsFeature.h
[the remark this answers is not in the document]
ryfu-msft: I agree the call site is weird passing shared ptrs around. I feel that this is important if dismhelper were to be extended to other scenarios so I changed it based on your suggestions. I created a helper function within dismhelper that creates the windows object for you, calling the constructor with the shared ptr. I made the WindowsFeature constructor protected and made dismHelper a friend so that only the dismhelper can create the WindowsFeature object.

PR #3006, opened 2023-02-23
Remove zip installer type from 1.0 and 1.1 manifests

ryfu-msft:
Continuation of #2996

Removes zip installer type from 1.0 and 1.1 manifests as zip is not supported until 1.4 client.
Also fixes the unit tests accordingly so that they do not rely on the zip installertype.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3006)

### DIFF
--- a/src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
+++ b/src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
@@ -4,24 +4,24 @@ Version: 1.7.32
 Publisher: Microsoft
 License: MIT License
 Description: A complex system reference test file.
-InstallerType: EXE
+InstallerType: exe
 PackageFamilyName: Microsoft.DesktopAppInstaller_8wekyb3d8bbwe
 ProductCode: "{Foo}"
 Installers:
-  - InstallerType: MSIX
+  - InstallerType: msix
     Arch: x86
     Url: https://rubengustorage.blob.core.windows.net/publiccontainer/msixsdkx86.zip
     Sha256: 69D84CA8899800A5575CE31798293CD4FEBAB1D734A07C2E51E56A28E0DF8C82
-  - InstallerType: MSI
+  - InstallerType: msi
     Arch: x86
     Url: https://rubengustorage.blob.core.windows.net/publiccontainer/msixsdkx86.zip
     Sha256: 69D84CA8899800A5575CE31798293CD4FEBAB1D734A07C2E51E56A28E0DF8C82
-  - InstallerType: MSIX
+  - InstallerType: msix
     Arch: x64
     Url: https://rubengustorage.blob.core.windows.net/publiccontainer/msixsdkx86.zip
     Sha256: 69D84CA8899800A5575CE31798293CD4FEBAB1D734A07C2E51E56A28E0DF8C82
     PackageFamilyName: Override_8wekyb3d8bbwe
-  - InstallerType: MSI
+  - InstallerType: msi
     Arch: x64
     Url: https://rubengustorage.blob.core.windows.net/publiccontainer/msixsdkx86.zip
     Sha256: 69D84CA8899800A5575CE31798293CD4FEBAB1D734A07C2E51E56A28E0DF8C82
--- a/src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
+++ b/src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
@@ -4,14 +4,10 @@ Version: 1.7.32
 Publisher: Microsoft
 License: MIT License
 Description: A complex system reference test file.
-InstallerType: Zip
+InstallerType: EXE
 PackageFamilyName: Microsoft.DesktopAppInstaller_8wekyb3d8bbwe
 ProductCode: "{Foo}"
 Installers:
-  - InstallerType: Zip
-    Arch: x86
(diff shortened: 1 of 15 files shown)


Trenly:
Holey smokes, I didn’t see they were implemented so many places! Thanks for catching my miss

ryfu-msft:
> Holey smokes, I didn’t see they were implemented so many places! Thanks for catching my miss

Nah, that's my mistake. I didn't realize the tests also were also affected since the pipelines don't get triggered on only schema file changes. 😅The pipeline run should be triggered on schema changes moving forward...

src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
```diff
@@ -4,14 +4,10 @@ Version: 1.7.32
 Publisher: Microsoft
 License: MIT License
 Description: A complex system reference test file.
-InstallerType: Zip
+InstallerType: EXE
```
florelis, line 7: Nit: You use inconsistent case (EXE vs exe) on these

src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
```diff
@@ -4,14 +4,10 @@ Version: 1.7.32
...
 PackageFamilyName: Microsoft.DesktopAppInstaller_8wekyb3d8bbwe
 ProductCode: "{Foo}"
 Installers:
-  - InstallerType: Zip
-    Arch: x86
-    Url: https://rubengustorage.blob.core.windows.net/publiccontainer/msixsdkx86.zip
-    Sha256: 69D84CA8899800A5575CE31798293CD4FEBAB1D734A07C2E51E56A28E0DF8C82
```
florelis, line 14: I'm assuming you made sure that we don't lose coverage by removing this?

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -93,10 +93,10 @@ TEST_CASE("ReadPreviewGoodManifestAndVerifyContents", "[ManifestValidation]")
     REQUIRE(installer1.Url == "https://rubengustorage.blob.core.windows.net/publiccontainer/msixsdkx86.zip");
     REQUIRE(installer1.Sha256 == SHA256::ConvertToBytes("69D84CA8899800A5575CE31798293CD4FEBAB1D734A07C2E51E56A28E0DF8C82"));
     REQUIRE(installer1.Locale == "en-US");
-    REQUIRE(installer1.BaseInstallerType == InstallerTypeEnum::Zip);
+    REQUIRE(installer1.BaseInstallerType == InstallerTypeEnum::Exe);
     REQUIRE(installer1.Scope == ScopeEnum::User);
     REQUIRE(installer1.PackageFamilyName == "");
-    REQUIRE(installer1.ProductCode == "");
+    REQUIRE(installer1.ProductCode == "{Foo}");
```
florelis, line 99: Why did this one change?

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -342,36 +342,31 @@ TEST_CASE("ComplexSystemReference", "[ManifestValidation]")
...
-    REQUIRE(manifest.Installers.size() == 5);
-
-    // Zip installer does not inherit
-    REQUIRE(manifest.Installers[0].BaseInstallerType == InstallerTypeEnum::Zip);
-    REQUIRE(manifest.Installers[0].PackageFamilyName == "");
-    REQUIRE(manifest.Installers[0].ProductCode == "");
+    REQUIRE(manifest.Installers.size() == 4);
 
     // MSIX installer does inherit
-    REQUIRE(manifest.Installers[1].BaseInstallerType == InstallerTypeEnum::Msix);
```
florelis, line 353: It's probably not worth changing, but we could do something like
```
auto& installer = manifest.Installer[0];
REQUIRE(installer.BaseInstallerType == InstallerTypeEnum::Msix);
...
```
so that we don't have to change the index in all places. Though I doubt we'll have to do it again...

florelis (approved):

src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
ryfu-msft: Yes, the v1.4 manifest tests cover this scenario and ensure that fields from the root get passed to the installer if they are not defined.

src/AppInstallerCLITests/TestData/Manifest-Good-SystemReferenceComplex.yaml
ryfu-msft: I changed all the installer types in this test manifest to lowercase to match other yamls.

src/AppInstallerCLITests/YamlManifest.cpp
ryfu-msft: I think for the older manifests, fields were not passed to the zip installer but that behavior has since changed. Since I changed the installertype to EXE, the productCode specified from the root is now being passed to the installer node.

src/AppInstallerCLITests/YamlManifest.cpp
ryfu-msft: Changed to suggested so that we never have to think about this again :)

Issue #3061, opened 2023-03-10
Separate "PackageVersion" from "Marketing Version"


Trenly:
### Description of the new feature / enhancement

Currently, there is the concept of two different versions
1) The Marketing Version
2) The DisplayVersion

Guidance has been to use `PackageVersion` as the marketing version and using the `AppsAndFeaturesEntries` to define the `DisplayVersion` when it differs from the marketing version.

I propose adding a new manifest field `MarketingVersion` - which would allow the version to be arbitrary and duplicative while still retaining a semantic version for `PackageVersion`

Trenly:
[Policy] Area-Manifest

PR #3062, opened 2023-03-10
Spec 3061 - Marketing Version

Trenly:
* Related to #3061

### DIFF
--- a/doc/specs/#3061 - Marketing Version.md
+++ b/doc/specs/#3061 - Marketing Version.md
@@ -1,0 +1,122 @@
+---
+author: Kaleb Luedtke (Trenly) trenlymc@gmail.com
+created on: 2023-03-10
+last updated: 2023-03-10
+issue id: 3061
+---
+
+# Marketing Version
+
+"For [#3061](https://github.com/microsoft/winget-cli/issues/3061)"
+
+## Abstract
+
+There are times when the version which is marketed to consumers on publishers' websites differs from what is written to the Apps and Features Entries data. Additionally, there are times when the same version is marketed for several different versions of the same package. The Windows Package Manager should be able to show an arbitrary version as determined by the publisher.
+
+## Inspiration
+
+The current design using `PackageVersion` for the displayed version and how it differs from `DisplayVersion` can be confusing.
+
+## Solution Design
+
+The Windows Package Manager manifest schema should have a `MarketingVersion` key added at the root level. For clarity, it should be placed immediately following the `PackageVersion` key.
+
+When the Windows Package Manager needs to display the version of a package, the value from the `MarketingVersion` key should be used. If the key is not present or has no value, the value of `PackageVersion` should be displayed instead, with the exception of `winget show <query> --versions` and `winget search <query> --versions` which should show the values as key pairs of `PackageVersion (MarketingVersion)`.
+
+When the `--version` parameter is used, matches with `PackageVersion` should be considered first and matches with `MarketingVersion` should be considered second. If there are multiple package versions of the same identifier matched using `MarketingVersion`, the highest `PackageVersion` shall be selected.
+
+When `MarketingVersion` is present, `PackageVersion` should be validated to be a semantic version.
+
+For the purposes of version comparison, mapping installed packages to versions, and all other behaviors, the current implementation using `PackageVersion` should be retained.
+
+## UI/UX Design
+
+Example of a manifest with a `MarketingVersion` specified:
+```yaml
+PackageIdentifier: Example.Package
+PackageVersion: 1.2.5.1
+MarketingVersion: 1.2.5
+Installers:
+  InstallerUrl: https://example.com
+  InstallerSha256: <SHA>
+  InstallerType: exe
+  AppsAndFeaturesEntries:
+    - DisplayVersion: 1.2.5-abcde
+...
+```
+
+### `winget list`
+```
+Name                Id               Version   Available
+---------------------------------------------------------- 
+Example Package     Example.Package  1.2.4     1.2.5
+```
+
+### `winget show Example.Package --versions`
+```
+Found Example Package [Example.Package]
+Version
+-------------
+1.2.5.1  (1.2.5)
+1.2.4.9  (1.2.4)
+1.2.4.8b (1.2.4 beta)
+1.2.4.7b (1.2.4 beta)
+```
+
+### `winget upgrade` when the installed version and newer version have the same marketing version
+```
+Name                Id               Version   Available    Source
+--------------------------------------------------------------------
+Example Package     Example.Package  1.2.5     * 1.2.5      winget
+```
+
+### `winget install Example.Package --version 1.2.4.7b`
+*This would install version 1.2.4.7b*
+```
+Found Badlion Client [Badlion.BadlionClient] Version 1.2.4 beta
+This application is licensed to you by its owner.
+...
+```
+
+### `winget install Example.Package --version '1.2.4 beta'`
+*This would install version 1.2.4.8b*
+```
+Found Badlion Client [Badlion.BadlionClient] Version 1.2.4 beta
+This application is licensed to you by its owner.
+...
+```
+
+## Capabilities
+Installed packages should still be mapped to `PackageVersion`, which will allow winget to update packages which change `DisplayVersion` but do not change `MarketingVersion`.
+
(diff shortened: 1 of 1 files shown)


denelon (approved):

denelon:
This looks good to me, but I want to get at least one other person from the engineering team to review.

doc/specs/#3061 - Marketing Version.md
mdanish-kh, line 1: I suppose Tab completion will work the same way as before by giving PackageVersion. If so, we should probably mention it in the spec.

Trenly:
@ryfu-msft - I believe @denelon is on PTO this week. Do you know who the right person from the team is to review this?

doc/specs/#3061 - Marketing Version.md
```diff
@@ -0,0 +1,124 @@
...
+
+For the purposes of version comparison, mapping installed packages to versions, and all other behaviors, the current implementation using `PackageVersion` should be retained.
+
+## UI/UX Design
+
+Example of a manifest with a `MarketingVersion` specified:
+```yaml
+PackageIdentifier: Example.Package
+PackageVersion: 1.2.5.1
+MarketingVersion: 1.2.5
```
yao-msft, line 38: Do you think MarketingVersion should exist in default locale manifest only Or it could be different for different locales?

doc/specs/#3061 - Marketing Version.md
```diff
@@ -0,0 +1,124 @@
...
+  AppsAndFeaturesEntries:
+    - DisplayVersion: 1.2.5-abcde
+...
+```
+
+### `winget list`
+```
+Name                Id               Version   Available
+---------------------------------------------------------- 
+Example Package     Example.Package  1.2.4     1.2.5
```
yao-msft, line 52: Just to confirm, the installed Version here is from arp entry display version and Available version is from marketing version? Would that cause confusion if Available version is way different from installed version?

yao-msft: Or were you thinking the installed Version would be mapped to marketing version as well?

yao-msft:
Just as a note, for this marketing version to take effect, this new marketing version needs to be taken into local index (we do not want to download every manifest to check marketing version for matching).

By the way, is this Marketing Version mainly for better display purpose? I do not see many functional changes around this new field. I guess this will also need to be exposed in Com api so at least our powershell cmdlet would display the same

doc/specs/#3061 - Marketing Version.md
Trenly: That's an interesting point. I would presume that it would be the same across locales, since there is usually a single version referenced on websites / documentation. I suppose it is possible that different locales could use different versions, but I would be a very small number of packages. Perhaps it would be better in the version manifest then, instead of the locale?

Trenly:
> Just as a note, for this marketing version to take effect, this new marketing version needs to be taken into local index (we do not want to download every manifest to check marketing version for matching).
> 
I want to be sure that I understand correctly; By local index is this referring to the index.db from the pre-packaged winget source, or the installed.db? Since it is theoretically possible that a publisher could change the marketing version at will, I was thinking it would need to be made available in the pre-indexed package and from REST source implementations. 

> By the way, is this Marketing Version mainly for better display purpose? I do not see many functional changes around this new field. I guess this will also need to be exposed in Com api so at least our powershell cmdlet would display the same

"Mainly" - yes. However, functionally, it has a large impact to user experience based on the way manifests are created. There is an issue open for handling versions with pre-ambles better - the marketing version would allow versions with pre-ambles to be handled without issue by using the DisplayVersion for correlation, MarketingVersion for display purposes, and PackageVersion for sorting. There is also an issue open for publishers that use dd-mm-yyyy date formats as versions, or where patches are denoted by adding "a" "b" "c" to the base version without a separator, or where there is a git commit hash, to allow them to be sorted correctly. This new marketing version, when used in combination with DisplayVersion and PackageVersion would effectively allow that sorting to happen at the manifest level instead of requiring unique handling in the cli, since the CLI would use DisplayVersion to map installed packages to PackageVersion, continues to perform version comparison using PackageVersion, but then displays the MarketingVersion.

doc/specs/#3061 - Marketing Version.md
yao-msft: This is assuming all installed versions are listed in winget repos, which is not the case most of the time. Currently installed to available mapping only happens when the manifest contains DisplayName in Apps and Features entries. If we want to achieve showing marketing version for installed, we need to map installed to available for all cases. And likely, we'll get `< SomeMarketVersion` as a result most of the time.

doc/specs/#3061 - Marketing Version.md
yao-msft: Version manifest is currently only used for validation and ignored during processing (merging in the service). If we want only one marketing version, we can limit it to just the default locale.

doc/specs/#3061 - Marketing Version.md
Trenly: Hmmm. That's true, but its also not different than what happens currently when there is a DisplayVersion specified and the installed version doesn’t match an available version

yao-msft:
> I want to be sure that I understand correctly; By local index is this referring to the index.db from the pre-packaged winget source, or the installed.db? Since it is theoretically possible that a publisher could change the marketing version at will, I was thinking it would need to be made available in the pre-indexed package and from REST source implementations.

I meant index.db. Marketing version will be added to installed.db though since both use the same code. But installed.db is only used for correlation heuristics, the marketing version in it will be ignored.

> "Mainly" - yes. However, functionally, it has a large impact to user experience based on the way manifests are created. There is an issue open for handling versions with pre-ambles better - the marketing version would allow versions with pre-ambles to be handled without issue by using the DisplayVersion for correlation, MarketingVersion for display purposes, and PackageVersion for sorting. There is also an issue open for publishers that use dd-mm-yyyy date formats as versions, or where patches are denoted by adding "a" "b" "c" to the base version without a separator, or where there is a git commit hash, to allow them to be sorted correctly. This new marketing version, when used in combination with DisplayVersion and PackageVersion would effectively allow that sorting to happen at the manifest level instead of requiring unique handling in the cli, since the CLI would use DisplayVersion to map installed packages to PackageVersion, continues to perform version comparison using PackageVersion, but then displays the MarketingVersion.

If package owners wrote these "commit hash versions" to arp entry, then it would not help improve sorting.
If package owners just wanted to show "commit hash versions" to user. Then I think this marketing version solves their needs. Then in this scenario, for displaying installed, should we really need to map to marketing version?

```
Name                Id               Version   Available
---------------------------------------------------------- 
Example Package     Example.Package  < 801d129d     43dba4c4
```

Trenly:
> If package owners wrote these "commit hash versions" to arp entry, then it would not help improve sorting.

But it would, since the `PackageVersion` could be set to 1.0.0, 1.0.1, 1.0.2, 2.0.1, etc; The arp matching would be handled by DisplayVersion; Then MarketingVersion would control the version displayed in outputs

doc/specs/#3061 - Marketing Version.md
yao-msft: Yep. We can only do relative version `< SomeVersion` if the mapping is performed. It's just the question that is showing `< SomeMarketVersion` is better or  `< SomePackageVersion` is better.

I was thinking to make the logic easier. We only do mapping when ARP DisplayVersion is specified in the manifest, most likely, packages that need marketing version are those packages needing mapping because they write non very standard versions in registry.

yao-msft:
> > If package owners wrote these "commit hash versions" to arp entry, then it would not help improve sorting.
> 
> But it would, since the `PackageVersion` could be set to 1.0.0, 1.0.1, 1.0.2, 2.0.1, etc; The arp matching would be handled by DisplayVersion; Then MarketingVersion would control the version displayed in outputs

Sorry, I meant if package owners wrote these "commit hash versions" to arp registry as DisplayVersion. From sorting perspective, marketing version is not involved in mapping from registry DisplayName to package version.

doc/specs/#3061 - Marketing Version.md
Trenly: That makes sense, and it would be a rarity to use MarketingVersion without DisplayVersion

Trenly:
> Sorry, I meant if package installers wrote these "commit hash versions" to arp registry as DisplayVersion. From sorting perspective, marketing version is not involved in mapping from registry DisplayName to package version.

Yes, agreed. Marketing version, in principle, is just a masking of the Package Version by changing what is displayed.

PR #3064, opened 2023-03-11
Rework extra packaged files mechanism

JohnMcPMS:
Learn more of the dark arts of msbuild syntax ✔️ 
Use that to implement recursive copying of files into the package ✔️ 

## Change
This change is only to how we pull files into the package for development purposes.  It better uses an `ItemGroup` and properties to enable the addition of files/directories by only adding one additional item, rather than 3 properties, 2 items, 1 message and 1 error.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3064)


JohnMcPMS:
The package on the pipeline artifacts works! 🎊 🚢 🍰

yao-msft (approved):

PR #3087, opened 2023-03-18
Add diagnostics to the processor component

JohnMcPMS:
## Change
This change adds the same diagnostics event to the `IConfigurationSetProcessorFactory` as the `ConfigurationProcessor` already has.  It also adds a property to control the minimum level of events, enabling filtering to be done at event publishing rather than consumption.

All of the public facing methods in the processor component are enlightened to send any escaping exceptions to the diagnostics before letting them generate an error.  Additionally, verbose logs are added throughout to enable a better understanding of what/when things are happening.

## Validation
Manual runs, including with verbose logging enabled.  Verified output in log file.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3087)

### DIFF
--- a/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
+++ b/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
@@ -6,6 +6,7 @@
 
 namespace Microsoft.Management.Configuration.Processor
 {
+    using System;
     using Microsoft.Management.Configuration;
     using Microsoft.Management.Configuration.Processor.ProcessorEnvironments;
     using Microsoft.Management.Configuration.Processor.Set;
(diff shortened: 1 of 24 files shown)


src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
```diff
@@ -29,27 +30,66 @@ public ConfigurationSetProcessorFactory(ConfigurationProcessorType type, IConfig
...
+                    }
                 }
+
+                this.OnDiagnostics(DiagnosticLevel.Verbose, "... done creating set processor.");
+
+                return new ConfigurationSetProcessor(processorEnvironment, set) { SetProcessorFactory = this };
+            }
+            catch (Exception ex)
+            {
+                this.OnDiagnostics(DiagnosticLevel.Error, ex.ToString());
```
msftrubengu, line 73: Do you know how much ToString outputs for a RuntimeException? If it doesn't include much from the ErrorRecord we should also log that. I found ScriptStackTrace to be really helpful.

https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.errorrecord?view=powershellsdk-7.3.0#properties

msftrubengu (approved):

PR #3094, opened 2023-03-21
Add logging around specific number of matches

Trenly:

In attempting to diagnose issues with search results, it's helpful to have the number of matches in the search result. This information is sent to telemetry, but is not ever put into the log.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3094)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
+++ b/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
@@ -851,7 +851,6 @@ namespace AppInstaller::CLI::Workflow
         auto& searchResult = context.Get<Execution::Data::SearchResult>();
 
         Logging::Telemetry().LogSearchResultCount(searchResult.Matches.size());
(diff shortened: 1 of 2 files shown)


src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -851,6 +851,7 @@ namespace AppInstaller::CLI::Workflow
         auto& searchResult = context.Get<Execution::Data::SearchResult>();
 
         Logging::Telemetry().LogSearchResultCount(searchResult.Matches.size());
+        AICLI_LOG(Repo, Verbose, << "Search result size: " << searchResult.Matches.size());
```
yao-msft, line 854:
```suggestion
        AICLI_LOG(Cli, Verbose, << "Search result size: " << searchResult.Matches.size());

```

Also, I think we should move this inside LogSearchResultCount() above to be consistent with other telemetry logging too?

yao-msft (approved):

PR #3100, opened 2023-03-23
Improve correlation by keeping arch info declared in manifest arp DisplayName entry

yao-msft:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3100)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3100)

### DIFF
--- a/src/AppInstallerCommonCore/NameNormalization.cpp
+++ b/src/AppInstallerCommonCore/NameNormalization.cpp
@@ -501,4 +501,28 @@ namespace AppInstaller::Utility
     {
         return m_normalizer->NormalizePublisher(publisher);
     }
+
+    std::string NormalizedName::GetNormalizedName(NormalizationField fieldsToInclude) const
+    {
+        std::string result = Name();
+
+        if (WI_IsFlagSet(fieldsToInclude, NormalizationField::Architecture) && m_arch != Utility::Architecture::Unknown)
+        {
+            result += '(' + std::string(Utility::ToString(m_arch)) + ')';
+        }
+
+        return result;
+    }
+
+    NormalizationField NormalizedName::GetNormalizedFields() const
+    {
+        NormalizationField result = NormalizationField::None;
+
+        if (m_arch != Utility::Architecture::Unknown)
+        {
+            result |= NormalizationField::Architecture;
+        }
+
+        return result;
+    }
 }
--- a/src/AppInstallerRepositoryCore/CompositeSource.cpp
+++ b/src/AppInstallerRepositoryCore/CompositeSource.cpp
@@ -1358,11 +1358,21 @@ namespace AppInstaller::Repository
                 {
                     // For installed package to available package mapping,
                     // try the search with NormalizedName with Arch first, if not found, try with all values.
-                    SearchRequest systemReferenceSearch = installedPackageData.CreateInclusionsSearchRequest(Utility::NormalizationField::Architecture);
-                    if (!tryAddAvailablePackage(systemReferenceSearch))
+                    // This can be extended in the future for more granular search requests.
+                    std::vector<SearchRequest> candidateSearches;
+                    if (installedPackageData.PackageNamesContainNormalizationField(Utility::NormalizationField::Architecture))
                     {
-                        systemReferenceSearch = installedPackageData.CreateInclusionsSearchRequest();
-                        tryAddAvailablePackage(systemReferenceSearch);
+                        candidateSearches.emplace_back(installedPackageData.CreateInclusionsSearchRequest(Utility::NormalizationField::Architecture));
+                    }
+                    candidateSearches.emplace_back(installedPackageData.CreateInclusionsSearchRequest());
+
+                    for (const auto& candidateSearch : candidateSearches)
+                    {
+                        if (tryAddAvailablePackage(candidateSearch))
+                        {
+                            // If found available package, break. Otherwise try next search request.
+                            break;
+                        }
                     }
                 }
 
--- a/src/AppInstallerRepositoryCore/CompositeSource.cpp
+++ b/src/AppInstallerRepositoryCore/CompositeSource.cpp
@@ -4,7 +4,6 @@
 #include "CompositeSource.h"
 #include "Microsoft/PinningIndex.h"
 #include <winget/ExperimentalFeature.h>
-#include <winget/NameNormalization.h>
 
 using namespace AppInstaller::Repository::Microsoft;
 using namespace AppInstaller::Settings;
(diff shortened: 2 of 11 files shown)


src/AppInstallerCommonCore/NameNormalization.cpp
```diff
@@ -501,4 +501,28 @@ namespace AppInstaller::Utility
...
         return m_normalizer->NormalizePublisher(publisher);
     }
+
+    std::string NormalizedName::GetNormalizedName(NormalizationField fieldsToInclude) const
+    {
+        std::string result = Name();
+
+        if (WI_IsFlagSet(fieldsToInclude, NormalizationField::Architecture) && m_arch != Utility::Architecture::Unknown)
+        {
+            result += '(' + std::string(Utility::ToString(m_arch)) + ')';
```
Trenly, line 511: What about for packages that just include their architecture and not the parens? Or where it isn't just the arch to string? Won't this still cause issues? I feel like where DisplayName is included in the manifest, the normalization shouldn't occur.
```
  13  AppsAndFeaturesEntries:
  14:   - DisplayName: Bluebeam Revu x64 21
```
```
  34: - DisplayName: calibre 64bit
  35    UpgradeCode: '{5DD881FF-756B-4097-9D82-8C0F11D521EA}'
```
```
  23    - ProductCode: SyncBackPro64_is1_is1
  24:     DisplayName: SyncBackPro x64
```
```
  87:   - DisplayName: 7-Zip 22.01 (x64 edition)
  88      Publisher: Igor Pavlov
```
```
  28    - Publisher: Adobe
  29:     DisplayName: Adobe Acrobat DC (64-bit)
```

yao-msft: Our normalization code uses a regex which should take care of different formats of arch representations. x64, (x64), 64bit etc.

The reason we still need to do normalization on DisplayName comes from various considerations:
1. Including every declared DisplayName explodes the index (which is currently around 13 mb), it will be at least several times bigger. Considering by default winget updates source every 3 mins if new version applicable, it's a lot of data downloading.
2. If DisplayName contains version (or any other Package Version specific string) and we don't normalize it, our correlation will rely on the repo containing every version of the package (with every package version declaring DisplayName). For example, if DisplayName is Foo v1.0.0 and Foo v2.0.0 (2 versions exist in winget-pkgs repo), and if we don't normalize them in index, a local installed version with Foo v1.5.0 will not be able to get correlated. We cannot guarantee winget-pkgs will have every version of a package listed and client should not rely on this assumption.

According to our brief analysis, the arch is the main source of false correlation. And it makes sense to keep arch info specifically. We could expand to include locale, or even version if we decide to in the future. But each expansion will come with the cost of making index multiple times larger, and less correlation success if the installed version is not in the repo. Unless we redesign the index structure in future winget v2 (if winget v2 is a thing).

Trenly: I see; thanks for the detailed explanation!

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -77,12 +78,13 @@ namespace AppInstaller::Repository
...
     {
         PackageMatchField Field;
+        Utility::NormalizationField NameNormalizationField = Utility::NormalizationField::None;
 
         PackageMatchFilter(PackageMatchField f, MatchType t) : RequestMatch(t), Field(f) { EnsureRequiredValues(); }
         PackageMatchFilter(PackageMatchField f, MatchType t, Utility::NormalizedString& v) : RequestMatch(t, v), Field(f) { EnsureRequiredValues(); }
         PackageMatchFilter(PackageMatchField f, MatchType t, const Utility::NormalizedString& v) : RequestMatch(t, v), Field(f) { EnsureRequiredValues(); }
         PackageMatchFilter(PackageMatchField f, MatchType t, Utility::NormalizedString&& v) : RequestMatch(t, std::move(v)), Field(f) { EnsureRequiredValues(); }
-        PackageMatchFilter(PackageMatchField f, MatchType t, std::string_view v1, std::string_view v2) : RequestMatch(t, v1, v2), Field(f) { EnsureRequiredValues(); }
+        PackageMatchFilter(PackageMatchField f, MatchType t, std::string_view v1, std::string_view v2, Utility::NormalizationField n = Utility::NormalizationField::None) : RequestMatch(t, v1, v2), Field(f), NameNormalizationField(n) { EnsureRequiredValues(); }
```
JohnMcPMS, line 87: The searching should not know or understand how normalization is being done.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -1054,14 +1099,17 @@ namespace AppInstaller::Repository
                 auto names = installedVersion->GetMultiProperty(PackageVersionMultiProperty::Name);
                 auto publishers = installedVersion->GetMultiProperty(PackageVersionMultiProperty::Publisher);
 
+                Utility::NameNormalizer normer(Utility::NormalizationVersion::Initial);
```
JohnMcPMS, line 1102: Different versions of the index will use different versions of the normalization; I don't think that this is future proof.  The goal was that from the outside, normalization was known to be used, but not understood.  I don't think that this level of detail is a workable solution for an arbitrary source to implement.

src/AppInstallerRepositoryCore/CompositeSource.cpp
JohnMcPMS, line 1: I wonder if this is part of a larger problem where we are taking the first match rather than treating multiple matches as an issue (and thus resulting in no matches, with some logging of course).

Also, some analysis that I hadn't thought is to look at how many normalized name and publisher values (or any of the system reference string cohort) map onto multiple package ids.  Any time that is happening is likely cause for concern.

JohnMcPMS (changes requested):

src/AppInstallerRepositoryCore/CompositeSource.cpp
Trenly: There's 336 package id's where the `norm_name` maps to another id, and 153 unique `norm_names` that map to more than one package.

[ID_Name_Map_DuplicatesOnly_2023-03-29.csv](https://github.com/microsoft/winget-cli/files/11107012/ID_Name_Map_DuplicatesOnly_2023-03-29.csv)

yao-msft:
After offline discussion, we will address above comments by moving all normalization related logic inside index search logic.
And add a new enum called SearchPurpose in search request, together with source type, SearchInternal can decide on the fallback logic if needed.

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -95,6 +95,17 @@ namespace AppInstaller::Repository
         }
     };
 
+    // The search purpose of the search request.
+    enum class SearchPurpose
+    {
+        // Default search purpose.
+        Default,
+        // The result is used for correlation to an installed package.
+        CorrelationToInstalled,
```
yao-msft, line 104: In the end I have to use 3 enums for search purpose because the sqlite index itself does not know if it's installed source, this info is only available at Source level

JohnMcPMS (approved):

Issue #3025, opened 2023-02-28
Change WinGet behavior when the Microsoft Store UI is disabled


denelon:
### Description of the new feature / enhancement

I want to be able to leverage WinGet to install packages in the Microsoft Store if the Microsoft Store UI is disabled.

*Note: The current behavior in WinGet is to check to see if the Microsoft Store UI is disabled. If the Microsoft Store UI is disabled, WinGet blocks installing Microsoft Store Apps.*

This is targeted as a 1.6 release milestone to afford enterprise customers the time necessary to configure policy appropriately given the nature of the change.

Related to:
* #3024

BarbieAction:
Settings Require Private Store only on Windows 11 will block the store and allow Winget.
https://learn.microsoft.com/en-gb/windows/client-management/mdm/policy-csp-ApplicationManagement?WT.mc_id=Portal-fx#requireprivatestoreonly

PR #3105, opened 2023-03-24
Support to Bypass Store Client App Policy When Called Through COM

PaulJYim:
This is part of the effort to support scenarios where the MS Store is disabled.

Added a flag that bypasses the store policy checks.

Related to https://github.com/microsoft/winget-cli/issues/3025
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3105)


src/PowerShell/Microsoft.WinGet.Client/Commands/Common/BaseInstallCommand.cs
```diff
@@ -57,7 +57,13 @@ public string Location
         /// Gets or sets a value indicating whether to skip the installer hash validation check.
         /// </summary>
         [Parameter(ValueFromPipelineByPropertyName = true)]
-        public SwitchParameter AllowHashMismatch { get; set; }
+        public SwitchParameter AllowHashMismatch { get; set; }
+
+        /// <summary>
```
yao-msft, line 62: We do not want to expose this in Powershell module yet, so please revert changes to this file

src/AppInstallerCLIE2ETests/Interop/InstallInterop.cs
```diff
@@ -599,5 +599,28 @@ public void GetApplicableInstaller()
             Assert.AreEqual(PackageInstallerScope.User, packageInstallerInfo.Scope);
             Assert.AreEqual("en-US", packageInstallerInfo.Locale);
         }
+
+        /// <summary>
+        /// Test install with BypassStorePolicy Enabled.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Test]
+        public async Task InstallMsixStoreBypass()
```
yao-msft, line 608: This test probably will not work. We don't have any e2e tests targeting msstore types

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -249,7 +255,9 @@ namespace AppInstaller::CLI::Workflow
 
         // Policy check
         AppInstallManager installManager;
-        if (installManager.IsStoreBlockedByPolicyAsync(s_StoreClientName, s_StoreClientPublisher).get())
+        bool bypassStorePolicy = WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::BypassStorePolicy);
+
+        if (installManager.IsStoreBlockedByPolicyAsync(s_StoreClientName, s_StoreClientPublisher).get() && !bypassStorePolicy)
```
yao-msft, line 260: Do we want to bypass GetIsAppAllowedToInstallAsync() check below? If not, we could use a better name to represent "we are only bypassing if the Store client is disabled, other checks continue to work as expected", like BypassIsStoreClientBlockedPolicyCheck?

yao-msft, line 260: nit:

if (!WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::BypassStorePolicy) && installManager.IsStoreBlockedByPolicyAsync())

To avoid a potential unnecessary store policy api call

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -115,6 +115,12 @@ namespace AppInstaller::CLI::Workflow
             }
             else if (result.Status() == GetEntitlementStatus::ServerError)
             {
+                bool bypassStorePolicy = WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::BypassStorePolicy);
```
yao-msft, line 118: This is not Store client policy check, please revert

yao-msft: Our previous conversation with the api owner says these apis do not check for the policy. I'll need to try locally to confirm

src/AppInstallerCLITests/InstallFlow.cpp
```diff
@@ -536,6 +536,30 @@ TEST_CASE("MSStoreInstallFlowWithTestManifest", "[InstallFlow][workflow]")
...
+TEST_CASE("MSStoreInstallFlowWithTestManifestBypassStoreClientBlockedByPolicy", "[InstallFlow][workflow]")
+{
+    TestCommon::TempFile installResultPath("TestMSStoreInstalled.txt");
+
+    std::ostringstream installOutput;
+    TestContext context{ installOutput, std::cin };
+    auto previousThreadGlobals = context.SetForCurrentThread();
+    OverrideForMSStore(context, false);
+    context.Args.AddArg(Execution::Args::Type::Manifest, TestDataFile("InstallFlowTest_MSStore.yaml").GetPath().u8string());
+    context.SetFlags(Execution::ContextFlag::BypassIsStoreClientBlockedPolicyCheck);
```
yao-msft, line 548: I think if there's no previous setup to disable the policy, the related code path is not tested. we should do the proper policy disable setup(and revert in test end) and do a positive, negative install attempt with and without the flag.

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
PaulJYim: Changed the name!

yao-msft (approved):

PR #3112, opened 2023-03-28
Pass COM caller name to rest source in request header

yao-msft:
Added unit tests and tested manually.


ryfu-msft:
>     }

Nit: These seem to be almost identical except for caller and "winget-cli". Maybe create a helper so we don't have 2 places with strings?

---
Refers to: src/AppInstallerCommonCore/Runtime.cpp:792 in d632e19. [](commit_id = d632e194dcf04f473d2ac23e35151273e9a15629, deletion_comment = False)

ryfu-msft (approved):

PR #3118, opened 2023-03-30
Enforce single install across winget processes

JohnMcPMS:
## Change
This change uses a named mutex to enforce a single install/uninstall operation is occurring for the entire session.  The current behavior is that every process was able to install a single thing at a time, but as we have more usage this could lead to a potential issue.

The mutex is acquired and held for the workflow task that routes out to the installer type specific tasks.  Some installer types are explicitly excluded, as they are known to contain their own synchronization internally.

In addition to the mutex, the ability to set a message to show with the indefinite progress (spinner) was added.  This allows for the message to be removed when the wait is over.  The message was not implemented into the progress bar, nor is it being sent to COM progress.

The COM scheduling was not made aware of the mutex, and so it will execute as it has in the past, blocking on the mutex when appropriate.  This could be improved in the future to enable the parallel capable installer types to operate.  One potential option would be a second installer processing thread that only operated on the excluded types, and thus would never block.

## Validation
Manually validated the user experience with an interactive install.
Automated tests are added to ensure that the lock is held during the installation workflow.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3118)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionProgress.cpp
+++ b/src/AppInstallerCLICore/ExecutionProgress.cpp
@@ -151,6 +151,24 @@ namespace AppInstaller::CLI::Execution
                 LOG_HR(E_UNEXPECTED);
             }
         }
+
+        void ProgressVisualizerBase::ClearLine()
+        {
+            if (UseVT())
+            {
+                m_out << TextModification::EraseLineEntirely << '\r';
+            }
+            else
+            {
+                // Best effort when no VT (arbitrary number of spaces that seems to work)
+                m_out << "\r                                                              \r";
+            }
+        }
+
+        void ProgressVisualizerBase::SetMessage(std::string_view message)
+        {
+            m_message = message;
+        }
     }
 
     void IndefiniteSpinner::ShowSpinner()
@@ -188,17 +206,19 @@ namespace AppInstaller::CLI::Execution
             }
 
             // Indent two spaces for the spinner, but three here so that we can overwrite it in the loop.
-            m_out << "   ";
+            std::string_view indent = "   ";
 
             for (size_t i = 0; !m_canceled; ++i)
             {
                 constexpr size_t repetitionCount = 20;
                 ApplyStyle(i % repetitionCount, repetitionCount, true);
-                m_out << '\b' << spinnerChars[i % ARRAYSIZE(spinnerChars)] << std::flush;
+                m_out << '\r' << indent << spinnerChars[i % ARRAYSIZE(spinnerChars)];
+                m_out.RestoreDefault();
+                m_out << ' ' << m_message << std::flush;
                 Sleep(250);
             }
 
-            m_out << "\b \r";
+            ClearLine();
 
             if (UseVT())
             {
@@ -217,6 +237,7 @@ namespace AppInstaller::CLI::Execution
             ClearLine();
         }
 
+        // TODO: Progress bar does not currently use message
         if (UseVT())
         {
             ShowProgressWithVT(current, maximum, type);
@@ -252,19 +273,6 @@ namespace AppInstaller::CLI::Execution
             }
 
             m_isVisible = false;
-        }
-    }
-
-    void ProgressBar::ClearLine()
-    {
-        if (UseVT())
-        {
-            m_out << TextModification::EraseLineEntirely << '\r';
-        }
-        else
-        {
-            // Best effort when no VT (arbitrary number of spaces that seems to work)
-            m_out << "\r                                                              \r";
         }
     }
 
--- a/src/AppInstallerCLICore/ExecutionProgress.cpp
+++ b/src/AppInstallerCLICore/ExecutionProgress.cpp
@@ -160,8 +160,7 @@ namespace AppInstaller::CLI::Execution
             }
             else
             {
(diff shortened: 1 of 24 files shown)


src/AppInstallerCLICore/ExecutionProgress.cpp
```diff
@@ -151,6 +151,24 @@ namespace AppInstaller::CLI::Execution
...
+        void ProgressVisualizerBase::ClearLine()
+        {
+            if (UseVT())
+            {
+                m_out << TextModification::EraseLineEntirely << '\r';
+            }
+            else
+            {
+                // Best effort when no VT (arbitrary number of spaces that seems to work)
+                m_out << "\r                                                              \r";
```
florelis, line 164: Nit: How about something like `m_out << '\r' << std::string(N, ' ') << '\r';` so that we know how many spaces there are and can extend it if needed?

src/AppInstallerCommonCore/Synchronization.cpp
```diff
@@ -259,4 +262,48 @@ namespace AppInstaller::Synchronization
 
         return result;
     }
+
+    CrossProcessInstallLock::CrossProcessInstallLock()
+    {
+        m_mutex.create(L"WinGetCrossProcessInstallLock", 0, SYNCHRONIZE);
```
florelis, line 268: Should we be worried about some third party acquiring this named mutex and just making winget unusable?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1821,4 +1821,7 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Reboot required to fully enable the Windows Feature(s); proceeding due to --force</value>
     <comment>"Windows Feature(s)" is the name of the options Windows features setting.</comment>
   </data>
+  <data name="InstallWaitingOnAnother" xml:space="preserve">
+    <value>Waiting on another install or uninstall to complete...</value>
```
florelis, line 1825: Nit: Shouldn't it be "Waiting for"?

src/AppInstallerCLITests/Synchronization.cpp
```diff
@@ -24,7 +24,7 @@ TEST_CASE("CPRWL_MultipleReaders", "[CrossProcessReaderWriteLock]")
     otherThread.detach();
 
     // Wait up to a second for the other thread to do one thing...
```
florelis, line 26: Nit: Update comment

src/AppInstallerCLITests/Synchronization.cpp
```diff
@@ -116,10 +116,60 @@ TEST_CASE("CPRWL_CancelEndsWait", "[CrossProcessReaderWriteLock]")
...
+    REQUIRE(!signal.wait(500));
 
     progress.Cancel();
 
     // Upon release of the writer, the other thread should signal
-    REQUIRE(signal.wait(1000));
+    REQUIRE(signal.wait(500));
+}
+
+TEST_CASE("CPIL_BlocksOthers", "[CrossProcessInstallLock]")
```
florelis, line 127: Nit: Do we really need to abbreviate CPIL?

src/AppInstallerCLITests/Synchronization.cpp
```diff
@@ -116,10 +116,60 @@ TEST_CASE("CPRWL_CancelEndsWait", "[CrossProcessReaderWriteLock]")
...
+    wil::unique_event signal;
+    signal.create();
+    AppInstaller::ProgressCallback progress;
+
+    CrossProcessInstallLock mainThreadLock;
+    mainThreadLock.Acquire(progress);
+
+    std::thread otherThread([&signal, &progress]() {
+        CrossProcessInstallLock otherThreadLock;
+        otherThreadLock.Acquire(progress);
```
florelis, line 163: Should we `REQUIRE(!otherThreadLock.Acquire(progres);`?

florelis (approved):

src/AppInstallerCommonCore/Synchronization.cpp
JohnMcPMS: We could add a random suffix to the mutex (and any other named mutexes), but we would have to write that information into a location that the unelevated user token can access.  While it would make it slightly harder to execute this attack, it wouldn't be any harder from a security perspective.

I suppose that we could put a conditional ACE on it to require our package identity (when packaged) to even read from it.  Even that is just another minor bump in the road that can be overcome with more code (but not any more security access).

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: Turns out that my southern US origins may be the source of this based on a brief look.  They both are "correct" but various sources give differing opinions, but that one vs the other is colloquial.  I have no problem changing it though.

src/AppInstallerCLICore/ExecutionProgress.cpp
JohnMcPMS: Changed it to get the current console width.

PR #3127, opened 2023-04-04
Bug fixes

msftrubengu:
This PR contains some bug fixes for Configuration Processor

- Set 2.* as MaximumVersion for PsDesiredStateConfiguration to explicitly not support DSC v3 DscModuleV2.cs
- Set ErrorAction to Stop for Invoke-DscResource calls. If the cmdlet write to the error stream, we will now throw an exception and add the error to the exception message.
- Support nested maps in settings.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3127)

### DIFF
--- a/src/Microsoft.Management.Configuration.Processor/DscModules/DscModuleV2.cs
+++ b/src/Microsoft.Management.Configuration.Processor/DscModules/DscModuleV2.cs
@@ -11,8 +11,10 @@ namespace Microsoft.Management.Configuration.Processor.DscModule
     using System.Collections.ObjectModel;
     using System.Linq;
     using System.Management.Automation;
+    using System.Text;
     using Microsoft.Management.Configuration.Processor.DscResourcesInfo;
     using Microsoft.Management.Configuration.Processor.Exceptions;
+    using Microsoft.Management.Configuration.Processor.Extensions;
     using Microsoft.Management.Configuration.Processor.Helpers;
     using Microsoft.PowerShell.Commands;
     using Windows.Foundation.Collections;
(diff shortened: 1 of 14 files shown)


src/Microsoft.Management.Configuration.Processor/DscModules/DscModuleV2.cs
```diff
@@ -26,6 +28,9 @@ internal class DscModuleV2 : IDscModule
         private const string InDesiredState = "InDesiredState";
         private const string RebootRequired = "RebootRequired";
 
+        // Fake max version to avoid using v3 and allowing new v2 versions to be used.
+        private const string MaxVersion = "2.*";
```
JohnMcPMS, line 32: Why not `Modules.PSDesiredStateConfigurationMaxVersion`?

src/Microsoft.Management.Configuration.Processor/DscModules/DscModuleV2.cs
```diff
@@ -95,9 +101,20 @@ public IReadOnlyList<DscResourceInfoInternal> GetAllDscResources(PowerShell pwsh
             var getResult = pwsh.AddCommand(this.InvokeDscResourceCmd)
                                 .AddParameters(PrepareInvokeParameters(name, settings, moduleSpecification))
                                 .AddParameter(Parameters.Method, DscMethods.Get)
-                                .Invoke()
+                                .InvokeAndStopOnError()
                                 .FirstOrDefault();
 
+            if (pwsh.HadErrors)
```
JohnMcPMS, line 107: Helper method?  I don't understand why it was pulled out of the other one, but at least it could be one here to prevent the block from being copied three times.

JohnMcPMS (approved):

Issue #2298, opened 2022-07-01
Display the path to an installed package


sohang3112:
### Description of the new feature / enhancement

After installing a package using `winget install <PACKAGE>`, running `winget show <PACKAGE>` should also show the absolute path of the installed executable.

### Proposed technical implementation details

`winget show <PACKAGE>` outputs several fields (like `Version`, `Description`, etc.) describing the package.
Another field `Installed Path` should also be added, that shows the absolute path of the installed executable.

wmstack:
I suggest that the subcommand should be called `which`, or at least that both `which` and `show` should be aliases for the same command. For example, that is how it is in Scoop, and that is how e.g., Linux does it with the `which` command. Having a `show ` would be counter-intuitive to already-existing users.

fredx30:
This thread was the first hit when i went looking for a default install dir for winget. 📦 👍

alex-jitbit:
I agree, I just installed `winget install ntop` and it took me ages to realize that winget had just thrown it into `%LOCALAPPDATA%\Microsoft\WinGet\Packages\gsass1.NTop_Microsoft.Winget.Source_8wekyb3d8bbwe` silently, without adding any `PATH`. Had to run an entire disk search.

denelon:
There are some subtle behaviors here. I'll try to clarify them just to help everyone's understanding.

Use case 1:
If you are in "Administrative" mode when you install, or if you have "Developer mode" enabled, WinGet creates a symbolic link in a common directory to the package. 

Use case 2:
If you are in "User" mode, then we have to add unique entries to the path per package.

In either case, the first time you install a portable package, the package will not immediately be available on the path. You will need to restart the terminal to be able to run the package.

For use case 1, after you have installed the first portable package and restarted the terminal, the path entry for the symbolic links is in place, so subsequent installations shouldn't require a terminal restart.

For use case 2, we have to add unique path entries, so each portable package will require a restart of the terminal to get the refereshed "Path" environment variable.

alex-jitbit:
@danelon I did restart the Windows terminal. I even checked the PATH env manually, the program was still not there. I don't know if it's winget or the package author, but providing a way to manually investigate problems like this manually would be helpful

denelon:
I ran `winget install ntop` in user mode in Windows Terminal (PowerShell).
I was given the message that I needed to restart my terminal.

[...]

PR #3128, opened 2023-04-04
Show Portable Installation Directories with --info

Trenly:
  * [#2298](https://github.com/microsoft/winget-cli/issues/2298)

```
PS C:\WINDOWS\system32> wingetdev --info
Windows Package Manager (Preview) v1.5.2-preview

Windows: Windows.Desktop v10.0.19044.2728
System Architecture: X64
Package: WinGetDevCLI v0.0.2.0

Winget Directories
--------------------------------------------------------------------------------------------------------------
Logs                               %LOCALAPPDATA%\Packages\WinGetDevCLI_8wekyb3d8bbwe\LocalState\DiagOutputDir
User Settings                      %LOCALAPPDATA%\Packages\WinGetDevCLI_8wekyb3d8bbwe\LocalState\settings.json
Portable Package Root (User)       %LOCALAPPDATA%\Microsoft\WinGet\Packages\Test
Portable Package Root (x64)        C:\Program Files\WinGet\Packages

Links

```

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3128)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3128)

### DIFF
--- a/src/AppInstallerCLICore/Commands/RootCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/RootCommand.cpp
@@ -225,6 +225,7 @@ namespace AppInstaller::CLI
 
             info << std::endl << Resource::String::Logs(Utility::LocIndView{ Runtime::GetPathTo(Runtime::PathName::DefaultLogLocation, true).u8string() }) << std::endl;
             info << std::endl << Resource::String::UserSettings(Utility::LocIndView{ UserSettings::SettingsFilePath(true).u8string() }) << std::endl;
+
             info << std::endl;
 
             Execution::TableOutput<2> links{ context.Reporter, { Resource::String::Links, {} } };
--- a/src/AppInstallerCLICore/Commands/RootCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/RootCommand.cpp
@@ -130,11 +130,11 @@ namespace AppInstaller::CLI
         void OutputPortableDirectories(Execution::Context& context)
         {
             Execution::TableOutput<2> portableRoots{ context.Reporter, { Resource::String::PortableDirectoriesHeader, {} } };
-            portableRoots.OutputLine({ "Portable Links Directory (User)", Runtime::GetPathTo(Runtime::PathName::PortableLinksUserLocation, true).string() });
-            portableRoots.OutputLine({ "Portable Links Directory (Machine)", Runtime::GetPathTo(Runtime::PathName::PortableLinksMachineLocation, true).string() });
-            portableRoots.OutputLine({ "Portable Package Root (User)", Runtime::GetPathTo(Runtime::PathName::PortablePackageUserRoot, true).string() });
-            portableRoots.OutputLine({ "Portable Package Root (x86)", Runtime::GetPathTo(Runtime::PathName::PortablePackageMachineRootX86, true).string() });
-            portableRoots.OutputLine({ "Portable Package Root (x64)", Runtime::GetPathTo(Runtime::PathName::PortablePackageMachineRootX64, true).string() });
+            portableRoots.OutputLine({ Resource::LocString{ Resource::String::PortableLinksUser }, Runtime::GetPathTo(Runtime::PathName::PortableLinksUserLocation, true).string() });
+            portableRoots.OutputLine({ Resource::LocString{ Resource::String::PortableLinksMachine }, Runtime::GetPathTo(Runtime::PathName::PortableLinksMachineLocation, true).string() });
+            portableRoots.OutputLine({ Resource::LocString{ Resource::String::PortableRootUser }, Runtime::GetPathTo(Runtime::PathName::PortablePackageUserRoot, true).string() });
+            portableRoots.OutputLine({ Resource::LocString{ Resource::String::PortableRoot64 }, Runtime::GetPathTo(Runtime::PathName::PortablePackageMachineRootX86, true).string() });
+            portableRoots.OutputLine({ Resource::LocString{ Resource::String::PortableRoot86 }, Runtime::GetPathTo(Runtime::PathName::PortablePackageMachineRootX64, true).string() });
             portableRoots.Complete();
             context.Reporter.Info() << std::endl;
         }
--- a/src/AppInstallerCLICore/Commands/RootCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/RootCommand.cpp
@@ -127,15 +127,31 @@ namespace AppInstaller::CLI
             adminSettingsTable.Complete();
         }
 
-        void OutputPortableDirectories(Execution::Context& context)
+        void OutputKeyDirectories(Execution::Context& context)
         {
-            Execution::TableOutput<2> portableRoots{ context.Reporter, { Resource::String::PortableDirectoriesHeader, {} } };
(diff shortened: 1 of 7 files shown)


src/AppInstallerCLICore/Commands/RootCommand.cpp
```diff
@@ -126,6 +126,18 @@ namespace AppInstaller::CLI
             }
             adminSettingsTable.Complete();
         }
+
+        void OutputPortableDirectories(Execution::Context& context)
+        {
+            Execution::TableOutput<2> portableRoots{ context.Reporter, { Resource::String::PortableDirectoriesHeader, {} } };
+            portableRoots.OutputLine({ "Portable Links Directory (User)", Runtime::GetPathTo(Runtime::PathName::PortableLinksUserLocation, true).string() });
```
yao-msft, line 133: I think we need to localize these table entry names as well?

src/AppInstallerCommonCore/Public/AppInstallerRuntime.h
```diff
@@ -112,10 +112,10 @@ namespace AppInstaller::Runtime
 
     // Gets the PathDetails used for the given path.
     // This is exposed primarily to allow for testing, GetPathTo should be preferred.
-    PathDetails GetPathDetailsFor(PathName path);
+    PathDetails GetPathDetailsFor(PathName path, bool anonymize = false);
 
     // Gets the path to the requested location.
-    std::filesystem::path GetPathTo(PathName path);
+    std::filesystem::path GetPathTo(PathName path, bool anonymize = false);
```
yao-msft, line 118: nit: I don't think name anonymize fits here very well. Anonymize is quite context specific. Maybe bool forDisplay?

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -511,7 +511,7 @@ namespace AppInstaller::Runtime
 
             if (path == PathName::DefaultLogLocationForDisplay)
```
yao-msft, line 512: DefaultLogLocationForDisplay can be removed if forDisplay is exposed.

src/AppInstallerCommonCore/Public/winget/Filesystem.h
```diff
@@ -41,6 +41,10 @@ namespace AppInstaller::Filesystem
     // Returns true if replacement happened, false otherwise.
     bool ReplaceCommonPathPrefix(std::filesystem::path& source, const std::filesystem::path& prefix, std::string_view replacement);
 
+    // If the path begins with a common prefix for APPDATA or LOCALAPPDATA, replace it with the anonymized version
+    // Returns true if replacement happened, false otherwise.
+    bool AnonymizeFilePath(std::filesystem::path& source);
```
yao-msft, line 46: I don't think this fits as a common util method. We may not be covering all cases. The "replace with env var" thing depends on caller intention, some  may want %Programfiles% replaced, some may want %UserProfile% replaced. By doing multiple attempts in one method here, at least some attempts will be unnecessary. For the GetPathTo() result, I think it would better call ReplaceCommonPathPrefix with corresponding "what to replace" directly inside the GetPathTo() implementation.

src/AppInstallerCLICore/Resources.h
```diff
@@ -311,6 +311,12 @@ namespace AppInstaller::CLI::Resource
         WINGET_DEFINE_RESOURCE_STRINGID(PortableHashMismatchOverridden);
         WINGET_DEFINE_RESOURCE_STRINGID(PortableHashMismatchOverrideRequired);
         WINGET_DEFINE_RESOURCE_STRINGID(PortableAliasAdded);
+        WINGET_DEFINE_RESOURCE_STRINGID(PortableDirectoriesHeader);
+        WINGET_DEFINE_RESOURCE_STRINGID(PortableLinksMachine);
```
yao-msft, line 315: nit: these definitions should be in alphabetical order

src/AppInstallerCLICore/Commands/RootCommand.cpp
```diff
@@ -213,7 +225,6 @@ namespace AppInstaller::CLI
 
             info << std::endl << Resource::String::Logs(Utility::LocIndView{ Runtime::GetPathTo(Runtime::PathName::DefaultLogLocationForDisplay).u8string() }) << std::endl;
```
yao-msft, line 226: This needs to be updated since this entry is removed

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -564,23 +567,31 @@ namespace AppInstaller::Runtime
...
             result.Path /= GetRuntimePathStateName();
+
             if (path == PathName::Temp)
             {
                 result.SetOwner(ACEPrincipal::CurrentUser);
                 result.ACL[ACEPrincipal::System] = ACEPermissions::All;
                 result.ACL[ACEPrincipal::Admins] = ACEPermissions::All;
             }
+
+            if (path == PathName::DefaultLogLocation && forDisplay)
```
yao-msft, line 589:
```suggestion
            if (forDisplay)

```

I understand you were trying to keep previous behavior, but for this one, forDisplay itself seems enough

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -542,6 +540,11 @@ namespace AppInstaller::Runtime
         case PathName::PortableLinksUserLocation:
         case PathName::PortableLinksMachineLocation:
             result = GetPathDetailsCommon(path);
+            if (forDisplay)
```
yao-msft, line 543: Sorry I may not make my previous comment clear enough. I was thinking to always call replace once. 

Is RoamingAppData used in these paths? I did not find the usage in this file. If not used, we should remove the RoamingAppData replace attempt.

If there are places using RoamingAppData, the paths should be broken down to 2 groups to avoid unnecessary replace prefix calls. Like:

case PathName::1:
case PathName::2:
case PathName::3:
            result = GetPathDetailsCommon(path);
            if (forDisplay)
            {
                ReplaceCommonPathPrefix(result.Path, GetKnownFolderPath(FOLDERID_LocalAppData), "%LOCALAPPDATA%");
            }
            break;
case PathName::4:
case PathName::5:
case PathName::6:
             result = GetPathDetailsCommon(path);
            if (forDisplay)
            {
                ReplaceCommonPathPrefix(result.Path, GetKnownFolderPath(FOLDERID_RoamingAppData), "%APPDATA%");
            }
            break;

Trenly: Winget doesn’t use roaming appdata by default, but since it is a runtime path, it's possible a user has changed the default to be in the roaming appdata folder, so I was trying to cover that possibility

yao-msft: Yes, I think the roaming replace attempt should be removed. We are just trying to replace the value returned from one line above GetPathDetailsCommon(), where only LocalAppData is used. If user configures to use other paths in settings, it can be any path, we'll not be guessing their settings.

By the way, after looking at the code a bit closer, I think we should just try the LocalAppData replace with PortableLinksUserLocation and PortablePackageUserRoot only, for others, the replace seems very unlikely.

Trenly: Will do

yao-msft (approved):

PR #3129, opened 2023-04-05
Use in-proc winget in PowerShell module when running as SYSTEM

JohnMcPMS:
## Change
This change uses the in-proc WinRT binaries and activation path when running as `SYSTEM` in the PowerShell module.  It also makes `Assert-WinGetPackageManager` succeed as long as the OS version is high enough in that case.

There are two known caveats with this that could be addressed in a future change if needed:
1. These newly added dlls are still dependent on the VC runtime, and thus will not work unless it has been installed. This seems acceptable for development scenarios. For shipping a module without this dependency, we will need to build a statically linked binary, but only on the internal, release pipeline.
2. The code currently doesn't behave well in an STA thread, meaning that when used from `SYSTEM` the module requires running PowerShell with the MTA option.

## Validation
Ran module in user and `SYSTEM` contexts using the `Initialize-LocalWinGetModules.ps1` script.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3129)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3129)


src/PowerShell/Microsoft.WinGet.Client/Common/WinGetIntegrity.cs
```diff
@@ -26,6 +26,18 @@ internal static class WinGetIntegrity
         /// <param name="expectedVersion">Expected version.</param>
         public static void AssertWinGet(PSCmdlet psCmdlet, string expectedVersion)
         {
+            // In-proc shouldn't have other dependencies and thus should be ok.
```
yao-msft, line 29: >// In-proc shouldn't have other dependencies and thus should be ok. [](http://example.com/codeflow?start=12&length=67)

In-proc requires unpackaged vc redistributable. Do we want to or have a good way to check that?

yao-msft (approved):
:shipit:

src/PowerShell/Microsoft.WinGet.Client/Common/WinGetIntegrity.cs
JohnMcPMS: This was caveat 1 from my PR summary.  We will need to statically link the runtime in to be able to release this in a self-contained form.  My hope is that we can do that with some form of property override on the internal release build.  If that doesn't work, then a future PR will need to create a new build flavor that only changes to static linkage.

PR #3135, opened 2023-04-06
Add support for bool, strings and arrays in Configuration settings

msftrubengu:
- Add support to winget's YAML wrapper parser to assign a bool or int type for unquoted scalar values if can be converted. If a string is desired, then our recommendation will be to quote it.
- Fix configuration show output to print non string settings values, specifically ints and bools.
- Processor now handles arrays by checking if a value set contains a "treatAsArray" key.
- Add tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3135)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
@@ -121,26 +121,26 @@ namespace AppInstaller::CLI::Workflow
         {
             Utility::LocIndString indentString{ std::string(indent, ' ') };
 
-            std::vector<IKeyValuePair<winrt::hstring, winrt::Windows::Foundation::IInspectable>> arrayValues;
+            std::vector<std::pair<int, winrt::Windows::Foundation::IInspectable>> arrayValues;
             for (const auto& arrayValue : valueSetArray)
             {
                 if (arrayValue.Key() != L"treatAsArray")
                 {
-                    arrayValues.emplace_back(arrayValue);
+                    arrayValues.emplace_back(std::make_pair(std::stoi(arrayValue.Key().c_str()), arrayValue.Value()));
                 }
             }
 
             std::sort(
                 arrayValues.begin(),
                 arrayValues.end(),
-                [](const IKeyValuePair<winrt::hstring, winrt::Windows::Foundation::IInspectable>& a, const IKeyValuePair<winrt::hstring, winrt::Windows::Foundation::IInspectable>& b)
+                [](const std::pair<int, winrt::Windows::Foundation::IInspectable>& a, const std::pair<int, winrt::Windows::Foundation::IInspectable>& b)
                 {
-                    return a.Key() < b.Key();
+                    return a.first < b.first;
                 });
 
             for (const auto& arrayValue : arrayValues)
             {
-                auto arrayObject = arrayValue.Value();
+                auto arrayObject = arrayValue.second;
                 IPropertyValue arrayProperty = arrayObject.try_as<IPropertyValue>();
 
                 out << indentString << "-";
--- a/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
@@ -104,7 +104,7 @@ namespace AppInstaller::CLI::Workflow
                 out << ' ' << Utility::ConvertToUTF8(property.GetString()) << '\n';
                 break;
             case PropertyType::Boolean:
-                out << ' ' << (property.GetBoolean() ? "true" : "false") << '\n';
+                out << ' ' << (property.GetBoolean() ? Utility::LocIndView("true") : Utility::LocIndView("false")) << '\n';
                 break;
             case PropertyType::Int64:
                 out << ' ' << property.GetInt64() << '\n';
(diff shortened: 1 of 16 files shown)


src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -96,36 +96,84 @@ namespace AppInstaller::CLI::Workflow
...
-        void OutputValueSet(OutputStream& out, const ValueSet& valueSet, size_t indent)
+        void OutputPropertyValue(OutputStream& out, const IPropertyValue property)
+        {
+            switch (property.Type())
+            {
+            case PropertyType::String:
+                out << ' ' << Utility::ConvertToUTF8(property.GetString()) << '\n';
+                break;
+            case PropertyType::Boolean:
+                out << ' ' << (property.GetBoolean() ? "true" : "false") << '\n';
```
JohnMcPMS, line 107: These boolean concept names should probably be localized... c'est vrais.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -96,36 +96,84 @@ namespace AppInstaller::CLI::Workflow
...
+
+        void OutputValueSet(OutputStream& out, const ValueSet& valueSet, size_t indent, bool isArray)
         {
             Utility::LocIndString indentString{ std::string(indent, ' ') };
+            bool processedFirstElement = false;
 
             for (const auto& value : valueSet)
             {
-                out << indentString << Utility::ConvertToUTF8(value.Key()) << ':';
+                if (isArray && !processedFirstElement)
```
JohnMcPMS, line 125: I would check for `treatAsArray` first and defer to a separate `OuputValueSetAsArray` function to keep things less complicated in this function.

PS: Thought experiment:
```YAML
objectButArray:
  0: Item0
  1: Item1
  treatAsArray: true
```

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -96,36 +96,84 @@ namespace AppInstaller::CLI::Workflow
...
+                    OutputPropertyValue(out, property);
                 }
                 else
                 {
                     // If not an IPropertyValue, it must be a ValueSet
                     ValueSet subset = object.as<ValueSet>();
+
                     out << '\n';
-                    OutputValueSet(out, subset, indent + 2);
+                    if (subset.HasKey(L"treatAsArray"))
```
JohnMcPMS, line 150: See, you basically already do check, just in here and you inlined the function yourself...  Basically I think that this can be more readable if you separate it and don't pre-optimize by just assuming any `ValueSet` might be an array.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -96,36 +96,84 @@ namespace AppInstaller::CLI::Workflow
...
+                    {
+                        size_t arrayIndent = indent + 2;
+                        Utility::LocIndString arrayIndentString{ std::string(arrayIndent, ' ') };
+                        for (const auto& arrayValue : subset)
```
JohnMcPMS, line 154: I don't think that there is any guarantee about the iteration order of a `ValueSet`.  This code would likely need to get items in their numerical order, or sort them all before iterating.

src/AppInstallerSharedLib/Yaml.cpp
```diff
@@ -146,6 +195,38 @@ namespace AppInstaller::YAML
...
+        // tag. Here we just try our best and assume that if the value is unquoted
+        // then is not necessarily a string.
+        // TODO: handle float and timestamps
+        if (!isQuoted && this->GetTagType() == TagType::Str)
+        {
+            // Integer
+            // 0 | -? [1-9] [0-9]*
+            try
+            {
+                this->as<int64_t>();
```
JohnMcPMS, line 212: This seems like it might be very noisy for logs.  Could we have a `try_as` that returns a `std::optional` and uses functions that don't end up logging?

src/Microsoft.Management.Configuration.Processor/Extensions/ValueSetExtensions.cs
```diff
@@ -38,5 +49,35 @@ public static Hashtable ToHashtable(this ValueSet valueSet)
...
+        private static List<object> ConvertValueSetToArray(ValueSet valueSet)
+        {
+            if (!valueSet.ContainsKey(TreatAsArray))
+            {
+                throw new InvalidOperationException();
+            }
+
+            var result = new List<object>();
+
+            foreach (var keyValuePair in valueSet)
```
JohnMcPMS, line 62: Again, iteration order might not be preserved.

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
msftrubengu: ei

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -115,24 +115,81 @@ namespace AppInstaller::CLI::Workflow
...
+                    arrayValues.emplace_back(arrayValue);
+                }
+            }
+
+            std::sort(
+                arrayValues.begin(),
+                arrayValues.end(),
+                [](const IKeyValuePair<winrt::hstring, winrt::Windows::Foundation::IInspectable>& a, const IKeyValuePair<winrt::hstring, winrt::Windows::Foundation::IInspectable>& b)
+                {
+                    return a.Key() < b.Key();
```
JohnMcPMS, line 138: These keys are strings and won't sort numerically.

src/AppInstallerSharedLib/AppInstallerStrings.cpp
```diff
@@ -175,6 +175,32 @@ namespace AppInstaller::Utility
         return result;
     }
 
+    std::optional<std::wstring> TryConvertToUTF16(std::string_view input, UINT codePage)
+    {
+        if (input.empty())
+        {
+            return {};
```
JohnMcPMS, line 182: This returns an empty optional, which isn't what I would expect.  It should return an optional with an empty string in it.

src/AppInstallerSharedLib/Public/winget/Yaml.h
```diff
@@ -128,6 +128,14 @@ namespace AppInstaller::YAML
             return as_dispatch(t);
         }
 
+        template <typename T>
+        std::optional<T> try_as() const
+        {
+            Require(Type::Scalar);
```
JohnMcPMS, line 134: This will throw if the node isn't a scalar.  Should it just return an empty optional instead?

src/AppInstallerSharedLib/Yaml.cpp
```diff
@@ -207,22 +207,19 @@ namespace AppInstaller::YAML
...
             }
-            catch (...)
-            {
-            }
-
-            // Either 'true' or 'false'
-            // Avoid THROW_HR to don't log.
-            if (Utility::CaseInsensitiveEquals(m_scalar, "true") ||
-                Utility::CaseInsensitiveEquals(m_scalar, "false"))
+            else
```
JohnMcPMS, line 215: I would continue to use the early return pattern rather than an `else`.

src/AppInstallerSharedLib/Yaml.cpp
```diff
@@ -319,38 +316,84 @@ namespace AppInstaller::YAML
...
     }
 
-    bool Node::as_dispatch(bool*) const
+    std::optional<int> Node::try_as_dispatch(int*) const
     {
-        if (Utility::CaseInsensitiveEquals(m_scalar, "true"))
+        try
         {
-            return true;
+            return std::optional{ static_cast<int>(std::stoll(m_scalar, 0, 0)) };
```
JohnMcPMS, line 361:
```suggestion
            return std::optional{ static_cast<int>(std::stol(m_scalar, 0, 0)) };
```

src/Microsoft.Management.Configuration.Processor/Extensions/ValueSetExtensions.cs
```diff
@@ -50,14 +49,19 @@ public static Hashtable ToHashtable(this ValueSet valueSet)
...
+        /// <returns>Ordered list.</returns>
+        public static IList<object> ToArray(this ValueSet valueSet)
         {
             if (!valueSet.ContainsKey(TreatAsArray))
             {
                 throw new InvalidOperationException();
             }
 
-            var result = new List<object>();
+            var sortedList = new SortedList<string, object>();
```
JohnMcPMS, line 64: string sort won't work once we hit 11 items.

JohnMcPMS (changes requested):

JohnMcPMS (approved):

PR #3137, opened 2023-04-08
Use FolderId_ProgramFiles for non x86 default machine portable folder

yao-msft:
This change keeps existing behavior for x64 and arm64. On x86 machines, both point to C:\program files, and it's not breaking (actually the else part will not even be triggered).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3137)


Trenly:
Thanks for fixing this! Sorry for not considering the emulated case

msftrubengu (approved):

yao-msft:
After more investigation, I feel this fix is just hiding some potential failures. I need to trace back to see why [this code](https://github.com/microsoft/winget-cli/blob/850a59967394adbfb9d0d8819fa54c7852ea232a/src/AppInstallerCLICore/PortableInstaller.cpp#L45) assumes either x86 or x64.

if a portable package is arm64, it should not be put under Program Files(x64) on an arm64 machine

PR #3150, opened 2023-04-13
DRAFT PR: Microsoft.WinGet.Client

msftrubengu:
Creating a draft PR so run the build and see all the things I broke. 🫠


JohnMcPMS (commented):
This doesn't match the [PowerShell example](https://learn.microsoft.com/en-us/powershell/scripting/dev-cross-plat/resolving-dependency-conflicts?view=powershell-7.3) exactly.  They required that the engine also be in the `Dependencies` directory, and from my reading of the documentation that was a very intentional thing.  I get the sense that once an ALC loads a binary, [it gets the first shot at loading any dependencies](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=net-7.0#application-usage).  Only if it fails does the Default context try, then if it also fails the `Resolving` event is raised.

The point is that you only want the `Resolving` event to be used for a binary whose name is not going to collide with any other module.  The engine is the entry point for this.  Then afterward, the context will be used to load all dependencies of the engine (recursively) via the context's `Load`.

If `Resolving` is allowed to load shared dependencies directly, then it is possible that multiple modules can still collide as the call order will determine which one of them wins.

I don't know how hard it will be to get things to work with the engine as an ALC dependency, but I think it is necessary.  I would further suggest two directories: `DirectDependencies` and `SharedDependencies`.  The direct dependencies can be loaded via the `Resolving` event and all have a name that is very unique (maybe it is only the engine).  The shared dependencies can only be loaded via `Load` and contain anything that might result in a conflict.  This design is of course completely dependent on my understanding of ALCs, which admittedly was only gained today.

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Acl/ModuleInit.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Initialization class for this module.
+    /// </summary>
+    public class ModuleInit : IModuleAssemblyInitializer, IModuleAssemblyCleanup
+    {
+        /// <inheritdoc/>
+        public void OnImport()
+        {
+            AssemblyLoadContext.Default.Resolving += WinGetAssemblyLoadContext.ResolvingHandler;
```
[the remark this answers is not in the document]
JohnMcPMS: I think that my overall comment has more understanding than this one I made before reading the docs.

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Acl/WinGetAssemblyLoadContext.cs
```diff
@@ -18,9 +18,13 @@ namespace Microsoft.WinGet.Client.Acl
...
     internal class WinGetAssemblyLoadContext : AssemblyLoadContext
     {
-        private static readonly string ExpectedDependencyPath = Path.Combine(
+        private static readonly string SharedDependencyPath = Path.Combine(
             Path.GetDirectoryName(typeof(WinGetAssemblyLoadContext).Assembly.Location),
-            "Dependencies");
+            "SharedDependencies");
+
+        private static readonly string DirectDependencyPath = Path.Combine(
+            Path.GetDirectoryName(typeof(WinGetAssemblyLoadContext).Assembly.Location),
```
JohnMcPMS, line 26: It looks like you inverted the name suggestions that I used (completely, so functionally it should work).  To me, the "shared" dependencies are the ones that might conflict (and so need to load from the ALC `Load`) and the "direct" ones are the ones that are directly dependencies of the main DLL (and so need to use `Resolving`).  Feel free to use whatever names you want, other than reversing my suggestion which is confusing to me. 🫠

msftrubengu: 🫠

JohnMcPMS (approved):

PR #3152, opened 2023-04-14
Add configuration telemetry events

JohnMcPMS:
## Change
This change adds telemetry events from configuration actions, as well as more details to improve the error reporting (both in telemetry and to the user). The telemetry events can be disabled in code by the front end. For `winget configure`, the winget setting that controls telemetry is flowed through into the configuration code.

Two telemetry events are added by this change:
- An event with details on a failed attempt to execute a configuration unit. This will only be logged for failures of publicly available units. It contains the name of the unit and module, as well as the names of the top level settings provided (but *not* the values).
- An event with a summary of processing a configuration set, containing the overall result and error attribution, as well as the counts of configuration units, runs, and failures.

A new value was added to the result information to attribute any error to the appropriate portion of the overall system. The current partitioning of the blame is:

|Source|Description|
|---|---|
|None|The source is not known, or more likely, there was no failure.|
|Internal|The result came from inside the configuration system; this is likely a bug.|
|ConfigurationSet|The configuration set was ill formed. For instance, referencing a configuration unit that does not exist or a dependency that is not present.|
|UnitProcessing|The external module that processes the configuration unit generated the result.|
|SystemState|The system state is causing the error.|
|Precondition|The configuration unit was not run due to a precondition not being met. For example, when an assert in the configuration set is not in the desired state, all of the units with Apply intent will have this set.|

Additionally, another string was added to the result information. The intention is that the existing string (`Description`) should be used for a "short", user presentable message. The new string (`Details`), can contain a longer value that is intended for logs or a "more details" type experience.

The PowerShell processor was changed to wrap almost all exceptions into the result information object now that the source can be clearly stated there. Failures coming directly from invoking the resource are attributed to the configuration unit processing, except in the cases where we detect the signature of an invalid setting value (then the configuration set [author] is blamed). All other exceptions are treated as an internal error.

## Validation
- A mechanism for using the diagnostics to examine telemetry event data was created.
- Existing unit tests were updated to verify the error attribution and telemetry fields.
- New tests were added to verify the telemetry controls and fields.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3152)


src/Microsoft.Management.Configuration.Processor/Exceptions/InvokeDscResourceException.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// </summary>
+        /// <param name="method">Method.</param>
+        /// <param name="resourceName">Resource name.</param>
+        /// <param name="module">Optional module.</param>
+        /// <param name="message">Message.</param>
+        /// <param name="configurationSetSource">If true, the source of this error is set to be the configuration.</param>
+        public InvokeDscResourceException(string method, string resourceName, ModuleSpecification? module, string message, bool configurationSetSource = false)
+            : base(CreateMessage(method, resourceName, module, message))
+        {
+            // The property 'Fake' cannot be found on this object. Verify that the property exists and can be set.
```
msftrubengu, line 63: This comment seems like it should be in ContainsPropertyError

msftrubengu (approved):

PR #3157, opened 2023-04-15
Fix bugs with display paths

JohnMcPMS:
## Change
This change fixes two bugs with display paths:
1. The recent change to use case insensitive ICU for directory name equality inverted the comparison, making the replacement not happen correctly.
2. The recent change to enable display paths generically left them marked for creation, resulting in errors if directories that required administrator privileges to create did not exist and were requested from an unelevated context (ex. run `winget --info` from medium IL process when the portable machine links location(s) did not exist).

In addition, I tried to generally make `forDisplay` apply broadly since it is now allowed to be requested on any path.

## Validation
`winget --info` now contains the expected output instead of an error (2) or the full paths containing the user's account name (1).
A few unit tests that were being hit by the error are also working now.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3157)

### DIFF
--- a/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj.filters
+++ b/src/AppInstallerCLITests/AppInstallerCLITests.vcxproj.filters
@@ -227,9 +227,6 @@
     <ClCompile Include="Archive.cpp">
       <Filter>Source Files\Common</Filter>
     </ClCompile>
-    <ClCompile Include="Filesystem.cpp">
-      <Filter>Source Files</Filter>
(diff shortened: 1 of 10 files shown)


Trenly:
Thank you for fixing everything I broke

yao-msft (approved):

msftrubengu:
Blaming cosmic rays 👾

PR #3158, opened 2023-04-16
Improve configure user experience

JohnMcPMS:
## Change
This change improves the `configure` user experience by:
1. Ensuring that the configuration file exists before doing any other initialization. This speeds up the time to getting the error dramatically if the file does not exist.
2. Creating a better mechanism to report progress of async activities and using it to give the user information on what is happening during the beginning of the `configure` command.
3. Adding specific error strings for every known configuration error and error source, as well as generic fallbacks.
4. Outputting the result description in the event that it is supplied, and informing the user that there is additional information in the log file in case it is truncated or the `Details` property is not empty.

It also logs the entire configuration file input stream to the log in verbose, and improves the progress reporting for the pre-check of the apply flow.

The initialization phases are reported as:
- "Initializing configuration system" while creating the server process and factory objects.
- "Reading configuration file" while parsing the configuration file (this is usually fast enough that you won't see it).
- "Retrieving configuration details" while we wait for the detailed information to come in about each configuration unit. This is reported repeatedly any time we are still waiting on more details.

## Validation
Only minor manual validation has been done at the time of PR creation.
Automated testing should certainly be added, but we may want to forego that at this time given how poor the current UX is and the desire to get a build out for testing.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3158)


src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -435,20 +492,40 @@ namespace AppInstaller::CLI::Workflow
...
+                        AICLI_LOG_LARGE_STRING(Config, Error, << "Configuration unit " << Utility::ConvertToUTF8(unit.UnitName()) << "[" << Utility::ConvertToUTF8(unit.Identifier()) << "] failed with code 0x"
+                            << Logging::SetHRFormat << resultInformation.ResultCode() << " and error message:\n" << description, Utility::ConvertToUTF8(resultInformation.Details()));
+
+                        UnitFailedMessageData messageData = GetUnitFailedData(unit, resultInformation);
+                        auto error = m_context.Reporter.Error();
+                        error << "  "_liv << messageData.Message << std::endl;
+
+                        if (messageData.ShowDescription && !description.empty())
+                        {
+                            constexpr size_t maximumDescriptionLines = 3;
```
Trenly, line 506: Might be nice if users could configure this. Maybe just a TODO for now until we know if being able to configure would be useful

src/AppInstallerCLICore/Workflows/ConfigurationFlow.h
```diff
@@ -5,6 +5,12 @@
 
 namespace AppInstaller::CLI::Workflow
 {
+    // Fails if the configuration file path does not exist.
+    // Required Args: ConfigurationFile
+    // Inputs: None
+    // Outputs: None
+    void EnsureConfigurationFileExists(Execution::Context& context);
```
Trenly, line 12: This name of this almost makes me think that this could be abstracted into a generic check for files existing where an enum could provide the logic of which arg to grab the path from . `winget import` is one that comes to mind, install/uninstall/show with `--manifest`, `winget hash`, etc. I'm surprised this isn’t genericised already. . .

JohnMcPMS: It actually is, and I wrote it. I just didn't look hard enough to rediscover `VerifyFile`.

msftrubengu (approved):

Issue #3155, opened 2023-04-14
Implement "winget list --upgradable"


Hermholtz:
Hello. Bringing to your attention a suggestion that had 67 positive reactions already. Thanks in advance.

```powershell
winget list # lists apps installed by winget
winget list --upgradable # lists only installed apps with available updates
```

_Originally posted by @jantari in https://github.com/microsoft/winget-cli/issues/119#issuecomment-631029541_

jantari:
Thanks for tagging me, I just want to add that that comment of mine is 3 years old and by now we have the PowerShell cmdlets for winget.

Therefore I don't think the legacy DOS command and its syntax matter anymore as it is unusable anyway (due to truncating output etc.)

`winget list --upgradable` is unnecessary because you can just use `Get-WinGetPackage | Where-Object IsUpdateAvailable` instead and get actually usable data from it.

Hermholtz:
On my system there's no "Get-WinGetPackage" nor any other fancy PowerShell stuff.
There _is_ a way to display only upgradable packages by using `winget upgrade` which is totally conterintuitive. Therefore the code is done, please just add this one option `--upgradable` to use it.

PR #3162, opened 2023-04-17
Add argument to list only upgradeable

Trenly:
- [x] This pull request is related to an issue - [3155](https://github.com/microsoft/winget-cli/issues/3155)

-----

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -158,6 +158,10 @@ namespace AppInstaller::CLI
         case Execution::Args::Type::ListVersions:
             return { type, "versions"_liv };
 
+        // List command
+        case Execution::Args::Type::Upgrade:
+            return { type, "upgrade"_liv};
+
         // Pin command
         case Execution::Args::Type::GatedVersion:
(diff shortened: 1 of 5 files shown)


yao-msft:
Change looks good. @denelon to see if we have a better arg name than `--upgrade`

denelon:
I don't want to require tons of typing, but `winget list --upgrade` almost feels like a synnonym for `winget upgrade --all`.

I'm not sure if "--upgrades" is better or "--upgrade-available" is better.

yao-msft:
"--upgrade-available" seems more accurate

Trenly:
> "--upgrade-available" seems more accurate

Travelling at the moment; Can you add as a suggestion so I can quickly commit?

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -158,6 +158,10 @@ namespace AppInstaller::CLI
         case Execution::Args::Type::ListVersions:
             return { type, "versions"_liv };
 
+        // List command
+        case Execution::Args::Type::Upgrade:
+            return { type, "upgrade"_liv};
```
yao-msft, line 163:
```suggestion
            return { type, "upgrade-available"_liv};

```

Issue #3141, opened 2023-04-10
[Pinning] - Usage of `--id` should bypass installed package check


Trenly:
### Description of the new feature / enhancement

The current implementation of pinning checks to try and map to an installed package before adding the pin. Since the pin only requires the package ID and source list, when `--id` is provided, there should be no need to attempt the mapping. The given ID can be added to the pinning database directly.

This will help when there are side-by-side installs or matching issues which may prevent pinning due to multiple packages found.

### Proposed technical implementation details

This behavior probably should not be default, but enabled through settings.

```json
"pinning" : {
    "directIdAddition" : true
}
```

florelis:
@Trenly Do you think this would still be needed after #3167?

Trenly:
> @Trenly Do you think this would still be needed after #3167?

Yes, because #3167 still fails when trying to pin Microsoft.Teams due to multiple strong matches when the Teams Machine Wide Installer is present

florelis:
I ended up including this in #3167, so I'm closing the issue. Feel free to re-open if I missed anything.

PR #3167, opened 2023-04-18
Record Product Codes in pinning table

### LINKED ISSUE #2974: `winget pin list` does not list all pins

Not sure if this is the same root cause as #2973 but I'm filing it separately just in case.
I have 3 pins configured, but only 2 are shown.

### Steps to reproduce

Add some pins, then list them

### Expected behavior

All pins to be listed

### Actual behavior

[Pin List - WinGet-2023-02-15-20-50-07.616.log](https://github.com/microsoft/winget-cli/files/10751697/WinGet-2023-02-15-20-50-07.616.log)

image

### Environment

```shell
PS C:\WINDOWS\system32> winget --info
Windows Package Manager (Preview) v1.5.441-preview

System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.441.0

Links

```

### LINKED ISSUE #2977: ID of added pin inconsistent with output

* Related to #2976

When I ran `winget pin CCleaner` on a new machine, it indicated that a pin was added for the ID `Piriform.CCleaner`. However, the pin was actually added for `Piriform.CCleaner.ProTrial`

### Steps to reproduce

1) `winget install --id Piriform.CCleaner -e`
2) `winget pin add CCleaner`
3) `winget pin list`

### Expected behavior

Pin added for `Piriform.CCleaner` as shown in the output of `Found CCleaner [Piriform.CCleaner]`

### Actual behavior

[Pin add #1 - WinGet-2023-02-15-21-44-39.723.log](https://github.com/microsoft/winget-cli/files/10752005/WinGet-2023-02-15-21-44-39.723.log)
[Pin add #2 - WinGet-2023-02-15-21-44-46.457.log](https://github.com/microsoft/winget-cli/files/10752006/WinGet-2023-02-15-21-44-46.457.log)
[Pin list - WinGet-2023-02-15-21-45-03.796.log](https://github.com/microsoft/winget-cli/files/10752008/WinGet-2023-02-15-21-45-03.796.log)

```
PS C:\WINDOWS\system32> winget pin add CCleaner
Found CCleaner [Piriform.CCleaner]
Pin added successfully
PS C:\WINDOWS\system32> winget pin add CCleaner
Found CCleaner [Piriform.CCleaner]
There is already a pin for package CCleaner Professional Trial
PS C:\WINDOWS\system32> winget pin list
Id                         Source Version
------------------------------------------
Microsoft.Office           winget Blocking
Microsoft.WindowsSDK       winget Blocking
Piriform.CCleaner.ProTrial winget Pinning
```

image

### Environment

```shell
PS C:\WINDOWS\system32> winget --info
Windows Package Manager (Preview) v1.5.441-preview

System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.441.0

Links

```

### LINKED ISSUE #3142: Handle `null` as version in pinning.db

After manually editing `pinning.db` using DB Browser for SQLite, winget returns null when running upgrade / list commands.

After further investigation, this is because the `version` field for one entry was `null` and not empty

### Steps to reproduce

Manually add an entry to the pinning database

image

### Expected behavior

Null to be handled seamlessly

### Actual behavior

Winget ceases to provide any package mapping during list / upgrade

### Environment

```shell
PS D:\Git\winget-pkgs> winget --info
Windows Package Manager (Preview) v1.5.441-preview

Windows: Windows.Desktop v10.0.19044.2728
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.441.0

Links

LocalArchiveMalwareScanOverride           Enabled
```

florelis:
This PR changes the pinning table to record an installed pinned app's Product Code or Package Family Name. This allows us to distinguish between multiple installed apps that we want to pin and have the same package ID, in particular the Windows SDK.

This includes a breaking change to the schema of the pinning DB. The pinning index now uses v1.1, but it is not backwards compatible with v1.0. If, when opening an existing index, we see that it has version 1.0, we delete it and create a new one with the new version. This will drop all existing pins; requiring them to be re-added.I could try to recover the pins, but I think it is better to re-add them to get the new fields I'm adding in this PR.

I'm adding the ProductCode or PackageFamilyName as an extra identifier in the PinKey. If an installed package has multiple, we create multiple pins.

Closes #2974
Closes #3142, as the new DB schema won't even allow null for the version field
Edit: Seems to also fix #2977

Edit: Updated note about deleting old DB

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3167)


Trenly:
> This includes a breaking change to the schema of the pinning DB. Opening an existing pinning DB will fail, and one has to manually delete the DB file located at %localappdata%\Packages\<dev or release package name>\LocalState\pinning.db. Since pinning has only been in preview; I consider this acceptable instead of creating a new DB schema version.

I think that it's acceptable to not create a new DB schema version. Like you said, it's still an experimental feature. That said, it is annoying to have to go in and manually delete a file, and I'm sure there will be a few issues opened about it and closed after we link back to this PR. @denelon may have a rough estimate of how many users are using the pin command?

To make it a bit more seamless, perhaps it may be beneficial to add a check that the pinning table matches at least one of the valid schemas. If it doesn't match any known schema, delete it and create a new one. Or perhaps make a method to restore the table to match the schema using the SQLite `AlterTable` functions to add any columns that are missing, reset the data type on existing columns, and drop any columns not present in the schema. 

The "delete" or "repair" functionality could potentially be useful in the future as the installed.db is leveraged more and as new functionality is added.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -16,21 +16,47 @@ namespace AppInstaller::CLI::Workflow
...
+            {
+            case Pinning::ExtraIdStringType::ProductCode:
+                searchRequest.Filters.emplace_back(PackageMatchField::ProductCode, MatchType::CaseInsensitive, pinKey.ExtraId);
+                break;
+            case Pinning::ExtraIdStringType::PackageFamilyName:
+                searchRequest.Filters.emplace_back(PackageMatchField::PackageFamilyName, MatchType::CaseInsensitive, pinKey.ExtraId);
+                break;
+            case Pinning::ExtraIdStringType::None:
+            default:
+                searchRequest.Filters.emplace_back(PackageMatchField::Id, MatchType::CaseInsensitive, pinKey.PackageId);
```
Trenly, line 54: Is there a reason the ID isn't always added as a filter? I'm thinking of the scenario where an organization has their own source which a user has added. The same application may be available from different sources, with different ID's, a user may want updates from one source but pin the ID from another source. On the other hand, that would require them to add two separate pins - one for each ID.

Unless you think that this case is already covered by `winget upgrade <query> -s <source> --include-pinned`

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -74,7 +91,7 @@ namespace AppInstaller::Repository::Microsoft
     PinningIndex::IdType PinningIndex::AddPin(const Pinning::Pin& pin)
     {
         std::lock_guard<std::mutex> lockInterface{ *m_interfaceLock };
-        AICLI_LOG(Repo, Verbose, << "Adding Pin for package [" << pin.GetPackageId() << "] from source [" << pin.GetSourceId() << "] with pin type " << Pinning::ToString(pin.GetType()));
+        AICLI_LOG(Repo, Verbose, << "Adding Pin for " << GetPinKeyLogString(pin.GetKey()) << " with pin type " << Pinning::ToString(pin.GetType()));
```
Trenly, line 94: If I'm not mistaken, the only time that the PinKey or the PinType need to be converted to a string is for logging. I think this may be better if you overload `Pinning::ToString` with `Pinning::ToString(PinKey)` or maybe even `Pinning::ToString(Pin)`.

You could even simplify the `with pin type` to be part of the string as `Type=[ ]` just to keep it all the same format

src/AppInstallerCommonCore/Public/winget/Pin.h
```diff
@@ -6,63 +6,116 @@
...
+                return SourceId < other.SourceId;
+            }
+            else if (ExtraIdType != other.ExtraIdType)
+            {
+                return ExtraIdType < other.ExtraIdType;
+            }
+            else
+            {
+                return ExtraId < other.ExtraId;
+            }
```
Trenly, line 85:
```suggestion
            // std::tie implements tuple comparison, wherein it checks the first item in the tuple,
            // iff the first elements are equal, then the second element is used for comparison, and so on
            return std::tie(PackageId, SourceId, ExtraIdType, ExtraId) < 
            	std::tie(other.PackageId, other.SourceId, other.ExtraIdType, other.ExtraId);
```
https://bajamircea.github.io/coding/cpp/2017/03/10/std-tie.html
https://en.cppreference.com/w/cpp/utility/tuple/operator_cmp

src/AppInstallerCLICore/Workflows/PinFlow.cpp
florelis: > Is there a reason the ID isn't always added as a filter?

That's what I did at first but I had some issues with it for the WindowsSDK. IIRC, when I used the ID, it would match the available package, which we would then try to correlate to the installed packages, find multiple and not know which one to pick. But using only the product code, we can first match the installed package, then correlate it back to the available package without conflicts.

And for the case you mention, I think it can be handled with `winget pin add <query> -s <source>`

Trenly: Makes sense!

florelis:
> > This includes a breaking change to the schema of the pinning DB. Opening an existing pinning DB will fail, and one has to manually delete the DB file located at %localappdata%\Packages<dev or release package name>\LocalState\pinning.db. Since pinning has only been in preview; I consider this acceptable instead of creating a new DB schema version.
> 
> I think that it's acceptable to not create a new DB schema version. Like you said, it's still an experimental feature. That said, it is annoying to have to go in and manually delete a file, and I'm sure there will be a few issues opened about it and closed after we link back to this PR. @denelon Demitrius Nelon FTE may have a rough estimate of how many users are using the pin command?
> 
> To make it a bit more seamless, perhaps it may be beneficial to add a check that the pinning table matches at least one of the valid schemas. If it doesn't match any known schema, delete it and create a new one. Or perhaps make a method to restore the table to match the schema using the SQLite `AlterTable` functions to add any columns that are missing, reset the data type on existing columns, and drop any columns not present in the schema.
> 
> The "delete" or "repair" functionality could potentially be useful in the future as the installed.db is leveraged more and as new functionality is added.

I changed it to actually use a new schema version. The pinning index now uses v1.1, but it is not backwards compatible with v1.0. (Now that I'm typing it, I realize that it doesn't respect sem-ver...). If, when opening an existing index, we see that it has version 1.0, we delete it and create a new one with the new version.

This would no longer require manual deletion of the index, but we're still dropping all existing pins. I could try to recover the pins, but I think it is better to re-add them to get the new fields I'm adding in this PR.

Trenly:
> I changed it to actually use a new schema version. The pinning index now uses v1.1, but it is not backwards compatible with v1.0. (Now that I'm typing it, I realize that it doesn't respect sem-ver...). If, when opening an existing index, we see that it has version 1.0, we delete it and create a new one with the new version.
> 
> This would no longer require manual deletion of the index, but we're still dropping all existing pins. I could try to recover the pins, but I think it is better to re-add them to get the new fields I'm adding in this PR.

I think dropping them all is fine since its experimental anyways. Thanks for at least removing the manual step!!

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -57,25 +83,32 @@ namespace AppInstaller::CLI::Workflow
...
+            const auto availablePackageId = availableVersion->GetProperty(PackageVersionProperty::Id).get();
+            const auto availableSourceId = availableVersion->GetProperty(PackageVersionProperty::SourceIdentifier).get();
+            const auto pinKeys = Pinning::GetPinKeysForAvailablePackage(availablePackageId, availableSourceId, installedProductCodes, installedPackageFamilyNames);
 
-            if (sources.insert(pinKey.SourceId).second)
+            for (const auto& pinKey : pinKeys)
             {
-                auto pin = pinningIndex->GetPin(pinKey);
-                if (pin)
+                if (sourcesAndIds.emplace(pinKey.SourceId, pinKey.ExtraId).second)
```
yao-msft, line 105: If the purpose is to avoid duplicate pins, should we just make pins an std::set and not worry about 2 specific fields in pinkey?

yao-msft:
The initial thought for pinning a specific installed version is to add a pin from installed source to the pinning table so the previous pinning table will not break.

PackageId                              SourceId
PFN or Product code             *installed

But if this approach works, it's ok too since this is still in experimental (and seems easier in implementation?)

florelis:
After talking offline with @yao-msft I realize that this change is breaking the basic use case of "I update this app outside of winget, so don't touch it" because once there is an update, the product code will change and the pin will no longer apply.

The case I was thinking for this PR was the one we see with Windows SDK, where multiple installed packages match with the same available package, so we need to pin the specific installed version, which is done with the product code.

Taking into account both cases makes it weird. For upgrades outside of winget, we have to pin just by package ID & source to be able to track across versions. But for packages where matching doesn't quite work or have multiple installed packages, we have to pin by product code. I'm thinking the solution will have to be guessing the intent at the time of adding a pin and either add it by ID or by product code. 

I like @yao-msft 's idea for the DB so we don't have to change the schema, so I'll change it to that. That way we don't have to deal with deleting old DBs. It also makes sense because for specific installed packages the source doesn't matter; there could be multiple sources for it but the moment one of them updates the package, the pin stops applying to all of them.

florelis:
I've updated the PR to follow the schema suggested by @yao-msft for representing the pins in the DB. Given how different it all is I could probably have made a new PR but I want to keep the discussion...

I'm introducing the concept of _pinning an available package_ or _pinning an installed package_
* Pinning an available package is done through its package ID and source. When we pin an available package, we can preserve the pin if there are updates outside of winget. The drawback is that we can't add a pin when there are multiple installed apps matching the available package (like Windows SDK). This is because when looking for the package to pin we can't find a single match. This would probably be fixed by implementing #3141
* Pinning an installed package is done through its product code or package family name. Doing this allows us to distinguish between multiple installed apps for the same available package. The drawback is that if the app updates outside of winget, we most likely would lose the pin as the product code would change. There's not much we can do about that because we wouldn't be able to tell if the update was on the pinned or on the other side-by-side install

By default pinning works on available. To pin a specific installed app I'm adding a --installed flag. Would be great if we can just guess the intent and avoid the arg, but I could not find an easy way to do it

Pins for an installed package are represented in the DB by using the ProductCode or PFN as package ID and *PredefinedInstalledSource for the source ID (this could be any string, doesn't actually have to match the real installed source ID)

Most of the work is done in the CompositeSource. When creating a composite package, we query the pinning index for pins on the installed and available packages that make it up. Then when querying the composite package (e.g. for IsUpdateAvailable) we just reason about what the pin state would be for each version.

Trenly:
>  Would be great if we can just guess the intent and avoid the arg,

Perhaps if pin by available fails (due to multiple packages, as one scenario) then pinning by installed could be a fallback?

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -185,9 +199,22 @@ namespace AppInstaller::CLI
...
+                Workflow::OpenSource();
+        }
+        else
+        {
+            // If not working from just ID, try matching a single installed package
+            context <<
+                Workflow::OpenSource() <<
+                Workflow::OpenCompositeSource(Repository::PredefinedSource::Installed);
+        }
+
```
Trenly, line 216: This section is repeated between PinAdd and PinRemove. Perhaps it would make sense to have a workflow task like `OpenSourceForPinning`, or even abstract it to something like `OpenSourceFor( OperationType::Pin)` so that the source opening for list/update/show/install could be brought back into a single workflow task.

src/AppInstallerRepositoryCore/Microsoft/Schema/Version.cpp
```diff
@@ -26,6 +26,13 @@ namespace AppInstaller::Repository::Microsoft::Schema
 
     std::ostream& operator<<(std::ostream& out, const Version& version)
     {
-        return (out << version.MajorVersion << '.' << version.MinorVersion);
+        if (version == Version::Latest())
```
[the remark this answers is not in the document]
florelis: Added

src/AppInstallerCommonCore/Public/winget/Pin.h
```diff
@@ -6,63 +6,110 @@
...
     };
 
     std::string_view ToString(PinType type);
     PinType ConvertToPinTypeEnum(std::string_view in);
 
-    // The set of values needed to uniquely identify a Pin
+    std::string_view ToString(ExtraIdStringType type);
+
+    // Determines which of two pin types is more strict.
+    PinType StrictestPinType(PinType first, PinType second);
```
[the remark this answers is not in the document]
florelis: Changed to IsStricter

src/AppInstallerCommonCore/Public/winget/Pin.h
```diff
@@ -6,63 +6,110 @@
...
+    // The set of values needed to uniquely identify a Pin.
+    // We use both a PackageId and a ProductCode or PackageFamilyName
+    // to be able to deal with packages that match with multiple installed apps.
+    // This helps us deal better with apps that we don't match properly elsewhere.
     struct PinKey
     {
         PinKey() {}
-        PinKey(const Manifest::Manifest::string_t& packageId, std::string_view sourceId)
-            : PackageId(packageId), SourceId(sourceId) {}
+        PinKey(const Manifest::Manifest::string_t& packageId, std::string_view sourceId, ExtraIdStringType extraIdType, std::string_view extraId = {})
```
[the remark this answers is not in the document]
florelis: Outdated; no longer applies with latest changes

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -156,7 +169,7 @@ namespace AppInstaller::Repository::Microsoft
             m_version.MajorVersion == 1 ||
```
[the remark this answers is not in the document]
florelis: Yeah, it should have been {1,1}, but I reverted it back to staying on 1.0

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -22,44 +22,84 @@ namespace AppInstaller::Repository
...
+            PinBehavior pinBehavior)
+        {
+            if (pinBehavior == PinBehavior::IgnorePins)
+            {
+                // No need to check anything
+                return Pinning::PinType::Unknown;
+            }
+
+            // For the pin in the available version, we can ignore it depending on the behavior and type.
+            // If it is gating, we don't need to check the version as that was already done.
```
yao-msft, line 48: Do you feel we should centralize the logic to do the gated version check here so this will be the one and only logic of "what is the pinned state of the version" and everywhere else should just call this function?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -741,14 +812,21 @@ namespace AppInstaller::Repository
                 }
             }
 
-            std::shared_ptr<IPackage> m_installedPackage;
+            std::optional<Pinning::Pin> GetInstalledPin() const
+            {
+                return (ExperimentalFeature::IsEnabled(ExperimentalFeature::Feature::Pinning) && m_installedPackage)
+                    ? m_installedPackage->GetPin()
+                    : std::optional<Pinning::Pin>{};
```
yao-msft, line 819: nit: std::nullopt  or  {}

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -634,16 +682,24 @@ namespace AppInstaller::Repository
...
                     static_cast<bool>(m_installedPackage) != static_cast<bool>(otherComposite->m_installedPackage) ||
-                    (m_installedPackage && !m_installedPackage->IsSame(otherComposite->m_installedPackage.get())) ||
                     m_availablePackages.size() != otherComposite->m_availablePackages.size())
                 {
                     return false;
                 }
 
+                if (m_installedPackage)
+                {
+                    if (m_installedPackage->GetSourceId() != otherComposite->m_installedPackage->GetSourceId() ||
```
yao-msft, line 692: Looks like it will cause a crash if otherComposite->m_installedPackage  is nullptr?

Also nit: the 2 if statements could be combined

yao-msft: nvm, I saw the `static_cast<bool>(m_installedPackage) != static_cast<bool>(otherComposite->m_installedPackage)` which ensures that.

But I think below `m_installedPackage->GetPackage()->IsSame(otherComposite->m_installedPackage->GetPackage().get())` already checked the sourceId is same, does not seem necessary to check explicitly here

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -701,26 +764,34 @@ namespace AppInstaller::Repository
...
 
             // Gets the information about the pins that exist for this package
-            void GetExistingPins(PinningIndex& pinningIndex, bool cleanUpStalePins)
+            void GetExistingPins(PinningIndex& pinningIndex)
             {
-                // If the package is installed, we need to add the pin information to the available packages from any source.
-                // If the package is not installed, we clean up stale pin information here.
+                if (!m_installedPackage)
+                {
+                    // Only installed packages have pins
```
yao-msft, line 771: Does not seem the case any more if pin by id where we do not ensure an installed version found

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_0/PinTable.cpp
```diff
@@ -51,14 +55,15 @@ namespace AppInstaller::Repository::Microsoft::Schema::Pinning_V1_0
         createTableBuilder.Column(ColumnBuilder(s_PinTable_PackageId_Column, Type::Text).NotNull());
         createTableBuilder.Column(ColumnBuilder(s_PinTable_SourceId_Column, Type::Text).NotNull());
         createTableBuilder.Column(ColumnBuilder(s_PinTable_Type_Column, Type::Int64).NotNull());
-        createTableBuilder.Column(ColumnBuilder(s_PinTable_Version_Column, Type::Text));
+        createTableBuilder.Column(ColumnBuilder(s_PinTable_Version_Column, Type::Text).NotNull());
```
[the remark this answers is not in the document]
florelis: I think we will just fail and not get pins if there is a null, which is the same as we currently do. The only way we can get a null is if someone goes to manually edit the database. I think the only thing this would do is prevent that if their db is created with the new code.

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -22,44 +22,84 @@ namespace AppInstaller::Repository
...
             };
         }
 
-        std::optional<PackageVersionKey> GetLatestAvailableVersionKeySatisfyingPin(const std::vector<PackageVersionKey>& availableVersionKeys, PinBehavior pinBehavior)
+        // Gets the pinned state for an available PackageVersionKey that may have a pin,
+        // and optionally an additional pin that could come from the installed version.
+        // If there are both installed and available pins, we return the one that is the most strict.
+        Pinning::PinType GetPinnedStateForVersion(
+            const PackageVersionKey& availableVersionKey,
+            const std::optional<Pinning::Pin>& pin,
```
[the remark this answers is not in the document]
florelis: It actually can be the available pin before we set the pinned state in the version key. I'll update the comment instead

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -113,16 +114,28 @@ namespace AppInstaller::CLI
...
         context <<
-            Workflow::OpenSource() <<
-            Workflow::OpenCompositeSource(Repository::PredefinedSource::Installed) <<
             Workflow::SearchSourceForSingle <<
             Workflow::HandleSearchResultFailures <<
             Workflow::EnsureOneMatchFromSearchResult(OperationType::Pin) <<
             Workflow::GetInstalledPackageVersion <<
             Workflow::ReportPackageIdentity <<
             Workflow::OpenPinningIndex() <<
-            Workflow::SearchPin <<
```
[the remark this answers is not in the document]
florelis: Yes, it wasn't used. I think I added it thinking of using it to check if it already existed before adding it, but I ended up doing that during the actual add so it wasn't needed.

src/AppInstallerRepositoryCore/CompositeSource.cpp
florelis: I think that's what I did? I had some duplication before but I thought I had moved it all here. Can't find another use of the `GetGatedVersion()`

src/AppInstallerRepositoryCore/CompositeSource.cpp
florelis: Updated the comment. A package that is not installed can have pins, but we don't care about them. For pin operations in those cases we bypass the composite source, and pins don't affect packages that are not installed

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -210,29 +229,59 @@ namespace AppInstaller::CLI::Workflow
...
+                    {
+                        packageName = availableVersion->GetProperty(PackageVersionProperty::Name);
+                        sourceName = availableVersion->GetProperty(PackageVersionProperty::SourceName);
+                    }
+                }
+
+                auto installedVersion = match.Package->GetInstalledVersion();
+                if (installedVersion)
+                {
+                    packageName = installedVersion->GetProperty(PackageVersionProperty::Name);
```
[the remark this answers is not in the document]
florelis: I think if we have it installed it's better to show that name instead of the available. That way it matches with what we show in list/upgrade commands. It also makes the differences clearer if a single available pin matches multiple installed packages.

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -113,16 +114,28 @@ namespace AppInstaller::CLI
 
     void PinAddCommand::ExecuteInternal(Execution::Context& context) const
     {
+        if (context.Args.Contains(Execution::Args::Type::Id))
+        {
+            // When we are given an ID, just pin that available package without checking for installed.
+            // This helps when there are matching issues, for example due to multiple side-by-side installs.
```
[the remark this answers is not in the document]
florelis: > So previously, our goal for pinning available package is: if we find one available package to pin, all other available packages from other sources are pinned, except if user provides -s <source>
>
> Do you think we should do this for --id pins as well?

I hadn't thought about that. With the code as it is, it will add pins to all sources matching the query (--id or otherwise). So if you do `pin add Microsoft.PowerToys` and multiple sources have `Microsoft.PowerToys` we'd be adding pins to all of them, which seems okay to me. But if the ID in one of the sources is just "PowerToys" we wouldn't pin it. That also seems okay to me because we can't guess what queries we'd have to make to match the package from another source as it can be totally different. And this applies whether we are using --id or some other filter

src/AppInstallerCLITests/PinningIndex.cpp
```diff
@@ -99,8 +98,8 @@ TEST_CASE("PinningIndex_AddUpdateRemove", "[pinningIndex]")
     TempFile tempFile{ "repolibtest_tempdb"s, ".db"s };
     INFO("Using temporary file named: " << tempFile.GetPath());
 
-    Pin pin = Pin::CreateGatingPin({ "pkgId", "srcId" }, { "1.0.*"sv });
-    Pin updatedPin = Pin::CreatePinningPin({ "pkgId", "srcId" });
+    Pin pin = Pin::CreateGatingPin({ "pkgId", "srcId"}, { "1.0.*"sv });
+    Pin updatedPin = Pin::CreatePinningPin({ "pkgId", "srcId"});
```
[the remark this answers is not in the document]
florelis: Yeah, botched search-and-replace from when I was removing the extra args I added. I added back the spaces, and also added them in some other places where they were missing.

src/AppInstallerCLICore/Commands/PinCommand.cpp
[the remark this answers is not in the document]
yao-msft: We could construct a query for systemReferenceStrings to find available packages from other sources. We can do it later if such asks emerges. But the behavior is different from what I specc'ed before.

src/AppInstallerRepositoryCore/CompositeSource.cpp
yao-msft: I feel we could just remove this check and always add pin info to the composite package, for correctness

yao-msft (dismissed):
:shipit:

src/AppInstallerRepositoryCore/CompositeSource.cpp
florelis: Removed

yao-msft (approved):

PR #3172, opened 2023-04-20
Make entitlement best effort

### LINKED ISSUE #2052: Install from msstore error

Install from msstore returns error.
```
Verifying/Requesting package acquisition...
Verifying/Requesting package acquisition failed: server error'
```

### Steps to reproduce

1. `winget install xxxxxx`, change to any msstore package hash (e.g., WA Beta: 9nbdxk71nk08)

### Expected behavior

Package installed

### Actual behavior

Packaged failed to install.
```
Verifying/Requesting package acquisition...
Verifying/Requesting package acquisition failed: server error
```

### Environment

```shell
Windows: Windows.Desktop v10.0.22000.593
Package: Microsoft.DesktopAppInstaller v1.17.10271.0
```

JohnMcPMS:
Fixes #2052 

## Change
Make entitlement acquisition best effort rather than required. A recent change caused some issues for us regarding the entitlement process, and since the entitlement isn't actually required to install, just attempt to get it and log if it failed.  Even if the underlying issue is resolved, this change will prevent future issues from having the same level of impact.

This may lead to an issue where the entitlement acquisition fails, the package is installed, but attempting to run it fails because the entitlement cannot be acquired then either (by the OS in the background).  Given the choice between the two failure modes though, it seems far better to remove these false negatives and create the (my gut says fewer) false positives that will result.

## Validation
Manually verified that I can get an error from entitlement acquisition and still have the install succeed (at least for the re-acquisition case).  Also verified independently that not getting an entitlement, installing, and then launching works.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3172)

### DIFF
--- a/src/AppInstallerCLICore/Resources.h
+++ b/src/AppInstallerCLICore/Resources.h
@@ -264,10 +264,6 @@ namespace AppInstaller::CLI::Resource
         WINGET_DEFINE_RESOURCE_STRINGID(MsixArgumentDescription);
         WINGET_DEFINE_RESOURCE_STRINGID(MsixSignatureHashFailed);
         WINGET_DEFINE_RESOURCE_STRINGID(MSStoreAppBlocked);
-        WINGET_DEFINE_RESOURCE_STRINGID(MSStoreInstallGetEntitlementNetworkError);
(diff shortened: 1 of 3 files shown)


yao-msft (approved):
:shipit:

PR #3180, opened 2023-04-24
Repair-WinGetPackage download VCLibs if needed

msftrubengu:
`Add-AppxPackage -Path <vclibs urls>` fails in Windows Sandbox. This change detects it and download the package and installs is from a file path. 

Add more debug information to Repair and Assert cmdlets.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3180)

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.Client.Engine/Common/WinGetIntegrity.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Engine/Common/WinGetIntegrity.cs
@@ -48,17 +48,17 @@ namespace Microsoft.WinGet.Client.Engine.Common
             }
             catch (Win32Exception e)
             {
-                psCmdlet.WriteDebug($"'winget.exe --version' Win32Exception {e}");
+                psCmdlet.WriteDebug($"'winget.exe' Win32Exception {e.Message}");
                 throw new WinGetIntegrityException(GetReason(psCmdlet));
             }
             catch (Exception e) when (e is WinGetCLIException || e is WinGetCLITimeoutException)
             {
-                psCmdlet.WriteDebug($"'winget.exe --version' WinGetCLIException {e}");
+                psCmdlet.WriteDebug($"'winget.exe' WinGetCLIException {e.Message}");
                 throw new WinGetIntegrityException(IntegrityCategory.Failure, e);
             }
             catch (Exception e)
             {
-                psCmdlet.WriteDebug($"'winget.exe --version' Exception {e}");
+                psCmdlet.WriteDebug($"'winget.exe' Exception {e.Message}");
                 throw new WinGetIntegrityException(IntegrityCategory.Unknown, e);
             }
 
--- a/src/PowerShell/Microsoft.WinGet.Client.Engine/Common/WinGetIntegrity.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Engine/Common/WinGetIntegrity.cs
@@ -46,16 +46,19 @@ namespace Microsoft.WinGet.Client.Engine.Common
                 var result = wingetCliWrapper.RunCommand("--version");
                 result.VerifyExitCode();
             }
-            catch (Win32Exception)
+            catch (Win32Exception e)
             {
+                psCmdlet.WriteDebug($"'winget.exe --version' Win32Exception {e}");
                 throw new WinGetIntegrityException(GetReason(psCmdlet));
             }
             catch (Exception e) when (e is WinGetCLIException || e is WinGetCLITimeoutException)
             {
+                psCmdlet.WriteDebug($"'winget.exe --version' WinGetCLIException {e}");
                 throw new WinGetIntegrityException(IntegrityCategory.Failure, e);
             }
             catch (Exception e)
             {
+                psCmdlet.WriteDebug($"'winget.exe --version' Exception {e}");
                 throw new WinGetIntegrityException(IntegrityCategory.Unknown, e);
             }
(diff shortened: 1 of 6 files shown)


src/PowerShell/Microsoft.WinGet.Client.Engine/Common/WinGetIntegrity.cs
```diff
@@ -46,16 +46,19 @@ public static void AssertWinGet(PSCmdlet psCmdlet, string expectedVersion)
                 var result = wingetCliWrapper.RunCommand("--version");
                 result.VerifyExitCode();
             }
-            catch (Win32Exception)
+            catch (Win32Exception e)
             {
+                psCmdlet.WriteDebug($"'winget.exe --version' Win32Exception {e}");
```
ryfu-msft, line 51: NIT: I think it would be better to just say `Failed to call winget.exe with Win32Exception {e}`. I don't think the --version is important and might be confusing.

ryfu-msft: Same on line 56 and 61

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/AppxModuleHelper.cs
```diff
@@ -210,5 +215,64 @@ private void InstallUiXaml()
...
+
+        private void AddAppxPackageAsUri(string packageUri)
+        {
+            var errorRecord = this.RunScript(string.Format(AddAppxPackageFormat, packageUri));
+
+            if (errorRecord != null)
+            {
+                if (errorRecord.CategoryInfo.Category == ErrorCategory.OpenError)
+                {
+                    this.psCmdlet.WriteDebug($"Failed adding package {packageUri}. Retrying downloading it.");
```
ryfu-msft, line 227: NIT: `Failed to add package with uri [{packageUri}]. Retry downloading the package and adding it.`

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/TempFile.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <summary>
+        /// Gets the file name.
+        /// </summary>
+        public string FileName { get; }
+
+        /// <summary>
+        /// Gets the full file name.
+        /// </summary>
+        public string FullFileName { get; }
```
ryfu-msft, line 66: NIT: I think `FullPath` would be a better name for this property

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/TempFile.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <summary>
+        /// IDisposable.Dispose .
```
ryfu-msft, line 69: Nit: Extra space at end before final period

ryfu-msft (approved):

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/AppxModuleHelper.cs
```diff
@@ -155,11 +158,14 @@ private PSObject GetAppxObject(string packageName)
         private IReadOnlyList<string> GetVCLibsDependencies()
         {
             var vcLibsDependencies = new List<string>();
-            var vcLibsPackageObjs = this.psCmdlet.InvokeCommand
-                .InvokeScript(string.Format(GetAppxPackageByVersionCommand, VCLibsUWPDesktop, VCLibsUWPDesktopVersion));
+
+            var cmd = string.Format(GetAppxPackageByVersionCommand, VCLibsUWPDesktop, VCLibsUWPDesktopVersion);
+            this.psCmdlet.WriteDebug($"Running command: '{cmd}'");
+            var vcLibsPackageObjs = this.psCmdlet.InvokeCommand.InvokeScript(cmd);
```
Trenly, line 164: Invoking scripts built by strings always makes me a bit nervous. Are there checks in place to be sure no malicious commands can be injected into the string?

florelis: > Invoking scripts built by strings always makes me a bit nervous

It's also against internal security recommendations to use scripts built by strings. I think they suggest you use ScriptBlocks

JohnMcPMS: Even though these are all constant strings, I agree that always using a ScriptBlock is the right thing to do to prevent any accidental injection attack possibilities.  It seems like it would be fairly straightforward to make wrapper functions for anything like this that contains the ScriptBlock and parameter addition to whatever `Invoke` function is for a ScriptBlock.

msftrubengu: The main reason I added InvokeScript is to avoid using the PowerShell obj and have to import Appx for PowerShell 7 every time. If there's no explicit import any cmdlet that uses this class will fail for pre 10.0.22453.0 builds. So I'm going back to use PowerShell, import Appx before any cmdlet in 7 and hide the warnings

JohnMcPMS (approved):

PR #3186, opened 2023-04-25
Enable module/resource syntax and improve file error reporting

JohnMcPMS:
## Change
This change adds a new syntax that can be used when referencing resources in a configuration file: `Module/Resource`. This is only added to a new 0.2 schema version, mainly to prevent an issue with older clients failing to run the newer files in a way that is less obvious than "unrecognized configuration file version".

As part of adding a new schema version, I also added the schema version as a property to the `ConfigurationSet` and `ConfigurationUnit`. This will enable the processor to handle any semantic differences that might arise as schemas evolve.

Finally, the result from opening a configuration file now contains the field name and value, as well as the line and column number for the error (all values supplied as appropriate).  All of this information is better presented to the user from the family of `winget configure` commands.

## Validation
New tests are added to validate the result of opening a configuration file.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3186)


src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
```diff
@@ -454,6 +462,15 @@ namespace Microsoft.Management.Configuration
...
         // The field that is missing/invalid, if appropriate for the specific ResultCode.
         String Field{ get; };
+
+        // The value of the field, if appropriate for the specific ResultCode.
+        String Value{ get; };
+
+        // The line number for the failure reason, if determined.
+        UInt32 Line{ get; };
+
+        // The line number for the failure reason, if determined.
```
msftrubengu, line 472: nit: line column

src/Microsoft.Management.Configuration/ConfigurationSetParser.cpp
```diff
@@ -86,15 +105,60 @@ namespace winrt::Microsoft::Management::Configuration::implementation
...
-        return std::make_unique<ConfigurationSetParserError>(WINGET_CONFIG_ERROR_UNKNOWN_CONFIGURATION_FILE_VERSION, versionNode.as<std::string>());
+        return (schemaVersion == SemanticVersion{ "0.1" } || schemaVersion == SemanticVersion{ "0.2" });
     }
+    catch (...) { LOG_CAUGHT_EXCEPTION(); return false; }
 
-    void ConfigurationSetParser::SetError(hresult result, std::string_view field)
+    hstring ConfigurationSetParser::LatestVersion()
     {
-        AICLI_LOG(Config, Error, << "ConfigurationSetParser error: " << AppInstaller::Logging::SetHRFormat << result << " [" << field << "]");
+        return hstring{ L"0.1" };
```
msftrubengu, line 129: shouldn't this be 0.2?

src/Microsoft.Management.Configuration/ConfigurationSetParser.cpp
```diff
@@ -86,15 +105,60 @@ namespace winrt::Microsoft::Management::Configuration::implementation
...
+        else if (schemaVersion.PartAt(0).Integer == 0 && schemaVersion.PartAt(1).Integer == 2)
+        {
+            return std::make_unique<ConfigurationSetParser_0_2>(std::move(document));
+        }
+
+        AICLI_LOG(Config, Error, << "Unknown configuration version: " << schemaVersion.ToString());
+        return std::make_unique<ConfigurationSetParserError>(WINGET_CONFIG_ERROR_UNKNOWN_CONFIGURATION_FILE_VERSION, GetFieldName(FieldName::ConfigurationVersion), versionNode.as<std::string>());
+    }
+
+    bool ConfigurationSetParser::IsRecognizedSchemaVersion(hstring value) try
```
msftrubengu, line 117: maybe there can be something like an enum that has the recognized versions, unknown and max. a helper method that returns the string or semantic representation and compare with the value given. then Create and current callers of IsRecognizedSchemaVersion throw if the result is unknown and LatestVersion just uses max. but maybe is too much.

src/Microsoft.Management.Configuration/ConfigurationSetParser_0_2.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    namespace
+    {
+        // Contains the qualified resource name information.
+        struct QualifiedResourceName
+        {
+            QualifiedResourceName(hstring input)
+            {
+                std::wstring_view inputView = input;
+                size_t pos = inputView.find('/');
```
msftrubengu, line 29: should we fail if there are more than one / ?

src/Microsoft.Management.Configuration/ConfigurationSetParser.cpp
JohnMcPMS: I don't think an enum is warranted here, and I would only expect the value to be set in the event that the caller knows what they are doing with it (and therefore knows the versions).

src/Microsoft.Management.Configuration/ConfigurationSetParser_0_2.cpp
JohnMcPMS: Given that `/` isn't a valid character in module or resource names, I don't think I mind if we break non-functional files at a later date.

msftrubengu (approved):

PR #3187, opened 2023-04-25
Start Microsoft.WinGet.Configuration module

msftrubengu:
This PR sets the foundation for the new PowerShell module Microsoft.WinGet.Configuration which is going to be an equivalent of `winget configure`

The module is going to be built as a binary module and it supports using its own ACL to load dependencies. The Microsoft.WinGet.Configuration.Cmdlets project will contain all the cmdlet entry points and Microsoft.WinGet.Configuration.Engine the actual logic.

This module consumes Microsoft.Management.Configuration which uses cswinrt and requires being run on an MTA. This is achieved via MtaCommand.cs. This class contains functions that will verify if the current thread is already running on an MTA (via pwsh.exe -MTA) and if not start a new MTA thread. Every command that consumes winrt objects must implement it.

MtaCommand.cs also implements methods to handle waiting for asynchronous calls and allows sending messages to the PowerShell host that are required to be executed in the main thread (only WriteDebug is implemented for now, but any other method follows the same approach). This is based on the async mechanism that [azure-powershell](https://github.com/Azure/azure-powershell/blob/main/src/Workloads/generated/runtime/AsyncCommandRuntime.cs) uses.

This PR includes `Invoke-WinGetConfiguration` but is just a proof-of-concept cmdlet and should not be taken seriously.

The next PR will be all the changes in the configuration processor to enable `ConfigurationProcessorType.Default`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3187)


src/PowerShell/Microsoft.WinGet.Configuration.Cmdlets/Cmdlets/InvokeWinGetConfigurationCmdlet.cs
```diff
@@ -0,0 +1,50 @@
...
+    [Cmdlet(VerbsLifecycle.Invoke, "WinGetConfiguration")]
+    public sealed class InvokeWinGetConfigurationCmdlet : PSCmdlet
+    {
+        /// <summary>
+        /// Gets or sets the configuration file.
+        /// </summary>
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true)]
+        public string File { get; set; }
```
JohnMcPMS, line 27: There should be a `Get-WinGetConfiguration` command that is the primary method to open a file and other commands should be taking in a file path only as a convenience.  Given that there might be an arbitrary number of mechanisms to load a configuration set, maybe we even force it to be loaded with the `Get`.

JohnMcPMS (approved):

g0t4:
About a week ago I stumbled on the `Microsoft.WinGet.Client` & `Microsoft.WinGet.Dsc` modules. I immediately incorporated them into my own machine init scripts. Work like a charm. I'm looking forward to what you're doing here with `winget configure` and `DevHome`... keep up the great work! 

Are you looking for feedback on this `Configuration` module?

denelon:
@g0t4 yes, we're always looking for feedback. Package Management is complex to begin with, and adding configuration management is probably an order of magnitude more complex. Feel free to create issues or discussions. We're always looking for user scenarios that either aren't covered, or where the behavior isn't quite right.

PR #3193, opened 2023-04-27
Enable building everything needed in PowerShell with static runtime

JohnMcPMS:
## Change
Adds a solution and project configurations for building everything needed for PowerShell with static runtime linkage.  This enables us to produce modules without any external dependencies (beyond PowerShell of course).

Also fixes a bug in the project setup that wasn't renaming one of the output files.

## Validation
Ensured that the static runtime linkage build for the client module works in `SYSTEM` context as that is the only time we currently use any of these changes.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3193)


src/PowerShell/WinGetPowerShell.sln
```diff
@@ -0,0 +1,849 @@
+﻿
```
msftrubengu, line 1: I know you love this so maybe you can add the new powershell projects, the server and processor 😁

msftrubengu (approved):

PR #3195, opened 2023-04-27
Fix tags for --count argument

### LINKED ISSUE #3194: Count argument is throwing an error

`winget search a -n 20` is throwing an error

### Steps to reproduce

`winget search a -n 20`

### Expected behavior

The top 20 packages are displayed.

### Actual behavior

An argument was provided that can only be used for single package

### Environment

```shell
winget --info
Windows Package Manager (Preview) v1.5.1081-preview

Windows: Windows.Desktop v10.0.22621.1555
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.1081.0
```

florelis:
This fixes the tags on the --count argument, which were causing an error during arg validation when it was used.

Closes #3194

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -48,7 +48,7 @@ namespace AppInstaller::CLI
         case Execution::Args::Type::Source:
             return { type, "source"_liv, 's', ArgTypeCategory::Source };
         case Execution::Args::Type::Count:
-            return { type, "count"_liv, 'n', ArgTypeCategory::MultiplePackages };
+            return { type, "count"_liv, 'n', ArgTypeCategory::PackageQuery | ArgTypeCategory::SinglePackageQuery };
         case Execution::Args::Type::Exact:
             return { type, "exact"_liv, 'e', ArgTypeCategory::PackageQuery };
 


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -48,7 +48,7 @@ namespace AppInstaller::CLI
         case Execution::Args::Type::Source:
             return { type, "source"_liv, 's', ArgTypeCategory::Source };
         case Execution::Args::Type::Count:
-            return { type, "count"_liv, 'n', ArgTypeCategory::MultiplePackages };
+            return { type, "count"_liv, 'n', ArgTypeCategory::PackageQuery | ArgTypeCategory::SinglePackageQuery };
```
JohnMcPMS, line 51: I don't see how `MultiplePackages` is incorrect from an English semantics standpoint (I'm not sure how it differs from `PackageQuery` either though, which itself seems like an issue).

This feels like it is masking the actual issue since there is no way that count applies to `SinglePackageQuery`.  I don't think we are interested in providing an "I'm feeling lucky" experience.

From looking more into the `ArgTypeCategory` design, what is needed is to make `Args::Type::Query` into `Args::Type::QueryForSingle` and `Args::Type::QueryForMany` (as example names). This will allow for different categories to match the different semantics for the commands.

JohnMcPMS: Although I suppose we haven't done what I suggested for the name, etc. so maybe we don't want to do that for now.

JohnMcPMS (approved):

src/AppInstallerCLICore/Argument.cpp
yao-msft: This was related to a previous feature to support installing multiple packages with 1 command invocation.

`winget install TestApp1 TestApp2` is categorized as Multiple
`winget install TestApp1` is categorized as SinglePackageQuery

The naming may seem a bit confusing with the concept of querying for single/multiple package

PR #3201, opened 2023-05-01
Add Friendlier Messages for Error Codes

Trenly:

Not all the error codes had a friendly message associated. This PR adds them.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3201)

### DIFF
--- a/src/AppInstallerSharedLib/Errors.cpp
+++ b/src/AppInstallerSharedLib/Errors.cpp
@@ -243,9 +243,9 @@ namespace AppInstaller
             case APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_FINISH:
                 return "Restart your PC to finish installation.";
             case APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_INSTALL:
-                return "Your PC will restart to finish installation.";
-            case APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_INITIATED:
                 return "Installation failed. Restart your PC then try again.";
+            case APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_INITIATED:
+                return "Your PC will restart to finish installation.";
             case APPINSTALLER_CLI_ERROR_INSTALL_CANCELLED_BY_USER:
                 return "You cancelled the installation.";
             case APPINSTALLER_CLI_ERROR_INSTALL_ALREADY_INSTALLED:
--- a/src/AppInstallerSharedLib/Errors.cpp
+++ b/src/AppInstallerSharedLib/Errors.cpp
@@ -266,14 +266,12 @@ namespace AppInstaller
             // Status values for check package installed status results.
             case WINGET_INSTALLED_STATUS_ARP_ENTRY_NOT_FOUND:
                 return "The Apps and Features Entry for the package could not be found.";
-
             case WINGET_INSTALLED_STATUS_INSTALL_LOCATION_NOT_APPLICABLE:
                 return "The install location is not applicable.";
             case WINGET_INSTALLED_STATUS_INSTALL_LOCATION_NOT_FOUND:
                 return "The install location could not be found.";
             case WINGET_INSTALLED_STATUS_FILE_HASH_MISMATCH:
                 return "The hash of the existing file did not match.";
-
             case WINGET_INSTALLED_STATUS_FILE_NOT_FOUND:
                 return "File not found.";
             case WINGET_INSTALLED_STATUS_FILE_FOUND_WITHOUT_HASH_CHECK:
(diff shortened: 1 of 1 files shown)


JohnMcPMS:
While this change isn't inherently bad, the correct solution that just hasn't been done is to have this function return an `AppInstaller::StringResource::String`, then have a function to get the internal, non-localized string message from that string name.  The CLI paths would prefer to get the localized resource value, falling back to the internal path if one isn't present (and hopefully never needing to do that).

Not expecting you to do any of this, just dumping my "we should do this one day" thoughts out there...

Trenly:
> While this change isn't inherently bad, the correct solution that just hasn't been done is to have this function return an `AppInstaller::StringResource::String`, then have a function to get the internal, non-localized string message from that string name. The CLI paths would prefer to get the localized resource value, falling back to the internal path if one isn't present (and hopefully never needing to do that).
> 
> Not expecting you to do any of this, just dumping my "we should do this one day" thoughts out there...

I completely agree with you; Definitely something to add to the "To-Do" list. If I have time this weekend I can look into that (maybe)

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -237,9 +237,9 @@ namespace AppInstaller
...
                 return "There's not enough memory available to install. Close other applications then try again.";
             case APPINSTALLER_CLI_ERROR_INSTALL_NO_NETWORK:
-                return "This application requires internet connectivity.Connect to a network then try again.";
+                return "This application requires internet connectivity. Connect to a network then try again.";
             case APPINSTALLER_CLI_ERROR_INSTALL_CONTACT_SUPPORT:
-                return "This application encountered an error during installation.Contact support.";
+                return "This application encountered an error during installation. Contact support.";
             case APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_FINISH:
                 return "Restart your PC to finish installation.";
             case APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_INSTALL:
```
florelis, line 245: I just realized that this and the next one have the strings swapped

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -262,6 +262,73 @@ namespace AppInstaller
                 return "Invalid parameter.";
             case APPINSTALLER_CLI_ERROR_INSTALL_SYSTEM_NOT_SUPPORTED:
                 return "Package not supported by the system.";
+            
+            // Status values for check package installed status results.
+            case WINGET_INSTALLED_STATUS_ARP_ENTRY_NOT_FOUND:
+                return "The Add & Remove Programs Entry for the package could not be found.";
```
yao-msft, line 268:
```suggestion
                return "The Apps and Features Entry for the package could not be found.";

```

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -262,6 +262,73 @@ namespace AppInstaller
...
+            
+            // Status values for check package installed status results.
+            case WINGET_INSTALLED_STATUS_ARP_ENTRY_NOT_FOUND:
+                return "The Add & Remove Programs Entry for the package could not be found.";
+            case WINGET_INSTALLED_STATUS_INSTALL_LOCATION_NOT_APPLICABLE:
+                return "The install location is not applicable.";
+            case WINGET_INSTALLED_STATUS_INSTALL_LOCATION_NOT_FOUND:
+                return "The install location could not be found.";
+            case WINGET_INSTALLED_STATUS_FILE_HASH_MISMATCH:
+                return "The hash of the existing file did not match. Please uninstall and try again.";
```
yao-msft, line 274:
```suggestion
                return "The hash of the existing file did not match.";

```
This is a status check of one file for callers like Intune to notify IT admins of package status. IT admins would decide what to do. Maybe the package is updated outside of winget etc.

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -262,6 +262,73 @@ namespace AppInstaller
...
+            case WINGET_CONFIG_ERROR_SET_DEPENDENCY_CYCLE:
+                return "The dependency graph contains a cycle which cannot be resolved.";
+            case WINGET_CONFIG_ERROR_INVALID_FIELD_VALUE:
+                return "The configuration has an invalid field value.";
+            case WINGET_CONFIG_ERROR_MISSING_FIELD:
+                return "The configuration is missing a field.";
+
+            // Configuration Processor Errors
+            case WINGET_CONFIG_ERROR_UNIT_NOT_INSTALLED:
+                return "The configuration unit was not in the module as expected.";
```
yao-msft, line 314:
```suggestion
                return "The configuration unit was not installed.";

```

yao-msft:
Nit: can you remove the extra line, not sure if it's github's apply suggestion issue.

yao-msft (approved):

PR #3204, opened 2023-05-01
Initial implementation of Microsoft.WinGet.Configuration cmdlets.

msftrubengu:
This PR adds 5 cmdlets

- Get-WinGetConfiguration: creates a configuration set given a file path.
- Get-WinGetConfigurationDetails: gets the details from a set.
- Invoke-WinGetConfiguration: applies the configuration and waits for completion.
- Start-WinGetConfiguration: starts applying the configuration asynchronously. Returns PSConfigurationTask. 
- Complete-WinGetConfiguration: waits for the PSConfigurationTask to be completed.

For now, a call the Get-WinGetConfiguration is required to start. It returns a PSConfigurationSet which can be pass to the other cmdlets. 

If Invoke-WinGetConfiguration or Start-WinGetConfiguration get called before Get-WinGetConfigurationDetails, they will retrieve the details.

There is no output in PowerShell except for debug messages and the objects. Future PRs will provide a better user experience similar to what winget configure does.

By default, Start-WinGetConfiguration won't write anything to the stream buffers  (except the returned object).because it needs to be executed in the main thread. When Complete-WinGetConfiguration gets called, writting to the streams gets enabled. For now, all the messages before Complete gets lost but in the future, we can store them and show them (similar to what Start-Job/Receive-Job does).

This PR also adds a new ConfigurationProcessorPolicy enum for creating a ConfigurationSetProcessor and can be added via IConfigurationProcessorFactoryProperties. The enum is a mirror of the PowerShell Execution Policies. By default, winget uses Unrestricted (we should change in the near future) and the module uses the same as the current PowerShell session.

I couldn't get the execution policy at the time of creating the set processor because it is not possible to create a runspace based on the current session. At that point, we are already running in another thread so the runspace can't be found. This means that the two types of runspaces are always "hosted" which just means that there's a new runspace created from them.

When we enable variables in configuration, there needs to be something that makes them visible from the current PowerShell session to our runspace (maybe using synchronized hashtable).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3204)

### DIFF
--- a/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
+++ b/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Management.Configuration.Processor
                 var envFactory = new ProcessorEnvironmentFactory(this.type);
                 var processorEnvironment = envFactory.CreateEnvironment(
                     this,
-                    this.properties != null ? this.properties.Policy : ConfigurationProcessorPolicy.Default);
+                    this.properties != null ? this.properties.Policy : ConfigurationProcessorPolicy.RemoteSigned);
 
                 if (this.properties is not null)
                 {
--- a/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
+++ b/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Management.Configuration.Processor
                 var envFactory = new ProcessorEnvironmentFactory(this.type);
                 var processorEnvironment = envFactory.CreateEnvironment(
                     this,
-                    this.properties != null ? this.properties.Policy : ConfigurationProcessorPolicy.RemoteSigned);
+                    this.properties?.Policy ?? ConfigurationProcessorPolicy.RemoteSigned);
 
                 if (this.properties is not null)
                 {
--- a/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
+++ b/src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
@@ -55,7 +55,9 @@ namespace Microsoft.Management.Configuration.Processor
                 this.OnDiagnostics(DiagnosticLevel.Verbose, $"Creating set processor for `{set.Name}`...");
 
                 var envFactory = new ProcessorEnvironmentFactory(this.type);
-                var processorEnvironment = envFactory.CreateEnvironment(this);
+                var processorEnvironment = envFactory.CreateEnvironment(
+                    this,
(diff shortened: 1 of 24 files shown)


src/Microsoft.Management.Configuration.Processor/Public/ConfigurationSetProcessorFactory.cs
```diff
@@ -55,7 +55,9 @@ public IConfigurationSetProcessor CreateSetProcessor(ConfigurationSet set)
                 this.OnDiagnostics(DiagnosticLevel.Verbose, $"Creating set processor for `{set.Name}`...");
 
                 var envFactory = new ProcessorEnvironmentFactory(this.type);
-                var processorEnvironment = envFactory.CreateEnvironment(this);
+                var processorEnvironment = envFactory.CreateEnvironment(
+                    this,
+                    this.properties != null ? this.properties.Policy : ConfigurationProcessorPolicy.RemoteSigned);
```
JohnMcPMS, line 60:
```suggestion
                    this.properties?.Policy ?? ConfigurationProcessorPolicy.RemoteSigned);
```
I'm pretty sure this is the fancy operator version.

src/PowerShell/Microsoft.WinGet.Configuration.Cmdlets/Cmdlets/CompleteWinGetConfigurationCmdlet.cs
```diff
@@ -0,0 +1,42 @@
...
+    public sealed class CompleteWinGetConfigurationCmdlet : PSCmdlet
+    {
+        /// <summary>
+        /// Gets or sets the configuration task.
+        /// </summary>
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipeline = true,
+            ValueFromPipelineByPropertyName = true)]
+        public PSConfigurationTask ConfigurationTask { get; set; }
```
JohnMcPMS, line 29:
```suggestion
        public PSConfigurationJob ConfigurationJob { get; set; }
```
I feel like `Job` is more parallel to the PowerShell naming.

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
```diff
@@ -13,74 +13,278 @@ namespace Microsoft.WinGet.Configuration.Engine.Commands
...
+
+            var runningTask = this.StartApplyInternal(psConfigurationSet);
+            this.WriteObject(new PSConfigurationTask(runningTask, this));
+        }
+
+        /// <summary>
+        /// Applies configuration.
+        /// </summary>
+        /// <param name="psConfigurationSet">PSConfigurationSet.</param>
+        public void Apply(PSConfigurationSet psConfigurationSet)
```
JohnMcPMS, line 131: You should create another helper(s) so that this method can always be as simple as `ContinueHelper(StartApplyHelper(set));`.

JohnMcPMS (approved):

PR #3217, opened 2023-05-04
Allow prerelease modules for configuration

msftrubengu:
This PR allows a configuration file to have a boolean allowPrerelease directive to use prerelease DSC resources.

```
properties:
  configurationVersion: 0.2
  resources:
    - resource: FakeModule/FakeResource
      directives:
        version: '1.0.0-preview'
        allowPrerelease: true
      settings:
        Args: 'args'
```

Internally, allow prerelease is used only for a module that it needs to get install. When is enabled, the processor will use `-AllowPrerelease` to the Find-DscResource command. 

It also fixes and infinite loop in `SplitIntoLines`. If the found offset is the one of "\r\n", it uses the same offset for `pos` in `find_first_of`, so it keeps returning the offset making `nextOffset - currentOffset = 0`

I don't know why VisualStudio decided to change a character...
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3217)

### DIFF
--- a/src/Microsoft.Management.Configuration.Processor/Helpers/PowerShellHelpers.cs
+++ b/src/Microsoft.Management.Configuration.Processor/Helpers/PowerShellHelpers.cs
@@ -48,16 +48,33 @@ namespace Microsoft.Management.Configuration.Processor.Helpers
 
             if (!string.IsNullOrEmpty(version))
             {
+                // Prerelease versions append the prerelease tag after a -
+                // PowerShell doesn't handle semantic versions.
+                if (version.Contains("-"))
+                {
+                    version = version[..version.IndexOf("-")];
+                }
+
                 moduleInfo.Add(Parameters.RequiredVersion, version);
             }
 
             if (!string.IsNullOrEmpty(minVersion))
             {
+                if (minVersion.Contains("-"))
+                {
+                    minVersion = minVersion[..minVersion.IndexOf("-")];
+                }
+
                 moduleInfo.Add(Parameters.ModuleVersion, minVersion);
             }
 
             if (!string.IsNullOrEmpty(maxVersion))
             {
+                if (maxVersion.Contains("-"))
+                {
+                    maxVersion = maxVersion[..maxVersion.IndexOf("-")];
+                }
+
                 // For some reason, the constructor of ModuleSpecification that takes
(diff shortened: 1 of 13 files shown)


src/Microsoft.Management.Configuration.Processor/Helpers/PowerShellHelpers.cs
```diff
@@ -48,24 +48,41 @@ internal static class PowerShellHelpers
 
             if (!string.IsNullOrEmpty(version))
             {
+                // Prerelease versions append the prerelease tag after a -
+                // PowerShell doesn't handle semantic versions.
+                if (version.Contains("-"))
```
JohnMcPMS, line 53: Maybe a function? `GetOnlyNumericVersion` or something?  Or a new type that splits them out.  It could be used to implicitly declare `allowPrerelease` by having something after the `-`.

src/Microsoft.Management.Configuration.Processor/ProcessorEnvironments/HostedEnvironment.cs
```diff
@@ -256,41 +256,54 @@ public void ImportModule(string path)
...
                 parameters.Add(Parameters.Repository, repository);
             }
 
             using PowerShell pwsh = PowerShell.Create(this.Runspace);
 
-            // TODO: Implement prerelease directive.
+            pwsh.AddCommand(Commands.FindDscResource)
+                .AddParameters(parameters);
+
+            if (allowPrerelease)
```
JohnMcPMS, line 298: Why isn't this placed into the `parameters` like the others?

JohnMcPMS (approved):

src/Microsoft.Management.Configuration.Processor/ProcessorEnvironments/HostedEnvironment.cs
msftrubengu: I was to lazy to see if a switch with bool was supported in the parameters hashtable

denelon:
Do we need to modify the JSON schema for this?

msftrubengu:
I can create the schema in another PR.

PR #3222, opened 2023-05-04
Queue write operations for Start-* cmdlets and hook up diagnostics

msftrubengu:
When a Start-* cmdlet starts its operation it won't write to any PowerShell stream because control is returned back to the user almost immediately.  This changes queue messages to be then displayed at Continue-* time. This way, we will have all the messages in the appropriate order. It applies for WriteDebug, WriteVerbose, WriteWarning, WriteError and WriteProgress. 

WriteProgress will only be called when the activity hasn't been completed. A caller must use `AsyncCommand.GetNewProgressActivityId` to generate a new activity id.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3222)

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.Configuration.Cmdlets/Cmdlets/GetWinGetConfigurationCmdlet.cs
+++ b/src/PowerShell/Microsoft.WinGet.Configuration.Cmdlets/Cmdlets/GetWinGetConfigurationCmdlet.cs
@@ -20,6 +20,7 @@ namespace Microsoft.WinGet.Configuration.Cmdlets
     public sealed class GetWinGetConfigurationCmdlet : PSCmdlet
     {
         private ExecutionPolicy executionPolicy = ExecutionPolicy.Undefined;
+        private bool canUseTelemetry = true;
 
         /// <summary>
         /// Gets or sets the configuration file.
@@ -35,6 +36,7 @@ namespace Microsoft.WinGet.Configuration.Cmdlets
         protected override void BeginProcessing()
         {
             this.executionPolicy = Utilities.GetExecutionPolicy();
+            this.canUseTelemetry = Utilities.CanUseTelemetry();
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/PowerShell/Microsoft.WinGet.Configuration.Cmdlets/Cmdlets/GetWinGetConfigurationCmdlet.cs
```diff
@@ -44,7 +44,7 @@ protected override void ProcessRecord()
         {
             CancellationTokenSource source = new ();
```
JohnMcPMS, line 45: Remove?  Repeat for all the other times.

JohnMcPMS (approved):

src/PowerShell/Microsoft.WinGet.Configuration.Engine/PSObjects/PSConfigurationProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                this.diagnosticCommand = newDiagnosticCommand;
+            }
+        }
+
+        private void LogConfigurationDiagnostics(DiagnosticInformation diagnosticInformation)
+        {
+            try
+            {
+                // This is expensive.
```
JohnMcPMS, line 66: The only reason I see the lock as necessary is to prevent multiple calls to write output at the same time.  If that is already being serialized (it was to some degree before) then this lock seems to protect the read/write of the `diagnosticCommand` reference.  [That shouldn't be needed](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/variables#96-atomicity-of-variable-references) if you just grab a reference to the object within the function to use for the duration.

JohnMcPMS (approved):

src/PowerShell/Microsoft.WinGet.Configuration.Engine/PSObjects/PSConfigurationProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+                lock (CmdletLock)
+                {
+                    switch (diagnosticInformation.Level)
+                    {
+                        // PowerShell doesn't have critical and critical isn't an error.
+                        case DiagnosticLevel.Critical:
+                        case DiagnosticLevel.Warning:
+                            this.diagnosticCommand.WriteWarning(diagnosticInformation.Message);
```
JohnMcPMS, line 74: Check for null or enforce that it is not null in the constructor/setter.

PR #3224, opened 2023-05-05
Abstract Source Open Workflows

Trenly:

It seemed to me that the way sources are opened is very repetitive across commands, and especially within subcommands. If this is something that isn't wanted, please close it, just thought I would make the PR and see where it went.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3224)


src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -117,7 +117,7 @@ namespace AppInstaller::CLI
         {
             context <<
                 Workflow::ReportExecutionStage(ExecutionStage::Discovery) <<
-                Workflow::OpenSource();
+                Workflow::OpenSourceFor(OperationType::Install);
 
             if (!context.Args.Contains(Execution::Args::Type::Force))
```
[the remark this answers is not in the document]
Trenly: Ah, I didn't see that down there. Should be resolved

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -444,6 +444,42 @@ namespace AppInstaller::CLI::Workflow
...
+        case OperationType::List:
+        case OperationType::Uninstall:
+        case OperationType::Upgrade:
+            context <<
+                Workflow::OpenSource() <<
+                Workflow::OpenCompositeSource(Workflow::DetermineInstalledSource(context));
+            break;
+        case OperationType::Pin:
+        case OperationType::Export:
+        case OperationType::Completion:
```
florelis, line 473: I'm not sure about `Completion`. There are other cases like search where we don't use the installed, and I see you had to use Search type there

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -119,14 +117,12 @@ namespace AppInstaller::CLI
             // When we are given an ID, just pin that available package without checking for installed.
             // This helps when there are matching issues, for example due to multiple side-by-side installs.
             context <<
-                Workflow::OpenSource();
+                Workflow::OpenSourceFor(OperationType::Search);
```
florelis, line 120: Also not a fan of having to use ::Search in the Pin command

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -168,9 +164,7 @@ namespace AppInstaller::CLI
 
     void PinRemoveCommand::Complete(Execution::Context& context, Args::Type valueType) const
     {
-        context <<
-            Workflow::OpenSource() <<
-            Workflow::OpenCompositeSource(Repository::PredefinedSource::Installed);
+        context << Workflow::OpenSourceFor(OperationType::Pin);
```
florelis, line 167: I forgot to change this. Maybe we should also conditionally open only the available source here

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
Trenly: Hmm. I can understand that point. One thing that I don't understand is the difference between the `PredefinedSource::Installed` and `DetermineInstalledSource`. The cases where completion uses the Predefined source are for the commands which use `DetermineInstalledSource`. 

Perhaps what would make the most sense is to have completion throw the invalid state and keep the current behavior of explicit management of sources for completion

src/AppInstallerCLICore/Commands/PinCommand.cpp
Trenly: Understandable; Will adjust

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
yao-msft: DetermineInstalledSource will filter the installed source to include only user scope | machine scope | default (previous behavior where for win32, user and machine scopes are included, for msix, only apps registered to user are included)

Issue #3211, opened 2023-05-04
Blocking & Gating Pin should require force Uninstall


Trenly:
### Description of the new feature / enhancement

Users may pin a package for a project they don't touch often but which is needed for that project. If a user goes to clean up their installed packages, running `winget uninstall` will not inform them that there is a pin, resulting in a broken build next time they touch their legacy project.

It would be nice if there was a warning that a pin existed for the packages, requiring either the pin to be removed or `--force` to be used to proceed with the uninstall.

Trenly:
I can add this feature once #3167 is merged

florelis:
> I can add this feature once #3167 is merged, or if @florelis Flor Elisa Chacon Ochoa FTE doesn't mind, I can create a PR into the branch on her repo to merge it with #3167

I would rather that be a separate change. #3167 is already large enough

PR #3236, opened 2023-05-09
Make pins block uninstalls

Trenly:
  - #3211 
-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3236)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
@@ -3,6 +3,7 @@
 #include "pch.h"
 #include "UninstallFlow.h"
 #include "InstallFlow.h"
+#include "PinFlow.h"
 #include "WorkflowBase.h"
 #include "DependenciesFlow.h"
 #include "ShellExecuteInstallerHandler.h"
@@ -62,6 +63,8 @@ namespace AppInstaller::CLI::Workflow
     {
         context <<
             Workflow::GetInstalledPackageVersion <<
+            Workflow::OpenPinningIndex() <<
+            Workflow::SearchPin <<
             Workflow::EnsureSupportForUninstall <<
             Workflow::GetUninstallInfo <<
             Workflow::GetDependenciesInfoForUninstall <<
@@ -369,6 +372,36 @@ namespace AppInstaller::CLI::Workflow
 
     void EnsureSupportForUninstall(Execution::Context& context)
     {
+        // If the installed package has a pin which is stricter than a "Pinning" pin, require --force
+        auto pins = context.Get<Execution::Data::Pins>();
+        if (!pins.empty())
+        {
+            bool pinRequiresForce = false;
+            for (const auto& pin : pins)
+            {
+                if (Pinning::IsStricter(pin.GetType(), Pinning::PinType::Pinning))
+                {
+                    pinRequiresForce = true;
+                    AICLI_LOG(CLI, Warning, << "Found Pin " << pin.ToString());
+                    break;
+                }
+            }
+
+            if (pinRequiresForce && context.Args.Contains(Execution::Args::Type::Force))
+            {
+
+                AICLI_LOG(CLI, Warning, << "Continuing with uninstall due to --force argument");
+                context.Reporter.Warn() << Resource::String::UninstallIsPinnedForced << std::endl;
+            }
+            else if (pinRequiresForce)
+            {
+                AICLI_LOG(CLI, Error, << "Package pin is preventing uninstall");
+                context.Reporter.Error() << Resource::String::UninstallIsPinnedUseForce << std::endl;
+                AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_PACKAGE_IS_PINNED);
+            }
+        }
+
+        // If the installed package is portable or MSIX and is machine scoped, require admin to uninstall
         auto installedPackageVersion = context.Get<Execution::Data::InstalledPackageVersion>();
         const std::string installedTypeString = installedPackageVersion->GetMetadata()[PackageVersionMetadata::InstalledType];
         auto installedType = ConvertToInstallerTypeEnum(installedTypeString);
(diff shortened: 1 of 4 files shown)


src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -369,6 +372,36 @@ namespace AppInstaller::CLI::Workflow
...
+                {
+                    pinRequiresForce = true;
+                    AICLI_LOG(CLI, Warning, << "Found Pin " << pin.ToString());
+                    break;
+                }
+            }
+
+            if (pinRequiresForce && context.Args.Contains(Execution::Args::Type::Force))
+            {
+
```
florelis, line 392: Nit: empty line

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -369,6 +372,36 @@ namespace AppInstaller::CLI::Workflow
...
+                AICLI_LOG(CLI, Warning, << "Continuing with uninstall due to --force argument");
+                context.Reporter.Warn() << Resource::String::UninstallIsPinnedForced << std::endl;
+            }
+            else if (pinRequiresForce)
+            {
+                AICLI_LOG(CLI, Error, << "Package pin is preventing uninstall");
+                context.Reporter.Error() << Resource::String::UninstallIsPinnedUseForce << std::endl;
+                AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_PACKAGE_IS_PINNED);
```
florelis, line 400: Does this error message have the right string? I may have said specifically "upgrade" in there

florelis (approved):

yao-msft:
When we first spec'ed the pinning feature, the community asks were in several categories:
1.  I do not want winget to manage package updates because I have other package managers managing these packages (blocking)
2.  I do not want to accidentally update some packages with upgrade --all (pinning)
3.  I want to limit some packages to specific version (gating)
4.  I prefer one source over another (I pin winget-pkgs source so that I always get updates from msstore source) (pin for one available package)
5.  I do not want some installed packages being updated (pin for one installed package)

To achieve above, we designed the pinning table to have pinning, gating and blocking (either for an available package or an installed package)

With the complexity of all the pinning types and package types, I personally do not think these pins should be mixed with uninstalling behaviors. For example, if I block upgrade some app from winget-pkgs source, it does not mean I cannot uninstall that app. If I pin package to v1.1.*, when v1.1.1 is available, I do still want to uninstall v1.1.0

Pinning is mainly for better `upgrade --all ` customizations, where we do not have batch uninstall (i.e. `uninstall --all`). I feel winget uninstall is almost always from user intent. User will just pass in --force to uninstall and leave a stale pin in the index.

yao-msft (changes requested):
I think we need to go through the scenarios to see if this is actually needed. If yes, whether we want to mix the upgrade pins with uninstall.

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
Trenly: I'm 99% sure it does, but will verify

Trenly:
> I do not want winget to manage package updates because I have other package managers managing these packages (blocking)

Along the same line of thinking, if another package manager is being used to manage the packages, using winget to uninstall it could cause unintended results in the other package manager. Requiring force for an uninstall is an extra reminder to the user that they were using a different package manager for that package.

> I do not want to accidentally update some packages with upgrade --all (pinning)

Pinning pins remain unaffected by this change

> I want to limit some packages to specific version (gating)

This is actually one of the reasons that I brought this up. Once a pin is added, it will stay there until a user intentionally removes it. It could be months after a pin is added and a user goes to uninstall the package, not remembering they needed the package at a specific version for some workflow they use only once or twice a year. Requiring force for an uninstall is an extra reminder to the user that they wanted it to be at that specific version.

> I do not want some installed packages being updated (pin for one installed package)

To me this is no different than the first scenario (blocking) or even a gating pin with a version range of a single version. It is entirely possible that some users just want an app not to be updated, but it's also possible that some users require specific versions and would want to also block uninstall.

> For example, if I block upgrade some app from winget-pkgs source, it does not mean I cannot uninstall that app.  If I pin package to v1.1.*, when v1.1.1 is available, I do still want to uninstall v1.1.0

That is true. I would have to see how the upgrade flow calls the uninstall flow in cases where uninstallPrevious is specified. 

> I feel winget uninstall is almost always from user intent.

I agree completely. However, user intent is not always as intentional as we like to believe. Most people can't remember what they had for lunch two Tuesday's ago, so I felt that a reminder of their past actions would be minimally invasive to confirm the intent. 

> User will just pass in --force to uninstall and leave a stale pin in the index.

In some cases yes, and in some cases no. Currently, all uninstalls of pinned packages leave a stale pin in the index, so that isn't really a concern IMO.. The thought behind `--force` was to act as a reminder that there was some reason the user had it pinned, while making it easy to bypass. The other thought I had was to perhaps add a prompt that "A pin exists for this package, are you sure you want to remove it", but I thought that requiring force was more consistent with the current UX.

yao-msft:
Just left my thought here. If we do wish to associate pins with uninstall, I feel it should be limited to "blocking pin for installed package".

Pins for available (remote) package just blocks packages from a specific source (for the scenario I prefer one source over another i.e. I pin winget-pkgs source so that I always get updates from msstore source)

Pins for installed packages makes more sense to prevent uninstall. And in the context of gating, I want my package to be latest within some range. So it's still possible to uninstall a version that's slightly older.

P.S. We are in the final stage of winget 1.5 rc. We may need to pick this up in winget 1.6.

yao-msft:
Just left my thought here. If we do wish to associate pins with uninstall, I feel it should be limited to "blocking pin for installed package".

Pins for available (remote) package just blocks packages from a specific source (for the scenario I prefer one source over another i.e. I pin winget-pkgs source so that I always get updates from msstore source)

Pins for installed packages makes more sense to prevent uninstall. And in the context of gating, I want my package to be latest within some range. So it's still possible to uninstall a version that's slightly older.

P.S. We are in the final stage of winget 1.5 rc. We may need to pick this up in winget 1.6.

PR #3237, opened 2023-05-10
Allow ${WinGetConfigRoot} variable expansion

msftrubengu:
Allow settings to use the `$WinGetConfigRoot` variable.

This variable means the directory containing the configuration file and is helpful for units that want reference paths that are relative to the location of the configuration file.

For example, our test resource can use it like this:
```
properties:
  configurationVersion: 0.2
  resources:
    - resource: xSimpleTestResource/SimpleFileResource
      settings:
        path: '${WinGetConfigRoot}\..\test.txt'
        content: 'Ella baila sola'
```

To make it PowerShelly, the variable expansion is done ignoring case.

Since this is still in preview, the name of the variable can change when `winget configure` release.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3237)


JohnMcPMS:
> This variable means the current location of the configuration file and is helpful for units that want reference paths that are relative to the location of the configuration file.

I assume you mean "the directory containing the configuration file".

src/Microsoft.Management.Configuration.Processor/Helpers/ConfigurationUnitInternal.cs
```diff
@@ -150,6 +159,52 @@ public string ToIdentifyingString()
...
+            if (!string.IsNullOrEmpty(value))
+            {
+                // TODO: since we only support one variable, this only finds and replace
+                // $WingetConfigRoot if found in the string when the work of expanding
+                // string is done it should take into account other operators like the subexpression operator $()
+                if (value.Contains(configRoot, StringComparison.OrdinalIgnoreCase))
+                {
+                    if (string.IsNullOrEmpty(this.origin))
+                    {
+                        throw new UnitSettingConfigRootException(this.Unit.UnitName, settingName);
```
JohnMcPMS, line 198: When will this end up happening in the overall lifetime of processing the configuration?  I would prefer that it be in response to the request to process a set, but before any actual processing takes place.

JohnMcPMS (changes requested):

src/Microsoft.Management.Configuration.Processor/Helpers/ConfigurationUnitInternal.cs
msftrubengu: Right now, the expansion of settings are done before right before any call to Invoke-DscResource

src/Microsoft.Management.Configuration.Processor/Exceptions/UnitSettingConfigRootException.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    internal class UnitSettingConfigRootException : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UnitSettingConfigRootException"/> class.
+        /// </summary>
+        /// <param name="unitName">Unit name.</param>
+        /// <param name="setting">Setting.</param>
+        public UnitSettingConfigRootException(string unitName, string setting)
+            : base($"Unit: {unitName} Setting {setting} requires the ConfigurationSet Origin")
```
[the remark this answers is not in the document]
msftrubengu: 🫠

src/Microsoft.Management.Configuration.Processor/Helpers/ConfigurationUnitInternal.cs
```diff
@@ -184,21 +185,32 @@ public ValueSet GetExpandedSettings()
...
+                if (value.Contains(configRootVar, StringComparison.OrdinalIgnoreCase))
                 {
-                    if (string.IsNullOrEmpty(this.origin))
+                    if (string.IsNullOrEmpty(this.configurationFilePath))
                     {
                         throw new UnitSettingConfigRootException(this.Unit.UnitName, settingName);
                     }
 
-                    return value.Replace(configRoot, this.origin, StringComparison.OrdinalIgnoreCase);
+                    if (!File.Exists(this.configurationFilePath))
```
JohnMcPMS, line 202: Do these checks in the constructor to prevent doing them every time through the loop.

JohnMcPMS (changes requested):

src/Microsoft.Management.Configuration.Processor/Helpers/ConfigurationUnitInternal.cs
```diff
@@ -6,28 +6,38 @@
...
     using Microsoft.PowerShell.Commands;
+    using Windows.Foundation.Collections;
 
     /// <summary>
     /// Wrapper around Configuration units and its directives. Creates a normalized directives map
     /// for consumption.
     /// </summary>
     internal class ConfigurationUnitInternal
     {
+        private static string configRootVar = "$WinGetConfigRoot";
```
JohnMcPMS, line 23: After thinking about this some, I don't think we should go with this form when we are doing a simple find and replace.  Without considering an actual syntax, we would convert `$WinGetConfigRootName => C:Name` if the file was `C:\config.yml`.

I think we should use one of the other forms of PowerShell syntax here to force it to be compliant (or at least, much more likely to be) while still being able to use a find and replace for time.  So either `${WinGetConfigRoot}` or `$($WinGetConfigRoot)`.

JohnMcPMS (approved):

PR #3242, opened 2023-05-11
Microsoft.WinGet.Configuration messages

msftrubengu:
This PR starts printing information messages for `Get-WingetConfiguration` and `Get-WingetConfigurationDetails`. The messages are similar to what winget is showing.

Also:
- Add resource file with necessary configuration messages. These are mostly the same as the one winget uses.
- Modify AsyncCommand to always use a BlockingCollection for storing messages. The messages are written to PowerShell when a cmdlet waits for the task to complete or a Complete-* cmdlet executes and the task is already completed. This completely removes the context switch (which might come back if we ever require a ShouldContinue/ShouldProgress in the middle of the task)

Next:
- Finish apply cmdlets messages
- add formats file for output objects
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3242)


src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/AsyncCommand.cs
```diff
@@ -51,62 +41,74 @@ public abstract class AsyncCommand
...
-                    this.canWriteToStream = value;
-                }
-            }
-        }
-
         /// <summary>
         /// Cancel this operation.
         /// </summary>
-        public virtual void Cancel()
+        public virtual void Complete()
```
JohnMcPMS, line 109: Did we lose `Cancel`?  The comment on this function is still for `Cancel`.

JohnMcPMS (dismissed):

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/AsyncCommand.cs
msftrubengu: I used to have a cancel that no one called. I'll bring it back when integrate your cancellation support.

JohnMcPMS (approved):

PR #3253, opened 2023-05-16
Microsoft.WinGet.Client throws System.InvalidOperationException: Attempt to update previously set global instance

msftrubengu:
Microsoft.WinGet.Client throws `System.InvalidOperationException:“Attempt to update previously set global instance.”` when used in winget configure with a resource from Microsoft.WinGet.DSC

This is because the custom AssemblyLoadContext implemented by the client module loads WinRT.Runtime.dll into the same processes where the configuration server already have WinRT.Runtime.dll loaded. When there are multiple versions of WinRT.Runtime.dll loaded in process this exception gets thrown. [More info](https://github.com/microsoft/CsWinRT/issues/623)

The fix is to always load WinRT.Runtime.dll into the default assembly context, if its not loaded by the time our handler is executed and always return null in the Load implementation of the custom ALC. 

This is a problem any PowerShell module that uses cswinrt will have (if there's another one out there)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3253)

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Acl/WinGetAssemblyLoadContext.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Acl/WinGetAssemblyLoadContext.cs
@@ -52,7 +52,7 @@ namespace Microsoft.WinGet.Client.Acl
         internal static Assembly ResolvingHandler(AssemblyLoadContext context, AssemblyName assemblyName)
         {
             string name = $"{assemblyName.Name}.dll";
-            if (DefaultContextAssemblies.Any(a => a == name))
+            if (DefaultContextAssemblies.Any(a => a.Equals(name, StringComparison.OrdinalIgnoreCase)))
             {
                 string sharedPath = Path.Combine(SharedDependencyPath, name);
                 if (File.Exists(sharedPath))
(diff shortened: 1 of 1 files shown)


src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Acl/WinGetAssemblyLoadContext.cs
```diff
@@ -41,6 +51,16 @@ private WinGetAssemblyLoadContext()
         /// <returns>The assembly, null if not in our assembly location.</returns>
         internal static Assembly ResolvingHandler(AssemblyLoadContext context, AssemblyName assemblyName)
         {
+            string name = $"{assemblyName.Name}.dll";
+            if (DefaultContextAssemblies.Any(a => a == name))
```
JohnMcPMS, line 55: This comparison is case sensitive, and this load request could be coming from anywhere.

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Acl/WinGetAssemblyLoadContext.cs
```diff
@@ -53,13 +73,19 @@ internal static Assembly ResolvingHandler(AssemblyLoadContext context, AssemblyN
         /// <inheritdoc/>
         protected override Assembly Load(AssemblyName assemblyName)
         {
-            string path = $"{Path.Combine(SharedDependencyPath, assemblyName.Name)}.dll";
+            string name = $"{assemblyName.Name}.dll";
+            if (DefaultContextAssemblies.Any(a => a == name))
```
JohnMcPMS, line 77: Also case sensitive.

JohnMcPMS (approved):

PR #3255, opened 2023-05-16
Add initial version of manifest 1.5

yao-msft:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3255)


schemas/JSON/manifests/v1.5.0/manifest.defaultLocale.1.5.0.json
```diff
@@ -0,0 +1,213 @@
+{
+  "$id": "https://aka.ms/winget-manifest.defaultlocale.1.5.0.schema.json",
```
JohnMcPMS, line 2: I reallllllllllllllllly wish we had a single version file so that we could actually see the diff.

If there was a `manifests/latest` directory with unversioned schemas and you just copied the new files' contents into them, I think that would achieve the desired result.  Thoughts?

One could achieve similar results with a little more effort for the reviewer if one always started with a commit of the previous version's files, just renamed.  Meaning, the first commit of such a PR is always the result of "copy version N-1 files to version N directory with new names".  That way the reviewer can look at everything but the first commit to get the schema diff.

yao-msft: It's a copy of v1.4 schemas(with only 1.4.0 replaced by 1.5.0), maybe I could make the copy and update to1.5.0 separate commits

JohnMcPMS: No, if you aren't changing it to make future diffs easy, then I think just putting that in the PR initial comment would be helpful.

JohnMcPMS (approved):

PR #3256, opened 2023-05-16
Disable warnings in external projects

florelis:
Disabling compilation warnings from external projects as we're probably not going to do anything about them and they hide other warnings we do care about, like static analysis results.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3256)


yao-msft (approved):

JohnMcPMS (approved):
The hero we need.

PR #3260, opened 2023-05-18
Log settings errors

Trenly:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3260)

### DIFF
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -933,6 +933,11 @@ namespace AppInstaller::CLI
             if (!Settings::User().GetWarnings().empty() &&
                 !WI_IsFlagSet(command->GetOutputFlags(), CommandOutputFlags::IgnoreSettingsWarnings))
             {
+                auto warnings = Settings::User().GetWarnings();
+                for (int i=0; i<warnings.size(); i++) 
+                {
+                    AICLI_LOG( CLI, Warning, << warnings.at(i).Path << ": " << warnings.at(i).Message << warnings.at(i).Data );
+                }
                 context.Reporter.Warn() << Resource::String::SettingsWarnings << std::endl;
             }
 


src/AppInstallerCLICore/Command.cpp
```diff
@@ -933,6 +933,11 @@ namespace AppInstaller::CLI
             if (!Settings::User().GetWarnings().empty() &&
                 !WI_IsFlagSet(command->GetOutputFlags(), CommandOutputFlags::IgnoreSettingsWarnings))
             {
+                auto warnings = Settings::User().GetWarnings();
```
yao-msft, line 936: The settings parsing error/warning logging should happen during instance creation, not when it is being used. Please move the logic to end of `UserSettings::UserSettings(const std::optional<std::string>& content) : m_type(UserSettingsType::Default)` in UserSettings.cpp

Trenly: When I try doing that and check for `!m_warnings.empty()` it never writes anything to the log. Even if I just add something like `AICLI_LOG(Core, Info, << "THIS IS A TEST LOG ITEM");` at the end or right after the ValidateAll function, nothing gets logged?

yao-msft: OK, looks like first instance is created [here](https://github.com/microsoft/winget-cli/blob/56060c44be069bf65126d7c8b2f46795b850ca4e/src/AppInstallerCLICore/Core.cpp#L56) before logging is enabled [here](https://github.com/microsoft/winget-cli/blob/56060c44be069bf65126d7c8b2f46795b850ca4e/src/AppInstallerCLICore/Core.cpp#L69).

Trenly: > OK, looks like first instance is created [here](https://github.com/microsoft/winget-cli/blob/56060c44be069bf65126d7c8b2f46795b850ca4e/src/AppInstallerCLICore/Core.cpp#L56) before logging is enabled [here](https://github.com/microsoft/winget-cli/blob/56060c44be069bf65126d7c8b2f46795b850ca4e/src/AppInstallerCLICore/Core.cpp#L69).

Interesting. So none of the existing log statements in the settings initialization will show up either. 

I'll poke around and see if the logging initialization can be moved before the settings initialization.

yao-msft: > > OK, looks like first instance is created [here](https://github.com/microsoft/winget-cli/blob/56060c44be069bf65126d7c8b2f46795b850ca4e/src/AppInstallerCLICore/Core.cpp#L56) before logging is enabled [here](https://github.com/microsoft/winget-cli/blob/56060c44be069bf65126d7c8b2f46795b850ca4e/src/AppInstallerCLICore/Core.cpp#L69).
> 
> Interesting. So none of the existing log statements in the settings initialization will show up either.
> 
> I'll poke around and see if the logging initialization can be moved before the settings initialization.

In cli case on dev builds. Correct.

Trenly: Either I'm doing something wrong, or there's something else going on here. I tried removing that instance of settings just to see if the logging worked and it still did not. Then I tried explicitly setting the level to verbose and changing the level to the user settings only after the logging is enabled. Either way, I don't see any logging output from the settings being initialized. 

I've created a separate branch [(SettingLogDebug)](https://github.com/Trenly/winget-cli/commit/f9e9b21c7992a62898948e5fe9f03573532764de) so you can see what I tried without messing up the commit history on this PR. I can't seem to figure out what would be causing this bug

Issue #1625, opened 2021-10-22
Support installing APK into windows subsystem for android (WSA)


oxygen-dioxide:
### Description of the new feature / enhancement

Windows 11 supports windows subsystem for android now. I hope that we can install apk with winget.

### Proposed technical implementation details

Firstly, download apk from the download link provided in manifest.
Secondly, use adb (bundled with winget package) to install this apk.

W2725730722:
It's good

upintheairsheep:
> Tangentially related, but does anyone know if there's a way to install APKs on the subsystem without ADB right now?

I think via some filesystem hacks it is.

upintheairsheep:
@Trenly This issue deserves a pull request, it is relatively easy to implement:
1. Windows Subsystem for Android is installed
2. ADB must be installed on the machine
3. It must connect to the Subsystem, bypassing the developer mode restriction, which I think the uninstallation in control panel already has.


A nice feature in the manifest would be for adding the adb install flags that allow for stuff like test apps to be installed.

RokeJulianLockhart:
@oxygen-dioxide, why https://github.com/microsoft/winget-cli/issues/1625#event-12035696430?

oxygen-dioxide:
> @oxygen-dioxide, why [#1625 (comment)](https://github.com/microsoft/winget-cli/issues/1625#event-12035696430)?

Because WSA isn't supported by Microsoft anymore. Besides, even if WSA is still in support, I don't think it is a good idea to manage APKs in the same tool that manages windows applications now. It would be better to have another tool to manage Android apps.

PR #3267, opened 2023-05-24
Docs - Create #1625 - Android app support

upintheairsheep:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3267)


doc/specs/#1625 - Android app support
```diff
@@ -0,0 +1,113 @@
...
+
+Windows 11 or newer  
+Windows Subsystem for Android
+ADB (if it ends up being the method of installation)
+
+### Android YAML manifest
+
+The "InstallerType" field in the YAML manifest file specifies the type of the application. A new InstallerType "Android" or "APK" will be added to show that a given application is an Android app.
+
+Each PWA would have its own manifest YAML file. A sample manifest file would look like this:
```
Trenly, line 32: PWA or APK?

doc/specs/#1625 - Android app support
```diff
@@ -0,0 +1,113 @@
...
+Name: F-Droid
+Publisher: F-Droid
+AppMoniker: fdroid
+License: Test
+InstallerType: Android
+Installers:
+- Arch: neutral
+  Url: https://f-droid.org/F-Droid.apk
+  InstallerType: Android
+  MinAndroidPlat: 31
```
Trenly, line 46: `MinAndroidPlat` should be called out in the spec as a new key which needs to be added to the manifest. What validation needs to be in place on this key? Does it need to be at the installer level, or can it also be at root level? Where would it fall for APKs distributed in zip files?

doc/specs/#1625 - Android app support
```diff
@@ -0,0 +1,113 @@
...
+The "InstallerType" field in the YAML manifest file specifies the type of the application. A new InstallerType "Android" or "APK" will be added to show that a given application is an Android app.
+
+Each PWA would have its own manifest YAML file. A sample manifest file would look like this:
+
+```
+Id: FDroid.FDroid
+Version: 1337.0
+Name: F-Droid
+Publisher: F-Droid
+AppMoniker: fdroid
```
Trenly, line 39: What is `AppMoniker`?

doc/specs/#1625 - Android app support
```diff
@@ -0,0 +1,113 @@
...
+WinGetCreate will download the package and extract metadata from AndroidManifest.xml, automatically filling out all data excluding the short desciption
+
+### Package installation
+
+Either ADB could be used to install the application, or the method that Microsoft Store installs Android apps could be utilized.
+### Flow of the install process
+
+1. Windows Subsystem for Android is installed
+2. ADB must be installed on the machine
+3. It must connect to the Subsystem, bypassing the developer mode restriction, which I think the uninstallation in control panel already has.
```
Trenly, line 62: I don’t think winget should be bypassing restrictions. If the user has developer mode disabled it could be due to group policy or intune restrictions and sysadmin decisions need to be respected

doc/specs/#1625 - Android app support
```diff
@@ -0,0 +1,113 @@
...
+A nice feature in the manifest would be for adding the adb install [flags](https://adbshell.com/commands/adb-install) for installation of apps that require it.
+## UI/UX Design
+
+Installing a PWA will be similar to installing other application types currently supported by winget, using the install command. The -i flag will not be implemented due to it's impracticality.
```
Trenly, line 66: PWA or APK?

doc/specs/#1625 - Android app support
```diff
@@ -0,0 +1,113 @@
...
+
+# A field to determine minumum platform SDK version of Android is required for the app to run on like 27 (Oreo), 30 (Red Velvet Cake), 31 (Snow Cone), 32 (12L), or 33 (Tiramisu)
+# A field for ADB flags to be put, like test apps.
+### Accessibility
+
+This should have no direct impact on accessibility.
+
+### Security
+
+Android has security concerns.
```
Trenly, line 92: How exactly would this featue impact security, more specifically than the blanket statement that there are security concerns?

doc/specs/#1625 - Android app support
```diff
@@ -0,0 +1,113 @@
...
+
+### Performance, Power, and Efficiency
+
+Installing some or most Android apps leaves a service running in the background, for stuff such as notifications.
+## Potential Issues
+
+
+## Future considerations
+
+1. Android 14 "Upside Down Cake" will restrict the installation of apps that target any version before 6.0 "Marshmallow" due to security concerns, however --bypass-low-target-sdk-block can bypass this limitation. Similar, certain new Android devices, such as the Pixel 7 and Pixel Tablet, have no support for 32-bit apps and will not be able to run 32 bit code.
```
Trenly, line 108: I don’t understand what the Pixel 7 or Pixel Tablet have to do with running apps on WSA

Trenly (commented):
1) What would be the error message if user doesn’t have Windows Subsystem for Android enabled, or is using Windows 10?
2) What would be the error message if ADB isn’t installed?
3) What about android app bundles like AAB/APKS/APKM/XAPK?
4) What would the uninstall process look like?
5) What would the upgrade process look like?
6) Would `--scope machine` be applicable or inapplicable?
7) What considerations need to be made on the validation pipelines to test APKs? Would the signature of all APKs need to be validated? Would we need to validate a debug cert wasn’t used?
8) What happens when the package conflicts with an existing package of the same name when attempting to install?
9) How would these apps be handled by export

upintheairsheep:
For Windows 10, the error would say “You are using Windows 10. Installation of Android apps requires Windows 11 or higher.” and if not installed: “Windows Subsystem for Android is not currently installed, please install via running this command: COMMAND”. If the user has Windows 11 but no requirements: “Use of Android apps requires 8 GB of RAM or more. Please upgrade the RAM on your computer.”, for CPU “Use of Android apps is not supported on your CPU. Please purchase a new CPU that supports Android apps.” and if using HHD, “Your storage device is too slow to run Android apps. Please upgrade to an SSD.”
“ADB is not installed on your computer. Please run this command to install: COMMAND”
Yes. Other android package files would be supported, including .aab, .apks, .apkm, .xapk, .obb, and possibly .ab.
Uninstallation of applications would either invoke the control panel, or use the adb uninstall command, and the --preserve option in WinGet activates the -k flag in adb uninstall.
The upgrade process is more straightforward, just complete the same process, but use the -r flag to replace the app while keeping user data.
--scope machine may be applicable, we need to run this command as each user account, or make a system built inside of windows to install an android application on all users
If it works, it works. We may need to increase the amount of RAM or install WSA on Windows 11 on them. A debug certificate can be bypassed via adb install -t .
If it conflicts with an Android package, it gives a confirmation dialouge to replace/update the app with -r flag. I’m not really sure about Android apps conflicting with Windows apps though.
I’m not sure I understand what you mean by “export”
I’ll add WinGetCreate docs later.
AppMonkier is the app monkier
I’m saying to bypass the toggle when developer mode in WSA is off by the user. The group policy should be respected, and there could be a policy premitting winget to install android apps exclusively.
I used the PWA support docs as a template.
This may affect security due to malicious APKs via “http” protocol but all http sources should be checked for malware first.
It was just to avoid confusion between the two similar news

florelis:
@upintheairsheep can you rename the file to have .md extension so it's easier to view here, please?

florelis:
Also, as a suggestion, I think it may be easier to follow the PR if you reply to each comment separately in the conversation for each one:

image

upintheairsheep:
> items

I did.

upintheairsheep:
Hello? Is anyone there?

denelon:
This seems to be a pair of specifications with APK (Android Package Kit) and PWA (Progressive Web Application). I don't fully understand the WSA( Windows Subsystem for Android) well enough yet to tease all this apart yet. I think we'll need a bit more investigation on our side as WinGet can't install the WSA and potentially a "Store" for APKs to make them available.

I think there may also need to be some additional work to reason about how the upgrade scenario might work.

denelon:
I think we also may run into cases where there is a "Windows App" and an APK for the same package. We will likely need some settings/preferences to determine which to install in that case.  I know there are some existing issues. We should link to those so we can figure out the right sequencing here.

upintheairsheep:
> This seems to be a pair of specifications with APK (Android Package Kit) and PWA (Progressive Web Application). I don't fully understand the WSA( Windows Subsystem for Android) well enough yet to tease all this apart yet. I think we'll need a bit more investigation on our side as WinGet can't install the WSA and potentially a "Store" for APKs to make them available.
> 
> I think there may also need to be some additional work to reason about how the upgrade scenario might work.

I used the PWA documentation as a base for this issue. Any writings related to PWAs are outdated.

upintheairsheep:
> I think we also may run into cases where there is a "Windows App" and an APK for the same package. We will likely need some settings/preferences to determine which to install in that case. I know there are some existing issues. We should link to those so we can figure out the right sequencing here.

Just append “-android” to the identifier, and add “(Android)” to the name

upintheairsheep:
> > I think we also may run into cases where there is a "Windows App" and an APK for the same package. We will likely need some settings/preferences to determine which to install in that case. I know there are some existing issues. We should link to those so we can figure out the right sequencing here.
> 
> Just append “-android” to the identifier, and add “(Android)” to the name

“.Android” may work as well similar to how Geekbench versions are PrimateLabs.Geekbench.5 for instance

doc/specs/#1625 - Android app support.md
```diff
@@ -0,0 +1,113 @@
...
+Either ADB could be used to install the application, or the method that Microsoft Store installs Android apps could be utilized.
+### Flow of the install process
+
+1. Windows Subsystem for Android is installed
+2. ADB must be installed on the machine
+3. It must connect to the Subsystem, bypassing the developer mode restriction, which I think the uninstallation in control panel already has.
+A nice feature in the manifest would be for adding the adb install [flags](https://adbshell.com/commands/adb-install) for installation of apps that require it.
+## UI/UX Design
+
+Installing a PWA will be similar to installing other application types currently supported by winget, using the install command. The -i flag will not be implemented due to it's impracticality.
```
denelon, line 66:
```suggestion
```

doc/specs/#1625 - Android app support.md
```diff
@@ -0,0 +1,113 @@
...
+
+Android is by far the most used operating system on the planet, excluding a few certain regions, where iOS dominates. Many mobile developers make apps for Android, and only have a sometimes inferior web app for desktop, if they have one at all.
+
+## Solution Design
+
+### System requirements
+
+Windows 11 or newer  
+Windows Subsystem for Android
+ADB (if it ends up being the method of installation)
```
denelon, line 26:
```suggestion
ADB (Android Debug Bridge) (if it ends up being the method of installation)
```

doc/specs/#1625 - Android app support.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Android YAML manifest
+
+The "InstallerType" field in the YAML manifest file specifies the type of the application. A new InstallerType "Android" or "APK" will be added to show that a given application is an Android app.
+
+Each PWA would have its own manifest YAML file. A sample manifest file would look like this:
```
denelon, line 32:
```suggestion
The "InstallerType" enumeration would include "APK" to indicate an Andoroid package. A sample installer manifest file would look like the following:
```

doc/specs/#1625 - Android app support.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Solution Design
+
+### System requirements
+
+Windows 11 or newer  
+Windows Subsystem for Android
+ADB (if it ends up being the method of installation)
+
+### Android YAML manifest
```
denelon, line 28:
```suggestion
### Android Installer Type manifest
```

vedantmgoyal2009:
> “.Android” may work as well similar to how versions that act as different packages like Geekbench and Scratch are PrimateLabs.Geekbench.5 for instance. The installer type should not apply as the packages are separate.

Denelon want to say that you can do `winget install PrimateLabs.Geekbench.5 --installertype <apk|android>` and `winget install Publisher.Scratch --installertype <apk|android>`. Using `.Android` in package identifiers would make the directory structure more nested, which we may not want to do. We will also have update `search` command to search for android apps specifically.

upintheairsheep:
> > “.Android” may work as well similar to how versions that act as different packages like Geekbench and Scratch are PrimateLabs.Geekbench.5 for instance. The installer type should not apply as the packages are separate.
> 
> Denelon wants to say that you can do `winget install PrimateLabs.Geekbench.5 --installertype <apk|android>` and `winget install Publisher.Scratch --installertype <apk|android>`. (I would suggest using `android` because as Trenly said, there is more than one installer type for Android as well (AAB, XAPK, etc.) Using `.Android` in package identifiers would make the directory structure more nested, which we may not want to do. We will also have to update the `search` command to search for Android apps specifically.
> 
> Edit: Using `.Android` in package identifiers would make it troublesome to search only for Android packages. Example: `Google.AndroidStudio` or `blueedge.android11react`.

Why not just for example 3u.3utools-android

vedantmgoyal2009:
> Why not just for example 3u.3utools-android

It would result in locale metadata duplication, making it harder for moderators to ensure consistency b/w the Windows and Android manifests for the same package. Hence, adding a new installer in the same manifests containing Windows installers would be easier. I don't see any good reason to create a whole new package identifier for the same package, except in cases of where we need WinGet to behave in a particular manner.

upintheairsheep:
> > Why not just for example 3u.3utools-android
> 
> One of the reasons would be locale metadata duplication, making it harder for moderators to ensure consistency b/w the Windows and Android manifests for the same package. Hence, adding a new installer in the same manifests containing Windows installers would be easier. I don't see any good reason to create a whole new package identifier for the same package, except in cases where we need WinGet to behave in a particular manner.

upintheairsheep:
> > “.Android” may work as well similar to how versions that act as different packages like Geekbench and Scratch are PrimateLabs.Geekbench.5 for instance. The installer type should not apply as the packages are separate.
> 
> Denelon wants to say that you can do `winget install PrimateLabs.Geekbench.5 --installertype <apk|android>` and `winget install Publisher.Scratch --installertype <apk|android>`. (I would suggest using `android` because as Trenly said, there is more than one installer type for Android as well (AAB, XAPK, etc.) Using `.Android` in package identifiers would make the directory structure more nested, which we may not want to do. We will also have to update the `search` command to search for Android apps specifically.
> 
> Edit: Using `.Android` in package identifiers would make it troublesome to search only for Android packages. Example: `Google.AndroidStudio` or `blueedge.android11react`.

What about .Mobile?

upintheairsheep:
> > “.Android” may work as well similar to how versions that act as different packages like Geekbench and Scratch are PrimateLabs.Geekbench.5 for instance. The installer type should not apply as the packages are separate.
> 
> Denelon wants to say that you can do `winget install PrimateLabs.Geekbench.5 --installertype <apk|android>` and `winget install Publisher.Scratch --installertype <apk|android>`. (I would suggest using `android` because as Trenly said, there is more than one installer type for Android as well (AAB, XAPK, etc.) Using `.Android` in package identifiers would make the directory structure more nested, which we may not want to do. We will also have to update the `search` command to search for Android apps specifically.
> 
> Edit: Using `.Android` in package identifiers would make it troublesome to search only for Android packages. Example: `Google.AndroidStudio` or `blueedge.android11react`.

—installertype should only be for EXE, MSI, MSIX, APPX that are of the same version of the package.

vedantmgoyal2009:
Well, then a flag `--android` or `--only-android` can be added to `search` command, but the underlying searching algorithm would be based on InstallerType only.

upintheairsheep:
You know we can just add apps only available for Android for now, with no
Windows version available on that app. Seriously, how did this thread go
downhill?

upintheairsheep:
Merge?

yao-msft:
> Merge?

Hello @upintheairsheep, 

We usually merge spec after or near when implementation is done. There may be unknowns or changes needed during the actual implementation. So we'd like to leave this pr open for discussion and it could be extended to include future dev spec.

Thanks

upintheairsheep:
> Well, then a flag `--android` or `--only-android` can be added to `search` command, but the underlying searching algorithm would be based on InstallerType only.

Good idea. But would require the interactive only thing to be done.

upintheairsheep:
Additionally, it is important to note that not all Android apps have Windows or PWA versions of themselves.

denelon:
We're doing work on "InstallerType" in the client for filtering as a part of WinGet download. I'm sure that will accrue to the work here.

mominshaikhdevs:
I believe this should be closed since WSA is deprecated.

Issue #3026, opened 2023-02-28
Add support for icons via COM APIs


denelon:
### Description of the new feature / enhancement

Related to:
* #820 

I want to be able to expose icons via the COM API so a GUI can display package icons.

denelon:
Hey everyone,

We're working on supporting icons for the [Dev Home](https://github.com/microsoft/devhome) UI. We're going to start populating icons based on what we detect during validation. This will be part of the 1.5 release. As an FYI, the community repository will not accept that metadata in PRs, but it will be supported via the REST API for LOB (Line of Business) packages and it will also improve the experience in Intune and will be available for other MDM implementations.

denelon:
* https://github.com/microsoft/winget-cli/pull/3273

PR #3273, opened 2023-05-26
Add icons entries to yaml manifest, rest interface and com interface

yao-msft:
Added tests for yaml manifest, rest interface and com interface

### DIFF
--- a/src/AppInstallerRepositoryCore/Rest/Schema/1_5/Json/ManifestDeserializer_1_5.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/Schema/1_5/Json/ManifestDeserializer_1_5.cpp
@@ -1,0 +1,67 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "ManifestDeserializer.h"
+#include <winget/JsonUtil.h>
+
+using namespace AppInstaller::Manifest;
+
+namespace AppInstaller::Repository::Rest::Schema::V1_5::Json
+{
+    namespace
+    {
+        // Locale
+        constexpr std::string_view Icons = "Icons"sv;
+        constexpr std::string_view IconUrl = "IconUrl"sv;
+        constexpr std::string_view IconFileType = "IconFileType"sv;
+        constexpr std::string_view IconResolution = "IconResolution"sv;
+        constexpr std::string_view IconTheme = "IconTheme"sv;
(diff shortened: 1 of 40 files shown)


src/AppInstallerRepositoryCore/Rest/Schema/1_5/Json/ManifestDeserializer_1_5.cpp
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    Manifest::Icon iconEntry;
+
+                    iconEntry.Url = JSON::GetRawStringValueFromJsonNode(iconNode, JSON::GetUtilityString(IconUrl)).value_or("");
+                    if (iconEntry.Url.empty())
+                    {
+                        continue;
+                    }
+
+                    iconEntry.FileType = Manifest::ConvertToIconFileTypeEnum(JSON::GetRawStringValueFromJsonNode(iconNode, JSON::GetUtilityString(IconFileType)).value_or(""));
```
JohnMcPMS, line 45: Type was also required in the YAML; shouldn't this result in a "if (empty) { continue; }"?

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -885,7 +885,65 @@ namespace Microsoft.Management.Deployment
         String DocumentUrl { get; };
     }
 
-    /// IMPLEMENTATION NOTE: SourceAgreement from AppInstaller::Repository::SourceAgreement
+    /// Icon resolution
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 6)]
+    enum IconResolution
+    {
+        Unknown,
+        Custom,
```
JohnMcPMS, line 893: Isn't `Custom` just an official way of saying `Unknown`?  Should we get rid of `Unknown` and just return `Custom` instead?

JohnMcPMS (dismissed):

src/Microsoft.Management.Deployment/PackageManager.idl
yao-msft: Had offline discussion with John. From caller perspective, Custom or Unknown both indicates "caller, figure it out by yourself".

"But still, "known to to you but not to me" is no different than "known to no one" if my code needs to handle strange sizes"

JohnMcPMS (approved):

denelon:
* https://github.com/microsoft/winget-cli/issues/3026

PR #3296, opened 2023-06-01
Refresh process path variable when installing package dependencies

### LINKED ISSUE #2939: [Dependency Issue]: Dependency can't be found by main installer after it's installed
### Please confirm these before moving forward

- [X] I have searched for my issue and not found a work-in-progress/duplicate/resolved issue.
- [X] I have not been informed if the issue is resolved in a preview version of the winget client.

### Category of the issue

Other

This installer has Git as a dependency. It finds Git in the Windows Path during install to register itself. But the Windows Path in most Command Prompt and PowerShell windows appears to only update at window launch. One window must be opened for the dependency install, and then another window opened for the main package install. Observed in
* https://github.com/microsoft/winget-pkgs/pull/90133

Similar behavior was observed in this PR - the installer completed, but validation failed as it used the same PowerShell window, and thus couldn't find Git in the Windows Path.
* https://github.com/microsoft/winget-pkgs/pull/90517

### Steps to reproduce

For 90133 - On new system, modify the package's manifest to have Git.Git as a dependency, then install in a VM.

### Actual behavior

The installer will error "Could not find Git; can not register Git LFS."

### Expected behavior

Winget could help the client window to update its local Windows Path variable from the registry, so subsequent installers could install successfully, and programs run from the same window could find Git to execute successfully.

### Environment

```shell
Windows Package Manager (Preview) v1.4.3132-preview

Windows: Windows.Desktop v10.0.19044.1288
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.19.3132.0
```

### Screenshots and Logs

_No response_

ryfu-msft:
Fixes #2939

This PR adds support for refreshing the process PATH environment variable when installing multiple packages simultaneously. This means that modifications to the PATH variable by a package dependency should now be present for the main installer to invoke. 

Changes:
- Added `RefreshPathVariableForCurrentProcess()` method to refresh the path variable for the winget process between package installs
- Refactors the install flow so that installing dependencies are separated into their own workflow.
- Added a `--skip-dependencies` argument and a `ignoreDependencies` install behavior setting to skip the dependencies flow if needed.
- Fixed a bug with the IndexCreationTool to retry adding manifests to the test source if they fail. This can occur if a package dependency has not yet been added to the test source.
- Updated PackageManager.idl to contract version 7 and added a new installOption() for skipDependencies.

Tests:
- Added COM E2E test to verify skip dependencies install option behavior
- Modified the AppInstallerTestExeInstaller to invoke a command when provided with the argument option to emulate an installer that invokes a command from PATH.
- Added E2E test that installs a portable package dependency with a specific command, which is later invoked by the main installer.
- Added unit tests to verify `--skip-dependencies`, `ignoreDependencies`, and refresh path behavior for both user and system.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3296)


src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -536,28 +538,33 @@ namespace AppInstaller::CLI::Workflow
...
     }
 
     void InstallSinglePackage(Execution::Context& context)
     {
         context <<
             Workflow::CheckForUnsupportedArgs <<
-            Workflow::DownloadSinglePackage <<
+            Workflow::ReportIdentityAndInstallationDisclaimer <<
+            Workflow::ShowPromptsForSinglePackage(/* ensureAcceptance */ true) <<
+            Workflow::InstallDependencies <<
```
Trenly, line 566: If `--force` is provided but not `--skip-dependencies` and a dependency fails to install, will the context terminate or will it be a warning that a dependency failed to install, but force was specified so execution is continuing?

ryfu-msft: As of right now, the `--force` argument does not affect the package dependencies. It will just terminate if one of the package dependencies fails to install. `SkipDependencies` is a way to completely skip all of the dependencies for a package. I haven't put too much thought into how `--force` will play a role in this (future PR) but it would warn the user that a dependency failed to install, but continue attempting to install the rest of the dependencies as well as the main package.

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -81,4 +95,10 @@ namespace AppInstaller::Registry::Environment
         std::wstring pathName = std::wstring{ s_PathName };
         m_key.SetValue(pathName, ConvertToUTF16(value), REG_EXPAND_SZ);
     }
+
+    void RefreshPathVariableForCurrentProcess()
+    {
+        const auto& pathValue = PathVariable(Manifest::ScopeEnum::User, true).GetPathValue() + PathVariable(Manifest::ScopeEnum::Machine, true).GetPathValue();
+        THROW_HR_IF(E_UNEXPECTED, _putenv_s("PATH", pathValue.c_str()) != 0);
```
yao-msft, line 102: Use wide string version to support non ascii characters.

E_EXPECTED seems too strong here. Use a custom error code for this failure.

By the way, do we want to make this best effort? Or at least overridable with --force upon failure. Most of the cases, refresh env variable is not needed. Only a small portion of packages depend on refreshing dependency packages' newly written PATH.

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -81,4 +95,10 @@ namespace AppInstaller::Registry::Environment
         std::wstring pathName = std::wstring{ s_PathName };
         m_key.SetValue(pathName, ConvertToUTF16(value), REG_EXPAND_SZ);
     }
+
+    void RefreshPathVariableForCurrentProcess()
+    {
+        const auto& pathValue = PathVariable(Manifest::ScopeEnum::User, true).GetPathValue() + PathVariable(Manifest::ScopeEnum::Machine, true).GetPathValue();
```
yao-msft, line 101: I think we need to make sure the first value ends with ; before concatenating?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1939,4 +1939,10 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="CancellingOperation" xml:space="preserve">
     <value>Cancelling operation</value>
   </data>
+  <data name="SkipDependenciesArgumentDescription" xml:space="preserve">
+    <value>Skip dependencies</value>
```
yao-msft, line 1943: nit: I think WindowsFeatures also honoring this? *ArgumentDescription strings should be more descriptive

src/AppInstallerCLITests/Dependencies.cpp
```diff
@@ -206,4 +208,69 @@ TEST_CASE("DependencyNodeProcessor_NoMatches", "[dependencies]")
...
+
+std::string GetCurrentProcessPathVariable()
+{
+    size_t requiredSize;
+    getenv_s(&requiredSize, nullptr, 0, "PATH");
+
+    if (requiredSize > 0)
+    {
+        char* buffer = new char[requiredSize];
+        errno_t errorResult = getenv_s(&requiredSize, buffer, requiredSize, "PATH");
```
yao-msft, line 221: nit: almost always prefer the wide string version when calling OS apis

src/AppInstallerCLITests/Dependencies.cpp
```diff
@@ -206,4 +208,69 @@ TEST_CASE("DependencyNodeProcessor_NoMatches", "[dependencies]")
...
+        }
+
+        delete[] buffer;
+    }
+    return {};
+}
+
+TEST_CASE("RefreshEnvironmentVariable_User", "[dependencies]")
+{
+    std::string testPathEntry = "testUserPathEntry";
```
yao-msft, line 234: Since we are running all these tests on our dev machines, I don't think we would want these tests mess with our registry directly.

For now, maybe just guard it behind admin check.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -623,6 +630,8 @@ namespace AppInstaller::CLI::Workflow
                 installContext.SetTerminationHR(Workflow::HandleException(installContext, std::current_exception()));
             }
 
+            RefreshPathVariableForCurrentProcess();
```
yao-msft, line 633: This should only be called after each dependency package installation. InstallMultiplePackages will be used by upgrade -all, import etc which in this implementation, an unnecessary amount of path refresh will be performed.

src/AppInstallerCLICore/Workflows/InstallFlow.h
```diff
@@ -142,11 +142,11 @@ namespace AppInstaller::CLI::Workflow
     // Outputs: None
     void InstallPackageInstaller(Execution::Context& context);
 
-    // Downloads the installer for a single package. This also does all the reporting and user interaction needed.
+    // Installs the dependencies for a single package.
     // Required Args: None
     // Inputs: Manifest, Installer
     // Outputs: InstallerPath
-    void DownloadSinglePackage(Execution::Context& context);
+    void InstallDependencies(Execution::Context& context);
```
yao-msft, line 149: nit: ManageDependencies because we'll expand this to do download only later?

src/AppInstallerCLICore/Commands/COMCommand.cpp
```diff
@@ -21,7 +22,9 @@ namespace AppInstaller::CLI
             Workflow::ReportExecutionStage(ExecutionStage::Discovery) <<
             Workflow::SelectInstaller <<
             Workflow::EnsureApplicableInstaller <<
-            Workflow::DownloadSinglePackage;
+            Workflow::ReportIdentityAndInstallationDisclaimer <<
+            Workflow::ShowPromptsForSinglePackage(/* ensureAcceptance */ true) <<
+            Workflow::DownloadInstaller;
```
yao-msft, line 27: Do you leave out Dependencies part on purpose? I don't see why it's necessary for this pr. But if you feel you need to, add a TODO here to add back dependency processing for COM (and we probabl;y need to fix this TODO before next preview release?)

doc/Settings.md
```diff
@@ -90,6 +90,15 @@ The `portablePackageMachineRoot` setting affects the default root directory wher
     },
 ```
 
+### Ignore Dependencies
+The 'ignoreDependencies' behavior affects whether dependencies are installed for a given package. Defaults to 'false' if value is not set or is invalid.
```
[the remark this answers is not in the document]
ryfu-msft: I chose 'ignoreDependencies' because this is an install behavior setting. 'Skip' feels more like a one-time behavior so did not change this.

src/IndexCreationTool/Program.cs
```diff
@@ -62,11 +62,22 @@ static void Main(string[] args)
                     foreach (string includeDir in includeDirList)
                     {
                         var fullPath = Path.Combine(rootDir, includeDir);
-                        foreach (string file in Directory.EnumerateFiles(fullPath, "*.yaml", SearchOption.AllDirectories))
+                        Queue<string> filesQueue = new(Directory.EnumerateFiles(fullPath, "*.yaml", SearchOption.AllDirectories));
+                        while (filesQueue.Count > 0)
```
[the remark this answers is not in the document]
ryfu-msft: Created a second queue that catches all of the failed manifests and retries once more.

src/AppInstallerCommonCore/Public/winget/PathVariable.h
```diff
@@ -6,9 +6,11 @@
 
 namespace AppInstaller::Registry::Environment
 {
+    void RefreshPathVariableForCurrentProcess();
+
     struct PathVariable
     {
-        PathVariable(Manifest::ScopeEnum scope);
+        PathVariable(Manifest::ScopeEnum scope, bool readOnly = false);
```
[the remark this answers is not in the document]
ryfu-msft: Enforced readonly to Remove and Append methods

src/AppInstallerCommonCore/PathVariable.cpp
ryfu-msft: Added check to verify value ends with ; before concatenating

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
ryfu-msft: Added description for package dependencies and windows features

src/AppInstallerCLITests/Dependencies.cpp
ryfu-msft: changed to wide string version

src/AppInstallerCLITests/Dependencies.cpp
ryfu-msft: Added admin check to unit test.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
ryfu-msft: added a new parameter to InstallMultiplePackages that controls whether path is refreshed. This is only set to true for package dependencies and the default is set to false.

src/AppInstallerCLICore/Commands/COMCommand.cpp
ryfu-msft: I added back dependencies and added a TODO statement. Will address this refactoring in a separate PR

doc/Settings.md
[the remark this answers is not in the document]
ryfu-msft: Changed to 'skipDependencies' based on discussion

yao-msft (dismissed):

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -81,4 +107,16 @@ namespace AppInstaller::Registry::Environment
...
+
+    bool RefreshPathVariableForCurrentProcess()
+    {
+        std::string userPathValue = PathVariable(Manifest::ScopeEnum::User, true).GetPathValue();
+        std::string systemPathValue = PathVariable(Manifest::ScopeEnum::Machine, true).GetPathValue();
+
+        EnsurePathValueEndsWithSemicolon(userPathValue);
+        EnsurePathValueEndsWithSemicolon(systemPathValue);
+
+        std::wstring pathValue = ConvertToUTF16(userPathValue + systemPathValue);
```
yao-msft, line 119: per offline discussion, we need to expand the paths before setting them
Also per https://web.archive.org/web/20111225045158/http://support.microsoft.com/kb/100843/EN-US
It should be system path and append with user path

ryfu-msft: I added a new split function to split the path entries with the ';' character, then expanded each path entry and concatenated them together with the ';' character. Also verified that this behavior is necessary by modifying the path with %USERPROFILE%/testPath and executing a batch script in that path within the same refreshed process environment.

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -78,7 +117,22 @@ namespace AppInstaller::Registry::Environment
...
     }
+
+    bool RefreshPathVariableForCurrentProcess()
+    {
+        // Path values must be expanded before assigning to process environment for proper refresh.
+        std::string userPathValue = ExpandPathValue(PathVariable(Manifest::ScopeEnum::User, true).GetPathValue());
+        std::string systemPathValue = ExpandPathValue(PathVariable(Manifest::ScopeEnum::Machine, true).GetPathValue());
+
+        EnsurePathValueEndsWithSemicolon(userPathValue);
+        EnsurePathValueEndsWithSemicolon(systemPathValue);
```
yao-msft, line 133: This seems redundant

ryfu-msft: removed

src/AppInstallerCommonCore/PathVariable.cpp
ryfu-msft: Changed so that systempath comes first and is appended with user path.

yao-msft (approved):

PR #3299, opened 2023-06-02
Stub upgrade self

msftrubengu:
This is in preparation of the new AppInstaller bundle that will contain stub packages for all the application packages. The stub package doesn't contain any of the configuration binaries and it will be the preferred package installed inbox on Windows. 

To install the main package `winget configuration --enable` must be called. This will detect which type of package is installed and install the full package if needed. `winget configuration --disable` will install the stub package.

By design we are not supporting `winget configuration --file config.yaml --enable` because winget itself is updating. Running the same command automatically after the update can only be done by starting a new shell window losing context of the current shell and making scripts hard to implement.

The next PR will be to handle gracefully when deployment is telling us to shut down.

Based on [JohnMcPMS/stub-upgrade-self](https://github.com/JohnMcPMS/winget-cli/tree/stub-upgrade-self)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3299)


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -176,6 +180,8 @@ namespace AppInstaller::CLI
             return { type, "file"_liv, 'f' };
         case Execution::Args::Type::ConfigurationAcceptWarning:
             return { type, "accept-configuration-agreements"_liv };
+        case Execution::Args::Type::ConfigurationEnable:
```
Trenly, line 183: Enable also implies a `--disable` - would be a good thing to add probably?

src/AppInstallerCLICore/Commands/ConfigureCommand.cpp
```diff
@@ -30,8 +31,9 @@ namespace AppInstaller::CLI
     {
         return {
             // Required for now, make exclusive when history implemented
-            Argument{ Execution::Args::Type::ConfigurationFile, Resource::String::ConfigurationFileArgumentDescription, ArgumentType::Positional, true },
+            Argument{ Execution::Args::Type::ConfigurationFile, Resource::String::ConfigurationFileArgumentDescription, ArgumentType::Positional },
             Argument{ Execution::Args::Type::ConfigurationAcceptWarning, Resource::String::ConfigurationAcceptWarningArgumentDescription, ArgumentType::Flag },
+            Argument{ Execution::Args::Type::ConfigurationEnable, Resource::String::ConfigurationEnableMessage, ArgumentType::Flag, Argument::Visibility::Help },
```
Trenly, line 36: With this change, will there also be a GPO to disallow enabling configuration?

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -222,6 +222,8 @@ namespace AppInstaller
                 return "One or more queries did not return exactly one match";
             case APPINSTALLER_CLI_ERROR_PACKAGE_IS_PINNED:
                 return "The package has a pin that prevents upgrade.";
+            case APPINSTALLER_CLI_ERROR_PACKAGE_IS_STUB:
+                return "The package currently installed is the stub package";
```
Trenly, line 226: This should also be added to `doc/windows/package-manager/winget/returnCodes.md`

Trenly (commented):
This work seems to assume the user has the store available to them. It is possible to install winget on devices that do not have the Microsoft Store. Can some additional information be provided as to what the GitHub releases will look like? Will both the full and stub packages be available on GitHub, or just the full package? If the stub package is available outside the store and someone manually installs the dependencies, then `winget configure --enable` would fail on systems where the msstore is not present.  What about the case where msstore is present but the user has explicitly indicated they don't want to use it by removing it as a source? What if msstore is blocked by policy? 

Some documentation on the Failure Modes and Effects for this type of change would be helpful, since it seems there may be some unvalidated assumptions

src/AppInstallerCLICore/Commands/ConfigureCommand.cpp
msftrubengu: same as store

msftrubengu:
@trenly On GitHub we publish the MSIX bundle that contains the application and resources package, which is the same that is published to the store. A stub package is contained within the bundle, so to answer your question: the bundle will contain the full and stub package. Eventually, Windows will preinstall the stub package.

If store is not available, winget can be installed the same way as it is installed from the GitHub release. The Add-AppxPackage cmdlet have options to select which package to install. One can also use Add-AppxProvisionedPackage with the preferred stub preference.

There are things that still need to happen to enable all this and change to the PowerShell cmdlets will be made.

src/AppInstallerCLICore/Commands/ConfigureCommand.cpp
```diff
@@ -30,8 +31,10 @@ namespace AppInstaller::CLI
     {
         return {
             // Required for now, make exclusive when history implemented
```
JohnMcPMS, line 33: nit: Could delete comment since you are making it not required now.

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -212,72 +317,71 @@ namespace AppInstaller::CLI::Workflow
...
+        if (FAILED(hr))
         {
-            context.Reporter.Info() << Resource::String::InstallFlowInstallSuccess << std::endl;
+            AICLI_TERMINATE_CONTEXT(hr);
+        }
+#else
+        if (IsStubPackage())
+        {
+            context.Reporter.Info() << Resource::String::ConfigurationEnablingMessage << std::endl;
+            HRESULT hr = AppInstallerUpdate(context, false);
```
JohnMcPMS, line 350: What is the value of returning an `HRESULT` if you are just going to terminate the context if it failed?  Why are you converting these away from workflow tasks?

My suggestion was to convert the entire thing into back end code and then the front end would need to be written to watch it.  I don't think that must be done here (although I don't think I like the messages being output in the stub updates), but I do think it should either remain workflow tasks (`void F(Context&)`) or be completely separated (taking no `Context` at all and implemented in CommonCore).

JohnMcPMS (changes requested):

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -140,24 +156,113 @@ namespace AppInstaller::CLI::Workflow
...
+            AppInstallManager installManager;
+
+            if (!WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::BypassIsStoreClientBlockedPolicyCheck) && installManager.IsStoreBlockedByPolicyAsync(s_StoreClientName, s_StoreClientPublisher).get())
+            {
+                context.Reporter.Error() << Resource::String::MSStoreStoreClientBlocked << std::endl;
+                AICLI_LOG(CLI, Error, << "Store client is blocked by policy. MSStore execution failed.");
+                return APPINSTALLER_CLI_ERROR_MSSTORE_BLOCKED_BY_POLICY;
+            }
+
+            //if (!installManager.GetIsAppAllowedToInstallAsync(productId).get())
```
JohnMcPMS, line 238: Revert

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -36,7 +34,7 @@ namespace AppInstaller::CLI::Workflow
             }
 
             HRESULT errorCode = S_OK;
-            context.Reporter.ExecuteWithProgress(
+            reporter.ExecuteWithProgress(
```
JohnMcPMS, line 37: Take in the `IProgressCallback`, not the reporter.  That means moving the `ExecuteWithProgress` to the caller of this.

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -172,17 +168,15 @@ namespace AppInstaller::CLI::Workflow
             return Utility::LocIndString{ ssError.str() };
         }
 
-        HRESULT MSStoreUpdateImpl(Execution::Context& context, const std::wstring& productId, Manifest::ScopeEnum scope, bool force)
+        HRESULT MSStoreUpdateImpl(const std::wstring& productId, Manifest::ScopeEnum scope, Execution::Reporter& reporter, bool isSilentMode, bool force)
```
JohnMcPMS, line 171: Same: take in progress callback.

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -194,82 +188,62 @@ namespace AppInstaller::CLI::Workflow
...
-                context.Reporter.Error() << Resource::String::MSStoreAppBlocked << std::endl;
                 AICLI_LOG(CLI, Error, << "App is blocked by policy. MSStore execution failed. ProductId: " << Utility::ConvertToUTF8(productId));
                 return APPINSTALLER_CLI_ERROR_MSSTORE_APP_BLOCKED_BY_POLICY;
             }
 
             return S_OK;
         }
 
-        HRESULT AppInstallerUpdate(Execution::Context& context, bool preferStub)
+        void AppInstallerUpdate(bool preferStub, Execution::ContextFlag contextFlags, Execution::Reporter& reporter)
```
JohnMcPMS, line 223: Same

src/AppInstallerCLICore/Workflows/MSStoreInstallerHandler.cpp
```diff
@@ -316,42 +290,65 @@ namespace AppInstaller::CLI::Workflow
...
+            if (hr == APPINSTALLER_CLI_ERROR_MSSTORE_BLOCKED_BY_POLICY)
+            {
+                context.Reporter.Error() << Resource::String::MSStoreStoreClientBlocked << std::endl;
+            }
+            else if (hr == APPINSTALLER_CLI_ERROR_MSSTORE_APP_BLOCKED_BY_POLICY)
+            {
+                context.Reporter.Error() << Resource::String::MSStoreAppBlocked << std::endl;
+            }
+
+            THROW_HR(hr);
```
JohnMcPMS, line 339:
```suggestion
            AICLI_TERMINATE_CONTEXT(hr);
```

JohnMcPMS (changes requested):

JohnMcPMS (approved):

PR #3301, opened 2023-06-02
Treat warning not as error for CS8785

ryfu-msft:
Build pipeline had a recent update that started treating CS8785 as an error instead of warning. Modified .csproj file so that the warning does not break the build.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3301)


JohnMcPMS (approved):

src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
```diff
@@ -16,6 +16,8 @@
     <ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
     <!-- Disable warning for CS1591 for cswinrt auto-generated files -->
     <NoWarn>1591</NoWarn>
+    <!-- Treat warning not as error for CS8785 to avoid compilation errors with SourceGenerator -->
+    <WarningsNotAsErrors>CS8785</WarningsNotAsErrors>
```
yao-msft, line 20: can we use NoWarn above so it's in one place?

yao-msft (approved):

PR #3377, opened 2023-06-23
Support winget installing AppInstaller

msftrubengu:
## Change
When winget is updating AppInstaller (aka. itself) via `winget install 9NBLGGH4NNS1 -s msstore --force` if appears the installation failed but in reality, it got updated successfully.

```
winget install 9NBLGGH4NNS1 -s msstore --force
Verifying/Requesting package acquisition...
Starting package install...
  ███████████████████████████     90%
Failed to install or upgrade Microsoft Store package. Error code: 0x80073d02
```
The failure means `ERROR_PACKAGES_IN_USE` and from my last PR #3299 I believe that error will go away because of `--force`, but a new one will occur. I need a real AppInstaller package to confirm this but because reasons I cannot build it right now. Regardless, the correct fix it to listen to messages from the lifetime manager. 

When the system is updating an app, the app lifetime manager will ask any the app to terminate if its running. It does it by:
1. Sending a WM_QUERYENDSESSION message to any app window.
2. Sending a CTRL-C signal.

This means that winget.exe needs to start listening to window messages because there's no real difference between the app lifetime manager and a user doing CTRL-C. This PR creates a hidden window and start the message loop. When WM_QUERYENDSESSION is received, we will ask to cancel any progress (as it currently does with CTRL-C) stating the reason as AppShutdown. By design, we are not going to cancel installations as maybe they get completed before they terminate us. Any other operation will be cancelled.

If the context gets cancelled with an AppShutdown reason while installing AppInstaller we will just set the progress as 100% and be done with it as we are going to get terminated soon.
```mermaid
pie title Progress of AppInstaller installation by winget
    "Real" : 9
    "Fake" : 1
```

## Validation
I created a new `winget test appshutdown` command that will verify the window handle is correct and the app shutdown event is signaled. The E2E test will run this command and reregister wingetdev to generate the message as deployment would do. I still need a real AppInstaller package to make sure the flow works as I expect.
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3377&drop=dogfoodAlpha


src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -79,43 +127,136 @@ namespace AppInstaller::CLI::Execution
...
             }
             else
             {
-                CtrlHandler::Instance().RemoveContext(context);
+                SignalTerminationHandler::Instance().RemoveContext(context);
+            }
+        }
+
+        // This starts getting called by CreateWindow.
+        // If it's a member function in SignalTerminationHandler we will have problems.
```
JohnMcPMS, line 232: If it is a static function in `SignalTerminationHandler` it should work just fine, and not need to be so far from the usage nor pre-declared.

src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -242,5 +400,15 @@ namespace AppInstaller::CLI::Execution
     {
         return (m_shouldExecuteWorkflowTask ? m_shouldExecuteWorkflowTask(task) : true);
     }
+
+    HWND Context::GetWindowHandle()
```
JohnMcPMS, line 404: These should be wrapped in `#ifndef AICLI_DISABLE_TEST_HOOKS` as well.

JohnMcPMS (approved):

PR #3385, opened 2023-06-28
Allow usage of portable files that are not .exe (such as .jar).

seppeon:

-----
This is related to allowing support for this:
https://github.com/microsoft/winget-pkgs/pull/110767

The issue for this is:
https://github.com/microsoft/winget-cli/issues/3386

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3385&drop=dogfoodAlpha

### DIFF
--- a/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
@@ -32,7 +32,7 @@ namespace AppInstaller::CLI::Workflow
         }
 
         // Get the file extension to be used for the installer file.
-        std::wstring_view GetInstallerFileExtension(Execution::Context& context)
+        std::wstring_view GetInstallerFileExtension(std::wstring extension, Execution::Context& context)
         {
             const auto& installer = context.Get<Execution::Data::Installer>();
             switch (installer->BaseInstallerType)
@@ -41,8 +41,10 @@ namespace AppInstaller::CLI::Workflow
             case InstallerTypeEnum::Exe:
             case InstallerTypeEnum::Inno:
             case InstallerTypeEnum::Nullsoft:
-            case InstallerTypeEnum::Portable:
                 return L".exe"sv;
+            case InstallerTypeEnum::Portable:
+                // When a portable installer has an extension, use it, it might be a .jar, or portable executable file.
+                return not extension.empty() ? extension : L".exe"sv;
             case InstallerTypeEnum::Msi:
             case InstallerTypeEnum::Wix:
                 return L".msi"sv;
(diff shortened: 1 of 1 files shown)


seppeon:
I believe I need to change in PortableFlow.cpp too, will make that change then report results :)

vedantmgoyal2009:
We also need to add a supported JDK/JRE as Dependencies in the manifests, and also, the command will be `java -jar <path-to-jar> <parameters-of-jar-by-user>`, but `java -jar` should be configured internally so that users can run it like normal EXEs.

seppeon:
> We also need to add a supported JDK/JRE as Dependencies in the manifests, and also, the command will be `java -jar <path-to-jar> <parameters-of-jar-by-user>`, but `java -jar` should be configured internally so that users can run it like normal EXEs.

seppeon:
That would also mean that documentation given by vendors like PlantUML would not match usage after installing a jar package.

For example, PlantUML tell users to run their application like this:
`java -jar plantuml.jar sequenceDiagram.txt` [source](https://plantuml.com/starting)

If this was wrapped, the doco from these authors may not match.

Trenly:
I don’t think this should be merged, as it would allow things like .ps1 scripts to be injected as portable packages, which is something that should be disallowed.

Having any file extension be available as part of the portable flow is a bad idea in my opinion. If support for Jar files is needed, it should be added as an entirely separate installer type for `jar`

seppeon:
Right, that's true. I don't mind adding .jar specifically. What is the central concern about ps1 scripts? An executable can presumably achieve the same thing as that script?

Trenly:
> Right, that's true. I don't mind adding .jar specifically. What are the concerns about ps1 scripts? An executable can presumably achieve the same thing as that script?

Presumably, yes, but if an EXE is malicious, it is generally easier to detect than a .ps1 or .bat script executing arbitrary code

seppeon:
An installer can already package and install scripts which end up very similar to this outcome though?
Additionally, chocolatey, apt, pacman all allow scripts afaik.

denelon:
The "Installation Notes" displayed after an install "could" help the user understand the special case for a Java Archive as a portable package.

mominshaikhdevs:
what's the latest update on this?

whindsaks:
Relying on ShellExecute to be configured the same on all systems makes this pretty fragile. On my system, .jar opens in 7-zip by default...

denelon:
We're going to need a specification for things other than .exe as "portable" applications. Java archives are a "well known" type of portable application, but Windows doesn't natively have a mechanism to launch them so this would also mean that WinGet would need a reliable way to "install" them and make sure they correctly launch with an appropriate runtime.

sredna:
For Java, people should be using a .exe launcher anyway, it does all the work of finding a compatible Java install. Only .cmd and .ps1 can be assumed to work out of the box.

Issue #2804, opened 2022-12-27
Add way to specify MSIX details such as Package Family Name for AppsAndFeatures Entries


Trenly:
### Description of the new feature / enhancement

I ran into an oddity with the Microsoft Windows App SDK Runtime where the exe installer is just a wrapper for several MSIX packages. It would be nice if the details of the MSIX package, such as the PackageFamilyName, could be specified in the apps and features entries 

https://github.com/Trenly/winget-pkgs/blob/9cf38f7be8b2f18cab84cefed19b00d957c5e9db/manifests/m/Microsoft/WindowsAppRuntime/1/2/1.2.2/Microsoft.WindowsAppRuntime.1.2.installer.yaml#L12-L15

```
PS C:\Users\WDAGUtilityAccount\Desktop\winget-pkgs> Get-AppxPackage |?{$_.name -match 'microsoft' -and $_.name -match 'runtime'}|Select-Object -Property Name,Version,Architecture,PackageFamilyName

Name                                            Version         Architecture PackageFamilyName
----                                            -------         ------------ -----------------
Microsoft.WindowsAppRuntime.1.2                 2000.707.2303.0          X64 Microsoft.WindowsAppRuntime.1.2_8wekyb3d8bbwe
Microsoft.WindowsAppRuntime.1.2                 2000.707.2303.0          X86 Microsoft.WindowsAppRuntime.1.2_8wekyb3d8bbwe
MicrosoftCorporationII.WinAppRuntime.Main.1.2   2000.707.2303.0          X64 MicrosoftCorporationII.WinAppRuntime.Main.1.2_8wekyb3d8bbwe
MicrosoftCorporationII.WinAppRuntime.Singleton  2000.707.2303.0          X64 MicrosoftCorporationII.WinAppRuntime.Singleton_8wekyb3d8bbwe
Microsoft.WinAppRuntime.DDLM.2000.707.2303.0-x6 2000.707.2303.0          X64 Microsoft.WinAppRuntime.DDLM.2000.707.2303.0-x6_8wekyb3d8bbwe
Microsoft.WinAppRuntime.DDLM.2000.707.2303.0-x8 2000.707.2303.0          X86 Microsoft.WinAppRuntime.DDLM.2000.707.2303.0-x8_8wekyb3d8bbwe
```

PR #3391, opened 2023-06-29
Do not attempt post install ARP correlation if PackageFamilyName is provided and present for the user

JohnMcPMS:
## Change
We had the case of an EXE installer that would install an MSI on lower OS versions and an MSIX on higher ones.  This change attempts to improve the correlation experience as best as we can, although given the current greedy correlation strategy this requires some careful work in the manifest construction.  Even then you can't get the best case results, but this allows more options.  By changing the package name slightly, one can give up matching pre-existing installs of the MSI in exchange for detecting that the MSIX has been uninstalled outside of winget.

The simple change is to not attempt the confidence interval-based correlation after installing an installer that has a `PackageFamilyName` value present, when that family name is found to be registered for the user.

## Validation
A new E2E test is added that recreates the targeted flow.
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3391&drop=dogfoodAlpha


src/AppInstallerCLIE2ETests/TestData/Manifests/TestExeInstaller_InstallMSIX.yaml
```diff
@@ -0,0 +1,20 @@
...
+Name: EXE Installer that Installs MSIX - extra
+Version: 1.0.0.0
+Publisher: AppInstallerTest
+License: Test
+Installers:
+  - Arch: x86
+    Url: https://localhost:5001/TestKit/AppInstallerTestExeInstaller/AppInstallerTestExeInstaller.exe
+    Sha256: <EXEHASH>
+    InstallerType: exe
+    PackageFamilyName: 6c6338fe-41b7-46ca-8ba6-b5ad5312bb0e_8wekyb3d8bbwe
```
Trenly, line 11: Is this the intended way of doing the correlation? I thought that's what the AppsAndFeaturesEntries were for, so that the different InstallerType could be specified. In fact, this isn't the first time I've seen the issue where an exe installs MSIX packages - 
* https://github.com/microsoft/winget-cli/issues/2804
* https://github.com/Trenly/winget-pkgs/blob/9cf38f7be8b2f18cab84cefed19b00d957c5e9db/manifests/m/Microsoft/WindowsAppRuntime/1/2/1.2.2/Microsoft.WindowsAppRuntime.1.2.installer.yaml#L12-L15

If this is the intended way, that's great, I would only ask that it be documented somewhere that the PackageFamilyName may be used for correlation even though it is outside of the AppsAndFeaturesEntries, since based on the current manifest documentation, I would have thought that it wouldn't be used. And then, since the package behaves differently based on the OS Version, that there would be an installer node for each `MinimumOSVersion` that has a unique `AppsAndFeaturesEntries` to make the correlation unambiguous.

JohnMcPMS: `AppsAndFeaturesEntries` is for the Add/Remove Programs data in the Uninstall registry key.  It doesn't contain the package family name:
https://github.com/microsoft/winget-cli/blob/531d2f1677717c5c4d59f407011e6390f8589506/schemas/JSON/manifests/v1.5.0/manifest.singleton.1.5.0.json#L568-L600

Yes, the WinApp SDK / Runtime is what caused us to open that value up to installer types that were not definitely MSIX based.  Where would you suggest that it be documented that `Installer > PackageFamilyName` be used when an MSIX is installed?

The OS version targeting portion is relevant to the real manifest in question, not the test one.  I don't think it should be necessary if we can help it though.

Trenly: Ah, I see, that clarifies things. I think it can be documented at pkgs in the manifest docs; I can do that after this is merged.

Thank you for the explanation!

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -691,6 +691,15 @@ namespace AppInstaller::CLI::Workflow
             return;
         }
 
+        // If the installer claims to have a PackageFamilyName, and that family name is currently registered for the user,
+        // let that be the correlated item and skip any attempt at further ARP correlation.
+        auto installer = context.Get<Execution::Data::Installer>();
```
yao-msft, line 696: nit const auto& to avoid a copy?

yao-msft (dismissed):

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
JohnMcPMS: The fact that `auto` doesn't preserve the reference in the type constantly escapes my brain, no matter how many times I relearn it...

yao-msft (dismissed):

yao-msft (approved):

msftrubengu (approved):

Chico1244 (commented):
[]()

Chico1244 (commented):
[]()

PR #3395, opened 2023-06-30
Improve packaged test log collection and fix crash

JohnMcPMS:
## Change
Collect all of the logs from the E2E test runs, rather than just the logs from the last one of them run.
Re-enable crash dumps on E2E tests runs.
Collect logs and crash dumps from OOP configuration tests as well.

Fixes for the following issues:
1. Asynchronous functions in OOP configuration need to keep a reference to the object
  a. Done using `get_strong` at the top of each async function
2. Thread globals being set and then switching threads in `OpenConfigurationSetAsync` (via `co_await`)
  a. Done by not using `co_await` and instead polling the operation and cancelling it if we are cancelled
3. `Diagnostics` delegates need to keep a reference to the processor
  a. Done by capturing `get_weak` into the lambda that we pass along to the factory
4. Diagnostics failures can lead to a recursion and stack exhaustion
  a. Done by putting a recursive mutex around the diagnostics call and writing the error that must have occurred sending the diagnostic message to the debugger

Also makes the file logger files share read so that they can be opened while the server is still running.

## Validation
Upon attempting to repro locally, only Release builds would hit anything.  After these fixes, I ran the OOP configuration tests in a loop ~50 times without issue.
###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3395&drop=dogfoodAlpha

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -92,6 +92,7 @@ Dbg
 debian
 deigh
 deleteifnotneeded
+DENYWR
 desktopappinstaller
 devhome
 dic
@@ -465,6 +466,7 @@ websites
 WERSJA
 wesome
 wfopen
+wfsopen
 wgetenv
 Whatif
 winapifamily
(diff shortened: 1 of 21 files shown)


yao-msft (approved):
:shipit:

PR #3422, opened 2023-07-10
Disable RTTI

### LINKED ISSUE #2015: WindowsPackageManager.dll size could be shrunk significantly by disabling RTTI

The WindowsPackageManager.dll binary has C++ Run-Time Type Information (RTTI) enabled in many projects, but uses this in very few of them.  RTTI is unfortunately not pay-for-play so this means a lot of RTTI costs are sneaking into the binary for no benefit.

I locally tried disabling RTTI in all the places it could be, and was able to save ~400KB, or about 13% of this binary's on-disk size.

Seems like a nice easy win for disk space.

### Steps to reproduce

Build the binary, look at how big it is and how much RTTI data is in it using SizeBench ( https://aka.ms/SizeBench )

### Expected behavior

Shouldn't be so much RTTI, since it's not being used

### Actual behavior

---

### Environment

```shell
---
```

JohnMcPMS:
Fixes #2015

## Change
Disable RTTI (run-time type information) in almost all native projects.  The only uses of it were in `IPackage` and `ISource` dynamic casts, so explicit implementations were created for these to allow them to continue to function as before.  Helper templates allow replacing `dynamic_cast` with `PackageCast` and `std::dynamic_pointer_cast` with `SourceCast` directly.

The file size dropped ~800 KB on ~5000 KB, which is about a 15% reduction in size.

## Validation
Current tests should validate that we do not hit any new UB due to using a feature that should not work (although `std::dynamic_pointer_cast` actually fails to compile with RTTI disabled, so hopefully everything is covered in MS VC).
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3422&drop=dogfoodAlpha


src/AppInstallerCommonCore/AppInstallerCommonCore.vcxproj
```diff
@@ -251,6 +251,11 @@
       <EnablePREfast Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">true</EnablePREfast>
       <EnablePREfast Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">true</EnablePREfast>
       <EnablePREfast Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</EnablePREfast>
+      <RuntimeTypeInfo Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">false</RuntimeTypeInfo>
+      <RuntimeTypeInfo Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">false</RuntimeTypeInfo>
+      <RuntimeTypeInfo Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">false</RuntimeTypeInfo>
+      <RuntimeTypeInfo Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">false</RuntimeTypeInfo>
```
yao-msft, line 257: nit: duplicates here and above

yao-msft (approved):

src/AppInstallerCommonCore/AppInstallerCommonCore.vcxproj
JohnMcPMS: VS does what VS wants apparently...

yao-msft (approved):

PR #3435, opened 2023-07-13
Fix using expired cert in tests

msftrubengu:
StoreLeaf1.cer expired today. This is a band-aid until 3/22/24.

Real fix should be using certs that expire in a long long time, but I don't want to keep PRs broken.
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3435&drop=dogfoodAlpha


yao-msft:
Those expired certs are actually real store certs. And they update yearly.

yao-msft (approved):

PR #3448, opened 2023-07-18
Generate manifest for Winget Download

ryfu-msft:
Related to: #2953 

This PR adds support for generating a manifest file to go along with the winget download command. This is so that users can have a reference to the installer metadata. The generated manifest will only contain the installer node that corresponds to the downloaded installer. The yaml manifest will follow the same naming patter as the installer.

Changes:
- Added YamlWriter to handle processing all manifest fields and outputting the proper yaml string as well as outputting the string to a yaml file.
- Modified the download command flow to export the manifest file for the main installer as well as for all downloaded dependent installers.
- Added a unit test that utilizes the existing manifest validation tests to verify that the generated manifest matches the same parsed object as the test manifest. A Test hook was added to override the default manifest type to 'singleton' since 'merged' manifests are not accepted by validation.
- Updated E2E tests to also check that the manifest files were created.
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3448&drop=dogfoodAlpha


src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -569,7 +572,7 @@ void VerifyV1ManifestContent(const Manifest& manifest, bool isSingleton, Manifes
     auto installer1Dependencies = installer1.Dependencies;
     REQUIRE(installer1Dependencies.HasExactDependency(DependencyType::WindowsFeature, "PreviewIIS"));
     REQUIRE(installer1Dependencies.HasExactDependency(DependencyType::WindowsLibrary, "Preview VC Runtime"));
-    REQUIRE(installer1Dependencies.HasExactDependency(DependencyType::Package, "Microsoft.MsixSdkDepPreview"));
+    REQUIRE(installer1Dependencies.HasExactDependency(DependencyType::Package, "Microsoft.MsixSdkDepPreview", "1.0.0"));
```
yao-msft, line 575: I was wondering all manifest_good* manifests use the same validation code, and in this change, only 1.5 singleton is modified, and still the tests passed.

Then I found the implementation of HasExactDependency() is incorrect

```
                if (dependency.MinVersion) {
                    if (dependency.MinVersion.value() == minVersion)
                    {
                        return true;
                    }
                }
                else {
                    return true;
                }
```

should be 

```
if (!minVersion.empty())
{
      return minVersion == dependency.MinVersion.value();
}
else
{
     return true;
}                
```

And all v1.1 to v1.5 singleton and multifile manifests should be updated with this

src/AppInstallerCLITests/TestHooks.h
```diff
@@ -228,4 +234,19 @@ namespace TestHook
     private:
         std::vector<AppInstaller::Repository::ExtractedIconInfo> m_extractedIcons;
     };
+
+    struct SetExportedManifestTypeResult_Override
+    {
+        SetExportedManifestTypeResult_Override(AppInstaller::Manifest::ManifestTypeEnum manifestType) : m_manifestType(manifestType)
```
yao-msft, line 240: I don't feel this test override is needed. Just recreate the manifest and turn off schema validation, merged could be read correctly.

With this you were limiting the test to singleton only, while most of the manifests are not singleton

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -862,6 +865,28 @@ TEST_CASE("ValidateV1_5GoodManifestAndVerifyContents", "[ManifestValidation]")
...
+
+    TempDirectory downloadDirectory{ "downloadDirectory" };
+    std::filesystem::path testManifestPath = downloadDirectory.GetPath() / "testManifest.yaml";
+    const auto& manifestYamlContent = YamlWriter::ManifestToYamlString(singletonManifest, singletonManifest.Installers[0]);
+    YamlWriter::OutputYamlFile(manifestYamlContent, testManifestPath);
+
+    REQUIRE(std::filesystem::exists(testManifestPath));
+
+    Manifest generatedManifest = YamlParser::CreateFromPath(downloadDirectory, validateOption);
+    VerifyV1ManifestContent(generatedManifest, true, ManifestVer{ s_ManifestVersionV1_5 });
```
yao-msft, line 887: Test a multifile scenario as well (multi installers and multi locales)

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -375,6 +387,12 @@ namespace AppInstaller::Manifest
 
     std::string_view IconResolutionToString(IconResolutionEnum iconResolution);
 
+    std::string_view BoolToString(bool value);
```
yao-msft, line 390: nit: BoolToString seems generic enough to fit in AppInstallerStrings.cpp

src/AppInstallerCommonCore/Public/winget/ManifestYamlWriter.h
```diff
@@ -0,0 +1,23 @@
...
+    /// <param name="installer">Manifest installer object.</param>
+    /// <returns>Yaml string.</returns>
+    std::string ManifestToYamlString(const Manifest& manifest, const ManifestInstaller& installer);
+
+    /// <summary>
+    /// Exports the manifest yaml string to a file.
+    /// </summary>
+    /// <param name="content">Manifest yaml string.</param>
+    /// <param name="out">Path of the yaml file.</param>
+    void OutputYamlFile(const std::string& content, const std::filesystem::path& out);
```
yao-msft, line 22: This should be 
void OutputYamlFile(const Manifest& manifest, const ManifestInstaller& installer, const std::filesystem::path& out);

And set the yaml emitter to the file stream, so we don't need to hold a big string in memory. Then update the caller to use this method directly.

The above ManifestToYamlString could still be kept for future use. And both ManifestToYamlString and OutputYamlFile could use a common function which takes in the emitter

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+#include "winget/Yaml.h"
+#include "winget/ManifestYamlWriter.h"
+
+namespace AppInstaller::Manifest::YamlWriter
+{
+    using namespace Utility::literals;
+
+    namespace
+    {
+        constexpr Utility::LocIndView PackageIdentifier = "PackageIdentifier"_liv;
```
yao-msft, line 15: nit: we are not dealing with outputting to user here, soLocIndView is not needed just string_view is enough

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+#define WRITE_BOOL_PROPERTY(emitter, key, value) \
+        { \
+            emitter << YAML::Key << key << YAML::Value << BoolToString(value); \
+        }
+
+#define WRITE_PROPERTY_IF_EXISTS(emitter, key, value) \
+        { \
+            if (!value.empty()) \
+            { \
+                emitter << YAML::Key << key << YAML::Value << value; \
```
yao-msft, line 146:
```suggestion
                WRITE_PROPERTY(emitter, key, value) \
```

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+            }
+
+            out << YAML::EndMap;
+        }
+
+        void ProcessInstallerFields(YAML::Emitter& out, const ManifestInstaller& installer)
+        {
+            if (installer.Arch != Utility::Architecture::Unknown)
+            {
+                WRITE_PROPERTY(out, Architecture, Utility::ToLower(ToString(installer.Arch)));
```
yao-msft, line 537: nit: Arch is required, always output even unknown?

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+            }
+
+            WRITE_PROPERTY_IF_EXISTS(out, InstallerType, InstallerTypeToString(installer.BaseInstallerType));
```
yao-msft, line 540:
```suggestion
            WRITE_PROPERTY(out, InstallerType, InstallerTypeToString(installer.BaseInstallerType));
```

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+        }
+
+        void ProcessLocaleFields(YAML::Emitter& out, const ManifestLocalization& manifest)
+        {
+            ProcessAgreements(out, manifest.Get<Localization::Agreements>());
+            ProcessDocumentations(out, manifest.Get<Localization::Documentations>());
+            ProcessIcons(out, manifest.Get<Localization::Icons>());
+            ProcessSequence(out, Tags, manifest.Get<Localization::Tags>());
+
+            WRITE_PROPERTY_IF_EXISTS(out, PackageLocale, manifest.Locale);
```
yao-msft, line 235:
```suggestion
            WRITE_PROPERTY(out, PackageLocale, manifest.Locale);
```

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+    {
+        s_ExportedManifestTypeResult_TestHook_Override = manifestType;
+    }
+#endif
+
+    std::string YamlWriter::ManifestToYamlString(const Manifest& manifest, const ManifestInstaller& installer)
+    {
+        YAML::Emitter out;
+        out << YAML::BeginMap;
+        WRITE_PROPERTY_IF_EXISTS(out, PackageIdentifier, manifest.Id);
```
yao-msft, line 619:
```suggestion
        WRITE_PROPERTY(out, PackageIdentifier, manifest.Id);
```

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+        WRITE_PROPERTY_IF_EXISTS(out, PackageVersion, manifest.Version);
```
yao-msft, line 620:
```suggestion
        WRITE_PROPERTY(out, PackageVersion, manifest.Version);
```

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+        WRITE_PROPERTY_IF_EXISTS(out, Channel, manifest.Channel);
+        WRITE_PROPERTY_IF_EXISTS(out, Moniker, manifest.Moniker);
+        ProcessLocaleFields(out, manifest.DefaultLocalization);
+
+        // Handles default installer fields in the root node.
+        ProcessInstallerFields(out, manifest.DefaultInstallerInfo);
```
yao-msft, line 626: DefaultInstaller is not needed, default info is already populated to individual installer if applicable.

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+        ProcessInstaller(out, installer);
+        WRITE_PROPERTY_IF_EXISTS(out, ManifestVersion, manifest.ManifestVersion.ToString());
```
yao-msft, line 628:
```suggestion
        WRITE_PROPERTY(out, ManifestVersion, manifest.ManifestVersion.ToString());
```

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+        ManifestTypeEnum manifestType = ManifestTypeEnum::Merged;
+        
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        if (s_ExportedManifestTypeResult_TestHook_Override)
+        {
+            manifestType = *s_ExportedManifestTypeResult_TestHook_Override;
+        }
+#endif
+
+        WRITE_PROPERTY_IF_EXISTS(out, ManifestType, ManifestTypeToString(manifestType));
```
yao-msft, line 639:
```suggestion
        WRITE_PROPERTY(out, ManifestType, ManifestTypeToString(manifestType));
```

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,661 @@
...
+
+    std::string YamlWriter::ManifestToYamlString(const Manifest& manifest, const ManifestInstaller& installer)
+    {
+        YAML::Emitter out;
+        out << YAML::BeginMap;
+        WRITE_PROPERTY_IF_EXISTS(out, PackageIdentifier, manifest.Id);
+        WRITE_PROPERTY_IF_EXISTS(out, PackageVersion, manifest.Version);
+        WRITE_PROPERTY_IF_EXISTS(out, Channel, manifest.Channel);
+        WRITE_PROPERTY_IF_EXISTS(out, Moniker, manifest.Moniker);
+        ProcessLocaleFields(out, manifest.DefaultLocalization);
```
yao-msft, line 623: Also populate additional locales if any

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -674,7 +674,11 @@ namespace AppInstaller::CLI::Workflow
 
                 currentContext << Workflow::DownloadInstaller;
 
-                if (!downloadInstallerOnly)
+                if (downloadInstallerOnly)
+                {
+                    currentContext << Workflow::ExportManifest;
```
yao-msft, line 679: >currentContext << Workflow::ExportManifest; [](http://example.com/codeflow?start=20&length=43)

This seems duplicate, already done by Workflow::DownloadInstaller above

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
ryfu-msft: I left this as is. Some fields like the installer's package locale don't get populated in the root of the manifest so the unit test will fail.

src/AppInstallerCommonCore/Public/winget/ManifestYamlWriter.h
ryfu-msft: Changed to suggested.

src/AppInstallerCLITests/YamlManifest.cpp
ryfu-msft: Added tests for multifile and singleton scenario for all versions v1.0 - 1.6

src/AppInstallerCLITests/TestHooks.h
ryfu-msft: Removed test cook and turned off validation

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
```diff
@@ -0,0 +1,668 @@
...
+        else
+        {
+            THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_CANNOT_MAKE), !std::filesystem::is_directory(parentDirectory));
+        }
+
+        YAML::Emitter emitter;
+        PopulateManifestYamlEmitter(emitter, manifest, installer);
+
+        std::ofstream outFileStream(out);
+        outFileStream << emitter.str();
```
yao-msft, line 665: >outFileStream << emitter.str(); [](http://example.com/codeflow?start=8&length=31)

std::ofstream outFileStream(out);
emitter.emit(outFileStream);
outFileStream.close();

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -886,6 +888,198 @@ TEST_CASE("ValidateV1_6GoodManifestAndVerifyContents", "[ManifestValidation]")
     VerifyV1ManifestContent(mergedManifest, false, ManifestVer{ s_ManifestVersionV1_6 });
 }
 
+TEST_CASE("WriteV1SingletonManifestAndVerifyContents", "[ManifestCreation]")
+{
+    ManifestValidateOption validateOption;
```
yao-msft, line 893: >ManifestValidateOption validateOption; [](http://example.com/codeflow?start=4&length=38)

nit: this is not needed if we are to use the defaults.

Same applies to all below new tests added

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
yao-msft: I think we should update the test to ignore DefaultInstallerInfo check for the exported manifest case.

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -886,6 +888,198 @@ TEST_CASE("ValidateV1_6GoodManifestAndVerifyContents", "[ManifestValidation]")
...
+        "ManifestV1-MultiFile-Locale.yaml" }, multiFileDirectory);
+
+    Manifest multiFileManifest = YamlParser::CreateFromPath(multiFileDirectory, validateOption);
+    TempDirectory exportedMultiFileDirectory{ "exportedMultiFile" };
+    std::filesystem::path generatedMultiFileManifestPath = exportedMultiFileDirectory.GetPath() / "testMultiFileManifest.yaml";
+    YamlWriter::OutputYamlFile(multiFileManifest, multiFileManifest.Installers[0], generatedMultiFileManifestPath);
+
+    REQUIRE(std::filesystem::exists(generatedMultiFileManifestPath));
+    Manifest generatedMultiFileManifest = YamlParser::CreateFromPath(exportedMultiFileDirectory, validateOption);
+    VerifyV1ManifestContent(generatedMultiFileManifest, true);
```
yao-msft, line 920: >VerifyV1ManifestContent [](http://example.com/codeflow?start=4&length=23)

The isSingleton should be false and additional locales should be verified. Regarding installer, we could have another bool called isExport then skip defaultInstallerInfo and additional installer if needed.

Ideally, the manifest export code should support exporting a single installer and exporting all installers if no specific installer is passed in. But I'm ok not doing exporting all installers case in this change since it's not used anywhere yet.

src/AppInstallerCommonCore/Manifest/YamlWriter.cpp
ryfu-msft: Changed to ignoring default installer info as discussed.

src/AppInstallerCLITests/YamlManifest.cpp
ryfu-msft: removed validate option for these tests.

src/AppInstallerCLITests/YamlManifest.cpp
ryfu-msft: Added a TODO comment for outputting all installers if no single installer is present. 

Updated tests to behave differently if the manifest is exported.

yao-msft (approved):
:shipit:

PR #3451, opened 2023-07-19
Add file logger to the statics object creation

JohnMcPMS:
## Change
Add a file logger for the factory creation (and anything else we wanted to log in the statics object), which isn't covered by any other logging/diagnostics.

## Validation
File is created.

### DIFF
--- a/src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
+++ b/src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
@@ -12,6 +12,7 @@
 #include <winget/ILifetimeWatcher.h>
 #include <winget/GroupPolicy.h>
 #include <winget/Security.h>
+#include <winget/ThreadGlobals.h>
 
 namespace ConfigurationShim
 {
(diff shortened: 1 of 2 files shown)


src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
```diff
@@ -22,11 +24,32 @@ namespace ConfigurationShim
...
     DECLSPEC_UUID(WINGET_OUTOFPROC_COM_CLSID_ConfigurationStaticFunctions)
     ConfigurationStaticFunctionsShim : winrt::implements<ConfigurationStaticFunctionsShim, winrt::Microsoft::Management::Configuration::IConfigurationStatics>
     {
-        ConfigurationStaticFunctionsShim() = default;
+        ConfigurationStaticFunctionsShim()
+        {
+            auto threadGlobalsRestore = m_threadGlobals.SetForCurrentThread();
+            auto& diagnosticsLogger = m_threadGlobals.GetDiagnosticLogger();
+            diagnosticsLogger.EnableChannel(AppInstaller::Logging::Channel::All);
+            diagnosticsLogger.SetLevel(AppInstaller::Logging::Level::Verbose);
```
yao-msft, line 50: Do we need to call ConvertLevel above to set corresponding level? The ConvertLevel does not seem to be called in this change.

yao-msft (approved):

Issue #3456, opened 2023-07-20
winget list --upgrade-available suggests "Use --include-unknown to see all results"


TonyWhitley:
### Brief description of your issue

e.g.
`winget list --upgrade-available`
`No installed package found matching input criteria.`
`5 package(s) have version numbers that cannot be determined. Use --include-unknown to see all results.`
but
`winget list --upgrade-available --include-unknown`
results in
```
Windows Package Manager v1.5.1881
Copyright (c) Microsoft Corporation. All rights reserved.

Argument name was not recognized for the current command: '--include-unknown'

### Steps to reproduce

winget list --upgrade-available --include-unknown


### Expected behavior

That it would list the unknown packages as well

### Actual behavior

An error message

### Environment

```shell
Windows Package Manager v1.5.1881
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22621.2066
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.1881.0
```

stephengillie:
Some packages don't write version data to the Windows Registry, and so the WinGet package manager doesn't have version data to know if they need an upgrade.

denelon:
Do you get the same output with `winget upgrade`?

TonyWhitley:
Good question, no, the unknown ones are listed and will update (again, obviously winget can't tell they are already up to date)
```
winget upgrade --accept-package-agreements --accept-source-agreements --all --silent --include-unknown
Name                            Id                        Version Available Source
----------------------------------------------------------------------------------
Erlang OTP 24.0 (12.0)          Erlang.ErlangOTP          Unknown 26.0.1    winget
FlightGear v2019.1.1            FlightGear.FlightGear     Unknown 2020.3.18 winget
SpeedFan (remove only)          Almico.SpeedFan           Unknown 4.52      winget
FotoSketcher 3.80 (64-bit)      DavidTHOIRON.FotoSketcher Unknown 3.80      winget
MSI Kombustor 4.1.20.0 (64-bit) MSI.Kombustor.4           Unknown 4.1.20.0  winget
5 upgrades available.

TonyWhitley:
I note that 
`winget list --help`
does not mention --include-unknown (fair enough) so I suppose the bug is that `winget list --upgrade-available` shouldn't suggest  `Use --include-unknown to see all results.` (though it would be useful thing to add to the `list` command).

denelon:
[...]

PR #3473, opened 2023-07-25
Allow `--include-unknown` in list `--upgrade-available`

florelis:
Allow using `--include-unknown` and `--include-pinned` when using `winget list --upgrade-available`. Also block those args from being used without `--upgrade-available`

Solves #3456
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3473&drop=dogfoodAlpha

### DIFF
--- a/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
+++ b/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
@@ -1328,6 +1328,10 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="IncludeUnknownArgumentDescription" xml:space="preserve">
     <value>Upgrade packages even if their current version cannot be determined</value>
   </data>
+  <data name="IncludeUnknownArgumentInListDescription" xml:space="preserve">
+    <value>List packages even if their current version cannot be determined. Can only be used with the --upgrade argument</value>
+    <comment>{Locked="--upgrade"}</comment>
+  </data>
   <data name="UpgradeUnknownVersionExplanation" xml:space="preserve">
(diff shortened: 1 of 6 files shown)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1328,6 +1328,10 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="IncludeUnknownArgumentDescription" xml:space="preserve">
     <value>Upgrade packages even if their current version cannot be determined</value>
   </data>
+  <data name="IncludeUnknownInListArgumentDescription" xml:space="preserve">
+    <value>List packages even if their current version cannot be determined. Can only be used with the --upgrade argument</value>
```
yao-msft, line 1332:
```suggestion
    <value>List packages even if their current version cannot be determined. Can only be used with the --upgrade-available argument</value>

```

And also update the locked="" below

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1689,6 +1697,10 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="IncludePinnedArgumentDescription" xml:space="preserve">
     <value>Upgrade packages even if they have a non-blocking pin</value>
   </data>
+  <data name="IncludePinnedInListArgumentDescription" xml:space="preserve">
+    <value>List packages even if they have a pin that prevents upgrade. Can only be used with the --upgrade argument</value>
```
yao-msft, line 1701: --upgrade-available here and the locked

yao-msft (dismissed):

yao-msft (approved):

Issue #3165, opened 2023-04-18
Reboot resilience


denelon:
### Description of the new feature / enhancement

I'd like to have WinGet be resilient to reboots.

Some packages require a reboot when they are updated. If I was in the middle of a `winget import packages.json` or `winget install <package 1> <package 2> <package 3>` or installing a package with a dependency that triggered a reboot, I'd like WinGet to be able to pick up where it left off.

Related to:
* #163

solomoncyj:
have winget after parsing input, create a temp a file in %temp% should an intaller restart the system, have winget run at startup to verify that all packages in the temp file have been instlled. if there are pacages left, resume install if not, exit

PR #3508, opened 2023-08-07
Add `resume` command and support saving the argument state.

ryfu-msft:
Related to:
#3165 

Changes:

1. Add a new `Resume` command that takes in a `--resume-id; -g` argument that points to a resume state.

2. Created a Checkpoint Index that creates/interacts with the following tables:
- CheckpointMetadataTable (Responsible for tracking important information about a resume i.e. command, args, clientVersion)
- CheckpointContextTable (Records the checkpoint and all relevant context data. Will be extended in a future PR to handle more complex context data)

3. The resume command will create a brand new context and initialize the CheckpointManager with the provided --resume-id argument. The CheckpointManager will be utilized to interface with the checkpointIndex and retrieve information for calling the correct command.  
4. Added a `Resume` function that an individual command can override if they have command-specific steps for resuming.

Tests:
Add unit tests to verify the following scenarios:
- Invalid resume id
- Guid resume state does not exist
- Resume state is empty and does not have any values.
- Successful install should clean up index.
- Failed install should not clean up index.

Added E2E tests to verify:
- Successful install cleans up checkpoint index after completion.
- Failed install leaves behind a checkpoint save file that can be called using the resume command to replicate the same outcome.
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3508&drop=dogfoodAlpha


src/AppInstallerCLICore/Commands/ResumeCommand.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+        THROW_HR_IF_MSG(E_UNEXPECTED, !commandToResume, "Command to resume not found.");
+
+        resumeContext.SetExecutingCommand(commandToResume.get());
+        resumeContext.SetTargetCheckpoint(checkpointManager.GetLastCheckpoint(rootContextId));
+        resumeContext.SetFlags(Execution::ContextFlag::Resume);
+
+        checkpointManager.Checkpoint(resumeContext, Execution::CheckpointFlag::CommandArguments);
+        
+        commandToResume->Execute(resumeContext);
```
Trenly, line 82: Because the command to be resumed is actually calling the command itself, I presume this already handles a change in state on the machine? i.e - User started a command with experimental feature for dependencies enabled, first dependency spawned a reboot, upon reboot, dependencies experimental feature was *magically* disabled by some script that ran during the reboot, and then the `winget resume` happens;  Would it fail due to a change in state? Proceed but just not install further dependencies? What if an argument was used originally that was behind an experimental feature and then was disabled? I think that would cause it to error out and show the "need to enable experimental feature for this", but would that also remove the resume from the checkpoint index?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2059,4 +2059,30 @@ Please specify one of them using the --source option to proceed.</value>
...
     <value>Installer is prohibited from being downloaded for later offline installation. </value>
   </data>
+  <data name="ResumeCommandLongDescription" xml:space="preserve">
+    <value>Resumes execution of a previously saved command by passing in the unique guid identifier of the saved command. This is used to resume an executed command that may have been terminated due to a reboot.</value>
+  </data>
+  <data name="ResumeCommandShortDescription" xml:space="preserve">
+    <value>Resumes execution of a previously saved command.</value>
+  </data>
+  <data name="ResumeGuidArgumentDescription" xml:space="preserve">
+    <value>The unique guid identifier of the saved state to resume</value>
```
Trenly, line 2069: `unique guid identifier` is redundant. Suggest switching to the full definition or removing `guid`
```suggestion
    <value>The globally unique identifier (guid) of the saved state to resume</value>
    <comment>{Locked="guid"} 'guid' is the short name form of a globally unique identifier.</comment>
```

```suggestion
    <value>The unique identifier of the saved state to resume</value>
```

src/AppInstallerCommonCore/Public/AppInstallerRuntime.h
```diff
@@ -49,7 +49,7 @@ namespace AppInstaller::Runtime
         SelfPackageRoot,
         // The location where user downloads are stored.
         UserProfileDownloads,
-        // Always one more than the last path; for being able to iterate paths in tests.
+        // The location where checkpoint database files are stored.
```
Trenly, line 52: Looks like you might have forgotten to add a runtime pathname?

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,221 @@
...
+    CheckpointManager::~CheckpointManager()
+    {
+        CleanUpIndex();
+    }
+
+    void CheckpointManager::SaveCheckpoint(Execution::Context& context, Execution::CheckpointFlag flag)
+    {
+        switch (flag)
+        {
+        case Execution::CheckpointFlag::CommandArguments:
```
JohnMcPMS, line 139: I think that the arguments should always be stored in the database metadata, so it seems unnecessary to have a flag to specify them.

Also, using a flag to specify what to store isn't particularly scalable here.  Instead, my suggestion is to use a list of context data items that are relevant at the checkpoint.

So the set of data to store is:
1. Automatic metadata: command, command line, COM info, whatever we want or need for debugging or potential resume use.
2. Checkpoint identifier: I would make this a checkpoint name rather than an enum, because a string is fine here and will likely be easier to debug.  The only downside is the potential for collisions, but if we enforce a descriptive name it should help with debugging and reduce the slim chance of that.
3. Execution context data items: Every item that was requested to be stored.  Additionally, I would keep all data from every checkpoint as the workflow progresses for debugging purposes.

The callers would `context.Checkpoint("MyCheckpoint", { Data1, Data2 });` and would be able to inspect the set of checkpoints during resumption and load the one they know about.

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    void CheckpointManager::Checkpoint(Execution::Context& context, Execution::CheckpointFlag checkpointFlag)
+    {
+        Execution::CheckpointFlag contextCheckpointFlag = context.GetTargetCheckpoint();
+
+        // If the context target checkpoint is ahead or equal to the current checkpoint, we have previously executed this state, load checkpoint state from index.
+        // If the context target checkpoint is behind the current checkpoint, we have not yet processed this state, save checkpoint state to index.
+        if (contextCheckpointFlag > checkpointFlag || contextCheckpointFlag == checkpointFlag)
+        {
+            LoadCheckpoint(context, checkpointFlag);
```
JohnMcPMS, line 54: This makes me think that you are attempting to design this where the command is not cooperating in the resume, but is simply calling a checkpoint function every so often.  From our discussions on this, I thought we agreed that that design was not going to work.

src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -103,6 +105,16 @@ namespace AppInstaller::CLI
 
     void InstallCommand::ExecuteInternal(Context& context) const
     {
+        bool resumeExperimentalFeatureEnabled = Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::Resume);
+
+        if (resumeExperimentalFeatureEnabled && WI_IsFlagClear(context.GetFlags(), Execution::ContextFlag::Resume))
+        {
+            auto& checkpointManager = CheckpointManager::Instance();
```
JohnMcPMS, line 112: This code should not be in an individual command.  I think that the manager should live in the context (and certainly not be a singleton).

JohnMcPMS (changes requested):
We should probably talk again.

src/AppInstallerRepositoryCore/Microsoft/Schema/Checkpoint_1_0/CheckpointArgumentsTable.cpp
```diff
@@ -0,0 +1,253 @@
...
+#include "CheckpointArgumentsTable.h"
+#include "SQLiteStatementBuilder.h"
+#include "Microsoft/Schema/ICheckpointIndex.h"
+
+namespace AppInstaller::Repository::Microsoft::Schema::Checkpoint_V1_0
+{
+    using namespace SQLite;
+    using namespace std::string_view_literals;
+
+    static constexpr std::string_view s_CheckpointArgumentsTable_Table_Name = "CheckpointArguments"sv;
```
Trenly, line 13: Instead of having a column for each argument type, wouldn't it make more sense to have an `ArgumentName` table which associates a unique numerical ID to each argument type, then use a many-to-one architecture with [dynamic typing](https://sqlite.org/faq.html#q3) for storing the arguments? It would reduce the amount of space needed in memory to not have to have empty columns for every argument, especially when arguments aren't supported by certain commands

image

Trenly (commented):
> Will be extended in a future PR to handle more complex context data

I presume this relates to sub-contexts, multi-installs, etc?

src/AppInstallerCLICore/Commands/ResumeCommand.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+        GUID checkpointId = Utility::ConvertToGuid(resumeGuidString);
+
+        if (!std::filesystem::exists(AppInstaller::Repository::Microsoft::CheckpointIndex::GetCheckpointIndexPath(checkpointId)))
+        {
+            context.Reporter.Error() << Resource::String::ResumeIdNotFoundError(Utility::LocIndView{ resumeGuidString }) << std::endl;
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_RESUME_ID_NOT_FOUND);
+        }
+
+        Execution::Context resumeContext{ std::cout, std::cin };
```
JohnMcPMS, line 55: Might have to make a new clone-type method on `Context` to properly clone the in/out streams from the incoming context.  But definitely need to inherit them rather than using cout/cin directly.

src/AppInstallerCLICore/Commands/ResumeCommand.cpp
```diff
@@ -0,0 +1,91 @@
...
+        auto previousThreadGlobals = resumeContext.SetForCurrentThread();
+        resumeContext.EnableSignalTerminationHandler();
```
JohnMcPMS, line 57: You should do this as the last thing before calling `Resume`.  I would say that you are still executing the resume command until you transfer to it.

src/AppInstallerCLICore/Commands/ResumeCommand.cpp
```diff
@@ -0,0 +1,91 @@
...
+        resumeContext.InitializeCheckpointManager(checkpointId);
+
+        const auto& checkpointClientVersion = resumeContext.GetClientVersionFromCheckpoint();
+        if (checkpointClientVersion != AppInstaller::Runtime::GetClientVersion().get())
+        {
+            context.Reporter.Error() << Resource::String::ClientVersionMismatchError(Utility::LocIndView{ checkpointClientVersion }) << std::endl;
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_CLIENT_VERSION_MISMATCH);
+        }
+
+        std::string commandName = resumeContext.GetCommandNameFromCheckpoint();
```
JohnMcPMS, line 67: Can we just have all of the new methods be from the `CheckpointManager` and not add so many helpers to the context?

src/AppInstallerCLICore/Commands/ResumeCommand.cpp
```diff
@@ -0,0 +1,91 @@
...
+        std::unique_ptr<Command> commandToResume;
+
+        // Find the command using the root command.
+        AICLI_LOG(CLI, Info, << "Resuming command: " << commandName);
+        for (auto& command : std::make_unique<RootCommand>()->GetCommands())
+        {
+            if (
+                Utility::CaseInsensitiveEquals(commandName, command->Name()) ||
+                Utility::CaseInsensitiveContains(command->Aliases(), commandName)
```
JohnMcPMS, line 76: You shouldn't need to look at the aliases; the full name should always be the one saved to the database.

src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -407,4 +409,57 @@ namespace AppInstaller::CLI::Execution
...
+    {
+        // TODO: Implementation for capturing each context data specified in the provided vector.
+        UNREFERENCED_PARAMETER(checkpointName);
+        UNREFERENCED_PARAMETER(contextData);
+    }
+
+    std::string Context::GetCommandLineFromArgs()
+    {
+        std::string commandLine;
+        for (auto type : Args.GetTypes())
```
JohnMcPMS, line 433: Hopefully this is just old and will be replaced with storing/loading the entire string directly.

src/AppInstallerCLICore/Workflows/ResumeFlow.h
```diff
@@ -0,0 +1,28 @@
...
+    // Required Args: None
+    // Inputs: ResumeGuid
+    // Outputs: None
+    void EnsureSupportForResume(Execution::Context& context);
+
+    // Applies a checkpoint to the context workflow.
+    struct Checkpoint : public WorkflowTask
+    {
+        Checkpoint(std::string_view checkpointName, std::vector<Execution::Data> contextData) :
+            WorkflowTask("ApplyCheckpoint"),
```
JohnMcPMS, line 18: Please use the same name as the type to make things easier.

src/AppInstallerCLICore/Workflows/ResumeFlow.h
```diff
@@ -0,0 +1,28 @@
...
+            m_checkpointName(checkpointName),
+            m_contextData(contextData) {}
```
JohnMcPMS, line 20: move

src/AppInstallerCLICore/Workflows/ResumeFlow.cpp
```diff
@@ -0,0 +1,25 @@
...
+    void Checkpoint::operator()(Execution::Context& context) const
+    {
+        if (!Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::Resume))
+        {
+            return;
+        }
+
+        if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::Resume))
+        {
+            // TODO: If the resume flag is set, resume workflow behavior.
```
JohnMcPMS, line 17: Since the `operator<<` overloads will need to know when to skip tasks (assuming that is the path chosen), it seems better to let them also handle knowing when to stop skipping tasks.  Alternatively, a new workflow virtual could be added that allows for this skip/resume decision.

src/AppInstallerRepositoryCore/Microsoft/CheckpointIndex.h
```diff
@@ -0,0 +1,69 @@
...
+// Licensed under the MIT License.
+#pragma once
+#include "SQLiteWrapper.h"
+#include "Microsoft/Schema/ICheckpointIndex.h"
+#include "Microsoft/SQLiteStorageBase.h"
+#include <winget/ManagedFile.h>
+
+namespace AppInstaller::Repository::Microsoft
+{
+    struct CheckpointIndex : SQLiteStorageBase
```
JohnMcPMS, line 11: nit: this isn't an index.  Not sure what the most precise term is though.

src/AppInstallerRepositoryCore/Microsoft/CheckpointIndex.h
```diff
@@ -0,0 +1,69 @@
...
+        static CheckpointIndex CreateNew(const std::string& filePath, Schema::Version version = Schema::Version::Latest());
+
+        // Opens an existing CheckpointIndex database.
+        static CheckpointIndex Open(const std::string& filePath, OpenDisposition disposition, Utility::ManagedFile&& indexFile = {})
+        {
+            return { filePath, disposition, std::move(indexFile) };
+        }
+
+        // Opens or creates a CheckpointIndex database on the default path.
+        static std::shared_ptr<CheckpointIndex> OpenOrCreateDefault(GUID guid, OpenDisposition openDisposition = OpenDisposition::ReadWrite);
```
JohnMcPMS, line 32: I don't really see any scenario where the disposition would be anything but `ReadWrite`.  I also don't see any scenario where you would want create *or* open; every new resumable should be create and every resume command should be open.

src/AppInstallerRepositoryCore/Microsoft/CheckpointIndex.h
```diff
@@ -0,0 +1,69 @@
...
+
+        // Gets the file path of the CheckpointIndex database.
+        static std::filesystem::path GetCheckpointIndexPath(GUID guid);
+
+        bool IsEmpty();
+
+        IdType SetClientVersion(std::string_view clientVersion);
```
JohnMcPMS, line 39: It will likely be less work in the long run to make an an enum and have `Set(Enum, value)` and `Get(Enum)`.

src/AppInstallerRepositoryCore/Microsoft/CheckpointIndex.h
```diff
@@ -0,0 +1,69 @@
...
+
+        IdType SetCommandName(std::string_view commandName);
+
+        std::string GetCommandName();
+
+        IdType SetCommandArguments(std::string_view commandArguments);
+
+        std::string GetCommandArguments();
+
+        IdType AddContextData(std::string_view checkpointName, int contextData, std::string_view name, std::string_view value);
```
JohnMcPMS, line 51: It might be better to have a mechanism to create a checkpoint (`CheckPoint CreateCheckPoint(name)`) and then store the additional data off of the `CheckPoint` methods.  That makes it easier to ensure some consistency and to have a checkpoint table to store metadata about each checkpoint.

src/AppInstallerRepositoryCore/Microsoft/CheckpointIndex.cpp
```diff
@@ -0,0 +1,212 @@
...
+        s_CheckpointIndexDirectoryOverride = std::move(checkpointIndexDirectory);
+    }
+#endif
+
+    std::filesystem::path CheckpointIndex::GetCheckpointIndexPath(GUID guid)
+    {
+        wchar_t checkpointGuid[256];
+        THROW_HR_IF(E_UNEXPECTED, !StringFromGUID2(guid, checkpointGuid, ARRAYSIZE(checkpointGuid)));
+
+        const auto DefaultIndexDirectoryPath = Runtime::GetPathTo(Runtime::PathName::LocalState) / s_Checkpoints;
```
JohnMcPMS, line 77: Add a new `Runtime` path, then use the existing runtime path test hooks to change it as needed.

src/AppInstallerRepositoryCore/Microsoft/CheckpointIndex.cpp
```diff
@@ -0,0 +1,212 @@
...
+        {
+            std::filesystem::create_directories(checkpointIndexDirectory);
+            AICLI_LOG(Repo, Info, << "Creating checkpoint index directory: " << checkpointIndexDirectory);
+        }
+        else
+        {
+            THROW_HR_IF(ERROR_CANNOT_MAKE, !std::filesystem::is_directory(checkpointIndexDirectory));
+        }
+
+        auto indexPath = checkpointIndexDirectory / checkpointGuid;
```
JohnMcPMS, line 94: The guid should be a directory name and the checkpoint file inside of it can be a fixed name.  There are other files that can be alongside the SQLite database that won't be cleaned up by simply removing the file.

src/AppInstallerRepositoryCore/Microsoft/CheckpointIndex.cpp
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    CheckpointIndex::IdType CheckpointIndex::SetClientVersion(std::string_view clientVersion)
+    {
+        std::lock_guard<std::mutex> lockInterface{ *m_interfaceLock };
+        AICLI_LOG(Repo, Verbose, << "Setting client version [" << clientVersion << "]");
+
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(m_dbconn, "checkpointindex_setclientversion");
+
+        IdType result = m_interface->SetClientVersion(m_dbconn, clientVersion);
```
JohnMcPMS, line 111: Similar to the other comment, it is much less work to have a generic method rather than piping in a new one every time you need a new value saved.

src/AppInstallerRepositoryCore/Microsoft/Schema/Checkpoint_1_0/CheckpointIndexInterface_1_0.cpp
```diff
@@ -0,0 +1,86 @@
...
+    {
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "createCheckpointTables_v1_0");
+        Checkpoint_V1_0::CheckpointContextTable::Create(connection);
+        Checkpoint_V1_0::CheckpointMetadataTable::Create(connection);
+        savepoint.Commit();
+    }
+
+    SQLite::rowid_t CheckpointIndexInterface::SetClientVersion(SQLite::Connection& connection, std::string_view clientVersion)
+    {
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "setClientVersion_v1_0");
```
JohnMcPMS, line 25: You shouldn't need a transaction to set a single row.

JohnMcPMS (commented):
I would also have a checkpoint table with rows like { RowID, Name, CheckpointTime } and then all of the data items would use the checkpoint table RowID rather than strings again.  The filtering/index is easier then.

src/AppInstallerRepositoryCore/Microsoft/Schema/Checkpoint_1_0/CheckpointTable.cpp
```diff
@@ -0,0 +1,80 @@
...
+    
+    void CheckpointTable::Create(SQLite::Connection& connection)
+    {
+        using namespace SQLite::Builder;
+
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "createCheckpointTable_v1_0");
+
+        StatementBuilder createTableBuilder;
+        createTableBuilder.CreateTable(s_CheckpointTable_Table_Name).BeginColumns();
+        createTableBuilder.Column(ColumnBuilder(s_CheckpointTable_Name_Column, Type::Text).Unique());
```
JohnMcPMS, line 27: Any time that you are going to use the row id as a foreign key in another table, you must specify it explicitly as a column or it can be changed on you when the table is rebalanced.  See the index code for the way that you do it (it has to be one of a few specific names and with specific properties).

JohnMcPMS, line 27: Not unique.

src/AppInstallerRepositoryCore/Microsoft/Schema/Checkpoint_1_0/CheckpointTable.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "pch.h"
+#include "CheckpointTable.h"
+#include "SQLiteStatementBuilder.h"
+
+namespace AppInstaller::Repository::Microsoft::Schema::Checkpoint_V1_0
+{
+    using namespace std::string_view_literals;
+    static constexpr std::string_view s_CheckpointTable_Table_Name = "Checkpoints"sv;
+    static constexpr std::string_view s_CheckpointTable_Name_Column = "Name";
+    static constexpr std::string_view s_CheckpointTable_WriteTime_Column = "WriteTime";
```
JohnMcPMS, line 12: This should be "CreationTime" unless you are planning to update it on every field write.

src/AppInstallerRepositoryCore/Microsoft/Schema/Checkpoint_1_0/CheckpointDataTable.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "createCheckpointDataTable_v1_0");
+
+        StatementBuilder createTableBuilder;
+        createTableBuilder.CreateTable(s_CheckpointDataTable_Table_Name).BeginColumns();
+        createTableBuilder.Column(ColumnBuilder(s_CheckpointDataTable_CheckpointId_Column, Type::Int).NotNull());
+        createTableBuilder.Column(ColumnBuilder(s_CheckpointDataTable_ContextData_Column, Type::Int));
+        createTableBuilder.Column(ColumnBuilder(s_CheckpointDataTable_Name_Column, Type::Text));
+        createTableBuilder.Column(ColumnBuilder(s_CheckpointDataTable_Value_Column, Type::Text));
+        createTableBuilder.Column(ColumnBuilder(s_CheckpointDataTable_Index_Column, Type::Int));
```
JohnMcPMS, line 64: Table should probably have a primary key on { checkpoint, data, name, index }.  Also, I don't think you want to allow nulls in any of the columns except for value, and then I think it would be better to have the design be that you simply don't insert the value rather than putting in a null one.

src/AppInstallerCLICore/Checkpoint.h
```diff
@@ -0,0 +1,88 @@
...
+        ClientVersion,
+        CommandName,
+        Arguments
+    };
+
+    struct CheckpointManager;
+
+    // A representation of a row in the Checkpoint table. 
+    template <typename T>
+    struct Checkpoint
```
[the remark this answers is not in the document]
ryfu-msft: Moved to repository core

src/AppInstallerRepositoryCore/Microsoft/CheckpointRecord.h
```diff
@@ -0,0 +1,76 @@
...
+// Licensed under the MIT License.
+#pragma once
+#include "SQLiteWrapper.h"
+#include "Microsoft/Schema/ICheckpointRecord.h"
+#include "Microsoft/SQLiteStorageBase.h"
+#include <winget/ManagedFile.h>
+
+namespace AppInstaller::Repository::Microsoft
+{
+    struct CheckpointRecord : SQLiteStorageBase
```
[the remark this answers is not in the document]
ryfu-msft: Renamed to CheckpointDatabase

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,123 @@
...
+    std::filesystem::path CheckpointManager::GetCheckpointRecordPath(GUID guid)
+    {
+        wchar_t checkpointGuid[256];
+        THROW_HR_IF(E_UNEXPECTED, !StringFromGUID2(guid, checkpointGuid, ARRAYSIZE(checkpointGuid)));
+
+        const auto checkpointsDirectory = Runtime::GetPathTo(Runtime::PathName::CheckpointsLocation) / checkpointGuid;
+
+        if (!std::filesystem::exists(checkpointsDirectory))
+        {
+            std::filesystem::create_directories(checkpointsDirectory);
```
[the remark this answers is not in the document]
ryfu-msft: Added a boolean parameter to indicate whether we should create the checkpoint directory.

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+        auto recordPath = checkpointsDirectory / s_CheckpointsFileName;
+        return recordPath;
+    }
+
+    CheckpointManager::CheckpointManager()
+    {
+        std::ignore = CoCreateGuid(&m_resumeId);
+        const auto& checkpointRecordPath = GetCheckpointRecordPath(m_resumeId);
+        m_checkpointRecord = std::make_shared<CheckpointRecord>(CheckpointRecord::CreateNew(checkpointRecordPath.u8string()));
```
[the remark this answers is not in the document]
ryfu-msft: Changed so that it just returns a `shared_ptr`

src/AppInstallerRepositoryCore/Microsoft/Schema/Checkpoint_1_0/CheckpointTable.cpp
ryfu-msft: Removed Guid and just made the resume id a string. Also removed the argument validation so that it simply checks if we can find the checkpoint database with the provided resume id string.

src/AppInstallerRepositoryCore/Microsoft/Schema/ICheckpointRecord.h
```diff
@@ -0,0 +1,50 @@
...
+        virtual SQLite::rowid_t AddCheckpoint(SQLite::Connection& connection, std::string_view checkpointName) = 0;
+
+        // Gets the id of a checkpoint by name.
+        virtual std::optional<SQLite::rowid_t> GetCheckpointIdByName(SQLite::Connection& connection, std::string_view checkpointName) = 0;
+
+        // Gets the data types associated with a checkpoint id.
+        virtual std::vector<int> GetCheckpointDataTypes(SQLite::Connection& connection, SQLite::rowid_t checkpointId) = 0;
+
+        // Returns a bool value indicating whether a field exists for a checkpoint data type.
+        virtual bool HasCheckpointDataField(SQLite::Connection& connection, SQLite::rowid_t checkpointId, int dataType, std::string_view name) = 0;
```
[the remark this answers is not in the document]
ryfu-msft: Changed function to return an `optional`

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -417,10 +418,15 @@ namespace AppInstaller::Runtime
             result.ACL[ACEPrincipal::System] = ACEPermissions::All;
             result.ACL[ACEPrincipal::Admins] = ACEPermissions::All;
             break;
+        case PathName::CheckpointsLocation:
         case PathName::LocalState:
         case PathName::UserFileSettings:
             result.Path.assign(appStorage.LocalFolder().Path().c_str());
             mayBeInProfilePath = true;
+            if (path == PathName::CheckpointsLocation)
```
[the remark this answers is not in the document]
ryfu-msft: Changed to constructing the path recursively.

src/AppInstallerCLICore/ExecutionContext.cpp
```diff
@@ -407,4 +426,54 @@ namespace AppInstaller::CLI::Execution
...
+
+    void Context::Checkpoint(std::string_view checkpointName, std::vector<Execution::Data> contextData)
+    {
+        UNREFERENCED_PARAMETER(checkpointName);
+        UNREFERENCED_PARAMETER(contextData);
+
+        if (!m_checkpointManager)
+        {
+            m_checkpointManager = std::make_unique<AppInstaller::Checkpoints::CheckpointManager>();
+            Checkpoints::Checkpoint<AutomaticCheckpointData> automaticCheckpoint = m_checkpointManager->CreateAutomaticCheckpoint();
```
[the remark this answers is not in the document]
ryfu-msft: Changed the method to take in the context and create the automatic checkpoint.

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,123 @@
...
+    {
+        CheckpointRecord::IdType startCheckpointId = m_checkpointRecord->AddCheckpoint(checkpointName);
+        Checkpoint<CLI::Execution::Data> checkpoint{ m_checkpointRecord, startCheckpointId };
+        return checkpoint;
+    }
+
+    std::vector<Checkpoint<CLI::Execution::Data>> CheckpointManager::GetCheckpoints()
+    {
+        std::vector<Checkpoint<CLI::Execution::Data>> checkpoints;
+        for (const auto& checkpointName : m_checkpointRecord->GetCheckpoints())
```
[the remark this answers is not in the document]
ryfu-msft: Changed to suggested. Now returns a list of ids in descending order with the automatic checkpoint id being the last id.

src/AppInstallerRepositoryCore/Microsoft/Schema/Checkpoint_1_0/CheckpointDataTable.cpp
ryfu-msft: Changed to suggested

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,150 @@
...
+        if (createCheckpointDirectory)
+        {
+            if (!std::filesystem::exists(checkpointsDirectory))
+            {
+                AICLI_LOG(Repo, Info, << "Creating checkpoint database directory: " << checkpointsDirectory);
+                std::filesystem::create_directories(checkpointsDirectory);
+            }
+            else
+            {
+                THROW_HR_IF(ERROR_CANNOT_MAKE, !std::filesystem::is_directory(checkpointsDirectory));
```
JohnMcPMS, line 33:
```suggestion
                THROW_HR_IF(MAKE_HRESULT_FROM_WIN32(ERROR_CANNOT_MAKE), !std::filesystem::is_directory(checkpointsDirectory));
```

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,150 @@
...
+    using namespace AppInstaller::Repository::Microsoft;
+    using namespace AppInstaller::Repository::SQLite;
+
+    // This checkpoint name is reserved for the starting checkpoint which captures the automatic metadata.
+    constexpr std::string_view s_AutomaticCheckpoint = "automatic"sv;
+    constexpr std::string_view s_CheckpointsFileName = "checkpoints.db"sv;
+
+    std::filesystem::path CheckpointManager::GetCheckpointDatabasePath(const std::string_view& resumeId, bool createCheckpointDirectory)
+    {
+
```
JohnMcPMS, line 21: nit: extre newline

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,150 @@
...
+        Checkpoint<CLI::Execution::Data> checkpoint{ m_checkpointDatabase, checkpointId };
+        return checkpoint;
+    }
+
+    std::vector<Checkpoint<CLI::Execution::Data>> CheckpointManager::GetCheckpoints()
+    {
+        auto checkpointIds = m_checkpointDatabase->GetCheckpointIds();
+
+        // Remove the last checkpoint (automatic)
+        checkpointIds.pop_back();
```
JohnMcPMS, line 112: In `GetAutomaticCheckpoint` you check for empty, but not here.

src/AppInstallerCLICore/CheckpointManager.cpp
```diff
@@ -0,0 +1,150 @@
...
+            if (std::filesystem::exists(checkpointDatabasePath))
+            {
+                std::error_code error;
+                if (std::filesystem::remove(checkpointDatabasePath, error))
+                {
+                    AICLI_LOG(CLI, Info, << "Checkpoint database deleted: " << checkpointDatabasePath);
+                }
+
+                const auto& checkpointDatabaseParentDirectory = checkpointDatabasePath.parent_path();
+                if (std::filesystem::remove(checkpointDatabaseParentDirectory, error))
```
JohnMcPMS, line 143: I feel like I've left this comment before, but why not simply delete the directory with a recursive delete method?

src/AppInstallerCLICore/Commands/ResumeCommand.cpp
```diff
@@ -0,0 +1,135 @@
...
+        {
+            context.Reporter.Error() << Resource::String::ClientVersionMismatchError(Utility::LocIndView{ checkpointClientVersion }) << std::endl;
+            AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_CLIENT_VERSION_MISMATCH);
+        }
+
+        const auto& checkpointCommand = automaticCheckpoint.Get(AutomaticCheckpointData::Command, {});
+        AICLI_LOG(CLI, Info, << "Resuming command: " << checkpointCommand);
+        std::unique_ptr<Command> commandToResume = FindCommandToResume(checkpointCommand);
+
+        for (const auto& fieldName : automaticCheckpoint.GetFieldNames(AutomaticCheckpointData::Arguments))
```
JohnMcPMS, line 106: For future maintainability, it would be better if this code that reads the values out were next to the code that writes the values in.  Even if that means a standalone helper, I would much rather have them in the same file (and even have the functions next to each other).

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2581,4 +2581,27 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="WINGET_CONFIG_ERROR_UNIT_SETTING_CONFIG_ROOT" xml:space="preserve">
     <value>A unit contains a setting that requires the config root.</value>
   </data>
+  <data name="ResumeCommandLongDescription" xml:space="preserve">
+    <value>Resumes execution of a previously saved command by passing in the unique guid identifier of the saved command. This is used to resume an executed command that may have been terminated due to a reboot.</value>
```
JohnMcPMS, line 2585:
```suggestion
    <value>Resumes execution of a previously saved command by passing in the unique identifier of the saved command. This is used to resume an executed command that may have been terminated due to a reboot.</value>

```

src/AppInstallerCLICore/Command.cpp
```diff
@@ -883,6 +883,12 @@ namespace AppInstaller::CLI
         }
     }
 
+    void Command::Resume(Execution::Context& context) const
+    {
+        context.Reporter.Error() << Resource::String::CommandDoesNotSupportResumeMessage << std::endl;
+        THROW_HR(E_NOTIMPL);
```
JohnMcPMS, line 889: Terminate the context rather than throwing I think...

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -31,6 +31,7 @@ namespace AppInstaller::Runtime
         constexpr std::string_view s_PortablePackageRoot = "WinGet"sv;
         constexpr std::string_view s_PortablePackagesDirectory = "Packages"sv;
         constexpr std::string_view s_LinksDirectory = "Links"sv;
+        constexpr std::string_view s_CheckpointsDirectory = "Checkpoints"sv;
```
JohnMcPMS, line 34: At some point I think we will need to have something going through and cleaning out old resume context data.  Doesn't need to be now.

src/AppInstallerSharedLib/Public/AppInstallerStrings.h
```diff
@@ -266,4 +266,6 @@ namespace AppInstaller::Utility
     // Converts the given boolean value to a string.
     std::string_view ConvertBoolToString(bool value);
 
+    // Converts the given GUID value to a string.
+    std::string ConvertGuidToString(GUID value);
```
JohnMcPMS, line 270:
```suggestion
    std::string ConvertGuidToString(const GUID& value);
```

JohnMcPMS (approved):

PR #3516, opened 2023-08-10
Add support for `--Installer-Type` argument for commands

### LINKED ISSUE #1166: Specify InstallerType
# Add an argument and settings to specify installer type

As a user I would like to be able to specify which installer type to use on my machine.

This could be an argument like: `winget install foo.bar --installer-type exe`

There could also be settings to specify a "preferred" installer type or a "required" installer type.

Given a "required" installer type, the search results could be trimmed to include only packages with manifests supporting my "required" installer type.

* Code updated to allow for an ordering
* Settings to enable preference/requirement and the ordering thereof
* Command line argument to specify immediate requirement
* COM API update to enable a choice there

ryfu-msft:
Resolves #1166 

Changes:
- Adds the `--installer-type` argument to the Install, Upgrade, and Show command.
- Modifies settings schema to allow for installer type preferences and requirements.
- Added the corresponding argument to the COM InstallOptions.
- Updated the manifest installer comparator to handle both installer type preferences and requirements.

Tests:
- Unit tests to verify the manifest comparator when providing the installerType arg, preference, and requirements.
- E2E tests to verify the behavior for COM install, Install, and Show.
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-cli/pull/3516&drop=dogfoodAlpha

### DIFF
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -140,6 +140,18 @@ The `architectures` behavior affects what architectures will be selected when in
     },
 ```
 
+### Installer Types
+
+The `installerTypes` behavior affects what installer types will be selected when installing a package. The matching parameter is `--installer-type`. Note that if multiple preferred installer types are available for installation, the order the installer types are specified will dictate which installer type is more preferred.
+
+```json
+    "installBehavior": {
+        "preferences": {
+            "installerTypes": ["msi", "msix"]
+        }
+    },
+```
+
 ### Default install root
 
 The `defaultInstallRoot` affects the install location when a package requires one. This can be overridden by the `--location` parameter. This setting is only used when a package manifest includes `InstallLocationRequired`, and the actual location is obtained by appending the package ID to the root.
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -142,7 +142,7 @@ The `architectures` behavior affects what architectures will be selected when in
 
 ### Installer Types
 
-The `installerTypes` behavior affects what installer types will be selected when installing a package. The matching parameter is `--installer-type`. Note that if multiple preferred installer types are available for installation, the order the installer types are specified will dictate which installer type is more preferred.
+The `installerTypes` behavior affects what installer types will be selected when installing a package. The matching parameter is `--installer-type`.
 
 ```json
     "installBehavior": {
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -93,6 +93,28 @@
             "minItems": 1,
             "maxItems": 4
           }
+        },
+        "installerTypes": {
+          "description": "The installerType(s) to use for a package install",
+          "type": "array",
+          "items": {
+            "uniqueItems": "true",
+            "type": "string",
+            "enum": [
+              "inno",
+              "wix",
+              "msi",
+              "nullsoft",
+              "zip",
+              "msix",
+              "exe",
+              "burn",
+              "msstore",
+              "portable"
+            ],
+            "minItems": 1,
+            "maxItems": 9
+          }
         }
       }
     },
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -282,17 +282,29 @@ namespace AppInstaller::CLI::Workflow
                 }
             }
 
+            std::string ExplainInapplicable(const Manifest::ManifestInstaller& installer) override
+            {
+                std::string result = "InstallerType does not match required type: ";
+                result += InstallerTypeToString(installer.EffectiveInstallerType());
+                result += "Required InstallerTypes: ";
+                result += m_requirementAsString;
+                return result;
+            }
+
+            bool ContainsInstallerType(const std::vector<InstallerTypeEnum>& selection, InstallerTypeEnum installerType)
+            {
+                return std::find(selection.begin(), selection.end(), installerType) != selection.end();
+            }
+
             InapplicabilityFlags IsApplicable(const Manifest::ManifestInstaller& installer) override
             {
                 if (!m_requirement.empty())
                 {
-                    for (auto requiredInstallerType : m_requirement)
+                    // The installer is applicable if the effective or base installer type matches.
+                    if (ContainsInstallerType(m_requirement, installer.EffectiveInstallerType()) ||
+                        ContainsInstallerType(m_requirement, installer.BaseInstallerType))
                     {
-                        // The installer is applicable if the installer type or nested installer type matches. (User should be allowed to specify 'zip')
-                        if (installer.EffectiveInstallerType() == requiredInstallerType || installer.BaseInstallerType == requiredInstallerType)
-                        {
-                            return InapplicabilityFlags::None;
-                        }
+                        return InapplicabilityFlags::None;
                     }
 
                     return InapplicabilityFlags::InstallerType;
@@ -303,15 +315,6 @@ namespace AppInstaller::CLI::Workflow
                 }
             }
 
-            std::string ExplainInapplicable(const Manifest::ManifestInstaller& installer) override
-            {
-                std::string result = "InstallerType does not match required type: ";
-                result += InstallerTypeToString(installer.EffectiveInstallerType());
-                result += "Required InstallerTypes: ";
-                result += m_requirementAsString;
-                return result;
-            }
-
             bool IsFirstBetter(const Manifest::ManifestInstaller& first, const Manifest::ManifestInstaller& second) override
             {
                 if (m_preference.empty())
@@ -319,26 +322,22 @@ namespace AppInstaller::CLI::Workflow
                     return false;
                 }
 
-                InstallerTypeEnum firstBaseInstallerType = first.BaseInstallerType;
-                InstallerTypeEnum firstEffectiveInstallerType = first.EffectiveInstallerType();
+                bool isFirstInstallerTypePreferred =
+                    ContainsInstallerType(m_preference, first.EffectiveInstallerType()) ||
+                    ContainsInstallerType(m_preference, first.BaseInstallerType);
 
-                InstallerTypeEnum secondBaseInstallerType = second.BaseInstallerType;
-                InstallerTypeEnum secondEffectiveInstallerType = second.EffectiveInstallerType();
+                bool isSecondInstallerTypePreferred = 
+                    ContainsInstallerType(m_preference, second.EffectiveInstallerType()) ||
+                    ContainsInstallerType(m_preference, second.BaseInstallerType);
 
-                for (auto const& preferredInstallerType : m_preference)
+                if (isFirstInstallerTypePreferred && isSecondInstallerTypePreferred)
                 {
-                    bool isFirstInstallerTypePreferred = (preferredInstallerType == firstBaseInstallerType) || (preferredInstallerType == firstEffectiveInstallerType);
-                    bool isSecondInstallerTypePreferred = (preferredInstallerType == secondBaseInstallerType) || (preferredInstallerType == secondEffectiveInstallerType);
+                    return false;
+                }
 
-                    if (isFirstInstallerTypePreferred && isSecondInstallerTypePreferred)
-                    {
-                        return false;
-                    }
-
-                    if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)
-                    {
-                        return isFirstInstallerTypePreferred;
-                    }
+                if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)
+                {
+                    return isFirstInstallerTypePreferred;
                 }
 
                 return false;
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -330,7 +330,7 @@ namespace AppInstaller::CLI::Workflow
                     bool isFirstInstallerTypePreferred = (preferredInstallerType == firstBaseInstallerType) || (preferredInstallerType == firstEffectiveInstallerType);
                     bool isSecondInstallerTypePreferred = (preferredInstallerType == secondBaseInstallerType) || (preferredInstallerType == secondEffectiveInstallerType);
 
-                    if (isFirstInstallerTypePreferred || isSecondInstallerTypePreferred)
+                    if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)
                     {
                         return isFirstInstallerTypePreferred;
                     }
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -332,7 +332,7 @@ namespace AppInstaller::CLI::Workflow
 
                     if (isFirstInstallerTypePreferred && isSecondInstallerTypePreferred)
                     {
-                        break;
+                        return false;
                     }
 
                     if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -284,9 +284,9 @@ namespace AppInstaller::CLI::Workflow
 
             std::string ExplainInapplicable(const Manifest::ManifestInstaller& installer) override
             {
-                std::string result = "InstallerType does not match required type: ";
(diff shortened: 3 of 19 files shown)


src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -306,15 +314,35 @@ namespace AppInstaller::CLI::Workflow
...
+
+                InstallerTypeEnum secondBaseInstallerType = second.BaseInstallerType;
+                InstallerTypeEnum secondEffectiveInstallerType = second.EffectiveInstallerType();
+
+                for (auto const& preferredInstallerType : m_preference)
+                {
+                    bool isFirstInstallerTypePreferred = (preferredInstallerType == firstBaseInstallerType) || (preferredInstallerType == firstEffectiveInstallerType);
+                    bool isSecondInstallerTypePreferred = (preferredInstallerType == secondBaseInstallerType) || (preferredInstallerType == secondEffectiveInstallerType);
+
+                    if (isFirstInstallerTypePreferred || isSecondInstallerTypePreferred)
```
Trenly, line 333: Just thinking through this, it seems to me that it should be an exclusive or here? Or maybe a check for both being preferred types needs to be added?

* If there is no installer type preference - the installers are equally preferred, return false
* If the first installer type is a preferred installer type and the second installer type is not a preferred installer type - first is better, return true
* If the first installer type is not a preferred installer type and the second installer type is a preferred installer type - second is better, return false

Current Implementation - 
* If both installers are of a preferred installer type - the first installer is preferred, return true

Ideal Implementation -
* If both installers are of a preferred installer type - the installers are equally preferred, return false

src/Microsoft.Management.Deployment/InstallOptions.h
```diff
@@ -43,12 +45,14 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         bool SkipDependencies();
         void SkipDependencies(bool value);
 
+
```
Trenly, line 48: nit: whitespace

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
ryfu-msft: Thanks for catching that mistake. I did not account for when both installers match the preferred installer type. 

I changed the logic such that:
- If both installers match the preferred installer type, the immediately return false since they are equally preferred. 
- If exactly one of these installers has a match, then we return the state of the first installer match.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -306,15 +314,40 @@ namespace AppInstaller::CLI::Workflow
...
+                InstallerTypeEnum secondEffectiveInstallerType = second.EffectiveInstallerType();
+
+                for (auto const& preferredInstallerType : m_preference)
+                {
+                    bool isFirstInstallerTypePreferred = (preferredInstallerType == firstBaseInstallerType) || (preferredInstallerType == firstEffectiveInstallerType);
+                    bool isSecondInstallerTypePreferred = (preferredInstallerType == secondBaseInstallerType) || (preferredInstallerType == secondEffectiveInstallerType);
+
+                    if (isFirstInstallerTypePreferred && isSecondInstallerTypePreferred)
+                    {
+                        return false;
```
Trenly, line 335: The more I think about this, I'm not sure the for loop is best here?

Presume for a moment `m_preference` is `{exe, nullsoft, inno}` - `first.BaseInstallerType` is `exe` - `second.BaseInstallerType` is `inno`

The very first time through the for loop, `preferredInstallerType` will be exe. When it goes through the comparison, it will return `true` indicating that the exe is better than the inno.

-----

This is good if the preferences list is intended to be in order of preference. But, if it's intended to be that all preferences are of equal preference, then this won't be accurate - we would need a member function to check if the preferences vector contains the installer type.
```c++
bool PreferenceContains(InstallerTypeEnum installerType) {
    return std::find(m_preference.begin(), m_preference.end(), installerType) != m_preference.end();
}
```

The other thing that may need to be considered is how this interacts with the other applicability filters. 
On an x64 system - if there is an `x86 - exe` and and `x64 - inno`, which would be selected? The architecture filter would prefer the x64, but if the user has specified preferences of `{exe, nullsoft, inno}` , would it still select the Exe as preferred since it's first in the list?

Trenly: Looked into the code further - Architecture preferences/requirements are handled where any architecture in the list is allowed, and order is not considered for preference. I'd suggest we use the same approach here

ryfu-msft: You bring up some great points, all of which I agree with. I had originally thought that order of preference would be a good idea, but this implementation implies that there is *always* an order and I think that could introduce an unnecessary behavior. A user would have no way of declaring preferences that are all equally preferred.

I also think that if other setting preferences don't have this notion of an order, I should keep the behavior the same and not introduce that just for installer types. 

I have updated preference to simply check if the installer type exists in the preferences and determining that result based on that.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -295,26 +315,38 @@ namespace AppInstaller::CLI::Workflow
...
+                {
+                    return false;
+                }
+
+                if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)
+                {
+                    return isFirstInstallerTypePreferred;
+                }
+
+                return false;
```
Trenly, line 343:
```suggestion
               return isFirstInstallerTypePreferred > isSecondInstallerTypePreferred
```
`(true > false)` returns `true` - equivalent to `isFirstInstallerTypePreferred != isSecondInstallerTypePreferred`
`(true > true)` returns `false` - equivalent to `isFirstInstallerTypePreferred && isSecondInstallerTypePreferred`
`(false > false)` returns `false` - equivalent to when neither if statement is matched
`(false > true)` returns `false` - equivalent to `isFirstInstallerTypePreferred != isSecondInstallerTypePreferred`

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -247,44 +247,64 @@ namespace AppInstaller::CLI::Workflow
...
+                std::string result = "InstallerType does not match required type: ";
+                result += InstallerTypeToString(installer.EffectiveInstallerType());
+                result += "Required InstallerTypes: ";
+                result += m_requirementAsString;
+                return result;
+            }
+
+            bool ContainsInstallerType(const std::vector<InstallerTypeEnum>& selection, InstallerTypeEnum installerType)
+            {
+                return std::find(selection.begin(), selection.end(), installerType) != selection.end();
```
Trenly, line 296: Also, thinking more about this, I wonder if it would make sense to have a templated utility function for checking if an object is contained within a container. 

Although, considering that this PR is required for 1.6, it might make sense not to create a utility function.

-----

I'm thinking something like this might work, but I'm not certain

```c++
template <template<typename, typename> typename Container, typename Allocator, typename Value>
bool ContainsObject(Container<Value, Allocator> container, Value value) {
  return std::find(container.begin(), container.end(), value) != container.end();
}
```

src/AppInstallerCLIE2ETests/Helpers/WinGetSettingsHelper.cs
```diff
@@ -196,5 +202,24 @@ public static void InitializeAllFeatures(bool status)
...
+        private static JObject GetJsonSettingsObject(string objectName, string propertyName)
+        {
+            JObject settingsJson = JObject.Parse(File.ReadAllText(TestSetup.Parameters.SettingsJsonFilePath));
+
+            if (!settingsJson.ContainsKey(objectName))
+            {
+                settingsJson[objectName] = new JObject();
+            }
+
+            var installBehavior = settingsJson[objectName];
```
florelis, line 215:
```suggestion
            var jsonObject = settingsJson[objectName];
```

src/AppInstallerCLIE2ETests/Helpers/WinGetSettingsHelper.cs
```diff
@@ -196,5 +202,24 @@ public static void InitializeAllFeatures(bool status)
             ConfigureFeature("windowsFeature", status);
             ConfigureFeature("download", status);
         }
+
+        private static JObject GetJsonSettingsObject(string objectName, string propertyName)
```
florelis, line 206: What does this function do?
It seems like it parses the settings file, then adds an object with a single empty property?
That is weird to me, and the method name doesn't help.

florelis: What about something like

```C#
private static JObject GetJsonSettingsObjectAndCreateEmptyProperty(string[] propertyPath...)
{
    JObject settingsJson = JObject.Parse(File.ReadAllText(TestSetup.Parameters.SettingsJsonFilePath));
    var currentNode = settings.Json;
    foreach (var key in propertyPath)
    {
        if (!currentNode.ContainsKey(key))
        {
            currentNode[key] = new JObject();
        }
        currentNode = currentNode[key];
    }

    return settingsJson;
}
```

(Not sure if this even compiles...)

doc/Settings.md
```diff
@@ -140,6 +140,18 @@ The `architectures` behavior affects what architectures will be selected when in
     },
 ```
 
+### Installer Types
+
+The `installerTypes` behavior affects what installer types will be selected when installing a package. The matching parameter is `--installer-type`.
```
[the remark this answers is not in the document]
Trenly: Zip uses the EffectiveInstallerType, but that brings up a good point - if someone has `{zip, exe}` as their preferences, because the check is against both base and effective type, a zip->msi could still be chosen even if a zip->exe exists

And another great point about the exe that installs an msi or msix; AppsAndFeaturesEntries->InstallerType might need to be considered

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -93,6 +93,28 @@
...
+            "type": "string",
+            "enum": [
+              "inno",
+              "wix",
+              "msi",
+              "nullsoft",
+              "zip",
+              "msix",
+              "exe",
+              "burn",
```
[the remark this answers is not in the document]
Trenly: Would this essentially mean that we would need an "InstallerTechnologyType" for each Installer? Effectively reducing it to portable, exe, msi, msix, and msstore?

src/AppInstallerCLIE2ETests/Helpers/WinGetSettingsHelper.cs
ryfu-msft: I changed the helper method to be a little more generic so that it only returns the json settings object that you specify and doesn't do anything with creating a single empty property (which I also agree is kind of weird).

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -345,6 +346,26 @@ namespace AppInstaller::Settings
...
 
+        WINGET_VALIDATE_SIGNATURE(InstallerTypePreference)
+        {
+            std::vector<Manifest::InstallerTypeEnum> installerTypes;
+            for (auto const& i : value)
+            {
+                Manifest::InstallerTypeEnum installerType = Manifest::ConvertToInstallerTypeEnum(i);
+                if (installerType == Manifest::InstallerTypeEnum::Unknown)
+                {
+                    return {};
```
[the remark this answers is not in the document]
ryfu-msft: I basically followed how we validate the other settings (like Architecture preference that drop if there is an invalid value) so I left it as is because I don't think installer type should behave any differently.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
ryfu-msft: Went with Flor's suggestions as I don't want to hide too much behind a boolean comparison as it would make it hard to reason.

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -93,6 +93,28 @@
...
+          "type": "array",
+          "items": {
+            "uniqueItems": "true",
+            "type": "string",
+            "enum": [
+              "inno",
+              "wix",
+              "msi",
+              "nullsoft",
+              "zip",
```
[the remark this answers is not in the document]
ryfu-msft: I don't think it will be a common scenario, but it should still be supported if people prefer that installer type.

schemas/JSON/settings/settings.schema.0.2.json
```diff
@@ -93,6 +93,28 @@
...
+              "msix",
+              "exe",
+              "burn",
+              "msstore",
+              "portable"
+            ],
+            "minItems": 1,
+            "maxItems": 9
```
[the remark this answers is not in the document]
ryfu-msft: All of the other settings arrays declare a maxItems so I followed it just to be consistent.

doc/Settings.md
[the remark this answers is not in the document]
ryfu-msft: For zip containing exe, I check both the baseInstallerType (for zip) and effective installer type to determine if an installer satisfies the preference/requirement so I believe that scenario is covered. I didn't do anything different for "exe that installs an msi" as I only consider what is shown in the manifest

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -247,44 +247,64 @@ namespace AppInstaller::CLI::Workflow
...
+            {
+                std::string result = "InstallerType [";
+                result += InstallerTypeToString(installer.EffectiveInstallerType());
+                result += "] does not match required InstallerTypes: ";
+                result += m_requirementAsString;
+                return result;
+            }
+
+            bool ContainsInstallerType(const std::vector<InstallerTypeEnum>& selection, InstallerTypeEnum installerType)
+            {
```
yao-msft, line 295: nit: move to private

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -295,26 +315,35 @@ namespace AppInstaller::CLI::Workflow
...
+
+                bool isFirstInstallerTypePreferred =
+                    ContainsInstallerType(m_preference, first.EffectiveInstallerType()) ||
+                    ContainsInstallerType(m_preference, first.BaseInstallerType);
+
+                bool isSecondInstallerTypePreferred = 
+                    ContainsInstallerType(m_preference, second.EffectiveInstallerType()) ||
+                    ContainsInstallerType(m_preference, second.BaseInstallerType);
+
+                if (isFirstInstallerTypePreferred == isSecondInstallerTypePreferred)
```
yao-msft, line 333: I'm a bit torn about whether making the preference ordered vs unordered. Making it unordered will limit the user's ability to prefer a specific installer type over another and all other preferences are ordered. Making it ordered seems too strong for installer type though. maybe a middle ground will be making it ordered and put this comparator in the last on line 765 so it's not that significant in the end (btw, putting the comparator to last should be done regardless of whether we decide it ordered or unordered)

yao-msft (approved):

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -295,27 +310,47 @@ namespace AppInstaller::CLI::Workflow
...
+                        second.BaseInstallerType == installerTypePreference;
+
+                    if (isFirstInstallerTypePreferred && isSecondInstallerTypePreferred)
+                    {
+                        return false;
+                    }
+
+                    if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)
+                    {
+                        return isFirstInstallerTypePreferred > isSecondInstallerTypePreferred;
```
yao-msft, line 337:
```suggestion
                        return isFirstInstallerTypePreferred;

```
do not compare bools

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -295,27 +310,47 @@ namespace AppInstaller::CLI::Workflow
...
+                    bool isSecondInstallerTypePreferred =
+                        second.EffectiveInstallerType() == installerTypePreference ||
+                        second.BaseInstallerType == installerTypePreference;
+
+                    if (isFirstInstallerTypePreferred && isSecondInstallerTypePreferred)
+                    {
+                        return false;
+                    }
+
+                    if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)
```
yao-msft, line 335:
```suggestion
                    else if (isFirstInstallerTypePreferred != isSecondInstallerTypePreferred)

```

yao-msft (approved):

PR #3536, opened 2023-08-18
Add ability to specify where to save modules for configuration

msftrubengu:
Before this PR, all the modules are installed using `Install-Module` with its default scope (current user). In the case where other applications (yes, I'm talking about OneDrive) decide to change the user profile while configuration it running, the PowerShell modules placed in MyDocuments\PowerShell\Modules get lost forever (more like moved into the OneDrive location).

To mitigate this issue, this PR adds the ability to specify where to install the modules via APIs.

### Locations

1. `CurrentUser`: the same location where `Install-Module -Scope CurrentUser` will install modules.
2. `AllUsers`: the same location where `Install-Module -Scope AllUsers` will install modules.
3. `WinGetModulePath`: the winget specific location `%LOCALAPPDATA%\Microsoft\WinGet\Configuration\Modules`
4. `Custom`: Any location the user wants to.

Currently, the default location is CurrentUser. Internally, the custom location will only be added to the PSModulePath when specified. The WinGetModulePath is always added to the PSModulePath. When a custom path is provided, it must be an absolute path.

### Winget flags
`winget configure file.yml --module-path currentuser`
`winget configure file.yml --module-path allusers`
`winget configure file.yml --module-path default`
`winget configure file.yml --module-path <path>`

### Microsoft.WinGet.Configuration
`Get-WinGetConfiguration -File file.yml -ModulePath CurrentUser`
`Get-WinGetConfiguration -File file.yml -ModulePath AllUsers`
`Get-WinGetConfiguration -File file.yml -ModulePath Default`
`Get-WinGetConfiguration -File file.yml -ModulePath <path>`

#### TODO
- Add a winget setting that specify the custom location to use. If location is passed, then this will be used. We could also add an additional module path setting just for the sake of it.
- Add a new cmdlet in the Microsoft.WinGet.Configuration module that adds the winget location to the $env:PSModulePath. This way, the same PowerShell session could use the downloaded modules after configuration is done,.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3536)


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -183,6 +183,14 @@ namespace AppInstaller::CLI
             return { type, "enable"_liv, ArgTypeCategory::None, ArgTypeExclusiveSet::StubType };
         case Execution::Args::Type::ConfigurationDisable:
             return { type, "disable"_liv, ArgTypeCategory::None, ArgTypeExclusiveSet::StubType };
+        case Execution::Args::Type::ConfigurationAllUsersLocation:
```
JohnMcPMS, line 186: What if we had one argument and it had special case text rather than all 4?
```
--module-path currentuser
--module-path allusers
--module-path default
--module-path Z:\my_dir
```
If you really want to use a directory named `allusers`, you just fully qualify it `D:\allusers`.

Alternately, I'm not really sure why we need to expose it at all at the moment.

src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
```diff
@@ -766,6 +774,12 @@ namespace Microsoft.Management.Configuration
 
             // The execution policy to apply; must be set before taking actions with the processor.
             PwshConfigurationProcessorPolicy Policy;
+
+            // The location to install modules. Default is CurrentUser
```
JohnMcPMS, line 778: I think we should change the default to `WinGetModulePath` to avoid the problem by default rather than requiring opt-in.

src/Microsoft.Management.Configuration.Processor/Helpers/PowerShellGetV2.cs
```diff
@@ -17,6 +17,8 @@ namespace Microsoft.Management.Configuration.Processor.Helpers
     /// </summary>
     internal class PowerShellGetV2 : IPowerShellGet
     {
+        private const string AllUsers = "AllUsers";
```
JohnMcPMS, line 20: Wouldn't there actually be an enum type for this value? Or do they not expose those?

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
```diff
@@ -74,19 +74,31 @@ public static bool ConfirmConfigurationProcessing(PSCmdlet psCmdlet, bool hasAcc
...
+        /// <param name="allUsers">If to use all users location.</param>
+        /// <param name="currentUser">If to use current user location.</param>
+        /// <param name="wingetModulePath">If to use the winget module path.</param>
+        /// <param name="customLocation">The custom location path.</param>
         /// <param name="executionPolicy">Execution policy.</param>
         /// <param name="canUseTelemetry">If telemetry can be used.</param>
-        public void Get(string configFile, ExecutionPolicy executionPolicy, bool canUseTelemetry)
+        public void Get(
+            string configFile,
+            bool allUsers,
```
JohnMcPMS, line 85: Couldn't we have a type that holds all of this information?
```
class ModuleStorage
{
enum Location { CU, AU, WG, Custom };

Location StorageLocation{get;set;} = WG;

String? CustomPath = null;
}
```

JohnMcPMS: Can types flow into the Engine?  Could be defined in the command side if so.  Or you can just say "this is part of the majesty of ALC design patterns..." 🤮

src/Microsoft.Management.Configuration.Processor/Helpers/PowerShellGetV2.cs
msftrubengu: noup, is just a validate set with two strings and even if they had an enum this is powershellget in psm1 files so no types in the sdk

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
msftrubengu: i changed it to be -ModulePath to be the same as winget, so we don't need it

src/AppInstallerCLICore/ConfigurationCommon.cpp
```diff
@@ -0,0 +1,67 @@
...
+                    location = SetProcessorFactory::PwshConfigurationProcessorLocation::CurrentUser;
+                }
+                else if (Utility::CaseInsensitiveEquals(modulePath, s_ModulePath_AllUsers))
+                {
+                    location = SetProcessorFactory::PwshConfigurationProcessorLocation::AllUsers;
+                }
+                else
+                {
+                    location = SetProcessorFactory::PwshConfigurationProcessorLocation::Custom;
+                    pwshFactory.CustomLocation(winrt::to_hstring(context.Args.GetArg(Execution::Args::Type::ConfigurationModulePath)));
```
JohnMcPMS, line 60: Should we have any validation on this path?  I get that it might not exist, but if I mistype `alusers` then I will end up with a directory created relative to the working directory of the PS runtime?  Maybe we should require that it be an absolute path to prevent such a thing.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2062,21 +2062,12 @@ Please specify one of them using the --source option to proceed.</value>
...
   <data name="ConfigurationAllUsersElevated" xml:space="preserve">
-    <value>all-users require administrator privileges to execute</value>
-    <comment>{Locked="all-users"}</comment>
-  </data>
-  <data name="ConfigurationCurrentUser" xml:space="preserve">
-    <value>Installs modules in a location that's accessible only to the current user of the computer</value>
-  </data>
-  <data name="ConfigurationCustomLocationPath" xml:space="preserve">
-    <value>Specifies the location on the local computer to store modules</value>
+    <value>--module-path allusers require administrator privileges to execute</value>
```
JohnMcPMS, line 2066:
```suggestion
    <value>`--module-path allusers` requires administrator privileges to execute.</value>
```

JohnMcPMS (approved):

PR #3537, opened 2023-08-19
Guard WinRT InProc Window Package Manager Deployment APIs by EnableAp…

Madhusudhan-MSFT:
…pInstaller GroupPolicy

**[WHY:]**
WINRT InProcess Deployment APIs are current not guarded by EnableAppInstaller GroupPolicy. where as COM InProcess/OutProcess API are guarded by group policy.

**[WHAT:]**
- This fix is introducing the Group Policy check for WinRT InProcess deployment APIs to block them if EnableAppInstaller is disabled.
   -  Added a group policy check in DllGetActivationFactory but this is to prevent external code calling into WinRT In Proc API goes through group policy check.
 - Added Group Policy check in our Projection layer
   - ActivationFactoryInstanceInitializer  : WinRT InProc object initializer
   - LocalServerInstanceInitializer : COM out of proc object initalizer
   This way consumer of this code doesn't have to rely too much on error message that bubbles from WinRT DllGetActionvationFactory implementation, and it also doesn't surface actual group policy error when there is a failure for consistent client experience.
- Update GroupPolicy Exception code to map appropriate error code based on the type of failure triggered the exception.
   
- Added some group policy tests for the InProcess/ OutOfProcess Interops to validate group policy indeed blocks the Object creation when policy is disabled.

**[How Tested:]**
- Copied the fix and applied wingetdev package
- GroupPolicyForInterops test run to ensure they all behave as expected.

Related to : https://microsoft.visualstudio.com/OS/_workitems/edit/45796608/

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3537)

### DIFF
--- a/src/AppInstallerCLIE2ETests/Constants.cs
+++ b/src/AppInstallerCLIE2ETests/Constants.cs
@@ -304,6 +304,8 @@ namespace AppInstallerCLIE2ETests
             public const int CONFIG_ERROR_UNIT_IMPORT_MODULE = unchecked((int)0x8A15C108);
             public const int CONFIG_ERROR_UNIT_INVOKE_INVALID_RESULT = unchecked((int)0x8A15C109);
             public const int CONFIG_ERROR_UNIT_SETTING_CONFIG_ROOT = unchecked((int)0x8A15C110);
+
+            public const int REGDB_E_CLASSNOTREG = unchecked((int)0x80040154);
         }
 
 #pragma warning restore SA1310 // Field names should not contain underscore
--- a/src/AppInstallerCLIE2ETests/Constants.cs
+++ b/src/AppInstallerCLIE2ETests/Constants.cs
@@ -128,10 +128,10 @@ namespace AppInstallerCLIE2ETests
         // Group Policy Error Message
         public const string BlockByWinGetPolicyErrorMessage = "This operation is disabled by Group Policy : Enable Windows Package Manager";
 
-    /// <summary>
-    /// Error codes.
-    /// </summary>
-    public class ErrorCode
+        /// <summary>
+        /// Error codes.
+        /// </summary>
+        public class ErrorCode
         {
             public const int S_OK = 0;
             public const int S_FALSE = 1;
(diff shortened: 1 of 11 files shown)


JohnMcPMS (approved):

Madhusudhan-MSFT:
**microsoft.winget-cli (Build x86_release) Failing after 68m — Build x86_release failed :** appears to be test timing issue and is not consistent. Also, it is particularly not related to this change.

Niranjan-Jyothi:
Hey, @yao-msft @denelon @Trenly @JohnMcPMS  I have been trying to use In-Proc Winget COM APIs directly onto my c# app (.net 8) using CsWinRT to eventually install/search applications but the initial PackageManger object creation itself fails with below message

System.TypeInitializationException: 'The type initializer for 'WinRT.ActivationFactory`1' threw an exception.'
Inner Exception
COMException: Class not registered (0x80040154 (REGDB_E_CLASSNOTREG))

image

Can someone point me the right direction on how to achieve this maybe a straightforward example code snippet or documentation. Analysing the powershell (as it uses In-Proc) was getting a bit complicated and did not see any registrations before hand in the code instead the class objects (PackageManager etc) were directly being created in the 
winget-cli\src\PowerShell\Microsoft.WinGet.Client.Engine\Helpers\**ManagementDeploymentFactory.cs** file

P.S : The reason for going with this approach was because we wanted to execute the commands in the SYSTEM context and using direct powershell modules was not approved.

yao-msft:
@Niranjan-Jyothi We have e2e tests running inproc winget apis. You can take a look at that for example as well if looking into Powershell invocation is too complex.

But mainly, did you have all necessary binaries beside your exe? The [InProc nuget package](https://www.nuget.org/packages/Microsoft.WindowsPackageManager.InProcCom/1.7.10091-preview) should contain all necessary binraries for inproc invocation. Also, the **Microsoft.Management.Deployment.InProc.dll** needs to be renamed to **Microsoft.Management.Deployment.dll** to match the api namespace (when invoked through ActivationFactory)

PR #3544, opened 2023-08-22
Rename portable file with command alias if symlinks are not supported

ryfu-msft:
Issue:
If creating a symlink is not supported AND the portable exe does not have the same name as the expected command alias, the portable alias will not be recognized. (example: Installing Microsoft.devTunnel without dev mode and running as non-admin)

Changes:
- Check whether symlink creation is supported at the start of registering ARP entry.
- If symlinkcreation is not supported, the target filename of the portable file entry is renamed with the expected command alias.
- Adding Install directory to Path is now its own step rather than a by product of failing to create a symlink. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3544)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/PortableFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/PortableFlow.cpp
@@ -6,6 +6,7 @@
 #include "WorkflowBase.h"
 #include "winget/Filesystem.h"
 #include "winget/PortableFileEntry.h"
+#include <AppInstallerRuntime.h>
 #include <Microsoft/PortableIndex.h>
 
 using namespace AppInstaller::Manifest;
@@ -168,11 +169,14 @@ namespace AppInstaller::CLI::Workflow
 
         const std::filesystem::path& targetInstallDirectory = portableInstaller.TargetInstallLocation;
         const std::filesystem::path& symlinkDirectory = GetPortableLinksLocation(portableInstaller.GetScope());
+        bool isSymlinkCreationSupported = AppInstaller::Runtime::IsSymlinkCreationSupported();
 
         // InstallerPath will point to a directory if it is extracted from an archive.
         if (std::filesystem::is_directory(installerPath))
         {
             portableInstaller.RecordToIndex = true;
+
+            const std::vector<Manifest::NestedInstallerFile>& nestedInstallerFiles = context.Get<Execution::Data::Installer>()->NestedInstallerFiles;
 
             for (const auto& entry : std::filesystem::directory_iterator(installerPath))
             {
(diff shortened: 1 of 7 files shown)


src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -187,16 +191,36 @@ namespace AppInstaller::CLI::Workflow
...
+                    if (isSymlinkCreationSupported)
+                    {
+                        entries.emplace_back(std::move(PortableFileEntry::CreateFileEntry(entryPath, targetPath, {})));
+                    }
+                    else
+                    {
+                        // Find matching nested installer file and rename the symlink target with the corresponding command alias if available.
+                        for (const auto& nestedInstallerFile : nestedInstallerFiles)
+                        {
+                            if (relativePath == std::filesystem::path{ ConvertToUTF16(nestedInstallerFile.RelativeFilePath) })
```
yao-msft, line 203: == just does raw string comparison (does not consider case insensitive, ..\, .\, etc), make sure it fits what we need here

yao-msft (approved):

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -187,16 +191,36 @@ namespace AppInstaller::CLI::Workflow
...
+                            {
+                                if (!nestedInstallerFile.PortableCommandAlias.empty())
+                                {
+                                    std::filesystem::path targetWithCommandAlias = targetInstallDirectory / ConvertToUTF16(nestedInstallerFile.PortableCommandAlias);
```
yao-msft, line 207: Probably we'll want to keep the nested path structure and only replace the filename to be safe

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -187,16 +191,36 @@ namespace AppInstaller::CLI::Workflow
...
+                                break;
+                            }
+                        }
+                    }
                 }
             }
 
-            const std::vector<Manifest::NestedInstallerFile>& nestedInstallerFiles = context.Get<Execution::Data::Installer>()->NestedInstallerFiles;
-
+            // Even if symlinks are not supported, symlink entries are still needed for accurate reporting of the command aliases.
```
yao-msft, line 219: After discussion, we'll not create these entries with incorrect data, instead we'll handle accordingly by setting a flag in the portable entry.

src/AppInstallerCLICore/Workflows/PortableFlow.cpp
```diff
@@ -239,7 +263,10 @@ namespace AppInstaller::CLI::Workflow
             AppInstaller::Filesystem::AppendExtension(fileName, ".exe");
             AppInstaller::Filesystem::AppendExtension(commandAlias, ".exe");
 
-            const std::filesystem::path& targetFullPath = targetInstallDirectory / fileName;
+            const std::filesystem::path& targetFullPath = isSymlinkCreationSupported ?
+                targetInstallDirectory / fileName :
+                targetInstallDirectory / commandAlias;
+
             entries.emplace_back(std::move(PortableFileEntry::CreateFileEntry(installerPath, targetFullPath, {})));
             entries.emplace_back(std::move(PortableFileEntry::CreateSymlinkEntry(symlinkDirectory / commandAlias, targetFullPath)));
```
yao-msft, line 271: Do not create symlink entry here if symlink creation not supported. Use the same flag in portable entry in archive portable case above.

Issue #3398, opened 2023-06-30
DonateUrl


SpecterShell:
### Description of the new feature / enhancement

Add a `DonateUrl` field to the manifest representing the URL to support authors or sponsor projects.

denelon:
We'll have to think about policy for this so we don't have people trolling to get donations.

PR #3571, opened 2023-08-30
Add DonationUrl Manifest Field

Trenly:
  - #3398

@denelon - I know there is a comment on the issue around needing to think about policy. I think that because this is a URL field it will be subject to the domain validation in the pipeline, which will ensure that it is associated with the publisher. This would trigger a manual review any time there are discrepancies. Also, this is only part of the "show" flow, which makes it less likely to troll.

However, I'm going to keep this as a draft while any further implications of adding this field are considered.

image

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3571)

### DIFF
--- a/schemas/JSON/manifests/v1.6.0/manifest.defaultLocale.1.6.0.json
+++ b/schemas/JSON/manifests/v1.6.0/manifest.defaultLocale.1.6.0.json
@@ -231,6 +231,10 @@
       "$ref": "#/$defs/Url",
       "description": "The purchase url for acquiring entitlement for the package."
     },
+    "DonationUrl": {
+      "$ref": "#/$defs/Url",
+      "description": "The url for making a donation to the publisher, author, or maintainer of the package."
+    },
     "InstallationNotes": {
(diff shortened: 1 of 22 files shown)


denelon:
Thanks. I do want to be careful that this doesn't get abused. Since we don't have the verified developer feature implemented on the business process side, we will need to be judicious about validating the field if it makes its way into the product. Most of the times I've donated, it's been to a particular developer and through a different third-party site.

Trenly:
> Thanks. I do want to be careful that this doesn't get abused. Since we don't have the verified developer feature implemented on the business process side, we will need to be judicious about validating the field if it makes its way into the product. Most of the times I've donated, it's been to a particular developer and through a different third-party site.

Agree. I would think that because this will trigger the validation-domain in most cases, it would be as "simple" as validating the URL matches what can be accessed from the PackageUrl page

PR #3573, opened 2023-08-31
Allow --accept-source-agreements with local manifests

### LINKED ISSUE #3447: Regression: WinGet 1.5.1881 doesn't correctly handle `--manifest` option with `--accept-source-agreements`

I've got an existing workflow that was working fine until the release of WinGet 1.5.1881. When using that release, the `winget install` command that was previously working is now failing with the error `Both local manifest and search query arguments are provided`. This happens even though no search query argument is provided. Specifically, it fails to work when used in conjunction with the `--accept-source-agreements` option. If I revert back to WinGet 1.5.1572, I get the expected behavior.

### Steps to reproduce

1. Open Windows Sandbox

2. Run the following script to install WinGet dependencies:

```powershell
$ErrorActionPreference = 'Stop'
$ProgressPreference = 'SilentlyContinue'
Set-StrictMode -Version 2.0

# Install XAML nuget package
$XAMLPackagePath = "Microsoft.UI.Xaml.2.7.0.appx"
$XAMLPackagePathZip = "Microsoft.UI.Xaml.2.7.0.zip"
$XAMLPackageSubDirectory = "Microsoft.UI.Xaml.2.7.0"

Write-Host "Downloading XAML nuget package"
Invoke-WebRequest 'https://www.nuget.org/api/v2/package/Microsoft.UI.Xaml/2.7.0' -OutFile $XAMLPackagePath

Write-Host "Expanding nuget package"
Move-Item -Path $XAMLPackagePath -Destination $XAMLPackagePathZip
Expand-Archive -Path $XAMLPackagePathZip -Destination $XAMLPackageSubDirectory

$OSArchitecture = (Get-WmiObject Win32_OperatingSystem).OSArchitecture
$XAMLPackageAppxPath = ""
Switch ($OSArchitecture) {
    "64-bit" { $XAMLPackageAppxPath = "$($XAMLPackageSubDirectory)\tools\AppX\x64\Release\Microsoft.UI.Xaml.2.7.appx" }
    "32-bit" { $XAMLPackageAppxPath = "$($XAMLPackageSubDirectory)\tools\AppX\x86\Release\Microsoft.UI.Xaml.2.7.appx" }
}

Write-Host "Installing XML Package"
Add-AppxPackage -Path $XAMLPackageAppxPath

Write-Host "Installing MS Visual C++ library"
Invoke-WebRequest -OutFile Microsoft.VCLibs.x64.14.00.Desktop.appx https://aka.ms/Microsoft.VCLibs.x64.14.00.Desktop.appx
Add-AppxPackage Microsoft.VCLibs.x64.14.00.Desktop.appx
```

3. Install WinGet

```powershell
$ErrorActionPreference = 'Stop'
$ProgressPreference = 'SilentlyContinue'
Set-StrictMode -Version 2.0

# Get metadata about the latest release of the winget CLI on GitHub
$release = Invoke-WebRequest -UseBasicParsing "https://api.github.com/repos/microsoft/winget-cli/releases/111680619" | ConvertFrom-Json
$installerAsset = $release.assets | Where-Object { $_.name -like "Microsoft.DesktopAppInstaller_*.msixbundle" }
$installerUrl = $installerAsset.browser_download_url
$licenseAsset = $release.assets | Where-Object { $_.name -like "*License*.xml" }
$licenseUrl = $licenseAsset.browser_download_url

try {
    Invoke-WebRequest -OutFile desktopAppInstaller.msixbundle $installerUrl    
    Invoke-WebRequest -OutFile license.xml $licenseUrl

    # Install the App Installer (winget CLI)
    Add-AppxProvisionedPackage -Online -PackagePath desktopAppInstaller.msixbundle -LicensePath license.xml
}
finally {
    Remove-Item desktopAppInstaller.msixbundle -ErrorAction Ignore
    Remove-Item license.xml -ErrorAction Ignore
}
```

4. Enable admin setting

```powershell
winget settings --enable LocalManifestFiles
```

5. Setup manifest location

```powershell
mkdir test
```

6. Download existing manifest files from https://github.com/microsoft/winget-pkgs/tree/master/manifests/m/Microsoft/DotNet/Runtime/7/7.0.9 and copy them to the `test` directory

7. Install manifest

```powershell
winget install --accept-source-agreements --manifest test
```

### Expected behavior

Successful installation

### Actual behavior

Error message: `Both local manifest and search query arguments are provided`

### Environment

```shell
Windows Package Manager v1.5.1881                                                                                                                                                                                     Copyright (c) Microsoft Corporation. All rights reserved.                                                                                                                                                                                                                                                                                                                                                                                   Windows: Windows.Desktop v10.0.19041.3208                                                                                                                                                                             System Architecture: X64                                                                                                                                                                                              Package: Microsoft.DesktopAppInstaller v1.20.1881.0                                                                                                                                                                                                                                                                                                                                                                                         Winget Directories                                                                                                                                                                                                    -------------------------------------------------------------------------------------------------------------------------------                                                                                       Logs                               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir                                                                                       User Settings                      %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\settings.json                                                                                       Portable Links Directory (User)    %LOCALAPPDATA%\Microsoft\WinGet\Links                                                                                                                                              Portable Links Directory (Machine) C:\Program Files\WinGet\Links                                                                                                                                                      Portable Package Root (User)       %LOCALAPPDATA%\Microsoft\WinGet\Packages                                                                                                                                           Portable Package Root              C:\Program Files\WinGet\Packages                                                                                                                                                   Portable Package Root (x86)        C:\Program Files (x86)\WinGet\Packages                                                                                                                                                                                                                                                                                                                                                                   Links                                                                                                                                                                                                                 ---------------------------------------------------------------------------                                                                                                                                           Privacy Statement   https://aka.ms/winget-privacy                                                                                                                                                                     License Agreement   https://aka.ms/winget-license                                                                                                                                                                     Third Party Notices https://aka.ms/winget-3rdPartyNotice                                                                                                                                                              Homepage            https://aka.ms/winget                                                                                                                                                                             Windows Store Terms https://www.microsoft.com/en-us/storedocs/terms-of-sale                                                                                                                                                                                                                                                                                                                                                                 Admin Setting                             State                                                                                                                                                                       --------------------------------------------------                                                                                                                                                                    LocalManifestFiles                        Enabled                                                                                                                                                                     BypassCertificatePinningForMicrosoftStore Disabled                                                                                                                                                                    InstallerHashOverride                     Disabled                                                                                                                                                                    LocalArchiveMalwareScanOverride           Disabled     
```
```

### PR DESCRIPTION
**Pull request by Trenly**
  - Resolves #3447
-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3573)

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -46,7 +46,7 @@ namespace AppInstaller::CLI
         case Execution::Args::Type::Command:
             return { type, "command"_liv, "cmd"_liv, ArgTypeCategory::PackageQuery | ArgTypeCategory::SinglePackageQuery };
         case Execution::Args::Type::Source:
-            return { type, "source"_liv, 's', ArgTypeCategory::Source };
+            return { type, "source"_liv, 's', ArgTypeCategory::QuerySource };
         case Execution::Args::Type::Count:
             return { type, "count"_liv, 'n', ArgTypeCategory::PackageQuery | ArgTypeCategory::SinglePackageQuery };
         case Execution::Args::Type::Exact:
@@ -219,13 +219,11 @@ namespace AppInstaller::CLI
             return { type, "force"_liv, ArgTypeCategory::CopyFlagToSubContext };
 
         case Execution::Args::Type::DependencySource:
-            return { type, "dependency-source"_liv, ArgTypeCategory::Source };
+            return { type, "dependency-source"_liv, ArgTypeCategory::ExtendedSource };
         case Execution::Args::Type::CustomHeader:
-            return { type, "header"_liv, ArgTypeCategory::Source };
-
-        // This must specifically excluded from being a source argument due to dependencies in local manifests potentially requiring a source search and source agreements
+            return { type, "header"_liv, ArgTypeCategory::QuerySource };
         case Execution::Args::Type::AcceptSourceAgreements:
-            return { type, "accept-source-agreements"_liv };
+            return { type, "accept-source-agreements"_liv, ArgTypeCategory::ExtendedSource };
 
         case Execution::Args::Type::ToolVersion:
             return { type, "version"_liv, 'v' };
@@ -461,7 +459,7 @@ namespace AppInstaller::CLI
 
         // If a manifest is specified, we cannot also have arguments for searching
         if (WI_IsFlagSet(categories, ArgTypeCategory::Manifest) &&
-            WI_IsAnyFlagSet(categories, ArgTypeCategory::PackageQuery | ArgTypeCategory::Source))
+            WI_IsAnyFlagSet(categories, ArgTypeCategory::PackageQuery | ArgTypeCategory::QuerySource))
         {
             throw CommandException(Resource::String::BothManifestAndSearchQueryProvided);
         }
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -222,8 +222,10 @@ namespace AppInstaller::CLI
             return { type, "dependency-source"_liv, ArgTypeCategory::Source };
         case Execution::Args::Type::CustomHeader:
             return { type, "header"_liv, ArgTypeCategory::Source };
+
+        // This must specifically excluded from being a source argument due to dependencies in local manifests potentially requiring a source search and source agreements
         case Execution::Args::Type::AcceptSourceAgreements:
-            return { type, "accept-source-agreements"_liv, ArgTypeCategory::Source };
+            return { type, "accept-source-agreements"_liv };
 
         case Execution::Args::Type::ToolVersion:
             return { type, "version"_liv, 'v' };
--- a/src/AppInstallerCLICore/Argument.h
+++ b/src/AppInstallerCLICore/Argument.h
@@ -58,15 +58,19 @@ namespace AppInstaller::CLI
         // Arguments for installer or uninstaller behavior, which do not work for multiple packages.
         // E.g.: --override
         SingleInstallerBehavior = 0x20,
-        // Arguments for selecting or interacting with the source.
-        // E.g.: --accept-source-agreements
-        Source = 0x40,
+        // Arguments for selecting or interacting with the source used for initial querying
+        // E.g.: --header
+        QuerySource = 0x40,
         // Arguments that only make sense when talking about multiple packages
         MultiplePackages = 0x80,
         // Flag arguments that should be copied over when creating a sub-context
         CopyFlagToSubContext = 0x100,
         // Arguments with associated values that should be copied over when creating a sub-context
         CopyValueToSubContext = 0x200,
+        // Arguments for selecting or interacting with dependencies or setting specific source behaviors
+        // E.g.: --dependency-source
+        // E.g.: --accept-source-agreements
+        ExtendedSource = 0x400,
     };
 
     DEFINE_ENUM_FLAG_OPERATORS(ArgTypeCategory);

### CODE REVIEW

File: src/AppInstallerCLICore/Argument.cpp [hunk 1]
```diff
@@ -216,8 +216,10 @@ namespace AppInstaller::CLI
             return { type, "dependency-source"_liv, ArgTypeCategory::Source };
         case Execution::Args::Type::CustomHeader:
             return { type, "header"_liv, ArgTypeCategory::Source };
+
+        // This must specifically excluded from being a source argument due to dependencies in local manifests potentially requiring a source search and source agreements
```
yao-msft, line 220: Does this mean this dependency-source (line 216) would also fail? dependency-source is specifically created for local manifest with dependencies testing purpose.

Anyway, this looks like a hacky fix, we should probably fix in the actual validation code instead of removing a correctly assigned tag

Trenly: I guess the quesion then is whether the tag is qctually correctly assigned, as it caused a semi-breaking behavior (according to the linked issue). 

If we are fine with leaving these as being mutually exclusive arguments, then this PR and the linked issue should be closed. 

If we need to fix the regression, then I'm not sure the best way to do it if this isn’t the desired path forward

yao-msft: `winget install -m <manifest> -s <winget>` is not valid, but `winget install -m <manifest> --dependency-source <winget>` is valid. I think we need to assign more granular tags to fix the issue correctly.

Trenly: But then what about when the dependency source requires agreements?

yao-msft: --accept-source-agreements should be allowed too

Trenly: Ah, now I see what you mean about needing to be more granular. 

Because of the `ArgTypeCategory::Source` both `--dependency-source` and `--accept-source-agreements` are unable to be used with the `--manifest` argument, but they are both correctly tagged with `ArgTypeCategory::Source` as they relate to the behavior of sources. Is that accurate? If it is accurate, I would still believe that `--accept-source-agreements` shouldn’t be tagged as a source argument because 1) It isn’t selecting a source and 2) it isn’t providing any data like a header, it's simply a flag stating how the CLI should behave *if* a source agreement is required. I'm thinking of it like `--force` where just because it is added doesn’t mean it will do anything - if there is nothing that needs to be forced then it is effectively ignored; if there are no source agreements - the behavior is the same whether the argument is present or not.

-----

I do agree that the dependency source shpuld be able to be used with local manifests, and based on the current implementation it cannot. 

So in order to be more granular we would either have to add an ArgTypeCategory for indicating that it should still be allowed with `--manifest` (or do the opposite and make allowed the default, flagging the ones that are incompatible). 

Or I suppose we could add a new argtypecategory that better describes what each does - but then I would think that there would need to be a specific one for agreements, and one for dependency source, since those source properties are allowed with local manifests while others aren't - in which case, being their own categories, it would be more straightforward to simply leave them untagged.

Trenly: @yao-msft - Any recommendations on how the additional granularity should be implemented? I ran into the issue with `--dependency-source` not working with `--manifest` today, so I'd like to get the issue resolved

florelis: From thinking about it for, like, five minutes..
I think I would change the `ArgTypeCategory`.

Replace `ArgTypeCategory::Source` with
* `ArgTypeCategory::SourceForQuery` - for args that refer specifically to the source we're initially querying
* `ArgTypeCategory::SourceOrDependencySource` - for args that apply to either the initial source or the dependencies

Then we can have the validation be
```c++
        if (WI_IsFlagSet(categories, ArgTypeCategory::Manifest) &&
            WI_IsAnyFlagSet(categories, ArgTypeCategory::PackageQuery | ArgTypeCategory::SourceForQuery)) 
```

Then tag
`--source, --header? = SourceForQuery`
`--dependency-source, --accept-source-agreements = SourceOrDependencySource`

I'm not sure if we'd need to do any validation on `SourceOrDependencySource`

Trenly: Ah, that makes perfect sense.

florelis (approved):

PR #3576, opened 2023-08-31
Enable configuration via IConfigurationStatics

msftrubengu:
Winget configuration is not going to be enabled by default. The user needs to opt-in by calling `winget configuration --enable`.  Callers that are interested in using the OOP Microsoft.Management.Configuration APIs need a way to ensure configuration is enabled and enable it if not. This PR address that.

### New IConfigurationStatics methods
- `IsConfigurationEnabled` - Checks if the configuration is enabled by checking if the current AppInstaller package is the stub or full package.  Inproc implementation always returns true.
- `EnableConfigurationAsync` - Asynchronously enables configuration. This changes the stub preference to full package and calls our own APIs to install AppIntaller. By design, this also updates AppInstaller to the newest version.

### Caveats
Callers need to understand that this is effectively saying "winget please update yourself". Which means that once `EnableConfigurationAsync` is done, configuration isn't ready yet. At this point, deployment scheduled AppInstaller for an update. The server process that owns the configuration statics object is already gone (or is about to) and a new instance of ConfigurationStaticFunctions needs to be created in order validate configuration is enabled. It is possible that creating that object fails with an error that the upgrade is in progress. Yes, this is awkward.

### Testing
It is complicated to write automated tests for this without faking most of the stack. I manually created an AppInstaller package with an older version to what we internally have to do these tests. Then I used the current OOP configuration tests to update it.

This is how I programmatically verified it
```
            {
                var statics1 = new ConfigurationStaticFunctions();

                Assert.False(statics1.IsConfigurationEnabled);

                var task = statics1.EnableConfigurationAsync();
                task.Progress = (asyncInfo, progress) =>
                {
                    // show progress
                };

                await task;
            }

            await System.Threading.Tasks.Task.Delay(120000);

            {
                var statics2 = new ConfigurationStaticFunctions(); // This failed with E_NOINTERFACE because the "newer" package in the store doesn't have my changes
                Assert.True(statics2.IsConfigurationEnabled);
            }
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3576)

### DIFF
--- a/src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
+++ b/src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
@@ -715,6 +715,12 @@ namespace Microsoft.Management.Configuration
 
         // Creates a processor from the given factory.
         ConfigurationProcessor CreateConfigurationProcessor(IConfigurationSetProcessorFactory factory);
+
+        // Whether configuration is enabled.
+        Boolean IsConfigurationEnabled{ get; };
+
+        // Enables configuration. Requires store access.
+        Windows.Foundation.IAsyncActionWithProgress<UInt32> EnableConfigurationAsync();
     }
 
     // Top level entry point for configuration, enabling easier usage in out-of-process scenarios.
--- a/src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
+++ b/src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
@@ -717,10 +717,10 @@ namespace Microsoft.Management.Configuration
         ConfigurationProcessor CreateConfigurationProcessor(IConfigurationSetProcessorFactory factory);
 
         // Whether configuration is enabled.
-        Boolean IsConfigurationEnabled{ get; };
+        Boolean IsConfigurationAvailable{ get; };
 
         // Enables configuration. Requires store access.
-        Windows.Foundation.IAsyncActionWithProgress<UInt32> EnableConfigurationAsync();
+        Windows.Foundation.IAsyncActionWithProgress<UInt32> EnsureConfigurationAvailableAsync();
     }
 
     // Top level entry point for configuration, enabling easier usage in out-of-process scenarios.
--- a/src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
+++ b/src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
@@ -174,7 +174,8 @@ namespace ConfigurationShim
 
             auto progress = co_await winrt::get_progress_token();
 
-            // Don't use UpgradePackageAsync
+            // Don't use UpgradePackageAsync, we don't support upgrade for packages from the msstore
+            // it has to be install and internally we know is an update.
             auto installTask = packageManager.InstallPackageAsync(catalogPackage, installOptions);
             installTask.Progress([progress](auto const&, InstallProgress installProgress)
             {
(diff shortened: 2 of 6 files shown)


src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
```diff
@@ -81,11 +103,78 @@ namespace ConfigurationShim
...
+        }
+
+        winrt::Windows::Foundation::IAsyncActionWithProgress<uint32_t> EnableConfigurationAsync()
+        {
+            if (IsConfigurationEnabled())
+            {
+                return;
+            }
+
+            co_await winrt::resume_background();
```
JohnMcPMS, line 128: Need to add:
```
auto strongSelf = get_strong();
```
To prevent the statics object from being destroyed while this runs.

src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
```diff
@@ -81,11 +103,78 @@ namespace ConfigurationShim
...
+
+            SetStubPreferred(false);
+
+            PackageManager packageManager;
```
JohnMcPMS, line 132: Does this actually create the object properly?  I remember seeing newer version of CppWinRT do handle that, but I don't think we have updated to them?

My gut would be that we need to create them with `make_self`.

src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
```diff
@@ -81,11 +103,78 @@ namespace ConfigurationShim
...
+            auto catalogPackage = matches.GetAt(0).CatalogPackage();
+
+            InstallOptions installOptions;
+            installOptions.AcceptPackageAgreements(true);
+            installOptions.AllowUpgradeToUnknownVersion(true);
+            installOptions.Force(true);
+
+            auto progress = co_await winrt::get_progress_token();
+
+            // Don't use UpgradePackageAsync
```
JohnMcPMS, line 161: ... because ...

src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
```diff
@@ -81,11 +103,78 @@ namespace ConfigurationShim
...
+            auto installTask = packageManager.InstallPackageAsync(catalogPackage, installOptions);
+            installTask.Progress([progress](auto const&, InstallProgress installProgress)
+            {
+                if (installProgress.State == PackageInstallProgressState::Downloading && installProgress.BytesRequired != 0)
+                {
+                    progress((uint32_t)(installProgress.DownloadProgress * 50));
```
JohnMcPMS, line 167: My guess is that the package download in this case is going to be more like 80% of the time.

src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
```diff
@@ -715,6 +715,12 @@ namespace Microsoft.Management.Configuration
 
         // Creates a processor from the given factory.
         ConfigurationProcessor CreateConfigurationProcessor(IConfigurationSetProcessorFactory factory);
+
+        // Whether configuration is enabled.
+        Boolean IsConfigurationEnabled{ get; };
+
+        // Enables configuration. Requires store access.
+        Windows.Foundation.IAsyncActionWithProgress<UInt32> EnableConfigurationAsync();
```
JohnMcPMS, line 723:
```suggestion
        Boolean IsConfigurationAvailable{ get; };

        // Enables configuration. Requires store access.
        Windows.Foundation.IAsyncActionWithProgress<UInt32> EnsureConfigurationAvailableAsync();
```
If there is a policy that disables configuration, I feel like the name of this becomes confusing.  A word that invokes "present" seems better.

src/WindowsPackageManager/ConfigurationStaticFunctions.cpp
msftrubengu: Talked offline

JohnMcPMS (approved):
I would prefer a static inside a function rather than a global, but I can live with the global for now.

Issue #3501, opened 2023-08-04
Error installing PowerShell module using PSResourceGet


mgreenegit:
### Brief description of your issue

Error occurs when installing the module using install-psresource.

### Steps to reproduce

install-psresource Microsoft.WinGet.Client -TrustRepository

### Expected behavior

Install module without error (as is the case for PowerShellGet/Install-Module).

### Actual behavior

Install-PSResource: At line:69 char:21
+ NestedModules = if ($env:PROCESSOR_ARCHITECTURE -like 'x86')
+                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~
A variable that cannot be referenced in restricted language mode or a Data section is being referenced. Variables that can be referenced include the following: $PSCulture, $PSUICulture, $true, $false, $null, $PSEdition, $PSScriptRoot.

### Environment

```shell
Pwsh7.4p4
```

kilasuit:
Seem this can be closed as I can't replicate this @denelon

PR #3622, opened 2023-09-14
Microsoft.WinGet.Client Any CPU

msftrubengu:
### Move to Any CPU projects
Moves `Microsoft.WinGet.Client.Cmdlet` and `Microsoft.WinGet.Client.Engine` to be Any CPU platform binaries. This is in preparation of PSResourceGet (aka PowerShellGet v3) release and addresses #3501 as $env:PROCESSOR_ARCHITECTURE is not an allowed variable in a psd1.

### Handle server disconnection
Before, we had a static PackageManager object for OOP calls. In the scenario where winget is upgraded, this object will be disconnected, and one would have to close its PowerShell 7 session and open a new one to keep using the cmdlets. This PR creates a wrapper that handles disconnections.

### Windows PowerShell support
Correctly throw NotSupported exception for cmdlets that use winget's COM APIs. For this I had to track down all the types being loaded per command and make sure they get loaded after the constructor of each one. This resulted in some weird cases that I need to pass a string instead of an enum in the virtual methods of some base classes.
Fix issue where `Microsoft.Win32.Registry.dll` was missing.
Add Pester tests specifically for Windows PowerShell

### Expected Layout
```
Microsoft.WinGet.Client\
	Format.ps1xml
	Microsoft.WinGet.Client.psd1
	net48\
		Microsoft.Win32.Registry.dll
		Microsoft.WinGet.Client.Cmdlets.dll
		Microsoft.WinGet.Client.Engine.dll
		Microsoft.WinGet.SharedLib.dll
		Newtonsoft.Json.dll
		Octokit.dll
	net6.0-windows10.0.22000.0\
		Microsoft.WinGet.Client.Cmdlets.dll
		DirectDependencies\
			Microsoft.WinGet.Client.Engine.dll
		SharedDependencies\
			Microsoft.Windows.SDK.NET.dll
			Microsoft.WinGet.SharedLib.dll
			Newtonsoft.Json.dll
			Octokit.dll
			WinRT.Runtime.dll
			x64\
				Microsoft.Management.Deployment.dll
				Microsoft.Management.Deployment.winmd
				WindowsPackageManager.dll
				winrtact.dll
			x86\
				Microsoft.Management.Deployment.dll
				Microsoft.Management.Deployment.winmd
				WindowsPackageManager.dll
				winrtact.dll
```
I didn't add the native binaries in the Windows PowerShell because they will never be loaded.

Additional changes:

- Remove unnecessary AnyCpu build in AppInstallerCLI.sln
- Fully remove Microsoft.WinGet.Client tests from AppInstallerCLIE2ETests in favor of Pester tests.
- Improve Initialize-LocalWinGetModules.ps1 for local development. You can now specify which module to initialize.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3622)


src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Acl/WinGetAssemblyLoadContext.cs
```diff
@@ -36,11 +36,33 @@ internal class WinGetAssemblyLoadContext : AssemblyLoadContext
...
+        private static readonly IEnumerable<string> ValidArchs = new string[] { "x86", "x64" };
+
         private static readonly WinGetAssemblyLoadContext WinGetAcl = new ();
 
+        private readonly string sharedArchDependencyPath;
+
         private WinGetAssemblyLoadContext()
             : base("WinGetAssemblyLoadContext", isCollectible: false)
         {
+            var arch = Environment.GetEnvironmentVariable("PROCESSOR_ARCHITECTURE", EnvironmentVariableTarget.Process);
```
JohnMcPMS, line 48: Why not `RuntimeInformation.ProcessArchitecture`?

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/PackageManagerWrapper.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        /// <summary>
+        /// Wrapper for InstallPackageAsync.
+        /// </summary>
+        /// <param name="package">The package to install.</param>
+        /// <param name="options">The install options.</param>
+        /// <returns>An async operation with progress.</returns>
+        public IAsyncOperationWithProgress<InstallResult, InstallProgress> InstallPackageAsync(CatalogPackage package, InstallOptions options)
+        {
+            return this.ExecuteWithRetry(
```
JohnMcPMS, line 46: Any of these functions that take in previously created objects are not likely to succeed on a retry from a server process termination (since the objects they refer to are also now disconnected).  They should simply release the static reference and rethrow.

I'm not sure that the retry part is even worth it for the other cases, but you could keep it.

JohnMcPMS (dismissed):

JohnMcPMS (approved):

PR #3631, opened 2023-09-15
Add experimental feature for initiating reboot for single package installs

ryfu-msft:
Related to: #3165 

This PR adds support for the `reboot` experimental feature. This only applies to single package and multiple package installs. This does not apply to any package dependencies that may require a restart.

Changes:
- Added an `AllowReboot` argument flag. When provided, if the return code of an installer is either `APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_FINISH` or `APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_INSTALL`, winget will automatically begin restarting your machine so that the installation can be completed.
- Reboot privilege is checked prior to kicking off a reboot and the appropriate error message is shown if the current user does not have the privilege. 

Tests:
- Verifies that a reboot is initiated for both single and multiple package installs if the return code matches the reboot required return codes. 
- Verifies that the error message is displayed to the user if the shutdown privilege is not met.
- Manually verified that restart is kicked off if requirements for a reboot are met.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3631)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2581,4 +2581,13 @@ Please specify one of them using the --source option to proceed.</value>
...
     <value>A unit contains a setting that requires the config root.</value>
   </data>
+  <data name="NoRebootPrivilegeError" xml:space="preserve">
+    <value>The current user does not have the privilege to initiate a reboot.</value>
+  </data>
+  <data name="AllowRebootArgumentDescription" xml:space="preserve">
+    <value>Allows a reboot if applicable</value>
+  </data>
+  <data name="InitiatingReboot" xml:space="preserve">
+    <value>Initiating reboot to complete installation...</value>
```
JohnMcPMS, line 2591: This is not quite true for `APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_INSTALL` until you hook up `resume`.  I assume that is the plan: `_TO_INSTALL` reboots with a resume but `_TO_FINISH` reboots without a resume?

Side thought, we might already have an issue with `_TO_FINISH` results being treated as an error and thus not recorded properly.  They should probably be treated as a success for post-install purposes.

src/AppInstallerCLITests/InstallFlow.cpp
```diff
@@ -1204,3 +1204,68 @@ TEST_CASE("InstallFlow_InstallAcquiresLock", "[InstallFlow][workflow]")
     REQUIRE(installResultStr.find("/custom") != std::string::npos);
     REQUIRE(installResultStr.find("/silentwithprogress") != std::string::npos);
 }
+
+TEST_CASE("InstallFlow_InstallMultipleWithReboot", "[InstallFlow][workflow][MultiQuery][reboot]")
```
JohnMcPMS, line 1208: Not a huge fan of a test that, if it "fails", reboots my machine...

Can we also have an override that replaces the reboot operation with a callback and you can just set a bool?  That would also allow for tests that actually "reboot".

src/AppInstallerCommonCore/Public/winget/Reboot.h
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+
+namespace AppInstaller::Reboot
+{
+    bool HasRebootPrivilege();
+
+    BOOL InitiateReboot();
```
JohnMcPMS, line 9:
```suggestion
    bool InitiateReboot();
```
Avoid using non-standard types if possible.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -152,6 +153,46 @@ namespace AppInstaller::CLI::Workflow
...
+                if (!Reboot::HasRebootPrivilege())
+                {
+                    AICLI_LOG(CLI, Info, << "Current process does not have reboot privilege.");
+                    context.Reporter.Error() << Resource::String::NoRebootPrivilegeError << std::endl;
+                    return;
+                }
+
+                AICLI_LOG(CLI, Info, << "Install requires reboot. Initiating reboot.");
+                context.Reporter.Info() << Resource::String::InitiatingReboot << std::endl;
+                THROW_LAST_ERROR_IF(Reboot::InitiateReboot());
```
JohnMcPMS, line 193: This feels like a place where we should instead have custom error message about failing to reboot.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -152,6 +153,46 @@ namespace AppInstaller::CLI::Workflow
...
+            if (shouldRebootOverride || ShouldInitiateReboot(context.GetTerminationHR()))
+            {
+                if (!Reboot::HasRebootPrivilege())
+                {
+                    AICLI_LOG(CLI, Info, << "Current process does not have reboot privilege.");
+                    context.Reporter.Error() << Resource::String::NoRebootPrivilegeError << std::endl;
+                    return;
+                }
+
+                AICLI_LOG(CLI, Info, << "Install requires reboot. Initiating reboot.");
```
JohnMcPMS, line 191: Move this logging into the reboot function, so that we always log that we are rebooting.  Change the message appropriately.

src/AppInstallerCommonCore/Reboot.cpp
```diff
@@ -0,0 +1,53 @@
...
+#ifndef AICLI_DISABLE_TEST_HOOKS
+    static bool* s_HasRebootPrivilegeResult_TestHook_Override = nullptr;
+
+    void TestHook_SetHasRebootPrivilegeResult_Override(bool* status)
+    {
+        s_HasRebootPrivilegeResult_TestHook_Override = status;
+    }
+#endif
+
+    bool HasRebootPrivilege()
```
JohnMcPMS, line 18: This function is not just checking for the presence of the privilege, but permanently adjusting the process token to enable it.

If there is need of it, we should have a function that *only* checks for the privilege in the token, not adjusting in.

Then, `IntiateReboot` should be the one responsible for enabling the privilege.  It can return value with more than two states to indicate no privilege, a failure to reboot, or success.  Additionally, the proper method for adjusting privileges in a multi-threaded process is to duplicate the token, adjust the duplicate's privilege, use it as the thread token, take privileged action, then remove the thread token.

src/AppInstallerCLITests/WorkflowCommon.cpp
```diff
@@ -635,6 +645,7 @@ namespace TestCommon
             std::filesystem::path temp = std::filesystem::temp_directory_path();
             temp /= "TestMsixInstalled.txt";
             std::ofstream file(temp, std::ofstream::out);
+            context.Add<Data::OperationReturnCode>(9);
```
yao-msft, line 648: This seems incorrect, it's overriding every test that tries to install an msix.

src/AppInstallerCLITests/TestData/InstallFlowTest_ExpectedReturnCodes.yaml
```diff
@@ -1,7 +1,7 @@
-PackageIdentifier: AppInstallerCliTest.TestInstaller
+PackageIdentifier: AppInstallerCliTest.ExpectedReturnCodes
```
yao-msft, line 1: Is this Id and name change intentional? Will it affect what other tests were testing?

src/AppInstallerCommonCore/Reboot.cpp
```diff
@@ -0,0 +1,53 @@
...
+        TOKEN_PRIVILEGES pTokenPrivileges;
+
+        // Get a token for this process. 
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_ADJUST_PRIVILEGES | TOKEN_QUERY, &hToken))
+        {
+            return false;
+        }
+
+        // Get the shutdown privilege for this process. 
+        THROW_IF_WIN32_BOOL_FALSE(LookupPrivilegeValue(NULL, SE_SHUTDOWN_NAME, &pTokenPrivileges.Privileges[0].Luid));
```
yao-msft, line 37: Maybe just return false instead of throw here and below for checking Privilege. Throw will cause the whole process to exit.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
yao-msft, line 2591:
```suggestion
    <value>Initiating reboot to complete operation...</value>

```

nit: in case it's upgrade, uninstall in the future

ryfu-msft: Made changes to ReportInstallerResult so that `APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_FINISH` is treated as a success and no longer reported as a failed installer result.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -730,6 +781,11 @@ namespace AppInstaller::CLI::Workflow
             }
         }
 
+        if (shouldReboot)
+        {
+           InitiateRebootIfApplicable(context, true);
```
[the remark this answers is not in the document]
ryfu-msft: Changed to setting a context flag so that indicates whether we should initiate a reboot so that it only takes place at the end of all the install workflows and is contained inside a single workflow function.

src/AppInstallerCLITests/InstallFlow.cpp
ryfu-msft: Changed to suggested and added tests that verifies if a reboot actually succeeds

src/AppInstallerCommonCore/Reboot.cpp
ryfu-msft: I made a mistake in the implementation. Turns out we don't need to check for the presence of the privilege. The token needs to be adjusted with the enabled shutdown privilege in order to enable shutdown. I have removed the privilege check and now just adjust the token privilege prior to initiating a reboot with the appropriate error logging if that fails.

src/AppInstallerCLICore/Commands/InstallCommand.cpp
```diff
@@ -50,6 +50,7 @@ namespace AppInstaller::CLI
             Argument::ForType(Args::Type::Rename),
             Argument::ForType(Args::Type::UninstallPrevious),
             Argument::ForType(Args::Type::Force),
+            Argument::ForType(Args::Type::AllowReboot),
```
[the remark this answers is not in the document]
ryfu-msft: Added `allowReboot` to the Upgrade and Import commands along with tests.

src/AppInstallerCLITests/TestData/InstallFlowTest_ExpectedReturnCodes.yaml
ryfu-msft: Name change is intentional as now the id clearly identifies what the purpose of the manifest. Also helps avoid conflicts when creating the test source for update/import/dependency sources.

src/AppInstallerCommonCore/Reboot.cpp
ryfu-msft: Changed to just returning false if this fails rather than throwing.

yao-msft (dismissed):

yao-msft (approved):

PR #3637, opened 2023-09-18
Add try/catch when reading upgrade codes

JohnMcPMS:
## Change
Despite the look of the diff on github, this change only adds two try/catch blocks to the upgrade code parsing code to prevent failures from propagating out.  The upgrade codes are just additional data to help with correlations and should not cause additional user impact beyond not being present.

The outer block protects against the primary key throwing, as well as the subkey enumeration.

The inner block protects against a single upgrade code key throwing, while allowing the loop to potentially continue.

## Validation
Code inspection only.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3637)


msftrubengu (approved):

yao-msft (approved):
:shipit:

PR #3642, opened 2023-09-19
Read registry value data on demand

JohnMcPMS:
## Change
When enumerating registry keys, move the code that reads the registry value data to be on request rather than during the iterator advancement.  This enables code that doesn't need the data to simply skip it, and code that does to better handle the error if desired.

## Validation
Regression only.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3642)


yao-msft (approved):
:shipit:

PR #3659, opened 2023-09-22
Invoke ShellExecute on dism.exe for enabling Windows Features

ryfu-msft:
Changes:
- All calls to dismApi have been replaced with invoking ShellExecute on the Dism.exe. This is because the DismAPI has issues when being used in packaged context. The behavior of enabling Windows Features has been kept the same and all relevant tests have been updated. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3659)


JohnMcPMS:
I disabled the tests `InstallWithWindowsFeatureDependency_FeatureNotFound` and `InstallWithWindowsFeatureDependency_Force`, as they were failing for the reason you are making this change.  Apparently there is some difference between the activation paths.

Please re-enable them in your branch.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -307,7 +347,142 @@ namespace AppInstaller::CLI::Workflow
...
+        }
+
+        const auto& rootDependencies = context.Get<Execution::Data::Installer>()->Dependencies;
+        DWORD result = 0;
+        bool force = context.Args.Contains(Execution::Args::Type::Force);
+        bool rebootRequired = false;
+
+        rootDependencies.ApplyToType(DependencyType::WindowsFeature, [&context, &result, &force, &rebootRequired](Dependency dependency)
+            {
+                if (FAILED(result) && !force || context.IsTerminated())
```
yao-msft, line 384: result does not seem to be an HResult, we would not use FAILED to check non hresult type

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -307,7 +347,142 @@ namespace AppInstaller::CLI::Workflow
...
+                if (!doesFeatureExistResult)
+                {
+                    // User aborted the operation.
+                    AICLI_TERMINATE_CONTEXT(E_ABORT);
+                }
+
+                auto doesFeatureExistExitCode = doesFeatureExistResult.value();
+                if (doesFeatureExistExitCode != ERROR_SUCCESS)
+                {
+                    AICLI_LOG(Core, Info, << "Windows Feature [" << featureName << "] does not exist");
```
yao-msft, line 410:
```suggestion
                    AICLI_LOG(Core, Warning, << "Windows Feature [" << featureName << "] does not exist");

```

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -307,7 +347,142 @@ namespace AppInstaller::CLI::Workflow
...
+        {
+            return;
+        }
+
+        const auto& rootDependencies = context.Get<Execution::Data::Installer>()->Dependencies;
+        DWORD result = 0;
+        bool force = context.Args.Contains(Execution::Args::Type::Force);
+        bool rebootRequired = false;
+
+        rootDependencies.ApplyToType(DependencyType::WindowsFeature, [&context, &result, &force, &rebootRequired](Dependency dependency)
```
yao-msft, line 382: I think we should keep this ApplyToType part in DependenciesFLow.cpp so all dependencies related logic are in that flow. ShellExecuteInstallerHandler.cpp will only expose EnableWindowsFeature flow for enabling only 1 feature, so in the future if we are to implement winget features --enable --disable, we can reuse the same code.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -307,7 +347,142 @@ namespace AppInstaller::CLI::Workflow
...
+                if (!enableFeatureResult)
+                {
+                    // User aborted the operation.
+                    AICLI_TERMINATE_CONTEXT(E_ABORT);
+                }
+
+                auto enableFeatureExitCode = enableFeatureResult.value();
+                if (enableFeatureExitCode == ERROR_SUCCESS_REBOOT_REQUIRED)
+                {
+                    AICLI_LOG(Core, Info, << "Reboot required for [" << featureName << "] with exit code: " << result);
```
yao-msft, line 435:
```suggestion
                    AICLI_LOG(Core, Info, << "Reboot required for [" << featureName << ']');

```
The result is not assigned, it's the value of last run

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -307,7 +347,142 @@ namespace AppInstaller::CLI::Workflow
...
+                else if (FAILED(enableFeatureExitCode))
+                {
+                    AICLI_LOG(Core, Error, << "Failed to enable Windows Feature [" << featureName << "] with exit code: " << enableFeatureExitCode);
+                    context.Reporter.Warn() << Resource::String::FailedToEnableWindowsFeature(locIndFeatureName) << std::endl
+                        << GetUserPresentableMessage(enableFeatureExitCode) << std::endl;
+                    result = enableFeatureExitCode;
+                }
+                else
+                {
+                    AICLI_LOG(Core, Info, << "Successfully enabled [" << featureName << "] with exit code: " << result);
```
yao-msft, line 447:
```suggestion
                    AICLI_LOG(Core, Info, << "Successfully enabled [" << featureName << ']');

```
The result is not assigned, it's the value of last run

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
ryfu-msft: I moved the ApplyToType part back to the DependenciesFlow as suggested. I created a separate workflow just for enabling a single feature that way it can be reused in the future.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -307,7 +347,142 @@ namespace AppInstaller::CLI::Workflow
             else
             {
                 context.Add<Execution::Data::OperationReturnCode>(uninstallResult.value());
+            }
+        }
+    }
+
+#ifndef AICLI_DISABLE_TEST_HOOKS
+    std::optional<DWORD> s_EnableWindowsFeatureResult_Override{};
```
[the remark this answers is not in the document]
ryfu-msft: There is another test hook that also passes an optional value like so:
`void TestHook_SetPinningIndex_Override(std::optional<std::filesystem::path>&& indexPath);`

Since I am also passing an optional DWORD value, I just did something similar to that.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -201,6 +201,45 @@ namespace AppInstaller::CLI::Workflow
 
             return args;
         }
+
+        std::filesystem::path GetDismExecutablePath()
+        {
+            return { ExpandEnvironmentVariables(L"%windir%\\system32\\dism.exe") };
```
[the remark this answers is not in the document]
ryfu-msft: Changed to calling GetExpandedPath from Filesystem.h

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
ryfu-msft: I replaced the FAILED(result) check with a boolean value that is only set to true if we encounter an error. Since there are only a few error codes that we are looking for that change our output behavior to the user, I felt that this was an easier approach rather than checking for failed exit codes.

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.h
```diff
@@ -34,4 +34,18 @@ namespace AppInstaller::CLI::Workflow
...
     // Outputs: InstallerArgs
     void GetInstallerArgs(Execution::Context& context);
+
+    // Enables the Windows Feature dependency by invoking ShellExecute on the DISM executable.
+    // Required Args: None
+    // Inputs: Windows Feature dependency
+    // Outputs: None
+    struct ShellExecuteEnableWindowsFeature : public WorkflowTask
+    {
+        ShellExecuteEnableWindowsFeature(const std::string& featureName) : WorkflowTask("ShellExecuteEnableWindowsFeature"), m_featureName(featureName) {}
```
yao-msft, line 44:
```suggestion
        ShellExecuteEnableWindowsFeature(std::string_view featureName) : WorkflowTask("ShellExecuteEnableWindowsFeature"), m_featureName(featureName) {}

```

yao-msft (approved):

src/AppInstallerCLICore/Workflows/DependenciesFlow.cpp
```diff
@@ -136,82 +135,85 @@ namespace AppInstaller::CLI::Workflow
...
-            context << Workflow::EnsureRunningAsAdmin;
-            if (context.IsTerminated())
+            return;
+        }
+
+        bool enableFeaturesFailed = false;
+        bool force = context.Args.Contains(Execution::Args::Type::Force);
+        bool rebootRequired = false;
+
+        rootDependencies.ApplyToType(DependencyType::WindowsFeature, [&context, &enableFeaturesFailed, &force, &rebootRequired](Dependency dependency)
```
yao-msft, line 154: I think the context should be from a subcontext

yao-msft (approved):

PR #3665, opened 2023-09-24
Log Com invocation startup telemetry and delay auto update time when invoked from explorer

yao-msft:
- Add telemetry event for PackageManager class creation
- Delay source update time to 7 days by default when invoked from explorer
- Also fixes Source agreements related crash found when doing the above work
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3665)

### DIFF
--- a/src/Microsoft.Management.Deployment/Helpers.cpp
+++ b/src/Microsoft.Management.Deployment/Helpers.cpp
@@ -118,4 +118,29 @@ namespace winrt::Microsoft::Management::Deployment::implementation
 
         return {};
     }
+
+    std::string GetCallerName()
+    {
+        // See if caller name is set by caller
+        std::string callerName = GetComCallerName("");
+
+        // Get process string
+        if (callerName.empty())
+        {
+            try
+            {
+                auto [hrGetCallerId, callerProcessId] = GetCallerProcessId();
+                THROW_IF_FAILED(hrGetCallerId);
+                callerName = AppInstaller::Utility::ConvertToUTF8(TryGetCallerProcessInfo(callerProcessId));
+            }
+            CATCH_LOG();
+        }
+
+        if (callerName.empty())
+        {
+            callerName = "UnknownComCaller";
+        }
+
+        return callerName;
+    }
 }
--- a/src/Microsoft.Management.Deployment/Helpers.cpp
+++ b/src/Microsoft.Management.Deployment/Helpers.cpp
@@ -130,8 +130,10 @@ namespace winrt::Microsoft::Management::Deployment::implementation
             try
             {
                 auto [hrGetCallerId, callerProcessId] = GetCallerProcessId();
-                THROW_IF_FAILED(hrGetCallerId);
-                callerName = AppInstaller::Utility::ConvertToUTF8(TryGetCallerProcessInfo(callerProcessId));
+                if (SUCCEEDED(hrGetCallerId))
+                {
+                    callerName = AppInstaller::Utility::ConvertToUTF8(TryGetCallerProcessInfo(callerProcessId));
+                }
             }
             CATCH_LOG();
         }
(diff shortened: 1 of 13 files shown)


src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -118,4 +118,29 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        // See if caller name is set by caller
+        std::string callerName = GetComCallerName("");
+
+        // Get process string
+        if (callerName.empty())
+        {
+            try
+            {
+                auto [hrGetCallerId, callerProcessId] = GetCallerProcessId();
+                THROW_IF_FAILED(hrGetCallerId);
```
JohnMcPMS, line 133: Don't use exceptions for control flow.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -15,42 +15,42 @@
...
+        try
+        {
+            auto [hrGetCallerId, callerProcessId] = GetCallerProcessId();
+            THROW_IF_FAILED(hrGetCallerId);
+            if (callerProcessId != GetCurrentProcessId())
             {
-                try
+                // OutOfProc case, we check for explorer.exe
+                auto callerNameWide = AppInstaller::Utility::ConvertToUTF16(GetCallerName());
+                auto processName = AppInstaller::Utility::ConvertToUTF8(std::filesystem::path{ callerNameWide }.filename().wstring());
```
JohnMcPMS, line 40: Please move this detection to a shared utility function as I suspect we will need it again (and I want to make it very explicit that we are doing this).

At the same time, wrap it in a `try/catch` so that we can ensure that an unexpected caller value being set doesn't blow up `path` parsing or something.  The values we are targeting should be ok, and so any exception can be treated as not the case we are looking for.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -140,8 +142,13 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
             else
             {
+                if (!m_acceptSourceAgreements && SourceAgreements().Size() != 0)
+                {
+                    return GetConnectSourceAgreementsNotAcceptedErrorResult();
+                }
+
                 source = m_sourceReference;
-                source.SetCaller(GetCallerName());
+                source.SetCaller(callerName);
```
JohnMcPMS, line 151: Don't you need to call `SetBackgroundUpdateInterval` here?  I'm not seeing it being set anywhere else...

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -15,42 +15,42 @@
...
 namespace winrt::Microsoft::Management::Deployment::implementation
 {
-    namespace
+    void PackageCatalogReference::Initialize(winrt::Microsoft::Management::Deployment::PackageCatalogInfo packageCatalogInfo, ::AppInstaller::Repository::Source sourceReference)
     {
-        std::string GetCallerName()
-        {
-            // See if caller name is set by caller
-            static auto callerName = GetComCallerName("");
+        static constexpr winrt::Windows::Foundation::TimeSpan s_PackageCatalogUpdateIntervalDelay = 168h; //1 week
```
JohnMcPMS, line 25: We might want to randomize this a bit as well, at least + 0/24 hours (worth of seconds or smaller) but maybe even up to a week extra.  That will hopefully prevent this from causing time based scheduled tasks to align on updates.

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -143,4 +145,26 @@ namespace winrt::Microsoft::Management::Deployment::implementation
 
         return callerName;
     }
+
+    bool ShouldDelayBackgroundUpdateByCaller()
```
JohnMcPMS, line 149: I really meant with a name that was like `IsBackgroundProcessForPolicy` or something.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -22,33 +22,19 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
-        }
-        CATCH_LOG();
+            static constexpr winrt::Windows::Foundation::TimeSpan s_PackageCatalogUpdateIntervalDelay_Base = 168h; //1 week
 
-        if (shouldDelayBackgroundUpdateInterval)
-        {
-            m_packageCatalogBackgroundUpdateInterval = s_PackageCatalogUpdateIntervalDelay;
+            // Add a bit of randomness to the default interval time
+            std::default_random_engine randomEngine(std::random_device{}());
+            std::uniform_int_distribution<long long> distribution(0, 168);
```
JohnMcPMS, line 35: We really want to spread this out, so multiple this by 3600 for second resolution.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -39,6 +39,15 @@ using namespace ::AppInstaller::CLI::Execution;
 
 namespace winrt::Microsoft::Management::Deployment::implementation
 {
+    PackageManager::PackageManager()
+    {
+        auto previousThreadGlobals = m_threadGlobals.SetForCurrentThread();
+        // Immediately reset as we only want the thread globals for logging within this object.
+        previousThreadGlobals.reset();
+        m_threadGlobals.GetTelemetryLogger().SetCaller(GetCallerName());
```
JohnMcPMS, line 47: Make sure that this *doesn't* send a summary event.  There is a bool that disables it as I remember.

JohnMcPMS (approved):

PR #3666, opened 2023-09-25
Reduce the size of the index

JohnMcPMS:
## Change
This change seeks to reduce the size of the index in two ways:
1. Better schema design for the 1:N mapping tables
2. Dropping some of the mapping data that is not particularly interesting per manifest (aka per version)

### Better schema design
This is achieved by the map table having no `rowid` and using a primary key with the value first.  This makes the table already sorted by the value, thus the reverse lookups are fast.  It also drops a fair amount of the data in the table itself to remove the `rowid`, given that it was ~1/3 of the rows.

### Dropping map data
We don't actually use the fact that we know that different versions have different tags (or any other data).  Thus, we can simply have one manifest entry per package identifier have all of the values and maintain the same functionality.  There is a slight loss of fidelity if one is reading through the values via API, but this is deemed acceptable given the large data savings.  I explicitly left the product codes alone, as this does have value to keep per version (even if we are not using it currently).

### Size comparison
| State | Size (bytes) | Percentage | Delta (relative) |
| --- | --- | --- | --- |
| Original | 18141184 | 100% | |
| Better schema | 12673024 | 70% | -30% |
| Map folding | 9256960 | 51% | -19% (-27%) |
| File name shortened | 8306688 | 45.8% | -5.2% (-10.3%) |

As a bonus, we plan to also shorten the file names of the manifests, but this is a service only change.

## Validation
Tests are added to verify the behavior is as expected for the various folded data.
The regression tests should help ensure that the schema changes are not functionally impactful.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3666)

### DIFF
--- a/src/AppInstallerCLITests/SQLiteIndex.cpp
+++ b/src/AppInstallerCLITests/SQLiteIndex.cpp
@@ -311,6 +311,17 @@ bool AreArpVersionsSupported(const SQLiteIndex& index, const Schema::Version& te
     return (index.GetVersion() >= Schema::Version{ 1, 5 } && testVersion >= Schema::Version{ 1, 5 });
 }
 
+bool IsMapDataFoldingSupported(const SQLiteIndex& index, const Schema::Version& testVersion)
+{
+    UNSCOPED_INFO("Index " << index.GetVersion() << " | Test " << testVersion);
+    return (index.GetVersion() >= Schema::Version{ 1, 7 } && testVersion >= Schema::Version{ 1, 7 });
+}
+
+bool IsMapDataFolded(const SQLiteIndex& index)
+{
+    return (index.GetVersion() >= Schema::Version{ 1, 7 });
+}
+
 std::string GetPropertyStringByKey(const SQLiteIndex& index, SQLite::rowid_t id, PackageVersionProperty property, std::string_view version, std::string_view channel)
 {
     auto manifestId = index.GetManifestIdByKey(id, version, channel);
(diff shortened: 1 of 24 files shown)


src/AppInstallerCLITests/SQLiteIndex.cpp
```diff
@@ -35,49 +35,39 @@ SQLiteIndex CreateTestIndex(const std::string& filePath, std::optional<Schema::V
     // If no specific version requested, then use generator to run against the last 3 versions.
     if (!version)
     {
-        version = GENERATE(Schema::Version{ 1, 2 }, Schema::Version{ 1, 3 }, Schema::Version{ 1, 4 }, Schema::Version::Latest());
+        version = GENERATE(Schema::Version{ 1, 5 }, Schema::Version{ 1, 6 }, Schema::Version::Latest());
```
Trenly, line 38: It seem to me that because this is always for the last 3 versions, that there should be an operator method to be able to subtract versions from each other; something like `Schema::Version::Latest() - Schema::Version{0, 2}` so these don't need to be updated manually

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.cpp
```diff
@@ -35,49 +36,53 @@ namespace AppInstaller::Repository::Microsoft
...
         }
-        else if (m_version == Version{ 1, 5 })
+        else if (version == Version{ 1, 5 })
         {
             return std::make_unique<V1_5::Interface>();
         }
-        else if (m_version == Version{ 1, 6 } ||
-            m_version.MajorVersion == 1 ||
-            m_version.IsLatest())
+        else if (version == Version{ 1, 6 })
```
Trenly, line 67: Should these be in reverse order to optimize the comparisons? I assume the newer schema versions would be more common, so having them at the top of the comparisons would result in slightly fewer comparisons

JohnMcPMS: I did one better.

src/AppInstallerCLITests/SQLiteIndex.cpp
```diff
@@ -35,39 +35,53 @@ SQLiteIndex CreateTestIndex(const std::string& filePath, std::optional<Schema::V
     // If no specific version requested, then use generator to run against the last 3 versions.
     if (!version)
     {
-        version = GENERATE(Schema::Version{ 1, 5 }, Schema::Version{ 1, 6 }, Schema::Version::Latest());
+        Schema::Version latestVersion = SQLiteIndex::GetLatestVersion();
+        if (latestVersion.MajorVersion != 1)
+        {
+            throw std::exception("You added major version 2, figure out how to deal with these tests that do back compat coverage!");
```
Trenly, line 41: I love this!

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndex.cpp
Trenly: That you did! Much cleaner!

src/AppInstallerCLITests/SQLiteIndex.cpp
JohnMcPMS: And most likely I'm the one who will be greeted by that message 😄

msftrubengu (approved):

PR #3673, opened 2023-09-26
Fix localized strings output

mdanish-kh:
- [x] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs).
- [ ] This pull request is related to an issue.

Current output:
image

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3673)

PR #3685, opened 2023-09-27
Cherry pick 3673 into 1.6

Trenly:
#3673 

cc @Johnmcpms / @ranm-msft 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3685)


JohnMcPMS:
We can't get updated localizations fast enough for the changes we want to make here.  The line endings I would take, but I don't love the idea of the english and other languages being out of sync.

Trenly:
> We can't get updated localizations fast enough for the changes we want to make here. The line endings I would take, but I don't love the idea of the english and other languages being out of sync.

I can revise so its just the line endings

JohnMcPMS (approved):

PR #3718, opened 2023-09-30
Honor 429 Retry-After

msftrubengu:
Based on [JohnMcPMS/honor-response](https://github.com/JohnMcPMS/winget-cli/tree/honor-response)

### Change

This pull request honors 429 Retry-After for pre indexed sources and throws `ServiceUnavailableException` when the response is 429 or 503.

- If there's a Retry-After header and its value is more than 60s or the date is more than 60s into the future
  - Save the next time it can be updated in the source details and not try again for that command.
  - Subsequent updates or open that source will be skipped if the time lapsed hasn't passed.
- If there's a Retry-After header and its value is less than 60s or the date value is less than 60 seconds into the future
  - Retry once after that amount of time.
- If no Retry-After header
  - Retry after a random amount of time.

For open, the failed to update source message will show up.
For update, a warning with Unavailable will show up.
The next available time will show up in the logs.

### Test
I ran manual validation using Fiddler to intercept winget's source.msix and return 429 or 503 with different Retry-Header

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3718)


src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -34,52 +34,127 @@ namespace AppInstaller::Repository
...
+
         template <typename MemberFunc>
-        bool AddOrUpdateFromDetails(SourceDetails& details, MemberFunc member, IProgressCallback& progress)
+        AddOrUpdateResult AddOrUpdateFromDetails(SourceDetails& details, MemberFunc member, IProgressCallback& progress)
         {
-            bool result = false;
+            AddOrUpdateResult result;
+
+            // Do not update if we are still before the update block time.
+            if (IsUpdateSuppressed(details))
```
JohnMcPMS, line 70: I had intentionally left it open for a manual update to still attempt this.  We are caught in the middle of respecting the user's request and the server's.  I wouldn't mind it being "harder" for the client to do (aka `--force`), but I feel like we should leave that option open.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -34,52 +34,127 @@ namespace AppInstaller::Repository
...
             // Attempt; if it fails, wait a short time and retry.
             try
             {
-                result = (factory.get()->*member)(details, progress);
-                if (result)
+                result.UpdateChecked = (factory.get()->*member)(details, progress);
+                if (result.UpdateChecked)
                 {
                     details.LastUpdateTime = std::chrono::system_clock::now();
+                    result.MetadataWritten = true;
```
JohnMcPMS, line 89: On success, we should probably zero out `DoNotUpdateBefore` to prevent odd scenarios (unless we are trying to support them).  For instance, if the 429/503 response said "wait 2 days", then a manual update attempt succeeded an hour later, would we still want to continue waiting 2 days for background updates?

JohnMcPMS (commented):
Code that could still use generating the exceptions:
1. Repo\Rest\Schema\HttpClientHelper.cpp :: ValidateAndExtractResponse
2. Common\HttpStream\HttpClientWrapper.cpp :: both PopulateInfoAsync and SendHttpRequestAsync
3. Common\DODownloader.cpp :: Somewhere in here after figuring out how they tell us about a 429

Code that could still be updated to handle the exception:
1. Manifest retrieval
2. Installer download

For this change I would request exception generation 1 and 2 be implemented.  The DO downloader is probably already handling this reasonably well, so we can wait until we have actual issues.  I would also request that we handle it in the installer download, in the event that the user has set us to use wininet.  The manifest retrieval has no retry, nor would it be easy or make much sense to integrate manifest retrieval with the index update.

msftrubengu:
Updated for
1. Repo\Rest\Schema\HttpClientHelper.cpp :: ValidateAndExtractResponse
2. Common\HttpStream\HttpClientWrapper.cpp :: both PopulateInfoAsync and SendHttpRequestAsync

For SendHttpRequestAsync is going to be more complicated to get the RetryCount without weird things.

I believe the installer download was already covered, no? If the setting is wininet it will go to WinINetDownloadToStream which I already implemented.

Missing manifest retrieval and verify ValidateAndExtractResponse

src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -409,12 +498,28 @@ namespace AppInstaller::Utility
...
+                THROW_IF_FAILED(CreateStreamOverRandomAccessStream(
+                    rasAsIUnknown,
+                    IID_PPV_ARGS(inputStream.ReleaseAndGetAddressOf())));
+            }
+            catch (const winrt::hresult_error& hre)
+            {
+                if (hre.code() == APPINSTALLER_CLI_ERROR_SERVICE_UNAVAILABLE)
+                {
+                    auto retryAfter = httpRandomAccessStream->RetryAfter();
+                    THROW_EXCEPTION(AppInstaller::Utility::ServiceUnavailableException(retryAfter));
```
JohnMcPMS, line 518: Did you verify this one?

src/AppInstallerCommonCore/MsixInfo.cpp
```diff
@@ -531,6 +531,8 @@ namespace AppInstaller::Msix
     MsixInfo::MsixInfo(std::string_view uriStr)
     {
         m_stream = Utility::GetReadOnlyStreamFromURI(uriStr);
+
+        // Here if thrown APPINSTALLER_CLI_ERROR_SERVICE_UNAVAILABLE somehow get RetryAfter
```
JohnMcPMS, line 535: No, this code should just let the exception escape.

src/AppInstallerCommonCore/Public/AppInstallerProgress.h
```diff
@@ -70,6 +70,9 @@ namespace AppInstaller
 
         // Sets a cancellation function that will be called when the operation is to be cancelled.
         [[nodiscard]] virtual CancelFunctionRemoval SetCancellationFunction(std::function<void()>&& f) = 0;
+
+        // Waits with cancellation. Returns false if cancelled before timeout.
+        virtual bool Wait(std::chrono::milliseconds millisecondsToWait) = 0;
```
JohnMcPMS, line 75: I meant a standalone helper function, or a static if you prefer.  It definitely doesn't need to be on the interface itself.

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: This is still here...

JohnMcPMS (commented):
I meant catching the exception in `DownloadInstallerFile` and waiting for the appropriate amount of time rather than the arbitrary (and short) time being waited for in that retry.

JohnMcPMS:
> For SendHttpRequestAsync is going to be more complicated to get the RetryCount without weird things.

Ok, the HEAD request handling it should hopefully cover it, and it is no longer the primary case anyway.

src/AppInstallerCommonCore/Downloader.cpp
msftrubengu: Yes

src/AppInstallerSharedLib/Public/AppInstallerErrors.h
```diff
@@ -125,6 +125,7 @@
 #define APPINSTALLER_CLI_ERROR_APPTERMINATION_RECEIVED              ((HRESULT)0x8A15006A)
 #define APPINSTALLER_CLI_ERROR_DOWNLOAD_DEPENDENCIES                ((HRESULT)0x8A15006B)
 #define APPINSTALLER_CLI_ERROR_DOWNLOAD_COMMAND_PROHIBITED          ((HRESULT)0x8A15006C)
+#define APPINSTALLER_CLI_ERROR_SERVICE_UNAVAILABLE                  ((HRESULT)0x8A15006D)
```
yao-msft, line 128: nit: please also update https://github.com/microsoft/winget-cli/blob/master/doc/windows/package-manager/winget/returnCodes.md

ryfu-msft (dismissed):

src/AppInstallerCommonCore/Downloader.cpp
```diff
@@ -16,12 +16,80 @@
...
+            }
+
+            return 0s;
+        }
+
+        std::chrono::seconds GetRetryAfter(const wil::unique_hinternet& urlFile)
+        {
+            std::wstring retryAfter = {};
+            DWORD length = 0;
+            if (!HttpQueryInfo(urlFile.get(),
```
yao-msft, line 59: nit: We usually explicitly use wide version for all win32 apis

HttpQueryInfoW and below usage

yao-msft (approved):

PR #3738, opened 2023-10-06
Make manifest retrieval choice more dynamic

JohnMcPMS:
## Change
This change lets the source implementation indicate whether manifests might contain more information than the search results.  It then only applies that to REST, as the implementation is not under our control.  This should have no net effect on the results quality, but should not attempt to retrieve manifests from the winget source when they are not providing any additional value currently.

## Validation
Manually ensured no manifest retrieved for winget based package lookup in dev when one was done for the current release.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3738)

### DIFF
--- a/src/AppInstallerCLITests/CompositeSource.cpp
+++ b/src/AppInstallerCLITests/CompositeSource.cpp
@@ -152,6 +152,12 @@ struct TestPackageHelper
         return *this;
     }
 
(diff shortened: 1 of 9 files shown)


yao-msft (approved):
:shipit:

PR #3759, opened 2023-10-12
Use package version as potential last update timestamp

JohnMcPMS:
## Change
This change moves the background update determination to be under the control of the source reference, and then uses the package version (which is based on the creation time for us) as a potential later "last update time".  This enables an update to the package from outside of our control to still prevent any update check from occurring.

## Validation
New tests were added for the version to `time_point` conversion.
Manually confirmed functionality against winget source.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3759)


src/AppInstallerRepositoryCore/SourceUpdateChecks.h
```diff
@@ -0,0 +1,20 @@
...
+namespace AppInstaller::Repository
+{
+    // Determines if the given source should update before opening.
+    bool ShouldUpdateBeforeOpen(ISourceReference* sourceReference, const std::optional<TimeSpan>& requestedUpdateInterval);
+
+    // Determines if the current time is before a previously stored "do note update before" time.
+    bool IsBeforeDoNotUpdateBeforeTime(const SourceDetails& details);
+
+    // Determines if the given details and desired update interval indicate an update check should occur.
+    bool IsAfterUpdateCheckTime(const SourceDetails& details, std::optional<TimeSpan> requestedUpdateInterval);
```
yao-msft, line 16: It seems this one is never being called.

yao-msft (approved):

PR #3763, opened 2023-10-12
Allow higher versions to satisfy the VCLibs dependency in Repair

JohnMcPMS:
## Change
Rather than looking for the explicit version of the VCLibs dependency, this change allows any version greater than or equal to satisfy the installed check.

This does not resolve the fact that we are using a hardcoded version, nor that we are not checking for the same set of architectures that we would install.

Also fixes the helper script to import the dev modules directly so that it can be used even when the release modules are present in the module path.

## Validation
Stepped through to confirm, added debug output as well.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3763)


dmgonch (approved):
:shipit:

yao-msft (approved):

PR #3787, opened 2023-10-19
Extend Configuration COM API for export

florelis:
This extends the Configuration COM API for the "export" scenario. Most of the changes follow the same pattern as for Get, except that I did not add a new Intent to the API.

* Extended the interface for the configuration unit processor to be able to export
* Implemented all the plumbing between the unit processor and the actual invoke of the dsc resource export, which is currently marked as not supported.
* Added some unit tests for export; this tests only the plumbing as there is no new functionality added yet
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3787)


src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
```diff
@@ -395,6 +410,13 @@ namespace Microsoft.Management.Configuration
         IApplySettingsResult ApplySettings();
     }
 
+    [contract(Microsoft.Management.Configuration.Contract, 2)]
+    interface IConfigurationUnitProcessor2
```
JohnMcPMS, line 414:
```suggestion
    interface IConfigurationUnitProcessor2 requires IConfigurationUnitProcessor
```
Although it apparently doesn't do much, this is the way to chain interfaces.

JohnMcPMS (commented):
This is missing the front end API surface for callers to operate on.  Is it coming in a different PR?

florelis:
> This is missing the front end API surface for callers to operate on. Is it coming in a different PR?

I knew I had to be missing something... Added.

src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
```diff
@@ -395,6 +410,13 @@ namespace Microsoft.Management.Configuration
         IApplySettingsResult ApplySettings();
     }
 
+    [contract(Microsoft.Management.Configuration.Contract, 2)]
+    interface IConfigurationUnitProcessor2 requires IConfigurationUnitProcessor
```
JohnMcPMS, line 414: ###### Sorry for not seeing this earlier...
Maybe this should be an explicit interface for `GetAllSettings`, rather than an interface in the generic evolution of the unit processor.  That way we can use it at the orchestration layer to know that export isn't expected to work rather than calling and having each processor need to give us the exact same error.

src/Microsoft.Management.Configuration/ConfigurationProcessor.cpp
```diff
@@ -626,6 +627,70 @@ namespace winrt::Microsoft::Management::Configuration::implementation
...
+        result->ResultInformation(*unitResult);
+
+        cancellation.ThrowIfCancelled();
+
+        IConfigurationUnitProcessor2 unitProcessor;
+
+        try
+        {
+            // TODO: Directives overlay to prevent running elevated for get
+            unitProcessor = setProcessor.CreateUnitProcessor(unit).as<IConfigurationUnitProcessor2>();
```
JohnMcPMS, line 666: RE: my other comment on changing the name of the interface.  This would become `try_as` and when `!unitProcessor` would set the consistent error in the event of it not being supported.

src/Microsoft.Management.Configuration.Processor/DscModules/DscModuleV2.cs
```diff
@@ -216,6 +216,16 @@ public IReadOnlyList<DscResourceInfoInternal> GetAllDscResources(PowerShell pwsh
             return setResult?.RebootRequired;
         }
 
+        /// <inheritdoc/>
+        public IList<ValueSet> InvokeExportResource(
```
JohnMcPMS, line 220: Again RE: interface.  Almost all (or all) of the changes to the processor changes could be removed with the interface semantic change.

I think that the DSC v3 integration does not benefit from being in .Net, so I don't know that any of this needs to be here when that comes either.

src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
florelis: Renamed the interface and made the unit processor here not implement it anymore.

src/Microsoft.Management.Configuration/ConfigurationProcessor.cpp
```diff
@@ -626,6 +627,59 @@ namespace winrt::Microsoft::Management::Configuration::implementation
...
+
+        IConfigurationSetProcessor setProcessor = m_factory.CreateSetProcessor(nullptr);
+        auto result = make_self<wil::details::module_count_wrapper<implementation::GetAllConfigurationUnitSettingsResult>>();
+        auto unitResult = make_self<wil::details::module_count_wrapper<implementation::ConfigurationUnitResultInformation>>();
+        result->ResultInformation(*unitResult);
+
+        cancellation.ThrowIfCancelled();
+
+        IGetAllSettingsConfigurationUnitProcessor unitProcessor;
+        if (setProcessor.CreateUnitProcessor(unit).try_as<IGetAllSettingsConfigurationUnitProcessor>(unitProcessor))
```
JohnMcPMS, line 662: I don't think that we should return a default constructed result if the processor doesn't support export; it should be an error and a specific `HRESULT`.  Probably a new one like ...`_NOT_SUPPORTED_BY_PROCESSOR` or some variant.

Alternately, we could add another field to the result object, but I think the error is cleaner.

This currently makes it seem like there were no results instead of a clear indication that you simply cannot do that.

JohnMcPMS (approved):

JohnMcPMS (approved):

PR #3789, opened 2023-10-19
Fix non-test hook code for Windows Feature

JohnMcPMS:
## Change
Move the Windows Feature `dism.exe` calling code around and make the test hook return from the primary function.  This ensures that all code is being compiled when test hooks are enabled, and the test hooks are simple, early returns.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3789)


yao-msft (approved):
:shipit:

PR #3790, opened 2023-10-19
Move SQLite base code

JohnMcPMS:
## Change
This change moves the core SQLite code to the shared lib.  It changes the namespace of all of the moved files, but the goal is absolutely zero functionality difference for now.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3790)


src/AppInstallerRepositoryCore/Microsoft/Schema/1_7/Interface.h
```diff
@@ -8,6 +8,10 @@ namespace AppInstaller::Repository::Microsoft::Schema::V1_7
 {
     using namespace std::string_view_literals;
 
+    // Version 1.7
+    static constexpr std::string_view s_MetadataValueName_MapDataFolded = "mapDataFolded"sv;
+    static constexpr char s_MetadataValue_MapDataFolded_Separator = ';';
```
yao-msft, line 13: I guess we would want to remove the corresponding entries in MetadataTable.h?

yao-msft (approved):

src/AppInstallerRepositoryCore/Microsoft/Schema/1_7/Interface.h
JohnMcPMS: They were removed.

Issue #3667, opened 2023-09-25
Windows Package Manager Server runs on high CPU


MarvinNorway:
### Brief description of your issue

I'm not sure if this is the right place to report this, but the "WinGET COM Server" with a child process of "WindowsPackageManagerServer.exe" is using up a lot of CPU on my system.
When killed, it just respawns immediately.

I'm not the only one with this issue:
https://www.reddit.com/r/techsupport/comments/162dj42/what_is_winget_com_server/

### Steps to reproduce

Not sure how to reproduce.

### Expected behavior

It shouldn't run on high CPU for extended periods of time.

### Actual behavior

It runs on super high CPU all the time.

### Environment

```shell
Microsoft.DesktopAppInstaller_1.20.2201.0_x64__8wekyb3d8bbwe
WindowsPackageManagerServer.exe version 1.20.2308.8001-v1.20WPMv1.5

Hope that helps...
```

sba923:
Same here. The process keeps running after I close its parent (a `pwsh.exe` instance). I can't copy the log from `%LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir` 'cos it's locked.

MozPri:
Same here. WindowsPackageManagerServer.exe is using 25% of the CPU and can't be killed.
Microsoft.DesktopAppInstaller_1.21.2701.0

PartedLine:
Same here, rather not have my OS frying my CPU

saggitar:
Same here ...

HyP3r-:
It is the same with me... when I end the process it is restarted immediately.

MarvinNorway:
Not sure if it helps anyone, but I left it running over night, and then it was quite the next day.
It happened one more time 1-2 days later, during the day, but I went out for a bit, and when I came back, it was over, as well.
Since then, it has been quiet. Would still be nice to know what the process does and if there's a way to stop this behavior...

chylex:
Mine has started crashing constantly, trying to stop it just restarts it and crashes again... how do I get rid of this???

![obrazek](https://github.com/microsoft/winget-cli/assets/3685160/09c11de2-534c-441d-ae8f-897dc6917d5c)

gibbed:
> Mine has started crashing constantly, trying to stop it just restarts it and crashes again... how do I get rid of this???
> 
[...]

PR #3808, opened 2023-10-24
Performance improvements

JohnMcPMS:
## Change
This change has several performance improvements based on profiling.

### Installed package caching
A global process cache is stored for the unfiltered install data.  It is very basic for updates right now to make it more reasonable to service, but future changes could improve things to examine the system data and only do delta updates.  It could also be saved to disk to improve the performance of future processes.

The cache itself is not used by any callers, but instead in-memory copies are handed out.  The cache sets up event listeners for the current data sources and waits for these events before it will attempt to update again.  A full cache hit requires only copying the memory database, which SQLite has an efficient method for.

The only performance gain to the update path that is provided by this implementation is to reuse the names of MSIX packages, but since this accounts for a huge amount of creating the index it results in ~50% time reduction in creating an updated installed source (without any MSIX changes).

### ICU regex caching
The regular expressions that we use were being compiled repeatedly.  Since this is a fixed set of expressions, they are all now cached and copies are used for actual operation.

### Version lookup (string)
The version lookup by string (such as `winget show FOO -v VER`) was doing a full table scan.  This change shifts to pulling all of the version for the given identifier out and using `Version` object comparisons to find the proper result.  This has the side effect of fixing a somewhat annoying user interaction where trailing 0s in a version were still required to find from the command line.  One can now provide the smaller version string and still find the expected package version (eg `1.2` will now find version `1.2.0`).

### Version lookup (available version)
The round trip from retrieving the available version information to getting the manifest has been improved by including the manifest id value directly in the available version data.  The available version package object now caches the version string to manifest id data as well, meaning that any usage of a `PackageVersionKey` that was retrieved from `GetAvailableVersions` will directly return the package version object without querying the index.

### Let property lookups handle an unknown manifest id
The `GetProperty` code was enforcing that the manifest id was valid and then reusing existing code that did not handle a missing manifest.  Creating new methods that can handle the missing manifest allows us to remove the initial check.

## Validation
Added tests where appropriate, regression covers many other cases.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3808)


denelon:
Related to:
* https://github.com/microsoft/winget-cli/issues/3667

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.h
```diff
@@ -34,6 +34,8 @@ namespace AppInstaller::Repository::Microsoft
             User,
             // Contains machine ARP and machine MSIX
             Machine,
+            // Same as None but creating the source reference causes the next Open to always update the cache
+            NoneWithForcedCacheUpdate,
```
florelis, line 38: Could we ever want to use anything else with forced cache update? If not, can you add a comment why not?

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
```diff
@@ -288,14 +288,34 @@ namespace AppInstaller::Repository::Microsoft
...
+                }
+
+                std::vector<SQLiteIndex::VersionKey> versions = source->GetIndex().GetVersionKeysById(m_idId);
+
+                for (const auto& vk : versions)
+                {
+                    std::string version = vk.VersionAndChannel.GetVersion().ToString();
+                    std::string channel = vk.VersionAndChannel.GetChannel().ToString();
+                    m_availableVersionKeys.emplace_back(source->GetIdentifier(), version, channel);
+                    m_availableVersionKeysMap.emplace(MapKey{ std::move(version), std::move(channel) }, vk.ManifestId);
```
florelis, line 315: Why does this one not take the source ID?

src/AppInstallerCLIE2ETests/Interop/InstallInterop.cs
```diff
@@ -657,5 +657,73 @@ public void GetApplicableInstaller()
...
+        /// Install exe and verify that we can find it as installed after.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Test]
+        public async Task InstallExe_VerifyInstalledCatalog()
+        {
+            var installedCatalogReference = this.packageManager.GetLocalPackageCatalog(LocalPackageCatalog.InstalledPackages);
+
+            // Ensure package is not installed
+            string targetPackageProductCode = "{A499DD5E-8DC5-4AD2-911A-BCD0263295E9}";
```
florelis, line 671: Nit: I had to look up what this GUID was. Can this be a constant somewhere else?

src/AppInstallerCLITests/TestHooks.h
```diff
@@ -39,6 +39,9 @@ namespace AppInstaller
     namespace Repository::Microsoft
     {
         void TestHook_SetPinningIndex_Override(std::optional<std::filesystem::path>&& indexPath);
+
+        using GetARPKeyFunc = Registry::Key(*)(void*, Manifest::ScopeEnum, Utility::Architecture);
```
florelis, line 43: Nit: Could we use `std::function` here?

src/AppInstallerCLITests/PredefinedInstalledSource.cpp
```diff
@@ -400,3 +402,89 @@ TEST_CASE("PredefinedInstalledSource_Search", "[installed][list]")
...
+        GetDatabaseIdentifier(source1)
+        ==
+        GetDatabaseIdentifier(source2)
+    );
+
+    // Get all packages
+    auto result1 = source1->Search({});
+    REQUIRE(!result1.Matches.empty());
+
+    for (const auto& match : result1.Matches)
```
florelis, line 427: Nit: Could you add a comment about what this is loop is doing?

florelis, line 427: Could/should we check that there are no packages in result2 that aren't in result1?

src/AppInstallerCLITests/ArpHelper.cpp
```diff
@@ -0,0 +1,71 @@
...
+    ARPHelper helper;
+
+    wil::unique_event callbackEvent;
+    callbackEvent.create();
+
+    ScopeEnum scopeCallback = ScopeEnum::Unknown;
+    Architecture architectureCallback = Architecture::Unknown;
+
+    ScopeEnum scopeTarget = ScopeEnum::User;
+    Architecture architectureTarget = Architecture::X64;
```
florelis, line 26: Will this test run on x86?

florelis (dismissed):

azure-pipelines.yml
```diff
@@ -295,18 +297,18 @@ jobs:
...
+  #   inputs:
+  #     testRunTitle: Microsoft.Management.Configuration.UnitTests (InProc)
+  #     testSelector: 'testAssemblies'
+  #     testAssemblyVer2: '**\Microsoft.Management.Configuration.UnitTests.dll'
+  #     searchFolder: '$(buildOutDir)\Microsoft.Management.Configuration.UnitTests'
+  #     codeCoverageEnabled: true
+  #     platform: '$(buildPlatform)'
+  #     configuration: '$(BuildConfiguration)'
+  #     diagnosticsEnabled: true
+  #   condition: succeededOrFailed()
```
florelis, line 311:
```suggestion
  - task: VSTest@2
    displayName: 'Run tests: Microsoft.Management.Configuration.UnitTests (InProc)'
    inputs:
      testRunTitle: Microsoft.Management.Configuration.UnitTests (InProc)
      testSelector: 'testAssemblies'
      testAssemblyVer2: '**\Microsoft.Management.Configuration.UnitTests.dll'
      searchFolder: '$(buildOutDir)\Microsoft.Management.Configuration.UnitTests'
      codeCoverageEnabled: true
      platform: '$(buildPlatform)'
      configuration: '$(BuildConfiguration)'
      diagnosticsEnabled: true
    enabled: false
    condition: succeededOrFailed()

```

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
JohnMcPMS: It isn't needed for the mapping.

florelis (approved):

yao-msft (approved):

src/AppInstallerSharedLib/Public/winget/SQLiteStorageBase.h
```diff
@@ -25,6 +25,9 @@ namespace AppInstaller::SQLite
         // Gets the last write time for the database.
         std::chrono::system_clock::time_point GetLastWriteTime();
 
+        // Gets the identifier written to the database when it was created.
+        std::string GetDatabaseIdentifier();
```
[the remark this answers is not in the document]
JohnMcPMS: I added it exclusively for test purposes at this time.  But it seemed like an interesting enough idea to just leave it in for all databases.

yao-msft (approved):

Girofox:
A possible trigger for debugging is manually starting backup in **Windows Backup** app, this instantly causes _WindowsPackageManagerServer.exe_ running at around 25 % cpu usage for some time. 

The log file in `%LocalAppData%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir` grows very quickly to more than 1.0 Mb.

PR #3816, opened 2023-10-27
Allow Microsoft.WinGet.Client to run in any PowerShell session running as system

msftrubengu:
This PR adds support for running the Microsoft.WinGet.Client module in system context without the need to start pwsh.exe -MTA. Running as MTA is required using inproc Microsoft.Management.Deployment.

If the module is running inproc and the current thread is not an MTA, it will create a new MTA thread and execute there. Otherwise, non inproc or already an MTA will use the current thread.

This was done by sharing AsyncCommand (renamed to PowerShellCmdlet) from Microsoft.WinGet.Configuration. Originally, I wanted to create a new shared lib, but decided to just share the files between the projects.

All cmdlets must inherit from PowerShellCmdlet. All cmdlets that use Microsoft.Management.Deployment must inherit ManagementDeploymentCommand and use Execute at the command engine entry point. As a safe mechanism, if any call to PackageManagerWrapper will verify the thread is not an STA if running inproc and fail.

I verified the cmdlets work in system context locally. A future PR will add running our pester tests using psexe.exe

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3816)

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/Common/FinderCommand.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/Common/FinderCommand.cs
@@ -35,31 +35,33 @@ namespace Microsoft.WinGet.Client.Engine.Commands.Common
         /// Gets or sets the field that is matched against the identifier of a package.
         /// </summary>
         [Filter(Field = PackageMatchField.Id)]
-        protected string Id { get; set; }
+        protected string? Id { get; set; }
 
         /// <summary>
         /// Gets or sets the field that is matched against the name of a package.
         /// </summary>
         [Filter(Field = PackageMatchField.Name)]
-        protected string Name { get; set; }
+        protected string? Name { get; set; }
 
         /// <summary>
         /// Gets or sets the field that is matched against the moniker of a package.
         /// </summary>
         [Filter(Field = PackageMatchField.Moniker)]
-        protected string Moniker { get; set; }
+        protected string? Moniker { get; set; }
 
         /// <summary>
         /// Gets or sets the name of the source to search for packages. If null, then all sources are searched.
         /// </summary>
-        protected string Source { get; set; }
+        protected string? Source { get; set; }
 
         /// <summary>
         /// Gets or sets how to match against package fields.
         /// </summary>
-        protected string[] Query { get; set; }
+#pragma warning disable SA1011 // Closing square brackets should be spaced correctly
+        protected string[]? Query { get; set; }
+#pragma warning restore SA1011 // Closing square brackets should be spaced correctly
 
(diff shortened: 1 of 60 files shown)


src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/Common/FinderCommand.cs
```diff
@@ -35,31 +35,33 @@ internal FinderCommand(PSCmdlet psCmdlet)
...
         /// </summary>
-        protected string Source { get; set; }
+        protected string? Source { get; set; }
 
         /// <summary>
         /// Gets or sets how to match against package fields.
         /// </summary>
-        protected string[] Query { get; set; }
+#pragma warning disable SA1011 // Closing square brackets should be spaced correctly
+        protected string[]? Query { get; set; }
```
JohnMcPMS, line 61: Academic question: Does `string?[]` mean a non-null array of nullable strings, while `string[]?` means a nullable array of non-null strings?  If so, then `string?[]?` is a lovingly confusing type name 😄

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/PackageManagerWrapper.cs
```diff
@@ -21,7 +22,7 @@ internal sealed class PackageManagerWrapper
     {
         private static readonly Lazy<PackageManagerWrapper> Lazy = new (() => new PackageManagerWrapper());
 
-        private PackageManager packageManager = null;
+        private PackageManager? packageManager = null;
```
JohnMcPMS, line 25: What is the point of making this nullable and then just overriding that everywhere?  Can you not do `null!` to trick it into letting you have a null value?  Maybe you need to do `((PackageManager?)null)!`.

JohnMcPMS (dismissed):

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/PackageManagerWrapper.cs
msftrubengu: `null!` is the way to go

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/Common/FinderCommand.cs
msftrubengu: Yup. There's another place where I needed a non-nullable array with nullable objects, so it was object?[]

JohnMcPMS (approved):

PR #3856, opened 2023-11-05
Anonymize display path in download command output

mdanish-kh:
    No; Thought it made sense for these paths to also respect `anonymizeDisplayedPaths` user setting

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3856)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
@@ -549,6 +549,12 @@ namespace AppInstaller::CLI::Workflow
 
             renamedDownloadedInstaller = downloadDirectory / GetInstallerDownloadOnlyFileName(context);
             Filesystem::RenameFile(installerPath, renamedDownloadedInstaller);
+
+            if (Settings::User().Get<Settings::Setting::AnonymizePathForDisplay>())
+            {
+                Runtime::ReplaceProfilePathsWithEnvironmentVariable(renamedDownloadedInstaller);
+            }
+
             context.Reporter.Info() << Resource::String::InstallerDownloaded(Utility::LocIndView{ renamedDownloadedInstaller.u8string() }) << std::endl;
         }
         else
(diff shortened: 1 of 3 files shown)


yao-msft:
We have many other places that outputs a path (winget log path, installer execution log, etc) during a package operation. And we have an override to output path [specifically](https://github.com/microsoft/winget-cli/blob/b96598a6038ac0fb0f58cf935eb466c4484dea3c/src/AppInstallerCLICore/ChannelStreams.cpp#L142) so any code path can output path if needed. This change is specifically about outputting some info regarding a particular package operation. It does not quite fall into the category of AnonymizeForDisplay.

I think AnonymizeForDisplay is originally for winget --info, so that when user files a bug, in the winget info section to be shared, their output is anonymized. And for specific package operations, we do not have that need yet.

If we believe we should anonymize on every path we output, it needs to be a more complete change to have a systematic way of applying anonymize to every path that winget code outputs, not a specific path in download flow. By the way, we do not anonymize path in the logs (they are not shared with winget team)

yao-msft (changes requested):
If we want to do this, we may need to review all the paths we output so they have consistent behaviors. Currently, anonymization are done in a central location GetPathFor() method, and are mostly for displaying in winget --info.

mdanish-kh:
> This change is specifically about outputting some info regarding a particular package operation. It does not quite fall into the category of AnonymizeForDisplay.

Ah, I misinterpreted the user setting to believe that it applies on every outputted path. The description of the setting strengthened that thought as well.

> I think AnonymizeForDisplay is originally for winget --info, so that when user files a bug, in the winget info section to be shared, their output is anonymized. And for specific package operations, we do not have that need yet.

On the same line of thinking, I would imagine someone doing a demo/video showcasing winget commands and may want the output to be anonymized. But since no one has requested that feature, I don't want to be making assumptions about its use case. I'll close this PR because it doesn’t involve a missed anonymization for the download command output, which was my initial assumption.

PR #3858, opened 2023-11-06
Register restart for resume

ryfu-msft:
Related to: #3165 

Changes:
- This PR calls [RegisterApplicationRestart](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerapplicationrestart) to register winget to restart upon a reboot. The resume command and the resume id are passed as arguments to the winget process so that it can resume from where it left off prior to reboot.
- Registering for a restart will only take place if the result of the installation is REBOOT_REQUIRED_FOR_INSTALL.
- The checkpoint database will only persist if the result of the installation is REBOOT_REQUIRED_FOR_INSTALL.
- Added a new settings value called `maxResumes` that limits the number of times a resume state can be "resumed". This is to ensure that continuous reboots don't occur.
- In order for the process to be restarted, the reboot must be kicked off while the process is still running. The winget process cannot be terminated normally or else Windows Error Reporting will not initiate the restart. My workaround is to have the process sleep while the reboot process is kicked off.
- Moved the InitiateReboot() flow to take place after `--wait` argument is processed so that the user is prompted to continue before a reboot is kicked off.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3858)

### DIFF
--- a/src/AppInstallerCommonCore/Runtime.cpp
+++ b/src/AppInstallerCommonCore/Runtime.cpp
@@ -478,14 +478,12 @@ namespace AppInstaller::Runtime
         case PathName::CLIExecutable:
             result.Path = GetKnownFolderPath(FOLDERID_LocalAppData);
             result.Path /= s_WindowsApps_Base;
-            if (IsReleaseBuild())
-            {
-                result.Path /= s_WinGet_Exe;
-            }
-            else
-            {
-                result.Path /= s_WinGetDev_Exe;
-            }
+            result.Path /= GetPackageFamilyName();
+#if USE_PROD_CLSIDS
+            result.Path /= s_WinGet_Exe;
+#else
+            result.Path /= s_WinGetDev_Exe;
+#endif
             result.Create = false;
             mayBeInProfilePath = true;
             break;


src/AppInstallerRepositoryCore/Public/winget/Checkpoint.h
```diff
@@ -53,6 +54,13 @@ namespace AppInstaller::Checkpoints
             m_checkpointDatabase->SetDataValue(m_checkpointId, dataType, fieldName, values);
         }
 
+        // Update a single existing field value for a data type.
+        void Update(T dataType, const std::string& fieldName, const std::string& value)
+        {
+            m_checkpointDatabase->RemoveDataType(m_checkpointId, dataType);
+            m_checkpointDatabase->SetDataValue(m_checkpointId, dataType, fieldName, { value });
```
JohnMcPMS, line 61: I'm fairly sure that you could fix this with an [upsert](https://sqlite.org/lang_upsert.html), but if that isn't to be done then a transaction should be used to preserve the atomic nature of updating the value.

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -212,7 +212,7 @@ namespace AppInstaller
             WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_INSTALL_NO_NETWORK, "This application requires internet connectivity. Connect to a network then try again."),
             WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_INSTALL_CONTACT_SUPPORT, "This application encountered an error during installation. Contact support."),
             WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_FINISH, "Restart your PC to finish installation."),
-            WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_TO_INSTALL, "Installation failed. Restart your PC then try again."),
+            WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_INSTALL_REBOOT_REQUIRED_FOR_INSTALL, "Installation failed. Restart your PC then try again."),
```
JohnMcPMS, line 215: This symbol is also the name of a localized string resource.  Changing it requires also changing that resource name in `winget.resw`.  Please place a few comments around that indicate this requirement also (here and in the header).

src/AppInstallerRepositoryCore/Public/winget/Checkpoint.h
ryfu-msft: I found this old comment that mentioned upsert: https://github.com/microsoft/winget-cli/blob/cd8e9920bcd5cc28bf9cb1217a82847352da0e09/src/AppInstallerRepositoryCore/Microsoft/Schema/1_1/ManifestMetadataTable.cpp#L94C1-L96

I'm assuming this comment still applies so I went ahead and used a transaction for updating the value.

src/AppInstallerCLICore/Commands/ResumeCommand.cpp
```diff
@@ -99,6 +99,18 @@ namespace AppInstaller::CLI
             AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_CLIENT_VERSION_MISMATCH);
         }
 
+        const auto& resumeCountString = automaticCheckpoint.Get(AutomaticCheckpointData::ResumeCount, {});
+        int resumeCount = std::stoi(resumeCountString);
+        if (resumeCount >= Settings::User().Get<Settings::Setting::MaxResumes>())
+        {
+            context.Reporter.Error() << Resource::String::ResumeLimitExceeded(Utility::LocIndView{ resumeCountString }) << std::endl;
```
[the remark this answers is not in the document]
ryfu-msft: Added the --ignore-resume-limit argument and display the command to execute manually.

src/AppInstallerSharedLib/Errors.cpp
ryfu-msft: Updated localized string resource and added comments to both header and cpp files.

src/AppInstallerCLICore/Workflows/ResumeFlow.cpp
```diff
@@ -16,31 +17,25 @@ namespace AppInstaller::CLI::Workflow
...
-        }
+            int argc = 0;
+            wil::unique_hlocal_ptr<LPWSTR> argv{ CommandLineToArgvW(GetCommandLineW(), &argc) };
+            THROW_LAST_ERROR_IF_NULL(argv);
 
-        if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::RebootRequired))
-        {
-            context.ClearFlags(Execution::ContextFlag::RebootRequired);
+            // RunOnce registry value must start with the full path of the executable.
+            std::string commandLine = Utility::ConvertToUTF8(argv.get()[0]) + " resume -g " + context.GetResumeId();
```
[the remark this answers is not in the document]
ryfu-msft: Created a new `Path` called CLIExecutable that points to the full executable path depending on how it is being invoked.

Packaged -> %LOCALAPPDATA%\Microsoft\WindowsApps\Winget.exe (wingetdev.exe if not release build)
Unpackaged -> Calls on GetBinaryDirectoryPath() and appends `winget.exe`

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -472,6 +475,20 @@ namespace AppInstaller::Runtime
             result = GetPathDetailsForPackagedContext(PathName::LocalState, forDisplay);
             result.Path /= s_CheckpointsDirectory;
             break;
+        case PathName::CLIExecutable:
+            result.Path = GetKnownFolderPath(FOLDERID_LocalAppData);
+            result.Path /= s_WindowsApps_Base;
+            if (IsReleaseBuild())
```
JohnMcPMS, line 481: This is not the right thing to key on here.  We have already overloaded `USE_PROD_CLSIDS` to cover targeting the dev vs prod package, so I'm fine with using that here.

Additionally, you should use the package family name in the path to ensure that it is going to the correct alias.  For instance, `%LOCALAPPDATA%\Microsoft\WindowsApps\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\winget.exe`

JohnMcPMS (approved):

ryfu-msft:
@microsoft-github-policy-service rerun

PR #3868, opened 2023-11-09
Move to latest cppwinrt package across all projects

JohnMcPMS:
## Change
Move all projects that use C++/WinRT, WIL, and C#/WinRT to the latest release.  Then fix all of the build breaks that this caused.

This also re-enables a test that needed a newer version of C++/WinRT to work.

After many attempts to debug the issue with the tests that crash on x64 only, I can only assume that there is some perfect collection of OS and dependency versions that will hopefully change at some point and allow us to re-enable them.  I have tried so many attempts at debugging it and ran it on a similar server VM to no avail.  Given that it is only affecting the inproc tests in x64, I can only assume that it something in codegen somewhere.  Will have to try to enable them later.

## Validation
Local build and unit tests succeeded; PR regression tests should cover the rest.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3868)


yao-msft (approved):
:shipit:

src/AppInstallerSharedLib/AppInstallerLogging.cpp
```diff
@@ -160,24 +160,27 @@ namespace AppInstaller::Logging
     }
 }
 
-std::ostream& operator<<(std::ostream& out, const std::chrono::system_clock::time_point& time)
+namespace std
```
florelis, line 163: Adding to the `std` namespace seems weird

florelis (approved):

src/AppInstallerSharedLib/AppInstallerLogging.cpp
JohnMcPMS: I don't know what properties the latest C++/WinRT is forcing on us, but they ended up breaking most of these `operator<<` overloads that were in the global namespace.  Leveraging [ADL](https://en.cppreference.com/w/cpp/language/adl) to overcome this requires putting these into `std` (or `std::chrono` here).

PR #3885, opened 2023-11-13
Add suggested dictionaries for spellchecking and remove unneeded words

florelis:
Added some pre-existing dictionaries suggested by the spellchecking action (e.g. for `win32`) and removed some that didn't make much sense for us to have (why did we even have `html` and `aws`?). Also removed the words that we don't need to have in our `excludes.txt` or `allow.txt` because they are already in one of the dictionaries (either existing or added).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3885)


florelis:
@jsoref why is the result different here than in [a run from my fork](https://github.com/florelis/winget-cli/actions/runs/6857166675)?

jsoref:
@florelis changes to the workflow file aren't applied until they're merged. -- You'll see that when I make my PR, I include items that will be removed in a subsequent run. 

Also, I should really send a PR to upgrade this repository to .22. -- what I do is copy over https://github.com/check-spelling/spell-check-this/tree/main/.github and work with it until it yields good results. 

Note that I'd counsel against using the win32 dictionary, I've found it contains too many typoed words. -- So bad that I've stopped suggesting it in .22.

florelis:
Thanks @jsoref
I removed the win32 dictionary and added back all the items I had removed from it.
I put all of the items that aren't needed with the new dictionary at the end of `allows.txt` and will remove them in another PR

yao-msft (approved):

PR #3886, opened 2023-11-14
Improve repair

msftrubengu:
This PR makes some improvements to the `Microsoft.WinGet.Client` module

**Use stub preference for AppInstaller**
The default behavior for `Add-AppxPackage` is to install the full package regardless of the preference. This PR adds `-StubPreference UsePreference` (if the cmdlet supports it) when installing the AppInstaller bundle. I originally want to allow the user to set the preference, but I realized I would have to blindly call `winget configure --enable|disable` since I don't know the state. I believe this should be done differently with perhaps different cmdlets and add something like `winget config --info` that prints a json with some information like if its enabled or not and the processor info.

**Add -Force to Repair-WinGetPackage**
On some occasions, `Add-AppxPackage` failed because AppInstaller files were in use. I added a `-Force` switch to the cmdlet that just sets `-ForceTargetApplicationShutdown` on the add appx package call.

**Module async friendly**
All async functions in the module required to be waited. Now that we can start an async context via `RunOnMTA`, I removed all the synchronous wrappers calls from GitHubClient and make it such as only the top-level cmdlet need to wait on the Task. This caused a lot of functions to be awaited, but at the end only the top-level cmdlet should care about manually waiting for the task which I believe is the right thing to do.

**Bring back synchronization with main PowerShell thread**
AppxModuleHelper requires its Appx Module calls to be executed in the main PowerShell thread otherwise a new runtime would need to be created for every cmdlet that uses this class. I brought back the old synchronization mechanism of `Microsoft.WinGet.Configure` to do this. Now any cmdlet that runs asynchronously and requires PowerShell Host can do it without creating its runspace.

**Redo progress for Install/Update/Uninstall-WinGetPackage**
I was not a huge fan of how progress was handled in these cmdlets. Given that I can now create an asynchronous context, I simplify how progress is done and follow a similar pattern as `Microsoft.WinGet.Configure`. The `IAsyncOperationWithProgress` is executed, set its Progress handler, converted to a Task and awaited all in one common method. I also fixed how the progress was calculated so it reflects almost the same as winget. .NET rounded the number so for some package I was 92.3 MB in winget and 92.4 the module (before it was like 97or something).

**Add proper can cancellation support**
Modify the Install/Update/Uninstall cmdlets to properly use PowerShell's `StopProcessing` for cancellation. The Task from `IAsyncOperationWithProgress` uses the PowerShellCmdlet CancellationToken, so everything is handled correctly. Also add cancellation to `Repair-WinGetPackage`. The repair state machine sees if the token is cancelled as the first thing on the loop.

**Download URL with progress.**
The module attempts to install a package using `Add-AppxPackage` with the url of the package. In the case it fails, like in Windows Sandbox, the package is downloaded to disk and then installed via the same cmdlet. Depending on the network, the download could take a while and there's no indication that work is being performed without `-Verbose`. This PR adds a PowerShell progress bar for the download when the response contains a Content-Length property. If cancellation is request during url download, the async calls will be cancelled the file will be deleted.


src/PowerShell/CommonFiles/PowerShellCmdlet.cs
```diff
@@ -239,21 +282,57 @@ internal void Wait(Task runningTask, PowerShellCmdlet? writeCmdlet = null)
...
             {
-                this.ConsumeAndWriteStreams(writeCmdlet);
+                if (this.pwshThreadActionReady.IsSet)
+                {
+                    // Someone needs the main thread.
+                    this.pwshThreadActionReady.Reset();
+
+                    if (this.pwshThreadAction != null)
+                    {
+                        this.pwshThreadAction();
```
JohnMcPMS, line 299: I would suggest that the try/catch be here and the exception rethrown via `WaitMainThreadActionCompletion`.

JohnMcPMS (approved):

src/PowerShell/CommonFiles/PowerShellCmdlet.cs
```diff
@@ -603,6 +596,11 @@ private void WaitMainThreadActionCompletion()
                 this.pwshThreadActionCompleted.WaitHandle,
             });
 
+            if (this.pwshThreadActionException != null)
+            {
+                throw this.pwshThreadActionException;
```
JohnMcPMS, line 601: Should this throw a new exception with `pwshThreadActionException` as the inner exception?  That way we get to keep the original stack trace.

JohnMcPMS (approved):

PR #3896, opened 2023-11-15
Upgrade to check-spelling v0.0.22

jsoref:
- [x] This pull request is related to https://github.com/microsoft/winget-cli/pull/3885#issuecomment-1809439464

-----

Changes:
* general spelling/grammar
* refreshed baseline metadata (roughly [spell-check-this@edccef0](https://github.com/check-spelling/spell-check-this/tree/edccef0d97da7e227dbec1d2c1faac5fd085f313))
* upgraded to [check-spelling v0.0.22](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.22)
   * ⚠️ there's one pattern that triggers in v0.0.21 (as part of this PR) which will not trigger once this PR is merged ([🩹 Pattern suggestions shouldn't trip over active patterns](https://github.com/check-spelling/check-spelling/wiki/Feature:-Suggest-patterns#multiple-patterns-for-a-line))
* refreshed workflow
  * slightly refreshed dictionaries -- I'm removing `cspell:cpp/src/cpp.txt`, but you could use it or `cspell:cpp/dict/cpp.txt` -- I think I missed that #3885 was adding it -- cpp suffers from the same problem as win32 -- too many misspelled words)
  * dropping commenting on pushes (people should migrate to github step summaries)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3896)


.github/actions/spelling/excludes.txt
```diff
@@ -88,7 +130,6 @@
 ^src/Valijson/
 ^src/Xlang/
 ^src/YamlCppLib/
-# Because it doesn't handle argument -Words well
-^tools/CorrelationTestbed/.*\.ps1$
-^tools/COMTrace/ComTrace.wprp$
```
jsoref, line 93: afaict this is no longer a problem, so I'm dropping it

.github/actions/spelling/expect.txt
```diff
@@ -400,35 +461,31 @@ STDMETHODCALLTYPE
...
+tli
 tombstoned
-TOperation
-TOptions
-TProgress
-TResult
-TReturn
 trimstart
-TState
-TStatus
```
jsoref, line 430: These `T[A-Z]` items are handled by a pattern below.

.github/actions/spelling/patterns.txt
```diff
@@ -60,7 +69,7 @@ REQUIRE\(RestHelper::GetRestAPIBaseUri\(".*"\) == L".*"
 # Automatically suggested patterns
 # hit-count: 407 file-count: 78
 # IServiceProvider
-\bI(?=(?:[A-Z][a-z]{2,})+\b)
+\b[IT](?=(?:[A-Z][a-z]{2,})+(?:[A-Z]|\b))
```
jsoref, line 72: There were a handful of `T...` items (flagged earlier), and it seemed worthwhile to add `T` here. There are also a couple of other letters which each have a couple of items, but it didn't really seem worthwhile to include a blanket -- it is worth noting that the pattern could be widened to cover them if that's desired.

.github/workflows/spelling3.yml
```diff
@@ -59,35 +83,41 @@ jobs:
       contents: read
       pull-requests: read
       actions: read
+      security-events: write
```
jsoref, line 86: this is required for sarif reporting

.github/workflows/spelling3.yml
```diff
@@ -59,35 +83,41 @@ jobs:
...
     outputs:
       followup: ${{ steps.spelling.outputs.followup }}
     runs-on: ubuntu-latest
-    if: "contains(github.event_name, 'pull_request') || github.event_name == 'push'"
+    if: ${{ contains(github.event_name, 'pull_request') || github.event_name == 'push' }}
```
jsoref, line 90: this is purely stylistic -- the vscode github integration doesn't like `"..."` whereas GitHub Actions do not mind.

.github/workflows/spelling3.yml
```diff
@@ -59,35 +83,41 @@ jobs:
...
       # note: If you use only_check_changed_files, you do not want cancel-in-progress
       cancel-in-progress: true
     steps:
     - name: check-spelling
       id: spelling
-      uses: check-spelling/check-spelling@v0.0.21
+      uses: check-spelling/check-spelling@v0.0.22
       with:
-        suppress_push_for_open_pull_request: 1
+        suppress_push_for_open_pull_request: ${{ github.actor != 'dependabot[bot]' && 1 }}
```
jsoref, line 100: This will let dependabot PRs warn about when changing the workflow will break things, especially `cspell:...` items.

.github/workflows/spelling3.yml
```diff
@@ -59,35 +83,41 @@ jobs:
...
         checkout: true
         check_file_names: 1
-        spell_check_this: check-spelling/spell-check-this@prerelease
+        spell_check_this: microsoft/winget-cli@master
```
jsoref, line 103: this is mostly irrelevant at this point (it's used when a branch doesn't have check-spelling metadata -- which basically happens when it's first added to a repository). I suppose the alternative would be to remove the lines entirely.

.github/workflows/spelling3.yml
```diff
@@ -59,35 +83,41 @@ jobs:
...
         post_comment: 0
         use_magic_file: 1
-        extra_dictionary_limit: 10
+        warnings: bad-regex,binary-file,deprecated-feature,ignored-expect-variant,large-file,limited-references,no-newline-at-eof,noisy-file,non-alpha-in-dictionary,token-is-substring,unexpected-line-ending,whitespace-in-dictionary,minified-file,unsupported-configuration,no-files-to-check
+        experimental_apply_changes_via_bot: ${{ github.repository_owner != 'microsoft' && 1 }}
+        use_sarif: ${{ (!github.event.pull_request || (github.event.pull_request.head.repo.full_name == github.repository)) && 1 }}
+        extra_dictionary_limit: 20
```
jsoref, line 109: The change from 10 to 20 is something I'm toying with in spell-check-this, but it might not make sense here.

.github/workflows/spelling3.yml
```diff
@@ -59,35 +83,41 @@ jobs:
...
         extra_dictionaries:
-          cspell:cpp/src/compiler-msvc.txt
-          cspell:cpp/src/cpp.txt
```
jsoref, line 84: this dictionary is lousy (way too many typos included, similar to the win32 dictionary) -- as of v0.0.22, I [🗑️ Removed (both) Dictionaries](https://github.com/check-spelling/check-spelling/wiki/Removed-Dictionaries) from the suggestion list due to their poor quality.

I did look into trying to fix them, but they had so many typos that I ran out of energy trying to go over them.

.github/workflows/spelling3.yml
```diff
@@ -59,35 +83,41 @@ jobs:
...
+          cspell:software-terms/dict/softwareTerms.txt
           cspell:cpp/src/stdlib-cpp.txt
+          cspell:cpp/src/stdlib-c.txt
+          cspell:dotnet/dict/dotnet.txt
           cspell:filetypes/filetypes.txt
-          cspell:fullstack/fullstack.txt
-          cspell:powershell/powershell.txt
-          cspell:software-terms/src/software-terms.txt
+          cspell:csharp/csharp.txt
+          cspell:cpp/src/compiler-msvc.txt
```
jsoref, line 117: order doesn't matter, so I should probably shift things so that they aren't changing in this PR.

.github/workflows/spelling3.yml
```diff
@@ -111,12 +141,38 @@ jobs:
...
+    concurrency:
+      group: spelling-update-${{ github.event.issue.number }}
+      cancel-in-progress: false
+    steps:
+    - name: apply spelling updates
+      uses: check-spelling/check-spelling@v0.0.22
+      with:
+        experimental_apply_changes_via_bot: 1
+        checkout: true
+        ssh_key: "${{ secrets.CHECK_SPELLING }}"
```
jsoref, line 178: This is, as noted, very optional, but it should make it easier for others to interact with the workflow instead of having to run programs on their own computer.

.github/actions/spelling/candidate.patterns
florelis, line 1: So much of this file seems to be specific to technologies we don't use in the project. I'm guessing we keep it all because it's easier to update if we don't make changes to the standard file?

.github/actions/spelling/excludes.txt
```diff
@@ -79,6 +103,24 @@
 ^src/AppInstallerCLITests/TestData/InputPublishers.txt$
 ^src/AppInstallerCLITests/TestData/NormalizationInitialIds.txt$
 ^src/AppInstallerCommonCore/external/do.h$
+^\Qsrc/AppInstallerCLIE2ETests/TestData/Configuration/Empty.yml\E$
```
florelis, line 106: This seems duplicated

.github/actions/spelling/excludes.txt
```diff
@@ -79,6 +103,24 @@
...
+^\Qsrc/PowerShell/ExternalModules/PackageManagement/1.4.8.1/PackageProviderFunctions.psm1\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/DscResources/MSFT_PSModule/en-US/MSFT_PSModule.strings.psd1\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/DscResources/MSFT_PSModule/MSFT_PSModule.psm1\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/DscResources/MSFT_PSModule/MSFT_PSModule.schema.mof\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/DscResources/MSFT_PSRepository/en-US/MSFT_PSRepository.strings.psd1\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/DscResources/MSFT_PSRepository/MSFT_PSRepository.psm1\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/DscResources/MSFT_PSRepository/MSFT_PSRepository.schema.mof\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/PowerShellGet.psd1\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/PSGet.Format.ps1xml\E$
+^\Qsrc/PowerShell/ExternalModules/PowerShellGet/2.2.5/PSModule.psm1\E$
```
florelis, line 123: Could this be a single line up to ExternalModules to exclude the whole directory?

.github/actions/spelling/patterns.txt
florelis: Does this still check the spelling of the word after the `T`? Or would it allow stuff like `TMispeled`?

.github/actions/spelling/expect.txt
```diff
@@ -502,3 +566,16 @@ yao
...
+Linq
+nameof
+netfx
+notmatch
+ools
+paramref
+Templating
+typeparam
+Unk
+websites
```
florelis, line 581: Nit: can these be sorted?

.github/workflows/spelling3.yml
florelis: I can see why you wouldn't want to suggest the lousy dictionaries, but I'm not totally convinced it is better for us to not use them. I'm not too invested in that position, though.

The downside I see to using them is that if we happen to make the same typo, we wouldn't notice. To me that seems somewhat unlikely, and if it does happen doesn't seem like that big of a deal. I assume the dictionary would get better over time, so those typos would eventually come up.
As for upsides, it will be easier for us to manage the lists of allow/expect if we have to add less words. If we have to add lots of words, I think it's more likely we're going to make another mistake like with "uknown"

.github/workflows/spelling3.yml
```diff
@@ -45,12 +68,13 @@ on:
...
     - "**"
-    tags-ignore:
-    - "**"
     types:
     - 'opened'
     - 'reopened'
     - 'synchronize'
+  issue_comment:
+    types:
+    - 'created'
```
florelis, line 77: What does this do? Check spelling when someone creates an issue?

florelis (approved):

.github/actions/spelling/candidate.patterns
jsoref: Yep.

.github/actions/spelling/excludes.txt
jsoref: Yes

.github/actions/spelling/expect.txt
jsoref: These will go away once it merges, they should be covered by one of the other dictionaries or some other part of the upgrade.

.github/workflows/spelling3.yml
jsoref: When someone comments on an issue/PR it will trigger the workflow. Each job independently decides whether to do anything.

For,
https://github.com/microsoft/winget-cli/blob/64d4149c1cf690f66c6a01cfaa9d1eb2fba18ac3/.github/workflows/spelling3.yml#L80
It'll do nothing:
https://github.com/microsoft/winget-cli/blob/64d4149c1cf690f66c6a01cfaa9d1eb2fba18ac3/.github/workflows/spelling3.yml#L90

For,
https://github.com/microsoft/winget-cli/blob/64d4149c1cf690f66c6a01cfaa9d1eb2fba18ac3/.github/workflows/spelling3.yml#L121
It'll do nothing:
https://github.com/microsoft/winget-cli/blob/64d4149c1cf690f66c6a01cfaa9d1eb2fba18ac3/.github/workflows/spelling3.yml#L128

For,
https://github.com/microsoft/winget-cli/blob/64d4149c1cf690f66c6a01cfaa9d1eb2fba18ac3/.github/workflows/spelling3.yml#L136
It'll do nothing:
https://github.com/microsoft/winget-cli/blob/64d4149c1cf690f66c6a01cfaa9d1eb2fba18ac3/.github/workflows/spelling3.yml#L144

It will only do work here:
https://github.com/microsoft/winget-cli/blob/64d4149c1cf690f66c6a01cfaa9d1eb2fba18ac3/.github/workflows/spelling3.yml#L153-L165
One of the criteria being that it isn't this repository.

If someone creates a PR within a single fork and it triggers a comment w/ a note about items that should be added to expect, the comment will include something like this from @check-spelling-bot:

.github/actions/spelling/patterns.txt
jsoref: Yes, the `(?=...)` is a [zero-width lookahead](https://www.regular-expressions.info/lookaround.html)

.github/workflows/spelling3.yml
JohnMcPMS: I think I have to veto on this permission.  I don't know what feature this removes, but allowing an action outside of our control to write is not going to work.  If they had `create` or similar permission level so that you could create new items but not overwrite existing ones, then it would likely be more acceptable.

JohnMcPMS (changes requested):
Cannot allow write permissions.

.github/workflows/spelling3.yml
jsoref: Yeah, I agree. I'm really annoyed that there isn't such a thing. If you can find someone to poke to get them to make a better api surface, I'd be overjoyed 😉.

JohnMcPMS: [community/discussions#29710](https://github.com/orgs/community/discussions/29710)

Closest thing I found.

denelon:
We use a "template: foo/bar/baz" type of convention for messages from bots so it's easy to build e-mail rules to filter those out. How hard would it be to add something like that to the spellcheck messages in PRs?

jsoref:
@denelon, add a line to https://github.com/microsoft/winget-cli/blob/master/.github/actions/spelling/advice.md

You could even add it of the form:
```
<!--
template: check-spelling
-->
```

So that it's only seen by mail filters and not by humans.

src/WinGetUtilInterop.UnitTests/TestCollateral/AllEquality.yaml
```diff
@@ -21,7 +21,7 @@ Installers:
       Language: /en-US
       Custom: /s
   - Arch: x64
-    Url: https://contosa.net/publiccontainer/contosainstaller64.exe
+    Url: https://contoso.com/publiccontainer/contosoinstaller64.exe
```
jsoref, line 24: I think I'd been ignoring this for a while... contoso.com is microsoft's domain. contosa is a typo of contoso, and contosa.net isn't registered.

If this needs to _not_ be contoso.com, it should be an example.com or some other microsoft controlled domain...

.github/actions/spelling/advice.md
```diff
@@ -23,3 +25,6 @@ https://www.regexplanet.com/advanced/perl/) yours before committing to verify it
   Note that patterns can't match multiline strings.
 
 </details>
+<!--
+template: check-spelling
```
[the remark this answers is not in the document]
jsoref: @denelon I hope this meets your needs, but you're welcome to change it to some other pattern.

jsoref:
Hmm.. I guess I need to rebase -- I was hoping @denelon or @JohnMcPMS would have reviewed it since...

At this point, I'm close to releasing the next version...

denelon:
Sorry about that! We've been super busy with some internal work and haven't been as good as we should have been about reviewing external PRs. Do you want me to get someone to review this or hold off until the next release?

.github/workflows/spelling3.yml
jsoref: I'm dropping this for now.

.github/workflows/spelling3.yml
```diff
@@ -62,61 +83,48 @@ jobs:
...
-            cspell:powershell/powershell.txt
-            cspell:software-terms/src/software-terms.txt
-
-  comment-push:
-    name: Report (Push)
-    # If your workflow isn't running on push, you can remove this job
-    runs-on: ubuntu-latest
-    needs: spelling
-    permissions:
-      contents: write
```
jsoref, line 98: I'm removing this job from the flow.

In the long term, I'm nudging people to rely on the step summary (which happens automatically) over adding comments, but in the near term, people can rely on comments when they make a pull request.

florelis (approved):

JohnMcPMS (approved):

jsoref:
@JohnMcPMS  could you add `check-spelling/actions-checkout@*` to the exemptions list? It isn't used at this time, but because of the logic, the runtime doesn't know.

```
Check Spelling
Bad request - check-spelling/actions-checkout@v4 is not allowed to be used in microsoft/winget-cli. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, or matching the following: check-spelling/check-spelling@*, check-spelling/checkout-merge@*, craigloewen-msft/gitgudsimilarissues@*.
```

denelon:
@jsoref, I just added it.

PR #3899, opened 2023-11-16
Add a script to bootstrap running Pester tests

JohnMcPMS:
## Change
Adds a script that can bootstrap from nothing to run the Pester tests.  Also updates the existing test scripts to enable this with the goal of not changing their current behavior.  Publishes the test scripts so that one can download the artifacts and run tests against whatever set of modules and client is interesting.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3899)


src/PowerShell/scripts/Execute-WinGetTests.ps1
```diff
@@ -0,0 +1,128 @@
...
+    [string]$ConfigurationModuleVersion,
+    
+    # The version of the client to use. Use 'existing' to skip updating.
+    [string]$ClientVersion,
+    
+    # The version of Powershell to use. Use 'existing' to skip updating.
+    [string]$PwshVersion,
+    
+    # The path to the binaries to run the web server.
+    [string]$LocalhostWebServerPath = (Join-Path $PSScriptRoot "..\LocalhostWebServer"),
```
Trenly, line 18: For any parameters which are paths, it may be a good idea to validate the path exists as part of the parameter validation
```powershell
[ValidateScript({if ($_){  Test-Path $_}})]
```

src/PowerShell/scripts/Execute-WinGetTests.ps1
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+[CmdletBinding()]
+param(
+    # The version of the client PS module to use.
+    [string]$ClientModuleVersion,
+
+    # The version of the configuration PS module to use.
```
msftrubengu, line 8: You could have a parameter set for to import local modules

msftrubengu (approved):

Issue #3934, opened 2023-11-29
App different versions, same ID, unable to uninstall


GustavoAlmeidaAboud:
### Brief description of your issue

Two different versions of the software installed with the same ID

### Steps to reproduce

I believe this was due a winget upgrade

### Expected behavior

Be able to uninstall winget app with the same ids

### Actual behavior

Unable to uninstall winget with the same id

Winget list keepass
-----------------------------------------------------------------------------
KeePass Password Safe 1.40.1 DominikReichl.KeePass   1.40.1  2.55.0    winget
KeePass Password Safe 2.55   DominikReichl.KeePass   2.55              winget
KeePass 2.53                 DominikReichl.KeePass   2.53.0  2.55.0    winget

### Environment

```shell
Windows Package Manager v1.6.3133
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22631.2715
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.21.3133.0
```

mdanish-kh:
This is a known issue with side-by-side installations and unfortunately WinGet does not handle this scenario all that well in its current implementation.

You'll have to manually uninstall the older version from `appwiz.cpl`. After that the uninstall command will work as expected with a single installed version of the package.

- Related to https://github.com/microsoft/winget-cli/issues/2129

denelon:
We've been working on the side-by-side scenarios. You could try the latest release and enable the experimental feature on:
* https://github.com/microsoft/winget-cli/releases/tag/v1.8.924-preview

Please provide feedback at:
* https://github.com/microsoft/winget-cli/discussions/4281

denelon:
You should now be able to uninstall a specific version of a package using WinGet 1.8.

Issue #3910, opened 2023-11-19
Winget says package is already installed, when it isn't


anthonyvdotbe:
### Brief description of your issue

2 flavors of FFmpeg cannot be installed alongside each other, but the error message doesn't make sense and is confusing.

### Steps to reproduce

* winget install --id 'Gyan.FFmpeg' --scope machine
* winget install --id 'Gyan.FFmpeg.Shared' --scope machine

### Expected behavior

Either both packages install, or the second command explains that `Gyan.FFmpeg.Shared` cannot be installed alongside `Gyan.FFmpeg` (ideally, it would also explain that `Gyan.FFmpeg.Shared` is a superset of `Gyan.FFmpeg`, and propose to uninstall `Gyan.FFmpeg` automatically before installing  `Gyan.FFmpeg.Shared`).

### Actual behavior

The second command just says the package is already installed, which is not true and confusing. What's particularly problematic to me, is that the commands use IDs, so it doesn't make sense to me that 2 packages with different IDs cause this behavior.

### Environment

```shell
Windows Package Manager v1.6.3133
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22631.2715
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.21.3133.0
```

denelon:
Are there any difference in Windows Apps & Feature via settings between how these two applications appear when they are installed?

anthonyvdotbe:
Yes. `Gyan.FFmpeg.Shared`:
![image](https://github.com/microsoft/winget-cli/assets/15714253/f7aa783b-79eb-4116-aae2-6bb8029285b9)

and `Gyan.FFmpeg`:
![image](https://github.com/microsoft/winget-cli/assets/15714253/f20abbf6-5a42-47d1-a603-0175590e9105)

denelon:
I see what's happening. The part in parenthesis is generally ignored. This is often a mechanism used to indicate scope or architecture but hasn't been reliable enough to use as authoritative since it's just a "displayName" field.

WinGet would see both entries as the same package. It may be possible to add Apps and Features entries to the manifests to clearly differentiate the "displayName" properties. We've been discussing changing the default logic to include what is in parenthesis when the Apps and Features entries are present in the manifest.

denelon:
We've been working on the side-by-side scenarios. You could try the latest release and enable the experimental feature on:
[...]

Issue #3485, opened 2023-07-29
Python.Python.2 Package v2.7.15150 is in a seemingly corrupt state 


the-simian:
### Brief description of your issue

Attempt to upgrade Python 2.7.15150 and get "                                                                                                                        
No installed package found matching input criteria."

I am not sure if this is related to Winget-cli itself or Winget UI, but I suspect its in the underlying cli, so submitting here.

### Steps to reproduce

As stated
![image](https://github.com/microsoft/winget-cli/assets/954596/d65c38b3-e978-496d-9faf-a752de1b0fa3)
And listed as installed:
![image](https://github.com/microsoft/winget-cli/assets/954596/907c25a3-23f3-4403-9e29-37e4227e8d91)

if I use the command line and run `list I get:
```
C:\Users\<my username>\Desktop>winget list Python
Name                     Id                                     Version    Available Source
-------------------------------------------------------------------------------------------
Python 2.7.15 (64-bit)   Python.Python.2                        2.7.15150  2.7.18150 winget
Python 3.12.0b4 (64-bit) Python.Python.3.12                     3.12.0b4             winget
WingetUI                 SomePythonThings.WingetUIStore         2.0.2                winget
Python 2.7.18 (64-bit)   Python.Python.2                        2.7.18150            winget
Python Launcher          {C6192300-FDE6-46C6-96A9-EB8C9370D520} 3.12.114.0
```

You can see that Winget (and winget ui) thinks that 2.7.15150 is there and can be updated, and also that the version it should update to is also there. 

### Expected behavior

I'd expect to be able to uninstall or update `2.7.15150` and `2.7.15150` would no longer appear in the 'needs-updating' list.

### Actual behavior

Version `2.7.15150` is still in the list, no matter what. I cannot update or remove it. I've got the version it seems to think it should update to installed. 

here is the related log operation entry in winget ui:
```
 Operation: update - Perform date 2023-07-29 15:39:22.374189
 Package: <Package: Python 2.7.15 (64-bit);Python.Python.2;2.7.15150;Winget: winget;<PackageManagers.winget.WingetPackageManager object at 0x00000275A4144810>;<genericCustomWidgets.TreeWidgetItemWithQAction object at 0x00000275A9FDBC80>>
[...]

Issue #3278, opened 2023-05-26
winget upgrade Microsoft.VCRedist.2013.x64  most current package missing?


georgewilkin:

denelon:
We've been working on the side-by-side scenarios. You could try the latest release and enable the experimental feature on:
* https://github.com/microsoft/winget-cli/releases/tag/v1.8.924-preview

Please provide feedback at:
* https://github.com/microsoft/winget-cli/discussions/4281

PR #3940, opened 2023-11-30
[Spec] First pass at Side-By-Side support.

denelon:

Related to:
* #2129

Possibly related to:
* #1392 
* #1619 
* #1778 
* #1831 
* #1865 
* #1886
* #2118 
* #2458
* #2634 
* #2645
* #2726 
* #3069 
* #3179 
* #3229 
* #3248 
* #3278 
* #3485 
* #3910 
* #3934
-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3940)

### DIFF
--- a/doc/specs/#2129 - Side-by-side.md
+++ b/doc/specs/#2129 - Side-by-side.md
@@ -1,0 +1,100 @@
+---
+author: Demitrius Nelon @denelon
+created on: 2023-11-30
+last updated: 2023-11-30
+issue id: 2129
+---
+
+# Better support for Side by Side packages
+
+[comment]: # Link to issue: "For [#2929](https://github.com/microsoft/winget-cli/issues/2129)"
+
+## Abstract
+
+WinGet should better support multiple versions of the same package being installed on the same machine. This happens, and WinGet currently tries to pretend it doesn't.
+
+## Inspiration
+
+Multiple [issues](https://github.com/microsoft/winget-cli/issues?q=is%3Aissue+is%3Aopen+label%3Aside-by-side) have been submitted with the "side-by-side" label.
+
+## Solution Design
+
+Per comment by @JohnMcPMS:
+> Replace PackageVersion InstalledVersion with PackageVersion[] InstalledVersions internally, then reasoning about how that affects everywhere that breaks.
+
+## UI/UX Design
+
+Several different commands will be impacted by this change. In some cases, additional information will be presented to the user so they are aware of the implications for the command executed.
+
+### WinGet Install
+
+When one or more versions (but not the latest) of a package are installed on the system and none of them are pinned, running `winget install <package>` will install the latest version of the package side-by-side unless the manifest specifies the previous version should be uninstalled or denied. The user may override this behavior with "--retain-previous".
+> Note: WinGet will not attempt to override the default behavior the installer may have related to the removal of previous versions of the package.
+
+### WinGet List
+
+When more than one version of a package is installed on the system, all installed versions will all be listed. Only the latest version of an installed package will show results in the "Available" column if a newer version is available in a configured source.
+
+### WinGet Upgrade
+
+When more than one version of a package is installed on the system, only the latest installed version will be listed if a newer version is available in a configured source.
+
+### WinGet Uninstall
+
+For `winget uninstall <package>`:
+
+When more than one version of a package is installed on the system, the user will be prompted to disambiguate. The user will be informed "Multiple package versions found matching input criteria. Please specify the version to be uninstalled or use "--all" to remove all versions of the package". All installed versions will be displayed so the user may either disambiguate by specifying the version to be removed or by passing "--all" as an argument to specify every installed version of the package should be uninstalled.
+
+For `winget uninstall <package> --version <version>`:
+
+If the version specified is one of the installed versions, only that version will be uninstalled. 
+
+### WinGet Pin
+
+The current behaviors associated with pinned packages will continue to be honored.
+
+## Capabilities
+
+Many different commands will be impacted as WinGet will understand the notion of multiple versions of the same package may be installed. This will necessarily be complicated when one or more versions of a package are pinned, and the pins should be prioritized in terms of behavior.
+
+### Accessibility
+
+There should be no noteworthy impacts to accessibility.
+
+### Security
+
+There should be no noteworthy impacts to security.
+
+### Reliability
+
+There should be no noteworthy impacts to reliability.
+
+### Compatibility
+
+There should be no noteworthy impacts to compatibility.
+
+### Performance, Power, and Efficiency
+
+## Potential Issues
+
+Some package installers may remove one or more previous versions of packages upon install. WinGet will not be able to override these behaviors, so in some cases, users may not be pleased with the results.
+
+Some packages are isolated into unique package identifiers based on major or minor versioning. When WinGet is attempting to correlate an installed version of a package during an upgrade scenario this may result in matching with more than one package in a configured source. This is often the case with runtimes or software development kits. 
+
+For example, consider the following packages are available.
+
+SDK.Runtime.1
+SDK.Runtime.2
+
+Assume SDK.Runtime.1 has versions 1.0, 1.1, and 1.2 available.
+Assume SDK.Runtime.2 has versions 2.0, and 2.1 available.
+
+The user may have SDK.Runtime.1 versions 1.0 and 1.1 installed as well as SDK.Runtime.2 version 2.0 installed. 
+
+If sufficient metadata isn't provided for WinGet to correlate both logical packages have updates available, the user may only be provided with the SDK.Runtime.2 version 2.1 as an available upgrade.
+
+## Future considerations
+
+Some packages are isolated into unique package identifiers based on major or minor versioning. In the future, WinGet may have some form or package providers or virtual packages to simplify getting or installing the latest version. 
+
+## Resources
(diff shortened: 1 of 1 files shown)


doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,100 @@
...
+Per comment by @JohnMcPMS:
+> Replace PackageVersion InstalledVersion with PackageVersion[] InstalledVersions internally, then reasoning about how that affects everywhere that breaks.
+
+## UI/UX Design
+
+Several different commands will be impacted by this change. In some cases, additional information will be presented to the user so they are aware of the implications for the command executed.
+
+### WinGet Install
+
+When one or more versions (but not the latest) of a package are installed on the system and none of them are pinned, running `winget install <package>` will install the latest version of the package side-by-side unless the manifest specifies the previous version should be uninstalled or denied. The user may override this behavior with "--retain-previous".
```
Trenly, line 31: The way this is written, `--retain-previous` would also override `UpgradeBehavior: deny`. This should be clarified that `deny` would take precedence

doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,100 @@
...
+> Note: WinGet will not attempt to override the default behavior the installer may have related to the removal of previous versions of the package.
```
Trenly, line 32:
```suggestion
> [!NOTE]  
> WinGet will not attempt to override the default behavior the installer may have related to the removal of previous versions of the package.
```
This will give better formatting

doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,100 @@
...
+
+### WinGet List
+
+When more than one version of a package is installed on the system, all installed versions will all be listed. Only the latest version of an installed package will show results in the "Available" column if a newer version is available in a configured source.
```
Trenly, line 36:
```suggestion
When more than one version of a package is installed on the system, all installed versions will be listed. Only the latest version of an installed package will show results in the "Available" column if a newer version is available in a configured source.
```
Nit: grammar

doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,100 @@
...
+
+### WinGet Upgrade
+
+When more than one version of a package is installed on the system, only the latest installed version will be listed if a newer version is available in a configured source.
```
Trenly, line 40:
```suggestion
When more than one version of a package is installed on the system, only the latest installed version will be listed if a newer version is available in a configured source. If the latest available version is in the collection of installed packages, no upgrade will be listed.
```
This is important since it's the heart of the bug. Right now the behavior is as follows, which is the behavior this feature is intending to fix -

Version 1 and Version 2 are installed on a machine. Version 2 is the latest in the source.
Version 1 will show as available for upgrade, even though Version 2 is already installed.

doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,100 @@
...
+
+### WinGet Uninstall
+
+For `winget uninstall <package>`:
+
+When more than one version of a package is installed on the system, the user will be prompted to disambiguate. The user will be informed "Multiple package versions found matching input criteria. Please specify the version to be uninstalled or use "--all" to remove all versions of the package". All installed versions will be displayed so the user may either disambiguate by specifying the version to be removed or by passing "--all" as an argument to specify every installed version of the package should be uninstalled.
+
+For `winget uninstall <package> --version <version>`:
```
Trenly, line 48: There is an implementation dependency here on -
* #1886

doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,100 @@
...
+
+### WinGet Upgrade
+
+When more than one version of a package is installed on the system, only the latest installed version will be listed if a newer version is available in a configured source.
+
+### WinGet Uninstall
+
+For `winget uninstall <package>`:
+
+When more than one version of a package is installed on the system, the user will be prompted to disambiguate. The user will be informed "Multiple package versions found matching input criteria. Please specify the version to be uninstalled or use "--all" to remove all versions of the package". All installed versions will be displayed so the user may either disambiguate by specifying the version to be removed or by passing "--all" as an argument to specify every installed version of the package should be uninstalled.
```
Trenly, line 46: This is related to -
* #2645

We should consider the behavior of `--uninstall-previous --all` and probably add a reference to #2645 in the notes of this spec

denelon: I linked to the issue. Draft away :)

denelon (commented):
update for deny upgrade behavior

doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,101 @@
...
+
+### WinGet Upgrade
+
+When more than one version of a package is installed on the system, only the latest installed version will be listed if a newer version is available in a configured source. If the latest available version is in the collection of installed packages, no upgrade will be listed.
+
+### WinGet Uninstall
+
+For `winget uninstall <package>`:
+
+When more than one version of a package is installed on the system, the user will be prompted to disambiguate. The user will be informed "Multiple package versions found matching input criteria. Please specify the version to be uninstalled or use "--all" to remove all versions of the package". All installed versions will be displayed so the user may either disambiguate by specifying the version to be removed or by passing "--all" as an argument to specify every installed version of the package should be uninstalled.
```
denelon, line 47:
```suggestion
When more than one version of a package is installed on the system, the user will be prompted to disambiguate. The user will be informed "Multiple package versions found matching input criteria. Please specify the version to be uninstalled or use "--all" to remove all versions of the package". All installed versions will be displayed so the user may either disambiguate by specifying the version to be removed or by passing "--all" as an argument to specify every installed version of the package should be uninstalled.

> [NOTE!]
> The use of "--all" for the uninstall command requires a package to be specified. Attempting to run `winget uninstall --all` is not supported.
```

mdanish-kh (commented):
Some packages (Microsoft.PowerShell, Microsoft.VisualStudioCode...screenshot attached below) allow side-by-side installations of same versions of different scope. WinGet should be able to disambiguate installed versions based on scope as well and allow users to:

`winget uninstall/install/upgrade Microsoft.VisualStudioCode --version <side_by_side_version> --scope machine` 

image

JohnMcPMS:
I updated the "implementation" section of the main issue with a brain dump that includes things beyond SxS but that are generally related to the greater concept of a more complex association graph.

As @mdanish-kh points out, this spec targets SxS *versions*, but disregards the potential for SxS within the same version, via scope, architecture, etc.  This is relevant for any commands that target the installed entity (`upgrade`, `uninstall`).

I want to request flexibility here on the UX of the CLI, as we will need to carefully balance adding more data to the index to achieve the `winget list` goals.  On the other side, we will also be relying on the manifest for deeper information but we only want to pull that down as needed, such as `winget list <single package ID>` to be able to enumerate every version and flavor of installed item.

denelon:
Locale is the other troublemaker.

I'll work on additional updates to the proposed specification.

During some very early discussions (2019), we were working on the characteristics that would uniquely define an installer within a manifest for a particular version of a package:
* Installer Type (MSIX, MSI, EXE)
* Architecture (x86, x64, ARM, ARM64)
* Target Platform (Desktop, Server)
* Locale (BCP-47 format)
* Scope (User vs. System)

The other concept mentioned in the updated description of the issue was called out as "group package". I've been thinking of that concept in terms of either a virtual package, or a provider concept.

The example for Python is fairly straight forward as there is one logical ISV (independent software vendor)/Publisher. I actually see Java (JRE/JDK) as more complex due to the fact that there could be more than one ISV/Publisher. In the case of dependencies where a package might need some minimum version of a JRE or a JDK, WinGet could potentially detect that the dependency was met by more than one ISV/Publisher. In the case of an install prerequisite (dependency) for another package, it's not obvious how a preference would be specified, or a default should be chosen.

Trenly:
> During some very early discussions (2019), we were working on the characteristics that would uniquely define an installer within a manifest for a particular version of a package:
> 
> * Installer Type (MSIX, MSI, EXE)
> * Architecture (x86, x64, ARM, ARM64)
> * Target Platform (Desktop, Server)
> * Locale (BCP-47 format)
> * Scope (User vs. System)
> 

I would add that there should likely be a priority order here - where if a user has a specific architecture installed, that should be respected even if the underlying installer type changed. Ideally all of these uniquely defined installers would be present in the manifest, but with many different permutations of install parameters, it's likely that one (or more) could be missing from the definitions. The client should be able to reasonably detect when an available version doesn't have an architecture matching the installed architecture, same with scope and, where possible, locale. 

Although, the more I think about it, that is probably more of a related feature than specific to SxS support

denelon:
Comments/Concerns:

@ImJoakim
@ItzLevvie
@jedieaston
@KaranKad
@mdanish-kh
@OfficialEsco
@quhxl
@russellbanks
@Trenly

doc/specs/#2129 - Side-by-side.md
```diff
@@ -0,0 +1,110 @@
...
+> [!NOTE]
+> The use of "--all" for the uninstall command requires a package to be specified. Attempting to run `winget uninstall --all` is not supported.
+
+For `winget uninstall <package> --version <version>`:
+
+If the version specified is one of the installed versions, only that version will be uninstalled. 
+
+### WinGet Pin
+
+The current behaviors associated with pinned packages will continue to be honored.
```
Trenly, line 64: Question on gating pins.

Assume I have the following pin:
```
Name                   Id                       Version Source Pin type
-----------------------------------------------------------------------
Microsoft .NET SDK 3.1 Microsoft.DotNet.SDK_3.1 3.1.42* winget Gating
```

If I have these two versions installed -
```
Name                                                Id                                     Version       Available Source
-------------------------------------------------------------------------------------------------------------------------
Microsoft .NET Core SDK 3.1.426 (x64)               Microsoft.DotNet.SDK.3_1               3.1.426                 winget
Microsoft .NET Core SDK 3.1.426 (x64)               Microsoft.DotNet.SDK.3_1               3.1.456                 winget
```

Would the behavior still be to not upgrade, even though I have 3.1.456 installed? clearly the 3.1.426 shouldn't be upgraded, but would 3.1.456 be omitted from upgrade?

denelon:
I'm going to go ahead and close this since we're taking a different approach on how we're slicing up the side-by-side support. I expect @JohnMcPMS will have a PR for part of this work coming soon™️

Issue #3824, opened 2023-10-27
Install-WingetPackage needs additional result properties


jdhitsolutions:
### Description of the new feature / enhancement

When I install a package using the `Install-WingetPackage` PowerShell command, I get a result like this:

```powershell
InstallerErrorCode Status RebootRequired ExtendedErrorCode CorrelationData
------------------ ------ -------------- ----------------- ---------------
0                  Ok     False
```

This is okay if I'm manually installing a single item. But If I'm installing multiple items or installing as a part of an unattended process, I have no way of telling which result goes with which product. I would suggest adding properties to show the package name and source. 

### Proposed technical implementation details

I am a proponent of writing rich objects to the pipeline. It is easier to format output to limit what the user sees by default. With that in mind, I would also suggest including the package version number and moniker.

PR #3954, opened 2023-12-06
Add package id, name, and source to install/update/uninstall result for PowerShell cmdlet

ryfu-msft:
Resolves: #3824 

Modifies the output of the install/update/uninstall result to display the id, name, and source of the package so that it is easier to correlate the result with the package. I wanted to also display the installed version but that would require calling the Get-WinGetPackage command and passing in the arguments.

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/InstallerPackageCommand.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/InstallerPackageCommand.cs
@@ -117,7 +117,6 @@ namespace Microsoft.WinGet.Client.Engine.Commands
 
             if (result != null)
             {
-
                 this.Write(StreamType.Object, new PSInstallResult(result, catalogPackage));
             }
         }
--- a/src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/InstallerPackageCommand.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/InstallerPackageCommand.cs
@@ -94,8 +94,6 @@ namespace Microsoft.WinGet.Client.Engine.Commands
             string psPackageFieldMatchOption,
             string psPackageInstallMode)
         {
-            CatalogPackage? catalogPackage = null;
-
             var result = this.Execute(
                 async () => await this.GetPackageAndExecuteAsync(
                     CompositeSearchBehavior.RemotePackagesFromRemoteCatalogs,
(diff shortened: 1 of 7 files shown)


src/PowerShell/Microsoft.WinGet.Client.Engine/PSObjects/PSInstallResult.cs
```diff
@@ -16,14 +15,50 @@ namespace Microsoft.WinGet.Client.Engine.PSObjects
...
+        }
+
+        /// <summary>
+        /// Gets the id of the installed package.
+        /// </summary>
+        public string Id
+        {
+            get
+            {
+                return this.catalogPackage != null ? this.catalogPackage.Id : string.Empty;
```
Trenly, line 38: For my own knowledge, Is there a reason to not use null coalescing for these?

ryfu-msft: I forgot about those 😓 , updated to use null coalescing operators

src/PowerShell/Microsoft.WinGet.Client.Engine/PSObjects/PSInstallResult.cs
```diff
@@ -16,14 +15,50 @@ namespace Microsoft.WinGet.Client.Engine.PSObjects
...
+        }
+
+        /// <summary>
+        /// Gets the id of the installed package.
+        /// </summary>
+        public string Id
+        {
+            get
+            {
+                return this.catalogPackage?.Id ?? string.Empty;
```
msftrubengu, line 38: You don't need the string.Empty if you make the property nullable

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/InstallerPackageCommand.cs
```diff
@@ -94,6 +94,8 @@ public sealed class InstallerPackageCommand : InstallCommand
             string psPackageFieldMatchOption,
             string psPackageInstallMode)
         {
+            CatalogPackage? catalogPackage = null;
+
             var result = this.Execute(
                 async () => await this.GetPackageAndExecuteAsync(
```
msftrubengu, line 100: Instead of this, you can modify the return of GetPackageAndExecuteAsync to be a tuple with TResult and the catalog package.

msftrubengu (approved):

PR #3956, opened 2023-12-07
Introduce strong and weak comparisons between installers

JohnMcPMS:
## Change
Rather than a `bool`, the installer comparison function now returns one of three values: negative, weak positive, and strong positive.  The comparison functions are updated to choose between weak and strong positive results as appropriate.  For instance, matching the system architecture when the alternative does not is a strong positive result.  Simply being "better" in the list of emulated architectures is a weak positive result.

The function that used these comparators in our priority order is updated to effectively do multiple passes, one looking for strong results and then one looking for weak results (in reality it is implemented with one pass as an optimization).

## Validation
Added a test to ensure that { System Architecture, Unknown Scope } is chosen over { Emulated Architecture, User Scope } by default.
Manually validated that the packages with this setup were now selecting the proper system architecture package as desired.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3956)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -217,8 +217,8 @@ namespace AppInstaller::CLI::Workflow
 
                 if (arch1 > arch2)
                 {
-                    // A match with the system architecture is strong, while another "emulated" architecture is weak.
-                    return (first.Arch == Utility::GetSystemArchitecture() ? details::ComparisonResult::StrongPositive : details::ComparisonResult::WeakPositive);
+                    // A match with the primary architecture is strong
+                    return (first.Arch == GetStrongArchitectureMatch() ? details::ComparisonResult::StrongPositive : details::ComparisonResult::WeakPositive);
                 }
 
                 return details::ComparisonResult::Negative;
@@ -244,6 +244,13 @@ namespace AppInstaller::CLI::Workflow
                     installer.UnsupportedOSArchitectures.end(),
                     Utility::GetSystemArchitecture());
                 return unsupportedItr != installer.UnsupportedOSArchitectures.end();
+            }
+
+            Utility::Architecture GetStrongArchitectureMatch()
+            {
+                // If we have a preferential order, treat the first entry as strong.
+                // Otherwise, treat the system architecture as strong (which is always first in the default order).
+                return m_allowedArchitectures.empty() ? Utility::GetSystemArchitecture() : m_allowedArchitectures.front();
             }
 
             std::vector<Utility::Architecture> m_allowedArchitectures;
@@ -849,6 +856,14 @@ namespace AppInstaller::CLI::Workflow
         for (auto comparator : m_comparators)
         {
             details::ComparisonResult forwardCompare = comparator->IsFirstBetter(first, second);
+            details::ComparisonResult reverseCompare = comparator->IsFirstBetter(second, first);
+
+            // Should not happen, but if it does it points at a serious bug that should be fixed.
+            if (forwardCompare != details::ComparisonResult::Negative && reverseCompare != details::ComparisonResult::Negative)
+            {
+                AICLI_LOG(CLI, Error, << "Installer " << first << " and " << second << " are both better than each other?");
+                THROW_HR(E_UNEXPECTED);
+            }
 
             if (forwardCompare == details::ComparisonResult::StrongPositive)
             {
@@ -856,9 +871,7 @@ namespace AppInstaller::CLI::Workflow
                 return true;
             }
 
-            details::ComparisonResult backwardCompare = comparator->IsFirstBetter(second, first);
-
-            if (backwardCompare == details::ComparisonResult::StrongPositive)
+            if (reverseCompare == details::ComparisonResult::StrongPositive)
             {
                 // Second is better by this comparator, don't allow a lower priority one to override that.
                 AICLI_LOG(CLI, Verbose, << "Installer " << second << " is better [strong] than " << first << " due to: " << comparator->Name());
@@ -873,18 +886,11 @@ namespace AppInstaller::CLI::Workflow
                     firstWeakComparator = comparator->Name();
                     firstWeakComparatorResult = true;
                 }
-                else if (backwardCompare == details::ComparisonResult::WeakPositive)
+                else if (reverseCompare == details::ComparisonResult::WeakPositive)
                 {
                     firstWeakComparator = comparator->Name();
                     firstWeakComparatorResult = false;
                 }
-            }
-
-            // Should not happen, but if it does it points at a serious bug that should be fixed.
-            if (forwardCompare == details::ComparisonResult::WeakPositive && backwardCompare == details::ComparisonResult::WeakPositive)
-            {
-                AICLI_LOG(CLI, Error, << "Installer " << first << " and " << second << " are both weakly better than each other?");
-                THROW_HR(E_UNEXPECTED);
             }
         }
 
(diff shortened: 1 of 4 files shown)


Trenly:
Does this new comparison logic give any consideration to  currently installed versions? i.e - if system architecture is x64, but a user has the x86 version of a package installed, would the x86 version be a strong positive for upgrade even though it doesn’t match the system arch?

Trenly:
> I think that you are correct here that a modification is needed. Currently my change always makes system architecture strong when it is allowed and considered better. Moving that to be "strong if is the first architecture in the preference list or system if list is empty" will be better. This allows caller preferences to drive the "strong" match, even if it isn't the system architecture.
> 
> For instance, if you have in your settings file { x86, x64 } on an x64 machine, this would make x86 the strong match because you put it first. Given how we build the preference list across the options, I believe this will achieve the desired goal.

I wasn't even thinking of that, but you're right that we should match the implementation of the preference list where it is ordered.

My initial thoughts were more around this issue - https://github.com/microsoft/winget-pkgs/issues/129427
A user has the x86 version of Microsoft.AzureCLI installed. Assuming no architecture preference is set, WinGet will still default to installing the x64 version of the application due to the system architecture.

JohnMcPMS:
> My initial thoughts were more around this issue - [microsoft/winget-pkgs#129427](https://github.com/microsoft/winget-pkgs/issues/129427) A user has the x86 version of Microsoft.AzureCLI installed. Assuming no architecture preference is set, WinGet will still default to installing the x64 version of the application due to the system architecture.

I suspect that the problem there is a lack of data indicating the architecture of the package on the device.  I believe that we do store when the user requests an architecture and then stick to that, but picking up the architecture from system artifacts only is not accurate enough to always have it.  There are likely improvements that we could make there, but it has thus far been extremely error prone.

For the issue you linked, I'm not even sure that I follow the problem.  The "hint" looks likely to be a manifested post-install message and the user seems to not want it to print if they already have x64 installed.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -825,19 +842,65 @@ namespace AppInstaller::CLI::Workflow
...
+                }
+                else if (backwardCompare == details::ComparisonResult::WeakPositive)
+                {
+                    firstWeakComparator = comparator->Name();
+                    firstWeakComparatorResult = false;
+                }
+            }
+
+            // Should not happen, but if it does it points at a serious bug that should be fixed.
+            if (forwardCompare == details::ComparisonResult::WeakPositive && backwardCompare == details::ComparisonResult::WeakPositive)
```
florelis, line 884: Would we gain anything by also checking for both being strong positives? (at the top of the block)

Trenly:
> I suspect that the problem there is a lack of data indicating the architecture of the package on the device. I believe that we do store when the user requests an architecture and then stick to that, but picking up the architecture from system artifacts only is not accurate enough to always have it. There are likely improvements that we could make there, but it has thus far been extremely error prone.
> 
> For the issue you linked, I'm not even sure that I follow the problem. The "hint" looks likely to be a manifested post-install message and the user seems to not want it to print if they already have x64 installed.

I see; I'm probably thinking too much into it then. It's probably just an unrelated issue that is presenting itself in a way that makes it appear related.

florelis (approved):

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -813,7 +837,7 @@ namespace AppInstaller::CLI::Workflow
             auto inapplicability = filter->IsApplicable(installer);
             if (inapplicability != InapplicabilityFlags::None)
             {
-                AICLI_LOG(CLI, Info, << "Installer " << installer << " not applicable: " << filter->ExplainInapplicable(installer));
+                AICLI_LOG(CLI, Verbose, << "Installer " << installer << " not applicable: " << filter->ExplainInapplicable(installer));
```
Trenly, line 840: I would prefer this to remain as info. When users are posting their logs, it can be an important debugging tool to identify why a certain installer is selected. Since the default is to not create verbose logs, having this at the info level makes it so we don't need to ask for verbose logs

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
```diff
@@ -825,19 +849,64 @@ namespace AppInstaller::CLI::Workflow
...
+        bool firstWeakComparatorResult = false;
+
         for (auto comparator : m_comparators)
         {
-            if (comparator->IsFirstBetter(first, second))
+            details::ComparisonResult forwardCompare = comparator->IsFirstBetter(first, second);
+            details::ComparisonResult reverseCompare = comparator->IsFirstBetter(second, first);
+
+            // Should not happen, but if it does it points at a serious bug that should be fixed.
+            if (forwardCompare != details::ComparisonResult::Negative && reverseCompare != details::ComparisonResult::Negative)
```
Trenly, line 862: I had added a few other comments above, but removed them and am condensing them here for consideration.

Basically, each comparison assumes that the first option is better if the preference is equal. Take for example when `m_preference` for installer type is empty - In this case, all installer types should be considered equal. However, both a forward compare and reverse compare will return a negative comparison result since the code has a preference to the first when there is no other preference specified. 

In addition, the `negative` and `positive` structure of the comparison result seems to ignore the implied case of equality. Shouldn't there be a `ComparisonResult::Neutral` which is used in the comparator?

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
JohnMcPMS: I changed it so that they were all at the same level.  I don't have a problem changing the entire selection process to be `Info`, but I do want it all, not just the filter pass.

Trenly: Hmm. I'll defer to the team on this then. Having it all the same makes the most sense, but I don't think it all needs to be in info. Verbose should be good enough, it's just a matter of asking for logs if we need them

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
JohnMcPMS: > However, both a forward compare and reverse compare will return a negative comparison result since the code has a preference to the first when there is no other preference specified.

Yes, I don't see the problem.  If it is related to the line selected for this comment, it will only be true when both are positive.  Both being negative should be fine.

The function is explicitly named `IsFirstBetter` to try and remove the semantic ambiguity of `compare` or other similar words.  They also happen to align with the semantics of `operator<` if "less than" means "a better choice of installer".  The same rules apply: `a < b => true` implies `b < a => false`, but `a < b => false` does not imply anything about `b < a`.  The equivalence relationship is then based on both results being false/negative.  Introducing an equality into the comparison functions would make them more complicated than is necessary.

Yes, all of this leads to the potential for weak ordering, and thus choosing simply based on being first in the list of options.  But that will only happen in cases where there is no implied ordering, and then I think it is actually a good thing to give the manifest author control in those cases (rare as they are).

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
JohnMcPMS: Usually by the time I see things, someone has run with / been asked to run with verbose.  But I don't know how common it is that an issue is solved with nothing more than inspecting the installer selection.

Most of the time this will generate ~5 lines in the log.  But there are some cases with 10s or over 100 installers that are a bit much.  Trying to log "the important" choices is difficult.

src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
Trenly: Ah, thanks for the great explanation! I was mis-reading the case of what would happen when they were both negative.

yao-msft (approved):
:shipit:

PR #3978, opened 2023-12-14
Apply MotW when installer hash does not match

florelis:
Current code sets MotW to trusted if an installer is from a trusted source, even if the hash does not match. And for untrusted sources, it does not set the MotW on hash mismatch.

Changing to always set MotW to "internet" for hash mismatch.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/3978)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
@@ -458,11 +458,11 @@ namespace AppInstaller::CLI::Workflow
     {
         if (context.Contains(Execution::Data::InstallerPath))
         {
-            if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerTrusted))
+            if (WI_AreAllFlagsSet(context.GetFlags(), Execution::ContextFlag::InstallerTrusted | Execution::ContextFlag::InstallerHashMatched))
             {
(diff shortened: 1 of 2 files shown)


src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
```diff
@@ -458,11 +458,11 @@ namespace AppInstaller::CLI::Workflow
     {
         if (context.Contains(Execution::Data::InstallerPath))
         {
-            if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::InstallerTrusted))
+            if (WI_AreAllFlagsSet(context.GetFlags(), Execution::ContextFlag::InstallerTrusted | Execution::ContextFlag::InstallerHashMatched))
```
yao-msft, line 461: Actually InstallerTrusted is only set when hash matches.

florelis: Oh, right. I got that wrong in the description. But still I think it would be good to check here as a precaution in case that code ever changes.

yao-msft: And for hash mismatch case, we leave the Motw as URLZONE_INTERNET. When we create the temp file before downloading, we set the Motw.

florelis:
Turns out I was misunderstanding things and this is not needed

PR #4034, opened 2024-01-05
azure-pipelines.nuget.yml Pipeline migration to 1ES

Madhusudhan-MSFT:
This commit migrates azure-pipelines.nuget.yml to 1ES template

**[How Tested:]**
TODOs:
 - Trigger the pipeline from the topic branch and ensure pipeline succeed and ensure pipeline artifacts matches before and after the migration

-----

### DIFF
--- a/azure-pipelines.nuget.yml
+++ b/azure-pipelines.nuget.yml
@@ -1,12 +1,9 @@
 trigger: none
 
 parameters:
-  - name: version
-    displayName: Version to stamp on binaries and nuget package. Should be in form of "major.minor", for example "1.0"
-    type: string
-
-pool:
-  vmImage: "windows-latest"
+- name: version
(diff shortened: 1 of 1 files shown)


azure-pipelines.nuget.yml
```diff
@@ -1,195 +1,208 @@
...
+  parameters:
+    pool:
+      name: Azure-Pipelines-1ESPT-ExDShared
+      image: windows-2022
+      os: windows
+    customBuildTags:
+    - ES365AIMigrationTooling
+
+    stages:
+    - stage: stage
```
yao-msft, line 31: >stage [](http://example.com/codeflow?start=13&length=5)

nit: should we have a better name then just "stage"

JohnMcPMS:
This pipeline is no longer used.

PR #4059, opened 2024-01-11
Support group processing of configurations

JohnMcPMS:
## Change
Adds support for test and apply processing to handle groups.  This is done by changing the top-level processing to only work on groups, and moving the existing processing code to a default group processor that is used when the set processor does not expose the group interface.  All telemetry is handled by the top level processing code so that we get consistent data regardless of the set processor.

Both set and unit processors can implement `IConfigurationGroupProcessor` to indicate that they can handle group processing on their own.  If the set processor does not expose this interface, the default processor will handle the set level units while checking the unit processors for group handling as well.  If the set processor does expose the interface, it is expected to handle all processing.  It is still necessary to implement unit processors to expose details about them, but they don't need to be able to function.
  
Also changed most `winrt::single_threaded_vector` to `winrt::multi_threaded_vector` as it was appropriate for their potential usage.

## Validation
Added new unit tests to cover group test and apply scenarios.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4059)


src/Microsoft.Management.Configuration/ConfigurationSetApplyProcessor.cpp
```diff
@@ -18,17 +18,30 @@ namespace winrt::Microsoft::Management::Configuration::implementation
...
+        }
+
+        bool InformFilter(ConfigurationUnitIntent intent)
+        {
+            return intent == ConfigurationUnitIntent::Inform;
+        }
+
+        bool ApplyFilter(ConfigurationUnitIntent intent)
+        {
+            return intent == ConfigurationUnitIntent::Apply || intent == ConfigurationUnitIntent::Unknown;
```
msftrubengu, line 34: Why does unknown implies apply? Is it because the group intent is unknown?

src/Microsoft.Management.Configuration/ConfigurationUnitResultInformation.cpp
```diff
@@ -29,10 +29,13 @@ namespace winrt::Microsoft::Management::Configuration::implementation
 
     void ConfigurationUnitResultInformation::Initialize(const Configuration::IConfigurationUnitResultInformation& other)
     {
-        m_resultCode = other.ResultCode();
-        m_description = other.Description();
-        m_details = other.Details();
-        m_resultSource = other.ResultSource();
+        if (other)
```
msftrubengu, line 32: How can this be null?

src/Microsoft.Management.Configuration/ConfigurationSetApplyProcessor.cpp
JohnMcPMS: The DSC v3 paradigm doesn't assign an intent at the unit level; groups are used to express the intent but we won't necessarily know what that intent is.  This is modeled with the `Unknown` intent, which is treated internally the same as `Apply`.

src/Microsoft.Management.Configuration/ConfigurationUnitResultInformation.cpp
JohnMcPMS: While one can create a null reference (ex. `int& i = *static_cast<int*>(nullptr);`), the actual referenced object is not null here.  This check is an implicit cast-to-bool, which in turn compares the contained pointer to null.

msftrubengu (approved):

PR #4081, opened 2024-01-17
Use std::variant in ManifestYamlPopulator

msftrubengu:
This PR uses `std::variant` instead of member pointers to manifest structs to be references used by the processing functions. `std::variant` contains pointers to manifest structs to be populated and is passed to the `FieldProcessInfo.ProcessFunc`. Each processing function is responsible of calling `std::get` or `std::get_if` with the correct pointer type. `ValidateAndProcessFields` takes the `std::variant` and pass it to the processing functions.

Processing functions that required the member types now take a pointer to the manifest struct. Some of these can go in the anonymous namespace, but I didn't want to add more moves.

The change will make implementing shadow manifests easier.

### DIFF
--- a/src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
+++ b/src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
@@ -1116,6 +1116,34 @@ namespace AppInstaller::Manifest
 
                 std::move(fields_v1_6.begin(), fields_v1_6.end(), std::inserter(result, result.end()));
             }
+
+            if (manifestVersion >= ManifestVer{ s_ManifestVersionV1_7 })
+            {
+                std::vector<FieldProcessInfo> fields_v1_7 =
+                {
+                    {
+                        "RepairBehavior",
+                        [forRootFields](const YAML::Node& value, std::any& any)->ValidationErrors
+                        {
+                            ManifestInstaller* installer;
+                            if (forRootFields)
+                            {
+                                Manifest* manifest = std::any_cast<Manifest*>(any);
+                                installer = &(manifest->DefaultInstallerInfo);
+                            }
+                            else
+                            {
+                                installer = std::any_cast<ManifestInstaller*>(any);
+                            }
+
+                            installer->RepairBehavior = ConvertToRepairBehaviorEnum(value.as<std::string>());
+                            return {};
+                        }
+                    },
+                };
+
+                std::move(fields_v1_7.begin(), fields_v1_7.end(), std::inserter(result, result.end()));
+            }
         }
 
         return result;
@@ -1213,6 +1241,18 @@ namespace AppInstaller::Manifest
                     (*switches)[InstallerSwitchType::Update] = value.as<std::string>();
                     return{};
                 });
+
+            if (manifestVersion >= ManifestVer{ s_ManifestVersionV1_7 })
+            {
+                result.emplace_back(
+                    "Repair",
+                    [](const YAML::Node& value, std::any& any)->ValidationErrors
+                    {
+                        auto switches = std::any_cast<std::map<InstallerSwitchType, Utility::NormalizedString>*>(any);
+                        (*switches)[InstallerSwitchType::Repair] = value.as<std::string>();
+                        return{};
+                    });
+            };
         }
 
         return result;
--- a/src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
+++ b/src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
@@ -12,42 +12,30 @@ namespace AppInstaller::Manifest
     namespace
     {
         template <typename Ptr>
-        Ptr* any_ptr(std::any& any) { return std::any_cast<Ptr*>(any); }
+        Ptr* variant_ptr(const VariantManifestPtr& v) { return std::get<Ptr*>(v); }
 
-        ManifestInstaller* GetManifestInstallerPtrFromManifest(std::any& any)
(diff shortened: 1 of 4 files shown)


src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -137,20 +137,52 @@ namespace AppInstaller::Manifest
...
-            { "Localization", [this](const YAML::Node& value)->ValidationErrors { m_p_localizationsNode = &value; return {}; } },
-            { "Channel", [this](const YAML::Node& value)->ValidationErrors { m_p_manifest->Channel = Utility::Trim(value.as<std::string>()); return {}; } },
+            { "ManifestVersion", [](const YAML::Node&, std::any&)->ValidationErrors { /* ManifestVersion already populated. Field listed here for duplicate and PascalCase check */ return {}; } },
+            { "Installers", [this](const YAML::Node& value, std::any&)->ValidationErrors { m_p_installersNode = &value; return {}; } },
+            { "Localization", [this](const YAML::Node& value, std::any&)->ValidationErrors { m_p_localizationsNode = &value; return {}; } },
+            {
+                "Channel",
+                [](const YAML::Node& value, std::any& any)->ValidationErrors
+                {
+                    Manifest* manifest = std::any_cast<Manifest*>(any);
```
JohnMcPMS, line 147: I think I would have gone with something to make this easier to remain one line, like:
```
template <typename Ptr>
Ptr* any_ptr(std::any& any) { return std::any_cast<Ptr*>(any); }
...
any_ptr<Manifest>(any)->Channel = ...;
...
```
Which is mostly in the not creating a local, but also is slightly shorter.  If you wanted to just not create a local it would be almost as short, but the function makes it easier to add something if needed.

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -187,9 +243,63 @@ namespace AppInstaller::Manifest
         // Common fields across versions
         std::vector<FieldProcessInfo> result =
         {
-            { "InstallerType", [this](const YAML::Node& value)->ValidationErrors { m_p_installer->BaseInstallerType = ConvertToInstallerTypeEnum(value.as<std::string>()); return {}; } },
-            { "PackageFamilyName", [this](const YAML::Node& value)->ValidationErrors { m_p_installer->PackageFamilyName = value.as<std::string>(); return {}; } },
-            { "ProductCode", [this](const YAML::Node& value)->ValidationErrors { m_p_installer->ProductCode = value.as<std::string>(); return {}; } },
+            {
+                "InstallerType",
+                [forRootFields](const YAML::Node& value, std::any& any)->ValidationErrors
```
JohnMcPMS, line 248: This seems fragile, I would really prefer to avoid a capture.  But I'm guessing it isn't nearly as easy to do as it is to say.

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -472,9 +1766,62 @@ namespace AppInstaller::Manifest
...
+                                localization = std::any_cast<ManifestLocalization*>(any);
+                            }
+
+                            localization->Add<Localization::InstallationNotes>(value.as<std::string>());
+                            return {};
+                        }
+                    },
+                    {
+                        "Documentations",
+                        [this, forRootFields](const YAML::Node& value, std::any& any)->ValidationErrors
```
JohnMcPMS, line 1809: I thought the point of this was to get rid of the `this` capture.

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
```diff
@@ -1066,8 +2691,8 @@ namespace AppInstaller::Manifest
             for (auto const& entry : m_p_localizationsNode->Sequence())
             {
                 ManifestLocalization localization;
-                m_p_localization = &localization;
-                auto errors = ValidateAndProcessFields(entry, LocalizationFieldInfos);
+                std::any any = &localization;
```
JohnMcPMS, line 2694: Maybe you should take in the `std::any` by value and just direct construct into the parameter rather than keeping the extra line.

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
msftrubengu: I can't get rid of it in all of them :(

src/AppInstallerCommonCore/Manifest/ManifestYamlPopulator.cpp
msftrubengu: I tried splitting `GetRootFieldProcessInfo` and calling ValidateAndProcessFields with root, root installer and root localization separately. but that will result in several `ManifestError::FieldUnknown` warning. I don't want to deal with looking at the errors and remove them.

I removed the capture, but it still needs to any_cast to manifest or installer/localization. I'm not sure if the capture is better than this try catch approach.

JohnMcPMS:
You are going to hate me, but I think `std::variant` might be the better approach.  I don't think it is reliant on `typeid`, which we want to avoid due to size bloat.  It will work the same way, except that you have to up front define the set of pointer types you would accept.

src/AppInstallerCommonCore/pch.h
```diff
@@ -59,6 +59,7 @@
 #include <type_traits>
 #include <unordered_set>
 #include <vector>
+#include <any>
```
JohnMcPMS, line 62: Replace with `<variant>`

JohnMcPMS (approved):

Issue #3966, opened 2023-12-11
Wrong translation en: targeting de: Zielgruppenadressierung


samuelspiza:
### Brief description of your issue

The error message "Für die folgenden Pakete ist ein Upgrade verfügbar, für das Upgrade ist jedoch eine explizite Zielgruppenadressierung erforderlich" makes no sense in german. It is the translation for "The following packages have an upgrade available, but require explicit targeting for upgrade". "Zielgruppenadressierung" is the translation for "targeting" in a marketing or user-tracking context. Here it would be more appropriate to use "Adressierung". I think the problem stems from the passivly worded "require explicit targeting" and could be avoided by just using an easily undersandable active wording like "To upgrade these packages, explicitly target them by ID." and "Um diese Pakete zu aktualisieren, adressieren Sie sie bitte explizit über die ID."

### Steps to reproduce

Have "Teams Machine-Wide Installer Microsoft.Teams" installed with german Language pack. Before winget is fixed in #3063 .

### Expected behavior

Error message does not use the word "Zielgruppenadressierung" in german.

### Actual behavior

German error message is "Für die folgenden Pakete ist ein Upgrade verfügbar, für das Upgrade ist jedoch eine explizite Zielgruppenadressierung erforderlich".

### Environment

```shell
Windows-Paket-Manager v1.6.3133
Copyright (c) Microsoft Corporation. Alle Rechte vorbehalten.

Windows: Windows.Desktop v10.0.19045.3693
Systemarchitektur: X64
Paket: Microsoft.DesktopAppInstaller v1.21.3133.0

WinGet-Verzeichnisse
-----------------------------------------------------------------------------------------------------------------------
Protokolle                                   %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\Local…
Benutzereinstellungen                        %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\Local…
Verzeichnis für portierbare Links (Benutzer) %LOCALAPPDATA%\Microsoft\WinGet\Links
Portables Linkverzeichnis (Computer)         C:\Program Files\WinGet\Links
Portierbarer Paketstamm (Benutzer)           %LOCALAPPDATA%\Microsoft\WinGet\Packages
Portierbarer Paketstamm                      C:\Program Files\WinGet\Packages
Portierbares Paketstamm (x86)                C:\Program Files (x86)\WinGet\Packages
[...]

PR #3250, opened 2023-05-14
Fix spelling mistake in german

Speedfighter:
WindowsTerminal_cd667uQz77

The spelling mistake on the screenshot will be fixed with this pull request.
There was a missing space.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/3250)

PR #4133, opened 2024-02-01
Fix German localization resource file, especially many odd translations regarding packages and pinning

igorakkerman:

-----

Fixes https://github.com/microsoft/winget-cli/issues/3966

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4133)

### DIFF
--- a/Localization/Resources/de-DE/winget.resw
+++ b/Localization/Resources/de-DE/winget.resw
@@ -898,16 +898,16 @@ Sie können über die Einstellungsdatei „winget settings“ konfiguriert werde
     <value>Microsoft Store-Quelle für den Windows-App-Installer aktivieren</value>
   </data>
   <data name="PolicyEnableWinGetSettings" xml:space="preserve">
-    <value>Einstellungen des Windows-Paket-Manager aktivieren</value>
+    <value>Einstellungen des Windows Package Manager aktivieren</value>
   </data>
   <data name="PolicyEnableWinGet" xml:space="preserve">
-    <value>Windows-Paket-Manager aktivieren</value>
+    <value>Windows Package Manager aktivieren</value>
   </data>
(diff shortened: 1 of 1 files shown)


igorakkerman:

Trenly:
> Files under Localization folder will be overwritten when we do a localization update, so no changes under Localization folder will take effect.
_Originally posted by @yao-msft in https://github.com/microsoft/winget-cli/issues/3250#issuecomment-1551990405_

Someone at MSFT will need to submit these to the internal localization team

igorakkerman:
Fixes #3966

yao-msft:
Thanks for your suggestion, @igorakkerman 

For winget team: I have created bug [BUG 48808197](https://microsoft.visualstudio.com/OS/_workitems/edit/48808197) for localization team consideration.

We'll close the bug after localization team takes in the suggestions, and after winget's next localization refresh.

denelon:
The latest localization patches were included in WinGet 1.8. I'm going to go ahead and close this PR.

igorakkerman:
I don't see any of the changes from this PR in the latest resource file. Is it safe to assume that they were rejected by your localization team in your internal review process?

ItzLevvie:
It is also important to note Microsoft's German style guide from https://aka.ms/german-styleguide which states "However, in order to offset the actual product name from the word that follows in a compound, the hyphen is placed right between the product name and the following term."

I am not German but based on that "Windows-Paket-Manager" could be correct here.

igorakkerman:
Thanks a lot, @JohnMcPMS . I agree that using the German term "Paket" for the English term "package" in German strings is preferable to what I suggested originally.

PR #4151, opened 2024-02-07
Refactor pinning evaluation

JohnMcPMS:
## Change
Moves the evaluation of pinning to when it is needed, rather than doing it for every composite search.  This both makes it more "pay for play" and enables more complex scenarios in the future by allowing the call sites to change behavior more easily.

Also fixes a few problems that I found along the way:
1. CLICore reaching in to RepoCore rather than strictly using publics
2. Move includes towards convention (`"file"` for files in the project, `<file>` for files outside of the project)
3. Applies pins to the COM `DefaultInstallVersion` and `IsUpdateAvailable` properties
    - Because the caller can still see all available versions, they can choose to ignore the pinned properties

## Validation
Existing regression tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4151)


src/AppInstallerRepositoryCore/Public/winget/PinningData.h
```diff
@@ -0,0 +1,105 @@
...
+            PinStateEvaluator(PinStateEvaluator&&) noexcept;
+            PinStateEvaluator& operator=(PinStateEvaluator&&) noexcept;
+
+            ~PinStateEvaluator();
+
+            // Gets the latest available package that fits within the pinning restrictions.
+            std::shared_ptr<AppInstaller::Repository::IPackageVersion> GetLatestAvailableVersionForPins(const std::shared_ptr<AppInstaller::Repository::IPackage>& package);
+
+            // Determines if the given version is an update to the installed version that this object was created with.
+            bool IsUpdate(const std::shared_ptr<AppInstaller::Repository::IPackageVersion>& availableVersion);
```
florelis, line 84: Does this consider the pins?

src/AppInstallerRepositoryCore/Public/winget/PinningData.h
```diff
@@ -0,0 +1,105 @@
...
+
+            PinStateEvaluator(const PinStateEvaluator&);
+            PinStateEvaluator& operator=(const PinStateEvaluator&);
+            PinStateEvaluator(PinStateEvaluator&&) noexcept;
+            PinStateEvaluator& operator=(PinStateEvaluator&&) noexcept;
+
+            ~PinStateEvaluator();
+
+            // Gets the latest available package that fits within the pinning restrictions.
+            std::shared_ptr<AppInstaller::Repository::IPackageVersion> GetLatestAvailableVersionForPins(const std::shared_ptr<AppInstaller::Repository::IPackage>& package);
```
florelis, line 81: Is the installed version in `package` used in any way? Do we have any expectations about it? (e.g., it is not set, or it matches what this object has)

src/AppInstallerRepositoryCore/Public/winget/PinningData.h
florelis, line 84: These functions assume that the available version passed as parameters are for the same package as the installed version, right? If so, I'd mention that in a comment

src/AppInstallerRepositoryCore/Public/winget/PinningData.h
```diff
@@ -0,0 +1,105 @@
...
+
+            // Determines the pin type to apply to the given version.
+            PinType EvaluatePinType(const std::shared_ptr<AppInstaller::Repository::IPackageVersion>& packageVersion);
+
+        private:
+            PinBehavior m_behavior;
+            std::shared_ptr<AppInstaller::Repository::Microsoft::PinningIndex> m_database;
+            std::optional<Pin> m_installedPin;
+            std::optional<Utility::VersionAndChannel> m_installedVersion;
+            std::map<PinKey, std::optional<Pin>> m_availablePins;
```
florelis, line 94: It took me a moment to get that this is only used as a cache for `EvaluatePinType()` and I had to look for all uses to make sure. May be worth adding a comment here to make it easier to understand.

src/AppInstallerRepositoryCore/PinningData.cpp
```diff
@@ -0,0 +1,213 @@
...
+                    || (pin->GetType() == PinType::Gating && !pin->GetGatedVersion().IsValidVersion(version)))
+                {
+                    return pin->GetType();
+                }
+            }
+
+            return PinType::Unknown;
+        }
+
+        // Gets the pinned state for an available PackageVersionKey that may have a pin,
```
florelis, line 35: `PackageVersionKey` -> `Version`?

florelis (dismissed):

src/AppInstallerRepositoryCore/Public/winget/RepositorySearch.h
```diff
@@ -351,31 +334,18 @@ namespace AppInstaller::Repository
...
         // Gets a specific version of this package.
         virtual std::shared_ptr<IPackageVersion> GetAvailableVersion(const PackageVersionKey& versionKey) const = 0;
 
-        virtual std::pair<std::shared_ptr<IPackageVersion>, Pinning::PinType> GetAvailableVersionAndPin(const PackageVersionKey& versionKey) const
-        {
-            return { GetAvailableVersion(versionKey), Pinning::PinType::Unknown };
-        }
-
-        // Gets a value indicating whether an available version is newer than the installed version.
-        virtual bool IsUpdateAvailable(PinBehavior pinBehavior) const = 0;
```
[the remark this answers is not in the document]
JohnMcPMS: Correct, it didn't seem right to leave this there to be misused.  Additionally, it can't properly consider multiple installed versions, which is a major factor driving this change.

src/AppInstallerRepositoryCore/Microsoft/PortableIndex.cpp
```diff
@@ -1,13 +1,20 @@
...
 #include "Schema/Portable_1_0/PortableIndexInterface.h"
-#include "winget/Filesystem.h"
+#include <winget/Filesystem.h>
 
 namespace AppInstaller::Repository::Microsoft
 {
+    PortableIndex::PortableIndex(PortableIndex&&) = default;
+    PortableIndex& PortableIndex::operator=(PortableIndex&&) = default;
+
+    PortableIndex::~PortableIndex() = default;
```
yao-msft, line 16: Just curious, is there particular reason we move these default constructor/destructor here where in other places we just define "= default" in header? Or this is the preferred convention in future?

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.h
```diff
@@ -29,6 +29,11 @@ namespace AppInstaller::Repository::Microsoft
             return { filePath, disposition, std::move(indexFile) };
         }
 
+        // Opens or creates a PinningIndex database on the default path.
```
yao-msft, line 32: nit: update comment

yao-msft (approved):

src/AppInstallerRepositoryCore/Microsoft/PortableIndex.cpp
JohnMcPMS: It is part of [pImpl](https://en.cppreference.com/w/cpp/language/pimpl); the destruction of the smart pointer has to happen in the compilation unit that has access to the destructor.  Leaving it in the header requires that the caller can do that, but moving it to the .cpp moves that requirement to the implementation.  This allows the `IPortableIndex` type to simply be a forward declaration for the public header and only a complete type in the .cpp.

This is not new convention, but application of a specific technique for this purpose.

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.h
JohnMcPMS: Will update the comment in my other change as soon as I merge them.

Issue #190, opened 2020-05-19
Proxy options support


KiruyaMomochi:
# Description of the new feature/enhancement

Something like:

```powershell
winget install git.git --proxy="http://127.0.0.1:2345"
```

This will make `winget` download `git` through the *proxy server*.

jantari:
Possible proxy configuration options are:

- Environment variables ( e.g. `WINGET_PROXY_URL`)
- Auto-detecting settings from inetcpl.cpl aka IE
- command line arguments
- config file
- registry config separate from IE. This would tie in nicely with ADMX (group policy) templates down the road

zuozishi:
1080端口，老局域网用户了

timiil:
please consider also proxy the download link inside the package , for example , 
```
winget install kodi --proxy http://xxx/
```

please also proxy the download link 'https://mirrors.kodi.tv/releases/windows/win64/kodi-18.7-Leia-x64.exe' which insided the installation package.

faustool:
> Possible proxy configuration options are:
> 
> * Environment variables ( e.g. `WINGET_PROXY_URL`)
> * Auto-detecting settings from inetcpl.cpl aka IE
> * command line arguments
> * config file
> * registry config separate from IE. This would tie in nicely with ADMX (group policy) templates down the road

I would recommend standard variables like HTTP_PROXY or HTTPS_PROXY

t-mxcom:
When adding proxy support, **please do not forget supporting proxies that require NTLM-authentication!!!**
In an enterprise environment where the workstations are joined to a domain, and the users use domain accounts, **entering a password should not be necessary!** (Windows Integrated Authentication)

Thank you!

klesh:
Any plan regarding this feature?

sharpninja:
Really need this!

HydrogenDeuterium:
Need this too!

hxse:
Need it!

sourcedelica:
Surprised this wasn't a 1.0 feature.

kybom:
Need it!

qiyuey:
Need it!

fMichaleczek:
Need it.

1.0 without proxy support ... lol

liuilin:
+1

logique233:
Need it

UM-RMeeker:
Need this as well.

Mindzy:
Also need for `winget upgrade`

zhaofeng-shu33:
[...]

PR #1776, opened 2021-12-05
Add basic proxy support

eternalphane:
- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [x] Are you working against an Issue?

-----

Partially resolves #190.

- [x] `--proxy` option
- [x] `--no-proxy` option (may need a better name, e.g. `proxy-bypass-list`?)
- [x] `network.proxy` settings
- [x] `network.proxyOverride` settings (or `proxyBypass`?)
- [x] `HTTP_PROXY`, `HTTPS_PROXY`, `FTP_PROXY`, `ALL_PROXY`, `NO_PROXY` environment variables

TODOs (not in this pr):

- [ ] authentication
- [ ] PAC
- [ ] protocol-specific proxy settings just like `aria2` (not sure there's much use case)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1776)

PR #4152, opened 2024-02-07
Add spec for proxy support

florelis:
Adding spec for #190 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4152)


doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,86 @@
...
+## Solution Design
+
+A new command line argument will be added to specify a proxy to use during a particular invocation of winget.
+
+An option to set a default proxy to use on every flow will be added, but it will be gated to require administrator permissions.
+
+New Group Policy will also be added for IT admins to control the use of proxies.
+The policies will be similar to those we already have for sources, so that a specific proxy can be required or only a predefined set of proxies can be allowed.
+
+Proxies will not be used for the configuration features for now.
```
yao-msft, line 25: I think if we do the proxy configuration at wininet, DO, restclient level, then winget configuration should already be covered. Is there something I missed?

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,86 @@
...
+
+If a proxy is configured, it will be used for accessing sources and downloading installers.
+Note that it will not affect the behavior of installers themselves, so an installation may still generate traffic outside of the proxy.
+
+## UI/UX Design
+
+We will add a command line argument taking the URI to the proxy.
+A separate parameter will be available to disable the use of proxy if there is a default set.
```
yao-msft, line 33: Do we have an admin setting to allow/disable proxy too? We try to make admin setting and group policy have parity.

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,86 @@
...
+The following mitigating factors will be in place:
+* (New) The ability to set a default proxy will be restricted to administrators, to prevent attackers from adding a proxy without the user realizing.
+* (New) A Group Policy will be available to block the use of proxies, or limit it to an approved list.
+* Pre-indexed sources need to be signed, and the publisher is required to match during source update.
+  When initially adding the source, administrator privileges are already required to limit misuse.
+* Pre-indexed sources include manifest hashes in the local database, to ensure that the manifest downloaded later is as expected.
+* For the Microsoft Store source, we use certificate pinning to ensure we are talking to the right server.
+* Manifests include a hash of the installer that is validated before executing it.
+  The ability to ignore installer hash mismatches requires administrator privileges.
+
```
yao-msft, line 68: Do we want to mention for 3rd party rest sources, we'll just rely on https cert CA not being compromised?

doc/specs/#190 - Proxy Support.md
florelis: I'm not familiar with the configuration code so I may be wrong, but the way I understand it is that that part of the code is mostly independent of the "package manager" side of things and some of it is written in .net. That's why I didn't include it in this.

I also don't know what network connections are done on that side. The only one that comes to mind is downloading the PS modules for each resource. If that's the only one, I don't see much case on going through the work of making it available on that side

JohnMcPMS: If PowerShell allows control over a proxy, then we could tunnel settings along to it.  Currently, the proxy would only affect the ability to retrieve a configuration document from the internet.

florelis: I'm fine with extending the use of proxies to configuration, but it doesn't make much sense to me if it will only affect retrieving the configuration file.

The [`Install-Module`](https://learn.microsoft.com/powershell/module/powershellget/install-module) cmdlet has a `-Proxy`  argument but it "ignores this parameter since it's not supported by `Install-PSResource`," so it doesn't really help.

Apparently we could also set the [`WebRequest.DefaultWebProxy` property](https://learn.microsoft.com/en-us/dotnet/api/system.net.webrequest.defaultwebproxy) to set it for the whole PSSession, but we'd need to try it to be sure it works for this.

yao-msft: btw, I'm fine with leaving configuration as future. I was just asking a naive question and I did not mean to add scope on this work.

yao-msft (dismissed):

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,89 @@
...
+### Performance, Power, and Efficiency
+
+There should not be any notable performance changes.
+
+## Potential Issues
+
+A faulty or misconfigured proxy could impact most of winget's functionality, but it can be worked around by disabling the use of the proxy.
+
+## Future considerations
```
Trenly, line 84: Would a future consideration be multiple proxies? Maybe even per-source proxies?

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,89 @@
...
+If a proxy is configured, it will be used for accessing sources and downloading installers.
+Note that it will not affect the behavior of installers themselves, so an installation may still generate traffic outside of the proxy.
+
+## UI/UX Design
+
+We will add a command line argument taking the URI to the proxy.
+A separate parameter will be available to disable the use of proxy if there is a default set.
+
+```
+> winget settings --enable ProxyCommandLineArgument
```
Trenly, line 37: Can we simplify here? Instead of having a separate command for managing the proxy, it could be built into the enable? Running the enable again would over-write the existing proxy. Thoughts @yao-msft?

`winget settings --enable DefaultProxy https://127.0.0.1:2345`
`winget settings --disable DefaultProxy`

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,89 @@
...
+
+This should have no direct impact on accessibility.
+
+### Security
+
+There is a possibility of an attacker using a malicious proxy to tamper with the data received from the source, or with the contents of the installer file.
+This is not much different from the risks of using a public network.
+The following mitigating factors will be in place:
+* (New) The ability to set a default proxy will be restricted to administrators, to prevent attackers from adding a proxy without the user realizing.
+* (New) A Group Policy will be available to block the use of proxies, or limit it to an approved list.
```
Trenly, line 63: Should there also be a mention of the Group Policy to force a default proxy? I see it mentioned above, but not specifically included here
> New Group Policy will also be added for IT admins to control the use of proxies.
The policies will be similar to those we already have for sources, so that a specific proxy can be required or only a predefined set of proxies can be allowed.

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,89 @@
...
+* Pre-indexed sources need to be signed, and the publisher is required to match during source update.
+  When initially adding the source, administrator privileges are already required to limit misuse.
+* Pre-indexed sources include manifest hashes in the local database, to ensure that the manifest downloaded later is as expected.
+* For the Microsoft Store source, we use certificate pinning to ensure we are talking to the right server.
+* When communicating with REST sources, the certificate used by the source for HTTPS needs to match the domain.
+* Manifests include a hash of the installer that is validated before executing it.
+  The ability to ignore installer hash mismatches requires administrator privileges.
```
Trenly, line 70: This is a bit misleading. Only administrators can enable the ability to ignore mismatches, but only non-elevated users can ignore the mismatch

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,89 @@
...
+## Potential Issues
+
+A faulty or misconfigured proxy could impact most of winget's functionality, but it can be worked around by disabling the use of the proxy.
+
+## Future considerations
+
+Things we may want to consider in the future:
+* Extend support for proxies to the Configuration feature
+* Add proxy support to the COM API
+* Add support for proxies that require authentication
```
Trenly, line 89: I believe there is an option to use DefaultNetworkCredentials when creating a proxy connection. Will this / should this be used in the initial implementation to enable domain-joined accounts to use their domain proxy? Just thinking that it may reduce the need to fully support authentication while providing a solution that addresses many enterprise security items

doc/specs/#190 - Proxy Support.md
yao-msft: We have thought about multiple proxies configured by admin and later usable by user. We have not thought about per source proxy during discussion though. The spec is what we came up with for first iteration. And yes, these could always be included in future consideration and revisited when community asks raise.

doc/specs/#190 - Proxy Support.md
yao-msft: I think Flor added this for admin/group policy to control whether non-elevated user could specify a different proxy in command line arg in one invocation. like "winget install TestPackage --Proxy https://MyTestProxy.com". This is a bit different from specifying default proxy for all users.

By the way, currently admin settings only supports bool, it may be extended to support string or it's put into separate proxy settings.

doc/specs/#190 - Proxy Support.md
yao-msft: Yea. This is the existing winget behavior. Admins control whether any winget invocation could ignore mismatch. But even admin allows hash mismatch, to ignore it, it must be run in non-elevated context to minimize the risk.

doc/specs/#190 - Proxy Support.md
Trenly: Ah, I see. So theres actually 2 different "admin settings" here. That makes sense

doc/specs/#190 - Proxy Support.md
florelis: I have added both "multiple proxies a standard user can choose from" and "default proxy by source" to the future considerations. But we can include them in the first iteration if it seems appropriate.

doc/specs/#190 - Proxy Support.md
florelis: I added a couple of lines to try and clarify the difference between "default proxy" and "standard user can set use proxy"

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,93 @@
...
+This functionality will first need to be enabled through an admin setting, similar to local manifests or hash override.
+
+An option to set a default proxy to use on every flow will be added, but it will require administrator permissions to be set.
+
+New Group Policy will also be added for IT admins to control the use of proxies.
+The policies will be similar to those we already have for sources, so that a specific proxy can be required or only a predefined set of proxies can be allowed.
+
+Proxies will not be used for the configuration features for now.
+
+If a proxy is configured, it will be used for accessing sources and downloading installers.
```
florelis, line 28: I started implementing the feature and just now realized that this is not possible in all cases the way we currently do things. I was assuming it would be basically what @eternalphane did in #1776, i.e., adding the proxy argument in the call to `InternetOpen()`, but we have several other ways of fetching remote content and not all of them expose functionality to specify proxy.

* Delivery Optimization cannot use a custom proxy, only the system-wide one.
* wininet _can_ use a custom proxy (that's `InternetOpen()`), but we basically only use it to download installers.
* The MSIX deployment APIs take a URI and have no option to configure a custom proxy. (I was assuming they took a stream and I could pass the proxied stream, but was wrong..)
* cpprestsdk can take a custom proxy to use
* `Windows.Web.Http.HttpClient` that we use for getting headers (for source update), downloading source MSIX packages and some other things, only allows us to use the system-wide proxy, not to set a custom one.

So, we can implement the feature for REST API requests and for downloading non-msix installers with wininet.
We definitely cannot implement the feature for using DO, or for MSIX installers (unless we download them first, but then we lose streaming installs).
And for the things that use `HttpClient`, we can't unless we decide to re-implement it using wininet or something else.

Would we be comfortable with saying "proxy is only used for non-MSIX installers (where it also forces the use of wininet over DO) and REST sources"? Or how else could we move the feature forward?

denelon: @florelis these are fantastic call-outs.

I think we should proceed with an experimental feature for proxy with the parts we can implement, and clearly state and link to the remaining details in Issues here (even if they are just to track work external to WinGet so the community can see as they are resolved).

Those other issues should be referenced in "Future Considerations" with links to the issues and we can continue to make progress over time. I have a fairly strong belief that anything we can implement will add value, and it will just continue getting better as we are able to track and close any remaining gaps.

doc/specs/#190 - Proxy Support.md
```diff
@@ -0,0 +1,98 @@
...
+> winget install Contoso.App --no-proxy
+```
+
+To configure the default proxy, a new `proxy` subcommand will be added to the `settings` command, with options to `set` and `reset` the default.
+This will require admin privileges and does not require `ProxyCommandLineArgument` to be enabled.
+
+```
+> winget settings proxy set https://127.0.0.1:2345
+> winget settings proxy reset
+```
```
florelis, line 53: The way I actually implemented this in #4203 was having `set` and `reset` subcommands for `settings` so that we can do
`winget settings set [--setting] <SomeAdminSetting> [--value] <SomeValue>` or `winget settings set [--setting] <SomeAdminSetting>`. Currently the only admin setting that can be used there would be `DefaultProxy`, but I think it could be extensible.

yao-msft (approved):

yao-msft (approved):

PR #4174, opened 2024-02-16
Refactor `IPackage`

JohnMcPMS:
## Change
To better reflect the underlying code and continue to move business logic out of the `CompositeSource`, this change refactors the `IPackage` interface into 3:

```C++
    // Contains a collection of package versions.
    struct IPackageVersionCollection
    {
        // Gets all versions of this package.
        // The versions will be returned in sorted, descending order.
        //  Ex. { 4, 3, 2, 1 }
        virtual std::vector<PackageVersionKey> GetVersionKeys() const = 0;

        // Gets a specific version of this package.
        virtual std::shared_ptr<IPackageVersion> GetVersion(const PackageVersionKey& versionKey) const = 0;

        // A convenience method to effectively call `GetVersion(GetVersionKeys[0])`.
        virtual std::shared_ptr<IPackageVersion> GetLatestVersion() const = 0;
    };

    // Contains information about a package and its versions from a single source.
    struct IPackage : public IPackageVersionCollection
    {
        // Gets a property of this package.
        virtual Utility::LocIndString GetProperty(PackageProperty property) const = 0;

        // Gets the source that this package is from.
        virtual Source GetSource() const = 0;

        // Determines if the given IPackage refers to the same package as this one.
        virtual bool IsSame(const IPackage*) const = 0;

        // Gets this object as the requested type, or null if it is not the requested type.
        virtual const void* CastTo(IPackageType type) const = 0;
    };

    // Contains information about the graph of packages related to a search.
    struct ICompositePackage
    {
        // Gets a property of this package result.
        virtual Utility::LocIndString GetProperty(PackageProperty property) const = 0;

        // Gets the installed package information.
        virtual std::shared_ptr<IPackage> GetInstalled() = 0;

        // Gets all of the available packages for this result.
        // There will be at most one package per source in this list.
        virtual std::vector<std::shared_ptr<IPackage>> GetAvailable() = 0;
    };
```

`IPackageVersionCollection` is simply a set of `IPackageVersion` objects with only the previous `IPackage` functions to enumerate them.  This enables some code that only wants to inspect versions to use a tighter set of APIs and thus some implementations need not worry about things like the appropriate property values to return.

The refactored `IPackage` inherits from `IPackageVersionCollection` and is itself a set of package versions that all come from a single source (as in `Source`).  It adds the properties and the `Source` that the versions are all from.

`ICompositePackage` is the new search result, resembling the previous `IPackage` in what it contains but exposing it more directly.  The installed `IPackage` object will eventually be able to contain more than one installed version in a future change.  The available packages are exposed directly as `IPackage` objects, rather than mixing the versions together in the `CompositeSource`.

The code in `PackageVersionSelection.cpp` extracts the business logic that was inside `CompositeSource` previously, as well as providing a few helpers to make the existing code work more succinctly.  The refactoring is not attempting to change behavior, but simply make it possible to change behavior more easily and precisely in the future.

I would have preferred to go further by splitting `CompositeSource` out into a public type, allowing the other sources to have a simpler model of only ever returning `IPackage` results from search.  This change would have created an even larger churn though and did not directly contribute towards the goal of supporting future side-by-side work.
  
Also changes the YAML object model to use a non-exceptional function to attempt to convert to an integer.

## Validation
Existing regression tests should cover all changes.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4174)


src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -628,7 +629,7 @@ namespace AppInstaller::CLI::Workflow
 
         for (size_t i = 0; i < searchResult.Matches.size(); ++i)
         {
-            auto latestVersion = searchResult.Matches[i].Package->GetLatestAvailableVersion();
+            auto latestVersion = GetAvailableVersionsForInstalledVersion(searchResult.Matches[i].Package, nullptr)->GetLatestVersion();
```
florelis, line 632: Could we add a `GetAvailableVersions(ICompositePackage)` so that we don't have to pass `nullptr`?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -327,163 +340,127 @@ namespace AppInstaller::Repository
             std::shared_ptr<IPackageVersion> m_trackingPackageVersion;
         };
 
-        // A composite package for the CompositeSource.
-        struct CompositePackage : public IPackage
+        // A composite package for the installed package of a CompositePackage.
```
florelis, line 343: "A composite package for [...] a CompositePackage"? Should it just be "A package for [...]"?

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -327,163 +340,127 @@ namespace AppInstaller::Repository
...
+            std::shared_ptr<IPackage> m_package;
+            Source m_trackingSource;
+            std::shared_ptr<IPackageVersion> m_trackingPackageVersion;
+            std::string m_overrideVersion;
+        };
+
+        // A composite package for the CompositeSource.
+        struct CompositePackage : public ICompositePackage
+        {
+            CompositePackage(const std::shared_ptr<ICompositePackage>& installedPackage, const std::shared_ptr<ICompositePackage>& availablePackage = {})
```
florelis, line 422: Maybe add comment that `availablePackage` can only include one package?

florelis (approved):

src/AppInstallerCLICore/ConfigurationWingetDscModuleUnitValidation.cpp
```diff
@@ -358,7 +358,8 @@ namespace AppInstaller::CLI::Configuration
                    {
                        if (!package.Version.empty())
                        {
-                           auto versionKeys = searchResult.Matches.at(0).Package->GetAvailableVersionKeys();
+                           std::shared_ptr<Repository::IPackage> availablePackage = searchResult.Matches.at(0).Package->GetAvailable().at(0);
```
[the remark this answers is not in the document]
JohnMcPMS: This code requires a source name, and at this time there is no way to name a composite source.  Thus, there will only ever be one available package.

As I mentioned in the description, this is a case where separating `CompositeSource` out from `ISource` would be very helpful.  But that change would be even larger.

src/AppInstallerCLICore/Workflows/CompletionFlow.cpp
```diff
@@ -70,11 +70,14 @@ namespace AppInstaller::CLI::Workflow
...
-        for (const auto& vc : context.Get<Execution::Data::Package>()->GetAvailableVersionKeys())
+        for (const auto& ap : context.Get<Execution::Data::Package>()->GetAvailable())
         {
-            if (word.empty() || Utility::ICUCaseInsensitiveStartsWith(vc.Version, word))
+            for (const auto& vc : ap->GetVersionKeys())
             {
-                OutputCompletionString(stream, vc.Version);
+                if (word.empty() || Utility::ICUCaseInsensitiveStartsWith(vc.Version, word))
+                {
+                    OutputCompletionString(stream, vc.Version);
```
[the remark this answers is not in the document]
JohnMcPMS: I think we should set a goal of output in "best match order"; aka the order that places the most likely completion first and so forth.  I believe that PowerShell's implementation does use the order as the tab completion cycle.

I assume what you are indicating is that the order will change when there are multiple available packages.  This is true, but I'm not sure that is incorrect in this case.

florelis (dismissed):

florelis (approved):

yao-msft (approved):
:shipit:

PR #4196, opened 2024-02-23
Make SQLite wrapper more resilient

JohnMcPMS:
We are seeing some errors from the SQLite usage that appear to largely be related to contention and the `FAIL_FAST` behavior for savepoint errors.  This change is attempting to address the issue by making contention less impactful and moving away from `FAIL_FAST`.

## Change
- Adds a busy timeout to SQLite connections by default and provides a method to change the value if callers desire.  This should prevent small contentions from causing errors.
- Moves from using `FAIL_FAST` on critical errors to simply "closing" the connection, making it non-functional for standard use.
- No longer requests this failure behavior for savepoint commit errors; the rollback can handle that if needed.

## Validation
Added tests for the busy timeout and connection closing.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4196)


src/AppInstallerSharedLib/SQLiteWrapper.cpp
```diff
@@ -144,49 +148,78 @@ namespace AppInstaller::SQLite
...
+        sqlite3** SharedConnection::GetPtr()
+        {
+            return &m_dbconn;
+        }
     }
 
     Connection::Connection(const std::string& target, OpenDisposition disposition, OpenFlags flags)
     {
+        m_dbconn = std::make_shared<details::SharedConnection>();
         m_id = GetNextConnectionId();
```
florelis, line 172: Is there any reason to not do these in initializer list?

florelis (dismissed):

src/AppInstallerSharedLib/SQLiteWrapper.cpp
JohnMcPMS: No, but there isn't really a big reason to do them in the list either.

ryfu-msft (approved):

PR #4205, opened 2024-02-27
Cherry-pick #4196 and #4204 to 1.7

JohnMcPMS:
Cherry-pick #4196 and #4204


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2804,14 +2804,17 @@ Please specify one of them using the --source option to proceed.</value>
...
   <data name="APPINSTALLER_CLI_ERROR_ADMIN_CONTEXT_REPAIR_PROHIBITED" xml:space="preserve">
     <value>Repair operations involving administrator privileges are not permitted on packages installed within the user scope.</value>
-  </data>  
+  </data>
   <data name="RepairFailedWithCode" xml:space="preserve">
     <value>Repair failed with exit code: {0}</value>
     <comment>{Locked="{0}"} Error message displayed when an attempt to repair an application package fails. {0} is a placeholder replaced by an error code.</comment>
   </data>
+  <data name="APPINSTALLER_CLI_ERROR_SQLITE_CONNECTION_TERMINATED" xml:space="preserve">
+    <value>The SQLite connection was terminated to prevent corruption.</value>
```
yao-msft, line 2818: We may need to trigger a new localization update for this.

yao-msft (approved):

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: I'm not sure that the error strings really necessitate that, although they can potentially be show to users not using the `error` command if the exception makes to the root.  They were not localized for a very long time though, and I never knew of anyone complaining.

So I wouldn't hold up any builds based on it.

Issue #4213, opened 2024-02-29
Export should allow suppressing missing source warnings


riotrah:
### Description of the new feature / enhancement

I know which packages are installed via winget's pkg repo source, and I want that exact list when `export`ing. 

If I run

`winget export --source winget -o ...`

I'll get a dozen `Installed package is not available from any source:` warnings.

I want to be able to suppress the warnings for unknown sources, especially when I understand that I have mixed-source app installations.

<details><summary>Personal use case context:</summary>
For context, I alias my package manager install/update/uninstall commands (`brew` on mac, `scoop`/`choco` on win) to shell fns that always `export` their lists and commit to dotfiles, so that I am always in sync for my cli tools. Winget's unknown source warnings makes these invocations very noisy.
</details>

I'd open a PR if I knew cpp.

### Proposed technical implementation details

A cli flag that allows me to suppress unknown source warnings

PR #4214, opened 2024-02-29
Add --ignore-warnings to Export command

Trenly:
  * #4213

Manually validated output files remained the same.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4214)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
@@ -67,6 +67,7 @@ namespace AppInstaller::CLI::Workflow
             bool checkVersion)
         {
             std::shared_ptr<IPackageVersionCollection> availableVersions = GetAvailableVersionsForInstalledVersion(package);
+            const bool suppressWarnings = context.Args.Contains(Execution::Args::Type::IgnoreWarnings);
 
             if (!checkVersion)
             {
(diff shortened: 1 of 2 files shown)


src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -98,6 +98,7 @@ namespace AppInstaller::CLI::Workflow
     {
         const auto& searchResult = context.Get<Execution::Data::SearchResult>();
         const bool includeVersions = context.Args.Contains(Execution::Args::Type::IncludeVersions);
+        const bool suppressWarnings = context.Args.Contains(Execution::Args::Type::IgnoreWarnings);
```
yao-msft, line 101: Should this be applied to line 89 as well?

Trenly: Makes sense to me

yao-msft (approved):

yao-msft:
Just leave my thoughts here. In the future, if we want to filter the output, we should do a refactor and do it in reporter level. So we don't need to check every place. Something like Reporter.DisableReportLevel(Level::Warning), then reporter will no longer output for warning level.

Trenly:
> Just leave my thoughts here. In the future, if we want to filter the output, we should do a refactor and do it in reporter level. So we don't need to check every place. Something like Reporter.DisableReportLevel(Level::Warning), then reporter will no longer output for warning level.

Good point; I can try taking a look into that (in a separate PR), though it may be above my skill level

yao-msft:
> Good point; I can try taking a look into that (in a separate PR), though it may be above my skill level

If we can do in the Reporter level, I would discard this pr. This one only disable the warning in this specific workflow task, it will not ignore warnings from previous steps in ExportCommand like HandleSearchResultFailures 

And we don't need to do spot checks at multiple places. the code will be cleaner.

Trenly:
If we can do it at the reporter level, I would revert these changes as part of the PR to handle the reporter refactor, since with that refactor the `--ignore-warnings` arg should likely be added to the common args list rather than individual commands anyways

yao-msft:
> If we can do it at the reporter level, I would revert these changes as part of the PR to handle the reporter refactor, since with that refactor the `--ignore-warnings` arg should likely be added to the common args list rather than individual commands anyways

That's fine too.

Trenly:
```
Failed ConfigureFromHttpsConfigurationFile [282 ms]
  Error Message:
     Expected: 0
  But was:  1
  Stack Trace:
     at AppInstallerCLIE2ETests.ConfigureCommand.ConfigureFromHttpsConfigurationFile() in D:\a\1\s\src\AppInstallerCLIE2ETests\ConfigureCommand.cs:line 182
```

Issue #4046, opened 2024-01-09
winget install hangs indefinitely with MotW Security Warning


jantari:
### Brief description of your issue

When winget-cli downloads an installer, it adds a MotW to it. If the source zone is not trusted, a GUI (!) warning popup shows when winget attempts to run the installer. This popup blocks the install forever and therefore breaks all unattended, automation and remoting scenarios.

### Steps to reproduce

I am unsure why this doesn't reproduce with installers downloaded from the actual www internet (super-duper untrusted), but I can reproduce it reliably with an internal REST source + internally hosted installer files.

My REST source and installer files are hosted with [rewinged](https://github.com/jantari/rewinged) on a routable domain (aka it looks like an internet website to Windows if it doesn't consider that it resolves to a RFC1918 IP? The domain wasn't explicitly added to the Intranet zone in `inetcpl.cpl`)

1. Set up an internal REST source, e.g. at `whatever.company.com`
2. Host installers for internal packages at `whatever.company.com` too
3. Ensure `whatever.company.com` isn't added to the trusted Intranet sites list in `inetcpl.cpl` / the registry
4. Issue a `winget install --id something --source internal` command to attempt a package installation
5. Observe how the MotW Security Warning pops up as a GUI window and the winget-cli process just hangs forever without an error, warning or timeout

### Expected behavior

winget should either ignore/remove or not set the MotW on downloaded installers or it should check for it and abort the installation with an error so that the winget process ends instead of hanging and communicates an actionable error back to the user.

Honestly, since winget is constantly downloading installers from the internet in its default configuration, without issues, I don't see why a MotW should be applied or respected for REST / internal sources. My preferred method would just be to either not create it on download or remove/ignore it before invoking the installer.

### Actual behavior

A MotW GUI (!) security warning pops up **IF** the user is in a scenario where a GUI is present (such as when logged in via RDP) and the winget process hangs indefinitely.

If the user is in a scenario where no GUI is present, such as when running as a service or connecting via ssh, there is no warning or popup or feedback of any kind and the winget install process just hangs indefinitely.

The following screenshot was captured in a GUI session via RDP to make the warning visible:
![image](https://github.com/microsoft/winget-cli/assets/25607767/456bc9b0-66c2-48b9-b1be-16f43301f6eb)

However please note that GUI access is rarely available in modern deployment scenarios and feedback on the CLI is mandatory.

[...]

PR #4216, opened 2024-02-29
Add support for setting source trust level

ryfu-msft:
Summary:
This PR adds support for specifying the trust level and making a source explicit when adding a new source. If a source is `explicit`, the source will not be discoverable unless explictly declared when searching for a package. i.e. `winget search foo.bar --source explicitSource`
- Added two new source arguments for --trust-level and --explicit. 
- Supports specifying multiple trust levels such as `--trust-level 'trusted|storeOrigin` 
- Also went through policy review and got approval for adding these optional fields to the registry. 
- `winget source list` will also show their current state. The only way to remove the trust level or explicit state is by removing the source entirely. 

Added unit and E2E tests to verify these behaviors.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4216)


src/AppInstallerRepositoryCore/ISource.h
```diff
@@ -88,6 +88,12 @@ namespace AppInstaller::Repository
         // Set caller.
         virtual void SetCaller(std::string) {}
 
+        // Require the source to be explicitly declared. Source must be removed for this to be reset.
+        virtual void SetExplicit() {}
```
JohnMcPMS, line 92: I'm not seeing the reason for `ISourceReference` to need these methods yet.  The data is stored in the details anyway, so why is that not the entry point for changing it?

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -45,6 +45,21 @@ namespace AppInstaller::Repository
 
     DEFINE_ENUM_FLAG_OPERATORS(SourceTrustLevel);
 
+    // Converts a string_view to the corresponding SourceTrustLevel enum.
+    SourceTrustLevel ConvertToSourceTrustLevelEnum(std::string_view trustLevel);
+
+    // Converts a SourceTrustLevel enum to the corresponding string.
+    std::string SourceTrustLevelToString(SourceTrustLevel trustLevel);
```
JohnMcPMS, line 52:
```suggestion
    std::string_view SourceTrustLevelToString(SourceTrustLevel trustLevel);
```

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -45,6 +45,21 @@ namespace AppInstaller::Repository
...
+    SourceTrustLevel ConvertToSourceTrustLevelEnum(std::string_view trustLevel);
+
+    // Converts a SourceTrustLevel enum to the corresponding string.
+    std::string SourceTrustLevelToString(SourceTrustLevel trustLevel);
+
+    // Converts a vector of trust level strings to the corresponding SourceTrustLevel enum flag.
+    SourceTrustLevel ConvertToSourceTrustLevelEnum(std::vector<std::string> values);
+
+    // Converts a SourceTrustLevel enum to a list of trust level strings.
+    std::vector<std::string> SourceTrustLevelToList(SourceTrustLevel trustLevel);
```
JohnMcPMS, line 58:
```suggestion
    std::vector<std::string_view> SourceTrustLevelToList(SourceTrustLevel trustLevel);
```

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -320,6 +322,120 @@ namespace AppInstaller::Repository
...
+        }
+
+        return result;
+    }
+
+    std::vector<std::string> SourceTrustLevelToList(SourceTrustLevel trustLevel)
+    {
+        if (trustLevel == Repository::SourceTrustLevel::None)
+        {
+            return { SourceTrustLevelToString(Repository::SourceTrustLevel::None) };
```
JohnMcPMS, line 393: It seems to me that it makes more sense to return an empty vector, which would mean just removing this entire half od the if/else.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -834,9 +966,18 @@ namespace AppInstaller::Repository
                 // to avoid the progress bar fill up multiple times.
                 AddOrUpdateResult updateResult = UpdateSourceFromDetails(details, progress);
 
+                // Update the trust level if it does not match what was previously recorded in the source list.
+                auto detailsInternal = sourceList.GetSource(details.Name);
+                if (detailsInternal->TrustLevel != details.TrustLevel)
+                {
+                    sourceList.RemoveSource(details);
```
JohnMcPMS, line 973: I'm not sure what this is accomplishing...

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -196,9 +210,9 @@ namespace AppInstaller::Repository
                 Utility::CaseInsensitiveEquals(left.Type, right.Type);
         }
 
-        bool ShouldBeHidden(const SourceDetailsInternal& details)
+        bool ShouldBeHidden(const SourceDetailsInternal& details, bool allowExplicitSources)
```
JohnMcPMS, line 213: My instinct is that `Explicit` only needs to be handled when constructing the composite source from the source list.  Passing a bool to all of these functions seems very fragile.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2861,4 +2861,16 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="SettingsSetCommandShortDescription" xml:space="preserve">
     <value>Sets the value of an admin setting.</value>
   </data>
+  <data name="SourceRequireExplicitArgumentDescription" xml:space="preserve">
+    <value>Excludes a source from discovery by default unless explicitly declared</value>
```
JohnMcPMS, line 2865: This is fairly long for the blurb in the help.

JohnMcPMS (commented):
I'm not sure that `StoreOrigin` should even be exposed.  It doesn't really hurt anything directly, but I also don't foresee anyone using it (except us for testing).  My expectation is only that it causes confusion.  I would suggest that it be supported but undocumented (including in the argument description and error messages).

src/AppInstallerCLICore/Workflows/SourceFlow.cpp
```diff
@@ -107,8 +107,23 @@ namespace AppInstaller::CLI::Workflow
...
             std::string_view arg = context.Args.GetArg(Args::Type::SourceArg);
             std::string_view type = context.Args.GetArg(Args::Type::SourceType);
+            bool isExplicit = context.Args.Contains(Args::Type::SourceExplicit);
 
-            Repository::Source sourceToAdd{ name, arg, type };
+            Repository::SourceTrustLevel trustLevel = Repository::SourceTrustLevel::None;
+            if (context.Args.Contains(Execution::Args::Type::SourceTrustLevel))
+            {
+                std::vector<std::string> trustLevelArgs = Utility::Split(std::string{ context.Args.GetArg(Execution::Args::Type::SourceTrustLevel) }, '|');
+                std::vector<std::string> trustLevels;
```
yao-msft, line 116: nit: a vector copy is probably not needed.

```
                for (auto& trustLevelArg : trustLevelArgs)
                {
                    Utility::Trim(trustLevelArg);
                }
```

src/AppInstallerCLICore/Command.cpp
```diff
@@ -740,6 +740,26 @@ namespace AppInstaller::CLI
             }
         }
 
+        if (execArgs.Contains(Execution::Args::Type::SourceTrustLevel))
```
yao-msft, line 743: This check is specific to "source add", move to ValidateArguments of SourceAdd command

src/AppInstallerSharedLib/Public/winget/GroupPolicy.h
```diff
@@ -84,6 +84,8 @@ namespace AppInstaller::Settings
         std::string Type;
         std::string Data;
         std::string Identifier;
+        std::vector<std::string> TrustLevel;
```
yao-msft, line 87: How is this finally converted to be usable in Source?

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -669,6 +696,8 @@ namespace AppInstaller::Repository
                         details.Data = additionalSource.Data;
                         details.Identifier = additionalSource.Identifier;
                         details.Origin = SourceOrigin::GroupPolicy;
+                        details.TrustLevel = Repository::ConvertToSourceTrustLevelEnum(additionalSource.TrustLevel);
```
yao-msft, line 699: try catch and make it to none

src/AppInstallerSharedLib/Public/winget/GroupPolicy.h
yao-msft: My bad, it's done in SourceList

src/AppInstallerCLICore/Workflows/SourceFlow.cpp
ryfu-msft: Updated Utility::Split so that it accepts an optional bool parameter for trimming.

src/AppInstallerSharedLib/GroupPolicy.cpp
```diff
@@ -336,6 +354,15 @@ namespace AppInstaller::Settings
         json["Arg"] = Arg;
         json["Data"] = Data;
         json["Identifier"] = Identifier;
+        json["Explicit"] = Explicit;
+
+        // Trust level is represented as an array of trust level strings since there can be multiple flags set.
+        int i = 0;
+        for (std::string entry : TrustLevel)
```
JohnMcPMS, line 361: nit: There is a `for` loop syntax that is designed just for this...

JohnMcPMS (dismissed):

src/AppInstallerSharedLib/GroupPolicy.cpp
ryfu-msft: 🤦

src/AppInstallerSharedLib/GroupPolicy.cpp
```diff
@@ -357,7 +357,8 @@ namespace AppInstaller::Settings
         json["Explicit"] = Explicit;
 
         // Trust level is represented as an array of trust level strings since there can be multiple flags set.
-        for (int i = 0; i < TrustLevel.size(); ++i)
+        int trustLevelLength = static_cast<int>(TrustLevel.size());
+        for (int i = 0; i < trustLevelLength; ++i)
```
JohnMcPMS, line 361: nit: The better solution is generally to use `size_t i`.

JohnMcPMS (approved):

jantari:
Is this the fix for https://github.com/microsoft/winget-cli/issues/4046 (installations from REST sources hanging indefinitely due to MotW)?

denelon:
@jantari, yes, this work will accrue to unblocking the hang.

PR #4221, opened 2024-03-01
Make Ignore Warnings suppress the output level entirely

### LINKED ISSUE #4213: Export should allow suppressing missing source warnings

Trenly:
  * Resolves #4213 
  * Closes #4214

Instead of handling each case individually, this PR makes the `--ignore-warnings` a common argument and suppresses all output from `context.Reporter.Warn()`. Support is added for enabling or disabling each reporter level individually.

cc @yao-msft 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4221)


src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -180,6 +201,9 @@ namespace AppInstaller::CLI::Execution
         wil::srwlock m_progressCallbackLock;
         std::atomic<ProgressCallback*> m_progressCallback;
         std::atomic<IProgressSink*> m_progressSink;
+
+        // Enable all levels by default
+        std::vector<Level> m_enabledLevels = { Level::Verbose, Level::Info, Level::Warning, Level::Error };
```
yao-msft, line 206: Make the Level enum as flag and use m_enabledLevels as flag

src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -165,9 +165,30 @@ namespace AppInstaller::CLI::Execution
...
+            else {
+                // Remove the level from the list of enabled levels
+                m_enabledLevels.erase(std::remove(m_enabledLevels.begin(), m_enabledLevels.end(), reporterLevel), m_enabledLevels.end());
+            }
+        }
+
     private:
         Reporter(std::shared_ptr<BaseStream> outStream, std::istream& inStream);
 
+        bool levelEnabled(Level reporterLevel) {
```
yao-msft, line 188: nit: code style: function name is PascalCase and our code block starts a new line with '{', applies to all new code added in this file
```suggestion
        bool LevelEnabled(Level reporterLevel)
        {

```

src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -165,9 +165,30 @@ namespace AppInstaller::CLI::Execution
             m_progressSink = sink;
         }
 
+        void SetLevelEnabled(Level reporterLevel, bool setEnabled = true) {
+            const bool isEnabled = levelEnabled(reporterLevel);
```
yao-msft, line 169:
```
void SetLevelEnabled(Level reporterLevel, bool setEnabled = true)
{
    if (setEnabled)
    {
        WI_SetFlag(m_enabledLevels, reportLevel)
    }
    else
    {
        WI_ClearFlag(m_enabledLevels, reportLevel)
    }
}

File: src/AppInstallerCLICore/ExecutionReporter.cpp [hunk 4]
```diff
@@ -54,6 +54,12 @@ namespace AppInstaller::CLI::Execution
 
     OutputStream Reporter::GetOutputStream(Level level)
     {
+        // If the level is not enabled, return a default stream which is disabled
+        if (!levelEnabled(level))
```
yao-msft, line 58:
```suggestion
        if (WI_IsFlagSet(m_enabledLevels, level))

```
LevelEnabled() function could be removed Or check the flag within LevelEnabled() function if you think we still need that function;

yao-msft (commented):
Thanks to previous build test failures, we can now truly "discard" the previous pr :)

src/AppInstallerCLICore/ExecutionReporter.h
Trenly: This makes a lot of sense, I see what you're getting at and how the flag (bitmask) works. What I don't quite get is how to make the Level enum as a flag

yao-msft: [Code](https://github.com/microsoft/winget-cli/blob/c5fbdd9fb9d53336bad182b82f255b2e3f691b10/src/AppInstallerCommonCore/Public/AppInstallerProgress.h#L41) to enable flags

Trenly: You're the best; Should have updates for this PR in a bit

yao-msft: You can define `All = Verbose | Info | Warning | Error` in the enum definition to make life easier as well

Trenly: Okay, this is getting to be an area where I clearly have no idea how anything works anymore. I've specified the bitmasks, and dropped in the `DEFINE_ENUM_FLAG_OPERATORS`, which results in `linkage specification is not allowed`

Then, in `SetLevelEnabled` I get the error `expression must have a constant value` pointing to `reporterLevel` which cannot be used as a constant value

denelon:
Thanks @Kaleb with a FANTASIC improvement for users! ❤️

src/AppInstallerCLICore/ExecutionReporter.h
yao-msft: Sorry for the confusion, please use WI_SetAllFlags and WI_ClearAllFlags for variables. WI_SetFlag and WI_ClearFlag require the second parameter to be compile time constant.

mdanish-kh (commented):
Output nit: We can get rid of the period at the end to make it consistent with the existing output
image

src/AppInstallerCLICore/ExecutionReporter.h
Trenly: > Sorry for the confusion, please use WI_SetAllFlags and WI_ClearAllFlags for variables. WI_SetFlag and WI_ClearFlag require the second parameter to be compile time constant.

That seemed to resolve that error, but there still seems to be something weird going on with the `DEFINE_ENUM_FLAG_OPERATORS`. If I put it before the Reporter struct, then I get that the `Reporter::Level` must be initialized first, but if I put it after the reporter struct then the functions inside the struct try to use an incorrect flag. If I put it inside the struct, I get the linkage error

image

Trenly: I just moved the enum outside the struct, since that seemed cleanest

src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -23,6 +23,18 @@ namespace AppInstaller::CLI::Execution
...
+    enum class ReporterLevel : uint32_t
+    {
+        None = 0x0,
+        Verbose = 0x1,
+        Info = 0x2,
+        Warning = 0x4,
+        Error = 0x8,
+        All = Verbose | Info | Warning | Error,
+    };
+    DEFINE_ENUM_FLAG_OPERATORS(ReporterLevel);
```
JohnMcPMS, line 36: If you previously had tried to place `DEFINE_ENUM_FLAG_OPERATORS` inside the `Reporter` struct alongside the enum, that was the issue.  It would need to be outside of the struct and reference a more qualified name, like: `DEFINE_ENUM_FLAG_OPERATORS(Reporter::Level)`.

src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -165,9 +168,20 @@ namespace AppInstaller::CLI::Execution
             m_progressSink = sink;
         }
 
+        void SetLevelEnabled(ReporterLevel reporterLevel, bool setEnabled = true)
```
JohnMcPMS, line 171: Passing `None` to this method has ambiguous meaning.  For instance, what do I want from `SetLevelEnabled(None, true)`?  Do I want no levels enabled?  And does `false` then mean I want no levels disabled?  If those are the cases, then the logic is wrong.  In fact, passing `None` will have no effect on the value of `m_enabledLevels` regardless of `setEnabled`.

I think the easy solution is a name change:
```suggestion
        void SetLevelMask(ReporterLevel reporterLevel, bool setEnabled = true)
```
To move the noun from "level" to "mask", which better indicates the bitrack and thus the meaning of what will happen when `None` is passed.  If you want to work "enabled" back in, that is probably fine too.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1064,7 +1064,7 @@ namespace AppInstaller::CLI::Workflow
             }
 
             auto promptString = m_isApply ? Resource::String::ConfigurationWarningPromptApply : Resource::String::ConfigurationWarningPromptTest;
-            if (!context.Reporter.PromptForBoolResponse(promptString, Reporter::Level::Warning))
+            if (!context.Reporter.PromptForBoolResponse(promptString, ReporterLevel::Warning))
```
JohnMcPMS, line 1067: This brings up an issue; if you have disabled `Warning` and we prompt at `Warning`, are we just going to sit waiting for user input without having actually issued a prompt on the console?

Maybe prompts need to enable all levels while they are active?

Alternatively, prompts at disabled levels could simply apply their non-interactive choice; after all, the user requested no warnings, and this prompt is a warning...

Trenly: I like the flow of appearing as their non-interactive choice; I think it makes for a cleaner flow. Will update

src/AppInstallerCLICore/ExecutionReporter.h
Trenly: In this case, I think the easiest solution is to remove `None` from the enum entirely, as it has no real purpose. It isn't necessary for any behaviors currently, and to turn off all reporting would be functionally the same as using `All` with `false`

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
JohnMcPMS: That might also require checking all of the prompts for the levels that they output at.  For instance, `InstallerAbortsTerminalPrompt` uses warnings by default and thus would need to check the warning level.

src/AppInstallerCLICore/ExecutionReporter.h
JohnMcPMS: The only purpose I see for `None` is to initialize the field if we wanted to start with no levels enabled instead of all.  Removing it is a valid option, but I personally would lean towards keeping `None` and changing the name of the method.  Not a big sticking point though.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
Trenly: > That might also require checking all of the prompts for the levels that they output at. For instance, `InstallerAbortsTerminalPrompt` uses warnings by default and thus would need to check the warning level.

It would require going through all the prompts and updating either way - either to re-enable the warning level temporarily, or to make it non-interactive.

The other approach could be to add an optional boolean to `Reporter.Warn()`, `Reporter.Info`, etc. for bypassing suppression

yao-msft: Or maybe expose the m_isEnabled in OutputStream class, then within the PromptFor* method implementations, after GetOutputStream(level), we can check if it's enabled and act accordingly.

src/AppInstallerCLICore/ExecutionReporter.h
yao-msft: I believe we have existing usage [here](https://github.com/microsoft/winget-cli/blob/464e177f3b355ffef3d9f96fdebed9004407d181/src/AppInstallerRepositoryCore/Microsoft/Schema/ISQLiteIndex.h#L121) Does making the variables fully qualified help?

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
Trenly: > Or maybe expose the m_isEnabled in OutputStream class, then within the PromptFor* method implementations, after GetOutputStream(level), we can check if it's enabled and act accordingly.

I tried that actually, but every time `Reporter.Warn()` is called it goes through `GetBasicOutputStream` it means that there is not a persistent stream for any of the reporter levels and so `m_enabled` doesn't actually have any effect. This is probably a flaw in the implementation, but not one that I know how to fix (as evidenced by all my troubles even using the simple enum as flags)

src/AppInstallerCLICore/ExecutionReporter.h
Trenly: Hmm. I thought I had tried that already, but perhaps I overlooked it. Will check

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
yao-msft: Maybe we are not looking at the same place? I was referring to [here](https://github.com/microsoft/winget-cli/blob/464e177f3b355ffef3d9f96fdebed9004407d181/src/AppInstallerCLICore/ExecutionReporter.cpp#L173), within  PromptForBool, PromptForEnter, PromptForPath implementation, after we get the stream, we can check if the stream is enabled(.m_enabled to be exposed) and act accordingly. So we don't need to go over every calling place.

Trenly: > Maybe we are not looking at the same place? I was referring to [here](https://github.com/microsoft/winget-cli/blob/464e177f3b355ffef3d9f96fdebed9004407d181/src/AppInstallerCLICore/ExecutionReporter.cpp#L173), within PromptForBool, PromptForEnter, PromptForPath implementation, after we get the stream, we can check if the stream is enabled(.m_enabled to be exposed) and act accordingly. So we don't need to go over every calling place.

Ah, I see. I thought you were talking about setting m_enabled to be false for the stream instead of using the flags. 

Yes, I think it would be easy enough to expose m_enabled and check it

riotrah:
Sorry for adding noise - I just wanted to say I'm really impressed with and appreciate the insane response time to #4213

src/AppInstallerCLICore/ExecutionReporter.h
Trenly: Even making the variables fully qualified results in the same error . . .

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
Trenly: After looking at the code, I thought it would be best to make the output stream check at the `IsInteractivityAllowed` level to take advantage of the already existing behaviors for when interactivity is not allowed. Since all of the `PromptFor` methods had a `Reporter::Level` defined, I removed the default of `Reporter::Level::Info` to make it more clear which level each prompt is at, and added that same level to the Interactivity check. Strictly speaking, there isn't the ability to disable the `Reporter::Level::Info` currently, but wanted to be sure that it wouldn't be missed if anyone goes to update the behavior in the future.

src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -165,9 +167,14 @@ namespace AppInstaller::CLI::Execution
             m_progressSink = sink;
         }
 
+        void SetLevelMask(Level reporterLevel, bool setEnabled = true);
+
+        Level GetLevelMask() {
```
yao-msft, line 172: nit: we would not publicly expose internal flags. Use the method IsLevelEnabled() in the previous iteration and let caller call it.

src/AppInstallerCLICore/Workflows/PromptFlow.cpp
```diff
@@ -5,13 +5,29 @@
...
 
+using namespace AppInstaller::CLI::Execution;
 using namespace AppInstaller::Settings;
 using namespace AppInstaller::Utility::literals;
 
 namespace AppInstaller::CLI::Workflow
 {
     namespace
     {
+        bool IsInteractivityAllowed(Execution::Context& context, Reporter::Level reporterLevel) {
```
yao-msft, line 16:
```
bool IsPromptAllowed(Execution::Context& context, Reporter::Level reporterLevel)
{
    return IsInteractivityAllowed(context) && context.Reporter.IsLevelEnabled(reporterLevel);
}
```

Use a different method name to more precisely describe the function. IsInteractivityAllowed is more of DisableInteractivity flag focused. Also, all reporter level checking should be performed within Reporter, it's not the caller's responsibility to interprete internal Reporter flags.

src/AppInstallerCLICore/ExecutionReporter.h
```diff
@@ -98,13 +100,13 @@ namespace AppInstaller::CLI::Execution
         void SetStyle(AppInstaller::Settings::VisualStyle style);
 
         // Prompts the user, return true if they consented.
-        bool PromptForBoolResponse(Resource::LocString message, Level level = Level::Info);
+        bool PromptForBoolResponse(Resource::LocString message, Level level);
```
yao-msft, line 103: We forgot the PromptForBoolResponse in ConfigurationFlow.cpp, which also needs to be handled. And that one is almost always prompted. It's the "Do you want to apply the device configuration?" before winget configure apply changes to the device.

Check IsEnabled() within PromptFor* has one benefit that all future invocations do not need to remember to check reporter level before prompt. Regarding the concern that different caller expect different results, we can make this method like `bool PromptForBoolResponse(Resource::LocString message, Level level, bool resultIfDisabled)` so in case the level is disabled, just return resultIfDisabled.

Same for PromptForPath, where caller can say return empty string is disabled. Then treat empty string as invalid imput and return "Install Location not provided."

But I'm ok with either way at the moment.

Trenly: Makes complete sense to me - Since checking within the PromptFor* means one less thing to remember, I went with that.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1064,7 +1064,7 @@ namespace AppInstaller::CLI::Workflow
             }
 
             auto promptString = m_isApply ? Resource::String::ConfigurationWarningPromptApply : Resource::String::ConfigurationWarningPromptTest;
-            if (!context.Reporter.PromptForBoolResponse(promptString, Reporter::Level::Warning))
+            if (!context.Reporter.PromptForBoolResponse(promptString, Reporter::Level::Warning, true))
```
yao-msft, line 1067:
```suggestion
            if (!context.Reporter.PromptForBoolResponse(promptString, Reporter::Level::Warning))

```

Let's make it default to false. Device configuration is quite a big change to the current device. We do want user confirm before execution. Plus we provided the ConfigurationAcceptWarning arg like AcceptPackageAgreements

src/AppInstallerCLICore/Workflows/PromptFlow.cpp
```diff
@@ -295,6 +296,10 @@ namespace AppInstaller::CLI::Workflow
 
                 AICLI_LOG(CLI, Info, << "Prompting for install root.");
                 m_installLocation = context.Reporter.PromptForPath(Resource::String::PromptForInstallRoot);
+                if (m_installLocation.empty())
+                {
```
yao-msft, line 300: Add logging and user message for this too
```
AICLI_LOG(CLI, Error, << "Install location is required but was not provided.");
context.Reporter.Error() << Resource::String::InstallLocationNotProvided << std::endl;
```

yao-msft (approved):

yao-msft (dismissed):

Trenly:
Thanks for all your patience and guidance @yao-msft, It really helps me understand a lot more about why certain things have been done the way they are

yao-msft:
@trenly I cannot merge due to merge conflicts. Can you merge to latest when you have time? Thanks.

Trenly:
> @Trenly I cannot merge due to merge conflicts. Can you merge to latest when you have time? Thanks.

Should be good now

yao-msft (approved):

Issue #3805, opened 2023-10-23
Winget PowerShell commands failing


RygelXVI:
### Brief description of your issue

attempting to use any winget powershell commands will fail with "Get-WinGetPackage: Failed to create instance: -2147483633"
using `Get-WinGetPackage -Verbose -Debug` does not produce further information.
using `try { get-wingetpackage } catch { $_.exception }` captures the following stacktrace:
```
ErrorCode      : -2147483633
TargetSite     : T Create[T](System.Type, System.Guid ByRef)
Message        : Failed to create instance: -2147483633
Data           : {}
InnerException :
HelpLink       :
Source         : Microsoft.WinGet.Client.Engine
HResult        : -2147483633
StackTrace     :    at Microsoft.WinGet.Client.Engine.Helpers.ComObjectFactory.Create[T](Type type, Guid& iid)
                    at
                 Microsoft.WinGet.Client.Engine.Helpers.ComObjectFactory.CreateCreateCompositePackageCatalogOptions()
                    at Microsoft.WinGet.Client.Engine.Commands.Common.FinderCommand.GetPackageCatalogReference(Composit
                 eSearchBehavior behavior)
                    at Microsoft.WinGet.Client.Engine.Commands.Common.FinderCommand.GetPackageCatalog(CompositeSearchBe
                 havior behavior)
                    at
                 Microsoft.WinGet.Client.Engine.Commands.Common.FinderCommand.FindPackages(CompositeSearchBehavior
                 behavior, UInt32 limit)
                    at Microsoft.WinGet.Client.Engine.Commands.Common.FinderExtendedCommand.FindPackages(CompositeSearc
                 hBehavior behavior)
                    at Microsoft.WinGet.Client.Engine.Commands.FinderPackageCommand.Get()
                    at Microsoft.WinGet.Client.Commands.GetPackageCmdlet.ProcessRecord()
                    at System.Management.Automation.CommandProcessor.ProcessRecord()
```

### Steps to reproduce

use a winget powershell command, e.g.
`get-wingetpackage | where { $_.Isupdateavailable }`

### Expected behavior

a list of packages with updates available to be displayed in the shell (in fact the same list of packages that would be displayed using `winget upgrade`

### Actual behavior

error message "Get-WinGetPackage: Failed to create instance: -2147483633" is returned

### Environment

```shell
Windows Package Manager (Preview) v1.7.2782-preview
Copyright (c) Microsoft Corporation. All rights reserved.

[...]

PR #4251, opened 2024-03-07
Add Arm64 to list of architectures in Module Initialization.

snickler:
- [X] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs).
- [X] This pull request is related to an issue.
  - Resolves: #4169
-----

This PR addresses an issue that prevents the WinGet Cmdlets from loading in an Arm64 PowerShell 7.x environment.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4251)

PR #4260, opened 2024-03-12
MSIX Workaround: Execute InProc when running from AppContainer

snickler:
- [X] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs).
- [X] This pull request is related to an issue.
Resolves: #3805 
-----

In troubleshooting efforts for this issue, I realized that I was able to properly execute the WinGet commands from an MSIX PowerShell 7.x distribution if I ran it as `SYSTEM`. This gave me the idea to make the commands execute InProc only if the current process has a `PackageFullName`. 

This effort is closely related to #4251.


snickler:
@msftrubengu - Mind checking this out? :)

JohnMcPMS:
There is currently a disadvantage of in-proc usage; it stores state in a completely separate location from the packaged out-of-proc (OOP) server.  This would lead to settings/pinning/tracking being split between running `winget.exe` and any PowerShell command, even from the same shell process.  I would prefer to avoid that if we can.

The blocker for this working is that our current implementation that enables calling OOP (admittedly, a workaround itself) is not in effect when run packaged.  The fix for that is to change the implementation to be more targeted in its scope, allowing it to be applied in all situations.

If you are extremely motivated:
1. Add a new export to the [winrtact.dll project](https://github.com/microsoft/winget-cli/tree/master/src/Xlang/UndockedRegFreeWinRT/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT) like `HRESULT SetWinRTNamespaceWinMDFilePath(string namespace, string filepath)` (where the strings are whatever the right marshallable-from-.NET type is)
2. Selectively remove the existing "If .winmd not found look next to *this* DLL" [implementation](https://github.com/microsoft/winget-cli/commit/95d3926d6c4c989ba81f3f3baa7855f9cc2cc746) (mostly the part in `ResolveThirdPartyType`)
3. In `ResolveThirdPartyType` (such that it applies to both packaged and not), use the namespace/filepath map to calculate a result if one was not found "normally".  This likely requires replicating the filename search algorithm, just using the map as the "filesystem".
4. Call the new export with the path to the .winmd, and probably move that file around a bit so that we don't need 4 copies (since it is currently required to be collocated).

See, an easy, ~~3~~ 4 step process...

mominshaikhdevs:
irrelevant information below.

### MSIX Packaging ≠ AppContainer. 

* **MSIX Packaging = _Applications Data/Resources Virtualization._**

* **AppContainer = _Processes Isolation/Sandboxing._** 
It's an NT Kernel feature. Any processes (including WSL1 processes) can be run inside AppContainer, through tools like  [AppContainer-Launcher](https://github.com/WildByDesign/AppContainer-Launcher) without any sort of MSIX Packaging. It just happens to be that Microsoft has advertised as if they are the same thing, but in reality, they are completely independent technology.

snickler:
I don't believe this is needed any longer due to #4392.

Issue #2129, opened 2022-04-27
Better support for Side by Side packages


JohnMcPMS:
### Description of the new feature / enhancement

winget should better support multiple versions of the same package being installed on the same machine.  This happens, and we currently try to pretend it doesn't.

### Proposed technical implementation details

Package detection on the system needs to be improved to correlate package versions together.

Replace `PackageVersion InstalledVersion` with `PackageVersion[] InstalledVersions` internally, then reasoning about how that affects everywhere that breaks.

denelon:
@JohnMcPMS I think this is a great call out. I think we should also take this into consideration as it relates to channels and how we can handle software that doesn't install side by side.

I know Ultimaker.Cura as an example of one package where the .exe during upgrade simply installs a new version of the package. We should also include the impact on the import scenario.

JohnLukeBentley:
> I think we should also take this into consideration as it relates to channels and how we can handle software that doesn't install side by side.

Another example in my case is with respect to Typora. https://typora.io/

I currently have a dev channel release installed, 1.3.3-dev (From https://typora.io/releases/all > [Dev/Beta Releases])

    > winget upgrade 

... gives ...

    Typora 1.3     Typora.Typora       1.3.3          1.3.6         winget
    
... That is, winget is not recognizing the dev channel and, if allowed, will (undesirably) upgrade the dev channel 1.3.3-dev to the stable release channel 1.3.6 (From https://typora.io/releases/all > [Stable releases]).

Unless, that is, that the Typora developer is failing to configure their winget manifest correctly (but given this issue is open presumably multiple channels is not supported in winget currently)

JohnLukeBentley:
#147 "Support different release channels [released, beta, alpha]" is the better place for the channels issue.

ppvnf:
I don't really understand why Microsoft Visual C++ Redistributables have different packages for x86 and x64 architectures, but .NET Desktop Runtimes don't have both as separate packages.

jedieaston:
Dependency management, mostly. We currently can't say in the manifest that a package relies on a certain architecture of another package, so we have to have per architecture pacakges for the VCRedists (many packages specifically want the x86 redists). And no one has added a package that relies on a certain architecture of the .NET Desktop Runtime.

Trenly:
[Policy] Area-Matching

dryo:
[...]

Issue #1886, opened 2022-02-01
Attempting to uninstall a specific version of a package doesn't work


denelon:
### Brief description of your issue

I tried to uninstall a specific version of a package, and it didn't appear to work. I have multiple versions of "Windows PC Health Check" and have been in the upgrade loop. I tried to uninstall the oldest version, and it didn't do what I expected

### Steps to reproduce

More than one version of "Windows PC Health Check" are installed.

Try to remove the "oldest" version that shows an upgrade is available.

`winget uninstall Microsoft.WindowsPCHealthCheck -v 2.5.2107.27001`

### Expected behavior

I expected the version specified to be removed.

### Actual behavior

A different version of the package is removed.

### Environment

```shell
Windows Package Manager Version: v1.2.10271
App Installer Version: v1.17.10271.0
```

denelon:
![Screenshot 2022-01-31 165926](https://user-images.githubusercontent.com/61799811/151897585-fb1216ed-77eb-4b69-b428-f8736c5a93a6.png)

denelon:
We've been working on the side-by-side scenarios. You could try the latest release and enable the experimental feature on:
* https://github.com/microsoft/winget-cli/releases/tag/v1.8.924-preview

Please provide feedback at:
* https://github.com/microsoft/winget-cli/discussions/4281

PR #4282, opened 2024-03-18
Add experimental support for multiple installed versions

JohnMcPMS:
Movement toward north star #2129

## Change
This change introduces a new experimental feature flag called `sideBySide` that enables multiple instances of an installed package to be found by search results.  The `CompositeSource` will fold these all together into a single package result through their common association with available packages.

> [!WARNING]
> This experimental feature affects the search results that the rest of the Windows Package Manager code relies on.  Enabling it may change the behavior of downstream components.

> [!IMPORTANT]
> Only the commands in `winget.exe` have been updated to account for this change.  Other interfaces will get reasonable defaults, but may not be able to interact properly in all cases.

### Installed identifier change
The most visible effect of this feature being enabled is that the identifier generated for installed items is no longer simply the registry key name or package family name but is instead constructed to make it fully unique.  This is mostly relevant to packages without an available counterpart.  For packages that are registered with the system through the registry, the identifier is now of the form `ARP\User\x64\ProductCode`.  For MSIX, the form is `MSIX\Package_Full_Name`.  The previous identifiers are still searchable data though, so it is hopefully not more difficult to `list` installed items.

### Installed item correlations
The implementation in `CompositeSource` tries to separate the steps into:
1. enumerating the basic results with each having at most one installed item and its correlated available packages.
2. folding the results together to create the final return value.

In an effort to prevent overly broad correlations from available to installed items, there is an additional attempt to ensure that there are no stronger correlations from installed to available.  As an example, if a package is split by version while still sharing a name that is reasonably close, searching for one of the version subsets should not attempt to associate all of the installed items so long as the other versions have stronger correlations.
1. Package `A1` has versions 1.0 - 1.5 with names like `A Package (1.0)`
2. Package `A2` has versions 2.0 - 2.2 with names like `A Package (2.0)`
3. Searching for `A2` would correlate with version `1.0` because the normalized names are the same
4. With the above check, it should not return installed version 1.0 as long as the product code is associated with `A1`

Once the basic results are discovered, the results are folded together.  A comment on that function indicates the methodology and possible future improvements:

```C++
// Group results in an attempt to have a single result that covers all installed versions.
// This is expected to be called immediately after the installed search portion,
// when each result will contain a single installed version and some number of available packages.
// 
// The folds that happen are:
//  1. When results have the same primary available package (the primary available package is set due to tracking data)
//  2. When a result has no primary available package, but another result does have a primary that matches one of the available
//      a. Choose the latest primary if there are multiple
//  3. When multiple results have no primary available package and share the same available package set
//      a. There are many potential additional rules that could be made here, but we will start with the simplest version.
//
// Potential improvements:
//  1. Attempting correlation of non-primary available packages to allow folding in more complex cases
//      a. For example, if installed A has {source1:package1, source2:package2} and installed B has {source1:package1}, can we
//          make sure that source1:package1 and source2:package2 are in fact "the same" to confidently say that installed A and B
//          are side by side versions.
//  2. Attempt correlation by installed data only
//      a. We can potentially detect multiple instances of the same installed item with the same correlation logic turned back on
//          the installed source.  This would allow for folding even when the package is not in any available source.
```

#### list
`list` has been updated to simply output all versions contained within a single search result as independent lines.  This keeps the resulting output basically the same, but that has the downside of not explicitly indicating the correlation between the items.

#### repair
`repair` was updated internally so that passing in a `--version` would cause that specific version to be targeted for repair.

#### uninstall
`uninstall` should behave as before if only a single installed version is present.  When multiple installed versions exist, an error is produced unless one of two new arguments is supplied:
1. `--all-versions` / `--all` will uninstall all installed versions
2. `--version` / `-v` will uninstall the targeted version only

### Discussion
I have created a discussion for this experimental feature: #4281 

The goal is to allow feedback to be better collected and threaded, although I expect that there will be issues created independently as well.

### Additionally...
I added a script in the `tools` that launches the sandbox with the dev package registered: Launch-DevPackageInSandbox.ps1

Since it is dev registered in the sandbox, all you need to do to update it is build and deploy from Visual Studio and the sandbox version will just use those latest files.

## Validation
Manual validation of the target scenario (multiple installed versions).
Unit tests added for core folding behavior.
More tests should be added before / as part of moving out of experimental.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4282)


Trenly:
> #### uninstall
> `uninstall` should behave as before if only a single installed version is present. When multiple installed versions exist, an error is produced unless one of two new arguments is supplied:
> 
> 1. `--all-versions` / `--all` will uninstall all installed versions
> 2. `--version` / `-v` will uninstall the targeted version only

Will this also resolve this issue once out of experimental?
* #1886

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -1349,7 +1368,38 @@ namespace AppInstaller::CLI::Workflow
...
+
+                    if (!installedVersion)
+                    {
+                        context.Reporter.Error() << Resource::String::GetManifestResultVersionNotFound(Utility::LocIndView{ versionKey.Version }) << std::endl;
+                        // This error maintains consistency with passing an available version to commands
+                        AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_NO_MANIFEST_FOUND);
+                    }
+
+                    context.Add<Execution::Data::InstalledPackageVersion>(std::move(installedVersion));
+                }
```
JohnMcPMS, line 1392: @Trenly , selection of target version.

JohnMcPMS:
> > #### uninstall
> > `uninstall` should behave as before if only a single installed version is present. When multiple installed versions exist, an error is produced unless one of two new arguments is supplied:
> > 
> > 1. `--all-versions` / `--all` will uninstall all installed versions
> > 2. `--version` / `-v` will uninstall the targeted version only
> 
> Will this also resolve this issue once out of experimental? I only ask because I don't see anywhere in the uninstall flow where `--version` is being checked, unless I'm missing it in one of the other function calls
> 
> * [Attempting to uninstall a specific version of a package doesn't work #1886](https://github.com/microsoft/winget-cli/issues/1886)

Yes, I believe it would resolve that issue.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
Trenly: Thanks! I must have missed that

yao-msft (approved):

src/AppInstallerRepositoryCore/CompositeSource.cpp
```diff
@@ -812,95 +1123,218 @@ namespace AppInstaller::Repository
...
-                        data.AddIfNotPresent(SystemReferenceString{
-                            PackageMatchField::NormalizedNameAndPublisher,
-                            names[i],
-                            publishers[j] });
+                        InstalledResultFoldData* latestPrimaryAvailable = nullptr;
+                        std::vector<InstalledResultFoldData*> availableFoldData;
+
+                        for (const auto& availablePackage : currentMatch.Package->GetAvailablePackages())
+                        {
+                            auto& packageFoldData = foldData.at(availablePackage);
```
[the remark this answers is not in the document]
JohnMcPMS: Yes, there is a non-explicit constructor from a `std::shared_ptr<IPackage>`.

yao-msft (approved):

Issue #4284, opened 2024-03-18
Winget upgrade allows updating the same package twice


ewanm89:
### Brief description of your issue

If you list the same package twice on winget upgrade, it attempts to install the package twice.

### Steps to reproduce

Have a package in need of updating then run winget upgrade <packageid> <packageid> with the same packageid twice.
![image](https://github.com/microsoft/winget-cli/assets/135458/7886b115-2e25-488d-b20e-e0136bf18e2f)

### Expected behavior

Ignore the accidental repeat and automatically not download and run the installer twice.

### Actual behavior

Runs the installer for the package twice.

### Environment

```shell
Windows Package Manager v1.7.10661

Windows: Windows.Desktop v10.0.19045.4123
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.22.10661.0
```

Trenly:
Can you share the output of `winget list "ShareX"`?

PR #4286, opened 2024-03-19
Remove duplicates from MultiQuery

Trenly:
  - Resolves https://github.com/microsoft/winget-cli/issues/4284
-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4286)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionArgs.h
+++ b/src/AppInstallerCLICore/ExecutionArgs.h
@@ -225,7 +225,7 @@ namespace AppInstaller::CLI::Execution
         {
             auto itr = m_parsedArgs.find(Type::MultiQuery);
             
-            // If there is not a value in MultiQuery, or there is only one value, it is already presumed to be unique
+            // If there is not a value in MultiQuery, or there is only one value, it is presumed to be unique
             if (itr == m_parsedArgs.end() || itr->second.size() == 1)
             {
                 return;
--- a/src/AppInstallerCLICore/ExecutionArgs.h
+++ b/src/AppInstallerCLICore/ExecutionArgs.h
@@ -232,10 +232,9 @@ namespace AppInstaller::CLI::Execution
             }
 
             std::set<std::string> querySet;
-            std::vector<std::string> queryStrings = itr->second;
+            std::vector<std::string>& queryStrings = itr->second;
 
             queryStrings.erase(std::remove_if(queryStrings.begin(), queryStrings.end(), [&](const std::string value) { return !querySet.insert(value).second; }), queryStrings.end());
-            m_parsedArgs[Type::MultiQuery].assign(queryStrings.begin(), queryStrings.end());
         }
 
         // If we get a single value for multi-query, we remove the argument and add it back as a single query.
--- a/src/AppInstallerCLICore/ExecutionArgs.h
+++ b/src/AppInstallerCLICore/ExecutionArgs.h
@@ -4,6 +4,7 @@
 #include <string>
 #include <string_view>
 #include <map>
+#include <unordered_set>
 #include <vector>
 
 namespace AppInstaller::CLI::Execution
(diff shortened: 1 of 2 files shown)


src/AppInstallerCLICore/ExecutionArgs.h
```diff
@@ -218,10 +219,30 @@ namespace AppInstaller::CLI::Execution
...
+            {
+                return;
+            }
+
+            std::unordered_set<std::string> queryStrings;
+            for (auto query : itr->second)
+            {
+                queryStrings.insert(query);
+            }
+            m_parsedArgs[Type::MultiQuery].assign(queryStrings.begin(), queryStrings.end());
```
yao-msft, line 240: This has possibility to change the execution order, but I guess it's ok?

Trenly: I specifically used an unordered set to hopefully avoid that; If there needs to be a strong preservation of order, then I suppose it might be possible to check whether or not the add to the set was successful, and if it was, emplace it into a new vector, then assign the new vector as the value of the arg.  However, I felt this was a cleaner way of doing it.

If it's okay with the team, I'd like to move forward with this approach first, and if ordering becomes an issue, it can be addressed as part of a separate issue

yao-msft: I'm personally ok with this approach. Just wanted to ask about your thought. btw, set will almost always break the order. If we need to preserve order, we'll have to create a new vector and dedupe ourselves (we can dedupe case insensitively too).

yao-msft (dismissed):

src/AppInstallerCLICore/ExecutionArgs.h
Trenly: I think it will be best to let this one as-is, and I'll do a follow up to preserve order by doing the new vector

Trenly:
I'm hoping this is a transient issue not related to the change?

```
  Failed ValidateWinGetDscResource_SourceOpenFailed [74 ms]
  Error Message:
     Expected: True
  But was:  False
  Stack Trace:
     at AppInstallerCLIE2ETests.ConfigureValidateCommand.ValidateWinGetDscResource_SourceOpenFailed() in D:\a\1\s\src\AppInstallerCLIE2ETests\ConfigureValidateCommand.cs:line 276
  Standard Output Messages:
 Starting command run. AICLI path: wingetdev.exe Command: configure validate Parameters: D:\a\1\s\src\x86\Release\AppInstallerCLIE2ETests\TestData\Configuration\WinGetDscResourceValidate_SourceOpenFailed.yml Timeout: 120000
 Command run completed with exit code: 1
```

yao-msft:
> I'm hoping this is a transient issue not related to the change?
> 
> ```
>   Failed ValidateWinGetDscResource_SourceOpenFailed [74 ms]
>   Error Message:
>      Expected: True
>   But was:  False
>   Stack Trace:
>      at AppInstallerCLIE2ETests.ConfigureValidateCommand.ValidateWinGetDscResource_SourceOpenFailed() in D:\a\1\s\src\AppInstallerCLIE2ETests\ConfigureValidateCommand.cs:line 276
>   Standard Output Messages:
>  Starting command run. AICLI path: wingetdev.exe Command: configure validate Parameters: D:\a\1\s\src\x86\Release\AppInstallerCLIE2ETests\TestData\Configuration\WinGetDscResourceValidate_SourceOpenFailed.yml Timeout: 120000
>  Command run completed with exit code: 1
> ```

Seems like so. All other builds are failing too.

JohnMcPMS:
> > I'm hoping this is a transient issue not related to the change?
> > ```
> >   Failed ValidateWinGetDscResource_SourceOpenFailed [74 ms]
> >   Error Message:
> >      Expected: True
> >   But was:  False
> >   Stack Trace:
> >      at AppInstallerCLIE2ETests.ConfigureValidateCommand.ValidateWinGetDscResource_SourceOpenFailed() in D:\a\1\s\src\AppInstallerCLIE2ETests\ConfigureValidateCommand.cs:line 276
> >   Standard Output Messages:
> >  Starting command run. AICLI path: wingetdev.exe Command: configure validate Parameters: D:\a\1\s\src\x86\Release\AppInstallerCLIE2ETests\TestData\Configuration\WinGetDscResourceValidate_SourceOpenFailed.yml Timeout: 120000
> >  Command run completed with exit code: 1
> > ```
> 
> Seems like so. All other builds are failing too.

I'm trying to get to the bottom of it, the failing tests seem to always have an exit code of 1 from winget and don't leave any log file to be collected into the artifacts.  My server coordination PR is likely to have the fix once it is determined.

src/AppInstallerCLICore/ExecutionArgs.h
Trenly: Since the build is failing I took the time to do this now rather than separately

src/AppInstallerCLICore/ExecutionArgs.h
```diff
@@ -218,10 +218,29 @@ namespace AppInstaller::CLI::Execution
...
+            auto itr = m_parsedArgs.find(Type::MultiQuery);
+            
+            // If there is not a value in MultiQuery, or there is only one value, it is presumed to be unique
+            if (itr == m_parsedArgs.end() || itr->second.size() == 1)
+            {
+                return;
+            }
+
+            std::set<std::string> querySet;
+            std::vector<std::string> queryStrings = itr->second;
```
yao-msft, line 235:
```suggestion
            std::vector<std::string>& queryStrings = itr->second;

```

And remove line 238 to avoid unnecessary copies

yao-msft (approved):

PR #4288, opened 2024-03-20
chore: remove repetitive words

vitalmotif:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4288)


vitalmotif:

JohnMcPMS:
These are all in subtrees that we do not want to update directly.  If the the (😈) existence of these repetitive words is really enough of an issue, the upstream repositories should be the target of the initial change.  If they accept the change, then we can pull the update here.

PR #4293, opened 2024-03-21
Fix `EnableWindowsPackageManagerCommandLineInterfaces` Policy Behavior

ryfu-msft:
Credit to @JohnMcPMS for initial implementation and guidance.

There is an issue where if the `EnableWindowsPackageManagerCommandLineInterfaces` policy is disabled, all COM calls would be blocked. This PR fixes that behavior by moving the CLI policy check to `CoreMain` in the CLICore project so that it only performs the check if being executed from the command line and unblocks COM calls.

Verified manually using the updated SampleUWPCaller. Also added E2E tests to check this behavior.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4293)


JohnMcPMS (approved):
I would prefer that we not throw an exception as control flow, but I can live with it.

selimfelex:
Hello  @Madhusudhan-MSFT , @ryfu-msft  could you please advise when this fix will be released, and will it require an update to the Intune management engine ?

makroground:
Deploying Intune Store Apps still failing with error "blocked by policy", if EnableWindowsPackageManagerCommandLineInterfaces policy is set. Tested with v1.7.11261 and also with v1.8.1133 in preview. Any news regarding this still existing issue @JohnMcPMS @ryfu-msft?

yao-msft:
Intune distributes a copy of winget binaries with their client release. The fix will be available in Intune client 2406 release (to be released end of June).

PR #1721, opened 2021-11-17
Updated Valijson to version 0.6.

jedieaston:
- [x] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?
- [ ] Are you working against an Issue?

-----

Per the discussion in #1720, I've moved the Valijson update into a separate PR.

The subtree was updated via a squash from the Valijson repo (as that was the way it was initially added in #697). For time travelers who don't want to spend half an hour reading esoteric Git documentation, the magic command to update is ~~`git pull --allow-unrelated-histories --no-commit --squash -s subtree git@github.com:tristanpenman/valijson.git v0.6` (changing "v0.6" to whatever tag/commit/ref you want).~~ (edit: magic command is actually below)

Tested: Built and ran tests inside Visual Studio 2019. Everything passed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-cli/pull/1721)

PR #4297, opened 2024-03-22
Update cpprestsdk to v2.10.19

Trenly:

-----

This PR updates cpprestsdk to version 2.0.19 which contains some bug fixes. Although I'm not sure if any of these bug fixes affect WinGet, with this being the first release in 2 years and the project being in maintenance mode, I felt it would be okay to update it.

cpprestsdk was updated via ` git subtree pull --squash -P src/cpprestsdk/cpprestsdk https://github.com/microsoft/cpprestsdk v2.10.19`, just like https://github.com/microsoft/winget-cli/pull/1721. There was a merge conflict (since we don't have the tests for cpprestsdk in the subtree in this repo), so I just kept the deleted files deleted.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4297)


yao-msft:
I reviewed the changelist and they do not seem to apply to winget usage now. Since we do not have e2e tests for rest source, we'll do an internal smoke test to take in the change (if we decide to).

The related code for security fix of 127.0.0.1 is not used by winget.
The new feature "get from token endpoint" does not apply to winget yet because winget uses WAM for auth requests.

The change itself seems small and safe though.

Issue #3483, opened 2023-07-28
Make InstallationNotes flow-dependent


Trenly:
### Description of the new feature / enhancement

Allow different messages to be shown on Install, Upgrade, and Uninstall.

### Proposed technical implementation details

To keep manifests clean, I would prefer the field be changed to an object rather than adding new manifests fields, although I suppose either would work. With the 2020-12 schema, it might also be possible to allow for backwards compatibility to prevent this from being a breaking change.

```yaml
InstallationNotes:
	- Install:
    	- Before: This message is shown before a package is installed
    	- After: This message is shown after a package is installed
	- Upgrade:
    	- Before: This message is shown before a package is upgraded
    	- After: This message is shown after a package is upgraded
	- Uninstall:
    	- Before: This message is shown before a package is uninstalled
    	- After: This message is shown after a package is uninstalled
```

stephengillie:
I'm assuming this would be displayed as console messages in the same window as WinGet was launched. This could be a good way to give interactive and other installer/uninstaller instructions. Possibly even give reboot notice.

Trenly:
> I'm assuming this would be displayed as console messages in the same window as WinGet was launched. This could be a good way to give interactive and other installer/uninstaller instructions. Possibly even give reboot notice.

Yes, basically the same way InstallationNotes work currently, but expanding the functionality

denelon:
The "pre" activity notes seem like they make sense in terms of a user at the CLI expecting a prompt.

In the context of a WinGet Configuration the mode would logically be "--disable-interactivity". We would expect WinGet to fail in those cases rather than expecting a prompt (unless "--force" was specified).

Users could still pass "--force" to tell WinGet they know what they are doing and it would essentially override "--disable-interactivity".

These notes are in reply to work being done to improve the "Deny" behavior for upgrade.

Several packages have different variations of reasons the upgrade behavior should be conditional:

**Discord.Discord** - Has an auto-updater built into the application, and doesn't update the version displayed in ARP. Installing a newer version erases user settings
> [!NOTE]
> This has been fixed by Discord.

**Anaconda.Miniconda3** - Has a special upgrade program "Conda" for upgrading. The only way to upgrade through winget is to uninstallPrevious, which erases user settings

**Chocolatey.Chocolatey** - Installer is "Install-Only", and the application self updates

**MSYS2.MSYS2** - Has a default install location and cannot be installed over,
[...]

PR #4300, opened 2024-03-23
Improve behavior of 'UpgradeBehavior: deny'

### LINKED ISSUE #4298: `upgrade --all` will execute update flow for denied packages

When a package has `UpgradeBehavior: deny` set, I expect `winget upgrade <Package>` and `winget upgrade --all` to block updating the package. Turns out `winget upgrade --all` still tries to update the execute flow.

### Steps to reproduce

1. Have a denied package that has an upgrade available (Take MSYS2.MSYS2 for example)
2. Run `winget upgrade MSYS2.MSYS2`
3. Package is blocked correctly
4. Run `winget upgrade --all`
5. winget tries to upgrade MSYS2.MSYS2

image

image

### Expected behavior

Package is blocked from upgrade

### Actual behavior

WinGet tries to upgrade the package

### Environment

```shell
Windows Package Manager v1.7.10661

Windows: Windows.Desktop v10.0.22631.3296
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.22.10661.0

Winget Directories
-------------------------------------------------------------------------------------------------------------------------------
Logs                               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir
User Settings                      %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\settings.json
Portable Package Root (User)       %LOCALAPPDATA%\Microsoft\WinGet\Packages
Portable Package Root              C:\Program Files\WinGet\Packages
Installer Downloads                %USERPROFILE%\Downloads

Links

InstallerHashOverride                     Enabled
```

### LINKED ISSUE #4299: Denied packages should be hidden from `winget upgrade` list

Packages that have `UpgradeBehavior: deny` set should be hidden from `winget upgrade` / `winget list --upgrade-available` list. We can have an argument `--include-denied` to show these packages. The current output for pinned, unknown and explicit targeted packages can be extended to show number of denied packages.

Current sample output
```
The following packages have an upgrade available, but require explicit targeting for upgrade:
Name                         Id              Version      Available            Source
-------------------------------------------------------------------------------------
Discord                      Discord.Discord 1.0.9016     1.0.9035             winget
Microsoft Teams              Microsoft.Teams 1.6.00.29964 23320.3027.2591.1505 winget
Teams Machine-Wide Installer Microsoft.Teams 1.5.0.8070   23320.3027.2591.1505 winget
1 package(s) have version numbers that cannot be determined. Use --include-unknown to see all results.
21 package(s) have pins that prevent upgrade. Use the 'winget pin' command to view and edit pins. Using the --include-pinned argument may show more results.
```

New sample output line
```
1 package(s) do not support upgrades through winget. Use --include-denied to see more results.
```

_No response_

Trenly:
  - Resolves #4298 
  - Resolves #4299 
-----

This PR extends the behavior of `UpgradeBehavior: deny` to the PackageTrackingCatalog so that packages which cannot be upgraded via winget are not included with `winget upgrade --all`. However, the package is still checked for whether or not an upgrade is available. If an upgrade is available, the user will be informed of the upgrade through a new table added at the end of the upgrade list. If the package is up to date, the table will not be shown.

```
PS D:\Git\winget-pkgs> wingetdev upgrade
Name           Id                   Version Available Source
------------------------------------------------------------
GitHub Desktop GitHub.GitHubDesktop 3.3.11  3.3.12    winget
1 upgrades available.

1 package(s) cannot be upgraded using winget. Please use the method provided by the publisher for upgrading these packages.
Name    Id              Version  Available Source
-------------------------------------------------
Discord Discord.Discord 1.0.9034 1.0.9035  winget

6 package(s) have pins that prevent upgrade. Use the 'winget pin' command to view and edit pins. Using the --include-pinned argument may show more results.
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4300)

### DIFF
--- a/src/AppInstallerCLICore/Resources.h
+++ b/src/AppInstallerCLICore/Resources.h
@@ -601,6 +601,7 @@ namespace AppInstaller::CLI::Resource
         WINGET_DEFINE_RESOURCE_STRINGID(UpdateNotApplicableReason);
         WINGET_DEFINE_RESOURCE_STRINGID(UpgradeArgumentDescription);
         WINGET_DEFINE_RESOURCE_STRINGID(UpgradeAvailableForPinned);
+        WINGET_DEFINE_RESOURCE_STRINGID(UpgradeBehaviorDenyCount);
         WINGET_DEFINE_RESOURCE_STRINGID(UpgradeBlockedByManifest);
         WINGET_DEFINE_RESOURCE_STRINGID(UpgradeBlockedByPinCount);
         WINGET_DEFINE_RESOURCE_STRINGID(UpgradeCommandLongDescription);
--- a/src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/UpdateFlow.cpp
@@ -211,6 +211,7 @@ namespace AppInstaller::CLI::Workflow
         bool updateAllFoundUpdate = false;
         int packagesWithUnknownVersionSkipped = 0;
         int packagesThatRequireExplicitSkipped = 0;
+        int packagesWithUpgradeBehaviorDeny = 0;
 
         for (const auto& match : matches)
         {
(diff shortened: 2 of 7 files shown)


yao-msft:
I believe UpgradeBehavior is the behavior when an older version is upgrading to this version. It's not meant for upgrading from this version. We'll only know upgrade is denied by the manifest from the new version. We may need another way to fix this issue.

Trenly:
> I believe UpgradeBehavior is the behavior when an older version is upgrading to this version. It's not meant for upgrading from this version. We'll only know upgrade is denied by the manifest from the new version. We may need another way to fix this issue.

While it is true that UpgradeBehavior is when an older version is upgrading (same with RequireExplicitUpgrade), the `deny` behavior is more similar to `RequireExplicitUpgrade` in that when a user installs a package with the behavior of `deny`, IMO it shouldn't be upgraded even when a new version is available. This is because (typically) upgrade behavior doesn’t change between versions, and that the `deny` behavior indicates publishers have a special way the program should be updated.

I've tested locally and this change does catch the use cases where the user installs a version that has `deny`, it then excludes it from future upgrades.

Regardless - in the current flow, if the upgrade for the package is called explicitly, the selected installer is checked for `UpgradeBehavior: deny` before it is installed, which should cover most edge cases this would miss. 

Without adding details on the UpdateBehavior to the index, this seemed the most appropriate route to me

yao-msft:
I understand this approach is the easiest at the moment and it probably works for most of the cases. But this approach is misusing the UpgradeBehavior deny concept. Once it's written in the tracking catalog, those packages will not be able to upgrade automatically if a new version removes the UpgradeBehavior deny in the future, unless user deletes the tracking catalog.

I think for improving user experience, we can show the package as skipped instead of failure when performing upgrading all. For code logic correctness, we have to check the latest version for UpgradeBehavior.

For winget upgrade list, we should filter out by latest selected installer Upgradebehavior (which may be heavy).
Or
A new manifest field like RequiresExplicitUpgrade for indicating deny all upgrading in the future (but it's still weird, we can not predict in the future that a new version which supports upgrade will never happen).
Or
As a workaround fix, for those packages that have UpgradeBehavior deny in the latest version, set RequiresExplicitUpgrade on all current manifests. This solves winget upgrade --all too,

Trenly:
> Once it's written in the tracking catalog, those packages will not be able to upgrade automatically if a new version removes the UpgradeBehavior deny in the future, unless user deletes the tracking catalog

I would argue that is the way it should be, especially for deny, since `Deny` is set when there are known issues with upgrading the package through WinGet

yao-msft:
Yes, RequireExplicitUpgrade will require explicit upgrade forever. I think it's an attribute for describing the installed version (the installed version attribute will not change regardless of future available version, it's meant for this installed version only). The UpgradeBehavior is an attribute for describing how to operate on incoming available version (which may change in future versions). Our tracking catalog is for recording installed version.

Implementation wise, I don't feel we'd mix the installed version attribute with available version attribute in the manifest. So I suggested a new manifest attribute just to describe "block upgrade from this installed version" if we want to. Unless if the team agrees with this mixed usage, writing to the tracking catalog would be the way to go.

Btw, this is a side question. RequireExplicitUpgrade is like pinning, it is soft. Deny is blocking, it's strong. Do we want manifest author to control blocking upgrade from an installed version and we write to index without user input?

denelon:
@Trenly and I were discussing some of the behavior changes over at:
* https://github.com/microsoft/winget-cli/issues/3483

denelon:
Maybe it makes sense to look at adding the additional notes fields to the manifest and then we can consider altering some of the behaviors in the client.

PR #4349, opened 2024-04-04
Add support for optional limitation set in configuration remote server and processors

yao-msft:
Added support for optional limitation set in configuration remote server and processors. When running in limitation mode, incoming units would need to match units in limitation set. And units can only be asserted or applied once.

Added tests. And some hacky manual verification. E2E tests will come along when calling side is updated in separate pr.

Also updated .net target OS version to 22000 to match other parts. (I thought there was a bug in IInputStreamAdaptor for MemoryStream)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4349)

### DIFF
--- a/src/ConfigurationRemotingServer/Program.cs
+++ b/src/ConfigurationRemotingServer/Program.cs
@@ -18,7 +18,6 @@ namespace ConfigurationRemotingServer
         private const string CommandLineSectionSeparator = "~~~~~~";
         private const string ExternalModulesName = "ExternalModules";
 
-
         static int Main(string[] args)
         {
             ulong memoryHandle = ulong.Parse(args[0]);
@@ -33,12 +32,13 @@ namespace ConfigurationRemotingServer
 
                 // Set default properties.
                 var externalModulesPath = GetExternalModulesPath();
-                if (!string.IsNullOrWhiteSpace(externalModulesPath))
+                if (string.IsNullOrWhiteSpace(externalModulesPath))
                 {
-                    // Set as implicit module paths so it will be always included in AdditionalModulePaths
-                    factory.ImplicitModulePaths = new List<string>() { externalModulesPath };
+                    throw new DirectoryNotFoundException("Failed to get ExternalModules.");
                 }
 
+                // Set as implicit module paths so it will be always included in AdditionalModulePaths
+                factory.ImplicitModulePaths = new List<string>() { externalModulesPath };
                 factory.ProcessorType = PowerShellConfigurationProcessorType.Hosted;
 
                 // Parse limitation set if applicable.
--- a/src/ConfigurationRemotingServer/Program.cs
+++ b/src/ConfigurationRemotingServer/Program.cs
@@ -1,14 +1,21 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+using Microsoft.Management.Configuration;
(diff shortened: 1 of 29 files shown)


src/ConfigurationRemotingServer/Program.cs
```diff
@@ -21,6 +31,73 @@ static int Main(string[] args)
 
                 PowerShellConfigurationSetProcessorFactory factory = new PowerShellConfigurationSetProcessorFactory();
 
+                // Set default properties.
+                var externalModulesPath = GetExternalModulesPath();
+                if (!string.IsNullOrWhiteSpace(externalModulesPath))
```
JohnMcPMS, line 36: This case should probably result in an error rather than silently skipping the external modules path.

src/Microsoft.Management.Configuration.Processor/Extensions/ValueSetExtensions.cs
```diff
@@ -90,5 +89,92 @@ public static IList<object> ToArray(this ValueSet valueSet)
...
+
+                var firstValue = keyValuePair.Value;
+                var secondValue = second[keyValuePair.Key];
+
+                // Empty value check.
+                if (firstValue == null && secondValue == null)
+                {
+                    continue;
+                }
+                else if ((firstValue == null && secondValue != null) ||
```
JohnMcPMS, line 121: nit: could be simplified to:
```
else if (firstValue == null || secondValue == null)
```
since the first condition above already filters out the `&&` case.

src/Microsoft.Management.Configuration.Processor/Set/ConfigurationSetProcessor.cs
```diff
@@ -24,16 +25,34 @@ namespace Microsoft.Management.Configuration.Processor.Set
...
-        public ConfigurationSetProcessor(IProcessorEnvironment processorEnvironment, ConfigurationSet? configurationSet)
+        /// <param name="isLimitMode">Whether the set processor should work in limitation mode.</param>
+        public ConfigurationSetProcessor(IProcessorEnvironment processorEnvironment, ConfigurationSet? configurationSet, bool isLimitMode = false)
         {
             this.ProcessorEnvironment = processorEnvironment;
             this.configurationSet = configurationSet;
+            this.isLimitMode = isLimitMode;
+
+            // In limit mode, configurationSet is the limitation set to be used. It cannot be null.
+            if (this.isLimitMode)
```
JohnMcPMS, line 44: Should probably use the property rather than the field.

src/Microsoft.Management.Configuration.Processor/Set/ConfigurationSetProcessor.cs
```diff
@@ -343,5 +406,41 @@ private void OnDiagnostics(DiagnosticLevel level, string message)
...
+            {
+                if (this.configurationSet == null)
+                {
+                    throw new InvalidOperationException("Configuration set should not be null in limit mode.");
+                }
+
+                var unitList = useLimitList ? this.limitUnitList : this.configurationSet.Units;
+
+                for (int i = unitList.Count - 1; i >= 0; i--)
```
JohnMcPMS, line 422: Why are you walking the list backward?

JohnMcPMS (approved):

PR #4462, opened 2024-05-08
Index V2 consumption

JohnMcPMS:
## Change
This change adds the ability to actually use the V2 index database.  This primarily required fixing the ignored search functionality from the previous PR and updating the `ISource` implementation to be aware of V2 and how to retrieve the package version manifest files.  Additionally, a few places in the code were updated to avoid requesting all of the package versions when we could instead use only the latest version.  This prevents the need to retrieve the package version file at all, which is the goal for all read-only operations (`search`, `list`, `update` [aka `list` but filtered to things with updates]).  In reality, we do need to get the package version files when we need to do ARP version mapping, but that constitutes only 5% of the total packages in `winget-pkgs` currently.

While the V2 usage code is always enabled, an experimental feature (`indexV2`) is required to attempt to retrieve the V2 index from the CDN.  Disabling this feature should naturally transition one back to V1 on the next update attempt.

File caching, based on the known SHA256 hash of the file, is added as a general utility.  This is then used for V1 manifest files and also for V2 package version data and manifest files.  While the files are generally small, it is not uncommon for them to be used a few times over a short period.

Thanks to @msftrubengu for pointing out that I forgot the include the exports in the .def file in the previous PR. 🥇 

## Validation
Added many new unit tests and updated many existing unit tests.
Updated E2E tests to generate a V2 index in addition to the V1, and added a test pass that runs all of the existing E2E tests using the V2 index.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4462)


src/AppInstallerCLITests/SQLiteIndex.cpp
```diff
@@ -1873,6 +1932,11 @@ TEST_CASE("SQLiteIndex_Search_VersionSorting", "[sqliteindex]")
 
     TestPrepareForRead(index);
 
+    if (!AreChannelsSupported(index))
+    {
+        return;
+    }
```
Trenly, line 1938: You repeat this a few times. Would a `RETURNIF(condition , [return value])` macro be useful to reduce code at all?

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -149,7 +150,7 @@ namespace AppInstaller::Settings
         SETTINGMAPPING_SPECIALIZATION(Setting::ProgressBarVisualStyle, std::string, VisualStyle, VisualStyle::Accent, ".visual.progressBar"sv);
         SETTINGMAPPING_SPECIALIZATION(Setting::AnonymizePathForDisplay, bool, bool, true, ".visual.anonymizeDisplayedPaths"sv);
         // Source
-        SETTINGMAPPING_SPECIALIZATION_POLICY(Setting::AutoUpdateTimeInMinutes, uint32_t, std::chrono::minutes, 5min, ".source.autoUpdateIntervalInMinutes"sv, ValuePolicy::SourceAutoUpdateIntervalInMinutes);
+        SETTINGMAPPING_SPECIALIZATION_POLICY(Setting::AutoUpdateTimeInMinutes, uint32_t, std::chrono::minutes, 15min, ".source.autoUpdateIntervalInMinutes"sv, ValuePolicy::SourceAutoUpdateIntervalInMinutes);
```
Trenly, line 153: Not a comment about the code, just a reminder for @denelon that this change will require a docs update - https://learn.microsoft.com/en-us/windows/package-manager/winget/settings#autoupdateintervalinminutes

doc/Settings.md
```diff
@@ -355,3 +355,20 @@ You can enable the feature as shown below.
        "configureExport": true
    },
 ```
+
+### indexV2
+
+This feature enables the `winget` source to retrieve the V2 index, which is significantly smaller.
+Regardless of the state of this feature, if the index on the machine contains a V2 index, it will be used.
```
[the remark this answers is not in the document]
JohnMcPMS: Correct.

src/AppInstallerCLIE2ETests/Helpers/WinGetSettingsHelper.cs
```diff
@@ -33,197 +38,226 @@ public static string GetUserSettingsPath()
...
+            Hashtable experimentalFeatures =
+                    new Hashtable()
+                    {
+                        { "experimentalArg", false },
+                        { "experimentalCmd", false },
+                        { "directMSI", false },
+                        { "windowsFeature", false },
+                        { "resume", false },
+                        { "reboot", false },
+                    };
```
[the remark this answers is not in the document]
JohnMcPMS: I agree that an empty table is currently equivalent, but that also means that not updating it is equivalent.  I'm not trying to change that right now.

src/AppInstallerCLITests/SQLiteIndex.cpp
JohnMcPMS: I don't feel like that is worth the additional cognitive load of knowing what it does to save the lines.

src/AppInstallerCLIE2ETests/Helpers/WinGetSettingsHelper.cs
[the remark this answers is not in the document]
JohnMcPMS: But since I was looking at the file and then touched it, removed them.

src/AppInstallerCommonCore/PackageVersionDataManifest.cpp
```diff
@@ -70,6 +72,14 @@ namespace AppInstaller::Manifest
         return "versionData.mszyml"sv;
     }
 
+    std::filesystem::path PackageVersionDataManifest::GetRelativeDirectoryPath(std::string_view packageIdentifier, std::string_view manifestHash)
+    {
+        std::filesystem::path result = "packages";
```
msftrubengu, line 77: static constexpr std::string_view?

src/AppInstallerCommonCore/ExperimentalFeature.cpp
```diff
@@ -52,6 +52,8 @@ namespace AppInstaller::Settings
                 return userSettings.Get<Setting::EFConfigureSelfElevation>();
             case ExperimentalFeature::Feature::StoreDownload:
                 return userSettings.Get<Setting::EFStoreDownload>();
+            case ExperimentalFeature::Feature::IndexV2:
```
msftrubengu, line 55: _triggers OCD_

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.cpp
```diff
@@ -31,375 +31,894 @@ namespace AppInstaller::Repository::Microsoft
             std::weak_ptr<SQLiteIndexSource> m_source;
         };
 
-        // The IPackageVersion impl for SQLiteIndexSource.
-        struct PackageVersion : public SourceReference, public IPackageVersion
+        namespace V1
```
msftrubengu, line 34: I'm sure this is easier, but it might be worthy to have them in separate files for maintainability.

msftrubengu (approved):

src/AppInstallerCommonCore/PackageVersionDataManifest.cpp
JohnMcPMS: If it was to be used elsewhere, but this function is meant to prevent that need.

src/AppInstallerCommonCore/ExperimentalFeature.cpp
JohnMcPMS: I'm going to assume that my lazy "Accept Both Changes" click in Code that put my case first is the issue here.  🦪

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.h
```diff
@@ -79,4 +81,19 @@ namespace AppInstaller::Repository::Microsoft
         // Removes a package version from the source.
         void RemovePackageVersion(const Manifest::Manifest& manifest, const std::filesystem::path& relativePath);
     };
+
+    namespace details
+    {
+        // The base for the package objects.
```
yao-msft, line 87: nit: seems like wrong description

yao-msft (approved):

doc/Settings.md
```diff
@@ -355,3 +355,20 @@ You can enable the feature as shown below.
...
+If there is a bug with the V2 index stopping the `winget` CLI from working, disable the feature in your settings file and run this command:
+```
+> winget uninstall -s msstore Microsoft.Winget.Source_8wekyb3d8bbwe
```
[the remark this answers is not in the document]
JohnMcPMS: No, that was an intentional "if the winget source is broken, lets avoid using it" flag.  Commands targeting the local data don't really need a source, but we don't have a way to use a null one.

src/AppInstallerRepositoryCore/Microsoft/SQLiteIndexSource.h
JohnMcPMS: No, it isn't wrong, but it probably made more sense before I refactored it away from the things that inherit it.

yao-msft (approved):

PR #4487, opened 2024-05-15
Add tests for configure mixed elevation

ryfu-msft:
Adds a test that runs two configuration with different elevation requirements using the dynamic runtime factory through COM. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4487)


IlCapitanoLoco (approved):

IlCapitanoLoco (approved):

JohnMcPMS (approved):

src/Microsoft.Management.Configuration.UnitTests/Tests/ConfigurationMixedElevationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+        /// <param name="log">Log helper.</param>
+        public ConfigurationMixedElevationTests(UnitTestFixture fixture, ITestOutputHelper log)
+            : base(fixture, log)
+        {
+            this.fixture = fixture;
+            this.log = log;
+        }
+
+        /// <summary>
+        /// Verifies that creating a high integrity unit processor. Must run elevated.
```
JohnMcPMS, line 41: I don't want a test that will actually request admin.  The basic test should still use the test mode, with two units (one high, one medium) and a special resource that writes the PID into a target file so that you can ensure that they were routed to different set processors.

src/Microsoft.Management.Configuration.UnitTests/Tests/ConfigurationMixedElevationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    using Microsoft.VisualBasic;
+    using Xunit;
+    using Xunit.Abstractions;
+
+    /// <summary>
+    /// Unit tests for verifying the processor behavior for handling mixed elevation scenarios.
+    /// </summary>
+    [Collection("UnitTestCollection")]
+    [OutOfProc]
+    [OutOfProcOnly]
```
JohnMcPMS, line 22: I assume that these tests still attempt to run if you select everything in the Test Explorer?  I'm not sure how to default them to not selected, but that would be better if we can.

src/Microsoft.Management.Configuration.UnitTests/Tests/ConfigurationMixedElevationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+                Assert.False(unitResult.PreviouslyInDesiredState);
+                Assert.False(unitResult.RebootRequired);
+                Assert.NotNull(unitResult.ResultInformation);
+                Assert.Null(unitResult.ResultInformation.ResultCode);
+                Assert.Equal(ConfigurationUnitResultSource.None, unitResult.ResultInformation.ResultSource);
+            }
+        }
+
+        /// <summary>
+        /// Verifies the behavior of running a restricted unit.
```
JohnMcPMS, line 83: I don't know what this test is actually assessing; it seems unnecessary right now.

src/Microsoft.Management.Configuration.UnitTests/Tests/ConfigurationMixedElevationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+                Assert.False(unitResult.PreviouslyInDesiredState);
+                Assert.False(unitResult.RebootRequired);
+                Assert.NotNull(unitResult.ResultInformation);
+                Assert.Null(unitResult.ResultInformation.ResultCode);
+                Assert.Equal(ConfigurationUnitResultSource.None, unitResult.ResultInformation.ResultSource);
+            }
+        }
+
+        /// <summary>
+        /// Verifies that creating a high integrity unit processor for a non elevated unit should return an invalid operation result.
```
JohnMcPMS, line 133: This test should have an elevated unit as well so that we can see one passing and one failing.

src/Microsoft.Management.Configuration.UnitTests/Tests/ConfigurationMixedElevationTests.cs
ryfu-msft: I was not able to find a way to have these skipped in Test Explorer. There is an option to do something like Fact[Skip="reason"] but that causes the test to always be skipped.

yao-msft (approved):

PR #4489, opened 2024-05-17
Fix limit mode metadata blocking

JohnMcPMS:
## Change
The serialization code should not be modifying the object that it is operating on.  Move it to instead skip the output of some values rather than removing them.

When the remoting server is operating in limited mode, the metadata must match between the initial limitation set and the incoming unit.  This removal was also breaking that, as the entry was now missing.

## Validation
Manual confirmation that the limit mode works after the change.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4489)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -345,7 +345,7 @@ jobs:
       testSelector: 'testAssemblies'
       testAssemblyVer2: '**\Microsoft.Management.Configuration.UnitTests.dll'
       searchFolder: '$(buildOutDir)\Microsoft.Management.Configuration.UnitTests'
-      codeCoverageEnabled: true
+      codeCoverageEnabled: false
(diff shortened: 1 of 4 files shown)


yao-msft (approved):
So looks like during serialization "module" will be excluded from metadata, I guess the incoming unit will have "module" dropped in the metadata somewhere else?

JohnMcPMS:
> So looks like during serialization "module" will be excluded from metadata, I guess the incoming unit will have "module" dropped in the metadata somewhere else?

No, `module` gets inserted from the resource name if present.  This was removing it for serialization for brevity, but by actually removing it the limit matching broke.

yao-msft (approved):

PR #4541, opened 2024-06-05
Add support for extracting archives using tar.exe

ryfu-msft:
We have received feedback noting that the shell api fails to extract archives in certain scenarios. This adds an alternative way of extracting archives by invoking the tar.exe executable which is included in version later than 17063.

- Added install behavior setting `extractArchiveWithTar` that takes in a boolean value. If no value is specified or value is not valid, the default extract behavior will utilize the Windows Shell API. 
- Added tests to verify that the archive extraction behavior still works. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4541)


src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
```diff
@@ -538,4 +538,36 @@ namespace AppInstaller::CLI::Workflow
...
         }
     }
-}
\ No newline at end of file
+
+    void ShellExecuteExtractArchive::operator()(Execution::Context& context) const
+    {
+        auto tarExecPath = AppInstaller::Filesystem::GetExpandedPath("%windir%\\system32\\tar.exe");
+
+        std::string args = "-xf " + m_archivePath.u8string() + " -C " + m_destPath.u8string();
```
yao-msft, line 546: Did you try paths with spaces (i.e. do we need to quote the path)?

tar.exe -xf c:\t e s t.zip -C C:\d e s t

doc/Settings.md
```diff
@@ -97,6 +97,15 @@ The 'skipDependencies' behavior affects whether dependencies are installed for a
...
         "skipDependencies": true
     },
+```
+
+### Extract Archive With Tar Executable
+The 'extractArchiveWithTar' behavior affects whether installer archives are extracted using the tar executable ('tar.exe'). Defaults to 'false' if value is not set or is invalid.
+
+```json
+    "installBehavior": {
+        "extractArchiveWithTar": true
```
yao-msft, line 107: I was expecting a string setting like below for extensibility

"archiveExtraction": "tar"  allowed values are (tar, shellApi)

src/AppInstallerCLICore/Workflows/ShellExecuteInstallerHandler.cpp
ryfu-msft: added single quotes to avoid issues with space in path

doc/Settings.md
ryfu-msft: changed to archiveExtractionMethod that takes in string values.

src/AppInstallerCLITests/InstallFlow.cpp
```diff
@@ -514,6 +514,24 @@ TEST_CASE("ExtractInstallerFromArchive_InvalidZip", "[InstallFlow][workflow]")
     REQUIRE(installOutput.str().find(Resource::LocString(Resource::String::ExtractArchiveFailed).get()) != std::string::npos);
 }
 
+TEST_CASE("ExtractInstallerFromArchiveWithTar_InvalidZip", "[InstallFlow][workflow]")
```
JohnMcPMS, line 517: Please have a success case in the unit tests.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -290,6 +290,23 @@ namespace AppInstaller::Settings
...
+            if (Utility::CaseInsensitiveEquals(value, s_archiveExtractionMethod_tar))
+            {
+                return Archive::ExtractionMethod::Tar;
+            }
+            else if (Utility::CaseInsensitiveEquals(value, s_archiveExtractionMethod_shellApi))
+            {
+                return Archive::ExtractionMethod::ShellApi;
+            }
+
+            return {};
```
yao-msft, line 307: what is "return {}" for enum return values? I guess we could add Archive::ExtractionMethod::Unknown and return the unknown enum

PR #4554, opened 2024-06-14
Trim Version When Parsing

Trenly:

-----

When parsing a version, it may be possible that a publisher included one or more whitespace characters at the end of the DisplayVersion key. Since a folder name cannot end with a whitespace character, this means that any PackageVersion in the community repository cannot have a trailing whitespace character.

Due to the way versions are parsed, the trailing whitespace gets considered as an "other" of the version part during the version comparison. This means that `5.8 ` is considered a lower version than `5.8` when parsed. 

This PR makes it so that when a version is assigned, as is the case when parsing the installed versions, both the final version and the version parts are trimmed of whitespace. This means that `5.0.13` is now considered equal to `5.0 .13`, `5.0.13 `, or even ` 5 .   0   .13 `
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4554)

### DIFF
--- a/src/AppInstallerSharedLib/Versions.cpp
+++ b/src/AppInstallerSharedLib/Versions.cpp
@@ -16,7 +16,7 @@ namespace AppInstaller::Utility
 
     Version::Version(std::string&& version, std::string_view splitChars)
     {
-        Assign(std::move(Utility::Trim(version)), splitChars);
+        Assign(std::move(version), splitChars);
     }
 
     Version::Version(Version baseVersion, ApproximateComparator approximateComparator) : Version(std::move(baseVersion))
@@ -438,12 +438,12 @@ namespace AppInstaller::Utility
 
     UInt64Version::UInt64Version(std::string&& version, std::string_view splitChars)
     {
-        Assign(std::move(Utility::Trim(version)), splitChars);
+        Assign(std::move(version), splitChars);
     }
 
     void UInt64Version::Assign(std::string version, std::string_view splitChars)
     {
-        Version::Assign(std::move(Utility::Trim(version)), splitChars);
+        Version::Assign(std::move(version), splitChars);
 
         // After trimming trailing parts (0 or empty),
         // at most 4 parts must be present
@@ -460,7 +460,7 @@ namespace AppInstaller::Utility
 
     SemanticVersion::SemanticVersion(std::string&& version)
     {
-        Assign(std::move(Utility::Trim(version)), DefaultSplitChars);
+        Assign(std::move(version), DefaultSplitChars);
     }
 
     void SemanticVersion::Assign(std::string version, std::string_view splitChars)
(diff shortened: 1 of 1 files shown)


src/AppInstallerSharedLib/Versions.cpp
```diff
@@ -16,7 +16,7 @@ namespace AppInstaller::Utility
 
     Version::Version(std::string&& version, std::string_view splitChars)
     {
-        Assign(std::move(version), splitChars);
+        Assign(std::move(Utility::Trim(version)), splitChars);
```
JohnMcPMS, line 19: You have `Assign` trimming; no need to do it multiple times.

src/AppInstallerSharedLib/Versions.cpp
```diff
@@ -437,12 +438,12 @@ namespace AppInstaller::Utility
 
     UInt64Version::UInt64Version(std::string&& version, std::string_view splitChars)
     {
-        Assign(std::move(version), splitChars);
+        Assign(std::move(Utility::Trim(version)), splitChars);
```
JohnMcPMS, line 441: Let `Assign` trim.  May require updating the custom `Assign` for this type.

src/AppInstallerSharedLib/Versions.cpp
```diff
@@ -437,12 +438,12 @@ namespace AppInstaller::Utility
...
     UInt64Version::UInt64Version(std::string&& version, std::string_view splitChars)
     {
-        Assign(std::move(version), splitChars);
+        Assign(std::move(Utility::Trim(version)), splitChars);
     }
 
     void UInt64Version::Assign(std::string version, std::string_view splitChars)
     {
-        Version::Assign(std::move(version), splitChars);
+        Version::Assign(std::move(Utility::Trim(version)), splitChars);
```
JohnMcPMS, line 446: Let `Assign` trim.

src/AppInstallerSharedLib/Versions.cpp
```diff
@@ -459,7 +460,7 @@ namespace AppInstaller::Utility
 
     SemanticVersion::SemanticVersion(std::string&& version)
     {
-        Assign(std::move(version), DefaultSplitChars);
+        Assign(std::move(Utility::Trim(version)), DefaultSplitChars);
```
JohnMcPMS, line 463: Let `Assign` trim.

src/AppInstallerSharedLib/Versions.cpp
Trenly: Thanks for the guidance! After reviewing the custom `Assign`, the method that takes in a string goes back to `Version::Assign`, so I'll let that handle the trimming, and the other method only takes in integers and shouldn't require trimming.

JohnMcPMS (approved):

Issue #2394, opened 2022-07-29
Improve version parsing to better support "preambles"


JohnMcPMS:
### Description of the new feature / enhancement

The version parsing could be improved to ignore leading, non-numeric values that are then followed by a more normal version number.  This would apply to versions like `v1.2.3`, as well as the package detailed in microsoft/winget-pkgs#67794

For example `ad 7.0.13` would equal `7.0.13` and `v1.2.3` would equal `1.2.3`.
`foo1` would equal `bar1`, but `1.foo2` would not equal `1.bar2`, as only leading non-numeric characters would be ignored.

### Proposed technical implementation details

For the first version part (where parts are split by the `.` character), determine if there is a number at the end (currently only decimal integers are supported) and use it as the integer portion.  Do not use the "preamble" in the version at all.

Trenly:
Related to #1298

luzpaz:
What's the tipping point for this issue to be solved ?

Pandoriaantje:
> What's the tipping point for this issue to be solved ?

There seems to be no mechanism to handle beta/release candidate, etc. version numbering, which leaves the users at the hands of however winget interprets the version number. This could lead to false 'upgrades'.

denelon:
That's correct, we still have a big chunk of work left in order to handle channels.
* #147

Trenly:
Resolved by #4558

PR #4558, opened 2024-06-14
Preambles

Trenly:
  - #2394 
 
-----


src/AppInstallerCLITests/Versions.cpp
```diff
@@ -134,9 +155,18 @@ TEST_CASE("VersionCompare", "[versions]")
...
     // Ensure whitespace doesn't affect equality
     RequireEqual("1.0", "1.0 ");
     RequireEqual("1.0", "1. 0");
+
+    // Ensure versions with preambles are sorted correctly
+    RequireEqual("1.0", "Version 1.0");
+    RequireEqual("foo1", "bar1");
+    RequireLessThan("v0.0.1", "0.0.2");
+    RequireLessThan("v0.0.1", "v0.0.2");
+    RequireLessThan("1.a1", "1.b2");
```
JohnMcPMS, line 168: This line is confusing; if the code removed preambles on all parts, it would still be expected to be true.

src/AppInstallerSharedLib/Public/AppInstallerStrings.h
```diff
@@ -10,6 +10,10 @@
 
 namespace AppInstaller::Utility
 {
+    // Same as std::isspace(char)
+#define AICLI_SPACE_CHARS " \f\n\r\t\v"sv
```
JohnMcPMS, line 14:
```suggestion
static constexpr s_AICLI_SPACE_CHARS " \f\n\r\t\v"sv;
```

src/AppInstallerSharedLib/Versions.cpp
```diff
@@ -64,7 +64,20 @@ namespace AppInstaller::Utility
             size_t newPos = baseVersion.find_first_of(splitChars, pos);
 
             size_t length = (newPos == std::string::npos ? baseVersion.length() : newPos) - pos;
-            m_parts.emplace_back(baseVersion.substr(pos, length));
+            std::string interimPart = baseVersion.substr(pos, length);
+
+            // For the first version part, if there is a digit before the split character, trim off all leading non-digit characters
```
JohnMcPMS, line 69: This seems like it would be better done by modifying `baseVersion` before the loop rather than having a conditional that we know will only ever be hit once.

src/AppInstallerSharedLib/Versions.cpp
```diff
@@ -64,7 +64,20 @@ namespace AppInstaller::Utility
...
+            {
+                size_t digitPos = interimPart.find_first_of(AICLI_DIGIT_CHARS);
+                if (digitPos != std::string::npos)
+                {
+                    interimPart.erase(0, digitPos);
+                }
+            }
+            // If the part contains any leading or trailing whitespace, it will be trimmed off as part of the constructor which
+            // emplaces it into the m_parts vector
+            m_parts.emplace_back(interimPart);
```
JohnMcPMS, line 80:
```suggestion
            m_parts.emplace_back(std::move(interimPart));
```

src/AppInstallerSharedLib/Public/AppInstallerStrings.h
Trenly: This define is used for creating the expressions used in the trim functions, which is why I didn't change it and only moved it; 
```
    constexpr std::string_view s_SpaceChars = AICLI_SPACE_CHARS;
    constexpr std::wstring_view s_WideSpaceChars = L"" AICLI_SPACE_CHARS;
```

Trenly:
> Microsoft.Management.Configuration.UnitTests.Tests.ConfigurationMixedElevationTests.ApplyMixedElevationUnits [FAIL]

Is this a transient error?

Trenly:
Interesting . . . Looks like the `HasExtension` function for ManifestVersion uses a Version to store just the extensions, which is quite confusing

src/AppInstallerSharedLib/Public/AppInstallerVersions.h
```diff
@@ -42,15 +42,15 @@ namespace AppInstaller::Utility
 
         Version() = default;
 
-        Version(const std::string& version, std::string_view splitChars = DefaultSplitChars) :
-            Version(std::string(version), splitChars) {}
-        Version(std::string&& version, std::string_view splitChars = DefaultSplitChars);
+        Version(const std::string& version, std::string_view splitChars = DefaultSplitChars, bool trimPrefix = true) :
```
JohnMcPMS, line 45: I think I would lean towards having a protected member field that controls the trim behavior and create a custom untrimmed version type.  That might make it harder to stumble into breaking the expectations somewhere, especially if the untrimmed type uses protected inheritance to prevent external use as a "standard" `Version` (at the cost of manually exposing all required methods via `using`).

JohnMcPMS:
```C++
struct Version {
...
protected:
    bool m_shouldRemovePreambles = true;
...
}

// Name for example
struct PreambleEnabledVersion : protected Version
{
PreambleEnabledVersion() { m_shouldRemovePreambles = false; }
}
```

And if my understanding is correct, then:

```C++
Version a{ "1.0" };
PreambleEnabledVersion b { "1.0" };
// Compilation error as b is not a `Version` outside of its own member functions.
if (a == b) ...
```

All of this is probably over-engineering, but I always try to design against shooting myself in the foot rather than only be careful not to.

Trenly: > All of this is probably over-engineering, but I always try to design against shooting myself in the foot rather than only be careful not to.

I appreciate this; I'd rather do it right the first time than have issues with it later

Trenly: I tried this, and when I use the example as written, I get the message that no overload exists which takes two parameters.  This is coming from the constructor for `PreambleEnabledVersion( std::string&& version, std::string_view splitChars)` not being present.  

I tried creating the constructor, but I think my understanding of how it works is flawed, since it doesn't seem to be properly setting the member field

```
        PreambleEnabledVersion() { m_shouldRemovePreambles = false; }
        
        PreambleEnabledVersion(const std::string& version, std::string_view splitChars = DefaultSplitChars) :
            Version(std::string(version), splitChars) {m_shouldRemovePreambles = false;}
        PreambleEnabledVersion(std::string&& version, std::string_view splitChars = DefaultSplitChars) :
            Version(version, splitChars) {m_shouldRemovePreambles = false;}
```

JohnMcPMS: The constructor if the base type will be run before anything in the derived type.  The easiest solution is to just use `Assign` in the derived constructor, rather than invoking the base constructor.

src/AppInstallerSharedLib/Public/AppInstallerVersions.h
```diff
@@ -119,12 +119,25 @@ namespace AppInstaller::Utility
...
     };
 
+    // Version that does not have leading non-digit characters trimmed
+    struct RawVersion : protected Version
+    {
+        RawVersion() { m_trimPrefix = false; }
+
+        RawVersion(const std::string& version, std::string_view splitChars = DefaultSplitChars) :
+            RawVersion(std::string(version), splitChars) { m_trimPrefix = false; }
+        RawVersion(std::string&& version, std::string_view splitChars = DefaultSplitChars);
```
JohnMcPMS, line 136: nit: just taking `std::string version` if you needed the memory for yourself is the simplest way.  l-values will copy construct (as you are doing manually) and r-values will move construct.

nit2: the `string&` constructor doesn't need to set `m_trimPrefix` again.

Trenly: I see. . .  I was just copying what `Version` had done, so wasn't paying attention to how the values actually get handled in memory. I presume the easiest is just to remove the `string&` constructor?

JohnMcPMS: Remove the `string&` and change `string&&` to `string`.

JohnMcPMS (approved):

PR #4564, opened 2024-06-19
Use digest verification APIs for MSIX when available

JohnMcPMS:
## Change
When available (10.0.23504.0 according to MSDN), use the MSIX digest APIs to verify the package during streaming installation.  When not available, download instead.

## Validation
Manual installation successful, with logs indicating usage of digest.  Manual installation with debugger memory change to invalidate the digest results in an error (although not the best error experience, it should also be extremely rare).

Existing regression tests should get some coverage as well.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4564)

### DIFF
--- a/src/AppInstallerCommonCore/Deployment.cpp
+++ b/src/AppInstallerCommonCore/Deployment.cpp
@@ -66,49 +66,108 @@ namespace AppInstaller::Deployment
 
             return S_OK;
         }
+
+        bool ShouldUseReputationCheck(const Options& options)
+        {
+            return options.ExpectedDigests.empty() && !options.SkipReputationCheck;
+        }
+
+        IAsyncOperationWithProgress<DeploymentResult, DeploymentProgress> StartAddPackage(PackageManager& packageManager, const winrt::Windows::Foundation::Uri& uri, const Options& options)
+        {
+            if (!options.ExpectedDigests.empty())
+            {
+                // Must use API that supports digests
+                THROW_WIN32_IF(ERROR_NOT_SUPPORTED, !IsExpectedDigestsSupported());
+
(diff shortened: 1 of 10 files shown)


src/AppInstallerCommonCore/Deployment.cpp
```diff
@@ -66,49 +66,108 @@ namespace AppInstaller::Deployment
...
+            {
+                return packageManager.AddPackageAsync(
+                    uri,
+                    nullptr, /*dependencyPackageUris*/
+                    DeploymentOptions::None,
+                    nullptr, /*targetVolume*/
+                    nullptr, /*optionalAndRelatedPackageFamilyNames*/
+                    nullptr, /*optionalPackageUris*/
+                    nullptr /*relatedPackageUris*/);
+            }
```
florelis, line 101: Is there a reason to have these two branches separate? I would find it easier to understand if we had as few different calls to `*AddPackage*Async()`
Only reason I can think of is if `AddPackageByUriAsync()` wasn't available downlevel, but it goes back to 19041 which I think is the oldest supported version

florelis (approved):

src/AppInstallerCommonCore/Deployment.cpp
JohnMcPMS: I don't want to remove it until we change the minimum version for the shipping package:
```
<TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.17763.0" MaxVersionTested="10.0.22621.0"/>
```

PR #4567, opened 2024-06-20
Launch COM server for elevated use via packaged path

JohnMcPMS:
## Change
Leverage the feature that allows launching a packaged process via its packaged executable location as long as there is a matching alias.  This better ensures that we are launching the correct process.

## Validation
While the dev usage can't use this path due to the way it and development deployed packages work, the code to find the executable is being run to ensure correct functionality.  It is then replaced with the alias path that was being used previously.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4567)

### DIFF
--- a/src/WinGetServer/WinGetServerManualActivation_Client.cpp
+++ b/src/WinGetServer/WinGetServerManualActivation_Client.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 #include "WinGetServer.h"
+#include "appmodel.h"
 #include "Utils.h"
 
 #include <wil/com.h>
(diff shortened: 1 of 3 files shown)


src/WinGetServer/WinGetServerManualActivation_Client.cpp
```diff
@@ -64,23 +61,112 @@ void InitializeRpcBinding()
...
+        UINT32 count = 0;
+        std::unique_ptr<PWSTR[]> fullNames;
+        UINT32 bufferLength = 0;
+        std::unique_ptr<WCHAR[]> buffer;
+        std::unique_ptr<UINT32[]> properties;
+
+        LONG result = FindPackagesByPackageFamily(pfn.c_str(), PACKAGE_FILTER_HEAD, &count, nullptr, &bufferLength, nullptr, nullptr);
+        THROW_WIN32_IF(result, result != ERROR_INSUFFICIENT_BUFFER);
+
+        for (size_t i = 0; i < 10 && result == ERROR_INSUFFICIENT_BUFFER; ++i)
```
yao-msft, line 118: Is this for "in case a new package is added between our calls"?

yao-msft (approved):

src/WinGetServer/WinGetServerManualActivation_Client.cpp
JohnMcPMS: It is defensive, yes.

PatrickSchmidtSE:
@JohnMcPMS Could this also help with general COM stuff ?.we are using the COM API to install/update packages. But performance is worse than using CLI . CLI works most of the time but COM method delivers "No applicable installer" for many packages. And cli with same options works just fine. 

If not I am happy to provide some details :)

JohnMcPMS:
> @JohnMcPMS Could this also help with general COM stuff ?.we are using the COM API to install/update packages. But performance is worse than using CLI . CLI works most of the time but COM method delivers "No applicable installer" for many packages. And cli with same options works just fine. Code must be also fine because some packages can be updates via COM.api. but success rate is 40% in COM vs 90% in cli
> 
> If not I am happy to provide some details :)

This should have no functional impact and probably wouldn't even affect your use.  Creating an issue with details is the best thing to do.  Make sure you include how exactly you are using the COM API (user, session, where you got the binaries, etc.)

PatrickSchmidtSE:
@JohnMcPMS https://github.com/microsoft/winget-cli/issues/4589 here is the issue :)

PR #4568, opened 2024-06-20
Update expired certificate with incoming one

JohnMcPMS:
## Change
Replace the expired store certificate with the newly generated one.

## Validation
Certificate is used as expected during a manual test.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4568)


yao-msft:
There are unit tests with Store certs chain 2 as test input. When updating the certs, we used to move chain 2 to chain 1, and put the new certs as chain 2. (Unless we fixed those unit tests to not test on Store certs)

JohnMcPMS:
> There are unit tests with Store certs chain 2 as test input. When updating the certs, we used to move chain 2 to chain 1, and put the new certs as chain 2. (Unless we fixed those unit tests to not test on Store certs)

From reviewing the tests, it doesn't look like it matters much which chain is 1 or 2.  The unit tests pass.

yao-msft:
> From reviewing the tests, it doesn't look like it matters much which chain is 1 or 2. The unit tests pass.

The unit tests pass because chain 2 certs have not expired yet. In Nov when chain 2 certs expired, the tests will fail. Like what Ruben did when chain 1 certs expired. https://github.com/microsoft/winget-cli/pull/3435

After that we try to make chain 2 certs the ones with later expiration dates. But I'm ok if we update the tests with our own created certs later.

yao-msft (approved):

PR #4570, opened 2024-06-21
Make Store package download stable

yao-msft:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4570)

### DIFF
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -323,17 +323,6 @@ Currently, this means that properly attributed configuration units (and only tho
    },
 ```
 
-### storeDownload
-
-This feature enables packages to be downloaded from the Microsoft Store.
-You can enable the feature as shown below.
-
-```json
-   "experimentalFeatures": {
-       "storeDownload": true
-   },
-```
-
 ### configureExport
 
 This feature enables exporting a configuration file.
--- a/schemas/JSON/settings/settings.schema.0.2.json
+++ b/schemas/JSON/settings/settings.schema.0.2.json
@@ -276,11 +276,6 @@
           "type": "boolean",
           "default": false
         },
-        "storeDownload": {
-          "description": "Enable support for downloading packages from the Microsoft Store",
-          "type": "boolean",
-          "default": false
-        },
         "configureExport": {
           "description": "Enable support for the configure export command",
           "type": "boolean",
--- a/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
@@ -247,7 +247,6 @@ namespace AppInstaller::CLI::Workflow
                 if (installerDownloadOnly)
                 {
                     context <<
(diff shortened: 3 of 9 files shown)


JohnMcPMS (approved):

rbalsleyMSFT:
@yao-msft Looks like this is removing the experimental feature. Is there an ETA on when this may be returning?

yao-msft:
Removing the experimental feature means this feature is now official. This feature will be in our next winget 1.8 release (and also 1.9 preview).

rbalsleyMSFT:
That's great to hear. Thanks!

Thanks,
Richard
________________________________

rbalsleyMSFT:
@yao-msft I'm using a PowerShell script to download multiple store apps with winget. I'm getting a consistent prompt for credentials when downloading store apps (I seem to get 2 prompts per app, one for the app itself and another for the dependencies I think, could also be due to multiple appx versions of the app being downloaded). I'm using an EntraID to download as expected, but I'm not signed into that account. 

Is there a way to invoke this to only be prompted once for credentials? Thanks.

cedricalfonsi:
@yao-msft We are encountering the same issue mentioned by @rbalsleyMSFT. We noticed that it's possible to download VSCode without authentication, so there might be a workaround to bypass this authentication, or is it reserved for Microsoft products?

yao-msft:
> @yao-msft I'm using a PowerShell script to download multiple store apps with winget. I'm getting a consistent prompt for credentials when downloading store apps (I seem to get 2 prompts per app, one for the app itself and another for the dependencies I think, could also be due to multiple appx versions of the app being downloaded). I'm using an EntraID to download as expected, but I'm not signed into that account.
> 
> Is there a way to invoke this to only be prompted once for credentials? Thanks.

The 2 prompts are: one for retrieving download links and the other for acquiring license. They are 2 different backend services. Winget uses Web Account Manager for Microsoft Entra Id authentications. Web Account Manager manages the token caches, it should behave the same with the Connect to Work or School experience of Windows Settings.

Did you try signing in System wide?
Did you use the Export-WingetPackage or the winget cmdline? For winget cmdline, you can specify the account to use using `--authentication-account`, if the account has been logged on recently, no popup will appear. If you are using Export-WingetPackage, currently it will try to logon with the default account (the current user session account), if failed, it will show the popup. PSCmdlet does not support specifying a specific account yet. If needed, please feel free to file an issue to ask for parity between cmdline version vs PSCmdlet for dowloading store apps (the work will require some winget COM api change then update the PSCmdlet)

yao-msft:
> @yao-msft We are encountering the same issue mentioned by @rbalsleyMSFT. We noticed that it's possible to download VSCode without authentication, so there might be a workaround to bypass this authentication, or is it reserved for Microsoft products?

Authentication is only required for installer type msstore (the appx/msix packages). VSCode is exe installer, it has download links in the manifest so it does not need authentication. A quick way to find out if authentication is needed is to look at package ids. If the Id starts with XP, it is non-msstore installer, no authentication needed. Otherwise, it's msstore installer type, it needs authentication.

PR #4577, opened 2024-06-22
Update winget server com security

yao-msft:
Change:

- Explicitly set COM access permissions for packaged com invocations. Leave access permissions as default and do not register COM objects for manual invocation so that only RPC channel can be used for manual activation.
- Update LaunchAndActivationString to allow Everyone and AppContainer only, require at least LowIL.
- Move Configuration to a separate COM server, only allow Everyone (No AppContainer access), require at least MediumIL.

A separate pr will be sent to update AppInstaller manifest.

Validation:
Validated manually with Microsoft Store invocation, Powershell invocation (elevated and non elevated), test sample code and Devhome invocation (on package management and configuration).

Also specifically validated Store invocation with Built-in Administrator sign-in (previously not working). 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4577)

### DIFF
--- a/src/AppInstallerCLIPackage/Package.appxmanifest
+++ b/src/AppInstallerCLIPackage/Package.appxmanifest
@@ -61,7 +61,7 @@
         </uap5:Extension>
         <com:Extension Category="windows.comServer">
           <com:ComServer>
-            <com:ExeServer Executable="WinGetServer\WindowsPackageManagerServer.exe" DisplayName="Windows Package Manager Server" LaunchAndActivationPermission="O:SYG:SYD:(A;;11;;;IU)(A;;11;;;SY)(A;;11;;;BA)(A;;11;;;AC)S:P(ML;;NX;;;S-1-16-4096)">
+            <com:ExeServer Executable="WinGetServer\WindowsPackageManagerServer.exe" DisplayName="Windows Package Manager Server" LaunchAndActivationPermission="O:SYG:SYD:(A;;11;;;IU)(A;;11;;;SY)(A;;11;;;BA)(A;;11;;;AC)S:P(ML;;NX;;;S-1-16-8192)">
               <com:Class Id ="74CB3139-B7C5-4B9E-9388-E6616DEA288C" DisplayName="PackageManager Server">
               </com:Class>
               <com:Class Id ="1BD8FF3A-EC50-4F69-AEEE-DF4C9D3BAA96" DisplayName="FindPackagesOptions Server">
--- a/src/AppInstallerCLIPackage/Package.appxmanifest
+++ b/src/AppInstallerCLIPackage/Package.appxmanifest
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Package 
   xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
   xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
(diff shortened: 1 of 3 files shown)


JohnMcPMS (commented):
Looks good to me; will forward to experts.

src/WinGetServer/WinMain.cpp
```diff
@@ -35,10 +36,10 @@ HRESULT WindowsPackageManagerServerInitializeRPCServer()
     RETURN_HR_IF(HRESULT_FROM_WIN32(status), status != RPC_S_OK);
 
     // The goal of this security descriptor is to restrict RPC server access only to the user in admin mode. 
-    // (ML;;NW;;;HI) specifies a high mandatory integrity level (requires admin).
+    // (ML;;NRNWNX;;;HI) specifies a high mandatory integrity level (requires admin).
     // (A;;GA;;;UserSID) specifies access only for the user with the user SID (i.e. self).
     wil::unique_hlocal_security_descriptor securityDescriptor;
-    std::string securityDescriptorString = "S:(ML;;NW;;;HI)D:(A;;GA;;;" + userSID + ")";
+    std::string securityDescriptorString = "S:(ML;;NRNWNX;;;HI)D:(A;;GA;;;" + userSID + ")";
```
brialmsft, line 42: Can't you use PS instead of calculating an explicit userSID?

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -61,7 +61,7 @@
         </uap5:Extension>
         <com:Extension Category="windows.comServer">
           <com:ComServer>
-            <com:ExeServer Executable="WinGetServer\WindowsPackageManagerServer.exe" DisplayName="Windows Package Manager Server" LaunchAndActivationPermission="O:SYG:SYD:(A;;11;;;WD)(A;;11;;;RC)(A;;11;;;AC)(A;;11;;;AN)S:P(ML;;NX;;;S-1-16-0)">
+            <com:ExeServer Executable="WinGetServer\WindowsPackageManagerServer.exe" DisplayName="Windows Package Manager Server" LaunchAndActivationPermission="O:SYG:SYD:(A;;11;;;WD)(A;;11;;;AC)S:P(ML;;NX;;;S-1-16-4096)">
```
brialmsft, line 64: Sanity check: you're fine with allowing non-AppContainer low IL to activate this?

JohnMcPMS, line 64: Set to medium, discuss in person.

src/WinGetServer/WinMain.cpp
```diff
@@ -100,6 +101,49 @@ extern "C" HRESULT CreateInstance(
     return S_OK;
 }
 
+HRESULT InitializeComSecurity()
+{
+    wil::unique_hlocal_security_descriptor securityDescriptor;
+    // Allow Everyone and AppContainer access. 3 is COM_RIGHTS_EXECUTE | COM_RIGHTS_EXECUTE_LOCAL
+    std::string securityDescriptorString = "O:SYG:SYD:(A;;3;;;WD)(A;;3;;;AC)";
```
hart-wilson, line 108: This is opening up permissions far more than is necessary to fix the bug as I understand it (High IL server processes that ran that way in a UAC-disabled account like the built-in Administrator account if enabled, rather than due to elevation, are not granting AC the way they would due to the defaults if they were running at Medium IL as usual). COM defaults for Medium IL processes grant PS, SY, BA, and AC. If you want the same permissions for both the usual Medium IL case and the less common High IL case, grant access to only those SIDs.

src/WinGetServer/WinMain.cpp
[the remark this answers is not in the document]
yao-msft: I'll revert the added access rights. Regarding why we use specific sid, the original intention was to ensure only whoever manually activated the server can access it.

src/AppInstallerCLIPackage/Package.appxmanifest
brialmsft: You can configure a SACL with a Medium mandatory label. This will block non-AppContainer low IL, but still allow AppContainer because you have a packaged SID in the DACL (ALL_APPLICATION_PACKAGES).

yao-msft: I just tested setting to 8192 and calls from AC still works. I'll set to 8192 as suggested.

src/WinGetServer/WinMain.cpp
```diff
@@ -100,6 +101,49 @@ extern "C" HRESULT CreateInstance(
     return S_OK;
 }
 
+HRESULT InitializeComSecurity()
+{
+    wil::unique_hlocal_security_descriptor securityDescriptor;
+    // Allow Everyone and AppContainer access. 3 is COM_RIGHTS_EXECUTE | COM_RIGHTS_EXECUTE_LOCAL
```
JohnMcPMS, line 107: Comment is no longer correct.

hart-wilson, line 107: Comment is out of date

hart-wilson (approved):

brialmsft (approved):

src/AppInstallerCLIPackage/Package.appxmanifest
hart-wilson: Ah, I forgot we just need a SACL to tell activation "we're handling that detail", using medium (8192) to allow only medium and above non-AC processes makes sense. Note that this is a new restriction though, existing non-AC < medium IL clients will be broken.

JohnMcPMS (approved):

yao-msft:
To unblock 1.8 servicing release process, I'll merge and not wait for build since it's only updating a comment from last success build and spelling task passed.

PR #4590, opened 2024-06-27
Configuration apply queueing

JohnMcPMS:
## Change
Adds a queue table to the configuration database and some code to synchronize the application of configurations.

Every apply in the queue puts a row in the table, with its instance identifier (it should also be in the history) and a named object that it will keep alive as long as it is in the queue.  This allows for other queued processes to check for dead queue items.

A global named mutex must be held in order to apply, or even check if one is at the front of the queue.  If not at the front of the queue, the waiting operation will release the mutex and wait for N * 100ms where N is their perceived position in the queue.  This should prevent repeated contention on the global mutex as the queued items sort themselves via the wait.

## Validation
Manual use with a configuration that allowed control of completion.
Added automated tests.
Plan to refactor out some of the internals in the future so that they can be directly unit tested.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4590)

### DIFF
--- a/src/Microsoft.Management.Configuration/Database/Schema/IConfigurationDatabase.cpp
+++ b/src/Microsoft.Management.Configuration/Database/Schema/IConfigurationDatabase.cpp
@@ -4,34 +4,74 @@
 #include "Database/Schema/IConfigurationDatabase.h"
 
 #include "Database/Schema/0_1/Interface.h"
+#include "Database/Schema/0_2/Interface.h"
 
 namespace winrt::Microsoft::Management::Configuration::implementation
 {
-    AppInstaller::SQLite::Version IConfigurationDatabase::GetLatestVersion()
+    namespace
     {
-        return { 0, 1 };
+        std::unique_ptr<IConfigurationDatabase> CreateForVersion(const AppInstaller::SQLite::Version& version, const std::shared_ptr<AppInstaller::SQLite::SQLiteDynamicStorage>& storage)
+        {
+            using StorageT = std::shared_ptr<AppInstaller::SQLite::SQLiteDynamicStorage>;
+
+            if (version.MajorVersion == 0)
+            {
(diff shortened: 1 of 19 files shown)


src/Microsoft.Management.Configuration/Database/Schema/IConfigurationDatabase.cpp
```diff
@@ -4,34 +4,74 @@
...
-            if (minorVersion >= 1 && minorVersion <= versionCreatorMap.size())
-            {
-                return versionCreatorMap[minorVersion - 1](std::move(storage));
-            }
+        AppInstaller::SQLite::Version latestVersion = GetLatestVersion();
+        if (allowMigration && version < latestVersion)
+        {
+            // Always migrate to the latest version until a reason comes along to not do that
+            std::unique_ptr<IConfigurationDatabase> latest = CreateForVersion(latestVersion, storage);
+            THROW_WIN32_IF(ERROR_NOT_SUPPORTED, !latest->MigrateFrom(result.get()));
```
yao-msft, line 54: Did we consider falling back to use previous version if migration failed? I saw the sequencer supports "database does not support queueing"

yao-msft (approved):

src/Microsoft.Management.Configuration/Database/Schema/IConfigurationDatabase.cpp
JohnMcPMS: I think that is probably reasonable past a 1.0 schema, especially if there is some actual potential for failure / data loss that we want to avoid.  The comment is intended to indicate that the hard failure isn't required forever.

yao-msft (approved):

Issue #1914, opened 2022-02-09
Switch to upgrade only minor version


Trenly:
### Description of the new feature / enhancement

* Related to #476 
* Requested in https://www.youtube.com/watch?v=7rBGVLsyc5w

As a user, I would like to be able to specify to only upgrade packages where a minor version is available.

### Proposed technical implementation details

`winget upgrade --minor` should list packages available for upgrade, where a minor version upgrade is available
`winget upgrade --all --minor` should install packages where a minor version upgrade is available

Other considerations -
It may be possible for there to be both a newer major version and a newer minor version available in the source. In this case, we should upgrade to the latest applicable minor version.

Example:
```
Installed Version:   1.0.1
Versions in Source:
  1.0.0
  1.0.1
  1.0.2
  1.0.3
  2.0.1

winget upgrade --all => Upgrades package to v2.0.1
winget upgrade --all --minor  => Upgrades package to v1.0.3
winget upgrade --minor => Shows available upgrade as v1.0.3
```

jedieaston:
Transcluded from #1915

We've been splitting manifests for major version numbers into different PackageIdentifiers in the community repo (when we notice, I still have to cleanup some of the dotnet bundles). It'd be nice for the client to have knowledge of minor vs major semvers, but as far as I know, people shouldn't be getting major version updates for Python, Java and Node.

(So if anyone thinks they are out of wack, please open a bug on [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) :D)

denelon:
This was just suggested by @shanselman in the Windows Wednesday [episode](https://www.youtube.com/watch?v=7rBGVLsyc5w) on Windows Package Manager.

OfficialEsco:
> We've been splitting manifests for major version numbers into different PackageIdentifiers in the community repo (when we notice, I still have to cleanup some of the dotnet bundles). It'd be nice for the client to have knowledge of minor vs major semvers, but as far as I know, people shouldn't be getting major version updates for Python, Java and Node.
> 
> (So if anyone thinks they are out of wack, please open a bug on [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) :D)

I actually think this is a different thought process

When we're splitting the PackageIdentifiers its because it can be installed side by side or considered a different package, however what this issue refers to is essentially applying the patches to the current major version because you're not sure what the Major change is (For example dropped support for some hardware)

[...]

PR #4602, opened 2024-06-30
Add Argument for Upgrading only minor versions

Trenly:
  - #1914
-----

This PR adds support for `--minor-version-only` and `--patch-version-only` in `winget upgrade`.  
`winget upgrade --minor` lists packages available for upgrade, where only a minor version upgrade is available.
`winget upgrade --all --minor` installs packages where only a minor version upgrade is available.

It may be possible for there to be both a newer major version and a newer minor version available in the source. In this case, the upgrade will be skipped entirely. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4602)

### DIFF
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -172,9 +172,9 @@ namespace AppInstaller::CLI
         case Execution::Args::Type::UninstallPrevious:
             return { type, "uninstall-previous"_liv, ArgTypeCategory::InstallerBehavior | ArgTypeCategory::CopyFlagToSubContext };
         case Execution::Args::Type::MinorVersionOnly:
-            return { type, "minor-version-only"_liv, NoAlias, "minor"_liv, ArgTypeCategory::MultiplePackages, ArgTypeExclusiveSet::VersionSelection };
+            return { type, "minor-version-only"_liv, NoAlias, "minor"_liv, ArgTypeCategory::MultiplePackages | ArgTypeCategory::CopyFlagToSubContext, ArgTypeExclusiveSet::VersionSelection };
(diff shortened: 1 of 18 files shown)


Trenly:
> This seems like a blocker to me, and I can't think of any technical reason why it can't be implemented.

The only reason that I can think of is my own incompetence. When looking at the current state of the Upgrade flow, there are actually two different paths that get followed. There is one workflow that does the filtering for showing the table of available upgrades and a completely different workflow that does the filtering for actually installing the available upgrades.

With my current skill level and understanding of the code, I don't think I'm able to reconcile these to a single workflow for selecting versions for upgrade.  Even without creating a combined workflow to select the versions for upgrade, I don't fully understand how the version selection works.

I do agree that it should be possible - in fact, thinking about it more, the logic seems to exist *somewhere* for the gating pin, I just don't know where or how.

> I also think it would be better to create a syntax for specifying version rather than new parameters. 

Yeah, this is what got me thinking about gating pins. It certainly would be a better implementation to have a syntax, but again, I don't really understand where to begin with that.


-----

I'll mark this as a draft, since I agree with the comments and don't think this implementation should be merged after further consideration. I'll have to do some more learning and digging in the code to better understand how the pinning database (and especially gating pins) work. I feel like once I have an understanding of how that needs to be used to filter versions for install, then I'll be better equipped to start looking at creating a syntax

Trenly:
After more careful review, I think creating a syntax is above my skill level. I'd love to learn more about it, but it isn’t a priority at this point. 

Since an argument isn’t the right way to do this, will close this PR

PR #4707, opened 2024-08-05
Validate configuration Urls

AmelBawa-msft:
## Description
- Added group policies for applying winget configuration based on url zone
- Added validation for urls from 'Internet' zone.
- In Dev mode, validation is omitted

## TODO
- [ ] Unit tests
- [ ] Finalize strings

## Follow ups
- Add MotW HostUrl validation (TODO link to issue)


src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -959,6 +961,87 @@ namespace AppInstaller::CLI::Workflow
...
+                context.Reporter.Error() << "Configuration is disabled for Zone: " << zone << std::endl;
+                return true;
+            }
+
+            AICLI_LOG(Config, Info, << "Configuration is configured in zone " << zone << " with value " << (isAllowed ? "allowed" : "blocked"));
+            return false;
+        }
+
+        // Validate smart screen for a given url.
+        bool IsBlockedBySmartScreen(Execution::Context& context, const std::string& url)
```
Trenly, line 1008: Should these be in a SmartScreen workflow file, or somewhere other than the configuration workflows? I feel there is opportunity for other URLs to eventually be run through SmartScreen, and segmenting the SmartScreen functionality into its own workflows would make it easier to follow in the future

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -959,6 +961,87 @@ namespace AppInstaller::CLI::Workflow
...
+            if (configurationPolicies->find(zone) == configurationPolicies->end())
+            {
+                AICLI_LOG(Config, Warning, << "Configuration is not configured in the zone " << zone);
+                return false;
+            }
+
+            auto isAllowed = configurationPolicies->at(zone);
+            if(!isAllowed)
+            {
+                context.Reporter.Error() << "Configuration is disabled for Zone: " << zone << std::endl;
```
AmelBawa-msft, line 999: TODO:
- Finalizing strings
- Move to resource file

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
AmelBawa-msft: Great feedback! Yes, ultimately this can be beneficial for workflows and other places where we need a consistent logic for dealing with URLs.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -959,6 +961,87 @@ namespace AppInstaller::CLI::Workflow
...
+        {
+            auto zone = GetUriZone(uri);
+            if(IsBlockedByGroupPolicy(context, zone))
+            {
+                return APPINSTALLER_CLI_ERROR_BLOCKED_BY_POLICY;
+            }
+
+            if (IsSmartScreenRequired(zone) && IsBlockedBySmartScreen(context, uri))
+            {
+                return APPINSTALLER_CLI_ERROR_SOURCE_NOT_SECURE;
```
AmelBawa-msft, line 1039: @JohnMcPMS, any suggestion if instead I should create a new error code here? e.g. 'blocked by smartscreen'?

JohnMcPMS: You should not use that error code.  I would also prefer to be able to differentiate between a policy-based block and one based on inspection of the URI by a service.  The relevant party to take things up with is different between these two cases.

Given that, I would suggest two new error codes: "blocked by policy" and "blocked by reputation service".

AmelBawa-msft: Great! Yes, that error code was awkward here 🙂 I will add a new one, thank you. For the "blocked by policy" one, I am using `APPINSTALLER_CLI_ERROR_BLOCKED_BY_POLICY` in the line above. Is your suggestion to create another one like APPINSTALLER_CLI_ERROR_URL_ZONE_BLOCKED_BY_POLICY?

JohnMcPMS: No, using that existing one is fine.  I didn't look into the change to see that you were already doing that.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -959,6 +961,87 @@ namespace AppInstaller::CLI::Workflow
...
+            }
+
+            return NO_ERROR;
```
JohnMcPMS, line 1042:
```suggestion
            return S_OK;
```
No functional difference, but `NO_ERROR` is a Win32 error code definition while `S_OK` is an HRESULT definition.

doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -122,6 +122,10 @@ If you disable this setting, users will not be able to use the Windows Package M
 If you disable or do not configure this setting, no proxy will be used by default.
 
 If you enable this setting, the specified proxy will be used by default.</string>
+      <string id="EnableDSCAllowedZones">Enable App Installer Allowed Zones for DSC</string>
+      <string id="EnableDSCAllowedZonesExplanation"></string>
```
JohnMcPMS, line 126: Waiting for PM input?

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -959,6 +961,87 @@ namespace AppInstaller::CLI::Workflow
...
+        }
+
+        // Validate smart screen for a given url.
+        bool IsBlockedBySmartScreen(Execution::Context& context, const std::string& url)
+        {
+            auto response = AppInstaller::UriValidation::UriValidation(url);
+            switch (response.Decision())
+            {
+            case AppInstaller::UriValidation::UriValidationDecision::Block:
+                context.Reporter.Error() << std::endl << "Blocked by smart screen" << std::endl << "Feedback: " << response.Feedback() << std::endl;
```
JohnMcPMS, line 1014: If `Feedback` isn't localized, it should probably only be sent to the log and not stdout.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -977,6 +1060,13 @@ namespace AppInstaller::CLI::Workflow
                     AICLI_TERMINATE_CONTEXT(ERROR_NOT_SUPPORTED);
                 }
 
+                auto uriValidation = EvaluateUri(context, argPath);
+                if (uriValidation != NO_ERROR)
```
JohnMcPMS, line 1064:
```suggestion
                if (FAILED(uriValidation))
```

src/Internal/UriValidation/UriValidation.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include "pch.h"
+#include "UriValidation.h"
+
+namespace AppInstaller::UriValidation
+{
+    UriValidationResult UriValidation(const std::string&)
+    {
+        // In Dev mode, allow all URIs
+        return UriValidationResult(UriValidationDecision::Allow, std::string());
```
JohnMcPMS, line 12: Needs test consideration, like have specific URIs that trigger specific results.

src/Internal/UriValidation/pch.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+
+#define NOMINMAX
```
JohnMcPMS, line 5: This seems like a copy/paste that doesn't need to have so much in it.  Given the single file nature of this lib, it doesn't even need a precompiled header.

src/Internal/UriValidation/UriValidation.cpp
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License.
+
+#include "pch.h"
+#include "UriValidation.h"
+
+namespace AppInstaller::UriValidation
+{
+    UriValidationResult UriValidation(const std::string&)
+    {
+        // In Dev mode, allow all URIs
```
Trenly, line 11: I'm a bit confused by this comment, can you help me understand it? The way it reads to me at the moment is that because you're currently testing you want to allow all URI's?

This new functionality may also benefit from having test hooks (similar to the Zip file malware scan) if it isn’t already needed for the tests John mentioned

https://github.com/microsoft/winget-cli/blob/2306d08b6f2ed2b6d23a3442258e1f2067c13b37/src/AppInstallerCommonCore/Archive.cpp#L52-L68

src/Internal/UriValidation/UriValidation.h
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+
+namespace AppInstaller::UriValidation
+{
+    // The decision made based on the Uri validation.
+    enum class UriValidationDecision
+    {
+        Allow,
```
Trenly, line 10: Isn't there also a `Warn` response that SmartScreen can return? 

https://learn.microsoft.com/en-us/defender-endpoint/web-protection-overview#order-of-precedence

src/Internal/UriValidation/UriValidation.h
```diff
@@ -0,0 +1,32 @@
...
+    enum class UriValidationDecision
+    {
+        Allow,
+        Block,
+    };
+
+    // The result of a Uri validation.
+    class UriValidationResult
+    {
+    private:
```
Trenly, line 17: Nit: In most other class definitions for this WinGet, the privates and member variables are defined at the bottom of the class; I'm not sure it this is a style convention or just happenstance

src/Internal/UriValidation/UriValidation.h
```diff
@@ -0,0 +1,32 @@
...
+        // The decision made based on the Uri validation.
+        UriValidationDecision m_decision;
+
+        // Uri to give feedback to smart screen about the decision.
+        std::string m_feedback;
+    public:
+        UriValidationResult(UriValidationDecision decision) : m_decision(decision), m_feedback(std::string()) {}
```
Trenly, line 24: Should there be a default (enpty) constructor?

src/Internal/UriValidation/UriValidation.cpp
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "pch.h"
+#include "UriValidation.h"
+
+namespace AppInstaller::UriValidation
+{
+    UriValidationResult UriValidation(const std::string&)
```
Trenly, line 9: The naming of the function seems a bit odd. Perhaps it's just me though?

src/Internal/UriValidation/UriValidation.vcxproj
```diff
@@ -0,0 +1,161 @@
...
+      <PrecompiledHeader>Create</PrecompiledHeader>
+    </ClCompile>
+    <ClCompile Include="UriValidation_internal.cpp" Condition="Exists('$(MSBuildProjectDirectory)\UriValidation_internal.cpp')" />
+    <ClCompile Include="UriValidation.cpp" Condition="!Exists('$(MSBuildProjectDirectory)\UriValidation_internal.cpp')" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" Condition="Exists('$(MSBuildProjectDirectory)\packages.config')" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Windows.CppWinRT" Version="2.0.230706.1" TargetFramework="native" />
```
Trenly, line 142: This doesnt seem to match the version of cppwinrt in NOTICE.txt.  Does the version need to be bumped across all the solution files, or is there something I’m missing?

https://github.com/microsoft/winget-cli/blob/2306d08b6f2ed2b6d23a3442258e1f2067c13b37/NOTICE#L769

doc/admx/en-US/DesktopAppInstaller.adml
AmelBawa-msft: Yes, but also, we're still thinking about the scope that this policy should cover (DSC only, or more ... )

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
AmelBawa-msft: All strings are not finalized/localized yet, there's an email thread for deciding on a couple things including localization

src/AppInstallerSharedLib/Public/winget/GroupPolicy.h
```diff
@@ -23,9 +23,20 @@ namespace AppInstaller::Settings
         AdditionalSources,
         AllowedSources,
         DefaultProxy,
+        ConfigurationAllowedZones,
         Max,
     };
 
+    // Enum for the configurable zones for the configuration policy.
+    enum class ConfigurationAllowedZonesOptions : DWORD
```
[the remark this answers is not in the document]
AmelBawa-msft: This enum reflects the internet zone values for a Url, and it is used here to make the values returned by MapUrlToZone  more descriptive/readable.

That said, the `enum` name is not finalized yet and may be updated before publishing the PR.

Related links:
- https://learn.microsoft.com/en-us/previous-versions/troubleshoot/browsers/security-privacy/ie-security-zones-registry-entries#zones
```
Value  Setting
------------------------------
0      My Computer
1      Local Intranet Zone
2      Trusted sites Zone
3      Internet Zone
4      Restricted Sites Zone
```

src/Internal/UriValidation/UriValidation.cpp
AmelBawa-msft: The actual SmartScreen logic/code will be injected in the release build pipeline and this method will be substituted. I plan on adding tests later. John mentioned to me offline that there are test hooks I can leverage, but I haven't had the chance to explore them yet. Thank you for providing an example, I will consider that once I add some tests.

src/Internal/UriValidation/UriValidation.h
AmelBawa-msft: You're right, there's a PUA response as well. However, PUA/Warn will also be treated as Blocked, but the details will be communicated in the logs instead and the logic will be in the injected code. I can potentially reduce this enum to a bool. I will revisit this later once things are a little bit more finalized.

doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -122,20 +122,24 @@ If you disable this setting, users will not be able to use the Windows Package M
 If you disable or do not configure this setting, no proxy will be used by default.
 
 If you enable this setting, the specified proxy will be used by default.</string>
+      <string id="EnableWindowsPackageManagerAllowedSecurityZones">Enable App Installer Allowed Zones for the Windows Package Manager</string>
+      <string id="EnableWindowsPackageManagerAllowedSecurityZonesExplanation"></string>
+      <string id="EnableWindowsPackageManagerSmartScreenCheck">Enable Microsoft SmartScreen checks for the Windows Package Manager</string>
+      <string id="EnableWindowsPackageManagerSmartScreenCheckExplanation"></string>
```
AmelBawa-msft, line 128: Adding @RDMacLachlan to help with the strings for the new group policies in the ADML

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2685,6 +2685,14 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Uri not well formed: {0}</value>
     <comment>{Locked="{0}"} Error message displayed when the provided uri is not well formed. {0} is a placeholder replaced by the provided uri.</comment>
   </data>
+  <data name="UriBlockedBySmartScreen" xml:space="preserve">
+    <value>This operation was blocked as unsafe by Microsoft Defender SmartScreen.</value>
```
AmelBawa-msft, line 2689: Adding @RDMacLachlan to help with the displayed messages (I copied the messages from App Installer with some modifications)

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+        return S_OK;
+    }
+
+    // Evaluate the configuration uri for group policy and smart screen.
+    HRESULT EvaluateConfigurationUri(Execution::Context& context)
+    {
+        if (context.Args.Contains(Execution::Args::Type::ConfigurationFile))
+        {
+            std::string argPath{ context.Args.GetArg(Execution::Args::Type::ConfigurationFile) };
+            return EvaluateUri(context, argPath);
```
florelis, line 149: Nit: You can pass the `string_view` you get from `GetArg()` directly to `EvaluateUri()` without copying it to a `string`

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+        case AppInstaller::UriValidation::UriValidationDecision::Allow:
+        default:
+            return false;
+        }
+    }
+
+    // Get Uri zone for a given uri or file path.
+    HRESULT GetUriZone(const std::string& uri, Settings::SecurityZoneOptions* zone)
+    {
+        if (!zone)
```
florelis, line 42: Probably better to pass as a reference (`Settings::SecurityZoneOptions& zone`) so that you don't have to worry about bad pointers

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+        #endif
+
+        DWORD dwZone;
+        auto pInternetSecurityManager = winrt::create_instance<IInternetSecurityManager>(CLSID_InternetSecurityManager, CLSCTX_ALL);
+        auto mapResult = pInternetSecurityManager->MapUrlToZone(AppInstaller::Utility::ConvertToUTF16(uri).c_str(), &dwZone, 0);
+
+        // Ensure MapUrlToZone was successful and the zone value is valid
+        if (FAILED(mapResult))
+        {
+            return mapResult;
```
florelis, line 65: Nit: Log the result

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    // Validate group policy for a given zone.
+    bool IsBlockedByGroupPolicy(Execution::Context& context, const Settings::SecurityZoneOptions zone)
+    {
+        if (!Settings::GroupPolicies().IsEnabled(Settings::TogglePolicy::Policy::AllowedSecurityZones))
+        {
+            AICLI_LOG(Core, Info, << "WindowsPackageManagerAllowedSecurityZones policy is disabled");
+            return false;
+        }
```
florelis, line 88: With my other suggestion you can get rid of this and just use the `allowedSecurityZones.has_value()` to know if it is enabled

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+        auto allowedSecurityZones = Settings::GroupPolicies().GetValue<Settings::ValuePolicy::AllowedSecurityZones>();
+        if (!allowedSecurityZones.has_value())
+        {
+            AICLI_LOG(Core, Warning, << "WindowsPackageManagerAllowedSecurityZones policy is not set");
+            return false;
+        }
+
+        if (allowedSecurityZones->find(zone) == allowedSecurityZones->end())
+        {
+            AICLI_LOG(Core, Warning, << "Security zone " << zone << " was not found in the group policy WindowsPackageManagerAllowedSecurityZones");
```
florelis, line 99: Nit: Should we log the zone as a string instead of the numerical value?
Nit: I think this would be easier to understand if it was along the lines of "No Group Policy configured for Security zone x"

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        auto allowedSecurityZones = Settings::GroupPolicies().GetValue<Settings::ValuePolicy::AllowedSecurityZones>();
+        if (!allowedSecurityZones.has_value())
+        {
+            AICLI_LOG(Core, Warning, << "WindowsPackageManagerAllowedSecurityZones policy is not set");
+            return false;
+        }
+
+        if (allowedSecurityZones->find(zone) == allowedSecurityZones->end())
```
florelis, line 97: Nit: Doesn't matter because of how few values there are, but I'd prefer to keep the iterator from `find()` to not have to do a search again with `at()`

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+        auto isAllowed = allowedSecurityZones->at(zone);
+        if(!isAllowed)
+        {
+            AICLI_LOG(Core, Error, << "Security zone " << zone << " is blocked by group policy");
+            context.Reporter.Error() << Resource::String::UriSecurityZoneBlockedByPolicy << std::endl;
+            return true;
+        }
+
+        AICLI_LOG(Core, Info, << "Configuration is configured in zone " << zone << " with value " << (isAllowed ? "allowed" : "blocked"));
```
florelis, line 111: "Configuration is configured"? Also, no need for the conditional at the end as you already know the value of `isAllowed`

src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
```diff
@@ -268,6 +257,25 @@ namespace AppInstaller::CLI::Workflow
...
+            CheckForExistingInstaller;
+
+        if (context.IsTerminated())
+        {
+            return;
+        }
+
+        context <<
+            ExecuteUriValidation(UriValidationSource::PackageCatalogSource) <<
+            DownloadInstallerInternal;
```
florelis, line 276: `<<` only executes if the context isn't terminated
```suggestion
        context <<
            ReportExecutionStage(ExecutionStage::Download) <<
            CheckForExistingInstaller <<
            ExecuteUriValidation(UriValidationSource::PackageCatalogSource) <<
            DownloadInstallerInternal;
```

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1298,14 +1299,21 @@ namespace AppInstaller::CLI::Workflow
...
         auto progressScope = context.Reporter.BeginAsyncProgress(true);
         progressScope->Callback().SetProgressMessage(Resource::String::ConfigurationInitializing());
 
         anon::ConfigureProcessorForUse(context, ConfigurationProcessor{ anon::CreateConfigurationSetProcessorFactory(context) });
     }
 
+    void CreateConfigurationProcessor(Context& context)
+    {
+        context <<
+            ExecuteUriValidation(UriValidationSource::ConfigurationSource) <<
```
florelis, line 1313: Could `ExecuteUriValidation` happen after we set up the progress reporting and message? Just in case it takes a while

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2834,6 +2842,14 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Enable Windows Package Manager proxy command line options</value>
     <comment>Describes a Group Policy that can enable the use of the --proxy option to set a proxy</comment>
   </data>
+  <data name="PolicyEnableSmartScreenValidation" xml:space="preserve">
+    <value>Enable Windows Package Manager smart screen validation</value>
```
florelis, line 2846:
```suggestion
    <value>Enable Windows Package Manager SmartScreen validation</value>
```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2834,6 +2842,14 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Enable Windows Package Manager proxy command line options</value>
     <comment>Describes a Group Policy that can enable the use of the --proxy option to set a proxy</comment>
   </data>
+  <data name="PolicyEnableSmartScreenValidation" xml:space="preserve">
+    <value>Enable Windows Package Manager smart screen validation</value>
+    <comment></comment>
+  </data>
+  <data name="PolicyEnableAllowedSecurityZones" xml:space="preserve">
+    <value>Enable Windows App Installer Allowed Security Zones</value>
```
florelis, line 2850:
```suggestion
    <value>Enable Windows Package Manager Allowed Security Zones</value>
```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2685,6 +2685,14 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Uri not well formed: {0}</value>
     <comment>{Locked="{0}"} Error message displayed when the provided uri is not well formed. {0} is a placeholder replaced by the provided uri.</comment>
   </data>
+  <data name="UriBlockedBySmartScreen" xml:space="preserve">
+    <value>This operation was blocked as unsafe by Microsoft Defender SmartScreen.</value>
+    <comment>Error message displayed when an operation is using a URI that was blocked by Microsoft Defender SmartScreen.</comment>
+  </data>
+  <data name="UriSecurityZoneBlockedByPolicy" xml:space="preserve">
+    <value>The operation you are attempting to apply has been blocked by your administrator.</value>
```
florelis, line 2693: This message could probably be more specific to zones. I think we already have a generic "not allowed by policy" string

src/AppInstallerCLITests/WorkflowCommon.h
```diff
@@ -13,6 +13,9 @@
     REQUIRE(_context_.IsTerminated()); \
     REQUIRE(_hr_ == _context_.GetTerminationHR())
 
+#define REQUIRE_OUTPUT_HAS_LOC(_output_, _resource_) \
+    REQUIRE(_output_.str().find(Resource::LocString(_resource_).get()) != std::string::npos);
```
florelis, line 17: I like this, we should use it everywhere we do this check

src/AppInstallerCLICore/Workflows/UriValidationFlow.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include "pch.h"
+#include "UriValidationFlow.h"
+#include <AppInstallerDownloader.h>
+#include <UriValidation/UriValidation.h>
+#include <regex>
+
+namespace AppInstaller::CLI::Workflow
+{
+    // Check if smart screen is required for a given zone.
+    bool IsSmartScreenRequired(Settings::SecurityZoneOptions zone)
```
florelis, line 12: We generally put all the internal helpers inside an anonymous namespace
```
{
    namespace
    {
        // Helpers here
    }
}
```

src/AppInstallerSharedLib/Public/AppInstallerErrors.h
```diff
@@ -152,6 +152,7 @@
 #define APPINSTALLER_CLI_ERROR_SFSCLIENT_PACKAGE_NOT_SUPPORTED      ((HRESULT)0x8A150084)
 #define APPINSTALLER_CLI_ERROR_LICENSING_API_FAILED_FORBIDDEN       ((HRESULT)0x8A150085)
 #define APPINSTALLER_CLI_ERROR_INSTALLER_ZERO_BYTE_FILE             ((HRESULT)0x8A150086)
+#define APPINSTALLER_CLI_ERROR_BLOCKED_BY_REPUTATION_SERVICE        ((HRESULT)0x8A150087)
```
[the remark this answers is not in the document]
mdanish-kh: And would be great if we could document the new error codes in [returnCodes.md](https://github.com/microsoft/winget-cli/blob/master/doc/windows/package-manager/winget/returnCodes.md)

florelis:
I think we should add a command line flag to bypass SmartScreen if it fails, and an admin setting to disable the SmartScreen checks without needing to use group policy.

src/AppInstallerSharedLib/Public/AppInstallerErrors.h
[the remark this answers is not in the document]
florelis: Maybe we should move this comment from the top of the list:
`// Changes to any of these errors require the corresponding resource string in winget.resw to be updated.`
and add all the other places that need to be updated

PR #4731, opened 2024-08-13
Generate Microsoft.WinGet.Client help docs in cmdlet output

ryfu-msft:
Builds the help doc xml and includes it in the Microsoft.WinGet.Client powershell module. 

`New-ExternalHelp` cmdlet does verify if the docs are formatted correctly so this is more of a sanity check. Will update AppInstaller pipeline to do the same before publishing.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4731)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -215,6 +215,15 @@ jobs:
       TargetFolder: $(buildOutDirAnyCpu)\PowerShell\Microsoft.WinGet.Configuration\SharedDependencies\$(BuildPlatform)
       flattenFolders: true
 
+  - task: PowerShell@2
+    displayName: Generate Microsoft.WinGet.Client Help Documentation
+    pwsh: true
+    inputs:
+      targetType: 'inline'
+      script: |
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -215,6 +215,16 @@ jobs:
       TargetFolder: $(buildOutDirAnyCpu)\PowerShell\Microsoft.WinGet.Configuration\SharedDependencies\$(BuildPlatform)
       flattenFolders: true
 
+  - task: PowerShell@2
```
JohnMcPMS, line 218: This seems better to be in the PowerShell job, which only runs once (not per-architecture) and because it is explicitly PowerShell related.  It could also target the artifacts directory for its output so that one could download the artifact and test with the help included (at least, I would hope it could work that way).

JohnMcPMS (commented):
Rerun failed tasks worked like expected with my pipeline changes!  This is the first time I have seen it tried.

PR #4732, opened 2024-08-13
Fix Reset-WinGetSource behavior

ryfu-msft:
Issue:

Reset-WinGetSource required a name which does not match the current behavior of `winget source reset`. 

Changes:
- Added '-All' parameter so that Reset-WinGetsource will reset all sources. I also created two separate parameter sets so that they can only be used separately.
- Made the '-Name' parameter also positional=0 for Reset-WinGetsource and Remove-WinGetsource (recommended by PowerShell docs team)
- Fixed the naming of PreRelease -> Prerelease as recommended by PowerShell docs team
- 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4732)

### DIFF
--- a/src/PowerShell/Help/Microsoft.WinGet.Client/Reset-WinGetSource.md
+++ b/src/PowerShell/Help/Microsoft.WinGet.Client/Reset-WinGetSource.md
@@ -17,13 +17,13 @@ Resets WinGet sources.
 ### DefaultSet (Default)
 
 ```
-Reset-WinGetSource [-Name <String>] [<CommonParameters>]
+Reset-WinGetSource -Name <String> [<CommonParameters>]
 ```
 
 ### OptionalSet
 
 ```
-Reset-WinGetSource [-All] [<CommonParameters>]
+Reset-WinGetSource -All [<CommonParameters>]
 ```
 
 ## DESCRIPTION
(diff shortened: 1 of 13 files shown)


src/PowerShell/Help/Microsoft.WinGet.Client/Reset-WinGetSource.md
```diff
@@ -15,17 +15,25 @@ Resets default WinGet sources.
 ## SYNTAX
 
 ```
-Reset-WinGetSource -Name <String> [<CommonParameters>]
+Reset-WinGetSource [-Name <String>] [<CommonParameters>]
```
JohnMcPMS, line 18: I worry somewhat about this being too easy to destroy state; the reason one needs to specify `--force` to winget.exe.

Maybe it should be two parameter sets, a default set with `-Name` and another one with `-All`.  That way just running `> Reset-WinGetSource` doesn't immediately try to reset everything, but instead asks for a name.

ryfu-msft: I created two parameters sets (DefaultSet and OptionalSet) so that we keep the default behavior of requiring the Name parameter. I also added an optional -All switch parameter that resets all sources but it is optional.

JohnMcPMS: Shouldn't `Name` and `All` be required within their respective parameter set?

JohnMcPMS (approved):

src/PowerShell/Help/Microsoft.WinGet.Client/Reset-WinGetSource.md
ryfu-msft: Removed brackets so they show up as being required.

JohnMcPMS (approved):

PR #4740, opened 2024-08-19
Remove preference for "installed" installer type

### LINKED ISSUE #2565: Wix and MSI not considered equal

When both WIX and MSI are specified in the manifest, architecture may be disregarded if the install technology is detected as MSI. This causes issues when packages require specific architecture to run

### Steps to reproduce

* Open a new 64-bit VM / Sandbox
* `winget install Tortoisesvn -v 1.14.29387`
* `winget upgrade tortoisesvn --verbose-logs`

### Expected behavior

Application to be upgraded successfully

### Actual behavior

```
2022-09-30 14:03:09.828 [CLI ] Starting installer selection.
2022-09-30 14:03:09.828 [CLI ] Installer [X64,wix,Machine,en-US] is current best choice
2022-09-30 14:03:09.828 [CLI ] Installer [X86,msi,Machine,en-US] is better than [X64,wix,Machine,en-US] due to: Installed Type
2022-09-30 14:03:09.828 [CLI ] Installer [X86,msi,Machine,en-US] is current best choice
2022-09-30 14:03:09.828 [CLI ] Installer [Arm64,msi,Machine,en-US] not applicable: Machine is not compatible with Arm64
```

Application cannot be installed as 32-bit installer will not run on x64

### Environment

```shell
PS C:\Windows\system32> winget --info
Windows Package Manager v1.3.2091

Windows: Windows.Desktop v10.0.19041.2006
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.18.2091.0

```

JohnMcPMS:
Fixes #2565 

## Change
Move (and rename) the `InstalledTypeComparator` from being a comparator to just a filter.  Given what we can (or rather, currently do) actually discern from system state, it doesn't make sense to give preference to the few types that we use to signify the compatibility sets.  If we truly want to preserve installer type selection within a compatibility set, we need to store the user selection into the tracking database similar to `UserIntentArchitecture`.

## Validation
Add a test for the case as called out by the issue; regression tests cover it as well and the only one that needed updating was relying on the intentionally changed behavior.

While this change is arguably "correct", it will cause different choices to be made if there are packages with multiple compatible installers that are equal in the other relevant dimensions.  I'm working on determining the extent of impact to winget-pkgs via a custom validation pass.

If it turns out to be overly broad, the most likely solution will be to keep the existing behavior but drop from `Strong` to `Weak` for the case of equal installer types.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4740)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
+++ b/src/AppInstallerCLICore/Workflows/ManifestComparator.cpp
@@ -364,12 +364,12 @@ namespace AppInstaller::CLI::Workflow
             }
         };
 
-        struct InstalledTypeComparator : public details::FilterField
+        struct InstalledTypeFilter : public details::FilterField
         {
-            InstalledTypeComparator(Manifest::InstallerTypeEnum installedType) :
+            InstalledTypeFilter(Manifest::InstallerTypeEnum installedType) :
                 details::FilterField("Installed Type"), m_installedType(installedType) {}
 
-            static std::unique_ptr<InstalledTypeComparator> Create(const Repository::IPackageVersion::Metadata& installationMetadata)
+            static std::unique_ptr<InstalledTypeFilter> Create(const Repository::IPackageVersion::Metadata& installationMetadata)
             {
                 auto installerTypeItr = installationMetadata.find(Repository::PackageVersionMetadata::InstalledType);
                 if (installerTypeItr != installationMetadata.end())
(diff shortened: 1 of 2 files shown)


JohnMcPMS:
The only issue I uncovered when searching for package versions with installers in the same compatibility set that would be indistinct for upgrades was with multiple versions of `ToEverything.AFFiNE` (since they added an Nullsoft version of their installer).  For example: https://github.com/microsoft/winget-pkgs/blob/master/manifests/t/ToEverything/AFFiNE/0.16.3/ToEverything.AFFiNE.installer.yaml

This is already an issue during upgrade, as for the same reason as the original issue we will prefer `exe`.  With my change, the two x64, user scope installers will be equivalent in priority for upgrade (as they are for install already).  We will still choose the `exe` as it comes first in the list rather than because it matches the detected type of `exe`.

This isn't really an issue, and it sets the likelihood of needing the `UserIntentInstallerType` to be implemented to very low.

yao-msft (approved):

PR #4750, opened 2024-08-22
Add pwsh support for trust level and explicit

ryfu-msft:
Adds support for `-TrustLevel` and `-Explicit`. 

The pwsh tests can now use the Add-winGetsource cmdlet to modify the trust level. 

Updated Get-WinGetsource to report the TrustLevel and Explicit flag.

Updated tests and docs to reflect these changes.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4750)

### DIFF
--- a/src/Microsoft.Management.Deployment/PackageManager.idl
+++ b/src/Microsoft.Management.Deployment/PackageManager.idl
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 namespace Microsoft.Management.Deployment
 {
-    [contractversion(10)]
+    [contractversion(11)] // For version 1.9
     apicontract WindowsPackageManagerContract{};
 
     /// State of the install
@@ -263,7 +263,7 @@ namespace Microsoft.Management.Deployment
         /// The trust level of the package catalog
         PackageCatalogTrustLevel TrustLevel { get; };
 
-        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 10)]
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 11)]
         {
             /// Excludes a source from discovery unless specified.
             Boolean Explicit{ get; };
--- a/src/Microsoft.Management.Deployment/PackageManager.idl
+++ b/src/Microsoft.Management.Deployment/PackageManager.idl
@@ -262,6 +262,12 @@ namespace Microsoft.Management.Deployment
         PackageCatalogOrigin Origin { get; };
         /// The trust level of the package catalog
         PackageCatalogTrustLevel TrustLevel { get; };
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 10)]
+        {
+            /// Excludes a source from discovery unless specified.
+            Boolean Explicit{ get; };
+        }
     }
 
     /// A metadata item of a package version.
--- a/src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddSourceCmdlet.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddSourceCmdlet.cs
@@ -46,7 +46,7 @@ namespace Microsoft.WinGet.Client.Cmdlets.Cmdlets
         /// <summary>
         /// Gets or sets the trust level of the source to add.
         /// </summary>
-        [Parameter(ValueFromPipeline = true, ValueFromPipelineByPropertyName = true)]
+        [Parameter(ValueFromPipelineByPropertyName = true)]
         public PSSourceTrustLevel TrustLevel { get; set; } = PSSourceTrustLevel.Default;
 
         /// <summary>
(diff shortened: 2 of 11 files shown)


src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -262,6 +262,12 @@ namespace Microsoft.Management.Deployment
         PackageCatalogOrigin Origin { get; };
         /// The trust level of the package catalog
         PackageCatalogTrustLevel TrustLevel { get; };
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 10)]
```
JohnMcPMS, line 266: You can't add things to contract 10, it shipped with 1.7.

On a related note, when you add contract 11 at the top of the file, put a comment above it like `// For version 1.9`

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/CliCommand.cs
```diff
@@ -70,17 +70,33 @@ public void GetSettings(bool asPlainText)
...
+        /// <param name="isExplicit">Make source explicit.</param>
+        public void AddSource(string name, string arg, string type, string trustLevel, bool isExplicit)
         {
             Utilities.VerifyAdmin();
+            string parameters = string.Empty;
+
             if (string.IsNullOrEmpty(type))
             {
-                _ = this.Run("source", $"add --name {name} --arg {arg}", 300000);
+                parameters = $"add --name {name} --arg {arg}";
```
JohnMcPMS, line 82: Not related to this change, but this code is not resilient to spaces in the arguments.

src/PowerShell/Microsoft.WinGet.Client.Engine/PSObjects/PSSourceResult.cs
```diff
@@ -39,5 +39,15 @@ internal PSSourceResult(Management.Deployment.PackageCatalogReference catalogRef
         /// Gets the type of the source.
         /// </summary>
         public string Type { get; private set; }
+
+        /// <summary>
+        /// Gets the trust level of the source.
+        /// </summary>
+        public string TrustLevel { get; private set; }
```
JohnMcPMS, line 46: Why is this a `string` and not a `PSSourceTrustLevel`?

Maybe that is normal in PowerShell, it just seems odd to me.

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddSourceCmdlet.cs
```diff
@@ -42,13 +43,42 @@ public sealed class AddSourceCmdlet : PSCmdlet
...
+            command.AddSource(this.Name, this.Argument, this.Type, this.ConvertPSSourceTrustLevelToString(this.TrustLevel), this.Explicit.ToBool());
+        }
+
+        private string ConvertPSSourceTrustLevelToString(PSSourceTrustLevel trustLevel)
+        {
+            string trustLevelString = string.Empty;
+            switch (trustLevel)
+            {
+                case PSSourceTrustLevel.None:
+                    trustLevelString = "None";
```
msftrubengu, line 74: You could do switch return with PSSourceTrusLevel.Default => string.Empty and _ being trustLevel.ToString()

src/PowerShell/Microsoft.WinGet.Client.Engine/PSObjects/PSSourceResult.cs
msftrubengu: The main reason is not load any dependency when the CLR loads this binary as the ALC hasn't been being configured yet at this point. For cswinrt enums, there are PSSomething equivalents that get converted to string and then converted to the cswinrt representation in the engine module. Since this doesn't have it he had to go with strings,

JohnMcPMS: Is it that we can't have a dependency inversion with `Engine` taking a dependency on `Cmdlets`?  And we can't have this kind of immediate dependency on `Engine` in types exposed on the interface of any cmdlet because we don't have the ALC yet to resolve it (thus preventing us from putting `PSSourceTrustLevel` in `Engine`)?

In any case, I don't have a problem with it being a string.  Just wasn't clear why.

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddSourceCmdlet.cs
JohnMcPMS: Changed it to your fancy switch pattern syntax.

src/PowerShell/Microsoft.WinGet.Client.Engine/PSObjects/PSSourceResult.cs
msftrubengu: Yeah, we don't want to have the dependency on Engine until the ALC is setup

msftrubengu (approved):

PR #4754, opened 2024-08-26
Allow for smaller configuration outputs

JohnMcPMS:
## Change
To prevent anything from outputting to our console, call `FreeConsole` in the configuration server.

To allow for output of primarily results, when `--accept-configuration-agreements` is passed, one can now also pass `--suppress-prologue` to prevent the initial details from being presented.  The only output (beyond the temporary progress spinner+text) will then be the disclaimer warning and the individual results.

## Validation
Confirmed that both changes work manually.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4754)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
@@ -1369,27 +1369,31 @@ namespace AppInstaller::CLI::Workflow
         auto getDetailsOperation = configContext.Processor().GetSetDetailsAsync(configContext.Set(), ConfigurationUnitDetailFlags::ReadOnly);
         auto unification = anon::CreateProgressCancellationUnification(std::move(progressScope), getDetailsOperation);
 
+        bool suppressDetailsOutput = context.Args.Contains(Args::Type::ConfigurationAcceptWarning) && context.Args.Contains(Args::Type::ConfigurationSuppressPrologue);
         anon::OutputHelper outputHelper{ context };
         uint32_t unitsShown = 0;
 
-        getDetailsOperation.Progress([&](const IAsyncOperationWithProgress<GetConfigurationSetDetailsResult, GetConfigurationUnitDetailsResult>& operation, const GetConfigurationUnitDetailsResult&)
-            {
-                auto threadContext = context.SetForCurrentThread();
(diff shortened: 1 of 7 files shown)


src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1369,27 +1369,31 @@ namespace AppInstaller::CLI::Workflow
         auto getDetailsOperation = configContext.Processor().GetSetDetailsAsync(configContext.Set(), ConfigurationUnitDetailFlags::ReadOnly);
         auto unification = anon::CreateProgressCancellationUnification(std::move(progressScope), getDetailsOperation);
 
+        bool suppressDetailsOutput = context.Args.Contains(Args::Type::ConfigurationAcceptWarning) && context.Args.Contains(Args::Type::ConfigurationSuppressPrologue);
```
florelis, line 1372: Should we validate that `supress-prologue` is not used on its own?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3109,4 +3109,7 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="StoreInstall_PackageNotAvailableForCurrentSystem" xml:space="preserve">
     <value>The package is not compatible with the current Windows version or platform.</value>
   </data>
+  <data name="ConfigurationSuppressPrologueArgumentDescription" xml:space="preserve">
+    <value>Suppress the configuration prologue when possible</value>
```
florelis, line 3113: I don't think this will be clear to either users or translators. Do we say "prologue" anywhere else?

florelis (dismissed):

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
JohnMcPMS: It only controls output, not the results.  I don't see it as harmful to be passed when it wouldn't do anything.

florelis (approved):

PR #4760, opened 2024-08-27
Run Unpackaged Unit Tests with winget-installed PsExec

ryfu-msft:
Currently forked builds do not have the permission to access secure files. 

To workaround using PsExec from a secure file, we will use a stable release of winget to install the latest PsTools and try invoking unit tests from the PsExec executable that is installed as part of the zip package.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -348,13 +348,23 @@ jobs:
       packageFeedSelector: 'nugetOrg'
 
   - pwsh: |
+      Add-AppxPackage "${env:ProgramFiles(x86)}\Microsoft SDKs\Windows Kits\10\ExtensionSDKs\Microsoft.VCLibs.Desktop\14.0\Appx\Retail\x64\Microsoft.VCLibs.x64.14.00.Desktop.appx"
       Install-Module Microsoft.WinGet.Client -Repository PSGallery -Force
       Repair-WingetPackageManager -AllUsers
-      winget install --id Microsoft.Sysinternals.PsTools -s winget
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -347,17 +347,23 @@ jobs:
...
+  - task: PowerShell@2
+    displayName: Install WinGet and PsExec
     inputs:
-      secureFile: 'PsExec.exe'
+      pwsh: true
+      targetType: 'inline'
+      script: |
+        Install-Module Microsoft.WinGet.Client -Repository PSGallery -Force
+        Repair-WingetPackageManager
+        winget install --id Microsoft.Sysinternals.PsTools --accept-source-agreements
```
florelis, line 358: Nit, not needed: Could also specify `-s winget` instead of accepting the agreements for msstore

azure-pipelines.yml
```diff
@@ -347,17 +347,23 @@ jobs:
...
       packageFeedSelector: 'nugetOrg'
 
-  - task: DownloadSecureFile@1
-    name: PsExec
-    displayName: 'Download PsExec.exe'
+  - task: PowerShell@2
+    displayName: Install WinGet and PsExec
     inputs:
-      secureFile: 'PsExec.exe'
+      pwsh: true
```
florelis, line 353: Nit, not needed: Could be simplified with the `pwsh` step alias

```yaml
- pwsh: |
    <script here>
  displayName: Install WinGet and PsExec
  condition: succeededOrFailed()
```

florelis (dismissed):

JohnMcPMS (approved):
Going to merge this and immediately fix the `iwr` so we don't need a new build.

PR #4791, opened 2024-09-04
PS Gallery tests are unreliable

JohnMcPMS:
## Change
These tests that rely on the PowerShell Gallery are unreliable.  Disable them.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4791)

### DIFF
--- a/src/AppInstallerCLIE2ETests/ConfigureCommand.cs
+++ b/src/AppInstallerCLIE2ETests/ConfigureCommand.cs
@@ -40,6 +40,7 @@ namespace AppInstallerCLIE2ETests
         /// Intentionally has no settings to force a failure, but only after acquiring the module.
         /// </summary>
         [Test]
+        [Ignore("PS Gallery tests are unreliable.")]
         public void ConfigureFromGallery()
         {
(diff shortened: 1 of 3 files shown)


src/AppInstallerCLIE2ETests/ConfigureCommand.cs
```diff
@@ -40,6 +40,7 @@ public void OneTimeTeardown()
         /// Intentionally has no settings to force a failure, but only after acquiring the module.
         /// </summary>
         [Test]
+        [Ignore("PS Gallery tests are unreliable.")]
```
florelis, line 43: Instead of ignoring them, is it possible to mark them as not-breaking somehow? Like changing the Assert for a warning or something.

florelis: Maybe do `Assert.Inconclusive()` if we get the specific failure we are seeing

florelis (approved):

PR #4795, opened 2024-09-05
Microsoft.WinGet.DSC improvements (and breaking change)

JohnMcPMS:
The goal of this PR is to get these resources to a state where we feel ready to release the module without the prerelease tag.  The `WinGetUserSettings`, `WinGetAdminSettings`, and `WinGetPackageManager` resources were deemed acceptable in their current state.  The `WinGetSources` resource was decided to be overly complicated and likely not heavily used yet (and containing some bugs), so it was removed in favor of a new `WinGetSource` resource for managing individual sources.  The `WinGetPackage` resource was updated to be better in line with best practices but should still be compatible unless one was using `Get`.

## Change
The `WinGetSources` resource was replaced with a `WinGetSource` resource that manages a single source.  The `WinGetSources` resource can continue to be used by specifying one of the module versions that contains it, but once released, new configurations should prefer the `WinGetSource` resource from the latest module.

While there is no way to include or discover help via the module, I documented the resources similarly to the cmdlet documentation format.  These can be used as the basis for official documentation, or we can put a link to them in the PowerShell Gallery listing (maybe updating the `Project Site` to point to a new landing page for the DSC resources).

The updates to `WinGetPackage` aim to have it follow best practices and make general improvements:
- Mark `Source` as a `Key`
- Validate with existing attributes where possible
- Not updating `$this` in calls to `Get`/`Set`/`Test`
- Using the configurable properties to report values with `Get`
  - `IsInstalled` is reported via `Ensure` (true => `Present`, false => `Absent`)
  - `InstalledVersion` is reported via `Version`
  - `IsUpdateAvailable` is reported via `UseLatest` (they will have the opposite boolean value from each other)

Any existing (non-`Get`) uses of the resource should continue to work as they did before.

Smaller updates worth mentioning:
- Updated the `winget configure validate` functionality to handle the new `WinGetSource` resource instead of `WinGetSources`.
- Removed the error when giving both a `Version` and `UseLatest` to the `WinGetPackage` resource. `UseLatest` will override any `Version` provided if it is true.
- Fixed a bug with COM `PackageCatalogInfo::TrustLevel` property reporting `None` improperly if the internal trust level contained both `Trusted` and another flag.
- Renamed all of the enums in the module as there is no namespacing for enums/classes in PowerShell.

## Validation
Manual verification of basics locally to ensure golden path functionality.
Updated existing tests to cover changes, including adding new fields.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4795)


src/AppInstallerCLICore/ConfigurationWingetDscModuleUnitValidation.cpp
```diff
@@ -206,56 +174,48 @@ namespace AppInstaller::CLI::Configuration
...
-               }
-           }
+            // Validate basic semantics.
+            if (source.Name.empty())
+            {
+                AICLI_LOG(Config, Error, << "WinGetSource unit missing required arg: Name");
+                context.Reporter.Error() << Resource::String::WinGetResourceUnitMissingRequiredArg(Utility::LocIndView{ UnitType_WinGetSource }, "Name"_liv) << std::endl;
+                foundIssues = true;
+            }
+            if (source.Arg.empty())
```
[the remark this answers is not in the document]
JohnMcPMS: This code wasn't originally written to be `Ensure` aware.  You are correct that during `Absent` only the `Name` is required.

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -193,130 +187,169 @@ class WinGetAdminSettings
...
+                $addSource = $true
+            }
+        }
+        else
+        {
+            if ($currentSource.Ensure -eq [WinGetEnsure]::Present)
+            {
+                $removeSource = $true
+            }
+            else
```
msftrubengu, line 278: I think I like more the else comment like 265, but at least both should be consistent

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -193,130 +187,169 @@ class WinGetAdminSettings
...
 
-        return $currentState -eq $this.Ensure
+            Add-WinGetSource @hashArgs
+        }
     }
-
-    # Sets the desired properties.
-    [void] Set()
+    
+    [bool] hidden TestAgainstCurrent([WinGetSource]$currentSource)
```
msftrubengu, line 320: I had to go back on why there's no comparison of Name. You don't want to call Get here because you don't want two Get-WinGetSource in Set. Maybe add a comment?

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -193,130 +187,169 @@ class WinGetAdminSettings
...
-
-    # An array of Hashtable with the key value properties that follows the source's group policy schema.
+    [DscProperty(Key, Mandatory)]
+    [ValidateNotNullOrWhiteSpace()]
+    [string]$Name
+    
     [DscProperty(Mandatory)]
-    [Hashtable[]]$Sources
-
+    [string]$Argument
```
msftrubengu, line 197: do we need `[ValidateNotNullOrWhiteSpace()]` here?

src/PowerShell/scripts/samples/WinGetPackageResourceSample.ps1
```diff
@@ -52,7 +52,7 @@ else
 {
     Write-Host "PowerToys 0.65.0 is not installed."
 
-    Invoke-DscResource @resource -Method Set
+    #Invoke-DscResource @resource -Method Set
```
msftrubengu, line 55: You don't want the sample to install PowerToys?

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
JohnMcPMS: No, because if `Ensure = Absent` we don't want to require an argument value (although from my testing it seems like the validation only happens during assignment).  If someone provides an empty argument they will still get an error, just from deeper in the stack.

src/PowerShell/scripts/samples/WinGetPackageResourceSample.ps1
JohnMcPMS: I forgot to revert that after testing the script.

msftrubengu (dismissed):

msftrubengu (approved):

Issue #4170, opened 2024-02-14
Implement source management in the COM API


denelon:
### Description of the new feature / enhancement

I'd like to be able to manage WinGet sources via the COM API.

This request will likely be associated with WinGet source management in [Dev Home](https://github.com/microsoft/devhome).

*I'll add the link to the issue when it's created*

PR #4813, opened 2024-09-16
win get source com api

Madhusudhan-MSFT:
Draft implementation of WinGet Source COM API for review.

-----


src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1351,5 +1565,15 @@ namespace Microsoft.Management.Deployment
         interface Windows.Foundation.Collections.IVectorView<MicrosoftEntraIdAuthenticationInfo>;
         interface Windows.Foundation.Collections.IVector<AuthenticationInfo>;
         interface Windows.Foundation.Collections.IVectorView<AuthenticationInfo>;
+        interface Windows.Foundation.Collections.IVector<AddPackageCatalogOptions>;
```
JohnMcPMS, line 1568: I don't think that these are needed.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1233,6 +1433,20 @@ namespace Microsoft.Management.Deployment
         /// (Installing, Installed) together at the same time.
         PackageCatalogReference CreateCompositePackageCatalog(CreateCompositePackageCatalogOptions options);
 
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 11)]
+        {
+            Windows.Foundation.IAsyncOperationWithProgress<AddPackageCatalogResult, AddPackageCatalogProgress> AddPackageCatalogAsync(AddPackageCatalogOptions options);
+
+            /// Removes a source from the Windows Package Manager source list.
```
JohnMcPMS, line 1440: We should only refer to them as catalogs within the COM API.  If there are other cases where we use source, they should be changed (unless it is explicitly a call out on the different nouns being used).

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1233,6 +1433,20 @@ namespace Microsoft.Management.Deployment
         /// (Installing, Installed) together at the same time.
         PackageCatalogReference CreateCompositePackageCatalog(CreateCompositePackageCatalogOptions options);
 
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 11)]
+        {
+            Windows.Foundation.IAsyncOperationWithProgress<AddPackageCatalogResult, AddPackageCatalogProgress> AddPackageCatalogAsync(AddPackageCatalogOptions options);
+
+            /// Removes a source from the Windows Package Manager source list.
+            Windows.Foundation.IAsyncOperationWithProgress<RemovePackageCatalogResult, RemovePackageCatalogProgress> RemovePackageCatalogAsync(String catalogName);
```
JohnMcPMS, line 1441: I would strongly consider making this (and the other non-Add methods) a member of `PackageCatalogReference`.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1214,6 +1214,206 @@ namespace Microsoft.Management.Deployment
...
     }
 
+    /// IMPLEMENTATION NOTE: AddPackageCatalogOptions
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 11)]
+    runtimeclass AddPackageCatalogOptions
+    {
+        AddPackageCatalogOptions();
+
+        /// The name of the catalog to add.
+        String CatalogName;
```
JohnMcPMS, line 1224:
```suggestion
        String Name;
```
Unless there is to be another name, it seems redundant to call out catalog again.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1214,6 +1214,206 @@ namespace Microsoft.Management.Deployment
...
+        /// The source URI of the catalog to add.
+        String SourceUri;
```
JohnMcPMS, line 1226: Technically this is an all-purpose string value that the code for the `Type` can interpret however it wants.  I get that this becomes awkward and requires more documentation on an API though.  The value will be returned through the `String Argument` property of `PackageCatalogInfo` though, so I think it makes sense to remain a string here.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1214,6 +1214,206 @@ namespace Microsoft.Management.Deployment
...
+        /// Type of the catalog to add.
+        String Type;
+        /// Additional arguments to pass to the catalog.
+        String Arguments;
```
JohnMcPMS, line 1230: I'm not sure what value this will map to; it should probably be removed.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1214,6 +1214,206 @@ namespace Microsoft.Management.Deployment
...
+        /// The trust level of the catalog to add.
+        PackageCatalogTrustLevel TrustLevel;
+        /// Custom header to pass to the catalog.
+        String CustomHeader;
+    };
```
JohnMcPMS, line 1235: Missing the `Explicit` property.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1214,6 +1214,206 @@ namespace Microsoft.Management.Deployment
...
+    runtimeclass AddPackageCatalogResult
+    {
+        AddPackageCatalogStatus Status { get; };
+
+        /// Error codes
+        HRESULT ExtendedErrorCode { get; };
+    };
+
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 11)]
+    enum AddPackageCatalogProgressState
```
JohnMcPMS, line 1258: This seems to be a copy-adapt of the package install state, but I don't think it properly represents the actual flow.  There is no queue unless we add one for COM.  The code is already multi-write resilient, so it shouldn't need a queue.

Also, we own the relevant actions, so there isn't a point at which we can't just revert state.

I don't think that this is really needed.  Discussing the larger progress type for all of these operations, I don't see a need for any of the *State enums.  We could probably just make the progress type `Double` for all of them.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1197,5 +1322,114 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            co_return GetAddPackageCatalogResult(APPINSTALLER_CLI_ERROR_INVALID_CL_ARGUMENTS);
+        }
+
+        HRESULT terminationHR = S_OK;
+        try {
+
+            // Check if running as admin
+            // [NOTE:] For OutOfProc calls, the Windows Package Manager Service executes in the context initiated by the caller process,
+            //so the same admin validation check is applicable for both InProc and OutOfProc calls.
+            THROW_HR_IF(APPINSTALLER_CLI_ERROR_COMMAND_REQUIRES_ADMIN, !AppInstaller::Runtime::IsRunningAsAdmin());
```
Madhusudhan-MSFT, line 1346: Is it required to check if the calling process should be running as an Admin for the packaged context? 

The scenario I want to confirm involves two processes, p1 and p2, both attempting to call OOP APIs. If p1, running as an admin, starts the WinGet COM Server initially, and p2, running in user context, tries to invoke PackageCatalogManagement APIs, 
1. will p2's requests be served by the WinGet COM Server running in p1's context, or 
2. will a new WinGet COM Server process be launched in p2's context to serve p2's requests?

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -289,4 +291,43 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+    winrt::Microsoft::Management::Deployment::RefreshPackageCatalogResult GetRefreshPackageCatalogResult(winrt::hresult terminationStatus)
+    {
+        winrt::Microsoft::Management::Deployment::RefreshPackageCatalogStatus status = GetPackageCatalogOperationStatus<RefreshPackageCatalogStatus>(terminationStatus);
+        auto updateResult = winrt::make_self<wil::details::module_count_wrapper<winrt::Microsoft::Management::Deployment::implementation::RefreshPackageCatalogResult>>();
+        updateResult->Initialize(status, terminationStatus);
+        return *updateResult;
+    }
+
+    winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::RefreshPackageCatalogResult, double> PackageCatalogReference::RefreshPackageCatalogAsync()
+    {
```
Madhusudhan-MSFT, line 304: Do we need to explicitly call `::AppInstaller::Logging::Telemetry().LogStartup(true);` here? 

Usually, a caller reaches this point after obtaining a PackageCatalogReference instance by calling one of the `PackageManager.*PackageCatalog*` methods, which in turn call `LogStartupIfApplicable(...)` to initialize the logs. Can we safely assume that the COM logging instance is already initialized before the caller attempts this operation?

yao-msft: Not at the moment. I don't think we have specific telemetry events for source operations.

src/Microsoft.Management.Deployment/PackageManager.cpp
JohnMcPMS, line 1346: This might need to allow for SYSTEM callers too.

src/Microsoft.Management.Deployment/ResetPackageCatalogResult.cpp
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
```
JohnMcPMS, line 1: Seems like leftover files that should be removed?

src/Microsoft.Management.Deployment/PackageManager.cpp
JohnMcPMS: As discussed, unelevated processes should use standard COM activation and end up in an unelevated server.  Elevated processes use manual activation and end up in an elevated server.  The elevated server does not allow unelevated clients to connect.

src/Microsoft.Management.Deployment/PackageManager.idl
Madhusudhan-MSFT: We decided to keep it part of PackageManger class in the API review.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1197,5 +1322,114 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
     }
 
+    winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::AddPackageCatalogResult, double> PackageManager::AddPackageCatalogAsync(winrt::Microsoft::Management::Deployment::AddPackageCatalogOptions options)
+    {
+        LogStartupIfApplicable();
+
+        // options must be set.
+        if (!options)
+        {
+            co_return GetAddPackageCatalogResult(APPINSTALLER_CLI_ERROR_INVALID_CL_ARGUMENTS);
```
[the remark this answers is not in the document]
Madhusudhan-MSFT: Returning an object with an error code is the pattern used by other APIs in PackageManager. For example, WINGET_RETURN_DOWNLOAD_RESULT_HR_IF(APPINSTALLER_CLI_ERROR_INVALID_CL_ARGUMENTS, !package) appears to throw an error code directly, but it actually creates an object with an error status behind the scenes.

I can implement the suggested change, but it might not be consistent with the existing pattern used by other APIs.

src/Microsoft.Management.Deployment/AddPackageCatalogOptions.h
```diff
@@ -0,0 +1,55 @@
...
+
+        bool Explicit();
+        void Explicit(bool const& value);
+
+        bool AcceptSourceAgreements();
+        void AcceptSourceAgreements(bool const& value);
+
+#if !defined(INCLUDE_ONLY_INTERFACE_METHODS)
+    private:
+        std::wstring m_name = L"";
```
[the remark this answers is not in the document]
Madhusudhan-MSFT: The pattern of using std::wstring for private fields storing strings is consistent across various APIs  *InputOptions runtimeclass implementation. However, the set and get methods use hstring. 

I can change it to hstring, but it may look inconsistent with the existing code pattern.

src/Microsoft.Management.Deployment/Converters.h
```diff
@@ -141,4 +141,95 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        case APPINSTALLER_CLI_ERROR_SOURCE_DATA_INTEGRITY_FAILURE:
+            return TStatus::CatalogError;
+        case APPINSTALLER_CLI_ERROR_INTERNAL_ERROR:
+        default:
+            return TStatus::InternalError;
+        }
+    }
+
+    template <typename TStatus>
+    TStatus GetAddPackageCatalogOperationStatus(winrt::hresult hresult)
```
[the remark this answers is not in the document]
Madhusudhan-MSFT: I agree with your comments that GetAddPackageCatalogOperationStatus and GetRemovePackageCatalogOperationStatus are only called with AddPackageCatalogStatus and RemovePackageCatalogStatus, respectively. Initially, I considered not using template methods for these functions, but implementing them in the header caused multiple linker errors due to symbol conflicts for GetAddPackageCatalogOperationStatus and GetRemovePackageCatalogOperationStatus. Therefore, I kept them as template functions. The right approach here is to move the non-template functions to converter.cpp, declare them here, and use templates for common error type methods.

src/Microsoft.Management.Deployment/PackageManager.cpp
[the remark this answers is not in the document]
Madhusudhan-MSFT: To simplify things, I'm considering moving this logic inside a try block and using THROW_HR_* constructs to throw invalid arguments. This way, the result object construction will be automatically handled at the point of return status in the following catch block.

src/Microsoft.Management.Deployment/Converters.h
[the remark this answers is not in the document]
JohnMcPMS: You didn't mark them as `inline`, which is required if implementing in a header.  But I would also have suggested to not implement them in the header.

src/Microsoft.Management.Deployment/PackageManager.cpp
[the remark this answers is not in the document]
Madhusudhan-MSFT: I have implemented the suggested PR feedback after reconsidering. Thanks, John, for explaining how this change aligns with the API Design guidelines and why it is beneficial.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    CompletionOnlyProgressSink::CompletionOnlyProgressSink(std::function<void(double)> progressReporter) :
+        m_progressReporter(progressReporter)
+    {
+    }
+
+    void CompletionOnlyProgressSink::OnProgress(uint64_t current, uint64_t maximum, AppInstaller::ProgressType type)
+    {
+        UNREFERENCED_PARAMETER(current);
```
JohnMcPMS, line 53: We prefer to leave the parameters unnamed rather than use `UNREFERENCED_PARAMETER`.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
```diff
@@ -0,0 +1,137 @@
...
+        for (const auto& weight : m_progressWeights)
+        {
+            if (weight.first != AppInstaller::ProgressType::None)
+            {
+                totalWeight += weight.second;
+            }
+        }
+
+        // If the total weight is greater than 1, throw an exception.
+        if (totalWeight != 1.0)
```
JohnMcPMS, line 94: I might be missing something, but it seems like the remove operation code will hit this check based on only setting bytes to 0 in the factory.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
```diff
@@ -0,0 +1,137 @@
...
+        double progress = static_cast<double>(current) / maximum;
+        m_progressValues[type] = progress;
+
+        double totalProgress = 0.0;
+
+        // Calculate the total progress.
+        for (const auto& [progressType, weight] : m_progressWeights)
+        {
+            // Adjust the total progress value based on the weight.
+            totalProgress += m_progressValues[progressType] * weight;
```
JohnMcPMS, line 116: This algorithm doesn't work right if one of the types never reports any progress (or doesn't report 100% as its last value).  What might be better (but a bit more complicated) is an ordered set of expected types and weights.  Then when calculating the overall value, the entire weight of any previous step is used rather than the reported value.

For this case, it would be similar except if we have seen any percent based progress we would always use the byte based progress weight value:
```
// New type for m_progressWeights to have an order
std::vector<std::pair<type, weight>> m_progressWeights;

doublt totalWeight = 0, calculatedProgress = 0;

for (const auto& [type, weight] : m_progressWeights)
{
    double typeWeight = m_progressValues[type];
    if (typeWeight != 0)
    {
        calculatedProgress = totalWeight;
    }
    totalWeight += weight;
    calculatedProgress += typeWeight * weight;
}
```
If we needed to support multiples of the same type it would be way more complicated.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    void PreIndexedPackageCatalogProgressSink::SetProgressMessage(std::string_view message)
+    {
+        UNREFERENCED_PARAMETER(message);
+    }
+
+    void PreIndexedPackageCatalogProgressSink::BeginProgress()
+    {
+        OnProgress(0, 1, AppInstaller::ProgressType::None);
```
JohnMcPMS, line 129: This is a no-op due to immediate return for type `None`.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -52,6 +57,113 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+                source.SetCustomHeader(customHeader);
+            }
+
+            try
+            {
+                auto sourceInfo = source.GetInformation();
+
+                if (sourceInfo.Authentication.Type == ::AppInstaller::Authentication::AuthenticationType::Unknown)
+                {
+                    throw winrt::hresult_error(APPINSTALLER_CLI_ERROR_AUTHENTICATION_TYPE_NOT_SUPPORTED);
```
JohnMcPMS, line 123:
```suggestion
                    THROW_HR(APPINSTALLER_CLI_ERROR_AUTHENTICATION_TYPE_NOT_SUPPORTED);
```
Always use macros for throwing.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -52,6 +57,113 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            }
+            catch (const winrt::hresult_error& hre)
+            {
+                if (hre.code() == APPINSTALLER_CLI_ERROR_AUTHENTICATION_TYPE_NOT_SUPPORTED)
+                {
+                    THROW_HR(APPINSTALLER_CLI_ERROR_AUTHENTICATION_TYPE_NOT_SUPPORTED);
+                }
+                else
+                {
+                    THROW_HR(APPINSTALLER_CLI_ERROR_SOURCE_OPEN_FAILED);
```
JohnMcPMS, line 134: Don't catch and normalize this way, it loses the actual HR value that we want exposed to the caller.

This really applies to the entire catch/throw you have here.  If the helper for extracting the HR from an exception doesn't provide a "use this HR for unknown", add a new parameter (with a default of whatever it uses internally now) that can do that.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1197,5 +1309,91 @@ namespace winrt::Microsoft::Management::Deployment::implementation
             true /*canCancelQueueItem*/, nullptr /*queueItem*/, package, options, std::move(callerProcessInfoString));
     }
 
+    winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::AddPackageCatalogResult, double> PackageManager::AddPackageCatalogAsync(winrt::Microsoft::Management::Deployment::AddPackageCatalogOptions options)
+    {
+        LogStartupIfApplicable();
+
+        // options must be set.
+        THROW_HR_IF_NULL(E_INVALIDARG, options);
```
JohnMcPMS, line 1317:
```suggestion
        THROW_HR_IF_NULL(E_POINTER, options);
```
From the same guidelines document I sent earlier, this is the correct HR for null parameters that should not be null.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1197,5 +1309,91 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+
+            packageCatalogProgressSink->BeginProgress();
+            ::AppInstaller::Repository::Source sourceToRemove = ::AppInstaller::Repository::Source{ matchingSource.value().Name };
+            ::AppInstaller::ProgressCallback progress(packageCatalogProgressSink.get());
+
+            // If the PreserveData option is set, this is equivalent to the WinGet CLI Reset command on a single source; otherwise, it removes the source.
+            if (options.PreserveData())
+            {
+                THROW_HR_IF(APPINSTALLER_CLI_ERROR_SOURCE_NAME_DOES_NOT_EXIST, !sourceToRemove.DropSource(matchingSource.value().Name));
+                packageCatalogProgressSink->OnProgress(100, 100, AppInstaller::ProgressType::Percent);
```
JohnMcPMS, line 1382: This feels like what `EndProgress` should effectively be doing all the time.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1370,6 +1492,15 @@ namespace Microsoft.Management.Deployment
         /// (Installing, Installed) together at the same time.
         PackageCatalogReference CreateCompositePackageCatalog(CreateCompositePackageCatalogOptions options);
 
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 12)]
+        {
+            /// Add a catalog to the Windows Package Catalogs.
```
JohnMcPMS, line 1497: Comment what the progress double value means here.  Include the range of values that should be expected.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
Madhusudhan-MSFT: I made a mistake in the logic. My intention was to include only the percentage progress type with a weight of 1.0 for the remove operation, but I incorrectly set it to 0. It is addressed now.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
Madhusudhan-MSFT: Good catch! I assumed the underlying operation code would always report progress until 100% completion. There's a potential gap when the initial operations complete successfully but stop reporting before full completion. It took a bit of time to understand this edge case, so I added an inline comment to explain why additional computation logic is necessary when applicable.

src/Microsoft.Management.Deployment/PackageManager.cpp
Madhusudhan-MSFT: For the sake of discussion, I've noticed that the E_POINTER error code translates to NullReferenceException when it passes through the project layer, but it should be ArgumentNullException.

According to the .NET MSDN documentation for both [NullReferenceException](https://learn.microsoft.com/en-us/dotnet/api/system.nullreferenceexception?view=net-8.0) and [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-8.0&form=MG0AV3)
both exceptions use the same HRESULT value 0x80004003 but in different contexts. It seems that the winrt projection layer might not be translating this correctly, and it should be ArgumentNullException instead of NullReferenceException.

JohnMcPMS: CsWinRT is responsible for translating the ABI layer HRESULT into an appropriate projection in C#.  There isn't really a way to know the difference between the two cases though, so I assume they chose the more generic (and still true) `NullReferenceException`.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -52,6 +57,95 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+
+            std::string customHeader = winrt::to_string(options.CustomHeader());
+            if (!customHeader.empty())
+            {
+                source.SetCustomHeader(customHeader);
+            }
+
+            auto sourceInfo = source.GetInformation();
+
+            if (sourceInfo.Authentication.Type == ::AppInstaller::Authentication::AuthenticationType::Unknown)
```
Madhusudhan-MSFT, line 119: @yao-msft - This check seems particularly relevant to the REST Source scenario. Do you think it's needed for the Add operations? I'm trying to confirm whether this is a redundancy check or if it becomes relevant when the caller invokes the PackageManagerReference.Connect API call. If it's redundant, I'll go ahead and remove this check and the associated error result, as we no longer encounter that scenario during the add call.

yao-msft: I think we can keep it. If a source is not supported by the client, we'd better return an error instead of adding it because all subsequent connect calls will fail in the future.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -940,6 +940,11 @@ namespace Microsoft.Management.Deployment
         [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 12)]
         {
             /// Updates the package catalog.
+            /// The progress value, represented as a double, indicates the percentage of operation completion.
+            /// For "Microsoft.PreIndexed.Package" Package Catalogs, it reflects the progress of downloading and
+            /// installing the index. When there is no meaningful progress to report for other Package Catalogs,
+            /// the progress will be set to 0% at the start and will be updated to 100% upon the operation's completion.
+            /// The progress range is from 0 to 100.
```
JohnMcPMS, line 947:
```suggestion
            /// The progress value, represented as a double, indicates the percentage of operation completion.
            /// The progress range is from 0 to 100.
```
I meant only for you to document the visible portion (range and meaning).  We should not document specifics on any of the internal implementation details.

This applies to the other progress comments as well.

JohnMcPMS (approved):

Madhusudhan-MSFT:
Related to:

#4170

PR #4828, opened 2024-09-23
Sixel support

JohnMcPMS:
## Hackathon
As part of a yearly event to work on personal choice projects, I implemented support for rendering [sixels](https://en.wikipedia.org/wiki/Sixel) in a few different scenarios.  You can try it out in any terminal that supports sixels, such as the latest [Windows Terminal Preview build](https://github.com/microsoft/terminal/releases/tag/v1.22.2362.0).  Both @lhecker and @j4james were very helpful with guidance on the nuances of sixels.

## Change
The foundational work is a wrapper around the [Windows Imaging Component (WIC) APIs](https://learn.microsoft.com/en-us/windows/win32/api/_wic/) and a renderer for converting the indexed images to sixel format.  WIC does all the heavy lifting of decoding images, palette optimization and dithering.  My additions are a simple compositor and sixel renderer.  There is also some new code for detecting VT extension support, including only enabling sixel rendering if the terminal advertises the extension.

Sixel use was added in various ways.  The winget icon is output any time the header would be (during `winget --info` and any time we output help):

winget-info

The package icon (if present) is displayed during `winget show`.  This is an example using a local manifest updated to point to an image located on the github repository of the package; actual package icon will likely be different.  Package icons extracted from actual installs are expected to arrive soon:

winget-show

A new progress visualization is available using sixels.  The indefinite progress (the current character spinner that iterates through various slashes, dash and pipe) looks like:

winget-indefinite

while the progress bar is a conveyor belt bringing your package to you:

winget-progress
_The conveyor belt base image is only the finest of dev art; it is very open to someone with more artistic skill to improve upon._

The tearing present in this gif does exist, but it is exaggerated by the capture and the slow steady progress.  Actual progress is much less steady, making any tearing that may occur far less noticeable.

## Settings
Use of sixels (including even the attempt to detect support in the terminal) is gated behind enabling them in the settings.  Two new/updated settings are available.

- `.visual.enableSixels` (`true` or `false`)
  - This controls the winget icon during help and the package icon during show.
- `.visual.progressBar`
  - `"sixel"` causes progress to use the sixel versions shown above.
  - `"disabled"` disables progress output.

## Validation
Many, many manual runs of various scenarios using the added `debug` commands.
Added a few unit tests to cover primary paths.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4828)


src/AppInstallerCLICore/Commands/DebugCommand.cpp
```diff
@@ -148,6 +154,186 @@ namespace AppInstaller::CLI
...
+            Argument{ "aspect-ratio", 'a', Args::Type::AcceptPackageAgreements, Resource::String::SourceListUpdatedNever, ArgumentType::Standard },
+            Argument{ "transparent", 't', Args::Type::AcceptSourceAgreements, Resource::String::SourceListUpdatedNever, ArgumentType::Flag },
+            Argument{ "color-count", 'c', Args::Type::ConfigurationAcceptWarning, Resource::String::SourceListUpdatedNever, ArgumentType::Standard },
+            Argument{ "width", 'w', Args::Type::AdminSettingEnable, Resource::String::SourceListUpdatedNever, ArgumentType::Standard },
+            Argument{ "height", 'h', Args::Type::AllowReboot, Resource::String::SourceListUpdatedNever, ArgumentType::Standard },
+            Argument{ "stretch", 's', Args::Type::AllVersions, Resource::String::SourceListUpdatedNever, ArgumentType::Flag },
+            Argument{ "repeat", 'r', Args::Type::Name, Resource::String::SourceListUpdatedNever, ArgumentType::Flag },
+            Argument{ "out-file", 'o', Args::Type::BlockingPin, Resource::String::SourceListUpdatedNever, ArgumentType::Standard },
```
florelis, line 169: Nit: Probably not too important for a debug command, but the arg types and strings just feel wrong...

JohnMcPMS: Yes, the goal was to not introduce any new arguments just for debugging purposes.  Rather than trying to find good type name fits, I just let autocomplete decide for me.  I then cheekily chose the string "Never" for the descriptions.

If the arbitrary type names are too much, I could define macros to make it easier to read.

JohnMcPMS: And that is what I did.

src/AppInstallerCLICore/VTSupport.cpp
```diff
@@ -17,92 +17,198 @@ namespace AppInstaller::CLI::VirtualTerminal
...
         {
-            LOG_LAST_ERROR_IF(!SetConsoleMode(GetStdHandle(STD_OUTPUT_HANDLE), m_previousMode));
+            LOG_LAST_ERROR_IF(!SetConsoleMode(GetStdHandle(m_handle), m_previousMode));
             m_token = false;
         }
     }
 
+    ConsoleModeRestore::ConsoleModeRestore() : ConsoleModeRestoreBase(STD_OUTPUT_HANDLE)
+    {
+        m_token = InitializeMode(STD_OUTPUT_HANDLE, m_previousMode, { ENABLE_VIRTUAL_TERMINAL_PROCESSING | DISABLE_NEWLINE_AUTO_RETURN, ENABLE_VIRTUAL_TERMINAL_PROCESSING });
```
lhecker, line 105: FWIW, I'd personally always recommend against using the old mode as a basis for setting the new mode. After all, why should a previous console application be allowed to determine what modes you want?

src/AppInstallerCLICore/VTSupport.cpp
```diff
@@ -17,92 +17,198 @@ namespace AppInstaller::CLI::VirtualTerminal
...
+                        }
                     }
                 }
             }
+
+            return false;
+        }
+
+        // Extracts a VT sequence, expected one of the form ESCAPE + prefix + result + suffix, returning the result part.
+        std::string ExtractSequence(std::istream& inStream, std::string_view prefix, std::string_view suffix)
```
lhecker, line 62: If you assume that your suffix is a single character you could use `inStream.getline()` to do all the seeking for you automatically. That would also avoid reading past the end of the DA1 response if you believe that may be a concern.

Personally, at least in the beginning, I'd not care about consuming more input than needed or handling situations where the sequence isn't close to the start of the input buffer, so I'd do something like this:
```cpp
char buf[1024];
const auto len = inStream.readsome(&buf[0], std::size(buf));

std::string_view response{ &buf[0], std::max(0, len) };
const auto beg = response.find('\x1b');
const auto end = response.find('c', beg);

std::string result;
if (beg < end) {
    result.assign(response, beg, end - beg);
}

return result;
```

src/AppInstallerCLICore/VTSupport.cpp
```diff
@@ -17,92 +17,198 @@ namespace AppInstaller::CLI::VirtualTerminal
...
 #define AICLI_VT_CSI        AICLI_VT_ESCAPE "["
 
 // The beginning of an Operating system command
 #define AICLI_VT_OSC        AICLI_VT_ESCAPE "]"
 
+    PrimaryDeviceAttributes::PrimaryDeviceAttributes(std::ostream& outStream, std::istream& inStream)
+    {
+        try
+        {
+            ConsoleInputModeRestore inputMode;
```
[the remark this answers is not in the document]
JohnMcPMS: While there are some mode flags that could probably be set for the entire process, disabling echo input is one that definitely needs to be done only while we receive the VT response.  We do have some cases where we prompt the user and I prefer the echo there.

src/AppInstallerCLICore/ExecutionProgress.cpp
```diff
@@ -151,289 +202,626 @@ namespace AppInstaller::CLI::Execution
...
+        SixelIndefiniteSpinner(BaseStream& stream, bool enableVT) :
+            ProgressVisualizerBase(stream, enableVT)
+        {
+            Sixel::RenderControls& renderControls = m_compositor.Controls();
+            renderControls.RenderSizeInCells(2, 1);
+
+            // Create palette from full image
+            std::filesystem::path imageAssetsRoot = Runtime::GetPathTo(Runtime::PathName::ImageAssets);
+
+            // This image matches the target pixel size. If changing the target size, choose the most appropriate image.
```
yao-msft, line 515: We should check if imageAssetsRoot is empty before using it?

src/AppInstallerCLICore/ExecutionProgress.cpp
```diff
@@ -151,289 +202,626 @@ namespace AppInstaller::CLI::Execution
...
-            size_t blocksOn = static_cast<size_t>(std::floor(percentage * blockWidth));
+            Sixel::RenderControls imageRenderControls;
+            imageRenderControls.RenderSizeInCells(2, 1);
 
-            for (size_t i = 0; i < blocksOn; ++i)
-            {
-                m_out << blockOn;
-            }
+            // This image matches the target pixel size. If changing the target size, choose the most appropriate image.
+            std::filesystem::path imageAssetsRoot = Runtime::GetPathTo(Runtime::PathName::ImageAssets);
```
yao-msft, line 636: Check empty path here too?

src/AppInstallerCLICore/Command.cpp
```diff
@@ -42,8 +43,35 @@ namespace AppInstaller::CLI
 
     void Command::OutputIntroHeader(Execution::Reporter& reporter) const
     {
+        auto infoOut = reporter.Info();
+        VirtualTerminal::ConstructedSequence indent;
+
+        if (reporter.SixelsEnabled())
```
yao-msft, line 49: Do we want to try CATCH_LOG since I saw a bunch of THROW_IF_FAILED in sixels.cpp code?

yao-msft (dismissed):

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -311,8 +313,13 @@ namespace AppInstaller::Runtime
             result = GetPathDetailsCommon(path, forDisplay);
             break;
         case PathName::SelfPackageRoot:
+        case PathName::ImageAssets:
             result.Path = GetPackagePath();
             result.Create = false;
+            if (path == PathName::ImageAssets)
+            {
+                result.Path /= (IsReleaseBuild() ? s_ImageAssetsDirectoryRelativeRelease : s_ImageAssetsDirectoryRelativePreview);
```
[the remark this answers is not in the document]
JohnMcPMS: Because I wasn't thinking through it.  It should be different paths for the different repositories.

yao-msft (approved):

SpecterShell:
The progress bar could be a nyancat.

Issue #166, opened 2020-05-15
Support for installation of fonts


sob:
# Description of the new feature/enhancement

As part of a new system build, I typically install a handful of fonts including Cascadia Code and other powerline enabled fonts. These fonts don't have an installer associated with them but instead need to be manually downloaded, extracted, and installed.

https://github.com/microsoft/cascadia-code/releases

Related to:
* https://github.com/microsoft/winget-pkgs/issues/262004

TravisSpomer:
Alternately, maybe just installing a font manager like [FontBase](https://fontba.se) or [SkyFonts](https://www.monotype.com/products/skyfonts) would work just as well without adding specific font-related features to winget. You'd have to sign into those though.

RobCannon:
Be sure to enable installing a font as a non-admin user!

BatmanAoD:
Installing a *CLI* font manager might be acceptable.

weitzhandler:
Related: https://github.com/JetBrains/JetBrainsMono/issues/260

WSLUser:
Neither are needed. All that's needed is for the font to be uninstalled first, then the updated version is installed after. Some apps or stuff in Windows will need a font-fallback during this process if they're configured to use that font. So winget just needs to support the unattended uninstallation and installation of the font.

ckarcz:
font support is not needed. install / uninstall script execution would not only allow the package to install/uninstall fonts, but install standalone apps (utilities) from zip, rar, 7zip, etc etc... even configure windows further than just fonts. we already trust that apps installed from the microsoft winget package repo are trusted so i don't see an issue with arbitrary scripts executing. they will be vetted when submitted to the MS pkg repo.

denelon:
Adding the link to [Cascadia Code](https://github.com/microsoft/cascadia-code/releases).

BEEDELLROKEJULIANLOCKHART:
Because the Microsoft Store is able to manage this already, implementation of this feature is important for compatibility with it. This was the purpose of the addition of the "msstore" repository also.

denelon:
Hey everyone [oh-my-posh](https://ohmyposh.dev/docs/) by @JanDeDobbeleer now has support for [installing nerd fonts](https://ohmyposh.dev/docs/installation/fonts#installation).

musm:
[...]

PR #4842, opened 2024-10-01
Font Spec

ryfu-msft:
Related to #166 

Adds the initial font specification document. This outlines the planned implementation for supporting fonts. This will be updated to match the actual implementation.

### DIFF
--- a/doc/specs/#166 - Font Support.md
+++ b/doc/specs/#166 - Font Support.md
@@ -69,6 +69,7 @@ To remove a font, the inverse operation of the steps above need to be completed.
 Manifest Example:
 ```yaml
 PackageIdentifier: Microsoft.CascadiaCode
+PackageName: Cascadia
 PackageVersion: 2404.23
 Installers:
 - InstallerType: zip
@@ -76,21 +77,48 @@ Installers:
   InstallerSha256: a911410626c0e09d03fa3fdda827188fda96607df50fecc3c5fee5906e33251b
   NestedInstallerFiles:
   - RelativeFilePath: ttf/CascadiaCodeItalic.ttf
-    FontName: Cascadia Code Italic
   - RelativeFilePath: ttf/CascadiaMonoNFItalic.ttf
-    FontName: Cascadia Mono NF Italic
 ManifestType: installer
 ManifestVersion: 1.9.0
 ```
 
+A new `PredefinedSource` will be created for `Fonts`.
+
+`PredefinedInstalledFontFactory` will be defined that creates a SQLite index from the installed font information.
+
+The table below shows how a font will be recorded in the index and highlights some of the similaries to an MSIX package.
+
+| Property | Package | Font |
+| -------- | ------- | ---- |
+| Id    |  MSIX\NotepadPlusPlus_1.0.0.0_neutral__7njy0v32s6xk6    |  Machine\Arial\Bold  |
+| Version    |  1.0.0.0  |  7.01  |
+| Name    |  NotepadPlusPlus   |  Arial Bold  |
+| PackageFamilyName    |  NotepadPlusPlus_1.0.0.0_neutral__7njy0v32s6xk6    |  Arial  |
+| InstalledLocation    |   N/A    | %LOCALAPPDATA%/Microsoft/Windows/Fonts/Arial.ttf |
+| InstalledScope    |       |   User |
+
+
+
+> The unique identifer for each font will be a combination of the scope, font font family name, and font face name.
+
+> The version of the font only exists at the font face level and not at the family level
 
 ## CLI Behavior / Implementation Details
 
-### `winget list --fonts`
-Fonts will not be shown by default to not interfere with listing packages. In order to list out all installed fonts, the user must include the `--fonts` argument. Only font families will be displayed to the user if no query is specified.
+New Command: `winget fonts`
+---
+Fonts should be completely separate from the existing package management experience. Fonts aren't as complex as installing packages so having new subcommands can help simplify the necessary arguments for proper functionality. To support this, a new command called `font` will be added. The following subcommands will be available.
+- `winget fonts list`
+- `winget fonts install`
+- `winget fonts uninstall`
+- `winget fonts upgrade`
+- `winget fonts show`
 
+`winget fonts list`
+---
+The default behavior of `winget fonts list` is to display all installed font families and the number of faces that belong to each font family.
 
-| Family Name | Font Count |
+| Family Name | Face Count |
 | -------- | ------- |
 | Arial    |   6    |
 | Bahnschrift | 1   |
@@ -139,18 +167,18 @@ The code snippet below shows how to enumerate the list of installed font family 
 
 If a user specifies a specific name such as : 
 
-`winget list --fonts --font-name 'Arial'`
+`winget fonts list --name 'Arial'`
 
 Then the tool will display all of the font faces related to that font family name along with the corresponding font file path.
 
-| Face Names | Font Family | Font File Path |
-| -------- | ------- | ------- |
-| Regular    |   Arial    | C:\Windows\Fonts\arial.ttf |
-| Narrow | Arial   | C:\Windows\Fonts\ARIALN.TTF |
-| Italic | Arial | C:\Windows\Fonts\ariali.ttf |
-| Narrow Italic | Arial | C:\Windows\Fonts\ARIALNI.TTF |
-| Narrow Bold | Arial | C:\Windows\Fonts\ARIALNB.TTF |
-| Narrow Bold Oblique | Arial | C:\Windows\Fonts\ARIALNB.TTF |
+| Face Name | Face Version | Font Family | Font File Path |
+| -------- | ------- | ------- | ------- |
+| Regular    | 7.01 |  Arial    | C:\Windows\Fonts\arial.ttf |
+| Narrow | 7.01 | Arial   | C:\Windows\Fonts\ARIALN.TTF |
+| Italic | 7.01 | Arial | C:\Windows\Fonts\ariali.ttf |
+| Narrow Italic | 7.01 | Arial | C:\Windows\Fonts\ARIALNI.TTF |
+| Narrow Bold | 7.01 | Arial | C:\Windows\Fonts\ARIALNB.TTF |
+| Narrow Bold Oblique | 7.01 | Arial | C:\Windows\Fonts\ARIALNB.TTF |
 
 > Note that some font faces share the same font file path. This will need to be handled properly during uninstall.
 
@@ -199,7 +227,8 @@ Sample Code:
         }
 ```
 
-### `winget install font`
+`winget fonts install`
+---
 
 #### 1. Initial Validation
 Winget will check that `effectiveInstallerType == font` before kicking off the font installation flow.
@@ -235,7 +264,8 @@ For all the entries above, we will check whether those files/entries already exi
 #### 3. Applying desired state
 Once all checks have been verified, we will apply the desired state so that the font file will exist in the appropriate location and a new registry entry will be created.
 
-### `winget uninstall font`
+`winget fonts uninstall`
+---
 
 1. Determine which font file(s) matches the font family/face name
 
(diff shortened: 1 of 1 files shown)


doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+        wchar_t name[512];
+        const auto fontCount = family->GetFontCount();
+        THROW_HR_IF(E_OUTOFMEMORY, nameLength > ARRAYSIZE(name));
+        THROW_IF_FAILED(names->GetString(0, &name[0], nameLength));
+    }
+```
+
+If a user specifies a specific name such as : 
+
+`winget fonts list --name 'Arial'`
```
mdanish-kh, line 164: Will this support a substring match like `winget list`? Can I do something like `winget fonts list --name 'Nerd Font'` (display all nerd fonts installed on my system) or `winget fonts list 'Italic'` (list out all italic fonts)?

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+### Uninstalling Fonts
+To remove a font, the inverse operation of the steps above need to be completed. 
+1. Remove the registry key entry for that specific font file
+2. Remove the file from the font directory
+
+
+## Manifest Changes
+- Addition of `font` to `installerType` and `nestedInstallerType`
+
+Manifest Example:
```
mdanish-kh, line 68: Is this example missing `NestedInstallerType: font`?

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+| Property | Package | Font |
+| -------- | ------- | ---- |
+| Id    |  MSIX\NotepadPlusPlus_1.0.0.0_neutral__7njy0v32s6xk6    |  Machine\Arial\Bold  |
+| Version    |  1.0.0.0  |  7.01  |
+| Name    |  NotepadPlusPlus   |  Arial Bold  |
+| PackageFamilyName    |  NotepadPlusPlus_1.0.0.0_neutral__7njy0v32s6xk6    |  Arial  |
+| InstalledLocation    |   N/A    | %LOCALAPPDATA%/Microsoft/Windows/Fonts/Arial.ttf |
+| InstalledScope    |       |   User |
+
+> The unique identifer for each font will be a combination of the scope, font font family name, and font face name.
```
mdanish-kh, line 96:
```suggestion
> The unique identifer for each font will be a combination of the scope, font family name, and font face name.
```

typo

doc/specs/#166 - Font Support.md
ryfu-msft: Forgot to add the one thing that I needed 😅

denelon:
From a PowerShell point of view are we thinking something like: WinGetFont as the Object type as opposed to a WinGetPackage?

JohnMcPMS:
> From a PowerShell point of view are we thinking something like: WinGetFont as the Object type as opposed to a WinGetPackage?

I would prefer not to have to create an entire COM and PowerShell interface that is 90% the same with different names.  That does mean a slant mapping of faces to versions.  But version can be used to describe different flavors of things, and faces of fonts are just versions of the primary theme.  So it still fits, somewhat.

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+For [#66](https://github.com/microsoft/winget-cli/issues/166)
+
+## Abstract
+This spec outlines the design for supporting the installation of fonts.
+
+## Context
+
+### Font File Types
+
+This feature will support font files with the following file types:
```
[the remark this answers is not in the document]
ryfu-msft: This is just for general knowledge, I don't plan to do anything different based on file type. Installation support will be based on the results from [IdWriteFontFile::Analyze](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritefontfile-analyze)

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+
+If a font family contains multiple font faces, the user must include the `--remove-multiple-fonts`. Otherwise, a blocking warning will be shown to the user.
+
+3. **Uninstalling a specific font face:**
+
+`winget fonts uninstall --name 'Bahnschrift' --font-face 'Semibold'`
+
+4. **Uninstalling a font face that shares the same font file as another font face:**
+
+`winget fonts uninstall --name 'Arial' --font-face 'Bold'`
```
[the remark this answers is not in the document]
ryfu-msft: I realized that using a font family that includes the word 'face' was slightly confusing... 

I've updated the arguments to more accurately reflect the difference between `--family-name` and `--face-name`

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+                wprintf(L"%s\n", &path[0]); 
+            }
+        }
+```
+
+`winget fonts install`
+---
+
+### 1. Initial Validation
+Winget will check that `effectiveInstallerType == font` before kicking off the font installation flow.
```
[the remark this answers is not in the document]
ryfu-msft: Updated spec to align with this. Made sure to call out that the workflow should not change post source selection.

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,416 @@
...
+- .otf (Open Type Font)
+
+The following font types to my knowledge are not supported:
+- .woff & .woff2 (developed by Google for the modern browser)
+- .eot (Embedded OpenType font file) Not used because of security issues.
+
+> Font types will not be treated differently. Support for font installation will be determined by calling [IDWriteFontFile::Analyze](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritefontfile-analyze).
+
+### Installing Fonts
+Typically, a user would download the font file and drag it to the `C:\Windows\Fonts` directory in the explorer view, but there are a couple things happening behind the scenes.
```
Trenly, line 31: Does the drag and drop flow trigger the same endpoints as double clicking the file in File Explorer? Just trying to understand if all the current methods of installing fonts end up at the same place

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,416 @@
...
+- [IDWriteFontList::GetFont](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritefontlist-getfont) - Gets the font from a family with a zero-based index
+- [IDWriteFont::GetInformationalStrings method](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritefont-getinformationalstrings) - Gets specific informational strings such as font full name, author, etc.
+- [IDWriteLocalFontFileLoader::GetFilePathFromKey](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritelocalfontfileloader-getfilepathfromkey) - Obtains the absolute font file path from a font file reference key.
+- [SHGetPropertyStoreFromParsingName](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-shgetpropertystorefromparsingname) - Returns the property store for a given path to obtain details like the file's title.
+
+
+## Manifest Changes
+- Addition of `font` to `installerType` and `nestedInstallerType`
```
Trenly, line 79: Is there any way for the manifest to specify the font name and not just the path to it? Thinking of a use case where the file is something like `ARandomFont.ttf` when it the name of the font is actually something like `Really Awesome Font`

doc/specs/#166 - Font Support.md
Trenly: Or also `--version`, `--exact`, etc.

ryfu-msft: I am thinking of creating an index in memory that we can use to better search the installed fonts like the examples you have above.

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+Once we have determined the matching font file and registry key entry, both of these items will be removed from the system and the font index will be updated.
+
+### Uninstall Scenarios:
+---
+
+1. **Uninstalling a font family with a single font face:**
+
+`winget fonts uninstall --name 'Baskerville Old Face'`
+
+Winget will uninstall a font family with a single font face without any warnings.
```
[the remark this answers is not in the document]
ryfu-msft: Updated spec to align with above.

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,416 @@
...
+            }
+        }
+```
+
+`winget fonts install`
+---
+
+### 1. Selection From Font Source
+Font source:
+There will be a separate source specifically for fonts. This source will be marked as explicit. The workflows following font source selection will be the same as if an executable package was selected. A user will be enter the following commands and experience the same behavior of installing a font.
```
Trenly, line 243: How will this new source interact with the GPO for allowed / disallowed sources?

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,416 @@
...
+| InstalledLocation    |   N/A    | %LOCALAPPDATA%/Microsoft/Windows/Fonts/Arial.ttf |
+| InstalledScope    |       |   User |
+
+> The unique identifer for each font will be a combination of the scope, font family name, and font face name.
+
+> The version of the font only exists at the font face level and not at the family level
+
+## CLI Behavior / Implementation Details
+
+New Command: `winget fonts`
```
Trenly, line 116: Will there be any new GPO to prevent the installation of fonts, or uninstallation of specific fonts? 

Thinking about how an enterprise could use a custom standard font for all of their branding / marketing and would want to ensure that users at the company can't uninstall it using WinGet

doc/specs/#166 - Font Support.md
```diff
@@ -0,0 +1,387 @@
...
+3. Remove the file and the registry entry.
+
+Once we have determined the matching font file and registry key entry, both of these items will be removed from the system and the font index will be updated.
+
+### Uninstall Scenarios:
+---
+
+1. **Uninstalling a font family with a single font face:**
+
+`winget fonts uninstall --name 'Baskerville Old Face'`
```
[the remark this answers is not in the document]
ryfu-msft: My thinking was that:
```
font face == package
font face version == version
font family == package family name
```
How should we reason about the versioning of these faces?

I haven't thought much about COM APIs but that is a good callout. Will address that later once the client implementation is a little more concrete.

doc/specs/#166 - Font Support.md
Trenly: Also, how will this interact with existing Font GPOs, if at all?

doc/specs/#166 - Font Support.md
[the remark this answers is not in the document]
JohnMcPMS: Your model is how I would put it into the existing index schema (if package == identifier).  But with the current code (at least my belief in how it should work) and a font available remotely, the search for the family would return one package result with multiple installed versions.  I would argue that this is a better model generally, and unlike executable packages we have a well-defined mechanism to do the correlation even without external data to drive it.  Thus, I would put some additional wrapper in place to always join the face "packages" together into one result based on the family name.  I feel like the existing composite code can already handle this.

doc/specs/#166 - Font Support.md
JohnMcPMS: I believe that the font files contain face and family metadata already and we are planning to always use that.  If that is not always the case, then we do probably need a way to specify the name.  The registry value names don't appear to have any actual impact on the enumerated fonts (from what I remember of @ryfu-msft investigation), but maybe they are important in the event that the file doesn't contain the relevant metadata.

doc/specs/#166 - Font Support.md
JohnMcPMS: Good question for @denelon 

The new source was just my thought on how we would separate the data and keep the index size from growing.  My expectation is that it would respect configured policy just the same, but the question is whether it would be in the umbrella of some existing controls or not (the one that controls the winget source for example).

denelon:
Closing this old PR. @dkbennett will author a new spec for fonts based on all the new changes and learnings.

PR #4844, opened 2024-10-02
Improve configuration self elevation flow

JohnMcPMS:
## Change
This change adds the ability to use schema 0.3 configuration files and set the module path when using the self-elevating configuration feature.  This required implementing the 0.3 serializer as well.

It also introduces an error when secure parameters would be passed across an integrity boundary.

## Validation
Adds tests covering serialization of 0.3, module location passing to server, and secure parameter blocking.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4844)


yao-msft (dismissed):
Update the new error codes to retuenCodes.md?

yao-msft (approved):

PR #4858, opened 2024-10-07
Add extended error codes to results without them

JohnMcPMS:
## Change
Adds `HRESULT ExtendedErrorCode{ get; };` to the `FindPackagesResult` and `ConnectResult` classes that did not have them.  They work just like the other cases, exposing the raw `HRESULT` value from their respective operations.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4858)


yao-msft (approved):
:shipit:

PR #4866, opened 2024-10-10
Make dependency lookup case insensitive

### LINKED ISSUE #3679: Dependencies should be Case Insensitive

When adding package dependencies, they are case sensitive. They probably don't need to be

image
 

_No response_

JohnMcPMS:
Fixes #3679 

## Change
Makes the dependency lookup case insensitive.

## Validation
Adds a new test that exercises adding a package with a dependency that does not match the case of the original package identifier.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4866)


Trenly:
* Resolves #3679

florelis (approved):

PR #4885, opened 2024-10-16
Change meaning of AddOrUpdate return bool

JohnMcPMS:
## Issue
The requestor of this new (1.10) API needs to know whether the operation was an add or an update.  At the same time, the "was the index modified" return value from Update has always been true for any changes since the manifest hash was added.

## Change
Return `true` when adding and `false` when updating.

## Validation
Updated the unit tests to reflect this semantic shift.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4885)


msftrubengu (approved):

msftrubengu:
So picky this requestor

JohnMcPMS:
> So picky this requestor

J'accuse

PR #4886, opened 2024-10-16
Add experimental feature for font list command

ryfu-msft:
Related to:
- #4842
- #392 

Changes:
- Adds experimental feature flag for `font` 
- Adds initial `winget font` command with `list` subcommand. 
- Adds functions for retrieving the installed font familes along with face names for each font. 
- Adds basic unit tests to verify that font enumeration works as expected.

`winget font list --family-name "Vivaldi"` will display all of the face names for the font family `Vivalidi`. At this point, this must be an exact match, but eventually this will allow partial search queries once the font index is created (separate PR).


src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "Resources.h"
+
+namespace AppInstaller::CLI
+{
+    using namespace AppInstaller::CLI::Execution;
+    using namespace AppInstaller::CLI::Workflow;
+    using namespace AppInstaller::Utility::literals;
+    using namespace std::string_view_literals;
+
+    Utility::LocIndView s_FontCommand_HelpLink = "https://aka.ms/winget-command-help"_liv;
```
mdanish-kh, line 17:
```suggestion
    Utility::LocIndView s_FontCommand_HelpLink = "https://aka.ms/winget-command-font"_liv;
```

This should be `font` i presume

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include "pch.h"
+#include <dwrite_3.h>
+#include <AppInstallerStrings.h>
+#include <winget/Fonts.h>
+
+namespace AppInstaller::Fonts
+{
+    namespace
+    {
+        std::vector<std::filesystem::path> GetFontFilePaths(const wil::com_ptr<IDWriteFontFace>& fontFace)
```
mdanish-kh, line 12: Seeing all upper case for returned file paths, can we update the casing for readability?

```
~ wingetdev font list --family-name "CaskaydiaCove Nerd Font Mono"
Face               Family                       Paths
-----------------------------------------------------------------------------------------------------------------------------------------------------------
book               CaskaydiaCove Nerd Font Mono C:\USERS\<USER>\APPDATA\LOCAL\MICROSOFT\WINDOWS\FONTS\CASKAYDIA COVE NERD FONT COMPLETE MONO.TTF
regular            CaskaydiaCove Nerd Font Mono C:\USERS\<USER>\APPDATA\LOCAL\MICROSOFT\WINDOWS\FONTS\CASKAYDIA COVE NERD FONT COMPLETE MONO REGULAR.OTF
italic             CaskaydiaCove Nerd Font Mono C:\USERS\<USER>\APPDATA\LOCAL\MICROSOFT\WINDOWS\FONTS\CASKAYDIA COVE NERD FONT COMPLETE MONO ITALIC.OTF
```

mdanish-kh:
Found myself juggling more than once between `winget font` / `winget fonts`. Would it make sense to add an alias `fonts` for subcommand `font`? _Though aliases are generally shorter and this one isn't so idk..._

ryfu-msft:
> Found myself juggling more than once between `winget font` / `winget fonts`. Would it make sense to add an alias `fonts` for subcommand `font`? _Though aliases are generally shorter and this one isn't so idk..._

@demitrius

src/AppInstallerCommonCore/Fonts.cpp
ryfu-msft: May also need to consider anonymizing path...

src/AppInstallerCLIPackage/Shared/Strings/en-us/Resources.resw
```diff
@@ -121,4 +121,10 @@
     <value>Not Localized</value>
     <comment>{Locked} This file is required to establish the basic expected subresource in the resource map. It is not used to facilitate localization with other projects.</comment>
   </data>
+  <data name="FontListCommandShortDescription" xml:space="preserve">
```
JohnMcPMS, line 124: Wrong file.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+        {
+            UINT32 fileCount = 0;
+            THROW_IF_FAILED(fontFace->GetFiles(&fileCount, nullptr));
+
+            wil::com_ptr<IDWriteFontFile> fontFiles[8];
```
JohnMcPMS, line 17: 1. static_assert that `sizeof(wil::com_ptr<IDWriteFontFile>) == sizeof(IDWriteFontFile*)`; means that it is a no-overhead smart pointer type and that you can use an array-like of them this way (regardless of the actual array-like type)
2. vector of `wil::com_ptr<IDWriteFontFile>`
3. resize to `fileCount`

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+                const void* fontFileReferenceKey;
+                UINT32 fontFileReferenceKeySize;
+                THROW_IF_FAILED(fontFiles[i]->GetReferenceKey(&fontFileReferenceKey, &fontFileReferenceKeySize));
+
+                if (const auto localLoader = loader.try_query<IDWriteLocalFontFileLoader>()) {
+                    UINT32 pathLength;
+                    THROW_IF_FAILED(localLoader->GetFilePathLengthFromKey(fontFileReferenceKey, fontFileReferenceKeySize, &pathLength));
+                    pathLength += 1; // Account for the trailing null terminator during allocation.
+
+                    wchar_t path[MAX_PATH];
```
JohnMcPMS, line 35: You have a size already, allocate some memory (preferably directly in a string so that you don't have to copy it).

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+                    THROW_HR_IF(E_OUTOFMEMORY, pathLength > ARRAYSIZE(path));
+                    THROW_IF_FAILED(localLoader->GetFilePathFromKey(fontFileReferenceKey, fontFileReferenceKeySize, &path[0], pathLength));
+                    std::filesystem::path fontFilePath = { AppInstaller::Utility::Normalize(std::wstring(path)) };
```
JohnMcPMS, line 38: `Normalize` does not seem like a necessary action to take here.  Are we planning on comparing these against each other later?

src/AppInstallerCommonCore/Fonts.cpp
JohnMcPMS: If we are given upper case from the API, I'm not sure it is worth the effort to use the filesystem to correct it back to the original casing.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+            return std::move(filePaths);
+        }
+
+        std::wstring GetFontFaceName(const wil::com_ptr<IDWriteFont>& font)
+        {
+            wil::com_ptr<IDWriteLocalizedStrings> faceNames;
+            THROW_IF_FAILED(font->GetFaceNames(faceNames.addressof()));
+
+            wchar_t localeNameBuffer[LOCALE_NAME_MAX_LENGTH];
+            const auto localeName = GetUserDefaultLocaleName(localeNameBuffer, LOCALE_NAME_MAX_LENGTH) ? localeNameBuffer : L"en-US";
```
JohnMcPMS, line 52: We already have some locale functionality for choosing installers and settings around this.  Prefer those methods over calling this function.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+            UINT32 faceNameIndex;
+            BOOL faceNameExists;
+            if (FAILED(faceNames->FindLocaleName(localeName, &faceNameIndex, &faceNameExists)) || !faceNameExists)
```
JohnMcPMS, line 56: It doesn't seem like this chooses a "best" locale to use.  We have the option of extracting all of the locale names and sorting them for best fit, much like installer selection.  Not saying that we should, but a comment may be warranted.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+            THROW_IF_FAILED(faceNames->GetStringLength(faceNameIndex, &faceNameLength));
+            faceNameLength += 1; // Account for the trailing null terminator during allocation.
+
+            wchar_t faceName[32]; // Max size for font name including null terminator is 32.
+            THROW_HR_IF(E_OUTOFMEMORY, faceNameLength > ARRAYSIZE(faceName));
+            THROW_IF_FAILED(faceNames->GetString(faceNameIndex, &faceName[0], faceNameLength));
+            return std::wstring(faceName);
+        }
+
+        std::wstring GetFontFamilyName(const wil::com_ptr<IDWriteFontFamily>& fontFamily)
```
JohnMcPMS, line 71: This is exactly the same function as `GetFontFaceName` save for the input type and function that is called to get the `IDWriteLocalizedStrings`.  Move all of the "get the correct localized string from an `IDWriteLocalizedStrings`" to a shared function.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+                std::wstring faceName = GetFontFaceName(font);
+
+                wil::com_ptr<IDWriteFontFace> fontFace;
+                THROW_IF_FAILED(font->CreateFontFace(fontFace.addressof()));
+
+                const auto& filePaths = GetFontFilePaths(fontFace);
+
+                FontFace fontFaceEntry;
+                fontFaceEntry.Name = std::wstring(faceName);
+                fontFaceEntry.FilePaths = filePaths;
```
JohnMcPMS, line 131: Apply `std::move` liberally through this type of function so that you aren't making a copy of something you are about to destroy anyway.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,188 @@
...
+        THROW_IF_FAILED(collection->FindFamilyName(familyName.c_str(), &familyIndex, &familyExists));
+
+        if (!familyExists)
+        {
+            return {};
+        }
+
+        wil::com_ptr<IDWriteFontFamily> family;
+        THROW_IF_FAILED(collection->GetFontFamily(familyIndex, family.addressof()));
+        const auto fontCount = family->GetFontCount();
```
JohnMcPMS, line 163: Factor the "create a `FontFamily` from a `IDWriteFontFamily`" into a shared function.

JohnMcPMS (changes requested):
It would be helpful as a reviewer if the PR description contained some examples of usage so that I could give feedback on the visual representation.

I also want to ensure that the ISource implementation containing installed items is well thought through to avoid the need to have specialized COM interfaces.  It doesn't even have to be a SQLite index, but if we can align the font data to the existing model we can save a lot of work exposing it.

The fun thing we could do with a custom ISource is to have the search results QI into these DWrite interfaces.

src/AppInstallerCommonCore/Fonts.cpp
Trenly: > May also need to consider anonymizing path...

And respecting the AnonymizeDisplayPaths setting

src/AppInstallerCLICore/Argument.cpp
```diff
@@ -198,6 +198,10 @@ namespace AppInstaller::CLI
         case Execution::Args::Type::IgnoreResumeLimit:
             return { type, "ignore-resume-limit"_liv, ArgTypeCategory::None };
 
+        // Font command
+        case Execution::Args::Type::FamilyName:
+            return { type, "family-name"_liv, ArgTypeCategory::None };
```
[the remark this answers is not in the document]
Trenly: Alternatively, why can't `--name` just be re-used for consistency with other winget commands? I think the spec in #4842 even calls it `--name` at [line 178](https://github.com/ryfu-msft/winget-cli/blob/91f932b7d9dc89e76076c209ac43630d4589be97/doc/specs/%23166%20-%20Font%20Support.md?plain=1#L178)

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -0,0 +1,86 @@
...
+    {
+        return {
+            Argument::ForType(Args::Type::FamilyName),
+            Argument::ForType(Args::Type::Moniker),
+            Argument::ForType(Args::Type::Source),
+            Argument::ForType(Args::Type::Tag),
+            Argument::ForType(Args::Type::Exact),
+            Argument::ForType(Args::Type::AuthenticationMode),
+            Argument::ForType(Args::Type::AuthenticationAccount),
+            Argument::ForType(Args::Type::AcceptSourceAgreements),
```
Trenly, line 56: * `-n / --count ` ?
* `--scope` ?
* `-q / --query` ? 

Will `FamilyName` be the default positional, so if a user typed `winget font list Arial` it would be the same as `winget font list --family-name Arial` ?

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    Utility::LocIndView FontListCommand::HelpLink() const
+    {
+        return s_FontCommand_HelpLink;
+    }
+
+    void FontListCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        context << Workflow::ReportInstalledFontsResult;
```
Trenly, line 84: nit: `Report` implies `Result` and is redundant; Therefore `Workflow::ReportInstalledFonts` would be sufficient

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+namespace AppInstaller::CLI::Workflow
+{
+    using namespace AppInstaller::CLI::Execution;
+
+    namespace
+    {
+        struct InstalledFontFamiliesTableLine
+        {
+            InstalledFontFamiliesTableLine(Utility::LocIndString familyName, int faceCount)
```
Trenly, line 16: Is `LocIndString` sufficient for the family name and face name? If I recall correctly, it is just a wrapper around `std::string`, will this appropriately handle family names that contain Unicode characters? Is it even possible for the names to contain Unicode characters?

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+            const auto& familyName = Utility::LocIndString(Utility::ConvertToUTF8(fontFamily->Name));
+            std::vector<InstalledFontFacesTableLine> lines;
+
+            for (const auto& fontFace : fontFamily->Faces)
+            {
+                bool isFirstLine = true;
+                for (const auto& filePath : fontFace.FilePaths)
+                {
+                    if (isFirstLine)
```
Trenly, line 80: I'm not sure that I like the idea of having table lines that have some fields being empty. While it can be appealing visually, I know that someone out there will want to script against it (even when it makes its way into the PowerShell module).  

You never know what someone will do with the output either - they could pipe it to `Sort-Object` and then it would be really confusing since all the lines would be out of order and they wouldn't contain all the information necessary to stand independently

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -0,0 +1,118 @@
...
+                table.OutputLine({ line.FaceName, line.FamilyName, line.FilePath.u8string() });
+            }
+
+            table.Complete();
+        }
+    }
+
+    void ReportInstalledFontsResult(Execution::Context& context)
+    {
+        if (context.Args.Contains(Args::Type::FamilyName))
```
Trenly, line 60: Will there be any provision for searching by Face Name? Not sure if it provides value or not

src/AppInstallerCLITests/Fonts.cpp
```diff
@@ -0,0 +1,33 @@
...
+constexpr std::wstring_view s_testFontName = L"Times New Roman";
+
+TEST_CASE("GetInstalledFonts", "[fonts]")
+{
+    std::vector<FontFamily> installedFontFamilies;
+    REQUIRE_NOTHROW(installedFontFamilies = GetInstalledFontFamilies());
+    REQUIRE(installedFontFamilies.size() > 0);
+}
+
+TEST_CASE("GetSingleFontFamily", "[fonts]")
```
Trenly, line 19: There should also be a test for getting multiple font families by substring

```
PS C:\Users\jluedtk> [Windows.Media.Fonts]::SystemFontFamilies | Select-Object -Property Source | ? { $_ -match 'Book' }

Source
------
Book Antiqua
Bookman Old Style
Bookshelf Symbol 7
Century Schoolbook
Franklin Gothic Book
```

src/AppInstallerCommonCore/Public/winget/Fonts.h
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    /// <returns>A list of installed font families.</returns>
+    std::vector<FontFamily> GetInstalledFontFamilies();
+
+    /// <summary>
+    /// Gets the installed font family from the provided family name.
+    /// </summary>
+    /// <param name="familyName">The font family name.</param>
+    /// <returns>The specified font family if found.</returns>
+    std::optional<FontFamily> GetInstalledFontFamily(const std::wstring& familyName);
```
Trenly, line 34: Seems a bit odd to me to have both a singular and a pluralized version of effectively the same function, with the only difference being that one takes a parameter to search for. Wouldn't it be possible to simply have `GetInstalledFontFamilies` with an optional parameter for the search string? Then if no families are found, you have an empty vector, if an exact match is found you'd have a vector with a single `FontFamily` in it, and if no `familyName` was specified, you'd have a vector of all the families.

Trenly (commented):
I'm not sure the way you've implemented the font families allows for Substring matching. It seems that if `--family-name` is present, you're expecting an exact match, which conflicts with the `--exact` argument. I presume the behavior should be the same as `winget list` where it defaults to a Substring match unless the `--exact` argument is specified.

src/AppInstallerCLICore/Argument.cpp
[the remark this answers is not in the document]
ryfu-msft: I'm still noodling on whether we can reuse `--name`, I think for now I will leave it as `--family` so that its clear what we're referring to. I did make it positional so you can now just do `wingetdev fonts list 'Times New Roman'`

src/AppInstallerCLICore/Commands/FontCommand.cpp
ryfu-msft: Made `--family` positional. I left out the args you mentioned above until we fully support it just so that it's clear what a user can do at this moment.

src/AppInstallerCLICore/Workflows/FontFlow.cpp
ryfu-msft: Yup, will come out in a later PR when I implement a searchable font source.

src/AppInstallerCLITests/Fonts.cpp
ryfu-msft: Currently doesn't support substrings but I'll add tests for that once I get there.

src/AppInstallerCommonCore/Public/winget/Fonts.h
ryfu-msft: Thanks for the suggestion, I agree that there's a lot of repeated code here so I combined the functionality as suggested.

src/AppInstallerCLICore/Workflows/FontFlow.cpp
ryfu-msft: I resorted to showing every value for each line. You bring up a good point and I'm sure someone may try to script with this table. I have yet to see a face that is supported by multiple font files so hopefully this doesn't even appear.

src/AppInstallerCLICore/Workflows/FontFlow.cpp
ryfu-msft: Yes this should support unicode characters because we are obtaining the localized wstring directly from the fonts library api and converting to UTF8 before creating the LocIndString.

ryfu-msft:
> I'm not sure the way you've implemented the font families allows for Substring matching. It seems that if `--family-name` is present, you're expecting an exact match, which conflicts with the `--exact` argument. I presume the behavior should be the same as `winget list` where it defaults to a Substring match unless the `--exact` argument is specified.

Once I create a fonts source and implement search functionality, I'll make this --exact argument work as expected.

src/AppInstallerCommonCore/Fonts.cpp
ryfu-msft: Added a comment noting that we may need to have better locale selection

src/AppInstallerCommonCore/Fonts.cpp
ryfu-msft: Changed to calling AppInstaller::Locale::GetUserPreferredLanguages() for the default locale.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,177 @@
...
+                const void* fontFileReferenceKey;
+                UINT32 fontFileReferenceKeySize;
+                THROW_IF_FAILED(fontFiles[i]->GetReferenceKey(&fontFileReferenceKey, &fontFileReferenceKeySize));
+
+                if (const auto localLoader = loader.try_query<IDWriteLocalFontFileLoader>()) {
+                    UINT32 pathLength;
+                    THROW_IF_FAILED(localLoader->GetFilePathLengthFromKey(fontFileReferenceKey, fontFileReferenceKeySize, &pathLength));
+                    pathLength += 1; // Account for the trailing null terminator during allocation.
+
+                    wchar_t* path = new wchar_t[pathLength];
```
JohnMcPMS, line 38: Memory leak.  A quick way to know is "using `new` probably isn't the answer".  If you can't solve it with a smart pointer, you are probably holding it wrong.

In this case, you should:
```
std::wstring path;
path.resize(pathLength);
... call the function
path.resize(pathLength - 1); // to remove the null char
```
This way you already have your `std::wstring` and don't need to create it by copying.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,177 @@
...
+                    THROW_IF_FAILED(localLoader->GetFilePathFromKey(fontFileReferenceKey, fontFileReferenceKeySize, &path[0], pathLength));
+                    filePaths.push_back(std::move(std::wstring(path)));
```
JohnMcPMS, line 40: This is an improper use of `std::move`.  All `std::move` does is to convert to r-value.

Does it have a name?
Yes means it is an l-value and will be copied.
No means it is an r-value and may be moved.
`std::move` allows you to "remove the name" from a named object.

Additionally, `push_back` takes a reference and won't move from it.  `emplace_back` can actually move from efficiently.

Note: you will need `move` here given my suggestion above.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,177 @@
...
+                }
+            }
+
+            return filePaths;
+        }
+
+        std::wstring GetLocalizedStringFromFont(const wil::com_ptr<IDWriteLocalizedStrings>& localizedStringCollection)
+        {
+            std::vector<std::string> locales = AppInstaller::Locale::GetUserPreferredLanguages();
+            std::wstring preferredLocale = Utility::ConvertToUTF16(!locales.empty() ? locales[0] : "en-US");
```
JohnMcPMS, line 50: My previous comment still stands; I don't think that defaulting to looking up `en-US` is correct.  If we can't find the preferred locale, we should just iterate through the list we have here (that seems easy enough to not leave the `//TODO`).

If we don't find one of the preferred locales, we should not use `en-US`, but rather just use index 0.

Additionally, from a performance perspective it is probably worth:
1. Creating a `std::wstring` version of `GetUserPreferredLanguages`
2. Holding onto the list of locales so that we don't have to create it so many times

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,177 @@
...
+            if (FAILED(localizedStringCollection->FindLocaleName(preferredLocale.c_str(), &index, &exists)) || !exists)
+            {
+                index = 0;
+            }
+
+            UINT32 length = 0;
+            THROW_IF_FAILED(localizedStringCollection->GetStringLength(index, &length));
+            length += 1; // Account for the trailing null terminator during allocation.
+
+            wchar_t* localizedString = new wchar_t[length];
```
JohnMcPMS, line 64: Memory leak.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,177 @@
...
+        {
+            wil::com_ptr<IDWriteLocalizedStrings> fontVersion;
+            BOOL exists;
+            THROW_IF_FAILED(font->GetInformationalStrings(DWRITE_INFORMATIONAL_STRING_VERSION_STRINGS, fontVersion.addressof(), &exists));
+            if (!exists)
+            {
+                return {};
+            }
+
+            std::string value = Utility::ConvertToUTF8(GetLocalizedStringFromFont(fontVersion));
```
JohnMcPMS, line 93: Why are you converting to UTF8 if you are going to convert back to UTF16 to return it?

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -0,0 +1,177 @@
...
+            UINT32 index;
+            BOOL exists;
+            THROW_IF_FAILED(collection->FindFamilyName(familyName.value().c_str(), &index, &exists));
+
+            if (!exists)
+            {
+                return {};
+            }
+
+            installedFontFamilies.emplace_back(GetFontFamilyByIndex(collection, index));
```
JohnMcPMS, line 163:
```suggestion
            if (exists)
            {
                installedFontFamilies.emplace_back(GetFontFamilyByIndex(collection, index));
            }
```
There are optimizations that are allowed by the compiler when only one thing is returned (can be multiple return statements, but must be the same object on each of them).  It would be more efficient to return `installedFontFamilies` than `{}`.

Additionally, the only thing you are doing after this is the `emplace_back`, so it would also be more efficient for readers to just do that in the conditional block.

src/AppInstallerCommonCore/Fonts.cpp
JohnMcPMS: Maybe a better question is just: Why are you converting back to UTF16 to return a value?  Why not our `Version` object?

https://learn.microsoft.com/en-us/typography/opentype/spec/name

> Version string. Should begin with the pattern “Version <number>.<number>” (upper case, lower case, or mixed, with a space between “Version” and the number).
>
> The string must contain a version number of the following form: one or more digits (0-9) of value less than 65,535, followed by a period, followed by one or more digits of value less than 65,535. Any character other than a digit will terminate the minor number. A character such as “;” is helpful to separate different pieces of version information.
> 
> The first such match in the string can be used by installation software to compare font versions. Some installers could require the string to start with “Version ”, followed by a version number as above.

**Should** start with "Version ".
**Must** contain a major.minor version component.

It sounds like the actual algorithm prescribed is (technically anywhere that it fails means start trying again, but we don't have to be so forgiving):
1. Find first digit
2. Read digits until non-digit
3. If anything except a `.`, error
4. If value >= 65535, error
5. Skip the `.`
6. Read digits until non-digit
7. If value >= 65535, error

But I think we can simplify it a bit:
1. Split on `.`
2. Ensure at least 2 parts
3. Part 0: Find first digit, parse number, ensure remainder consumed
4. Part 1: Ensure starts with digit, parse number, toss remainder
5. `Version` is the two parts with no "other"

I would agree that throwing out the preceding part (which is most likely "Version ") makes sense.  We could keep everything after the minor part as the "other" if we wanted to, but since this prescribes it to have no meaning, we probably shouldn't.

I would create an `OpenTypeFontVersion` (derived from `Version` of course) that encoded this, allowing a consistent construction.  This also makes it easy to test various strings, as this seems to be completely open to whatever arbitrary input someone wants to put in.

Trenly: With the new preamble stripping, a derived type may not even be necessary

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -10,6 +10,8 @@ namespace AppInstaller::Fonts
 {
     namespace
     {
+        std::vector<std::wstring> preferredLocales = AppInstaller::Locale::GetUserPreferredLanguagesUTF16();
```
JohnMcPMS, line 13: A global static is not how we should be reducing the number of calls to `GetUserPreferredLanguagesUTF16`.  Global statics bring along several issues, made even worse when they are not statically initialized.

A better approach would be to have the font enumeration functions actually be methods of a type.  The caller then controls the lifetime of anything that the font enumeration caches, like a member field of locales.

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -476,6 +480,15 @@ namespace AppInstaller::Runtime
         return result;
     }
 
+    // Try to replace LOCALAPPDATA first as it is the likely location, fall back to trying USERPROFILE.
+    void ReplaceProfilePathsWithEnvironmentVariable(std::filesystem::path& path)
```
Trenly, line 484: This also looks like a bad merge; I thought this function already existed?

https://github.com/microsoft/winget-cli/blob/fe939b80c32e40303a260b3d91b48dc1de56e8e4/src/AppInstallerCommonCore/Runtime.cpp#L167-L173

ryfu-msft: This function did exist before, I just moved it from the anonymous namespace so that I could use it elsewhere.

src/AppInstallerCommonCore/Fonts.cpp
ryfu-msft: Created a struct called `FontCatalog` that hold all of these font enumeration methods as well as a private member field for the preferred locales.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -46,13 +50,19 @@ namespace AppInstaller::Fonts
...
+            UINT32 index = 0;
+            BOOL exists = false;
 
-            UINT32 index;
-            BOOL exists;
-            // TODO: Aggregate available locales and find best alternative locale if preferred locale does not exist.
-            if (FAILED(localizedStringCollection->FindLocaleName(preferredLocale.c_str(), &index, &exists)) || !exists)
+            for (const auto& locale : preferredLocales)
+            {
+                if (FAILED(localizedStringCollection->FindLocaleName(locale.c_str(), &index, &exists)) || exists)
```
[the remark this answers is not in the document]
ryfu-msft: Changed to suggested.

JohnMcPMS (approved):

PR #4910, opened 2024-10-27
Use AMSI for archive malware scanning

pl4nty:
Replace pure with AMSI for archive malware scanning, to allow archive types other than zip. Added tests for `tgz` with `archiveExtractionMethod tar` and a stub for testing a large file. I tested with a 1GB compressed 7z and scanning took slightly under two minutes on a 4 core VM.

This was partially an experiment to see how Copilot Workspace handles C++/Windows, no worries if this feature isn't useful.

-----

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/winget-cli/issues/2899?shareId=5667db8d-1716-4173-8d5b-3808457075f4).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4910)


src/AppInstallerCLITests/TestData/TestTarGz.tar.gz
florelis, line 1: This may be me being paranoid but I would be slightly concerned about having a binary checked without knowing what it is. Would be great if you could share how it gas generated so that we can validate that it's that (assuming it's a deterministic process)

src/AppInstallerCommonCore/AppInstallerCommonCore.vcxproj
```diff
@@ -274,6 +274,21 @@
       <SubSystem Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">Windows</SubSystem>
       <SubSystem Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">Windows</SubSystem>
     </Link>
+    <Lib>
+      <AdditionalDependencies Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">amsi.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
florelis, line 278: Huh, I wonder why we don't have the other dependencies here. I think they are listed in the projects that produce the final binaries, like `WindowsPackageManager.vcxproj`. Maybe we could avoid some duplication by having the dependencies at the point they are introduced...
(Not a comment for this PR)

src/AppInstallerCommonCore/pch.h
```diff
@@ -18,6 +18,7 @@
 #include <wincrypt.h>
 #include <deliveryoptimization.h>
 #include <deliveryoptimizationerrors.h>
+#include <amsi.h>
```
florelis, line 21: I'm not sure if we need it in `pch.h` if it is only used in `Archive.cpp`. But I do see the TODO comment from before about moving it 🤔

src/AppInstallerCommonCore/Archive.cpp
```diff
@@ -1,79 +1,93 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
-#include "pch.h"
-#include "Public/winget/Archive.h"
-
-// TODO: Move include statement to pch.h and resolve build errors
-#pragma warning( push )
-#pragma warning ( disable : 4189 4244 26451 )
-#include <pure.h>
```
florelis, line 9: IF we do decide to go with AMSI instead of PureLib, there are a bunch other places we have to remove it from. We have the source code at `src/PureLib/`, all the references in `AdditionalIncludeDirectories` and `AdditionalDependencies`, and removing the project from the solution. There's also the some housekeeping to do with the cgmanifest and the notice file

src/AppInstallerCommonCore/Archive.cpp
```diff
@@ -1,79 +1,93 @@
...
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        if (s_ScanArchiveResult_TestHook_Override)
+        {
+            return *s_ScanArchiveResult_TestHook_Override;
+        }
+#endif
+
+        HRESULT hr = S_OK;
+        HAMSICONTEXT amsiContext;
+        HAMSISESSION amsiSession;
```
florelis, line 66: These should use one of the WIL magic RAII wrappers to make the destruction easier and ensure it happens. I don't see any specific for AMSI, so it would probably be a `wil::unique_any`

src/AppInstallerCLITests/Archive.cpp
```diff
@@ -1,34 +1,82 @@
...
+#include <Workflows/ShellExecuteInstallerHandler.h>
+
+using namespace AppInstaller::Archive;
+using namespace AppInstaller::CLI::Workflow;
+using namespace AppInstaller::Settings;
+using namespace TestCommon;
+
+constexpr std::string_view s_ZipFile = "TestZip.zip";
+constexpr std::string_view s_TarGzFile = "TestTarGz.tar.gz";
+constexpr std::string_view s_Large7zFile = "TestLarge7z.7z";
```
florelis, line 16: This is missing

florelis (commented):
If AMSI was a suitable replacement to PureLib, I'd be all for changing it just to have one less external dependency.

But I don't think it is a good replacement based on the documentation. I only see mentions of AMSI supporting scripts or .NET binaries, while PureLib is for scanning ZIP files specifically. For example, it's not clear to me that AMSI would catch a zip bomb.

SpecterShell:
> For example, it's not clear to me that AMSI would catch a zip bomb.

Most antivirus software should be able to detect zip bomb. I'm using ESET and just tried out AMSI with this code:

https://gist.github.com/SpecterShell/c2b5507ba9e5c531faead38c7d0b877e

And ESET passed me a notification for detecting the zip bomb.

EDIT: Microsoft Defender doesn't report on the zip bomb though.

pl4nty:
I'm so sorry Flor, I completely forgot about this PR! Thanks for the thorough review. I'm digging into this further - I suspect IOfficeAntivirus (via IAttachmentExecute) may already invoke AV scanning for zips, so AMSI could be redundant. And I assume PureLib would need to stay if Defender is still missing zip bombs, like Charlie found

pl4nty:
I've analysed >50 zip/tar test cases, with some interesting findings
* Identical results from AMSI and IAttachmentExecute. I suspect AVs that register for AMSI usually register for IOfficeAntivirus too. So this PR is probably redundant
* Microsoft Defender (via IOfficeAntivirus) detected malicious content in zips and all the compressed tars (gz etc). But not structural exploits like zip bombs
* [Other AVs detected structural zip exploits](https://www.virustotal.com/gui/file/bbd05de19aa2af1455c0494639215898a15286d9b05073b6c4817fe24b2c36fa)
* `bsdtar` [mitigates some risks by default](https://man.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&format=html#SECURITY), like absolute paths, traversal, and symlinks. Adding `--keep-old-files` (`-k`) could provide additional protection against file overwrites
* PureLib detects excessive file size/counts, but `bsdtar` doesn't. I couldn't find any equivalent tar libraries. Maybe listing contents with `tar -tf` could be enough of a mitigation, rather than importing libarchive?

Issue #4919, opened 2024-10-29
The configuration unit was not in the module as expected


Gijsreyn:
### Brief description of your issue

When applying the following on a fresh created Hyper-V box through DevHome:

```yaml
#configure.devbox.config.yaml
properties:
  resources:
  - resource: Microsoft.WinGet.DSC/WinGetPackage
    directives:
      description: Installing Git
      allowPrerelease: true
    settings:
      id: "Git.Git"
      source: winget
    id: Git.Git | Git
  - resource: GitDsc/GitClone
    directives:
      description: 'Cloning: winget-dsc.git'
      allowPrerelease: true
      securityContext: current
    settings:
      httpsUrl: https://github.com/microsoft/winget-dsc.git
      rootDirectory: C:\Repos\winget-dsc.git
    id: 'Clone winget-dsc.git: C:\Repos\winget-dsc.git'
    dependsOn:
    - Git.Git | Git
  configurationVersion: 0.2.0
```

I get the error: The configuration unit was not in the module as expected.

Related: https://github.com/microsoft/winget-cli/issues/4021

### Steps to reproduce

Create Hyper-V box through DevHome with default settings. Create a file with the above and execute winget configure configure.devbox.config.yaml --accept-configuration-agreements

### Expected behavior

Clone repository on C:\repos

### Actual behavior

The configuration unit was not in the module as expected.

### Environment

Windows Package Manager v1.8.1911
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22621.3880
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.23.1911.0

Winget Directories
-----------------------------------------------------------------------------------------------------------------------
Logs                               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\Diag…
User Settings                      %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\sett…
Portable Links Directory (User)    %LOCALAPPDATA%\Microsoft\WinGet\Links
Portable Links Directory (Machine) C:\Program Files\WinGet\Links
Portable Package Root (User)       %LOCALAPPDATA%\Microsoft\WinGet\Packages
Portable Package Root              C:\Program Files\WinGet\Packages
Portable Package Root (x86)        C:\Program Files (x86)\WinGet\Packages
[...]

PR #4932, opened 2024-10-31
Remove attribute in DSC

JohnMcPMS:
## Change
Removes the attribute `ValidateNotNullOrWhiteSpace` as it is only available starting in 7.4.  The attributes also only validate when the value is assigned, so enforcing that a value is set is more complete.

## Validation
Loads in PS 7.2.8 now.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4932)

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
+++ b/src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
@@ -190,7 +190,6 @@ class WinGetAdminSettings
 class WinGetSource
 {
     [DscProperty(Key, Mandatory)]
-    [ValidateNotNullOrWhiteSpace()]
     [string]$Name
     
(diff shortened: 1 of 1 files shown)


src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -471,6 +474,11 @@ class WinGetPackage
 
     [WinGetPackage] Get()
     {
+        if ([String]::IsNullOrWhiteSpace($this.Name))
```
ryfu-msft, line 477:
```suggestion
        if ([String]::IsNullOrWhiteSpace($this.Id))

```

ryfu-msft (approved):

denelon:
Related to:
* #4919

PR #4960, opened 2024-11-11
Update curl dependency due to CVE-2024-9681

florelis:
See https://curl.se/docs/CVE-2024-9681.html
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4960)

### DIFF
--- a/src/vcpkg.json
+++ b/src/vcpkg.json
@@ -38,5 +38,5 @@
       "./VcpkgCustomTriplets"
     ]
   },
-  "builtin-baseline": "f176b58f35a75f9f8f54099cd9df97d2e2793a2e"
+  "builtin-baseline": "e9235c7244fccb9881139976f579df1a816f5159"
 }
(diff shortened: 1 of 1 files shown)


yao-msft (approved):

yao-msft:
The build error could probably be fixed by updating the baseline.

yao-msft:
Looks like the official repo has not been updated from 8.10.1 yet... https://vcpkg.io/en/package/curl

florelis:
Guess this will have to wait then...
It isn't in any of the commits marked as "release" but the update is already in [vcpkg's repo](https://github.com/microsoft/vcpkg/commit/e9235c7244fccb9881139976f579df1a816f5159), so I thought setting the baseline to that commit would be enough. Weird that it worked locally but not on the pipeline

yao-msft (approved):

PR #4969, opened 2024-11-14
Add ODS logger and use it during init

JohnMcPMS:
## Change
The user settings (and then group policy) initialization happen before we get the file logger set up.  To be able to see into what is happening there more easily, this change creates a logger that uses `OutputDebugString` and uses it during the initial file logger creation.

## Validation
Debugged and saw the group policy and settings logging come through the output.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4969)

### DIFF
--- a/src/AppInstallerCLICore/Core.cpp
+++ b/src/AppInstallerCLICore/Core.cpp
@@ -186,10 +186,17 @@ namespace AppInstaller::CLI
     void ServerInitialize()
     {
 #ifndef AICLI_DISABLE_TEST_HOOKS
+        // We have to do this here so the auto minidump config initialization gets caught
+        Logging::OutputDebugStringLogger::Add();
+        Logging::Log().EnableChannel(Logging::Channel::All);
+        Logging::Log().SetLevel(Logging::Level::Verbose);
+
         if (Settings::User().Get<Settings::Setting::EnableSelfInitiatedMinidump>())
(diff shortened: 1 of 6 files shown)


src/AppInstallerCLICore/Core.cpp
```diff
@@ -66,21 +67,34 @@ namespace AppInstaller::CLI
...
 #endif
 
         Logging::UseGlobalTelemetryLoggerActivityIdOnly();
 
         Execution::Context context{ std::cout, std::cin };
-        auto previousThreadGlobals = context.SetForCurrentThread();
+        auto previousThreadGlobals = context.SetForCurrentThread();
+
+        // Set up debug string logging during initialization
+        Logging::OutputDebugStringLogger::Add();
```
florelis, line 90: I don't see why we have to remove the ODS logger only to add it immediately after. Couldn't we just add it before the `#ifndef` and remove it once in line 97?

florelis (approved):

src/AppInstallerCLICore/Core.cpp
JohnMcPMS: The first one is added to the global logger, while the second one is added to the context logger.

PR #4974, opened 2024-11-15
Add default module root setting for configuration

ryfu-msft:
Changes:
- Adds a new settings object for `configureBehavior`
- Adds a new field for `defaultModuleRoot` that determines the default directory where modules are installed to.

Tests:
- Verify settings value is parsed correctly
- Added E2E test to check that the module is installed to the directory specified from the settings.json file.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4974)

### DIFF
--- a/src/AppInstallerCLIE2ETests/ErrorCommand.cs
+++ b/src/AppInstallerCLIE2ETests/ErrorCommand.cs
@@ -1,4 +1,4 @@
-﻿// -----------------------------------------------------------------------------
+// -----------------------------------------------------------------------------
 // <copyright file="ErrorCommand.cs" company="Microsoft Corporation">
 //     Copyright (c) Microsoft Corporation. Licensed under the MIT License.
 // </copyright>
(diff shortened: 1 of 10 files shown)


JohnMcPMS (approved):

src/AppInstallerCLIE2ETests/ErrorCommand.cs
```diff
@@ -14,6 +14,15 @@ namespace AppInstallerCLIE2ETests
     /// </summary>
     public class ErrorCommand
     {
+        /// <summary>
+        /// Reset settings file to avoid affecting output from error command.
```
florelis, line 18: Why did you add this? I'm guessing changing the settings in the Configure tests was affecting this? If it's that, maybe we should also add it as a OneTimeTearDown to these tests, so that it does not affect other tests if the order ever changes. But also keep it here if these tests expect clean settings

florelis (approved):

src/AppInstallerCLIE2ETests/ErrorCommand.cs
ryfu-msft: Yup that's right. One of the tests in the error command checks if there are 2 lines returned from the error command but if the settings file is invalid, it will return >2. I think this should be sufficient since the error command only outputs to console. It doesn't change the settings file during the tests so it won't impact other tests.

florelis (approved):

florelis (approved):

PR #4990, opened 2024-11-19
[WinGet DSC] Add WinGetConfigRoot cmdlet

ryfu-msft:
Added a new dsc resource for ensuring that the $WinGetConfigRoot environment variable exists and is set. Added pester tests to verify that variables are set for both user and machine scope.

Some commands I tried out:

`Invoke-DscResource -Name WinGetConfigRoot -Method Get -ModuleName Microsoft.WinGet.DSC -Property @{ Exist=0; Path="C:\Foo\Bar"; WinGetScope='User' }`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4990)

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
+++ b/src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
@@ -52,7 +52,7 @@ enum WinGetTrustLevel
     Trusted
 }
 
-enum Scope
+enum WinGetScope
 {
     User
     Machine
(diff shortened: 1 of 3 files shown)


src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -446,6 +452,78 @@ class WinGetPackageManager
...
+    {
+        if (-not $this.Test())
+        {
+            if ($this.Exist)
+            {
+                [System.Environment]::SetEnvironmentVariable($this.WinGetConfigRoot, $this.Path, $this.WinGetScope)
+            }
+            else
+            {
+                [System.Environment]::SetEnvironmentVariable($this.WinGetConfigRoot, "", $this.WinGetScope)
```
florelis, line 521: Nit: this could set the variable to `$null` to delete it instead of setting it to the empty string. According to the docs, `""` does not delete the variable since .NET 9 (no idea which one this uses...)

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -52,6 +52,12 @@ enum WinGetTrustLevel
     Trusted
 }
 
+enum WinGetScope
```
florelis, line 55: Seems like this needs to be a subset of `EnvironmentVariableTarget`. I didn't know you could use enums like that

Small nit: if it is related to the other enum, I would add a comment for the future

florelis (approved):

src/PowerShell/tests/Microsoft.WinGet.DSC.Tests.ps1
```diff
@@ -242,3 +242,36 @@ Describe 'WinGetPackageManager' {
...
+
+        InvokeWinGetDSC -Name WinGetConfigRoot -Method Set -Property @{ Exist = $true; Path = 'C:\Foo\Bar'; WinGetScope = $_ }
+
+        # Verify that $WinGetConfigRoot variable is set.
+        $result = InvokeWinGetDSC -Name WinGetConfigRoot -Method Get -Property @{WinGetScope = $_}
+        $result.Exist | Should -Be $true
+        $result.Path | Should -Be 'C:\Foo\Bar'
+        $result.WinGetScope | Should -Be $_
+
+        # Remove $WinGetConfigRoot variable.
```
Trenly, line 265: Before you remove it, it may be a good idea to check that if a different Path is specified while the variable exists, then the test returns false.

src/PowerShell/tests/Microsoft.WinGet.DSC.Tests.ps1
```diff
@@ -242,3 +242,36 @@ Describe 'WinGetPackageManager' {
...
+        $test = InvokeWinGetDSC -Name WinGetConfigRoot -Method Test -Property @{ Exist = $true; Path = 'C:\Foo\Bar'; WinGetScope = $_ }
+        $test.InDesiredState | Should -Be $false
+
+        InvokeWinGetDSC -Name WinGetConfigRoot -Method Set -Property @{ Exist = $true; Path = 'C:\Foo\Bar'; WinGetScope = $_ }
+
+        # Verify that $WinGetConfigRoot variable is set.
+        $result = InvokeWinGetDSC -Name WinGetConfigRoot -Method Get -Property @{WinGetScope = $_}
+        $result.Exist | Should -Be $true
+        $result.Path | Should -Be 'C:\Foo\Bar'
+        $result.WinGetScope | Should -Be $_
```
Trenly, line 263: What is the expected result of test if the path is not specified, but the user has requested `Exist: true`?

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -446,6 +452,78 @@ class WinGetPackageManager
...
 }
 
+[DSCResource()]
+class WinGetConfigRoot
+{
+    [DscProperty(Key)]
+    [string]$WinGetConfigRoot = "WinGetConfigRoot"
+
+    [DscProperty()]
+    [string]$Path = ""
```
Trenly, line 462: There should be validation that `Exist: $true` cannot be used with an empty path in `Set`, otherwise that results in a state conflict where the user has requested a `Exist` to be true, but the result of the setter will always result in `Exist` being false

Trenly (changes requested):

PR #5034, opened 2024-12-03
Adding the client code for supporting experiments

AmelBawa-msft:
Adding the client-side code for supporting running experiments in winget-cli.

Group policy:
- EnableExperiments: Disabling this group policy will prevent any experiment from running.

User settings:
- "experiments": Enable the user to control whether or not an experiment should run or not.

-----


doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -25,6 +25,12 @@ If you disable this setting, users will not be able to change settings for the W
 If you enable or do not configure this setting, users will be able to enable experimental features for the Windows Package Manager.
 
 If you disable this setting, users will not be able to enable experimental features for the Windows Package Manager.</string>
+      <string id="EnableExperiments">Enable Windows Package Manager Experiments</string>
```
AmelBawa-msft, line 28: Adding @RDMacLachlan to help with the strings to be used in the adml file

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -760,6 +776,20 @@ namespace AppInstaller::Logging
...
 
+    Settings::ExperimentState TelemetryTraceLogger::GetExperimentState(Experiment::Key key)
+    {
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        m_summary.Experiments.clear();
+#endif
+
+        if (m_summary.Experiments.find(key) == m_summary.Experiments.end())
+        {
+            m_summary.Experiments[key] =  Experiment::GetStateInternal(key);
```
florelis, line 787: Do we need to cache this? Doesn't seem like we use it anywhere but the summary event, which fires only once AFAIK

src/AppInstallerSharedLib/Public/winget/JsonUtil.h
```diff
@@ -33,6 +33,9 @@ namespace AppInstaller::JSON
 
     template<>
     std::optional<std::vector<std::string>> GetValue<std::vector<std::string>>(const Json::Value& node);
+
+    template<>
+    std::optional<std::map<std::string, bool>> GetValue(const Json::Value& node);
```
florelis, line 38: Why does this one not have the type after `GetValue`? Like how above it is `GetValue<std::vector<std::string>>` instead of just `GetValue`?

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -278,6 +304,7 @@ namespace AppInstaller::Settings
         WINGET_VALIDATE_PASS_THROUGH(UninstallPurgePortablePackage)
         WINGET_VALIDATE_PASS_THROUGH(NetworkWingetAlternateSourceURL)
         WINGET_VALIDATE_PASS_THROUGH(MaxResumes)
+        WINGET_VALIDATE_PASS_THROUGH(Experiments)
```
florelis, line 307: We don't do any validation of these?

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -0,0 +1,63 @@
...
+        ExperimentToggleSource m_toggleSource;
+        bool m_isEnabled;
+    };
+
+    struct Experiment
+    {
+        enum class Key : unsigned
+        {
+            None = 0x0,
+            CDN = 0x1,
```
florelis, line 34: Is this going to be one of these enums where we increase by 1 or by powers of 2? I'm guessing the former

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    ExperimentState Experiment::GetStateInternal(Key key)
+    {
+        return GetExperimentStateInternal(key, User());
+    }
+
+    ExperimentState Experiment::GetState(Key key)
+    {
+        return Logging::Telemetry().GetExperimentState(key);
```
florelis, line 80: It doesn't feel right to have this get the state from the telemetry..

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -0,0 +1,63 @@
...
+#endif
+        };
+
+        using Key_t = std::underlying_type_t<Key>;
+
+        Experiment(std::string_view name, std::string_view jsonName, std::string_view link, std::string key) :
+            m_name(name), m_jsonName(jsonName), m_link(link), m_key(key) {}
+
+        static ExperimentState GetState(Key feature);
+        static ExperimentState GetStateInternal(Key feature);
```
florelis, line 48: What makes it internal if it is public?

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -0,0 +1,109 @@
...
+            if (!GroupPolicies().IsEnabled(TogglePolicy::Policy::Experiments))
+            {
+                AICLI_LOG(Core, Info, <<
+                    "Experiment " << Experiment::GetExperiment(key).Name() <<
+                    " is disabled due to group policy: " << TogglePolicy::GetPolicy(TogglePolicy::Policy::Experiments).RegValueName());
+                return { false, ExperimentToggleSource::Policy };
+            }
+
+            auto experiments = userSettings.Get<Setting::Experiments>();
+            auto experiment = Experiment::GetExperiment(key);
```
florelis, line 30: Nit: Can you make the naming more clear on how these are different or related?

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -0,0 +1,109 @@
...
+            auto jsonName = std::string(experiment.JsonName());
+            if (experiments.find(jsonName) != experiments.end())
+            {
+                auto isEnabled = experiments[jsonName];
+                AICLI_LOG(Core, Info, <<
+                    "Experiment " << Experiment::GetExperiment(key).Name() <<
+                    " is set to " << isEnabled << " in user settings");
```
florelis, line 37: I think this will log as 0 or 1 (unless we have `<<` overriden for bool?). Is that what you want?

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -200,7 +202,11 @@ namespace AppInstaller::Settings
         SETTINGMAPPING_SPECIALIZATION(Setting::LoggingChannelPreference, std::vector<std::string>, Logging::Channel, Logging::Channel::Defaults, ".logging.channels"sv);
         // Interactivity
         SETTINGMAPPING_SPECIALIZATION(Setting::InteractivityDisable, bool, bool, false, ".interactivity.disable"sv);
-        
+
+        // Experiments
+        using Experiments_t = std::map<std::string, bool>;
+        SETTINGMAPPING_SPECIALIZATION(Setting::Experiments, Experiments_t, Experiments_t, {}, ".experiments"sv);
```
florelis, line 208: This allows to enable/disable individual experiments. As a user, I would like to be able to disable all experimentation. I.e., the same we have for group policy, but available for a single user. If somebody wanted to disable all experiments, they would have to stay on the look for any experiments we add, and that's not good

src/AppInstallerCommonCore/Public/winget/UserSettings.h
florelis, line 1: Changes to settings require updating the JSON schema and doc file

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -207,6 +209,16 @@ namespace AppInstaller::Repository
...
+
+        std::string_view GetWingetCommunitySource()
+        {
+            if (Settings::Experiment::GetState(Settings::Experiment::Key::CDN).IsEnabled())
+            {
+                return s_Source_WingetCommunityExperimental_Arg;
+            }
+
+            return s_Source_WingetCommunityDefault_Arg;
+        }
```
florelis, line 221: We do some validation when a user tries to add one of the default sources. For example, to block it if the group policy restricts the default source. This may allow for an unintended way to bypass that validation. E.g., somebody has the experiment and the group policy, which means we block adding the cdn2.winget... source, but they can still add cdn.winget...

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -278,7 +290,7 @@ namespace AppInstaller::Repository
 
     std::optional<WellKnownSource> CheckForWellKnownSourceMatch(std::string_view name, std::string_view arg, std::string_view type)
     {
-        if (name == s_Source_WingetCommunityDefault_Name && arg == s_Source_WingetCommunityDefault_Arg && type == Microsoft::PreIndexedPackageSourceFactory::Type())
+        if (name == s_Source_WingetCommunityDefault_Name && arg == GetWingetCommunitySource() && type == Microsoft::PreIndexedPackageSourceFactory::Type())
```
florelis, line 293: See my previous comment. Not sure if this is the only place that needs something like this
```suggestion
        if (name == s_Source_WingetCommunityDefault_Name && (arg == s_Source_WingetCommunityDefault_Arg || arg == s_Source_WingetCommunityExperimental_Arg) && type == Microsoft::PreIndexedPackageSourceFactory::Type())
```

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -0,0 +1,109 @@
...
+        default:
+            THROW_HR(E_UNEXPECTED);
+        }
+    }
+
+    std::vector<Experiment> Experiment::GetAllExperiments()
+    {
+        std::vector<Experiment> result;
+
+        for (Key_t i = 0x1; i < static_cast<Key_t>(Key::Max); i = i << 1)
```
florelis, line 102: So the keys go by powers of two. Please add a comment to the enum definition so that we get it right when adding a new one. Although I think it only needs to be powers of two if it was a bitmask; so you could instead simply change the step here to `++i`

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -0,0 +1,63 @@
...
+
+        std::string_view Name() const { return m_name; }
+        Utility::LocIndView JsonName() const { return m_jsonName; }
+        std::string_view Link() const { return m_link; }
+        std::string GetKey() const { return m_key; }
+
+    private:
+        std::string_view m_name;
+        Utility::LocIndView m_jsonName;
+        std::string_view m_link;
```
florelis, line 60: I would think `m_name` and `m_link` need to have ownership of the data and not be `string_view`

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return Logging::Telemetry().GetExperimentState(key);
+    }
+
+    Experiment Experiment::GetExperiment(Key key)
+    {
+        switch (key)
+        {
+        case Key::CDN:
+            return Experiment{ "CDN experiment", "CDN", "https://aka.ms/winget-settings", "CDN"};
```
florelis, line 88: The first field looks like something that needs to be a loc string.

doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -25,6 +25,12 @@ If you disable this setting, users will not be able to change settings for the W
 If you enable or do not configure this setting, users will be able to enable experimental features for the Windows Package Manager.
 
 If you disable this setting, users will not be able to enable experimental features for the Windows Package Manager.</string>
+      <string id="EnableExperiments">Enable Windows Package Manager Experiments</string>
+      <string id="EnableExperimentsExplanation">This policy controls whether users can enable experiments in the Windows Package Manager.
```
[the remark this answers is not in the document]
AmelBawa-msft: +1 this is a placeholder text I added and will be replaced by PM's suggestion 😊

src/AppInstallerCLITests/AppInstallerCLITests.vcxproj.filters
```diff
@@ -371,6 +371,9 @@
     <ClCompile Include="Fonts.cpp">
       <Filter>Source Files\Common</Filter>
     </ClCompile>
+    <ClCompile Include="Experiment.cpp">
+      <Filter>Source Files\Repository</Filter>
```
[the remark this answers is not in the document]
AmelBawa-msft: Moved to Common next to ExperimentFeatures

doc/admx/en-US/DesktopAppInstaller.adml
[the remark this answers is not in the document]
denelon: I would use "Experimentation" rather than "Experiments"

The policy controls whether users can have experiments run in WinGet on their device.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
AmelBawa-msft: Getting the state of an experiment is used when applying that experiment in a specific use case and again that state will be reported in the summary event. This cache will also only contain the experiments that has been used during a workflow, and only those will be reported in the summary event.

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -838,7 +870,8 @@ namespace AppInstaller::Logging
                     AICLI_TraceLoggingStringView(m_summary.RepairExecutionType, "RepairExecutionType"),
                     TraceLoggingUInt32(m_summary.RepairErrorCode, "RepairErrorCode"),
                     TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance | PDT_ProductAndServiceUsage | PDT_SoftwareSetupAndInventory),
-                    TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES));
+                    TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+                    AICLI_TraceLoggingJsonString(experimentsJson, "ExperimentsJson"));
```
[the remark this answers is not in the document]
AmelBawa-msft: Good question, I am not entirely sure why we handle it that way. I know the Json API `Json::writeString` returns `std::string`, but it would be helpful to understand why some strings are reported as `wstring` and others are `string`. Adding @JohnMcPMS to help clarify this 😊

src/AppInstallerCommonCore/Experiment.cpp
AmelBawa-msft: I primarily used `ExperimentalFeature` as a [reference for this class's implementation](https://github.com/microsoft/winget-cli/blob/6034e7829616f838c691c1e2b10bc4cf6413a8ee/src/AppInstallerCommonCore/ExperimentalFeature.cpp#L77). But I can localize it if necessary.

src/AppInstallerCommonCore/Experiment.cpp
AmelBawa-msft: You're right! I copy pasted the code from `ExperimentalFeature` and missed that line. Good catch 😊

src/AppInstallerCommonCore/Public/winget/UserSettings.h
AmelBawa-msft: I like this idea. I can add a boolean `enableExperiments` field in user settings. Disabling this would result in disabling all experiments. Thoughts?

src/AppInstallerRepositoryCore/SourceList.cpp
JohnMcPMS: Yes, I think the following should cover it:

1. Remove `GetWellKnownSourceArg`
2. Update `CheckForWellKnownSourceMatch` to
  a. be case insensitive
  b. handle both DNS entries
  c. allow for not checking certain values (ex. name), maybe by passing an empty `string_view`
4. Replace use of `GetWellKnownSourceArg` in `GetPolicyBlockingUserSource` with the updated `CheckForWellKnownSourceMatch`

It didn't look like any other paths were relevant there.

src/AppInstallerSharedLib/JsonUtil.cpp
```diff
@@ -67,6 +67,27 @@ namespace AppInstaller::JSON
 
         return std::nullopt;
     }
+
+    template<>
+    std::optional<std::map<std::string, bool>> GetValue(const Json::Value& node)
+    {
+        std::map<std::string, bool> result;
```
JohnMcPMS, line 74: Move into inner scope.

src/AppInstallerSharedLib/JsonUtil.cpp
```diff
@@ -67,6 +67,27 @@ namespace AppInstaller::JSON
...
+    {
+        std::map<std::string, bool> result;
+
+        if (node.isObject())
+        {
+            for (const auto& entry : node.getMemberNames())
+            {
+                if (node[entry].isBool())
+                {
+                    result[entry] = node[entry].asBool();
```
JohnMcPMS, line 82: Get `node[entry]` once and use twice.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
denelon: As long as the default value in the client is set so that experiments are enabled by default (excluding GPO), I'm fine with having a setting a user could add to the client to disable experimentation. The real concern here is that if we're doing experiments, we need to know when a user disables them if an experiment is running so they can be excluded from the results.

Experimentation is only valid when a user doesn't know if they are in the control group or the treatment group. The same is true if a user is receiving an experiment and they disable an experimental feature that was enabled through experimentation.

We would benefit most by having the ability to enable experimental treatments for devices on Windows Insider Canary and Dev builds for features that we're developing (once they are ready) so we can determine if they are working properly, or if users are getting a better desired outcome.

I think we'll need to send a telemetry event if a user opts-out of experimentation and an experiment is running. I also think we'll need to send a telemetry event related to the experimental treatment if a user disables a feature that was enabled via an experiment.

Obviously if a user has disabled telemetry, we wouldn't receive any events, and if an experiment is running and telemetry is disabled, we will no longer get that data, so I'd also say we should disable any experiments if a user disables telemetry.

doc/admx/DesktopAppInstaller.admx
florelis, line 1: I imagine there is already a policy to disable experimentation across Windows as a whole. I found [AllowExperimentation](https://learn.microsoft.com/en-us/windows/client-management/mdm/policy-csp-system#allowexperimentation) but I'm not sure if it does what I think.

If there is such a policy, I think we would be better off not having our own and just pointing people to that one. If somebody disables it system-wide, then our policy won't do anything because the infrastructure we use will be disabled. The only case where it would be useful to have both is if somebody wants to have experimentation in Windows as a whole, but not on winget specifically, which I don't think is a common scenario

doc/admx/en-US/DesktopAppInstaller.adml
florelis, line 28: I would suggest the policy be called "Allow" instead of "Enable" even if it is not consistent with the other policies.
To me, "Enable" means that the experiments will be turned on but that's not what happens. What happens is that experiments _may_ be turned on depending on the control/treatment groups; which is better described by "Allow"

It probably should also mention that they are Microsoft-run experiments for new features/changes

src/AppInstallerCommonCore/Public/winget/UserSettings.h
florelis: > The real concern here is that if we're doing experiments, we need to know when a user disables them if an experiment is running so they can be excluded from the results.

In the telemetry he's already including whether the experiment is enabled and the reason, i.e. whether it was disabled by GP, overridden from settings, or came from being in the control/treatment group. That should be enough to make a distinction; we just have to add a new reason for the experiment state - "disabled by settings"

I imagine for the comparison we would ignore all groups except people who do not have any override, which should have about the same split in population size as the general population

> I can add a boolean `enableExperiments` field in user settings. Disabling this would result in disabling all experiments. Thoughts?

Sounds good to me, except that I would call it `allow` instead of `enable`

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -0,0 +1,63 @@
...
+        ExperimentState(bool isEnabled, ExperimentToggleSource toggleSource) : m_isEnabled(isEnabled), m_toggleSource(toggleSource) {}
+        bool IsEnabled() const { return m_isEnabled; }
+        ExperimentToggleSource ToggleSource() const { return m_toggleSource; }
+        std::string ToJson() const;
+    private:
+        ExperimentToggleSource m_toggleSource;
+        bool m_isEnabled;
+    };
+
+    struct Experiment
```
florelis, line 29: Could we make experiments part of the experimental features?

Experimental features already have most of the things we may want: settings to be turned on/off, GPO to disable them wholesale, an API to check the state, a command to show the current state. The only thing we would need to add is the option that they may be set remotely for an experiment if there is no local override.

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -0,0 +1,109 @@
...
+            {
+                return { true, ExperimentToggleSource::Default };
+            }
+
+            if (!GroupPolicies().IsEnabled(TogglePolicy::Policy::Experiments))
+            {
+                AICLI_LOG(Core, Info, <<
+                    "Experiment " << Experiment::GetExperiment(key).Name() <<
+                    " is disabled due to group policy: " << TogglePolicy::GetPolicy(TogglePolicy::Policy::Experiments).RegValueName());
+                return { false, ExperimentToggleSource::Policy };
```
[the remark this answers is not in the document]
JohnMcPMS: Maybe we should consider the meaning of the boolean associated with an experiment not "enabled" or "disabled", but rather, "In the experimental group?"  That makes the default state of "false" mean "Not in the experimental group", and thus more directly maps to group policy that disables running experiments.  A comment on `ExperimentState::IsEnabled` indicating this semantic and counter-indicating use of the "inverted" experiment would be sufficient.

src/AppInstallerCLI.sln
```diff
@@ -212,25 +212,39 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Microsoft.Management.Deploy
...
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Internal", "Internal", "{800529ED-3374-49C4-BAD1-9B27647E4359}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Experiment", "Experiment", "{13E55125-0ABF-4975-9B5B-DD6E5527D47F}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Experiment", "Internal\Experiment\Experiment.vcxproj", "{98920AB6-27B0-4C0F-B336-FA49DE57A1BA}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|ARM = Debug|ARM
```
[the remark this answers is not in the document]
JohnMcPMS: VS is constantly trying to put it back if you add a project with an ARM configuration.

Do please remove.

src/AppInstallerCommonCore/Public/winget/Experiment.h
JohnMcPMS: I think we might still want a new GP for "allow experimental feature state to be changed remotely", but other than that it was my original thought that all experiments would likely have a feature.  I forget the reasons why I softened on that, but I do think that there may be some value to separating them as they are slightly different semantics.  One potential large reason is that if we use EFs exclusively, then a disabled GP for EFs also disables experiments.  Moving experiments out to their own thing means an admin can disable in-development features while still allowing for experiments.

Typically, an EF is work-in-progress and will eventually be released.  Using an experiment on an EF is more akin to controlled feature rollout.  I think it will probably happen that we hook that up so that an EF can be controlled by an experiment and thus we get CFR.

On the other hand, a true experiment is typically a small configuration or code change that is itself not risky from a local functionality perspective (the impact, such as changing the CDN *could* have major impacts on the overall system of course).  We might be testing something that we want to make permanent like where the DNS for the CDN points, or we might be checking to see if customer satisfaction numbers are higher if the default progress style is rainbow.

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -13,35 +13,30 @@ namespace AppInstaller::Settings
...
             {
-                AICLI_LOG(Core, Info, <<
-                    "Experiment " << Experiment::GetExperiment(key).Name() <<
+                AICLI_LOG(Core, Info, << "Experiment " << Experiment::GetExperiment(key).Name() <<
                     " is disabled due to group policy: " << TogglePolicy::GetPolicy(TogglePolicy::Policy::Experiments).RegValueName());
                 return { false, ExperimentToggleSource::Policy };
             }
 
-            auto experiments = userSettings.Get<Setting::Experiments>();
+            if (key == Experiment::Key::None)
```
JohnMcPMS, line 23: Move this back to first.  While I'm not sure that we need a `None`, if we have one then checks against it should not end up logging about group policy being disabled.

src/AppInstallerCommonCore/Experiment.cpp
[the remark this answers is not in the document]
JohnMcPMS: It looks like you did this, but only to uses *after* the existing `experiment` local was created.  Move the local earlier so that it can be truly constructed once.

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -41,23 +41,23 @@ namespace AppInstaller::Settings
 
         using Key_t = std::underlying_type_t<Key>;
 
-        Experiment(std::string_view name, std::string_view jsonName, std::string_view link, std::string key) :
-            m_name(name), m_jsonName(jsonName), m_link(link), m_key(key) {}
+        Experiment(std::string name, std::string jsonName, std::string link, std::string key) :
+            m_name(std::move(name)), m_jsonName(jsonName), m_link(std::move(link)), m_key(std::move((key))) {}
```
JohnMcPMS, line 45:
```suggestion
            m_name(std::move(name)), m_jsonName(std::move(jsonName)), m_link(std::move(link)), m_key(std::move((key))) {}
```
Accompanied by making `m_jsonName` not a view.

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -41,23 +41,23 @@ namespace AppInstaller::Settings
...
+        Experiment(std::string name, std::string jsonName, std::string link, std::string key) :
+            m_name(std::move(name)), m_jsonName(jsonName), m_link(std::move(link)), m_key(std::move((key))) {}
 
         static ExperimentState GetState(Key feature);
         static ExperimentState GetStateInternal(Key feature);
         static Experiment GetExperiment(Key key);
         static std::vector<Experiment> GetAllExperiments();
 
-        std::string_view Name() const { return m_name; }
+        std::string Name() const { return m_name; }
```
JohnMcPMS, line 52:
```suggestion
        const std::string& Name() const { return m_name; }
```
And the rest as `const&`.

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -41,23 +41,23 @@ namespace AppInstaller::Settings
...
         Utility::LocIndView JsonName() const { return m_jsonName; }
-        std::string_view Link() const { return m_link; }
+        std::string Link() const { return m_link; }
         std::string GetKey() const { return m_key; }
 
     private:
-        std::string_view m_name;
+        std::string m_name;
         Utility::LocIndView m_jsonName;
```
JohnMcPMS, line 59: This does not need to be a localized value, does it?

src/AppInstallerCommonCore/Public/winget/Experiment.h
florelis: > as they are slightly different semantics

If users are going to come across the two concepts, do we think the difference is clear enough for them to understand?

> an admin can disable in-development features while still allowing for experiments

Do we expect there will be admins wanting to do that?

src/Internal/Experiment/Experiment.vcxproj
```diff
@@ -0,0 +1,414 @@
...
+      <PrecompiledHeader>NotUsing</PrecompiledHeader>
+      <PreprocessorDefinitions>_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <WarningLevel>Level4</WarningLevel>
+      <AdditionalOptions>%(AdditionalOptions) /permissive- /D _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING</AdditionalOptions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
+    <ClCompile>
+      <Optimization>Disabled</Optimization>
+      <PreprocessorDefinitions>_NO_ASYNCRTIMP;_SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING;_DEBUG;%(PreprocessorDefinitions);CLICOREDLLBUILD</PreprocessorDefinitions>
```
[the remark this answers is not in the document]
AmelBawa-msft: I copied this file from existing vcxproj, but this marcro is not needed here 😉 Good catch!

src/AppInstallerCommonCore/Public/winget/Experiment.h
JohnMcPMS: I don't think we want users to come across experiments really; the major reason for the control via settings is to enable *us* to test.

As to admins wanting to do that; I'm not sure.  But if we are not going to build any user facing interfaces around them, then I see the minor additional cost as acceptable.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
AmelBawa-msft: - Updated the JSON schema
- TODO Updated the md file

src/AppInstallerCommonCore/Experiment.cpp
```diff
@@ -0,0 +1,109 @@
...
+        ExperimentState GetExperimentStateInternal(Experiment::Key key, const UserSettings& userSettings)
+        {
+            if (key == Experiment::Key::None)
+            {
+                return { true, ExperimentToggleSource::Default };
+            }
+
+            if (!GroupPolicies().IsEnabled(TogglePolicy::Policy::Experiments))
+            {
+                AICLI_LOG(Core, Info, <<
```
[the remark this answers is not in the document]
AmelBawa-msft: Should all the logs in this method be Verbose?

src/AppInstallerSharedLib/JsonUtil.cpp
AmelBawa-msft: This currently lives in the same namespace scope as the other `GetValue` definitions. Let me know if you meant something else.

src/AppInstallerCommonCore/Public/winget/Experiment.h
```diff
@@ -0,0 +1,66 @@
...
+        UserSettingIndividualControl,
+        UserSettingGlobalControl,
+    };
+
+    struct ExperimentState
+    {
+        ExperimentState() = default;
+        ExperimentState(bool isEnabled, ExperimentToggleSource toggleSource) : m_isEnabled(isEnabled), m_toggleSource(toggleSource) {}
+
+        /// <summary>
```
florelis, line 24: We don't use this doc string format in C++. I think we just add the comment above.

src/AppInstallerCLITests/Experiment.cpp
```diff
@@ -0,0 +1,110 @@
...
+#define ASSERT_EXPERIMENT(_isEnabled_, _toggleSource_) \
+    auto testExperimentState = Experiment::GetState(ExperimentKey::TestExperiment); \
+    REQUIRE(_isEnabled_ == testExperimentState.IsEnabled()); \
+    REQUIRE(_toggleSource_ == testExperimentState.ToggleSource());
+
+TEST_CASE("Experiment_GroupPolicyControl", "[experiment]")
+{
+    SECTION("Not configured")
+    {
+        Experiment::ResetStates();
```
florelis, line 22: I would prefer this to be a struct that resets the state on destruction, like `GroupPolicyTestOverride` and `TestUserSettings`. Just to be sure that nothing we set here can escape and affect other tests

src/AppInstallerCLITests/Experiment.cpp
florelis, line 1: A couple of extra tests that would be good:
* Policy=Disabled and Global Allow=True -> disabled
* Policy=Enabled and Global Allow=False -> disabled
* Policy=Enabled and Individual Allow=False-> Disabled

AmelBawa-msft: Good point! The first one is already covered by "Experiment_GroupPolicyDisabled_ReturnFalse", but I added a test case for the other two

src/AppInstallerCommonCore/Public/winget/Experiment.h
AmelBawa-msft: Correct! It does not surface outside the logs.

doc/admx/en-US/DesktopAppInstaller.adml
RDMacLachlan: **Title:** Enable Windows Package Manager Feature Experimentation
**Description:**
This policy controls whether a user will be the recipient of feature experimentation while using the Windows Package Manager. 
If you enable, or do not configure this setting, user may receive experiences intended to analyze features and functionality experiences in the Windows Package Manager.
If you disable this setting, user will be excluded from receiving any modified experience in the Windows Package Manager.

doc/admx/en-US/DesktopAppInstaller.adml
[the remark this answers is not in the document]
AmelBawa-msft: Renamed all occurrences of Experiments to Experimentation following the GPO text suggested in the previous comment.

src/AppInstallerCommonCore/Experiment.cpp
AmelBawa-msft: Agree that this is currently a little hacky and piggybacking off telemetry to benefit from the lifetime of the thread-globals is a little awkward. At the moment I don't have an enhanced approach. I can file a follow up issue if we want to revisit this in the future or if we notice that more features need to benefit from the same scope.

doc/admx/en-US/DesktopAppInstaller.adml
AmelBawa-msft: I remember discussing this with Roy offline last Wednesday, but tagging him here in case he wants to add further information. Correct me if I am wrong Roy, but we will keep this Enable instead to align with the other policies.

CC: @RDMacLachlan

src/AppInstallerCommonCore/AppInstallerTelemetry.cpp
```diff
@@ -77,6 +80,19 @@ namespace AppInstaller::Logging
                 return "unknown"sv;
             }
         }
+
+        std::string GetExperimentsJson(const std::map<Experiment::Key, ExperimentState>& experiments)
```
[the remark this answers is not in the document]
JohnMcPMS: Still feel this should be done.

saltman007web (commented):
reviewed

Issue #3401, opened 2023-07-02
Expose more settings to Microsoft.WinGet.DSC/WinGetPackage


rlabrecquefsg:
### Description of the new feature / enhancement

I'm migrating over to `winget configure` from my bespoke `winget install` scripts, and am running into some missing features in WinGetPackage. This is a request to expose more of `winget install` and `Install-WinGetPackage`'s settings via `Microsoft.WinGet.DSC/WinGetPackage` so they can be used with `winget configure`, and `devhome`.

Notably missing  `--custom` and `--scope` for my needs.
I imagine some people would require `--location` and `--architecture`, and `--override` to make use of this feature as well.

For my current needs these only need to be set during install/upgrade.

In the long term it would be nice to have `Microsoft.WinGet.DSC/WinGetPackage` go down the re-install/upgrade path if say the location, scope, or custom/override settings change. But this should be a seperate ticket once both this ticket and https://github.com/microsoft/winget-cli/issues/3281 are complete!

Sorry @denelon, you knew this was coming😄

### Proposed technical implementation details

Adding more settings here:
https://github.com/microsoft/winget-cli/blob/e4b9205cff88701b2ad1078dd8ed311ac716b8b6/src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1#L545C12-L549C14

Like so: https://github.com/microsoft/winget-cli/blob/master/src/PowerShell/Microsoft.WinGet.Client/Examples/Sample_InstallPackage.ps1#L25

mdanish-kh:
[Policy] Command-Configure

denelon:
Yep 👍 I knew it was coming.

The goal here would be to essentially provide parity with the CLI in the DSC resource in terms of arguments you can pass to ensure a package is installed.

We'll have to make a laundry list for all the arguments that make sense.

egerlach:
Just ran into a use case for `--installer-type` in my configuration.

OranguTech:
Lack of `--scope` makes this a non-starter for us.

denelon:
Another bug we need to address with this is the scenario where an older version of WinGet with an expired certificate for the "msstore" causes the configuration to fail when building the composite source. Those users can remove the "msstore" source, but since the DSC resource is specifying "winget" as the source for the package, it shouldn't worry about trying to connect to the "msstore" source (or any other source for that matter).

When the source is specified in the configuration, it should be treated as a "targeted" source.

[...]

PR #5062, opened 2024-12-13
Clarify settings behavior

Gijsreyn:
- [x] This pull request is related to a discussion.

-----

When looking through the discussions, I noticed the following [question](https://github.com/microsoft/winget-cli/discussions/4985) asked by Tristan. After trying out some options, I thought the `ConfigureBehavior` would do the trick. Unfortunately, after trying it out, I took the conclusion from the description that only `winget install` is affected by these settings. This pull request clarifies it a bit more.

### DIFF
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -129,6 +129,10 @@ Some of the settings are duplicated under `preferences` and `requirements`. `pre
 
 Any arguments passed on the command line will effectively override the matching `requirement` setting for the duration of that command.
 
+> [!NOTE]
+> - These settings are only applied for the `winget install` command.
+> - Other commands like `winget configure` are not affected by these settings.
+
 ### Scope
 
 The `scope` behavior affects the choice between installing a package for the current user or for the entire machine. The matching parameter is `--scope`, and uses the same values (`user` or `machine`).


Trenly (changes requested):
They affect more than just `install`. If I'm not mistaken, they also affect upgrade and import.

In fact, I would think it's nore likely a bug in DSC if they aren’t being considered when used with the WinGetPackage resource; especially becausethefe doesn’t seem to be `Scope` or `Locale` or `InstallerType` properties on the DSC Resource (@denelon - should these be added? Should I open an issue for that?)

Gijsreyn:
Thanks for the reply Kaleb. If that's the case, then I would even think of renaming the setting name, as it states `installBehavior`, but that might be a preference and opinion.

Trenly:
> Thanks for the reply Kaleb. If that's the case, then I would even think of renaming the setting name, as it states `installBehavior`, but that might be a preference and opinion.

Unfortunately I think that renaming the setting would be considered a breaking change (unless both the old name and the new name would work)

Gijsreyn:
I agree; maybe only introducing `upgradeBehavior` would suffice, as `uninstallBehavior` is already there.

Trenly:
> I agree; maybe only introducing `upgradeBehavior` would suffice, as `uninstallBehavior` is already there.

Then you get into weird semantics because an upgrade is just a special case of install

denelon:
In general, a WinGet Configuration shouldn't be affected by a user's Package Management settings. That's probably something we'll need to clarify.

Trenly:
> In general, a WinGet Configuration shouldn't be affected by a user's Package Management settings. That's probably something we'll need to clarify.

In that case, does the DSC Resource need to be updated to support Scope, Locale, and installer type? It seems like a gap between the DSC Resource and the WinGet Client if there isn’t a way to specificy which is the preferred scope, or preferred installer

denelon:
Yes, we have an issue open for that work as well.

* https://github.com/microsoft/winget-cli/issues/3401

denelon:
* https://github.com/microsoft/winget-cli/issues/5064

Trenly:
> Yes, we have an issue open for that work as well.
> 
> * [Expose more settings to Microsoft.WinGet.DSC/WinGetPackage #3401](https://github.com/microsoft/winget-cli/issues/3401)

Ah, I knew I was missing something!

JohnMcPMS:
They should be affecting `configure`.  A package install via configure is: `winget.exe` -> `Hosted PowerShell: WinGet DSC` -> `Hosted PowerShell: WinGet Client` -> `WinGet COM install`.  The COM server uses the same settings as winget.exe.

I think that the statements in this change are inaccurate; at a minimum with desired behavior but I think also potentially with actual behavior.  If there is a discrepancy that would likely be a bug.

PR #5074, opened 2024-12-16
Implement export WinGetPackage DSC resource

Gijsreyn:

-----

Adds the export method on `WinGetPackage` class, just like some other resources in the [winget-dsc](https://github.com/microsoft/winget-dsc) repository.

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
+++ b/src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
@@ -79,7 +79,7 @@ class WinGetUserSettings
     {
         $userSettings = Get-WinGetUserSetting
         $result = @{
-            SID = ''
+            SID      = ''
             Settings = $userSettings
         }
         return $result
@@ -135,7 +135,7 @@ class WinGetAdminSettings
         # Get admin setting values.
 
         $result = @{
-            SID = ''
+            SID      = ''
             Settings = $settingsJson.adminSettings
         }
         return $result
@@ -216,10 +216,12 @@ class WinGetSource
 
         $currentSource = $null
 
-        try {
+        try
+        {
             $currentSource = Get-WinGetSource -Name $this.Name
         }
-        catch {
+        catch
+        {
         }
 
         $result = [WinGetSource]::new()
@@ -295,7 +297,7 @@ class WinGetSource
         if ($addSource)
         {
             $hashArgs = @{
-                Name = $this.Name
+                Name     = $this.Name
                 Argument = $this.Argument
             }
 
@@ -396,11 +398,13 @@ class WinGetPackageManager
             if ($this.UseLatest)
             {
                 $hashArgs.Add("Latest", $true)
-            } elseif ($this.UseLatestPreRelease)
+            }
+            elseif ($this.UseLatestPreRelease)
             {
                 $hashArgs.Add("Latest", $true)
                 $hashArgs.Add("IncludePrerelease", $true)
-            } elseif (-not [string]::IsNullOrWhiteSpace($this.Version))
+            }
+            elseif (-not [string]::IsNullOrWhiteSpace($this.Version))
             {
                 $hashArgs.Add("Version", $this.Version)
             }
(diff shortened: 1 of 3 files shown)


denelon:
@Gijsreyn what does the output look like for export. I want to make sure we have parity in the resource to be able to use "set" on the same output so logically (assuming there are no dependencies/conflicts), the resource can install those packages.

Gijsreyn:
@denelon here you have an example output:

image

Trenly:
> @denelon here you have an example output:
> 
> image

If you use `ConvertTo-JSON` and pipe that output toa file, can that raw JSON file be used directly with Winget Import (from the cli, not cmdlet)

Gijsreyn:
image

denelon:
Does it return an array of all packages?

We have a command for generating a configuration file from the current state of the device:
`winget configure export -o <output file>` in experimental status.
The goal is to be able to run that command and the output is a configuration file of the set of installed packages (the configurations for the ones that are DSC v3, and the Windows Settings.

The current "export" / "import" format would still be supported, but I'd like to make the command smart enough that it could do either the packages.json format or the WinGet Configuration YAML format over the fullness of time.

Gijsreyn:
Thanks for both replies. Yes it indeed returns an array of all packages. I knew about the experimental feature; currently, it is YAML, if I'm not mistaken. It doesn't fully follow the DSC schematics. However, if the command does it with the V3 schema (which is also in experimental?) in the future, this PR becomes obsolete.

denelon:
We're adding experimental support to leverage DSC v3 for configuration (WinGet will also support the current configuration schema). We're planning to make WinGet a DSC v3 resource so we can use that to get/set/test/export. With the new experimental feature enabled at that point, we should be closer to where we want to be.

denelon:
How would you envision the work in this PR being used. I just want to make sure we're not confusing folks with all the different mechanisms :) As long as we have good guidance and the right use cases, I'm fine with adding export to the current Microsoft.WinGet.DSC resource, but I want to move intentionally :)

Gijsreyn:
Got it Demitrius. I was just filling the gap between winget-dsc and the one in this repository. Since `Export` is the new method introduced in the DSC repository, I thought it was beneficial for users to beware of it. The `Microsoft.WinGet.DSC` module is more popular, so showing the users that it is available might help bridge the line between DSC, WinGet, and PowerShell.

denelon:
I understand. My main concern is that if someone were to run the following command, they would subsequently be able to run `winget configure packages.winget` and all the packages they had on their system coming from a configured source would be installed on the new system:

`winget configure export --package Microsoft.AppInstaller --module Microsoft.WinGet.DSC --resource WinGetPackage --output packages.winget`

Gijsreyn:
Mhm... I get your point, Demitrius. I have a strong feeling that there is an underlying question hidden here. Executing the above command will not really work because the properties within the DSC resource are different unless they can be added to the command call.

Nevertheless, the output of a DSC resource will not follow the schematics defined in WinGet. Now, I'm thinking of solutions to bridge the gap between WinGet and DSC. It would easily be solvable by providing a static method that converts it to a JSON string, expecting the format WinGet expects. 

Maybe a talk with Steve and Mikey would be worth a shot to discus the options?

denelon:
Yeah, we're actively in discussions around how we're going to support both DSC v2 resources as well as DSC v3 resources, and how the WinGet Configuration Export is going to behave.

src/PowerShell/Microsoft.WinGet.DSC/Microsoft.WinGet.DSC.psm1
```diff
@@ -579,6 +585,26 @@ class WinGetPackage
             }
         }
     }
+
+    static [WinGetPackage[]] Export()
+    {
+        $packages = Get-WingetPackage
+        $out = [List[WinGetPackage]]::new()
+        foreach ($package in $packages)
```
JohnMcPMS, line 593: For our existing `export` CLI command, we only export packages that we can install later.  This would require filtering the packages to those that have an available version.  While we could do that in the DSC, I think it would be better to add a filter parameter to `Get-WingetPackage` for that purpose.

Gijsreyn: What filter parameter are you thinking about?

JohnMcPMS: Something like `Get-WingetPackage -Export`, but there might be a more idiomatic name for such a parameter.  It would filter to only the packages that have an available version.

Gijsreyn: Perhaps a new cmdlet `Export-WinGetPackage` would also fit the use case...

Gijsreyn: Interestingly. Is there something that needs to be changed in this PR, or is it better to abandon it with all the possibilities already flying around?

JohnMcPMS: Sorry that this got abandoned via lack of communication, but if you want to proceed with this PR I would update `Get-WinGetPackage` with a switch to only get packages with available versions and leverage that here.

PR #5111, opened 2025-01-06
Add missing top-level exception handling for arg validation

### LINKED ISSUE #5098: winget silent termination/exit: no stdout/stderror output, 0 output

No error or any kind of feeback to the user.
Is Winget going the MS usual style of meaningful feedback?

### Steps to reproduce

Winget doesnt output any useful information and just exits without anything.

### Expected behavior

I want to have some **meaningful** error message, not the usual Windows behaviour where processes exit without any kind of output or with some useless 0x0800RANDOM code - the usual

### Actual behavior

Image

### Environment

```shell
Winget 1.10.40
windows 10.0.19045.4651
```

florelis:
The CLI arg validation code has gotten more complicated and now there is more room for exceptions to occur. For example, if there is an error when reading the admin settings or group policies that determine if an argument is available.

This adds generic exception handling at the top-level of arg validation so that we can exit gracefully and show an error, instead of just crashing.

Partial fix for #5098. This addresses the no output, but not the cause of the error

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5111)

### DIFF
--- a/src/AppInstallerCLICore/Core.cpp
+++ b/src/AppInstallerCLICore/Core.cpp
@@ -173,6 +173,10 @@ namespace AppInstaller::CLI
             context.Reporter.Error() << Resource::String::DisabledByGroupPolicy(policy.PolicyName()) << std::endl;
             return APPINSTALLER_CLI_ERROR_BLOCKED_BY_POLICY;
         }
+        catch (...)
+        {
+            return Workflow::HandleException(context, std::current_exception());
+        }
 
         return Execute(context, command);
     }


src/AppInstallerCLICore/Core.cpp
```diff
@@ -172,6 +172,10 @@ namespace AppInstaller::CLI
             AICLI_LOG(CLI, Error, << "Operation blocked by Group Policy: " << policy.RegValueName());
             context.Reporter.Error() << Resource::String::DisabledByGroupPolicy(policy.PolicyName()) << std::endl;
             return APPINSTALLER_CLI_ERROR_BLOCKED_BY_POLICY;
+        }
+        catch (...)
+        {
+            return Workflow::HandleException(context, std::current_exception());
```
yao-msft, line 178: nit: Workflow::HandleException handles GroupPolicyException above. Btw I don't have preference on whether we would list GroupPolicyException above explicitly or let HandleException handle all non-command exceptions.

yao-msft (approved):

src/AppInstallerCLICore/Core.cpp
Trenly: I'm in favor of having explicit handling in this case, since the message for Group Policy is quite clear about what is happening

PR #5126, opened 2025-01-12
Yaml manifest schemaheader validation for V.1.10.0 and above

Madhusudhan-MSFT:
### Enable schema header validation for YAML manifests from version 1.10.0 onwards.

**YAML Manifest Validation Changes:**

1. Introduce functions to search, parse, and validate schema headers in YAML manifests. 
2. Add/Update structs and enums to support schema header validation. 
3. Add new error messages and validation options, including treating schema header validation errors as warnings.

Schema Header Validation Includes:
1. Validate presence of Manifest Schema Header.
2. Ensure Schema Header format correctness.
3. Verify Schema Header URL Pattern accuracy.
4. Confirm Schema Header Manifest Type matches ManifestType Property.
5. Check Schema Header Manifest Version matches ManifestVersion Property.

**WinGet CLI Validation Command Changes:**

- Treat schema header validation errors as warnings for 'winget validate --manifest  <<path>>' command
   - Introduced a new option `SchemaHeaderValidationAsWarning` to the `validateOption` object and set it to `true`. 
   - This change treats schema header validation issues as warnings instead of errors, making the validation process more lenient.

[NOTE:]
1. SchemaHeaderValidation errors should be treated as warnings for the winget CLI validate command.
2. SchemaHeaderValidation errors should be treated as errors for wingetsvc community manifests.

**[Test Coverage:]**
-  Added new test data files to validate various schema header errors in YAML manifests, including
   - invalid headers,
   -  type mismatches,
   - version mismatches,
   - missing headers, and
   - URL pattern mismatches.
 - Added test coverage for WinGet Utils Interop binary and WinGet CLI Validated command

**[How Validated:]**
- Compiled the AppInstaller end-to-end solution incorporating the changes.
- Performed the YAML manifest validation tests, WinGet Utils tests, and WinGet CLI Validate command tests, ensuring all tests passed.
- Locally validated the winget CLI validation command with various schema header errors, ensuring they are treated as warnings and that no warnings are reported for a valid manifest.

**[Test Results:]**

YAML Manifest Validation Test Results:
image

WinGet Utils Interop Test Results:
image

Validate Command Test Results:
image

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5126)


src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -98,6 +110,167 @@ namespace AppInstaller::Manifest::YamlParser
...
+                errors.emplace_back(ValidationError::MessageContextValueLineLevelWithFile(ManifestError::InvalidSchemaHeader, "", schemaHeader, manifestSchemaHeader.Line, manifestSchemaHeader.column, errorLevel, manifestSchemaHeader.FileName));
+            }
+
+            return errors;
+        }
+
+        bool ParseSchemaHeaderUrl(const std::string& schemaHeaderValue, std::string& schemaType, std::string& schemaVersion)
+        {
+            // Use regex to match the pattern of @"winget-manifest\.(?<type>\w+)\.(?<version>[\d\.]+)\.schema\.json$"
+            std::regex schemaUrlPattern(R"(^\$schema=https://aka\.ms/winget-manifest\.(\w+)\.([\d\.]+)\.schema\.json$)");
```
Trenly, line 189: It isn’t guaranteed that the URL will always use the aka.ms link - in the case of new schema versions, the aka.ms link may not even exist, in which case the github link to the raw file is used instead

Madhusudhan-MSFT: sure, I'll exclude the domain name from the regex pattern.

Madhusudhan-MSFT: The right way, I believe, is to read the JSON schema definition file, extract the $id to get the schema URL, and then compare it with the schema header URL string.

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
```diff
@@ -479,6 +479,14 @@ namespace AppInstaller::Manifest::YamlParser
             {
                 errors = ValidateManifest(manifest);
                 std::move(errors.begin(), errors.end(), std::inserter(resultErrors, resultErrors.end()));
+
+                // Validate the schema header for manifest version 1.10 and above
+                if (manifestVersion >= ManifestVer{ s_ManifestVersionV1_10 })
```
Madhusudhan-MSFT, line 484: @denelon - Hopefully, you agree that enforcing Schema Header validation only for manifests V.1.10.0 and above will avoid unnecessary confusion, as older manifests without the schema header may start to fail.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -98,6 +110,213 @@ namespace AppInstaller::Manifest::YamlParser
...
+                std::string schemaId = schemaFile["$id"].asString();
+
+                // Prefix schema ID with "schema=" to match the schema header URL pattern and compare it with the schema header URL
+                schemaId = "$schema=" + schemaId;
+
+                if (std::equal(schemaId.begin(), schemaId.end(), schemaHeaderUrlString.begin(), schemaHeaderUrlString.end(),
+                    [](char a, char b) { return tolower(a) == tolower(b); }))
+                {
+                    return true;
+                }
```
Madhusudhan-MSFT, line 240: Do we need to perform a case-insensitive comparison, or should we enforce case sensitivity (all lowercase)?

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
Madhusudhan-MSFT, line 484: @denelon - Hopefully, you agree that enforcing Schema Header validation only for manifests V.1.10.0 and above will avoid unnecessary confusion, as older manifests without the schema header may start to fail.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -63,7 +63,12 @@ namespace AppInstaller::Manifest
...
-                { AppInstaller::Manifest::ManifestError::ShadowManifestNotAllowed, "Shadow manifest is not allowed." }
+                { AppInstaller::Manifest::ManifestError::ShadowManifestNotAllowed, "Shadow manifest is not allowed." },
+                { AppInstaller::Manifest::ManifestError::SchemaHeaderNotFound, "Schema header not found." },
+                { AppInstaller::Manifest::ManifestError::InvalidSchemaHeader , "The schema header is invalid. Please verify that the schema header is present and formatted correctly as a valid YAML node with the appropriate commented syntax."sv },
+                { AppInstaller::Manifest::ManifestError::SchemaHeaderManifestTypeMismatch , "The manifest type in the schema header does not match the ManifestType property value in the manifest."sv },
+                { AppInstaller::Manifest::ManifestError::SchemaHeaderManifestVersionMismatch, "The manifest version in the schema header does not match the ManifestVersion property value in the manifest."sv },
+                { AppInstaller::Manifest::ManifestError::SchemaHeaderUrlPatternMismatch, "The schema header URL does not match the expected pattern."sv },
```
Madhusudhan-MSFT, line 71: I just created this resource string to indicate errors related to Schema Header validation. Please let me know if you have any feedback to make it clearer.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
Trenly: > However, the concern is that the manifest might include a random domain name, which may not be valid, but it will still be treated as a valid manifest header.

The only two domains that are in use today (that I know of) are the `aka.ms` links, and `raw.githubusercontent.com`

I don’t think it's a bad idea to limit it to known-valid domains, but do understand the concern about proliferation.

> The right way, I believe, is to read the JSON schema definition file, extract the $id to get the schema URL, and then compare it with the schema header URL string.

This is probably true, but then the question could arise about using URLs from forks of the repo, or linking to a schema that is always treated as valid.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -98,6 +110,213 @@ namespace AppInstaller::Manifest::YamlParser
...
         }
+
+        bool SearchForManifestSchemaHeaderString(std::shared_ptr<std::istream> yamlInputStream, const size_t& rootNodeBeginsAtLine, ManifestSchemaHeader& schemaHeader)
+        {
+            std::string line;
+            size_t currentLine = 1;
+            schemaHeader.SchemaHeaderString.clear();
+
+            // Search for the schema header string in the comments before the root node.
+            while (currentLine < rootNodeBeginsAtLine && std::getline(*yamlInputStream, line))
```
JohnMcPMS, line 121: This code assumes that the input stream actually contains 8-bit characters.  The YAML parser (wrapper) supports detecting the encoding on the input stream and handling it properly.  The `Parser` type is the one that ends up containing the full string text of the stream, while `Document` contains probably the level that a schema should apply to (we don't really attempt to do anything past the first document in a stream).

Since the YAML parser ignores comments, I would suggest moving the code that detects the schema into the wrapper (probably `Document Parser::Load()`) and making it a property of a `Document`.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
JohnMcPMS: I think it is fine to be case-sensitive here as I don't think anyone will be typing the value in by hand.

src/AppInstallerCLICore/Commands/ValidateCommand.cpp
```diff
@@ -46,6 +46,7 @@ namespace AppInstaller::CLI
             {
                 ManifestValidateOption validateOption;
                 validateOption.FullValidation = true;
+                validateOption.SchemaHeaderValidationAsWarning = true;
```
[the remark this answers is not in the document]
Madhusudhan-MSFT: This option allows the YAML parser to treat SchemaHeader violations as either errors or warnings. For the winget CLI validate command, these violations should be warnings, but for API calls through WinGetUtils, they should be treated as errors

src/AppInstallerCommonCore/Manifest/YamlParser.cpp
Madhusudhan-MSFT: It was confirmed in a separate discussion that this is the intended behavior.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
Madhusudhan-MSFT: For the manifest file type `defaultLocale`, we use `defaultlocale `(all lowercase) in the schema file URL reference, but use camelCase in other references. To simplify the comparison, I will keep this case insensitive for now.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -98,6 +110,175 @@ namespace AppInstaller::Manifest::YamlParser
...
+
+            return errors;
+        }
+
+        std::vector<ValidationError> ValidateYamlManifestSchemaHeader(const YamlManifestInfo& manifestInfo, const ManifestVer& manifestVersion, ValidationError::Level errorLevel)
+        {
+            std::vector<ValidationError> errors;
+            std::string schemaHeaderString;
+
+            ManifestSchemaHeader schemaHeader{};
```
JohnMcPMS, line 265: All of this information is in the `manifestInfo`, which is also passed along to `ValidateSchemaHeaderUrl`.  Do we really need to copy it into a custom struct?

Madhusudhan-MSFT: Yes, it is no longer required.

src/AppInstallerCommonCore/Manifest/ManifestSchemaValidation.cpp
```diff
@@ -98,6 +110,175 @@ namespace AppInstaller::Manifest::YamlParser
...
+
+            if (actualHeaderVersion != expectedManifestVersion)
+            {
+                errors.emplace_back(ValidationError::MessageContextValueLineLevelWithFile(ManifestError::SchemaHeaderManifestVersionMismatch, "", schemaHeader.ManifestVersion, schemaHeader.Line, actualHeaderVersionIndex, errorLevel, schemaHeader.FileName));
+            }
+
+            return errors;
+        }
+
+        bool IsValidSchemaHeaderUrl(const std::string& schemaHeaderUrlString, const YamlManifestInfo& manifestInfo, const ManifestVer& manifestVersion)
```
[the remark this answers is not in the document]
Madhusudhan-MSFT: As we discussed offline, the purpose of `ParseSchemaHeaderUrl `is to extract the _ManifestType_ and _ManifestVersion_ details from the schema header, which we then compare against the `ManifestType` and `ManifestVersion `property values in the manifest. This is an optional requirement for the work item.

src/AppInstallerSharedLib/YamlWrapper.cpp
```diff
@@ -319,6 +334,12 @@ namespace AppInstaller::YAML::Wrapper
             }
         }
 
+        DocumentSchemaHeader schemaHeader;
+        if (result.HasRoot() && FindManifestSchemaHeaderString(result.GetRoot().Mark().line, schemaHeader))
```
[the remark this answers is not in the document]
Trenly: > (we could allow an arbitrary number of blank lines and whitespace characters before it).

I think part of it too, is that there could be any number of other comments before the schema comment. For example, wingetcreate will put a comment stating `# Created using wingetcreate {version}`

In fact, there should probably be a test case for it not being first-line-first-character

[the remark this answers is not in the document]
Madhusudhan-MSFT: Thank you for pointing this out. It appears that the call to `GetRoot().Mark.line `unintentionally caused a performance hit by reloading the entire document unnecessarily. I will address this by moving the logic to the` LoadDocument()` call. 

To clarify the need for marking the root node line:  It is beneficial to have the root node line marking when performing a search to retrieve the schema header string. The assumption is that the schema header string always exists at the beginning of the YAML document, before the first YAML node (i.e., the root node). Knowing this, we don't need to search/parse the entire YAML document comments to find the Schema Header string. However, I agree that we shouldn't call `GetRoot()` more than once for this to happen.

JohnMcPMS (approved):

PR #5139, opened 2025-01-15
Add configuration environment

JohnMcPMS:
## Change
This change adds a `ConfigurationEnvironment` type which is a property of a `ConfigurationUnit`.  It defines the environment in which the unit should be run.  This currently encompasses two properties; the security context and the processor.

The values are parsed from the metadata, but the object is considered authoritative at runtime and when serializing.  They are also inherited in schema 0.3, allowing the entire set to be defined in a single environment in it's metadata.

The code that was using the `securityContext` metadata has been updated to use the environment property instead.  It also uses the unique environment calculation function for the set to more efficiently determine which contexts are present.

## Validation
New tests are added for parsing, serialization, and unique environment calculation.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5139)


src/Microsoft.Management.Configuration/ConfigurationSetUtilities.h
```diff
@@ -38,10 +39,33 @@ namespace winrt::Microsoft::Management::Configuration::implementation
...
     };
 
     // Gets the name value of the configuration field.
     std::string_view GetConfigurationFieldName(ConfigurationField fieldName);
 
     winrt::hstring GetConfigurationFieldNameHString(ConfigurationField fieldName);
+
+    // Attempts to parse a security context from a string.
+    // Returns true if successful; false otherwise.
+    bool TryParseSecurityContext(hstring value, SecurityContext& result);
```
yao-msft, line 55: nit: Do we need `const hstring&` for this and below?

src/Microsoft.Management.Configuration/ConfigurationEnvironment.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    Windows::Foundation::Collections::IMapView<hstring, hstring> ConfigurationEnvironment::ProcessorPropertiesView() const
+    {
+        return m_processorProperties.GetView();
+    }
+
+    void ConfigurationEnvironment::DeepCopy(const implementation::ConfigurationEnvironment& toDeepCopy)
+    {
+        m_context = toDeepCopy.m_context;
```
yao-msft, line 68: This is duplicate code with line 23. Is it possible to make it a common method that works on IConfigurationEnvironmentView?

src/Microsoft.Management.Configuration/ConfigurationSetSerializer.h
```diff
@@ -34,16 +35,19 @@ namespace winrt::Microsoft::Management::Configuration::implementation
...
-        void WriteYamlValue(AppInstaller::YAML::Emitter& emitter, const winrt::Windows::Foundation::IInspectable& value);
-        void WriteYamlValueIfNotEmpty(AppInstaller::YAML::Emitter& emitter, ConfigurationField key, const winrt::Windows::Foundation::IInspectable& value);
-        void WriteYamlStringValueIfNotEmpty(AppInstaller::YAML::Emitter& emitter, ConfigurationField key, hstring value);
+        static void WriteYamlValue(AppInstaller::YAML::Emitter& emitter, const winrt::Windows::Foundation::IInspectable& value);
+        static void WriteYamlValueIfNotEmpty(AppInstaller::YAML::Emitter& emitter, ConfigurationField key, const winrt::Windows::Foundation::IInspectable& value);
+        static void WriteYamlStringValueIfNotEmpty(AppInstaller::YAML::Emitter& emitter, ConfigurationField key, hstring value);
 
         std::wstring_view GetSchemaVersionCommentPrefix();
+
+    private:
```
yao-msft, line 50: nit: not needed

yao-msft (dismissed):

src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
```diff
@@ -393,6 +443,12 @@ namespace Microsoft.Management.Configuration
             // Setting SchemaUri to a different value will change SchemaVersion.
             Windows.Foundation.Uri SchemaUri;
         }
+
+        [contract(Microsoft.Management.Configuration.Contract, 3)]
+        {
+            // Gets the union of environments as defined by all of the active units within the set.
+            Windows.Foundation.Collections.IVector<IConfigurationEnvironmentView> GetUnitEnvironments();
```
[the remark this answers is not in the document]
JohnMcPMS: A view would be a read only collection, but I'm giving back a collection that the caller now owns.  They can change it all they want, but it contains views that they cannot change.

But this has made me reconsider having `IConfigurationEnvironmentView` in the first place, which was originally part of a more complex change.  I don't think it is needed any more and could even lead to some incorrect assumptions given how things are implemented.  I'm going to return a vector of environments that are copies of the ones they represent and remove the environment view completely.

PR #5140, opened 2025-01-15
Update with new Store certificate

JohnMcPMS:
## Change
Rotate the new Store certificate in and the old one out.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5140)


yao-msft (approved):

yao-msft:
I don't think we addressed this yet? https://github.com/microsoft/winget-cli/pull/3435
Either we swap the cert1 and cert 2 to delay the test errors Or we'll do the real fix before cert2 expires.

JohnMcPMS:
> I don't think we addressed this yet? #3435 Either we swap the cert1 and cert 2 to delay the test errors Or we'll do the real fix before cert2 expires.

I always rotate Leaf2 to Leaf1 and make the new certificate Leaf2.  While we should fix that issue, I think it is fairly low priority.  If we ever stop updating the pinning certificates, we also probably stopped supporting the product generally.

PR #5156, opened 2025-01-27
Add support to export all installed packages in winget configure export

yao-msft:
The feature is in parity with `winget export` command. Except the exported file is in winget configuration yaml format. The exported file is ready to be used with winget configure commands.

New additions:
- When exporting a specific package (existing functionality), in addition we'll search the source to find the package before export. If the package is not from well known source, the source is exported as well. Supports exporting the version like `winget export` too.
- Added `--all` to export all installed packages. We reuse the workfow in `winget export` to collect packages to export. For non well known sources, the sources are exported as well. `--all` cannot be used with other specific package export arguments.

Manually validated and added e2e tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5156)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
@@ -1084,7 +1084,6 @@ namespace AppInstaller::CLI::Workflow
             ValueSet directives;
             directives.Insert(s_Directive_Module, PropertyValue::CreateString(s_Module_WinGetClient));
             directives.Insert(s_Directive_Description, PropertyValue::CreateString(winrt::to_hstring(description.get())));
-            directives.Insert(s_Directive_AllowPrerelease, PropertyValue::CreateBoolean(true));
             unit.Metadata(directives);
 
             ValueSet settings;
@@ -1113,7 +1112,6 @@ namespace AppInstaller::CLI::Workflow
             ValueSet directives;
             directives.Insert(s_Directive_Module, PropertyValue::CreateString(s_Module_WinGetClient));
             directives.Insert(s_Directive_Description, PropertyValue::CreateString(winrt::to_hstring(description.get())));
-            directives.Insert(s_Directive_AllowPrerelease, PropertyValue::CreateBoolean(true));
             unit.Metadata(directives);
 
             ValueSet settings;


src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1064,36 +1069,73 @@ namespace AppInstaller::CLI::Workflow
...
+            unit.Type(s_Unit_WinGetSource);
+            unit.Identifier(sourceUnitIdWide);
+            unit.Intent(ConfigurationUnitIntent::Apply);
+
+            auto description = Resource::String::ConfigureExportUnitDescription(Utility::LocIndView{ sourceUnitId });
+
+            ValueSet directives;
+            directives.Insert(s_Directive_Module, PropertyValue::CreateString(s_Module_WinGetClient));
+            directives.Insert(s_Directive_Description, PropertyValue::CreateString(winrt::to_hstring(description.get())));
+            directives.Insert(s_Directive_AllowPrerelease, PropertyValue::CreateBoolean(true));
```
JohnMcPMS, line 1087: Remove prerelease; should not be needed now.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1064,36 +1069,73 @@ namespace AppInstaller::CLI::Workflow
...
+            auto description = Resource::String::ConfigureExportUnitInstallDescription(Utility::LocIndView{ package.Id });
 
-                ValueSet settings;
-                settings.Insert(s_Setting_Id, PropertyValue::CreateString(packageIdWide));
-                settings.Insert(s_Setting_Source, PropertyValue::CreateString(s_WinGetSource));
-                unit.Settings(settings);
+            ValueSet directives;
+            directives.Insert(s_Directive_Module, PropertyValue::CreateString(s_Module_WinGetClient));
+            directives.Insert(s_Directive_Description, PropertyValue::CreateString(winrt::to_hstring(description.get())));
+            directives.Insert(s_Directive_AllowPrerelease, PropertyValue::CreateBoolean(true));
```
JohnMcPMS, line 1116: Remove

src/AppInstallerCLICore/ConfigureExportCommand.cpp
```diff
@@ -39,26 +43,30 @@ namespace AppInstaller::CLI
     {
         context <<
             VerifyIsFullPackage <<
+            SearchSourceForPackageExport <<
             CreateConfigurationProcessor <<
```
JohnMcPMS, line 47:
```suggestion
            CreateConfigurationProcessorWithoutFactory <<
```
The current code launches another process for the factory and then never uses it (at least from my reading).  This will speed things up a lot.

yao-msft: Discussed with John, we'll leave it as is. Other configuration resources need it and we may also explore exporting configurations related to exported packages later.

JohnMcPMS (approved):

PR #5158, opened 2025-01-28
Update Newtonsoft.Json to be same version across all projects

Trenly:

According to the NuGet management tools in Visual Studio, `Newtonsoft.Json` had different versions being used across different projects within the solution. This PR bumps them all to the same version, moving some from `13.0.2`  up to `13.0.3`. In the cases where the version specification was added, it is due to the package being a transient dependency

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5158)

### DIFF
--- a/src/ConfigurationRemotingServer/ConfigurationRemotingServer.csproj
+++ b/src/ConfigurationRemotingServer/ConfigurationRemotingServer.csproj
@@ -27,6 +27,10 @@
   <Import Project="..\targets\ReferenceEmbeddedCsWinRTProject.targets" />
 
   <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\Microsoft.Management.Configuration.Processor\Microsoft.Management.Configuration.Processor.csproj" />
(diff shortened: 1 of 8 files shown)


Trenly:
Resolved merge confilct

src/ConfigurationRemotingServer/ConfigurationRemotingServer.csproj
```diff
@@ -26,6 +26,10 @@
 
   <Import Project="..\targets\ReferenceEmbeddedCsWinRTProject.targets" />
 
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+  </ItemGroup>
+
```
yao-msft, line 32: This one is probably not needed.

Trenly: I believe that Visual Studio added it it to prevent different versions within transient dependencies

yao-msft: Ah, then no worry. Thanks for clarification.

PR #5170, opened 2025-01-30
Formatting changes for localized content

DuckDuckStudio:

### Harmonization of `:` in zh-CN and zh-TW localization
`： ` → `: `
`：` → `: `

### Completion of missing left and right spaces
For example:
`...将使用Windows 程序包管理器设置中指定的...` → `...将使用 Windows 程序包管理器设置中指定的...`

**See comments and commit messages below for other changes.**

-----

- Superseded by #5192

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5170)

### DIFF
--- a/Localization/Resources/zh-CN/winget.resw
+++ b/Localization/Resources/zh-CN/winget.resw
@@ -126,11 +126,11 @@
     <comment>{Locked="{0}"} Error message displayed when the user provides an adjoined flag alias argument that was not found. {0} is a placeholder replaced by the user input argument (e.g. '-ab').</comment>
   </data>
   <data name="AvailableArguments" xml:space="preserve">
-    <value>以下参数可用：</value>
+    <value>以下参数可用: </value>
     <comment>Message displayed to inform the user about the available command line arguments.</comment>
   </data>
   <data name="AvailableCommandAliases" xml:space="preserve">
-    <value>以下命令别名可用：</value>
+    <value>以下命令别名可用: </value>
     <comment>Message displayed to inform the user about the available command line alias arguments.</comment>
   </data>
   <data name="AvailableCommands" xml:space="preserve">
@@ -142,11 +142,11 @@
     <comment>As in "a new version is available to upgrade to".</comment>
   </data>
   <data name="AvailableOptions" xml:space="preserve">
-    <value>下列选项可用：</value>
+    <value>下列选项可用: </value>
     <comment>Message displayed to inform the user about the available options.</comment>
   </data>
   <data name="AvailableSubcommands" xml:space="preserve">
-    <value>以下子命令可用：</value>
+    <value>以下子命令可用: </value>
     <comment>Message displayed to inform the user about the available nested commands that run in context of the selected command.</comment>
   </data>
   <data name="AvailableUpgrades" xml:space="preserve">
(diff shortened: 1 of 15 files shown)


Trenly:
These changes will get overwritten any time there is a new localization patch from the translations team. These changes need to be made to the localization team's internal process I believe

JohnMcPMS:
> These changes will get overwritten any time there is a new localization patch from the translations team. These changes need to be made to the localization team's internal process I believe

I forwarded it to the localization team (Bug 56032525 for MS internal tracking).  These choices seem stylistic to me (a non-localization expert without any experience with any of the languages involved).  Given that, the localization team may not accept them as they may have internal guidelines governing these choices.

Localization/Resources/zh-CN/winget.resw
```diff
@@ -3147,14 +3147,14 @@
     <value>家庭</value>
   </data>
   <data name="FontFaces" xml:space="preserve">
-    <value>脸</value>
+    <value>Faces</value>
```
JohnMcPMS, line 3150: This seems like a mistake (and the one below and one in the zh-TW file).  I can't imagine that using English here is correct.

Using bing translation, the current translation might be literally "face", like the part of a human being containing the eyes, nose and mouth.  It suggests that `字体`, `字样`, or `字面` might be better fits for "typeface".  @yao-msft might be able to weigh in.

DuckDuckStudio: I think it should be translated as "字体样式" (or "字样"), but I don't have the context to know where they appear, and there's another `Face` in the document (line 3157), so I'm using the original English.
In any case, "脸" is definitely an incorrect translation.
In the zh-TW translation, it is translated as "表情" in one place (line 3150), but I think this is also incorrect based on the `<comment>`.

Localization/Resources/zh-CN/winget.resw
```diff
@@ -3147,14 +3147,14 @@
...
     <comment>"Faces" represents the typeface of the font family such as 'Bold' or 'Italic'</comment>
   </data>
   <data name="FontFamilyNameArgumentDescription" xml:space="preserve">
     <value>按家庭名称筛选结果</value>
```
DuckDuckStudio, line 3154:
```suggestion
    <value>按字体集名称筛选结果</value>
```

Just reacted. And this.

DuckDuckStudio:
Is a `}` missing here?
image

JohnMcPMS:
> Is a `}` missing here?

Yes, but if those comments are programmatically consumed they didn't complain at us.  It would need to be fixed in: [/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw)

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2662,7 +2662,7 @@ Please specify one of them using the --source option to proceed.</value>
   </data>
   <data name="ResumeIdNotFoundError" xml:space="preserve">
     <value>The resume state does not exist: {0}</value>
-    <comment>{Locked="{0}" Error message displayed when the user provides a guid that does not correspond to a valid saved state. {0} is a placeholder replaced by the provided guid string.</comment>
+    <comment>{Locked="{0}"} Error message displayed when the user provides a guid that does not correspond to a valid saved state. {0} is a placeholder replaced by the provided guid string.</comment>
```
yao-msft, line 2665: If you want, can you make a new pr with only this file's change? In the end, this pr is only a reference for the localization team for translation suggestions. It will be discarded after Localization team makes decisions on the translation suggestions.

DuckDuckStudio: - #5192

PR #5171, opened 2025-01-30
Add Allowed Architectures to FindPackagesOptions Com Api

yao-msft:
Interface change only. We are cutting a release candidate soon and we do not want 2 contract versions across servicing releases. So we'll get the interface changes in first. Implementation will follow later before the official 1.10 release.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5171)


yao-msft:
Not needed.

PR #5178, opened 2025-02-04
Use embedded CsWinRT

JohnMcPMS:
## Issue
Updating the version of CsWinRT that we use lead to a problem with conflicting WinRT.Runtime assemblies (microsoft/winget-dsc#140).  While that issue was mitigated by delisting the offending version, in order to ship any release with the newer version of CsWinRT, we need to embed the WinRT.Runtime in our assemblies.

## Change
Move to use embedded CsWinRT, specifically in `Microsoft.WinGet.Client.Engine` and `Microsoft.Management.Configuration.Processor`.  The dependent projects are updated to no longer reference CsWinRT, but requires some special handling to prevent errors.

Since the configuration code was leveraging a projection assembly, it is being replaced by the processor.  The projected types are shared out to those dependent binaries via `InternalsVisibleTo`.

Also fixes an annoying behavior where an exception escaping from a command execution would prevent the `--logs` option from opening the log location (error cases being more likely to be the time one would want the logs of course).

## Validation
Minor local validation that things work. Build regression tests will be the true test.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5178)

### DIFF
--- a/src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
+++ b/src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
@@ -73,29 +73,23 @@
     <CsWinRTComponent>false</CsWinRTComponent>
     <CsWinRTIncludes>
       Microsoft.Management.Configuration;
-      Windows.ApplicationModel;
-      Windows.Data;
-      Windows.Devices;
+      Windows.Data.Text.TextSegmen;
+      Windows.Devices.Geolocation;
       Windows.Foundation;
-      Windows.Globalization;
-      Windows.Graphics;
-      Windows.Management;
-      Windows.Media;
-      Windows.Networking;
-      Windows.Perception;
-      Windows.Security;
-      Windows.Services;
+      Windows.Globalization.DayOfWee;
+      Windows.Networking.Connectivity;
+      Windows.Networking.DomainNameTyp;
+      Windows.Networking.EndpointPai;
+      Windows.Networking.IEndpointPai;
+      Windows.Networking.HostNam;
+      Windows.Networking.IHostNam;
+      Windows.Security.Cryptography.Certificates;
       Windows.Storage;
-      Windows.System;
-      Windows.UI;
-      Windows.Web;
+      Windows.System.Use;
+      Windows.System.IUse;
     </CsWinRTIncludes>
     <CsWinRTExcludes>
-      Windows.ApplicationModel.Store.Preview;
-      Windows.Devices.PointOfService;
       Windows.Foundation.PropertyType;
-      Windows.UI.Input.Preview;
-      Windows.UI.Xaml;
     </CsWinRTExcludes>
     <!-- Ensure Support for Windows 10, Version 1809 -->
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
--- a/src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
+++ b/src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
@@ -22,7 +22,7 @@
     !!! Remove or update this on the next Microsoft.Windows.CsWinRT package version update. !!!
     -->
     <WindowsSdkPackageVersion>10.0.22000.53</WindowsSdkPackageVersion>
-    <WinGetCsWinRTEmbedded>true</WinGetCsWinRTEmbedded>
+    <WinGetCsWinRTEmbedded Condition="'$(WinGetCsWinRTEmbedded)'==''">true</WinGetCsWinRTEmbedded>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)'=='Release'">
@@ -31,8 +31,6 @@
 
   <PropertyGroup Condition="'$(Configuration)'=='ReleaseStatic'">
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <!-- From the name, ReleaseStatic would be a better fit for embedded. However, we will use embedded at all times except the inproc nuget package release, so it is better if it is in this configuration. -->
-    <WinGetCsWinRTEmbedded>false</WinGetCsWinRTEmbedded>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
+++ b/src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
@@ -77,7 +77,6 @@
       Windows.Data;
       Windows.Devices;
       Windows.Foundation;
-      Windows.Gaming;
       Windows.Globalization;
       Windows.Graphics;
       Windows.Management;
@@ -92,7 +91,11 @@
       Windows.Web;
     </CsWinRTIncludes>
     <CsWinRTExcludes>
+      Windows.ApplicationModel.Store.Preview;
+      Windows.Devices.PointOfService;
       Windows.Foundation.PropertyType;
+      Windows.UI.Input.Preview;
+      Windows.UI.Xaml;
     </CsWinRTExcludes>
     <!-- Ensure Support for Windows 10, Version 1809 -->
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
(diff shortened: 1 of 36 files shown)


src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
```diff
@@ -52,14 +53,53 @@
...
+  <PropertyGroup Condition="'$(WinGetCsWinRTEmbedded)'=='true'">
+    <DefineConstants>$(DefineConstants);WinGetCsWinRTEmbedded</DefineConstants>
+    <CsWinRTComponent>false</CsWinRTComponent>
+    <CsWinRTIncludes>
+      Microsoft.Management.Configuration;
+      Windows.ApplicationModel;
+      Windows.Data;
+      Windows.Devices;
+      Windows.Foundation;
+      Windows.Gaming;
```
Trenly, line 80: Out of curiosity - why do we need `Windows.Gaming`?

JohnMcPMS: It appears to be in the closure of things we need.  The list was generated by iteratively building and adding namespaces until the unrecognized namespace/type errors stopped.  It is possible that a more targeted set of types could be determined to prevent quite so much from being generated.

I will give that a try to see if it isn't too painful, as it will reduce build time and potentially allow us to not need to set up trimming.

JohnMcPMS: Nope, very painful.

I did however try removing just `Windows.Gaming` and found that it was `Windows.UI.Input.Preview.Injection` that was pulling it in.  Since we can selectively exclude things, this pruning seems like a more sane way to get rid of some of the excess.

Trenly: Well, 25% reduction isn't too bad for just trimming off one small portion. Thanks!!

Trenly: Does the same optimization apply to Engine, or is that one more painful?

src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
```diff
@@ -52,14 +53,56 @@
...
+      Windows.ApplicationModel;
+      Windows.Data;
+      Windows.Devices;
+      Windows.Foundation;
+      Windows.Globalization;
+      Windows.Graphics;
+      Windows.Management;
+      Windows.Media;
+      Windows.Networking;
+      Windows.Perception;
```
florelis, line 85: In the same line as @Trenly's comment, could do without the Windows.Perception namespace, and maybe Windows.Graphics?

src/Microsoft.Management.Configuration.Processor/Public/PowerShellConfigurationSetProcessorFactory.cs
```diff
@@ -15,12 +15,17 @@ namespace Microsoft.Management.Configuration.Processor
...
 
     /// <summary>
     /// ConfigurationSetProcessorFactory implementation.
     /// </summary>
-    public sealed class PowerShellConfigurationSetProcessorFactory : IConfigurationSetProcessorFactory, IPowerShellConfigurationProcessorFactoryProperties
+#if WinGetCsWinRTEmbedded
+    internal sealed class PowerShellConfigurationSetProcessorFactory : IConfigurationSetProcessorFactory, IPowerShellConfigurationProcessorFactoryProperties, IPwshConfigurationSetProcessorFactoryProperties
+#else
+    public sealed class PowerShellConfigurationSetProcessorFactory : IConfigurationSetProcessorFactory, IPowerShellConfigurationProcessorFactoryProperties, IPwshConfigurationSetProcessorFactoryProperties
+#endif
```
florelis, line 28: Nit: This may be worse but if the access modifier is the only thing that changes, could we have only that be conditional? Not sure if C# would allow it

```C#
#if WinGetCsWinRTEmbedded
    internal
#else
    public
#endif
         sealed class PowerShellConfigurationSetProcessorFactory : IConfigurationSetProcessorFactory, IPowerShellConfigurationProcessorFactoryProperties, IPwshConfigurationSetProcessorFactoryProperties
```

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Microsoft.WinGet.Configuration.Engine.csproj
```diff
@@ -41,15 +36,8 @@
...
-  <!-- This project doesn't reference CSWinRT, but specify this everywhere it will fail with NETSDK1130 and MSB3271. -->
-  <PropertyGroup>
-    <ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
-  </PropertyGroup>
-  
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.1.6" />
-  </ItemGroup>
+
+  <Import Project="..\..\targets\ReferenceEmbeddedCsWinRTProject.targets" />
```
florelis, line 40: I assume none of these are affected when the Processor is not using Embedded?

florelis (dismissed):

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Microsoft.WinGet.Configuration.Engine.csproj
JohnMcPMS: The only use case for that build is the nuget; these projects are not intended to ever ship with it after the merge.

src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
JohnMcPMS: In terms of size reduction, I am assuming that generated code file size translates linearly to resulting binary size.  The Xaml file(s) were drastically larger than any other, so its removal resulted in the largest drop.  The remaining namespaces are all a more gradual size gradient.  The largest ones would result in ~500 KB each if linear (current size is ~24 MB, which is already a size savings over the SDK assembly we get to leave out).

I tried removing some of them, but it takes around a minute to iterate and there are some things in each of these namespaces that are required.  The best thing someone could do would be to write code to analyze the SDK use of a non-embedded build, then output the list of types needed.  Presumably a huge list of types would be faster and a perfect pruning for output size.

JohnMcPMS:
@Trenly and @florelis 

I took another run at building up the types required and was able to get the output down to 2.7 MB (and build time to a few seconds rather than a minute.  Some lessons:

1. Just include all of `Windows.Foundation` to start
2. The inclusion filtering seems to require that the type be *longer* than the string given, so you have to cut off at least one character to get a specific type
3. This will definitely be a fragile list in the event of new code or a CsWinRT version update

But given the massive build time and binary size savings it seems to be worth the effort.

src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
```diff
@@ -73,29 +73,23 @@
...
       Windows.Foundation;
-      Windows.Globalization;
-      Windows.Graphics;
-      Windows.Management;
-      Windows.Media;
-      Windows.Networking;
-      Windows.Perception;
-      Windows.Security;
-      Windows.Services;
+      Windows.Globalization.DayOfWee;
```
florelis, line 79: I love this. It sounds so festive to me. "Day of weeeeee~"
But brits may disagree with my interpretation 😆

florelis (approved):

src/PowerShell/Microsoft.WinGet.Client.Engine/Microsoft.WinGet.Client.Engine.csproj
```diff
@@ -98,6 +97,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == '$(DesktopFramework)'">
+    <LangVersion>10</LangVersion>
```
yao-msft, line 100: Is the move from global to DesktopFramework only intentional?

yao-msft (approved):

src/PowerShell/Microsoft.WinGet.Client.Engine/Microsoft.WinGet.Client.Engine.csproj
JohnMcPMS: Yes, as the embedded code requires a higher version.

PR #5182, opened 2025-02-05
Update configuration environment behavior

JohnMcPMS:
## Change
In response to some other discussions, update the configuration environment behavior:
1. Add an environment at the set level
2. Don't inherit environment data into child units, and don't promote environment data when serializing

Updated the one consumer of environment (dynamic factory) to be responsible for the flow of the security context from the set to the immediate child units.

Also improved serialization of groups to output the non-resource properties.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5182)

### DIFF
--- a/src/Microsoft.Management.Configuration.Processor/Extensions/ValueSetExtensions.cs
+++ b/src/Microsoft.Management.Configuration.Processor/Extensions/ValueSetExtensions.cs
@@ -9,6 +9,8 @@ namespace Microsoft.Management.Configuration.Processor.Extensions
     using System;
     using System.Collections;
     using System.Collections.Generic;
+    using System.Text;
+    using System.Xml.Linq;
     using Windows.Foundation.Collections;
 
     /// <summary>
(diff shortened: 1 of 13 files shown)


src/Microsoft.Management.Configuration.Processor/Extensions/ValueSetExtensions.cs
```diff
@@ -175,5 +177,61 @@ public static bool ContentEquals(this ValueSet first, ValueSet second)
 
             return true;
         }
+
+        /// <summary>
+        /// Converts the value set to YAML like output.
+        /// </summary>
+        /// <param name="set">The set to output.</param>
+        /// <returns>The string.</returns>
+        public static string ToYaml(this ValueSet set)
```
yao-msft, line 186: I saw this is only used in tests. Is it expected to be used in product code in the future? If not, shall we move the code to test common?

src/Microsoft.Management.Configuration/Microsoft.Management.Configuration.idl
```diff
@@ -432,6 +432,11 @@ namespace Microsoft.Management.Configuration
 
         [contract(Microsoft.Management.Configuration.Contract, 3)]
         {
+            // The environment in which to process the configuration set.
```
yao-msft, line 435: How is the configuration contract version defined? Is it tied to the schema version? (i.e. not tied to our releases since it's still in preview only?)

yao-msft (dismissed):

src/Microsoft.Management.Configuration.Processor/Extensions/ValueSetExtensions.cs
JohnMcPMS: I honestly didn't notice that this was in the processor; I just assumed that it was in the unit tests as that is how I navigated there (via F12 from a test).

yao-msft (approved):

PR #5188, opened 2025-02-06
Convert cpprestsdk to WIL exception for better handling

JohnMcPMS:
## Change
To better expose the underlying result code, convert the `http_exception` to a `ResultException` so that the HRESULT can be properly extracted.

## Validation
Using the PowerShell module and interfering with my connection, I was able to see that the connect result's extended error was originally `8007023E`, but changed to the appropriate `80072EE2` with the change.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5188)

### DIFF
--- a/src/AppInstallerCommonCore/HttpClientHelper.cpp
+++ b/src/AppInstallerCommonCore/HttpClientHelper.cpp
@@ -242,7 +242,7 @@ namespace AppInstaller::Http
         return response.extract_json().get();
     }
 
-    [[noreturn]] void HttpClientHelper::RethrowAsWilException(web::http::http_exception& exception)
+    [[noreturn]] void HttpClientHelper::RethrowAsWilException(const web::http::http_exception& exception)
     {
         THROW_WIN32_MSG(exception.error_code().value(), "%hs", exception.what());
     }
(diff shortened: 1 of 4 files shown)


src/AppInstallerCommonCore/HttpClientHelper.cpp
```diff
@@ -233,4 +241,9 @@ namespace AppInstaller::Http
 
         return response.extract_json().get();
     }
+
+    [[noreturn]] void HttpClientHelper::RethrowAsWilException(web::http::http_exception& exception)
```
yao-msft, line 245: Should it be "const &" ?

yao-msft (approved):

src/AppInstallerCommonCore/HttpClientHelper.cpp
```diff
@@ -115,6 +115,10 @@ namespace AppInstaller::Http
 
         return ValidateAndExtractResponse(httpResponse);
     }
+    catch (web::http::http_exception& exception)
```
yao-msft, line 118: nit: the 2 catches here and below could be const too

yao-msft (approved):

src/AppInstallerCommonCore/HttpClientHelper.cpp
JohnMcPMS: Putting const on the catch is like putting const on a by-value parameter.  It is only there to stop you from changing something you have decided you don't want to change accidentally.  I will leave this one.

Issue #5211, opened 2025-02-16
Error 0x80070057 when running winget upgrade or list commands


uaiqop:
### Brief description of your issue

The error `0x80070057` occurs when running `winget upgrade` or `winget list` commands.

log:
[WinGet-2025-02-16-13-21-43.919.log](https://github.com/user-attachments/files/18813758/WinGet-2025-02-16-13-21-43.919.log)

### Steps to reproduce

**Steps to reproduce**  
1. Open Windows Terminal **as Administrator**.  
2. Run `winget upgrade --verbose` → Error `0x80070057` occurs.  
3. Run `winget list --verbose` → Same error appears.  
4. Reset winget v2, but the error still persists.

```
 ⚡user ❯❯ winget upgrade --verbose
执行此命令时发生意外错误：
0x80070057 : ��������
 ⚡user ❯❯ winget list --verbose
执行此命令时发生意外错误：
0x80070057 : ��������
```

### Expected behavior

**Under normal circumstances:**
- Running `winget upgrade` should list available updates for installed packages.
- Running `winget list` should display all installed packages without errors.
- Both commands should complete successfully without returning error code `0x80070057`.

### Actual behavior

**When running the commands:**
- winget upgrade → Fails to list updates and returns error code 0x80070057.
- winget list → Fails to display installed packages and throws the same error 0x80070057.

```
 ⚡user ❯❯ winget upgrade --verbose
执行此命令时发生意外错误：
0x80070057 : ��������
 ⚡user ❯❯ winget list --verbose
执行此命令时发生意外错误：
0x80070057 : ��������
```

### Environment

```shell
Windows 程序包管理器 v1.10.300
版权所有 (C) Microsoft Corporation。保留所有权利。

Windows: Windows.Desktop v10.0.26120.3281
系统体系结构: X64
软件包: Microsoft.DesktopAppInstaller v1.25.300.0

Winget 目录
-------------------------------------------------------------------------------------------------------------------
日志                   %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir
用户设置               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\settings.json
可移植链接目录(用户)   %LOCALAPPDATA%\Microsoft\WinGet\Links
可移植链接目录(计算机) C:\Program Files\WinGet\Links
[...]

PR #5213, opened 2025-02-16
Don't rely on parameter order when creating VersionRange

Trenly:
  - #5211

I don't know whether or not this will resolve the above issue, but I also don't think it can hurt?

Instead of relying on the parameters to be in the correct order for minimum and maximum when creating a VersionRange, the code can instead just put them in the right order.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5213)

### DIFF
--- a/src/AppInstallerCLITests/Versions.cpp
+++ b/src/AppInstallerCLITests/Versions.cpp
@@ -372,12 +372,12 @@ TEST_CASE("VersionRange", "[versions]")
 {
     // Create
     REQUIRE_NOTHROW(VersionRange{ Version{ "1.0" }, Version{ "2.0" } });
+    REQUIRE_NOTHROW(VersionRange{ Version{ "1.0" }, Version{ "1.0" } });
     REQUIRE_NOTHROW(VersionRange{ Version{ "2.0" }, Version{ "1.0" } });
-    REQUIRE_THROWS(VersionRange{ Version{ "2.0" }, Version{ "2.0" } });
 
     // Overlaps
     REQUIRE(VersionRange{ Version{ "1.0" }, Version{ "2.0" } }.Overlaps(VersionRange{ Version{ "2.0" }, Version{ "3.0" } }));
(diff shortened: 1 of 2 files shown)


JohnMcPMS:
The thing I worry about with this is code that would have failed previously will now get through and may continue on with a flawed view of the range.

JohnMcPMS:
> I'd be interested to know how the values in the index were flip-flopped between the max and min

That is the real root issue (other than our continued discussions on the whether version ranges make any sense at all 😄).

yao-msft:
> > I'd be interested to know how the values in the index were flip-flopped between the max and min
> 
> That is the real root issue (other than our continued discussions on the whether version ranges make any sense at all 😄).

That's probably due to this change https://github.com/microsoft/winget-cli/pull/5001 where the version comparison was fixed to correctly compare versions. It's likely the index was created with an old wingetutil and the new client has fixed code. I think a rebuild of the index will fix the index.

Trenly:
> > > I'd be interested to know how the values in the index were flip-flopped between the max and min
> > 
> > 
> > That is the real root issue (other than our continued discussions on the whether version ranges make any sense at all 😄).
> 
> That's probably due to this change #5001 where the version comparison was fixed to correctly compare versions. It's likely the index was created with an old wingetutil and the new client has fixed code. I think a rebuild of the index (with latest wingetutil) will fix the index.

Ah, that makes sense; The old code would have parsed it as `6.1` instead of `6.1.0`, and in that case would have returned the incorrect version comparison

JohnMcPMS:
Given that this is the root cause, I'm inclined to say that the best resolution likely is to accept a change along these lines.  If we do rebuild, or in fact if a similar situation arises after this change makes it to the service, then <1.10 clients will fail similarly.  The good thing is that while I am concerned about the potential for some code to make a mistake that is harder to diagnose after this change, we can simply review current usage to see if that is happening.

JohnMcPMS:
On review, I didn't see any concerning usage.

Also, GH copilot will write the commit message for you when you edit here.  Fun.

yao-msft (approved):

PR #5233, opened 2025-02-20
Validate URLs map to unique hashes

### LINKED ISSUE #2852: SHA-256 collisions between different installer URLs, for different architectures, are not flagged as a manifest bug?

winget currently does not flag as a manifest warning /error when it encounters the same SHA-256 checksum for installers for different cpu architectures?

This can lead to 64-bit programs being "upgraded" to 32-bit versions.

### Steps to reproduce

have something like this in a package manifest: (same checksum, different installer URLs, different architectures)

```
Installers:
- Architecture: x64
  InstallerUrl: https://edge.dropboxstatic.com/dbx-releng/client/Dropbox%20165.4.4300%20Offline%20Installer.exe
  InstallerSha256: 6AFE5324E7D2C18D565B2E5BBD98853C8CCD44DA3F0DCE03608B54DBBBCA6D1F
- Architecture: x86
  InstallerUrl: https://edge.dropboxstatic.com/dbx-releng/client/Dropbox%20165.4.4300%20Offline%20Installer.x86.exe
  InstallerSha256: 6AFE5324E7D2C18D565B2E5BBD98853C8CCD44DA3F0DCE03608B54DBBBCA6D1F
```
also, see issue https://github.com/microsoft/winget-pkgs/issues/93678

### Expected behavior

at least show a warning message in the console about the same checksum being present in the manifest file, but for different CPU architectures.

### Actual behavior

because of the manifest specifying the same installer for different architectures, winget behaves "normally" and "upgrades" a 64-bit program to a 32-bit version.
This might be "normal" for some programs that detect the architecture directly in the installer, but not all of them do. Those that can detect this should have a warning-suppression flag in the manifest for it.

Case in point: the Dropbox installer from the manifest quoted above will happily "upgrade" a 64-bit version to a 32-bit one, because the winget manifest specifies the same file for both x86 and x64 architectures.

### Environment

```shell
> winget --info
Windows Package Manager (Preview) v1.5.101-preview

System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.101.0

Links
```

Trenly:
  - Resolves #2852

## Description:
This PR adds validation for URL to SHA256 mappings. It is presumed that a single URL should only return a single SHA256, so if a manifest has different SHA256 for the same URL, this will now throw a validation *error*. This will also help reduce potential load on the Validation Pipelines at winget-pkgs as the `InstallerSha256` mismatch will be caught during manifest validation and the validation will exit before it proceeds to installer validation.

Additionally, as noted in the initial issue, it would be unlikely (but possible) for two different URLS to point to the same SHA256. Validation was added for this case, but will only throw a validation *warning*. It is also necessary for this to be a warning so that this change does not become a breaking change.

{D367385B-6718-46F4-818D-4198F75D36C9}

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5233)


denelon:
Nice!

src/AppInstallerCLITests/YamlManifest.cpp
```diff
@@ -791,8 +791,10 @@ TEST_CASE("ReadBadManifests", "[ManifestValidation]")
         { "Manifest-Bad-DuplicateKey-DifferentCase-lower.yaml", "Duplicate field found in the manifest." },
         { "Manifest-Bad-DuplicateReturnCode-ExpectedCodes.yaml", "Duplicate installer return code found." },
         { "Manifest-Bad-DuplicateReturnCode-SuccessCodes.yaml", "Duplicate installer return code found." },
+        { "Manifest-Bad-DuplicateSha256.yaml", "Multiple Installer URLs found with  to the same InstallerSha256. Please ensure the accuracy of the URLs", true },
```
mdanish-kh, line 794:
```suggestion
        { "Manifest-Bad-DuplicateSha256.yaml", "Multiple Installer URLs found with a match  to the same InstallerSha256. Please ensure the accuracy of the URLs", true },
```

is this a classic case of Trenly forgetting to commit local changes? 😁

Trenly: Yep

Trenly: Its actually a case of Trenly being an idiot

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -223,6 +229,31 @@ namespace AppInstaller::Manifest
...
+                    std::string url = installer.Url;
+
+                    auto [urlIterator, urlInserted] = urlToChecksum.try_emplace(url, checksum);
+                    auto [checksumIterator, checksumInserted] = checksumToUrl.try_emplace(checksum, url);
+
+                    if (!urlInserted && urlIterator->second != checksum)
+                    {
+                        // If the URL was not inserted, and the value in the map does not match the current Sha256, then
+                        // a single URL corresponds to multiple SHA256 and an error should be thrown
+                        resultErrors.emplace_back(ManifestError::InconsistentInstallerHash);
```
JohnMcPMS, line 247:
```suggestion
                        resultErrors.emplace_back(ManifestError::InconsistentInstallerHash, "InstallerUrl", url);
```
I see that this may have caused a build error due to ambiguity.  I would just static cast `url` to `string_view` if that is the case.

Trenly: I'll take a look. It seemed to build just fine on my end, with `ManifestInstaller::string_t` being implicitly cast to `std::string`. I can try a static cast if necessary, though

Trenly:
I hope I wasn’t the one who broke it
```
Failed ExportTestPackageWithVersion [8 s]
  Error Message:
     Expected: True
  But was:  False
  Stack Trace:
     at AppInstallerCLIE2ETests.ConfigureExportCommand.ExportTestPackageWithVersion() in D:\a\1\s\src\AppInstallerCLIE2ETests\ConfigureExportCommand.cs:line 94
  Standard Output Messages:
 Starting command run. AICLI path: wingetdev.exe Command: configure export Parameters: --package-id AppInstallerTest.TestPackageExport --include-versions -o D:\a\1\Build.x64release\AppInstallerCLIE2ETests\WorkDirectory\23pgi1ni.pnp\exported.yml Timeout: 60000
 Command run completed with exit code: 0
 Command run output. Output:
 Unexpected error while loading settings. Please verify your settings by running the 'settings' command.
 
    - 
                                                                                                                         
 
    - 
                                                                                                                         
 
    - Initializing configuration system
 ```

File: src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp [hunk 3]
```diff
@@ -27,6 +27,7 @@ namespace AppInstaller::Manifest
                 { AppInstaller::Manifest::ManifestError::FieldNotSupported, "Field is not supported."sv },
                 { AppInstaller::Manifest::ManifestError::FieldValueNotSupported, "Field value is not supported."sv },
                 { AppInstaller::Manifest::ManifestError::DuplicateInstallerEntry, "Duplicate installer entry found."sv },
+                { AppInstaller::Manifest::ManifestError::DuplicateInstallerHash, "Multiple Installer URLs found with the same InstallerSha256. Please ensure the accuracy of the URLs"sv },
```
yao-msft, line 30: nit: consistency here and below. All other messages end with '.' .

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -223,6 +229,31 @@ namespace AppInstaller::Manifest
                 {
                     resultErrors.emplace_back(ManifestError::FieldNotSupported, "ProductId");
                 }
+
+                // Ensure that each URL has a one to one mapping with a Sha256 and
+                // warn if a Sha256 has a one to many mapping with a URL
+                if (!installer.Url.empty() && !installer.Sha256.empty())
```
yao-msft, line 235: Should we do this only when fullValidation is true? The validation pipeline uses fullValidation. Some other manifest read scenarios do not use fullValidation to save time.

src/AppInstallerCLITests/AppInstallerCLITests.vcxproj
```diff
@@ -539,12 +539,18 @@
     <CopyFileToFolders Include="TestData\Manifest-Bad-DuplicateReturnCode-SuccessCodes.yaml">
       <DeploymentContent>true</DeploymentContent>
     </CopyFileToFolders>
+    <CopyFileToFolders Include="TestData\Manifest-Bad-DuplicateSha256.yaml">
+      <DeploymentContent>true</DeploymentContent>
+    </CopyFileToFolders>
```
yao-msft, line 544: nit: missing corresponding entries in AppInstallerCLITests.vcxproj.filters

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
yao-msft: nit: also the messages in YamlManifest.cpp if I'm not being annoying..

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
Trenly: I think that may make sense

Trenly:
Aha. . .  Looks like I only checked the tests for the bad manifests, and didn't check the tests for the good ones

yao-msft (approved):

Issue #5087, opened 2024-12-18
WinGet supports installing PowerShell modules as a portable application


Trenly:
### Description of the new feature / enhancement

I want to be able to install PowerShell modules from an archive using WinGet. I'd like installed modules to be available in all versions of PowerShell with which they are compatible. I'd like to be able to specify to install in either the user scope or machine scope.

Issue #5086, opened 2024-12-18
WinGet supports .jar as a portable application


denelon:
### Description of the new feature / enhancement

I want to be able to install portable Java applications using WinGet.

The Java archive format .jar requires a JRE to be present on the system, and that comes with challenges associated with dependencies.

Issue #5085, opened 2024-12-18
WinGet supports .cmd as a portable application


denelon:
### Description of the new feature / enhancement

As an enterprise customer I want to support .cmd as a portable application format so I can add portable applications to my private WinGet source.

> [!NOTE]
> The WinGet Community Repository blocks these formats due to policy.

Issue #5084, opened 2024-12-18
WinGet supports .bat as a portable application


denelon:
### Description of the new feature / enhancement

As an enterprise customer I want to support .bat as a portable application format so I can add portable applications to my private WinGet source.

> [!NOTE]
> The WinGet Community Repository blocks these formats due to policy.

PR #5237, opened 2025-02-21
Add validation option for portable installer type in archives

Trenly:

## Description
For the `portable` installer type, the `winget-pkgs` repository has a policy of not allowing scripts as installers. This includes `.bat`, `.ps1`, and `.cmd` files.

In addition to the policy at `winget-pkgs`, the client itself is not designed to handle portable files that are not `.exe` types. Issues have been filed asking for WinGet to support these in the future, but it currently does not -
* https://github.com/microsoft/winget-cli/issues/5083
* https://github.com/microsoft/winget-cli/issues/5086
* https://github.com/microsoft/winget-cli/issues/5087

This PR adds a validation option which validates the `RelativeFilePath` within a zip file is always a `.exe`, but can be extended to include additional formats as the requests above are addressed. This was added as a validation option (similar to the Schema Header Validation) as private repositories may want to not have the file type restrictions that `winget-pkgs` does - as described in these issues -
* https://github.com/microsoft/winget-cli/issues/5084
* https://github.com/microsoft/winget-cli/issues/5085

This should allow for the validation pipeline at `winget-pkgs` to automatically reject manifests that contain `NestedInstallerFiles` that are not supported, while allowing maintainers of private repositories to suppress the restriction by turning the validation option to false.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5237)

### DIFF
--- a/src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
+++ b/src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
@@ -302,7 +302,7 @@ namespace AppInstaller::Manifest
 
                 std::set<std::string> commandAliasSet;
                 std::set<std::string> relativeFilePathSet;
-                const bool isPortable = installer.NestedInstallerType == InstallerTypeEnum::Portable;
+                bool isPortable = installer.NestedInstallerType == InstallerTypeEnum::Portable;
 
                 for (const auto& nestedInstallerFile : installer.NestedInstallerFiles)
                 {
@@ -337,7 +337,7 @@ namespace AppInstaller::Manifest
                     // If running full validation, check filetype
                     if (fullValidation && isPortable)
                     {
-                        if (fullPath.has_extension() && allowedPortableFiletypes.find(fullPath.extension()) == allowedPortableFiletypes.end())
+                        if (fullPath.has_extension() && s_AllowedPortableFiletypes.find(fullPath.extension()) == s_AllowedPortableFiletypes.end())
                         {
                             resultErrors.emplace_back(ManifestError::InvalidPortableFiletype, "RelativeFilePath", nestedInstallerFile.RelativeFilePath);
                         }
--- a/src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
+++ b/src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
@@ -71,6 +71,7 @@ namespace AppInstaller::Manifest
                 { AppInstaller::Manifest::ManifestError::SchemaHeaderManifestTypeMismatch , "The manifest type in the schema header does not match the ManifestType property value in the manifest."sv },
                 { AppInstaller::Manifest::ManifestError::SchemaHeaderManifestVersionMismatch, "The manifest version in the schema header does not match the ManifestVersion property value in the manifest."sv },
                 { AppInstaller::Manifest::ManifestError::SchemaHeaderUrlPatternMismatch, "The schema header URL does not match the expected pattern."sv },
+                { AppInstaller::Manifest::ManifestError::InvalidPortableFiletype, "The file type of the referenced file is not allowed."sv },
             };
 
             return ErrorIdToMessageMap;
@@ -301,6 +302,7 @@ namespace AppInstaller::Manifest
 
                 std::set<std::string> commandAliasSet;
                 std::set<std::string> relativeFilePathSet;
+                const bool isPortable = installer.NestedInstallerType == InstallerTypeEnum::Portable;
 
                 for (const auto& nestedInstallerFile : installer.NestedInstallerFiles)
                 {
@@ -330,6 +332,15 @@ namespace AppInstaller::Manifest
                     {
                         resultErrors.emplace_back(ManifestError::DuplicatePortableCommandAlias, "PortableCommandAlias");
                         break;
+                    }
+
+                    // If running full validation, check filetype
+                    if (fullValidation && isPortable)
+                    {
+                        if (fullPath.has_extension() && allowedPortableFiletypes.find(fullPath.extension()) == allowedPortableFiletypes.end())
+                        {
+                            resultErrors.emplace_back(ManifestError::InvalidPortableFiletype, "RelativeFilePath", nestedInstallerFile.RelativeFilePath);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


yao-msft:
Basically same comment as the other pr, I think fullValidation should just work instead of a new ValidateOption. (Just to be clear, any new ValidateOption require a winget service side code change to enable the option).
Regarding warning vs error, if there's no existing manifests that violate this rule, error seems ok. Otherwise, maybe use warning to prevent unexpected failures in winget service.

Trenly:
> Basically same comment as the other pr, I think fullValidation should just work instead of a new ValidateOption. (Just to be clear, any new ValidateOption require a winget service side code change to enable the option). Regarding warning vs error, if there's no existing manifests that violate this rule, error seems ok. Otherwise, maybe use warning to prevent unexpected failures in winget service.

Wouldn't ValidateOption be required in order for private repositories to be able to disable the validation of this, similar to how the schema header validation is?

yao-msft:
> > Basically same comment as the other pr, I think fullValidation should just work instead of a new ValidateOption. (Just to be clear, any new ValidateOption require a winget service side code change to enable the option). Regarding warning vs error, if there's no existing manifests that violate this rule, error seems ok. Otherwise, maybe use warning to prevent unexpected failures in winget service.
> 
> Wouldn't ValidateOption be required in order for private repositories to be able to disable the validation of this, similar to how the schema header validation is?
> 
> I'll do a run through of winget-pkgs locally using the dev build and verify that there are no manifests that would be impacted by this being an error

For all manifest read operations, fullValidation is false. So private repositories should be fine. ValidateOptions is for granular control of winget service to choose which validation are performed during different steps. Currently I don't see a need to have a special switch just for this specific validation.

Trenly:
I'll submit corrections to these 9 manifests -

image

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -71,13 +71,15 @@ namespace AppInstaller::Manifest
...
                 { AppInstaller::Manifest::ManifestError::SchemaHeaderManifestVersionMismatch, "The manifest version in the schema header does not match the ManifestVersion property value in the manifest."sv },
                 { AppInstaller::Manifest::ManifestError::SchemaHeaderUrlPatternMismatch, "The schema header URL does not match the expected pattern."sv },
+                { AppInstaller::Manifest::ManifestError::InvalidPortableFiletype, "The file type of the referenced file is not allowed."sv },
             };
 
             return ErrorIdToMessageMap;
         }
     }
-    std::vector<ValidationError> ValidateManifest(const Manifest& manifest, bool fullValidation)
+    std::vector<ValidationError> ValidateManifest(const Manifest& manifest, ManifestValidateOption& validateOption)
```
yao-msft, line 80: Can you revert all these changes back to (and update the callers)
 std::vector<ValidationError> ValidateManifest(const Manifest& manifest, bool fullValidation)
 
I feel the ValidationOptions is supposed to be processed at top level, this is the actual internal operation.

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -249,8 +251,10 @@ namespace AppInstaller::Manifest
         bool m_warningOnly;
     };
 
+    static const std::unordered_set<std::filesystem::path> allowedPortableFiletypes = {".exe"};
```
yao-msft, line 254:
```suggestion
    static const std::unordered_set<std::filesystem::path> allowedPortableFiletypes = { L".exe" };

```
nit: Windows uses wide strings

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -331,6 +334,15 @@ namespace AppInstaller::Manifest
                         resultErrors.emplace_back(ManifestError::DuplicatePortableCommandAlias, "PortableCommandAlias");
                         break;
                     }
+
+                    // If running full validation, check filetype
+                    if (fullValidation && isPortable)
+                    {
+                        if (fullPath.has_extension() && allowedPortableFiletypes.find(fullPath.extension()) == allowedPortableFiletypes.end())
```
yao-msft, line 341: Do we want to allow .ExE?

yao-msft: Sorry, never mind. After reading the implementation code, it only supports .exe.  But, the code gets the final name from either PortableCommandAlias or RelativeFilePath. So the alias should be checked against ".exe" as well (if it has extention).

Trenly: Since Windows is case-insensitive, I think it makes sense to force `.exe` here, though I could add a utility function for making paths lowercase for comparison. My main concern is avoiding the proliferation of having to have `.exe`, `.Exe`, `.eXe`, `.exE`, `.EXe`, `.ExE`, `.eXE`, and `.EXE` all in the allow list

yao-msft: Yes, I think we can leave the code as is.

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -249,6 +251,8 @@ namespace AppInstaller::Manifest
         bool m_warningOnly;
     };
 
+    static const std::unordered_set<std::filesystem::path> allowedPortableFiletypes = { L".exe" };
```
yao-msft, line 254: nit: AllowedPortableFiletypes   or   s_AllowedPortableFiletypes  for static const

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -301,6 +302,7 @@ namespace AppInstaller::Manifest
 
                 std::set<std::string> commandAliasSet;
                 std::set<std::string> relativeFilePathSet;
+                const bool isPortable = installer.NestedInstallerType == InstallerTypeEnum::Portable;
```
yao-msft, line 305: nit: I don't think the const is needed here
```suggestion
                bool isPortable = installer.NestedInstallerType == InstallerTypeEnum::Portable;

```

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
Trenly: Fixed

yao-msft (approved):

PR #5252, opened 2025-02-26
Experimental support for DSC v3 processing

JohnMcPMS:
## Change
Adds experimental support for DSC v3 processing of configurations.  When enabled, one can use DSC v3 instead of PowerShell DSC v2 by setting their processor to `dscv3` for the configuration like:
```yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
metadata:
  winget:
    processor: dscv3
resources:
<continue with schema 0.3 resource definitions>
```
This is currently on-par with DSC v2 in terms of functionality, except for 2 things:
1. It does not attempt to ensure dsc.exe is present.  It will find and use the preview MSIX packaged version of dsc.exe (Store ID: `9PCX3HX4HZ0Z`) if installed.  Otherwise, you can specify the `--processor-path` to dsc.exe.
2. Resources for DSC v3 must be present on the system.  There is not currently any mechanism to find and install new resources, as the paradigm is that they are part of the configurable item.  There is no special handling of DSC v2 resources in the configuration; that could potentially come later.

Also makes 0.3 schema not experimental and fixes configuration history for environments.

## Validation
Added unit and E2E tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5252)


src/AppInstallerCLICore/ConfigurationSetProcessorFactoryRemoting.cpp
```diff
@@ -298,9 +331,54 @@ namespace AppInstaller::CLI::ConfigurationRemoting
...
     {
-        return winrt::make<RemoteFactory>(useRunAs, properties, restrictions);
+        Utility::Version schemaVersion{ Utility::ConvertToUTF8(set.SchemaVersion()) };
+
+        if (schemaVersion <= Utility::Version{ "0.3" })
+        {
+            ProcessorEngine result = ProcessorEngine::Unknown;
+
+            winrt::hstring processorIdentifier = set.Environment().ProcessorIdentifier();
+            if (processorIdentifier.empty() || processorIdentifier == L"pwsh")
```
yao-msft, line 350: Do we want to be more permissive by doing case insensitive checks?

yao-msft: nit: btw, the tests in ConfigurationSetAuthoringTests.cs used dsc3 as ProcessorIdentifier,  we may want to update those to avoid confusion.

yao-msft (approved):

yao-msft (approved):

Gijsreyn:
@JohnMcPMS, nice work! I really love seeing the content coming and WinGet's integration with DSC v3. Perhaps a tiny tiny remark. I can see that you've implemented a workaround in the azure-pipelines to install `dsc.exe` from the `winget` source. If I'm not mistaken, that will install the portable version, which is not the latest rc.1 version. I couldn't get the latest version in the winget-pkgs, because the .msixbundle was not added in the assets (see [discussion](https://github.com/PowerShell/DSC/discussions/638)).

PR #5275, opened 2025-03-05
Add max nest level limit to yaml parser

yao-msft:
This fixes an oom issue found in fuzzing run.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5275)


JohnMcPMS (approved):
I think you can achieve this with recursive references as well.  I would guess that is what fuzzing stumbled into.

yao-msft:
> I think you can achieve this with recursive references as well. I would guess that is what fuzzing stumbled into.

Yes, fuzzing test failed on a recursive reference.

PR #5279, opened 2025-03-07
Update winget.resw

pressRtowin:
Pedantic grammatical fixes.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5279)

### DIFF
--- a/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
+++ b/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
@@ -171,7 +171,7 @@
     <value>This command requires administrator privileges to execute.</value>
   </data>
   <data name="CompleteCommandLongDescription" xml:space="preserve">
-    <value>This command can be used to request context sensitive command line completion. The command line, cursor position, and word to be completed are passed in. The output is a set of potential values based on the inputs, one possible value per line.</value>
+    <value>This command can be used to request context sensitive command line completion. The command line, cursor position, and word to be completed are passed in. The output is a set of potential values based on the inputs, with one possible value per line.</value>
   </data>
   <data name="CompleteCommandShortDescription" xml:space="preserve">
     <value>Enables context sensitive command line completion</value>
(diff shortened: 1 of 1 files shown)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1597,7 +1597,7 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Version to which to pin the package. The wildcard '*' can be used as the last version part</value>
   </data>
   <data name="PinAddBlockingArgumentDescription" xml:space="preserve">
-    <value>Block from upgrading until the pin is removed, preventing override arguments</value>
+    <value>Block from upgrading until the pin is removed, preventing override arguments.</value>
```
Trenly, line 1600: Why only here and not the next line, and every other description?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2147,7 +2147,7 @@ Please specify one of them using the --source option to proceed.</value>
     <value>The source type is invalid</value>
   </data>
   <data name="APPINSTALLER_CLI_ERROR_PACKAGE_IS_BUNDLE" xml:space="preserve">
-    <value>The MSIX file is a bundle, not a package</value>
+    <value>The MSIX file is a bundle, not a package.</value>
```
florelis, line 2150: Seems like most other strings for error codes don't end with a period, but some do. Not sure which one is better, but if we're changing it we should make it consistent across the board

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
florelis: For the context: All the ones with name ending in `ArgumentDescription` are what is shown when you do `--help` for a command. Not sure if we should have a period or not in those, but it should be consistent since it's a pretty well defined category

florelis (commented):
Most changes seem good to me, except the adding of periods to some strings. I would prefer to have a larger change that standardizes that for all argument descriptions / error codes (or any other category where it makes sense). I'll approve if you undo these two changes so that we can have a separate PR with all the periods, or make the larger change in this one.

florelis (approved):

PR #5280, opened 2025-03-07
Dscv3 command base

JohnMcPMS:
## Change
Adds infrastructure for writing DSC v3 resource commands.
Adds a new hidden command `dscv3` to put the resources under.
Adds a file content resource (`DscTestFileResource`) that is only available when tests hooks are enabled.

The resource infrastructure generates the resource manifest based on the defined functions and modifiers given to the base via `winget dscv3 <resource> --manifest`.  It directs the `DscFunctions` it is given to the appropriate virtual method (`--get` maps to `ResourceFunctionGet`).

Resources can define properties using the helper macro `WINGET_DSC_DEFINE_COMPOSABLE_PROPERTY` and create an object to represent their properties with `DscComposableObject`.  This makes extracting and writing properties to JSON easy, as well as generating a schema.

## Validation
Updated E2E tests to use this test resource rather than the built-in registry resource from DSC v3.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5280)


yao-msft (approved):

mdanish-kh:
Following up from https://github.com/microsoft/winget-cli/pull/5252#issuecomment-2695202822

We now have the stable release of DSCv3 merged into winget-pkgs as [`Microsoft.DSC`](https://github.com/microsoft/winget-pkgs/tree/master/manifests/m/Microsoft/DSC/3.0.0). Should we update the azure-pipelines file to use this instead of the preview? That preview release doesn't work well with the CLI & breaks if invoked via symlink (https://github.com/PowerShell/DSC/issues/618). That _may_ get removed from winget-pkgs at some point because of other challenges with separate release variants for portable without channels support in the CLI, but we wouldn't want to do it if it's referenced here. The only concern (if it's a concern) being that the MSIX variant isn't available in the stable release assets

https://github.com/microsoft/winget-cli/blob/a3cea271bb84cf2a3358d97242effc348cce18a5/azure-pipelines.yml#L381-L384

JohnMcPMS:
I will make a separate change for that, once we get the MSIX in the release artifacts:
https://github.com/PowerShell/DSC/issues/697

denelon:
@SteveL-MSFT are you planning on adding the MSIX version of DSC to the release artifacts?

src/AppInstallerCLICore/Public/ConfigurationSetProcessorFactoryRemoting.h
```diff
@@ -35,6 +35,9 @@ namespace AppInstaller::CLI::ConfigurationRemoting
         // The path to the dsc.exe executable, as discovered.
         // Read only.
         FoundDscExecutablePath,
+        // Whether to request detailed traces from the processor.
+        // Read / Write
+        DiagnosticTraceLevel,
```
yao-msft, line 40: nit: the name "*TraceLevel" makes me think it's taking a trace level setting like warning, verbose while after reading the code, it's more like "DiagnosticTraceEnabled"

yao-msft (approved):

src/AppInstallerCLICore/Public/ConfigurationSetProcessorFactoryRemoting.h
JohnMcPMS: I will do that first thing in my next change so I don't need to wait for another build cycle.

PR #5302, opened 2025-03-17
DSC v3 adapter support

JohnMcPMS:
## Change
If we don't find a resource in DSC v3 at the top level, look in all of the adapters for it.  `dsc.exe` does this automatically when using the resource, so we only need to do it when finding the resource details.

Also improves the diagnostics handling and fixes an issue with the type of `implementedAs`.

## Validation
Adds an E2E test that uses the test v2 resource through the v3 processor.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5302)


src/AppInstallerCLIE2ETests/TestData/Configuration/Configure_TestRepo_DSCv3.yml
```diff
@@ -0,0 +1,10 @@
+$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
```
Gijsreyn, line 1: With the DSC v3.0 GA, shouldn't this schema be changed to the bundled one and aka.ms?

yao-msft (approved):
:shipit:

PR #5304, opened 2025-03-18
Fix DynamicProcessor crash when working on individual units without a set

yao-msft:
Added unit test. Manually validated as well.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5304)


JohnMcPMS (dismissed):

JohnMcPMS:
There could still be attempts to use `m_configurationSet` when `CreateSetProcessorForIntegrityLevel` wants to create a `Security::IntegrityLevel::High` processor.  It calls `Serialize*` and both of those functions use it.

I'm not sure what you can really do with that case though as the security design doesn't really allow for arbitrary unit processing in High IL.  So it should probably just fail (not AV of course).

JohnMcPMS (approved):

PR #5359, opened 2025-04-07
Fixes for pipeline stability

JohnMcPMS:
## Issue
Due to an issue with dsc.exe creating child processes (which after much investigation appears to be a race that only occurs on older OS versions), we get arbitrary failures.

We also occasionally get spurious failures of configuration tests.

## Change
Move to `windows-2025` to avoid the process creation failures from dsc.exe.
Move to Windows 11 SDK 26100 as 22000 is not installed by default on `windows-2025`.
Flush output at the end of a configuration unit info dump.
Add missing interface to manifest (@yao-msft was right).
Use `pwsh` task at least once to ensure that it is present.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5359)

### DIFF
--- a/.vsconfig
+++ b/.vsconfig
@@ -50,7 +50,7 @@
     "Microsoft.VisualStudio.Component.VC.ASAN",
     "Microsoft.VisualStudio.Component.Vcpkg",
     "Microsoft.VisualStudio.Component.Windows10SDK.19041",
-    "Microsoft.VisualStudio.Component.Windows11SDK.22000",
+    "Microsoft.VisualStudio.Component.Windows11SDK.26100",
     "Microsoft.VisualStudio.Workload.NativeDesktop",
     "Microsoft.Component.NetFX.Native",
     "Microsoft.VisualStudio.ComponentGroup.UWP.NetCoreAndStandard",
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,7 +15,7 @@ pr:
     - schemas/JSON/manifests/*
 
 pool:
-  vmImage: 'windows-latest'
+  vmImage: 'windows-2025'
 
 variables:
   solution: 'src\AppInstallerCLI.sln'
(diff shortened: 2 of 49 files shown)


florelis (approved):

.vsconfig
```diff
@@ -50,7 +50,7 @@
     "Microsoft.VisualStudio.Component.VC.ASAN",
     "Microsoft.VisualStudio.Component.Vcpkg",
     "Microsoft.VisualStudio.Component.Windows10SDK.19041",
```
florelis, line 52: Do we need to keep 19041?

azure-pipelines.yml
```diff
@@ -186,7 +186,7 @@ jobs:
           Microsoft.Management.Deployment\Microsoft.Management.Deployment.winmd
           WindowsPackageManager\WindowsPackageManager.dll
           UndockedRegFreeWinRT\winrtact.dll
-      TargetFolder: $(buildOutDirAnyCpu)\PowerShell\Microsoft.WinGet.Client\net8.0-windows10.0.22000.0\SharedDependencies\$(BuildPlatform)
+      TargetFolder: $(buildOutDirAnyCpu)\PowerShell\Microsoft.WinGet.Client\net8.0-windows10.0.26100.0\SharedDependencies\$(BuildPlatform)
```
florelis, line 189: This is going to be fun to update in the internal pipeline...

florelis (approved):

azure-pipelines.yml
JohnMcPMS: <fingers in ears> la la la la, I can't hear you

.vsconfig
JohnMcPMS: Probably not, I don't see anything using it in this solution.  I can remove it in another PR to prevent the need to rebuild this one.

PR #5389, opened 2025-04-16
Better DefaultInstallVersion

yao-msft:
DefaultInstallVersion being null causes larger than expected issues. The change proposes to return latest applicable version if exists, to minimize duplicate installs/ upgrade failures. If no latest applicable version, return latest available version if exists.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5389)

### DIFF
--- a/src/Microsoft.Management.Deployment/CatalogPackage.cpp
+++ b/src/Microsoft.Management.Deployment/CatalogPackage.cpp
@@ -67,9 +67,9 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         return m_availableVersions.GetView();
     }
 
-    void CatalogPackage::InitializeDefaultInstallVersion()
+    void CatalogPackage::InitializeLatestApplicableVersion()
     {
-        std::call_once(m_defaultInstallVersionOnceFlag,
+        std::call_once(m_latestApplicableVersionOnceFlag,
             [&]()
             {
                 using namespace AppInstaller::Pinning;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Management.Deployment/CatalogPackage.cpp
```diff
@@ -126,15 +126,27 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
             });
     }
 
     Microsoft::Management::Deployment::PackageVersionInfo CatalogPackage::DefaultInstallVersion()
     {
-        InitializeDefaultInstallVersion();
-        return m_defaultInstallVersion;
+        InitializeLatestApplicableVersion();
+
+        if (m_latestApplicableVersion)
```
yao-msft, line 138: If people absolutely do not want applicability check for DefaultInstallVersion, I can remove this if statement. But I feel having applicability check will lower chances of upgrade failures.

src/Microsoft.Management.Deployment/CatalogPackage.cpp
```diff
@@ -126,15 +126,27 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        {
+            return m_latestApplicableVersion;
+        }
+        else if (AvailableVersions().Size() > 0)
+        {
+            return GetPackageVersionInfo(AvailableVersions().GetAt(0));
```
JohnMcPMS, line 144: I would rather that this decision be made in the `call_once`, but I also refactored this entire function elsewhere in my upcoming PR.  So I will move that when I handle the merge conflict that this creates...

JohnMcPMS (approved):

PR #5395, opened 2025-04-17
Dscv3 package resource

JohnMcPMS:
## Change
Add a new DSC v3 resource: `package`.  This largely follows the design of the v2 resource, but uses the v3 paradigm for existence.

Other minor changes:
- Adds a `StdErrLogger` and attaches it during resource execution.  This currently writes all logs at `Error` or higher to `std::cerr`, allowing `dsc.exe` to receive (and pass on) failure information.
- Uses the type of the standard input handle to determine if it should be read from (will not read JSON input from the interactive console, only piped input).
- Adds support for localized schema generation for resources.
- Moves `ManifestComparator` to common for re-use.

## Validation
Added E2E tests to invoke the package resource in a variety of situations.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5395)


src/AppInstallerCLICore/Commands/DscPackageResource.cpp
```diff
@@ -0,0 +1,565 @@
...
+                            AICLI_LOG(CLI, Info, << "Reinstalling package to change scope");
+                            data.Reinstall();
+                        }
+                        if (!data.TestLatest())
+                        {
+                            // Install will swap to update flow
+                            AICLI_LOG(CLI, Info, << "Installing package to update to latest");
+                            data.Install();
+                        }
+                        else // (!data.TestVersion())
```
yao-msft, line 453: nit: cleanup?

yao-msft (approved):

src/AppInstallerCLICore/Commands/DscPackageResource.cpp
```diff
@@ -0,0 +1,565 @@
...
+                output.Source(source.Details.Name);
+
+                if (!package.VersionAndChannel.GetVersion().IsEmpty())
+                {
+                    output.Version(package.VersionAndChannel.GetVersion().ToString());
+                }
+
+                if (package.Scope != Manifest::ScopeEnum::Unknown)
+                {
+                    output.Scope(ConvertScope(Manifest::ScopeToString(package.Scope), true));
```
yao-msft, line 553: If we export Scope, during apply, will it be enforced during installer selection and may cause failures? i.e. we have many installers that do not specify scope.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3247,4 +3247,43 @@ Please specify one of them using the --source option to proceed.</value>
...
+    <value>The install mode to use if needed.</value>
+  </data>
+  <data name="DscResourcePropertyDescriptionPackageScope" xml:space="preserve">
+    <value>The target scope of the package.</value>
+  </data>
+  <data name="DscResourcePropertyDescriptionPackageAcceptAgreements" xml:space="preserve">
+    <value>Indicates whether to accept agreements for package installs.</value>
+  </data>
+  <data name="CorrelationArgumentDescription" xml:space="preserve">
+    <value>An unused argument for logging</value>
```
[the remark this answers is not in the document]
JohnMcPMS: I did consider making it test specific, but it seems reasonable to allow it to be passed in by anyone so that log files can be tagged.

src/AppInstallerCLICore/Commands/DscPackageResource.cpp
```diff
@@ -0,0 +1,565 @@
...
+                return std::nullopt;
+            }
+
+            std::string lowerValue = Utility::ToLower(value.value());
+
+            if (lowerValue == "equals")
+            {
+                return MatchType::Exact;
+            }
+            else if (lowerValue == "equals""case""insensitive")
```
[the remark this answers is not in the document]
JohnMcPMS: Yes, the spell checker is ok with this and C rules means that this is the same as `"equalscaseinsensitive"`.

src/AppInstallerCLICore/Commands/DscPackageResource.cpp
JohnMcPMS: Good point; I think we would need to support the "OrUnknown" variants for this to work as simply as it is now.  Alternately, we would need to check that the current scope is an available installer or track scope intent as we do for some other values.

denelon: Yeah, the other thing (for export) besides scope that could be a challenge is the install location.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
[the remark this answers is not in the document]
yao-msft: nit: maybe a description like "hidden argument" would be better than "unused"

yao-msft (approved):
:shipit:

PR #5420, opened 2025-05-01
Enable NuGet Central Package Management

florelis:
This pull request introduces the NuGet [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management) feature to streamline dependency management. Key benefits and changes include:

- **Centralized Version Management**: All NuGet package versions are now managed centrally in the `Directory.Packages.props` file. This ensures consistency across all projects in the repository.
  
- **Simplified Project Files**: The individual project files are simplified, as they no longer need to specify version numbers for NuGet packages. This change reduces duplication and the potential for version conflicts.
  
- **Ease of Updating Dependencies**: Updating a package version is now a matter of changing it in one place, making the process of keeping dependencies up-to-date more straightforward and less error-prone.

(Description copied from the automated PR to the internal repo :D )

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -1,0 +1,34 @@
+﻿<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageVersion Include="coverlet.collector" Version="3.1.2" />
+    <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.21.0" />
+    <PackageVersion Include="Microsoft.AspNet.WebApi.Client" Version="5.2.9" />
+    <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
+    <PackageVersion Include="Microsoft.Msix.Utils" Version="2.1.1" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+    <PackageVersion Include="Microsoft.PowerShell.SDK" Version="7.4.6" />
+    <PackageVersion Include="Microsoft.Win32.Registry" Version="5.0.0" />
+    <PackageVersion Include="Microsoft.Windows.Compatibility" Version="6.0.9" />
+    <PackageVersion Include="Microsoft.Windows.CsWinRT" Version="2.2.0" />
+    <PackageVersion Include="Microsoft.Windows.SDK.Contracts" Version="10.0.26100.1742" />
+    <PackageVersion Include="Moq" Version="4.18.2" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageVersion Include="nunit" Version="3.12.0" />
+    <PackageVersion Include="NUnit3TestAdapter" Version="3.15.1" />
+    <PackageVersion Include="Octokit" Version="4.0.3" />
+    <PackageVersion Include="PowerShellStandard.Library" Version="5.1.1" />
+    <PackageVersion Include="Semver" Version="2.3.0" />
+    <PackageVersion Include="StyleCop.Analyzers" Version="1.1.118" />
+    <PackageVersion Include="System.Data.SqlClient" Version="4.8.6" />
+    <PackageVersion Include="System.Net.Http" Version="4.3.4" />
+    <PackageVersion Include="System.Private.Uri" Version="4.3.2" />
+    <PackageVersion Include="System.Security.Principal.Windows" Version="5.0.0" />
+    <PackageVersion Include="System.Text.RegularExpressions" Version="4.3.1" />
+    <PackageVersion Include="xunit" Version="2.4.2" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.5" />
+    <PackageVersion Include="YamlDotNet" Version="16.3.0" />
+  </ItemGroup>
+</Project>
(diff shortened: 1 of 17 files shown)


src/Directory.Packages.props
```diff
@@ -0,0 +1,34 @@
+﻿<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
```
JohnMcPMS, line 3: This props file is just automagically included by some built-in targets?

src/Directory.Packages.props
```diff
@@ -0,0 +1,34 @@
...
+    <PackageVersion Include="coverlet.collector" Version="3.1.2" />
+    <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.21.0" />
+    <PackageVersion Include="Microsoft.AspNet.WebApi.Client" Version="5.2.9" />
+    <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
+    <PackageVersion Include="Microsoft.Msix.Utils" Version="2.1.1" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+    <PackageVersion Include="Microsoft.PowerShell.SDK" Version="7.4.6" />
+    <PackageVersion Include="Microsoft.Win32.Registry" Version="5.0.0" />
+    <PackageVersion Include="Microsoft.Windows.Compatibility" Version="6.0.9" />
+    <PackageVersion Include="Microsoft.Windows.CsWinRT" Version="2.2.0" />
```
JohnMcPMS, line 15: This cannot just be moved forward without changes that are not centrally managed in this change.  Fortunately, I think that with the embedded setup we shouldn't have to worry about any other compat problems (hopefully).

In the projects that do use CsWinRT, they currently have a property:
```
<WindowsSdkPackageVersion>10.0.26100.53</WindowsSdkPackageVersion>
```
That is dependent on the SDK that we use (10.0.26100) and the CsWinRT version (the 53 part, which is for 2.1.6).

Maybe, like the comment near those says, we can actually remove this.  I don't remember the exact details of why it broke without this, but I don't think it would build.  But if we don't remove it, it should probably be put into this props file so that it can also be updated singularly.  And of course either the 53 [needs to be updated](https://github.com/microsoft/CsWinRT/releases) or the 2.2.0 needs to be reverted to 2.1.6.

src/Microsoft.Management.Configuration.Processor/Microsoft.Management.Configuration.Processor.csproj
```diff
@@ -47,15 +47,15 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.1.6" />
-    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.4.6" GeneratePathProperty="true">
+    <PackageReference Include="Microsoft.Windows.CsWinRT" />
+    <PackageReference Include="Microsoft.PowerShell.SDK">
```
JohnMcPMS, line 51: Does central management just always generate the path property?

src/PowerShell/Microsoft.WinGet.Client.Engine/Microsoft.WinGet.Client.Engine.csproj
```diff
@@ -43,19 +43,19 @@
...
     </PackageReference>
-    <PackageReference Include="System.Security.Principal.Windows" Version="5.0.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" Condition="'$(TargetFramework)' == '$(DesktopFramework)'" />
-    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.1.6" Condition="'$(TargetFramework)' == '$(CoreFramework)'" />
-    <PackageReference Include="Microsoft.Windows.SDK.Contracts" Version="10.0.26100.1742" PrivateAssets="all" Condition="'$(TargetFramework)' == '$(DesktopFramework)'" />
-    <PackageReference Include="System.Net.Http" Version="4.3.4" Condition="'$(TargetFramework)' == '$(DesktopFramework)'" />
+    <PackageReference Include="System.Security.Principal.Windows" />
+    <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="Microsoft.CSharp" />
```
JohnMcPMS, line 55: We can't just drop these conditions; they are relevant to the .Net Framework vs .Net Core builds.

src/Directory.Packages.props
florelis: Correct

src/Directory.Packages.props
```diff
@@ -0,0 +1,34 @@
...
+    <PackageVersion Include="Octokit" Version="4.0.3" />
+    <PackageVersion Include="PowerShellStandard.Library" Version="5.1.1" />
+    <PackageVersion Include="Semver" Version="2.3.0" />
+    <PackageVersion Include="StyleCop.Analyzers" Version="1.1.118" />
+    <PackageVersion Include="System.Data.SqlClient" Version="4.8.6" />
+    <PackageVersion Include="System.Net.Http" Version="4.3.4" />
+    <PackageVersion Include="System.Private.Uri" Version="4.3.2" />
+    <PackageVersion Include="System.Security.Principal.Windows" Version="5.0.0" />
+    <PackageVersion Include="System.Text.RegularExpressions" Version="4.3.1" />
+    <PackageVersion Include="xunit" Version="2.4.2" />
```
florelis, line 30: I just realized that we use both nunit and xunit in different projects, which seems a bit weird

src/Directory.Packages.props
florelis: Reverted to 2.1.6. I had it as 2.2.0 because I just copied it from the other automated PR in the internal repo, and that had 2.2.0 due to some internal projects

JohnMcPMS (approved):

PR #5433, opened 2025-05-05
Place DSCv3 resource manifests in the package

JohnMcPMS:
## Change
Adds a method to write all manifests to a directory with one command.  Uses that to write all DSCv3 resource manifests and add them to the package files.

Also changes the "module" name for the dev package to add ".Dev" and updated export to use the shared definition of the module name.  This involved a minor refactor to remove static globals and unnecessary string constructions.

Finally, moves to using the DSC v3 `UserSettingsFile` resource rather than the v2 version.

## Validation
The package builds successfully as desired locally.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5433)


src/AppInstallerCLICore/Commands/DscCommandBase.cpp
```diff
@@ -224,14 +222,23 @@ namespace AppInstaller::CLI
 
     WINGET_DSC_FUNCTION_FOREACH(WINGET_DSC_FUNCTION_METHOD);
 
-    void DscCommandBase::ResourceFunctionManifest(Execution::Context& context) const
+    std::string_view DscCommandBase::ModuleName()
+    {
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        return "Microsoft.WinGet.Dev"sv;
```
yao-msft, line 228: I think this needs to be added to exclusion list in export --all too.

yao-msft (dismissed):

src/AppInstallerCLICore/Commands/DscCommand.cpp
```diff
@@ -12,8 +12,22 @@
 
 namespace AppInstaller::CLI
 {
+    namespace
+    {
+        Argument GetOutputFileArgument()
+        {
+            return { Execution::Args::Type::OutputFile, Resource::String::OutputFileArgumentDescription, ArgumentType::Standard };
```
yao-msft, line 19: nit: Do we consider differentiating output file vs output directory?

JohnMcPMS: Changed the description, left the arg type the same.

yao-msft (dismissed):

yao-msft (approved):

PR #5443, opened 2025-05-08
Attempt to use DSCv3 resources even if they were not found

JohnMcPMS:
## Change
Even if not found, attempt to use DSCv3 resources.  This is largely to avoid PowerShell/DSC#786, which only seems to affect dsc.exe's ability to run processes when listing adapted resources.

Also applies the `DiagnosticTraceEnabled` state to the elevated processor.  This is set when the config logging area is enabled and set to verbose.

## Validation
When running manually, the logs indicate that the adapted `list` invocations are failing, but the attempt to use the resource still succeeds.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5443)

### DIFF
--- a/src/AppInstallerCLICore/ConfigurationDynamicRuntimeFactory.cpp
+++ b/src/AppInstallerCLICore/ConfigurationDynamicRuntimeFactory.cpp
@@ -436,12 +436,8 @@ namespace AppInstaller::CLI::ConfigurationRemoting
                             }
                         });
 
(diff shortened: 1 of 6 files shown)


yao-msft (approved):
Should we set that Microsoft.Windows.Settings elevation to true since this change will mitigate the issue?

AmelBawa-msft (approved):

PR #4583, opened 2024-06-26
Component Governance fix: Update libyaml

yao-msft:
Component governance fix: update libyaml

Since a new release has not been created since after 0.2.5, pull to latest master. I reviewed the commits and they look safe.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/4583)

PR #5455, opened 2025-05-12
Use more recent version for libyaml

florelis:
We have a Component Governance alert for libyaml. There is no release of libyaml with this issue fixed, so the guidance was to apply the patch manually, and that's what I did when moving to vcpkg, but that doesn't play nicely with CG because it can't determine that the patch to fix the vulnerability was applied.

Instead of manually patching, this PR uses a more recent commit of libyaml (without an official release), which already has the changes we want. It uses the same commit from the last time we did a subtree update in #4583


src/VcpkgPortOverlay/libyaml/portfile.cmake
```diff
@@ -5,13 +5,12 @@ endif()
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO yaml/libyaml
-    REF 2c891fc7a770e8ba2fec34fc6b545c672beb37e6 # 0.2.5
-    SHA512 7cdde7b48c937777b851747f7e0b9a74cb7da30173e09305dad931ef83c3fcee3e125e721166690fe6a0987ba897564500530e5518e4b66b1c9b1db8900bf320
+    REF 840b65c40675e2d06bf40405ad3f12dec7f35923 # 0.2.5
```
JohnMcPMS, line 8:
```suggestion
    REF 840b65c40675e2d06bf40405ad3f12dec7f35923 # Unreleased
```

JohnMcPMS (commented):
Did using an unrelease version break the validation?

PR #5461, opened 2025-05-17
Fix debug build with latest VS version

yao-msft:
The u8"" string literal causes a new compiler warning 5321 with latest update, disable it in the project level. 

Also updated fuzzing path, again.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5461)


src/AppInstallerCLI.sln
```diff
@@ -606,23 +606,23 @@ Global
 		{846FB88B-BF1B-4F33-9883-E589CEC99739}.TestRelease|ARM64.ActiveCfg = Release|Any CPU
 		{846FB88B-BF1B-4F33-9883-E589CEC99739}.TestRelease|x64.ActiveCfg = Release|Any CPU
 		{846FB88B-BF1B-4F33-9883-E589CEC99739}.TestRelease|x86.ActiveCfg = Release|Any CPU
-		{68808357-902B-406C-8C19-E8E26A69DE8A}.Debug|ARM64.ActiveCfg = Debug|ARM64
+		{68808357-902B-406C-8C19-E8E26A69DE8A}.Debug|ARM64.ActiveCfg = Debug|x64
```
JohnMcPMS, line 609:
```suggestion
		{68808357-902B-406C-8C19-E8E26A69DE8A}.Debug|ARM64.ActiveCfg = Debug|ARM64
```
Shouldn't all of these ARM64 configs point to ARM64?

florelis:
What is warning 5321? I can't find any docs online mentioning it

src/AppInstallerCLI.sln
yao-msft: The tests project only has x86 and x64 configured, it's not built in arm64. This is why the configuration manager is complaining it can't find debug|arm64 configuration.

yao-msft:
> What is warning 5321? I can't find any docs online mentioning it

I searched online and cannot find it too when I encountered it. The closest I found is https://github.com/microsoft/STL/wiki/Changelog/c5aa92bec22b0d09c41282a79437855e9b269ebf

Which basically says: for cross platform reasons, don't use u8"\xffff" like literals since it's not supported in other platforms. But since our project is Windows only, I think we can safely ignore it.

JohnMcPMS (approved):

PR #5475, opened 2025-05-22
Fix catalog connection failure due to AppCapability::CreateWithProcessIdForUser throwing E_INVALIDARG when called with a null user on Windows 10 v1903

### LINKED ISSUE #5378: Connecting to a catalog reference returns `E_INVALIDARG` on Windows 10 1903 due to `AppCapability::CreateWithProcessIdForUser` not accepting a null user when called from an unpackaged app

https://github.com/microsoft/winget-cli/blob/1b74fd4d7ded2874254e1e8da3a3605ef20b3530/src/Microsoft.Management.Deployment/Helpers.cpp#L81C72-L81C134

`EnsureProcessHasCapability` calls `AppCapability::CreateWithProcessIdForUser` and passes `nullptr` as the user (which is not something the documentation for that method allows, but it wouldn't be the first underdocumented method...). When that call happens in a program without package identity on Windows 10 1903, this fails with `E_INVALIDARG`, which makes `PackageCatalogReference::Connect` fail.

### Steps to reproduce

1. Get a `PackageCatalogReference` object.
2. Call `Connect` on Windows 10 1903.
3. The extended error code will be `E_INVALIDARG`.

[Reproducer](https://github.com/user-attachments/files/19703725/WinGetBug-Capability.zip)

### Expected behavior

Connecting to a package catalog succeeds.

### Actual behavior

Connecting to a package catalog fails with `E_INVALIDARG`.

```
onecore\base\devices\cam\winrt\lib\appcapabilityaccessfactory.cpp(32)\CapabilityAccessManagerClient.dll!6F6ADE82: (caller: 7027946E) ReturnHr(1) tid(181c) 80070057 The parameter is incorrect.
WindowsPackageManager.dll!70279491: LogHr(1) tid(181c) 80070057 The parameter is incorrect.
(b6c.181c): Windows Runtime Originate Error - code 40080201 (first chance)
EXCEPTION:  WinRTOriginateError
hr = 0x80070057
Message:  The parameter is incorrect.
```

### Environment

```shell
Microsoft.WindowsPackageManager.InProcCom.1.10.340 from NuGet
```

Fulgen301:
`AppCapability::CreateWithProcessIdForUser` does not accept a null user on Windows 10 v1903. Treat this case as if the API didn't exist.

Fixes #5378.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5475)

### DIFF
--- a/src/Microsoft.Management.Deployment/Helpers.cpp
+++ b/src/Microsoft.Management.Deployment/Helpers.cpp
@@ -92,7 +92,7 @@ namespace winrt::Microsoft::Management::Deployment::implementation
             {
                 status = capability.CheckAccess();
 
-                return (status == winrt::Windows::Security::Authorization::AppCapabilityAccess::AppCapabilityAccessStatus::Allowed);
+                return ((status == winrt::Windows::Security::Authorization::AppCapabilityAccess::AppCapabilityAccessStatus::Allowed) ? S_OK : E_ACCESSDENIED);
             }
         }
 
--- a/src/Microsoft.Management.Deployment/Helpers.cpp
+++ b/src/Microsoft.Management.Deployment/Helpers.cpp
@@ -69,12 +69,6 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         }
     }
 
-    bool HasPackageId()
-    {
-        std::uint32_t bufferLength{};
-        return GetCurrentPackageId(&bufferLength, nullptr) != APPMODEL_ERROR_NO_PACKAGE;
-    }
-
     HRESULT EnsureProcessHasCapability(Capability requiredCapability, DWORD callerProcessId)
     {
         bool allowed = false;
--- a/src/Microsoft.Management.Deployment/Helpers.cpp
+++ b/src/Microsoft.Management.Deployment/Helpers.cpp
@@ -69,6 +69,12 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         }
     }
 
+    bool HasPackageId()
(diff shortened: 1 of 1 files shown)


JohnMcPMS (commented):
I spent quite a while trying to find a way to get a non-null `User` to pass, but never found one that works for us.  I did find the change that made null work, but it ultimately ends up ignoring the value, not using it as a signal to retrieve some default.

So while it irks me, I will not make the perfect stand in the way of the good.

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -78,24 +84,34 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            catch (const winrt::hresult_invalid_argument&)
             {
-                allowed = AppInstaller::Security::IsCurrentIntegrityLevelAtLeast(requiredIntegrityLevel);
             }
+
+            if (capability)
+            {
+                status = capability.CheckAccess();
+
+                return (status == winrt::Windows::Security::Authorization::AppCapabilityAccess::AppCapabilityAccessStatus::Allowed);
```
yao-msft, line 101: The method expects an hresult return type

Fulgen301: Ugh, good catch, fixed.

yao-msft (approved):

PR #5490, opened 2025-05-29
Fix: Prevent reinstallation of already installed versions during upgrade

Kenxpx:
This PR fixes an issue where `winget upgrade` would show packages for upgrade even when the exact version was already installed.

### Problem
If multiple versions of a package are installed (e.g., via different channels), `winget upgrade` doesn't correctly match all installed versions. It may show an upgrade even if a matching version is already installed.

### Fix
This fix iterates over all installed versions and uses the first one that matches the available version and channel. If a match is found, it skips the reinstallation.

### Impact
Prevents redundant upgrade suggestions and improves reliability of `winget upgrade`.

Fixes #2960

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5490)

### DIFF
--- a/src/AppInstallerRepositoryCore/PackageInstalledStatus.cpp
+++ b/src/AppInstallerRepositoryCore/PackageInstalledStatus.cpp
@@ -17,7 +17,6 @@ namespace AppInstaller::Repository
             HRESULT installLocationStatus = WINGET_INSTALLED_STATUS_INSTALL_LOCATION_NOT_APPLICABLE;
             if (!installedLocation.empty())
             {
-                // Use the none throw version, if the directory cannot be reached, it's treated as not found and later file checks are not performed.
                 std::error_code error;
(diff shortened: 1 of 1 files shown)


Kenxpx:

yao-msft (changes requested):
I doubt this change will affect winget upgrade behavior. The code being touched was only used for COM api reporting of a specific installed package version, it was never used in any winget upgrade flow.

Btw, the pr needs some cleanup as it's removing a bunch of comments not related to the change itself.

And if we are fixing incorrect behavior, please add an e2e test (install 2 different versions, and call winget upgrade to verify the correct behavior).

nit: Also the issue linked in pr description points to a winget pre-release, did you type the wrong issue number?

PR #5511, opened 2025-06-09
Refresh PATH in configuration remoting server

yao-msft:
Added a hidden window to listen for UserPreferenceChanged event (WM_SETTINGCHANGE) and update the process's PATH if needed.

Manually validated by running a configuration file with multiple resources, change the env var during the run, and verified in logs file that new path got included in runs after the change.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5511)

### DIFF
--- a/src/Microsoft.Management.Configuration.Processor/Public/PathEnvironmentVariableHandler.cs
+++ b/src/Microsoft.Management.Configuration.Processor/Public/PathEnvironmentVariableHandler.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Management.Configuration.Processor.Helpers
         /// </summary>
         public static void UpdatePath()
         {
-            List<string> paths = new List<string>(Environment.GetEnvironmentVariable(PathEnvironmentVariable)?.Split(';') ?? Array.Empty<string>());
+            HashSet<string> paths = new HashSet<string>(Environment.GetEnvironmentVariable(PathEnvironmentVariable)?.Split(';') ?? Array.Empty<string>());
             var originalPathsSize = paths.Count;
 
             AddPathsIfNotExist(paths, Environment.GetEnvironmentVariable(PathEnvironmentVariable, EnvironmentVariableTarget.Machine)?.Split(';'));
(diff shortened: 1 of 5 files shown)


src/Microsoft.Management.Configuration.Processor/Public/PathEnvironmentVariableHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+            get { return EnvironmentVariableLock; }
+        }
+
+        /// <summary>
+        /// Updates the process's PATH environment variable if new paths added.
+        /// Only adds new paths since we add to PATH in other code which may not be in the registry.
+        /// </summary>
+        public static void UpdatePath()
+        {
+            List<string> paths = new List<string>(Environment.GetEnvironmentVariable(PathEnvironmentVariable)?.Split(';') ?? Array.Empty<string>());
```
JohnMcPMS: Should this be a `HashSet<string>` to make the looped `Contains` calls in `AddPathsIfNotExist` more efficient?

src/Microsoft.Management.Configuration.Processor/Public/PathEnvironmentVariableHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+        private static void AddPathsIfNotExist(List<string> currentPaths, string[]? paths)
+#pragma warning restore SA1011 // Closing square brackets should be spaced correctly
+        {
+            if (paths is not null)
+            {
+                foreach (var path in paths)
+                {
+                    if (!currentPaths.Contains(path))
+                    {
+                        currentPaths.Add(path);
```
JohnMcPMS, line 63: This always adds to the end, making it the last location to look.  That may cause issues in some corner case; probably ok for now but maybe add a comment.

The "proper" thing to do would probably be to calculate the full new list of paths (what one would expect to get from a new process launch) and use a line merge algorithm with a strategy that puts the ephemeral entries before the new permanent ones.  But that is a fair bit more complicated and could wait until later.

JohnMcPMS (commented):
Please confirm that the server process exits cleanly when the client completes.

yao-msft:
> Please confirm that the server process exits cleanly when the client completes.

I can confirm ConfigurationRemotingServer.exe exits right after configuration completes.

JohnMcPMS (approved):

PR #5519, opened 2025-06-10
Add text file with the release notes

florelis:
This adds a text file with the notes we add to each release here in GitHub.
I intend to update the publishing script to include these notes automatically so I don't have to manually copy them each time.

These notes are not intended to be a full changelog of all versions, but rather just about the changes in the current version. (Although maybe we should have a changelog for all major versions at least.)

It would be nice if we manage to keep the notes up to date every time we make a change, but if we forget we can still edit the release after the fact.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5519)

### DIFF
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -1,7 +1,7 @@
 ## New in v1.11
 * Dropped support for running on 32-bit ARM
 * Support for [Microsoft Desired State Configuration (DSC) v3](https://learn.microsoft.com/powershell/dsc/overview?view=dsc-3.0).
-* Support for exporting the configuration of the current device. This includes Windows Settings, packages from configured WinGet sources, and  package settings from DSC v3 enabled packages.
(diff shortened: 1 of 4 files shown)


yao-msft (approved):

doc/ReleaseNotes.md
```diff
@@ -0,0 +1,20 @@
+## New in v1.11
+* Dropped support for running on 32-bit ARM
+* Support for [Microsoft Desired State Configuration (DSC) v3](https://learn.microsoft.com/powershell/dsc/overview?view=dsc-3.0).
+* Support for exporting the configuration of the current device. This includes Windows Settings, packages from configured WinGet sources, and  package settings from DSC v3 enabled packages.
```
yao-msft:
```suggestion
* Support for exporting the configuration of the current device. This includes Windows Settings, packages from configured WinGet sources, and package settings from DSC v3 enabled packages.
```
nit nit

JohnMcPMS (approved):

PR #5521, opened 2025-06-12
Ensure required modules for configuration export and import

yao-msft:
- Added Microsoft.PowerShell as dependent unit of other Predefined modules.
- Use RunCommandOnSet as awork around for ensuring v2 modules before v3 resource is available
- Also added always apply logic for certain units i.e. Microsoft.DSC.Transitional/RunCommandOnSet
- Added and updated tests

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
@@ -53,7 +53,6 @@ namespace AppInstaller::CLI::Workflow
         constexpr std::wstring_view s_UnitType_WinGetSource_DSCv3 = WINGET_DSCV3_MODULE_NAME_WIDE L"/Source";
         constexpr std::wstring_view s_UnitType_WinGetUserSettingsFile_DSCv3 = WINGET_DSCV3_MODULE_NAME_WIDE L"/UserSettingsFile";
         constexpr std::wstring_view s_UnitType_WinGetAdminSettings_DSCv3 = WINGET_DSCV3_MODULE_NAME_WIDE L"/AdminSettings";
-        constexpr std::wstring_view s_UnitType_PowerShellModuleGet = L"PowerShellGet/PSModule";
 
         constexpr std::wstring_view s_Module_WinGetClient = L"Microsoft.WinGet.DSC";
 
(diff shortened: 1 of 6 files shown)


src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1246,17 +1243,58 @@ namespace AppInstaller::CLI::Workflow
...
 
-        ConfigurationUnit CreatePowerShellModuleGetUnit(const std::wstring& moduleName)
+        ConfigurationUnit CreatePowerShellPackageUnit()
         {
-            ConfigurationUnit unit = CreateConfigurationUnitFromUnitType(s_UnitType_PowerShellModuleGet, Utility::ConvertToUTF8(moduleName));
+            ConfigurationUnit unit = CreateConfigurationUnitFromUnitType(s_UnitType_WinGetPackage_DSCv3, "Microsoft.PowerShell");
 
             ValueSet settings;
-            settings.Insert(s_Setting_PowerShellGet_ModuleName, PropertyValue::CreateString(moduleName));
+            settings.Insert(s_Setting_WinGetPackage_Id, PropertyValue::CreateString(L"Microsoft.PowerShell"));
```
JohnMcPMS: Constants?

Also, we should consider extracting this from / filtering out of the package results.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1536,15 +1574,27 @@ namespace AppInstaller::CLI::Workflow
         {
             ConfigurationContext& configContext = context.Get<Data::ConfigurationContext>();
 
+            // PowerShell package needs to be present for certain predefined modules to work.
+            ConfigurationUnit powerShellPackageUnit = CreatePowerShellPackageUnit();
+            configContext.Set().Units().Append(powerShellPackageUnit);
+
+            // Apply the unit to make sure it's on the system.
+            context.Reporter.Info() << Resource::String::ConfigurationExportInstallRequiredModule(Utility::LocIndView{ "Microsoft PowerShell Package" }) << std::endl;
+            auto applyPowerShellResult = ApplyUnit(context, powerShellPackageUnit);
```
JohnMcPMS, line 1586: We always apply it without testing?

yao-msft: For the wingetpackage dsc resource, it tests it (in the resource implementation) before applying. So I'm just applying it to save 2 calls.

PR #5525, opened 2025-06-12
Support for DSC minimum version

JohnMcPMS:
## Change
Adds a state machine for finding a version of DSC that meets our minimum requirements.  It first attempts to use the stable release, falling back to the preview version if it is not new enough.  This allows us to have functional, if less than ideal, user experience while we wait for DSC to ship a GA version.

Both callers (winget and PS module) are updated to handle the state machine transition mechanics, which means they need to install stable/preview versions when requested.

Also fixes some issues with the PS module install flow.

## Validation
Manually confirmed that being up to date for both packages works, as well as having neither present.

### DIFF
--- a/src/AppInstallerCLICore/Public/ConfigurationSetProcessorFactoryRemoting.h
+++ b/src/AppInstallerCLICore/Public/ConfigurationSetProcessorFactoryRemoting.h
@@ -40,6 +40,9 @@ namespace AppInstaller::CLI::ConfigurationRemoting
         // Whether to request detailed traces from the processor.
         // Read / Write
         DiagnosticTraceEnabled,
+        // The path to the dsc.exe executable, as discovered.
+        // Read only.
+        FindDscStateMachine,
     };
 
     // Gets the string for a property name.
--- a/src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
+++ b/src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
@@ -6,6 +6,14 @@
 
 namespace Microsoft.WinGet.Configuration.Engine.Commands
 {
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Management.Automation;
+    using System.Management.Automation.Runspaces;
+    using System.Text;
+    using System.Threading.Tasks;
     using Microsoft.Management.Configuration;
     using Microsoft.Management.Configuration.Processor;
     using Microsoft.Management.Configuration.Processor.PowerShell.Extensions;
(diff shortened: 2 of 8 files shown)


src/AppInstallerCLICore/Public/ConfigurationSetProcessorFactoryRemoting.h
```diff
@@ -40,6 +40,9 @@ namespace AppInstaller::CLI::ConfigurationRemoting
         // Whether to request detailed traces from the processor.
         // Read / Write
         DiagnosticTraceEnabled,
+        // The path to the dsc.exe executable, as discovered.
```
yao-msft: nit: copy paste comment

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
```diff
@@ -408,18 +407,42 @@ private async Task<IConfigurationSetProcessorFactory> CreateDSCv3ProcessorFactor
...
+                    }
+                    else if (nextTransition == "NotFound")
+                    {
+                        this.Write(StreamType.Warning, Resources.ConfigurationInstallDscPackageFailed);
+                        throw new FileNotFoundException(Resources.DscExeNotFound, "dsc.exe");
+                    }
+                    else
+                    {
+                        this.Write(StreamType.Warning, $"Unrecognized value from FindDscStateMachine: {nextTransition ?? "<null>"}");
+                        throw new FileNotFoundException(Resources.DscExeNotFound, "dsc.exe");
```
yao-msft: Should we throw some other exception to show it's internal error?

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
```diff
@@ -434,18 +457,19 @@ private async Task InstallDSCv3Package(OpenConfigurationParameters openParams)
...
                 }}
                 ");
 
             await installDSCv3.InvokeAsync();
 
             if (installDSCv3.HadErrors)
             {
-                this.Write(StreamType.Error, Resources.ConfigurationInstallDscPackageFailed);
+                this.Write(StreamType.Verbose, installDSCv3.GetErrorMessage() ?? "<Unknown error>");
+                this.Write(StreamType.Warning, Resources.ConfigurationInstallDscPackageFailed);
```
yao-msft, line 472: Is there a reason we change to warning here?

yao-msft (dismissed):

src/PowerShell/Microsoft.WinGet.Configuration.Engine/Commands/ConfigurationCommand.cs
JohnMcPMS: `Error` must be an `ErrorRecord` object, not a string.  I can look at creating one, but there isn't an easy "from a string" method to do so.

JohnMcPMS: Also, I think that writing to `Error` and throwing an exception are effectively the same thing except for execution flow.

PR #5526, opened 2025-06-13
Fix AV in portable flow

JohnMcPMS:
## Change
Only use the installed version's scope when we have one to use.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5526)


JohnMcPMS:
I looked at *why* we were getting into this state, but didn't find anything obvious.  Probably need a live repro to determine the root cause.

yao-msft (approved):

yao-msft (approved):

PR #5566, opened 2025-06-30
Font Install, Uninstall, additional Font List

dkbennett:
Implements a good chunk of #166  Key features and changes:

All of the below are experimental behind the fonts experimental feature. Font install is implemented as a winget-common installer that models most of the font install behavior of the operating system in its key components. It deviates from a normal font install in that the install locations and registry key paths are a little different in order to map fonts to an identifiable version-specific deterministic location. This is so we may more easily identify font packages and remove them and avoid font collisions, torn state, or removing fonts from other packages.

**Font List update**
- Added new font list "--files" option that lists fonts by installed files and their known validity state
   - Fonts are listed by file where they are located and have an assigned identifier for later use in removal.
   - The Font Status reflects whether the registry matches an actual font file on disk (fonts without a file are labeled "Corrupt").
   - The intent here is to add filters later to get a better idea of what fonts are on the system, where they came from, and which might need repair or removal. At present my system had over 300 font files installed so the list is quite large.

**Font Install**
- Add font support for install. This is the base WinGet install command, not the 'font' sub-command.
- Requires font experimental feature to be enabled.
- Added the core Font evaluation code to AppInstallerCommonCore so it is not strictly bound to the CLI.
- The font install scheme puts WinGet installed font packages in sub folders and subkeys where they can exist side-by-side with other font packages without worry of file collision and easier high-confidence removal/upgrade and rollback.
- Registers the font with the current session and notifies other apps of the new font availability.
- Configures the font to be loaded on subsequent sessions.
- Install works for machine and user fonts.

**Font Uninstall*
- Added font support for uninstall. This is the base WinGet uninstall command, not the 'font' sub-command.
- Requires font Experimental feature to be enabled.
- Uninstall works for WinGet-installed fonts only at this time.

** Font List*
- Added font package detection to the PredefinedInstalledSource.
- This allows font packages to be detected for uninstall.

**Workflows**
- Support for Font Install and Uninstall

**Tested**
- Added many unit tests which exercise the the Font install, uninstall, information gathering, listing, for machine and user scopes. All of the new core code added is exercised in unit tests.
- All scopes for install using unit tests and local web server.
- Install and uninstall with manifest an local server. 
- Install and uninstall and list with package from a private package source.
- Verified ACLs on the folders match expected permissions for visibility.
- Verified per-user and per-machine fonts are installed correctly and show up in Font Settings immediately.

**NOT part of this PR and is likely a future PR**
- Font upgrade (a new version will be installed side-by-side)
- Uninstall of non-packaged fonts. This is in plan, just not in the scope of this PR.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5566)


src/AppInstallerCLICore/Argument.cpp
```diff
@@ -470,6 +472,8 @@ namespace AppInstaller::CLI
             return Argument{ type, Resource::String::NoProxyArgumentDescription, ArgumentType::Flag, TogglePolicy::Policy::ProxyCommandLineOptions, BoolAdminSetting::ProxyCommandLineOptions };
         case Args::Type::Family:
             return Argument{ type, Resource::String::FontFamilyNameArgumentDescription, ArgumentType::Positional, false };
+        case Args::Type::Files:
+            return Argument{ type, Resource::String::FontFamilyNameArgumentDescription, ArgumentType::Flag, false };
```
mdanish-kh:
```suggestion
            return Argument{ type, Resource::String::FontFilesArgumentDescription, ArgumentType::Flag, false };
```

doc/ReleaseNotes.md
```diff
@@ -15,4 +15,8 @@ The following snippet enables experimental support for fonts via `winget setting
   }
 }
 
+Experimental initial Font Install and Uninstall via manifest for user and machine scopes has been added.
+
+The font 'list' command has been updated with a new '--files' feature for an alternate view of the installed fonts.
+
```
mdanish-kh: Should be written outside the ``` JSON block

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -222,6 +222,13 @@ namespace AppInstaller
             WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_SFSCLIENT_PACKAGE_NOT_SUPPORTED, "The Microsoft Store package does not support download."),
             WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_LICENSING_API_FAILED_FORBIDDEN, "Failed to retrieve Microsoft Store package license. The Microsoft Entra Id account does not have the required privilege."),
             WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_INSTALLER_ZERO_BYTE_FILE, "Downloaded zero byte installer; ensure that your network connection is working properly."),
+            WINGET_HRESULT_INFO(APPINSTALLER_CLI_ERROR_FONT_INSTALL_FAILED, "Failed to install font package."),
```
mdanish-kh: We should also update [returnCodes.md](https://github.com/microsoft/winget-cli/blob/master/doc/windows/package-manager/winget/returnCodes.md) with the new error codes

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3370,4 +3370,78 @@ Please specify one of them using the --source option to proceed.</value>
...
+    <value>Font uninstall failed. The font may not be in a good state. Try uninstalling after a restart.</value>
+  </data>
+  <data name="APPINSTALLER_CLI_ERROR_FONT_VALIDATION_FAILED" xml:space="preserve">
+    <value>"Font validation failed."</value>
+  </data>
+  <data name="APPINSTALLER_CLI_ERROR_FONT_ROLLBACK_FAILED" xml:space="preserve">
+    <value>Font rollback failed. The font may not be in a good state. Try uninstalling after a restart.</value>
+  </data>
+  <data name="FontFilesArgumentDescription" xml:space="preserve">
+    <value>Show fonts by installed font file.</value>
```
mdanish-kh:
```suggestion
    <value>Show fonts by installed font file</value>
```

nit: for style consistency since rest of the argument descriptions don't end with a period. I know there's this PR #5281 that should put this argument (whether to have a period or no) to rest _once it gets merged_, but right now we should be consistent?

doc/ReleaseNotes.md
mdanish-kh: Also, I suppose bumping the schema version to 1.12.0 with support for `InstallerType: font` is something that should also be called out in the release notes?

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -43,17 +49,13 @@ namespace AppInstaller::CLI
...
         return {
             Argument::ForType(Args::Type::Family),
-            Argument::ForType(Args::Type::Moniker),
-            Argument::ForType(Args::Type::Source),
-            Argument::ForType(Args::Type::Tag),
-            Argument::ForType(Args::Type::Exact),
-            Argument::ForType(Args::Type::AuthenticationMode),
-            Argument::ForType(Args::Type::AuthenticationAccount),
-            Argument::ForType(Args::Type::AcceptSourceAgreements),
+            Argument::ForType(Args::Type::Files),
```
Trenly: I don't think we should be removing the common arguments
* What is preventing a font from having a moniker?
* Why shouldn't I be able to add a private font source and specify that specific source?
* Why would exact matching be prevented?
* Why wouldn't EntraID Sources be supported?
* Why force interactive acceptance of source agreements if there are any?

See also https://github.com/microsoft/winget-cli/pull/4842/files#r1783193297

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -83,4 +85,113 @@ namespace AppInstaller::CLI
...
+
+
+    // FontInstallCommand
+
+    std::vector<Argument> FontInstallCommand::GetArguments() const
+    {
+        return {
+            Argument::ForType(Args::Type::Manifest),
+            Argument{ Args::Type::InstallScope, Resource::String::InstallScopeDescription, ArgumentType::Standard, Argument::Visibility::Help },
+            Argument::ForType(Args::Type::Force),
```
Trenly: See above comment about common arguments

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -121,4 +187,153 @@ namespace AppInstaller::CLI::Workflow
...
+        if (context.Args.Contains(Execution::Args::Type::InstallScope))
+        {
+            scope = Manifest::ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope));
+        }
+
+        // Scope may still be unknown, which is a failure for an install operation.
+        if (scope == Manifest::ScopeEnum::Unknown)
+        {
+            // We will default to User scope.
+            scope = Manifest::ScopeEnum::User;
```
Trenly: Is there a reason to not default it to user when instantiating the variable and instead handle the defaulting here?

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -83,4 +85,113 @@ namespace AppInstaller::CLI
...
+    {
+        return s_FontCommand_HelpLink;
+    }
+
+    void FontInstallCommand::ValidateArgumentsInternal(Execution::Args& execArgs) const
+    {
+        Argument::ValidateCommonArguments(execArgs);
+    }
+
+    void FontInstallCommand::ExecuteInternal(Execution::Context& context) const
```
Trenly: Is the MultiQuery flow that packages use also supported for fonts? If not, should it be?

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -83,4 +85,113 @@ namespace AppInstaller::CLI
...
+    {
+        if (context.Args.Contains(Execution::Args::Type::Manifest))
```
Trenly: Manifest is required for fonts??
If future work is planned for querying from a source, the command should show an appropriate error message that `--manifest` must be used and the exit code for not implemented should be thrown

src/AppInstallerCLICore/Workflows/InstallFlow.h
```diff
@@ -49,6 +49,12 @@ namespace AppInstaller::CLI::Workflow
         // Inputs: Installer, InstallerPath, Manifest
         // Outputs: None
         void ArchiveInstall(Execution::Context& context);
+
+        // Runs the flow for intsalling a font package.
```
Trenly:
```suggestion
        // Runs the flow for installing a font package.
```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3370,4 +3370,78 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="ResetAllAdminSettingsSucceeded" xml:space="preserve">
     <value>All admin settings reset.</value>
   </data>
+  <data name="FontInstallCommandShortDescription" xml:space="preserve">
+    <value>Install a font package</value>
+  </data>
+  <data name="FontInstallCommandLongDescription" xml:space="preserve">
+    <value>Installs the selected font package, either found by searching a configured source or directly from a manifest.</value>
+  </data>
+  <data name="APPINSTALLER_CLI_ERROR_FONT_INSTALL_FAILED" xml:space="preserve">
```
Trenly, line 3393: This isn't the same message that you have in `winget-cli/src/AppInstallerSharedLib/Errors.cpp`. Is that intentional?

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+    }
+
+    std::filesystem::path GetFontFileInstallPath(const FontContext& context)
+    {
+        auto path = std::filesystem::path();
+
+        switch (context.Scope)
+        {
+        case Manifest::ScopeEnum::Machine:
+            path /= Runtime::GetPathTo(Runtime::PathName::FontsMachineInstallLocation);
```
Trenly: These paths should be added to `winget --info`

 https://github.com/microsoft/winget-cli/blob/26e5eca45967b4949e96883aa9f5f953ceb84518/src/AppInstallerCLICore/Commands/RootCommand.cpp#L147-L161

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+                else
+                {
+                    fileInfo.PackageIdentifier = s_UnknownInstallerPrefixUser.data() + filePath.filename().wstring();
+                }
+            }
+
+            break;
+        }
+
+        // For font file paths in the registry, we may have a relative filename (becuase the path is assumed).
```
Trenly:
```suggestion
        // For font file paths in the registry, we may have a relative filename (because the path is assumed).
```

src/AppInstallerCLICore/Commands/FontCommand.cpp
dkbennett: I added arguments that have implementation. Other arguments will be added as the implementation supports more functionality. This is very much the first cut and more will be added in future PRs.

Trenly: Makes sense, thanks!

src/AppInstallerCLICore/Workflows/FontFlow.cpp
dkbennett: I prefer defaulting to unknown when it is in fact unknown and letting the interface decide what the behavior should be when it is unknown. Defaulting to a value could very easily lead to an error where it wasn't being set appropriately and we wouldn't know. Because I set it to unknown I very quickly found that scope may be undefined, which now makes the behavior explicit for caller to decide.

src/AppInstallerCLICore/Commands/FontCommand.cpp
dkbennett: It isn't at present; it should be in the future. I expect a lot of updates to the command flow for install and uninstall in the near future once the font installer code is solid.

src/AppInstallerCLICore/Commands/FontCommand.cpp
dkbennett: Yes, this is not the end implementation of fonts; it is the beginning. We plan on fully supporting most of what other packages support.

src/AppInstallerCLICore/Workflows/InstallFlow.h
dkbennett: I'm going to do a spelling pass update shortly, including the updating the spellcheck list where appropriate.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
dkbennett: No not intentional, I updated one and forgot to update the other. I'll do a pass on all the strings and make sure they're consistent.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: The font installer is not like most things winget does, it is an actual installer, not a host for an installer. I'm wary of implementation details being put into something like that that would invite people mucking with it and corrupting the installs. I'm not familiar with the purpose of key directories, but I will look into it and add if appropriate.

src/AppInstallerCLICore/Commands/FontCommand.cpp
dkbennett: Note that these are arguments for the List command, not the install command. I removed them because they weren't being used at all. This is an experimental feature and nothing stops us from adding them back when they are functional. To that end I will also remove a few unused ones in Uninstall.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: I'm going to leave these paths out of winget info because we are installing things to normal system locations, not something winget manages. They will be functional without winget present at all, similar to if another installer did the installation. They are tagged with winget-identifiable names, but that is for identification and layout considerations, not something anyone should ever need to look at.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
dkbennett: Fixed in latest update.

src/AppInstallerCLICore/Commands/FontCommand.cpp
dkbennett: Removed unnecessary uninstall arguments so now all font commands reflect arguments that are supported.

src/AppInstallerSharedLib/Errors.cpp
dkbennett: Updated

doc/ReleaseNotes.md
dkbennett: Fixed

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -83,4 +85,110 @@ namespace AppInstaller::CLI
...
+
+    Resource::LocString FontInstallCommand::LongDescription() const
+    {
+        return { Resource::String::FontInstallCommandLongDescription };
+    }
+
+    void FontInstallCommand::Complete(Execution::Context& context, Args::Type valueType) const
+    {
+        UNREFERENCED_PARAMETER(valueType);
+        context.Reporter.Error() << Resource::String::PendingWorkError << std::endl;
```
JohnMcPMS: This is probably more harmful than helpful.  The base implementation does a decent job of handling common argument types and falls back to the list of files in the working directory.  If you want a TODO sign, call the base and leave a comment.

The same applies to all commands in this file, new and old.

src/AppInstallerCLICore/Commands/FontCommand.cpp
JohnMcPMS: My expectation is that many of the `font` commands should be able to (eventually) be implemented as:

1. Set up the source state in the context
2. Jump over to the existing non-font workflow

Some like `show` and `list` probably need more special handling due to the nature of fonts.

I do agree that the command should report an error if the manifest is not supplied, but that can be accomplished by making it a required argument for now.

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -83,4 +85,110 @@ namespace AppInstaller::CLI
...
+    }
+
+    void FontUninstallCommand::ExecuteInternal(Execution::Context& context) const
+    {
+        context.SetFlags(Execution::ContextFlag::TreatSourceFailuresAsWarning);
+
+        context <<
+            Workflow::ReportExecutionStage(ExecutionStage::Discovery) <<
+            Workflow::OpenSource() <<
+            Workflow::OpenCompositeSource(Workflow::DetermineInstalledSource(context));
```
JohnMcPMS: I don't think that this is needed currently; it doesn't appear to be adding anything at least.

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -83,4 +85,110 @@ namespace AppInstaller::CLI
...
+                Workflow::InstallSinglePackage;
+        }
+    }
+
+    // Font Uninstall Command
+
+    std::vector<Argument> FontUninstallCommand::GetArguments() const
+    {
+        return {
+            Argument::ForType(Args::Type::Manifest),
```
JohnMcPMS: Make required for now.

src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
```diff
@@ -78,22 +78,28 @@ namespace AppInstaller::CLI::Workflow
...
 
-            // Assuming that we find a safe stem value in the URI, use it.
-            // This should be extremely common, but just in case fall back to the older name style.
-            if (filename.has_stem() && ((filename.wstring().size() + installerExtension.size()) < MAX_PATH))
-            {
-                filename = filename.stem();
-            }
-            else
+            // Installers that support multiple extensions will use the Url.
+            if (!DoesInstallerTypeSupportMultipleFileExtensions(installer->BaseInstallerType))
```
JohnMcPMS: You probably need to change this so that `GetInstallerFileExtension` returns the extension from the URL if needed so that the filename length check still happens.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -467,6 +467,8 @@ namespace Microsoft.Management.Deployment
         MSStore,
         /// Portable type.
         Portable,
+        /// Font type.
```
JohnMcPMS: This needs to be protected by a contract version block, and the version likely needs to be incremented (check what it is in the 1.11 branch to be sure).

src/ManifestSchema/ManifestSchema.h
```diff
@@ -62,3 +62,9 @@
 #define IDX_MANIFEST_SCHEMA_V1_10_INSTALLER           244
 #define IDX_MANIFEST_SCHEMA_V1_10_DEFAULTLOCALE       245
 #define IDX_MANIFEST_SCHEMA_V1_10_LOCALE              246
+
+#define IDX_MANIFEST_SCHEMA_V1_12_SINGLETON           247
+#define IDX_MANIFEST_SCHEMA_V1_12_VERSION             248
+#define IDX_MANIFEST_SCHEMA_V1_12_INSTALLER           249
+#define IDX_MANIFEST_SCHEMA_V1_12_DEFAULTLOCALE       250
+#define IDX_MANIFEST_SCHEMA_V1_12_LOCALE              251
```
JohnMcPMS, line 70: Hopefully we are not limited to one byte for these ids...

src/WindowsPackageManager/WindowsPackageManager.vcxproj
```diff
@@ -257,7 +257,7 @@
       <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">Source.def</ModuleDefinitionFile>
       <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">Source.def</ModuleDefinitionFile>
       <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">Source.def</ModuleDefinitionFile>
-      <AdditionalDependencies Condition="'$(Configuration)'=='Debug'">wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;winhttp.lib;onecoreuap.lib;msi.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies Condition="'$(Configuration)'=='Debug'">wininet.lib;shell32.lib;winsqlite3.lib;shlwapi.lib;icuuc.lib;icuin.lib;urlmon.lib;Advapi32.lib;winhttp.lib;onecoreuap.lib;msi.lib;gdi32.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
JohnMcPMS, line 255: Should probably be added to the delayload list.

src/AppInstallerCLIE2ETests/Test.runsettings
```diff
@@ -30,13 +30,14 @@
     <Parameter name="AICLIPackagePath" value="AppInstallerCLIPackage.appxbundle" />
     <Parameter name="LooseFileRegistration" value="false" />
     <Parameter name="InvokeCommandInDesktopPackage" value="false" />
-    <Parameter name="StaticFileRootPath" value="\TestLocalIndex" />
+    <Parameter name="StaticFileRootPath" value="D:\dev\TestLocalIndex" />
```
JohnMcPMS: This path seems specific to your machine.

src/AppInstallerCommonCore/Public/winget/Fonts.h
```diff
@@ -4,9 +4,38 @@
...
 {
+    enum class InstallerSource
+    {
+        // Installer source is not guaranteed to be absolutely correct, as it is just
+        // filesystem and registry. We make assumptions based on the structure of how
+        // the font is installed to make a best-effort determination.
+        Unknown,    // Can be any external font installer
+        UWP,        // Font installed by a Universal Windows Platform app
+        WinGet,     // Font installed by WinGet
+        Any,        // Can be any installer.
```
JohnMcPMS: How are `Unknown` and `Any` different? [comment made before reading Fonts.cpp]

src/AppInstallerCommonCore/Public/winget/Fonts.h
```diff
@@ -20,20 +49,95 @@ namespace AppInstaller::Fonts
...
+
+    struct FontOperationResult
+    {
+        winrt::hresult HResult = winrt::hresult(S_OK);
+        FontResult Result();
+
+        // TODO: Add optional rollback context to unwind installs if one operation succeeded but a subsequent one failed.
+    };
+
+    std::wstring GetFontRegistryPath(const FontContext& context);
```
JohnMcPMS: Why not a member of `FontContext`?  The same will apply to any of the functions here that take in one of the types you define above.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
 
             return filePaths;
         }
+
+        void RemoveAllFontResources(std::filesystem::path filePath)
+        {
+            // The recommended uninstall method of a font is to call RemoveFontResource until it fails,
```
JohnMcPMS, line 188: My reading of the documentation is that you would only do this when the font is under your control.  It also suggests calling `AddFontResource` the same number of times after replacing the file.  This makes me suspicious of the usage this way.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+        }
+        else
+        {
+            return FontResult::Success;
+        }
+    }
+
+    void FontContext::AddPackageFile(const std::filesystem::path& filePath)
+    {
+        if (!PackageFiles.has_value())
```
JohnMcPMS: Is there value in differentiating between not having a vector and an empty one?

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+    {
+        auto path = std::filesystem::path();
+
+        switch (context.Scope)
+        {
+        case Manifest::ScopeEnum::Machine:
+            path /= Runtime::GetPathTo(Runtime::PathName::FontsMachineInstallLocation);
+            break;
+        case Manifest::ScopeEnum::User:
+            path /= Runtime::GetPathTo(Runtime::PathName::FontsUserInstallLocation);
```
JohnMcPMS: Use `GetRootFontPath` rather than having another bit of code with the same switch.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+            break;
+        case InstallerSource::WinGet:
+            // WinGet path adds the WinGet prefix and the package name
+            if (context.PackageIdentifier.has_value() && !context.PackageIdentifier.value().empty())
+            {
+                path << s_RegistrySeparator << s_FontsWinGetPrefix << s_RegistrySeparator << context.PackageIdentifier.value();
+            }
+            else
+            {
+                throw std::runtime_error("WinGet Font paths require a package name.");
```
JohnMcPMS: If you use std exceptions, you will have to put a catch-wall at the front-end so that they can be presented to the user properly (localized messages).  Prefer `THROW_HR` with our errors.  When the front-end catches those it will output the localized version of the error message automatically.

The exception (PUN :100:) to that rule would be if these are all logic errors that you only expect to occur due to bugs, not based on user or data malformities.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+        {
+            auto fontCatalog = FontCatalog();
+            fileInfo.WinGetSupported = fontCatalog.IsFontFileSupported(fileInfo.FilePath, fileInfo.FileType);
+
+            if (fileInfo.WinGetSupported && fileInfo.Title.empty())
+            {
+                try
+                {
+                    // The GetFontTitle call can fail with E_INVALIDARG if the font file is located in
+                    // a subdirectory of the system Fonts folder. The system does not seem to expect
```
JohnMcPMS: This suggests that we should use more verbose file names than subdirectories.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+        }
+
+        return fileInfo;
+    }
+
+    FontValidationResult ValidateFontPackage(FontContext& context)
+    {
+        auto result = FontValidationResult();
+
+        AICLI_LOG(Core, Info, << L"Validating font package: " << context.PackageIdentifier.value_or(L"Unknown").c_str());
```
JohnMcPMS:
```suggestion
        AICLI_LOG(Core, Info, << L"Validating font package: " << context.PackageIdentifier.value_or(L"<Unknown>").c_str());
```
Putting `<>` catches the eye more to me, making the default value stand out more (and differentiating it from the potentially valid value of "Unknown").

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +67,647 @@ namespace AppInstaller::Fonts
...
+
+        auto fontFileInfos = std::vector<FontFileInfo>();
```
JohnMcPMS: Unused

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -172,4 +840,120 @@ namespace AppInstaller::Fonts
...
+    // This will not include fonts that are temporarily installed for the session.
+    std::vector<FontFileInfo> GetInstalledFontFiles()
+    {
+        auto fontFiles = std::vector<FontFileInfo>();
+
+        try
+        {
+            // Iterate through scopes for machine and user
+            std::vector<Manifest::ScopeEnum> scopes = { Manifest::ScopeEnum::Machine, Manifest::ScopeEnum::User };
+            for (const auto& scope : scopes)
```
JohnMcPMS:
```suggestion
            for (const auto& scope : { Manifest::ScopeEnum::Machine, Manifest::ScopeEnum::User })
```
Removes the unnecessary allocation of a vector.

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -121,4 +187,153 @@ namespace AppInstaller::CLI::Workflow
...
+        catch (...)
+        {
+            context.Add<Execution::Data::OperationReturnCode>(Workflow::HandleException(context, std::current_exception()));
+            context.Reporter.Warn() << Resource::String::FontInstallFailed << std::endl;
+
+            try
+            {
+                // The Font Install code handles rollback where appropriate. If we hit an
+                // unexpected exception, try to uninstall anyway as this is idempotent.
+                auto uninstallResult = Fonts::InstallFontPackage(fontContext);
```
JohnMcPMS:
```suggestion
                auto uninstallResult = Fonts::UninstallFontPackage(fontContext);
```
?

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -121,4 +187,153 @@ namespace AppInstaller::CLI::Workflow
...
+        }
+        catch (...)
+        {
+            context.Add<Execution::Data::OperationReturnCode>(Workflow::HandleException(context, std::current_exception()));
+            context.Reporter.Warn() << Resource::String::FontInstallFailed << std::endl;
+
+            try
+            {
+                // The Font Install code handles rollback where appropriate. If we hit an
+                // unexpected exception, try to uninstall anyway as this is idempotent.
```
JohnMcPMS: If this was a "force" to repair, uninstall is not idempotent.

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -121,4 +187,153 @@ namespace AppInstaller::CLI::Workflow
...
+        fontContext.InstallerSource = InstallerSource::WinGet;
+        fontContext.Scope = scope;
+
+        // Fonts are single instanced by an identifier which is:
+        //   PackageId + '_' + PackageVersion
+        // The PackageIdentifier is used to uniquely identify this package.
+        auto manifest = context.Get<Execution::Data::Manifest>();
+        const auto& packageId = ConvertToUTF16(manifest.Id);
+        const auto& packageVersion = ConvertToUTF16(manifest.Version);
+        fontContext.PackageIdentifier = packageId + L"_" + packageVersion;
```
JohnMcPMS: Shares a lot of code with install; could de-dupe.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: Yes, I need to change this to use a more verbose name because it simply does not work any other way with machine fonts, and the system does not recognize subfolders. This is a known issue I must fix and you are correct that this is what we should use for machine fonts.

src/AppInstallerCLICore/Commands/FontCommand.cpp
dkbennett: In the latest version of this PR font uninstall and install are now in the non-font workflow and do not require a manifest argument.

src/AppInstallerCLICore/Commands/FontCommand.cpp
```diff
@@ -3,8 +3,10 @@
 #include "pch.h"
 #include "FontCommand.h"
 #include "Workflows/CompletionFlow.h"
-#include "Workflows/WorkflowBase.h"
 #include "Workflows/FontFlow.h"
+#include "Workflows/InstallFlow.h"
```
JohnMcPMS: nit: Could remove unnecessary headers now.

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -138,7 +138,7 @@ namespace AppInstaller::CLI::Workflow
                 AICLI_TERMINATE_CONTEXT(APPINSTALLER_CLI_ERROR_INVALID_MANIFEST);
             }
 
-            if (installer.NestedInstallerType != InstallerTypeEnum::Portable && nestedInstallerFiles.size() != 1)
+            if (!DoesInstallerTypeSupportMultipleNestedInstallers(installer.NestedInstallerType) && (nestedInstallerFiles.size() != 1))
             {
                 AICLI_LOG(CLI, Error, << "Multiple nested installers specified for non-portable nested installerType");
```
JohnMcPMS: Log line still references portable.  Looks like resource string probably needs to be updated as well given its name.

src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
```diff
@@ -63,6 +63,18 @@ namespace AppInstaller::CLI::Workflow
...
+            case InstallerTypeEnum::Font:
+                {
+                    const auto& fileName = GetFileNameFromURI(installer->Url);
+                    if (fileName.has_extension())
+                    {
+                        return fileName.extension().c_str();
+                    }
+                    else
+                    {
+                        THROW_HR(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED));
```
JohnMcPMS, line 75: Is the file extension used by OS APIs to determine the encoding?  If not, could we just use `.font` or something if we don't have a better option?

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -121,4 +198,160 @@ namespace AppInstaller::CLI::Workflow
             OutputInstalledFontFamiliesTable(context, lines);
         }
     }
+
+    void FontInstallImpl(Execution::Context& context)
+    {
+        if (!Settings::ExperimentalFeature::IsEnabled(Settings::ExperimentalFeature::Feature::Font))
```
JohnMcPMS: There is an existing workflow task that handles this with a message and everything: `EnsureFeatureEnabled`.  Just put that into your task stream in InstallFlow (and anywhere else you need it).

src/AppInstallerCLICore/Workflows/FontFlow.cpp
```diff
@@ -121,4 +198,160 @@ namespace AppInstaller::CLI::Workflow
...
+            Fonts::FontContext fontContext;
+            fontContext.InstallerSource = InstallerSource::WinGet;
+            if (context.Args.Contains(Execution::Args::Type::Manifest))
+            {
+                const auto& manifest = context.Get<Execution::Data::Manifest>();
+                fontContext.PackageId = ConvertToUTF16(manifest.Id);
+                fontContext.PackageVersion = ConvertToUTF16(manifest.Version);
+                if (context.Args.Contains(Execution::Args::Type::InstallScope))
+                {
+                    fontContext.Scope = Manifest::ConvertToScopeEnum(context.Args.GetArg(Execution::Args::Type::InstallScope));
```
JohnMcPMS: Determine the scope once (here or above).

src/AppInstallerCommonCore/Fonts.cpp
JohnMcPMS: This continues to be suspicious; if anyone is using the font we might be ripping it out from under them.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -44,20 +185,647 @@ namespace AppInstaller::Fonts
...
+    FontFileInfo CreateFontFileInfo(const FontContext& context, const std::filesystem::path& filePath, const std::optional<std::wstring>& title)
+    {
+        auto fileInfo = FontFileInfo();
+        fileInfo.FilePath = filePath;
+        fileInfo.PackageIdentifier = context.PackageIdentifier;
+        fileInfo.PackageId = context.PackageId;
+        fileInfo.PackageVersion = context.PackageVersion;
+        fileInfo.InstallerSource = context.InstallerSource;
+        fileInfo.Scope = context.Scope;
+        if (title.has_value())
```
JohnMcPMS: This doesn't appear to need `optional`; an empty title string seems to be equivalent.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -85,6 +853,24 @@ namespace AppInstaller::Fonts
         return installedFontFamilies;
     }
 
+    bool FontCatalog::IsFontFileSupported(const std::filesystem::path& filePath, DWRITE_FONT_FILE_TYPE& fileType)
+    {
+        if (!std::filesystem::exists(filePath))
```
JohnMcPMS, line 828: `CreateFontFileInfo` checks if the file exists before calling this function, which checks if the file exists.  That outer check could be removed at the cost of creating a `FontCatalog` when it wasn't necessary.  But if that is expensive, then `CreateFontFileInfo` should be receiving one rather than creating one every time it is called.

src/AppInstallerCommonCore/Public/winget/Fonts.h
```diff
@@ -20,20 +47,103 @@ namespace AppInstaller::Fonts
...
+        Manifest::ScopeEnum Scope = Manifest::ScopeEnum::Unknown;
+        DWRITE_FONT_FILE_TYPE FileType = DWRITE_FONT_FILE_TYPE_UNKNOWN;
+        InstallerSource InstallerSource = InstallerSource::Unknown;
+        FontStatus Status = FontStatus::Unknown;
+        bool WinGetSupported = false;
+        bool IsFontFileInstalled = false;
+        bool IsFontFileRegistered = false;
+
+        // Registry Install path is where the installation with the system is located.
+        std::optional<std::wstring> RegistryInstallPath;
```
JohnMcPMS: Many instances of `optional` use that probably don't need it.  If the default value for a type is not a valid value semantically, then there is no effective difference between `std::nullopt` and `T{}` for the rest of the codebase.

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
```diff
@@ -263,6 +264,22 @@ namespace AppInstaller::Repository::Microsoft
                 PopulateIndexFromMSIX(index, Manifest::ScopeEnum::Machine);
             }
 
+            // Put Installed Fonts into the index.
```
JohnMcPMS, line 267: The installed items are cached and several signals are watched to know when to update that cache.  A registry watcher entry should be added for the relevant locations for fonts (if this remains the way).

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -172,4 +958,184 @@ namespace AppInstaller::Fonts
...
+    }
+
+    std::vector<FontPackageInfo> GetInstalledFontPackages(Manifest::ScopeEnum scope)
+    {
+        auto fontPackages = std::vector<FontPackageInfo>();
+
+        try
+        {
+            auto hive = scope == Manifest::ScopeEnum::Machine ? HKEY_LOCAL_MACHINE : HKEY_CURRENT_USER;
+            auto root = Registry::Key::OpenIfExists(hive, s_FontsPathSubkey.data(), 0UL, KEY_ALL_ACCESS);
```
JohnMcPMS:
```suggestion
            auto root = Registry::Key::OpenIfExists(hive, s_FontsPathSubkey.data(), 0UL, KEY_READ);
```
?

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -172,4 +958,184 @@ namespace AppInstaller::Fonts
...
+
+            for (const auto& subkey : root)
```
JohnMcPMS, line 1045: This seems like it does not work for machine scope, which doesn't have any subkeys.

src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
dkbennett: We have an explicit dwrite function to validate the font is a supported type, but I do not know if those APIs expect file extensions to be a certain way as part of their validation.  I have not come across any fonts that did not have one of the expected font filetype extensions. I expect this condition to be rare also, most font distributions are zip files with many fonts within them. Distributing a single font outside this method is unusual, and one without a filetype association would be quite exceptional.  I think this is best left as an error condition.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: This does work for machine scope, because it does have subkeys. We store the winget package information the same way in both machine and user scope. The difference is the actual font registration works both ways for User scope, but for machine scope a separate value is created for each file in the machine root, because the system only recognizes values in the machine root and ignores subkeys. This is how we keep package information separate from what the files on disk actually look like so we are keeping the font files human-readable as close to their original filename as we can while still avoiding collisions and functioning very similarly to user-installed fonts in every other way.

src/AppInstallerCLICore/Workflows/FontFlow.cpp
dkbennett: Fixed

src/AppInstallerCLICore/Workflows/FontFlow.cpp
dkbennett: Fixed and moved to using the workflow task.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: Fixed, no more redundant checking and setting.

src/AppInstallerCommonCore/Public/winget/Fonts.h
dkbennett: Removed almost all uses of optional, a few remain where I consider it meaningful of a difference between empty and null.

src/AppInstallerRepositoryCore/Microsoft/PredefinedInstalledSourceFactory.cpp
dkbennett: Added a Registry Watcher for the fonts, very similar to the ARPHelper example, and added a test for the helper which tests the watcher, also similar to ARPHelper.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: Fix this and in another place where only READ_KEY is necessary.

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
dkbennett: Fixed with new string, removed old string.

src/AppInstallerCLICore/Commands/FontCommand.cpp
dkbennett: Fixed.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -24,10 +24,10 @@ namespace AppInstaller::Fonts
         constexpr std::wstring_view s_FontsPathSubkey = L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Fonts";
         constexpr std::wstring_view s_FontsUserInstallFolder = L"Microsoft\\Windows\\Fonts";
         constexpr std::wstring_view s_TrueType = L" (TrueType)";
-        constexpr std::wstring_view s_UnknownInstallerPrefixUser = L"U\\F\\";
-        constexpr std::wstring_view s_UnknownInstallerPrefixMachine = L"M\\F\\";
+        constexpr std::wstring_view s_UnknownInstallerPrefixUser = L"FONT\\User\\";
+        constexpr std::wstring_view s_UnknownInstallerPrefixMachine = L"FONT\\Machine\\";
         constexpr std::wstring_view s_RegistrySeparator = L"\\";
-        constexpr std::wstring_view s_FontsWinGetPrefix = L"winget_v1";
+        constexpr std::wstring_view s_FontsWinGetPrefix = L"Microsoft.DesktopAppInstaller_8wekyb3d8bbwe";
```
JohnMcPMS, line 29: And this won't cause any issues if our package is removed or updated?

src/AppInstallerCommonCore/Fonts.cpp
JohnMcPMS: Bumping this again.

src/AppInstallerCommonCore/Fonts.cpp
JohnMcPMS: Bump.

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -478,43 +463,40 @@ namespace AppInstaller::Fonts
 
         auto result = FontValidationResult();
 
-        AICLI_LOG(Core, Info, << L"Validating font package: " << context.PackageIdentifier.value_or(L"<Unknown>").c_str());
+        AICLI_LOG(Core, Info, << L"Validating font package: " << context.PackageId.c_str() << L" " << context.PackageVersion.c_str());
```
JohnMcPMS: Logging takes UTF-8 strings, make the literals ASCII and convert the dynamic ones to UTF-8 or we will end up with garbage in the file for non-ASCII values.

This applies multiple other places in this file (and potentially others).

[I should make this not compile.]

src/AppInstallerCommonCore/Fonts.cpp
```diff
@@ -542,10 +524,10 @@ namespace AppInstaller::Fonts
             throw std::logic_error("Only WinGet format of font install is supported.");
         }
 
-        if (!context.PackageFiles.has_value() || context.PackageFiles.value().size() == 0)
+        if (context.PackageFiles.size() == 0)
         {
             result.HResult = winrt::hresult(APPINSTALLER_CLI_ERROR_FONT_FILE_NOT_FOUND);
-            AICLI_LOG(Core, Error, << L"Font package has no files: " << context.PackageId.value().c_str() << " - " << result.HResult);
+            AICLI_LOG(Core, Error, << L"Font package has no files: " << context.PackageId.c_str() << " - " << result.HResult);
```
JohnMcPMS: There is a helpful stream manipulator that sets things up to output an HRESULT in hex (`AppInstaller::Logging::SetHRFormat`).

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: This is the package family name so it is version agnostic. The reason I'm putting it here is because the system recognizes subfolders by package name or package family name, and this will be consistent with other apps that install fonts like Terminal.

src/AppInstallerRepositoryCore/Microsoft/FontHelper.cpp
```diff
@@ -0,0 +1,98 @@
...
+    void FontHelper::PopulateIndex(SQLiteIndex& index, Manifest::ScopeEnum scope) const
+    {
+        const auto& fontPackages = AppInstaller::Fonts::GetInstalledFontPackages(scope);
+        for (const auto& fontPackage : fontPackages)
+        {
+            try
+            {
+                Manifest::Manifest manifest;
+                manifest.DefaultLocalization.Add<Manifest::Localization::Tags>({ "Font" });
+                manifest.Id = ConvertToUTF8(fontPackage.PackageId);
```
JohnMcPMS: This should always match the identifier style of the other enumerators; something akin to the unknown font format you have which appears to be copied off of them (`FONT\<SCOPE>\etc.`).

JohnMcPMS: Then you make the correlation between what this generates and the tracking catalog function by putting the package identifier in the product code of both databases.

src/AppInstallerCommonCore/Fonts.cpp
JohnMcPMS: My question is "Given that the system uses PFN when installing an MSIX package with font(s), does it also just remove everything under a PFN when it uninstalls/updates an MSIX package?".  We don't want to store changes to the system requested by the user in a way that they are removed if our package is reset/removed/etc.

dkbennett: This is safe for two reasons. 1) we use a package family name while the app platform uses package full name, so they wont match. 2) Appinstaller doesn't use the manifest extensions that execute that code.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: Fixed all literals in the logging to be std string instead of wstring.

src/AppInstallerRepositoryCore/Microsoft/FontHelper.cpp
dkbennett: Updated the ID to be the FONT\Scope\Id as suggested, and moved that code into the core Font code so it is generated in one place and called the "PackageIdentifier" so PackageId can be specifically for the Id field of manifests to avoid confusion.

Added Correlation data as per recommendation similar to what Portable does.

Verified that this now has better matching for uninstall and list.

src/AppInstallerCommonCore/Fonts.cpp
dkbennett: Ty, updated to use this!

dkbennett:
Since i can't reply to the comment about AddFontResource or RemoveFontResource, I will update on that here.

I did some extensive testing and investigation of using and not using AddFontResource and RemoveFontResource. Using it was the guidance of font experts but the documentation on learn seems a little off. I verified using Notepad++ and Office Word that these two things need to be used as the code uses them or the Font Resources are not properly added or removed to the current session and the apps cannot detect the new font being available without it. They are used in this code as it was recommended that I use it and is consistent with internal font install behavior.

So these methods add and remove the font internally to the current active session and allows apps to use the fonts and detect them immediately. It is not tied to the current app; adding the font resource makes it available to all apps, and removing it removes it from all apps. I did not encounter issues of files-in-use when removing fonts the apps were actively using, this appears to be handled gracefully.

Moreover, removing a font while it is being used is handled well by the apps. Word was particularly graceful - adding a font was immediately available to the running app, you could use it, then when the font is removed via WinGet the document updated the text using that font to a similiar font. After re-adding the font via WinGet, Word updated the text back to the desired font.

src/AppInstallerCommonCore/Public/winget/Fonts.h
```diff
@@ -81,6 +81,7 @@ namespace AppInstaller::Fonts
         FontStatus Status = FontStatus::Unknown;
         std::wstring PackageId;
         std::wstring PackageVersion;
+        std::wstring PackageIdentifier;
```
JohnMcPMS, line 84: nit: Having both an `Id` and an `Identifier` is somewhat confusing.  Doesn't need to change for merge, but maybe think about renaming one of them in the future.

JohnMcPMS (approved):

src/AppInstallerCommonCore/Public/winget/Fonts.h
dkbennett: Next time I touch it I'll change it to "UniqueIdentifier" or something similar away from the Package similarity.

PR #5585, opened 2025-07-10
Handle Byte Order Mark during validation

### LINKED ISSUE #5502: YAML with UTF-8 BOM are not handled correctly

If you use YAMLs for the manifest and use UTF-8 with BOM, the first line is completely ignored. The first line usually holds the schema header. This might be a problem for a long time, but headers are checked now and if you use UTF-8 with BOM, the checks will fail.
Currently UTF-8 with BOM only works, if you place the schema header in the second line, leaving the first line blank. Or using UTF-8 without BOM.

### Steps to reproduce

Use a YAML with UTF-8 with BOM, place the schema-header in the first line and run the validate

### Expected behavior

The obe described usage should work

### Actual behavior

The first line is ignored

### Environment

```shell
Windows-Paket-Manager (Vorschau) v1.11.320-preview
Copyright (c) Microsoft Corporation. Alle Rechte vorbehalten.

Windows: Windows.Desktop v10.0.26100.4061
Systemarchitektur: X64
Paket: Microsoft.DesktopAppInstaller v1.26.320.0

The current azure tests will fail also!
```

Trenly:
  * Resolves #5502
-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5585)

### DIFF
--- a/src/AppInstallerSharedLib/YamlWrapper.cpp
+++ b/src/AppInstallerSharedLib/YamlWrapper.cpp
@@ -350,7 +350,7 @@ namespace AppInstaller::YAML::Wrapper
         // If input has a BOM, we want to remove it to prevent errors with checking for comments within the input document.
        
         // Check for UTF-16 BOMs
-        if (m_input.size() >= 2 && (m_input[0] == c_utf16LEBOM[0] && m_input[1] == c_utf16LEBOM[1]))
+        if (m_input.size() >= sizeof(c_utf16LEBOM) && std::memcmp(m_input.data(), c_utf16LEBOM, sizeof(c_utf16LEBOM)) == 0)
         {
             AICLI_LOG(YAML, Verbose, << "Found UTF-16 LE BOM");
 			yaml_parser_set_encoding(&m_parser, YAML_UTF16LE_ENCODING); // Without the BOM, the encoding must be explicitly set
@@ -358,7 +358,7 @@ namespace AppInstaller::YAML::Wrapper
             return;
         }
 
-        if (m_input.size() >= 2 && (m_input[0] == c_utf16BEBOM[0] && m_input[1] == c_utf16BEBOM[1]))
+        if (m_input.size() >= sizeof(c_utf16BEBOM) && std::memcmp(m_input.data(), c_utf16BEBOM, sizeof(c_utf16BEBOM)) == 0)
         {
             AICLI_LOG(YAML, Verbose, << "Found UTF-16 BE BOM");
             yaml_parser_set_encoding(&m_parser, YAML_UTF16BE_ENCODING); // Without the BOM, the encoding must be explicitly set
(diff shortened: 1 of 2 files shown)


src/AppInstallerSharedLib/Yaml.cpp
```diff
@@ -106,11 +106,19 @@ namespace AppInstaller::YAML
             std::string line;
             size_t currentLine = 1;
 
+            // convert BOMs to strings
```
yao-msft: BOM should only be the beginning of the document, so the correct way of doing it will be like how we handle it in PrepareInput(), compare the first 2 - 3 bytes and remove them if needed (a new yamlDocumentProcessed string without BOM).

By the way, not your fault, but if BOM is utf16, below code needs to use std::wstring and convert to utf8 string before returning. I remember somewhere we enforced utf-8 for winget-pkgs manifests? If that's the case, this might not be a problem and only utf8 BOM needs to be checked.

Trenly: Thanks for the advice! I'd originally not implemented it there due to the comment in `PrepareInput()` that the BOM was intentionally passed through. However, it seems this was only done so the parser could automatically determine the input encoding.

I've updated the PR to be entirely contained within `PrepareInput()`, removing the BOM if needed (as suggested) and directly setting the parser encoding based on the BOM.

src/AppInstallerSharedLib/YamlWrapper.cpp
```diff
@@ -343,23 +343,35 @@ namespace AppInstaller::YAML::Wrapper
...
+        // If input has a BOM, we want to remove it to prevent errors with checking for comments within the input document.
+       
         // Check for UTF-16 BOMs
-        if (m_input.size() >= 2 &&
-            ((m_input[0] == c_utf16BOM[0] && m_input[1] == c_utf16BOM[1]) || (m_input[0] == c_utf16BOM[1] && m_input[1] == c_utf16BOM[0])))
+        if (m_input.size() >= sizeof(c_utf16LEBOM) && std::memcmp(m_input.data(), c_utf16LEBOM, sizeof(c_utf16LEBOM)) == 0)
         {
-            AICLI_LOG(YAML, Verbose, << "Found UTF-16 BOM");
+            AICLI_LOG(YAML, Verbose, << "Found UTF-16 LE BOM");
+			yaml_parser_set_encoding(&m_parser, YAML_UTF16LE_ENCODING); // Without the BOM, the encoding must be explicitly set
```
yao-msft: nit: replace tabs with spaces. Several occurrences below too.

yao-msft (approved):
Do we enforce utf8 in our winget-pkgs manifests? If not, I feel the real fix need to include ExtractDocumentHeader side fixes. (I'm not asking for it now as it's a winget-pkgs only thing).

src/AppInstallerSharedLib/YamlWrapper.cpp
Trenly: I think this is an issue with how GitHub Online is showing it. I've explicitly used spaces (I've even copied and pasted the spaces from above into where GHO is showing tabs), but the result is always the same. I'll try again just to be sure though

Trenly:
> Do we enforce utf8 in our winget-pkgs manifests? If not, I feel the real fix need to include ExtractDocumentHeader side fixes. (I'm not asking for it now as it's a winget-pkgs only thing).

I believe we do enforce it at winget-pkgs, not certain about private repos

yao-msft (approved):

PR #5610, opened 2025-07-22
Initial MCP Server implementation

JohnMcPMS:
Discussion on further MCP work: https://github.com/microsoft/winget-cli/discussions/5609

## Change
This change adds an MCP server that provides two tools:

- Find packages
  - Searches for packages based on a single input value, returning information about each one found.
- Install package
  - Installs a package given a single input value, which is expected to be an identifier, name or moniker (same rules as `winget` CLI)

The package response looks like:
```JSON
{
  "identifier":"Microsoft.VisualStudioCode",
  "name":"Microsoft Visual Studio Code",
  "catalog":"winget",
  "isInstalled":true,
  "installedVersion":"1.102.1",
  "installedLocation":"%LOCALAPPDATA%\\Programs\\Microsoft VS Code\\",  <- Is actually real path, not tokenized
  "isUpdateAvailable":false
}
```

Also includes:
- A group policy that determines whether the MCP server is allowed to operate
- An `mcp` command that helps with manually configuring an MCP client
- A change to the default COM caller behavior that makes it only use the file name
  - Also changes to not use the package family name if it is the same as the current process (so we can better differentiate our own callers)
- A `Version` property on the `PackageManager` COM object that returns the version string as `1.2.3` or `1.2.3-preview`.
  - Start preferentially using this in the PowerShell module to get the version rather than running the CLI

## Validation
I have ensured that the various portions of this change function within the context of the dev package.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5610)


src/AppInstallerCLICore/Commands/ConfigureListCommand.cpp
```diff
@@ -3,6 +3,7 @@
 #include "pch.h"
 #include "ConfigureListCommand.h"
 #include "Workflows/ConfigurationFlow.h"
+#include "Workflows/MSStoreInstallerHandler.h"
```
florelis, line 6: This is silly but it bothers me that the stubbing stuff is in this header because it doesn't feel right to have this include...
(No need to change it though)

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3370,4 +3371,20 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="ResetAllAdminSettingsSucceeded" xml:space="preserve">
     <value>All admin settings reset.</value>
   </data>
+  <data name="McpCommandShortDescription" xml:space="preserve">
+    <value>MCP information</value>
+    <comment>MCP stands for Model Context Protocol and should probably remain as-is</comment>
```
florelis, line 3376: Anthropic's docs use MCP in other languages. For example, https://docs.anthropic.com/es/docs/mcp
So maybe this should be `{Locked}`

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1975,30 +1975,31 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="InstallFullPackageDescription" xml:space="preserve">
     <value>Install the full package for AppInstaller</value>
   </data>
-  <data name="ConfigurationEnableMessage" xml:space="preserve">
-    <value>Enable configuration components. Requires store access.</value>
+  <data name="ExtendedFeaturesEnableMessage" xml:space="preserve">
+    <value>Enable extended features. Requires store access.</value>
```
florelis, line 1979: I'm a bit concerned about there being confusion between "experimental" and "extended" features, but I can't think of a better naming.

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -50,6 +50,8 @@ namespace AppInstaller::Runtime
         constexpr std::string_view s_WindowsApps_Base = "Microsoft\\WindowsApps"sv;
         constexpr std::string_view s_WinGetDev_Exe = "wingetdev.exe";
         constexpr std::string_view s_WinGet_Exe = "winget.exe";
+        constexpr std::string_view s_WinGetMCPDev_Exe = "WindowsPackageManagerMCPServerDev.exe";
```
florelis, line 53: Doesn't really matter, but is there a reason to go with the long WPM name for this?

src/AppInstallerSharedLib/Public/winget/Filesystem.h
```diff
@@ -119,4 +119,7 @@ namespace AppInstaller::Filesystem
     // Gets the PathDetails used for the given path.
     // This is exposed primarily to allow for testing, GetPathTo should be preferred.
     PathDetails GetPathDetailsFor(PathName path, bool forDisplay = false);
+
+    // Gets the path to the executable for the given process id.
```
florelis: Are `HANDLE`s just the process ID? 🤯 
(Or is the comment wrong?)

src/AppInstallerCLIPackage/Package.appxmanifest
```diff
@@ -97,6 +97,18 @@
...
       </Extensions>
     </Application>
+    <Application Id="WinGetMCPServer" Executable="WinGetMCPServer\WinGetMCPServer.exe" EntryPoint="Windows.FullTrustApplication">
+      <uap:VisualElements DisplayName="WinGet Dev MCP Server" Square150x150Logo="Images\MedTile.png" Square44x44Logo="Images\AppList.png" Description="The WinGet MCP server." BackgroundColor="#0078d7" AppListEntry="none" >
+        <uap:DefaultTile/>
+      </uap:VisualElements >
+      <Extensions>
+        <uap5:Extension Category="windows.appExecutionAlias">
+          <uap5:AppExecutionAlias>
+            <uap5:ExecutionAlias Alias="WindowsPackageManagerMCPServerDev.exe" />
```
florelis, line 107: So we compile it with a name and use a different one for the App Execution Alias?
I think the people that like poking in the `WindowsApps` folder won't like it

src/AppInstallerCLIPackage/AppInstallerCLIPackage.wapproj
```diff
@@ -246,6 +246,13 @@
         <PackagePath>ConfigurationRemotingServer</PackagePath>
         <Recurse>true</Recurse>
       </WinGetAdditionalPackageFile>
+      <WinGetAdditionalPackageFile Include="$(WinGetAdditionalPackageFileRoot)\$(PlatformTarget)\$(Configuration)\WinGetMCPServer\net8.0-windows10.0.26100.0\$(ConfigServerRid)\**\*">
+        <PackagePath>WinGetMCPServer</PackagePath>
+        <Recurse>true</Recurse>
+      </WinGetAdditionalPackageFile>
+      <WinGetAdditionalPackageFile Include="$(WinGetAdditionalPackageFileRoot)\$(PlatformTarget)\$(Configuration)\Microsoft.Management.Deployment.OutOfProc\Microsoft.Management.Deployment.OutOfProc.dll">
+        <PackagePath>WinGetMCPServer\Microsoft.Management.Deployment.dll</PackagePath>
```
florelis, line 254: Is this just the projection?

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 namespace Microsoft.Management.Deployment
 {
-    [contractversion(12)] // For version 1.10
+    [contractversion(13)] // For version 1.12
```
florelis, line 5: I love this 😆

src/AppInstallerCLICore/Commands/McpCommand.cpp
```diff
@@ -0,0 +1,83 @@
...
+    McpCommand::McpCommand(std::string_view parent) :
+        Command("mcp", {}, parent, Settings::TogglePolicy::Policy::McpServer)
+    {
+    }
+
+    std::vector<Argument> McpCommand::GetArguments() const
+    {
+        return {
+            Argument{ Execution::Args::Type::ExtendedFeaturesEnable, Resource::String::ExtendedFeaturesEnableMessage, ArgumentType::Flag, Argument::Visibility::Help },
+            Argument{ Execution::Args::Type::ExtendedFeaturesDisable, Resource::String::ExtendedFeaturesDisableMessage, ArgumentType::Flag, Argument::Visibility::Help },
```
florelis, line 21: So now we can do both `winget mcp --disable` and `winget configure --disable`? Won't this be confusing?
I don't have a problem with enabling both together, but if I did `winget mcp --disable` I wouldn't expect to also lose configuration

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/WinGetVersion.cs
```diff
@@ -71,6 +71,12 @@ public WinGetVersion(string version)
         /// <returns>The WinGetVersion.</returns>
         public static WinGetVersion InstalledWinGetVersion(PowerShellCmdlet pwshCmdlet)
         {
+            string? comVersion = PackageManagerWrapper.Instance.GetVersion();
+            if (comVersion != null)
```
florelis, line 76: I'm guessing this can happen if using a newer version of the module with an older version of the package?
I would appreciate a comment about that. (Unless that's a common thing to check in all the code for the cmdlets. I'm not too familiar with it.)

src/WinGetMCPServer/WinGetMCPServer.csproj
```diff
@@ -0,0 +1,72 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <SupportedOSPlatformVersion>10.0.17763.0</SupportedOSPlatformVersion>
+    <Platforms>x64;x86;arm64</Platforms>
+    <OutputPath>$(SolutionDir)$(Platform)\$(Configuration)\$(MSBuildProjectName)\</OutputPath>
+    <SelfContained>true</SelfContained>
+    <RuntimeIdentifiers>win-x64;win-x86;win-arm64</RuntimeIdentifiers>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(Platform)' == 'x64' ">
```
florelis, line 16: Could just be a single `<RuntimeIdentifier>win-$(Platform)</RuntimeIdentifier>`

src/WinGetMCPServer/WinGetMCPServer.csproj
```diff
@@ -0,0 +1,72 @@
...
+
+  <Import Project="..\targets\EmbeddedCsWinRT.targets" />
+
+  <PropertyGroup>
+    <CsWinRTIncludes>
+      Microsoft.Management.Deployment;
+      Windows.Data.Text.TextSegmen;
+      Windows.Devices.Geolocation;
+      Windows.Foundation;
+      Windows.Globalization.DayOfWee;
```
florelis, line 49: Day of Wee always makes me laugh 😆 
Is there any reason to not have the full words?

src/WinGetMCPServer/WinGetMCPServer.csproj
```diff
@@ -0,0 +1,72 @@
...
+      Windows.System.Use;
+      Windows.System.IUse;
+    </CsWinRTIncludes>
+    <CsWinRTExcludes>
+      Windows.Foundation.PropertyType;
+      Windows.Storage.Provider;
+    </CsWinRTExcludes>
+    <CsWinRTWindowsMetadata>$(TargetWindowsVersion)</CsWinRTWindowsMetadata>
+    <!-- Ensure Support for Windows 10, Version 1809 -->
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
```
florelis, line 69: Could/should we tie this to `SupportedOSPlatformVersion`?

src/WinGetMCPServer/ServerConnection.cs
```diff
@@ -0,0 +1,40 @@
...
+// -----------------------------------------------------------------------------
+
+namespace WinGetMCPServer
+{
+    using Microsoft.Management.Deployment;
+
+    /// <summary>
+    /// Maintains the connection to the server.
+    /// </summary>
+    internal static class ServerConnection
```
florelis, line 14: If this means the winget COM server, I think it would be worth it to spell it out to not confuse it with the MCP server

src/WinGetMCPServer/ServerConnection.cs
```diff
@@ -0,0 +1,40 @@
...
+                if (packageManager == null)
+                {
+                    packageManager = new PackageManager();
+                }
+
+                try
+                {
+                    _ = packageManager.Version;
+                }
+                catch
```
florelis: Why is this needed?
My guess is that it is to be able to restart the COM server if it crashes. If that's the case, I would appreciate a comment about it

src/WinGetMCPServer/WingetPackageTools.cs
```diff
@@ -0,0 +1,266 @@
...
+            ReadOnly = false,
+            Destructive = true,
+            Idempotent = false,
+            OpenWorld = false)]
+        [Description("Install or update a package using WinGet")]
+        public async Task<CallToolResponse> InstallPackage(
+            [Description("The identifier of the WinGet package")] string identifier,
+            IProgress<ProgressNotificationValue> progress,
+            CancellationToken cancellationToken,
+            [Description("The catalog containing the package")] string? catalog = null)
```
florelis: I think we should use the word "source" instead of "catalog". I have no idea how MCP works, but I'm assuming the LLM would use this description in some way to prompt the user. End users see winget through the CLI and PS, which use "source". "Catalog" is only used in the context of COM

src/WinGetMCPServer/WingetPackageTools.cs
```diff
@@ -0,0 +1,266 @@
...
+                var packageCatalog = ConnectCatalog(catalog);
+
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    return PackageResponse.ForCancelBeforeSystemChange();
+                }
+
+                // First attempt a more exact match
+                var findResult = FindForIdentifier(packageCatalog, identifier, expandedFields: false);
+
```
florelis, line 99: Maybe check for cancellations in-between searches

src/WinGetMCPServer/WingetPackageTools.cs
```diff
@@ -0,0 +1,266 @@
...
+                // If nothing is found, expand to a looser search
+                if ((findResult.Matches?.Count ?? 0) == 0)
+                {
+                    findResult = FindForIdentifier(packageCatalog, identifier, expandedFields: true);
```
florelis, line 108: Do these not offer progress notifications?

src/WinGetMCPServer/WingetPackageTools.cs
```diff
@@ -0,0 +1,266 @@
...
+
+                var packageCatalog = ConnectCatalog(catalog);
+
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    return PackageResponse.ForCancelBeforeSystemChange();
+                }
+
+                // First attempt a more exact match
+                var findResult = FindForIdentifier(packageCatalog, identifier, expandedFields: false);
```
florelis, line 98: The search is using "FindForQuery", so it may return different results. I'm pretty sure people don't like that in the CLI, but I don't know how this ends up being called by the hosts to know if that would be an issue here.

src/WinGetMCPServer/WingetPackageTools.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        private static ProgressNotificationValue CreateInstallProgressNotification(ref InstallProgress installProgress)
+        {
+            string? message = null;
+
+            switch (installProgress.State)
+            {
+                case PackageInstallProgressState.Queued:
+                    message = "The install operation is queued";
```
florelis, line 241: Are we planning on localizing strings?

src/WinGetMCPServer/Response/FindPackageResult.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public string? Catalog { get; set; }
+
+        public bool? IsInstalled { get; set; }
+
+        public string? InstalledVersion { get; set; }
+
+        public string? InstalledLocation { get; set; }
+
+        public bool? IsUpdateAvailable { get; set; }
```
florelis, line 26: We don't include the available version?

src/WinGetMCPServer/Response/InstallOperationResult.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public string? Message { get; set; }
+
+        public bool? RebootRequired { get; set; }
+
+        public int? ErrorCode { get; set; }
+
+        public uint? InstallerErrorCode { get; set; }
+
+        public FindPackageResult? InstalledPackageInformation { get; set; }
```
florelis, line 22: Does MCP allow us to say what any of these mean?

src/WinGetMCPServer/Response/InstallOperationResult.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace WinGetMCPServer.Response
+{
+    /// <summary>
+    /// Contains information about an install operation.
+    /// </summary>
+    internal class InstallOperationResult
+    {
+        public string? Message { get; set; }
+
+        public bool? RebootRequired { get; set; }
```
florelis, line 16: We have information about this in the manifest, right? Could we report it ahead of time using the progress message? I think it would be nice to not get that information at the very end.

florelis (approved):

florelis:
Is this not going to be behind an experimental feature?

src/AppInstallerCLICore/Commands/ConfigureListCommand.cpp
JohnMcPMS: I don't disagree..., but I didn't feel like refactoring it out either.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: I wanted to leave it open to the Microsoft official terms usage rather than dictate it to them.

florelis:
Also, I'd really appreciate it if you could update doc/ReleaseNotes.md with this feature

src/AppInstallerCommonCore/Runtime.cpp
JohnMcPMS: Matches the name of the existing COM server executable.

src/AppInstallerSharedLib/Public/winget/Filesystem.h
JohnMcPMS: The comment is wrong because I wrote it with process id at first and then changed it when I realized what I actually had to pass in.  Might hold on changing it in case there aren't any other substantial issues.

src/AppInstallerCLIPackage/AppInstallerCLIPackage.wapproj
JohnMcPMS: No, it is the OOP activation thunk that makes the projection function.

src/AppInstallerCLICore/Commands/McpCommand.cpp
JohnMcPMS: There may come a time in the future when they aren't coupled.  If we think that it is a problem, maybe we need to conceptualize "extended features" more fully and integrate with experimental and the `feature` command.  @denelon FYI

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/WinGetVersion.cs
JohnMcPMS: This should be common any time two independently versionable processes interact.  And yes, the potential for failure is due to the module being newer than the COM server.

src/WinGetMCPServer/WinGetMCPServer.csproj
JohnMcPMS: Because of the way that CsWinRT does (did?) the matching here; it would only match types that start with this value AND are not equal to it (I'm sure that the code isn't doing that on purpose, but it is the effective result).

doc/admx/DesktopAppInstaller.admx
```diff
@@ -176,6 +176,16 @@
         <decimal value="0" />
       </disabledValue>
     </policy>
+    <policy name="EnableWindowsPackageManagerMcpServer" class="Machine" displayName="$(string.EnableWindowsPackageManagerMcpServer)" explainText="$(string.EnableWindowsPackageManagerMcpServerExplanation)" key="Software\Policies\Microsoft\Windows\AppInstaller" valueName="EnableWindowsPackageManagerMcpServer">
```
yao-msft, line 179: In case we have not done it, we'll need to talk to policy team about updating the policy template later.

src/Directory.Packages.props
```diff
@@ -10,13 +10,15 @@
...
     <PackageVersion Include="Microsoft.AspNet.WebApi.Client" Version="5.2.9" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="9.0.6" />
     <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageVersion Include="Microsoft.Msix.Utils" Version="2.1.1" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
     <PackageVersion Include="Microsoft.PowerShell.SDK" Version="7.4.6" />
     <PackageVersion Include="Microsoft.Win32.Registry" Version="5.0.0" />
     <PackageVersion Include="Microsoft.Windows.CsWinRT" Version="2.1.6" />
     <PackageVersion Include="Microsoft.Windows.SDK.Contracts" Version="10.0.26100.1742" />
+    <PackageVersion Include="ModelContextProtocol" Version="0.2.0-preview.3" />
```
yao-msft, line 21: nit: do we want to add a todo to update to stable version later?

src/AppInstallerCLICore/ConfigureExportCommand.cpp
```diff
@@ -3,6 +3,7 @@
 #include "pch.h"
 #include "ConfigureExportCommand.h"
 #include "Workflows/ConfigurationFlow.h"
+#include "Workflows/MSStoreInstallerHandler.h"
```
yao-msft, line 6: Now I'm wondering, why did we not put this command under Commands folder ...

JohnMcPMS:
> Is this not going to be behind an experimental feature?

I don't see it as needed given that the user needs to opt-in to using it by manually configuring their client.  The implementation of that block would either require reading the config in the MCP server or hacking in a check based on the caller info.

@denelon for opinion.

src/WinGetMCPServer/WingetPackageTools.cs
JohnMcPMS: Find does not provide progress.

src/Microsoft.Management.Deployment/Helpers.cpp
```diff
@@ -135,15 +137,18 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+                    return { packageFamilyName };
+                }
             }
 
             // if the caller doesn't have an AppUserModelID then fall back to the executable name
-            wil::unique_cotaskmem_string imageName = nullptr;
-            if (SUCCEEDED(wil::QueryFullProcessImageNameW(processHandle.get(), 0, imageName)) &&
-                (imageName.get() != nullptr))
+            std::filesystem::path executablePath = AppInstaller::Filesystem::GetExecutablePathForProcess(processHandle.get());
+            if (executablePath.has_filename())
```
yao-msft, line 150: If the executable path does not have file name, should we return the full path instead of empty?

src/WinGetMCPServer/WingetPackageTools.cs
JohnMcPMS: The search tool does:
1. ID equals OR Name equals OR Moniker equals (what the CLI uses for "targeted" commands)
2. ID contains OR Name contains OR Moniker equals (a looser version to provide some flexibility to the agent)

only using option 2 if nothing is found with option 1.

The install tool does:
1. Id equals
2. Id equals OR Name equals OR Moniker equals

similarly falling back to 2 only if nothing is found with 1.

The expectation is that the agent will always use the search tool to gather information, then use the result from that to inform the identifier field to use in the install tool.

src/WinGetMCPServer/WingetPackageTools.cs
JohnMcPMS: I don't know that it is helpful for the responses, but the progress might very well benefit from that.  I would prefer that to be a next step though for now.

src/WinGetMCPServer/Response/FindPackageResult.cs
JohnMcPMS: COM doesn't provide it easily.

src/WinGetMCPServer/Response/InstallOperationResult.cs
JohnMcPMS: No, as far as I know, there is no ability to provide *output* schemas, despite the MCP documentation suggesting that you should have them.

src/WinGetMCPServer/Response/InstallOperationResult.cs
JohnMcPMS: If I can get it in the search results, then I will add it there so that the agent can better reason over that fact.

src/WinGetMCPServer/WingetPackageTools.cs
```diff
@@ -0,0 +1,266 @@
...
+//     Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+// </copyright>
+// -----------------------------------------------------------------------------
+
+namespace WinGetMCPServer
+{
+    using Microsoft.Management.Deployment;
+    using ModelContextProtocol.Protocol;
+    using ModelContextProtocol.Server;
+    using System.ComponentModel;
```
yao-msft: nit: I guess Stylecop is not turned on for this project? It should complain that System. should be on the top.

src/WinGetMCPServer/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+            builder.Services
+                .AddMcpServer(configureOptions =>
+                {
+                    // TODO: More options setup?
+                    configureOptions.ServerInfo = new Implementation() { Name = ServerName, Version = ServerConnection.Instance.Version };
+                })
+                .WithStdioServerTransport()
+                .WithTools<WingetPackageTools>();
+
+            builder.Build().Run();
```
yao-msft, line 32: Does this handle shutdown events automagically by the framework?

yao-msft (approved):

src/Microsoft.Management.Deployment/Helpers.cpp
JohnMcPMS: I don't think so; that sounds like an error and logging would be appropriate.

src/WinGetMCPServer/Program.cs
JohnMcPMS: No, I doubt it, but I had thought of that.  Maybe my change to improve our quiesce story should include an event to make it easier for clients to act appropriately without needing to re-implement the entire "window to listen for shutdown messages".

Issue #5644, opened 2025-08-07
Group Policy Definition for Font Source


dkbennett:
### Description of the new feature / enhancement

Create a group policy definition for Font Source so that IT pro adopters can be prepared for the new policy existing with WinGet.

### Proposed technical implementation details

Add font source definition to group policy and related data structures. This is not adding the actual source definition, just the policy which it will use.

PR #5646, opened 2025-08-08
Add Font source group policy support

dkbennett:
This is for issue [5644](https://github.com/microsoft/winget-cli/issues/5644)

The purpose of this is to create a group policy definition in anticipation of the upcoming font support so IT administrators can be prepared for its inclusion and lead time.

It is not the intent of this checkin to actually implement the font source, but to create the policy so enterprises can be better prepared for it and have more lead time to adjust.

Changes:
* Added the Font Source group policy and associated tests and strings.

Verified build and groupPolicy tests.

-----


JohnMcPMS (dismissed):

dkbennett:
Just adding a comment here that this won't be checked in until there is a valid font index available , per #5647

This will stay in draft state until that is ready and then re-evaluated and updated if necessary and checked in with the font source availability.

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -48,6 +48,11 @@ namespace AppInstaller::Repository
         constexpr std::string_view s_Source_DesktopFrameworks_Data = "Microsoft.Winget.Platform.Source_8wekyb3d8bbwe"sv;
         constexpr std::string_view s_Source_DesktopFrameworks_Identifier = "Microsoft.Winget.Platform.Source_8wekyb3d8bbwe"sv;
 
+        constexpr std::string_view s_Source_Font_Name = "font"sv;
```
JohnMcPMS: I would prefer `winget-fonts`, but not sure if this has been discussed with @denelon yet.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -967,6 +967,10 @@ namespace Microsoft.Management.Deployment
         OpenWindowsCatalog,
         MicrosoftStore,
         DesktopFrameworks,
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 13)]
+        {
+            Font,
```
JohnMcPMS:
```suggestion
            OpenWindowsCatalogFonts,
```

src/AppInstallerRepositoryCore/SourceList.cpp
dkbennett: Renamed to winget-font per discussion with @denelon.

src/Microsoft.Management.Deployment/PackageManager.idl
dkbennett: Did this but had Font singular to be consistent with the rest of the source naming being singular.

PR #5652, opened 2025-08-11
Improve COM server quiescing

JohnMcPMS:
## Issue
When the package needs to be updated, the COM server does not currently shut down in response to the signals that it receives.  This is because it only shuts down when all COM object references have been released.  We need to be more proactive in termination as the process will be terminated shortly after we are notified either way.

## Change
This change refactors the existing shutdown signaling code out from `ExecutionContext.cpp` into its own, public location.  It then adds another system for coordinating complex shutdown that is triggered by the signaler.

Individual subsystems can register with the shutdown synchronizer with functions for:
- Disable new work
- Cancel current work
- Wait for quiesce

and on shutdown initialization, all subsystems are called for each stage in order.  When the final subsystem returns from waiting, the existing "all COM objects have been released" event is signaled and the process will exit.

Additionally, the shutdown signal handler waits for shutdown synchronization to complete in the window message function as after we return from that we are expected to be gone shortly after.  While this will lead to some increased number of non-quiescing processes, we want to do the best that we can to cooperatively exit.

> [!NOTE]
> Any actions being taken by callers outside of those described below will **NOT** block the process from exiting.

### Package Management
Registers with synchronization on `PackageManager` creation.  All of its functions lead back to the `ContextOrchestrator` which has been enlightened to handle the shutdown (and to operate as a contained set of instances for test purposes).

- **Disable new work** causes the orchestrator to refuse to queue items with an exception for new items and a cancellation for existing ones.
- **Cancel current work** causes all queued items to be cancelled, with different handling based on current status.
- **Wait for quiesce** waits on an event for each queue that is signaled when the queue is empty.

### Configuration
Registers with synchronization on `ConfigurationStaticFunctionsShim` creation (the `IConfigurationStatics` for OOP use).  All of its functions are routed through `IConfigurationStaticsInternals` to a new shutdown handler.  It tracks all of the async token objects used in the "background" sections of the `ConfigurationProcessor` functions.  This tracking is performed automatically by new types that manage that tracking with the lifetime.  It also enables non-async invocations to be cancelled by providing a dummy promise into the existing system.

- **Disable new work** causes an error whenever a new async token object is created.
- **Cancel current work** causes all tracked async tokens to have their `Cancel` function invoked.
- **Wait for quiesce** waits until all async tokens are destroyed.

## Validation
New tests are added for the `ContextOrchestrator`, the package management shutdown handling, and the configuration shutdown handling.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5652)


src/AppInstallerCLI.sln
```diff
@@ -968,6 +972,30 @@ Global
 		{9406322E-6272-487E-902A-9953889719EA}.ReleaseStatic|x64.Build.0 = ReleaseStatic|Any CPU
 		{9406322E-6272-487E-902A-9953889719EA}.ReleaseStatic|x86.ActiveCfg = ReleaseStatic|Any CPU
 		{9406322E-6272-487E-902A-9953889719EA}.ReleaseStatic|x86.Build.0 = ReleaseStatic|Any CPU
+		{7139ED6E-8FBC-0B61-3E3A-AA2A23CC4D6A}.Debug|ARM64.ActiveCfg = Debug|x64
```
yao-msft: nit: through all the new entries, I think only debug/release x64/x86 need Build.0 entries, all other combinations should be set to "not build"

src/AppInstallerCLICore/ContextOrchestrator.h
```diff
@@ -119,8 +124,25 @@ namespace AppInstaller::CLI::Execution
...
+        static void StaticCancelQueuedItems(CancelReason reason);
+        static void StaticWaitForRunningItems();
+
+        void Disable(CancelReason reason);
+        void CancelQueuedItems(CancelReason reason);
+        void WaitForRunningItems();
+
+        // Waits for running items to complete; waits up to full time out in *each* queue.
+        // Returns true to indicate all queues are empty before the timeout.
+        // Intended for test use.
```
yao-msft: nit: should we wrap around AICLI_DISABLE_TESTES if it's just for test purpose?

src/AppInstallerCLICore/ContextOrchestrator.h
```diff
@@ -151,19 +173,32 @@ namespace AppInstaller::CLI::Execution
...
+        // Cancels and "removes" all items in the queue.
+        void CancelAllItems(CancelReason reason);
+
+        // Waits until the empty queue event is signaled.
+        // Intended for use during shutdown handling.
+        void WaitForEmptyQueue();
+
+        // Waits until the empty queue event is signaled.
+        // Returns true to indicate the queue is empty before the timeout.
+        // Intended for tests.
```
yao-msft: nit: should we wrap around AICLI_DISABLE_TESTES if it's just for test purpose?

src/AppInstallerCLI.sln
JohnMcPMS: Agree on the configurations, disagree on removing from the ARM64 build (but updated things to make that work).

src/WinGetTestCommon/WinGetServerInstance.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        /// <summary>
+        /// The window name for the COM server message window.
+        /// </summary>
+        public const string TargetWindowName = "WingetMessageOnlyWindow";
+
+        /// <summary>
+        /// Gets the process for the server.
+        /// </summary>
+        public required Process Process { get; init; }
```
florelis, line 37: I didn't know `required` was a thing. That's cool

src/AppInstallerCLICore/ContextOrchestrator.cpp
```diff
@@ -145,13 +157,103 @@ namespace AppInstaller::CLI::Execution
...
+            queue.second->WaitForEmptyQueue();
+        }
+    }
+
+    bool ContextOrchestrator::WaitForRunningItems(DWORD timeoutMilliseconds)
+    {
+        std::lock_guard<std::mutex> lock{ m_queueLock };
+        for (const auto& queue : m_commandQueues)
+        {
+            if (!queue.second->WaitForEmptyQueue(timeoutMilliseconds))
```
florelis, line 222: Nit: I would expect this function to wait for _all_ items at most `timeoutMilliseconds`, but it waits that long for each of them. This may be moot depending on how many command queues we have or how big the timeout is, though.

src/AppInstallerCLICore/ContextOrchestrator.cpp
```diff
@@ -299,14 +407,71 @@ namespace AppInstaller::CLI::Execution
...
+
+    bool OrchestratorQueue::WaitForEmptyQueue(DWORD timeoutMilliseconds)
+    {
+        return m_queueEmpty.wait(timeoutMilliseconds);
+    }
+
+    std::string OrchestratorQueue::GetStatusString()
+    {
+        std::ostringstream stream;
+        stream << m_commandName << '[' << m_allowedThreads << "]\n";
```
florelis, line 450: nit: `endl`

yao-msft (approved):

PR #5659, opened 2025-08-13
Update localization files to standardize 'ID' to 'Id' across multiple…

6rube:
… languages

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5659)

### DIFF
--- a/Localization/Resources/de-DE/winget.resw
+++ b/Localization/Resources/de-DE/winget.resw
@@ -463,7 +463,7 @@ Sie können über die Einstellungsdatei „winget settings“ konfiguriert werde
     <value>Suchen und Anzeigen grundlegender Informationen zu Paketen</value>
   </data>
   <data name="SearchId" xml:space="preserve">
-    <value>ID</value>
+    <value>Id</value>
     <comment>Abbreviation of Identifier.</comment>
   </data>
   <data name="SearchMatch" xml:space="preserve">
--- a/Localization/Resources/fr-FR/winget.resw
+++ b/Localization/Resources/fr-FR/winget.resw
@@ -463,7 +463,7 @@ Elles peuvent être configurées par le biais du fichier de paramètres « wing
     <value>Rechercher et afficher des informations de base sur les packages</value>
   </data>
   <data name="SearchId" xml:space="preserve">
-    <value>ID</value>
+    <value>Id</value>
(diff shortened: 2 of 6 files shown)


6rube:

Trenly (changes requested):
Please only update the en-US file. Any changes to other files will be overwritten when the localization team re-processes the localizations.

Otherwise, if the en-US file is already accurate, someone on the engineering team needs to submit a defect to the localization team

> Files under Localization folder will be overwritten when we do a localization update, so no changes under Localization folder will take effect.
_Originally posted by @yao-msft in https://github.com/microsoft/winget-cli/issues/3250#issuecomment-1551990405_

JohnMcPMS:
The correct thing to do is replace `Id` with `Identifier` (in the en-us file) because abbreviations are not the best for localization.  The only advantage that the series of letters `i`+`d` have is that they match the series of letters used in the flag `--id`.  But we have localized descriptions that should contain the word `identifier`, hopefully matching the word that would be used in translating `Identifier` as a single noun.

6rube:
> Please only update the en-US file. Any changes to other files will be overwritten when the localization team re-processes the localizations.
> 
> Otherwise, if the en-US file is already accurate, someone on the engineering team needs to submit a defect to the localization team
> 
> > Files under Localization folder will be overwritten when we do a localization update, so no changes under Localization folder will take effect.
> > _Originally posted by @yao-msft in [#3250 (comment)](https://github.com/microsoft/winget-cli/pull/3250#issuecomment-1551990405)_

Is there a way to contribute to the the localization team re-processes ? I just wanted to fix the inconsistency I saw when dealing with winget in different languages. Would you otherwise so kind and give this information to someone on the engineering team ?

6rube:
> The correct thing to do is replace `Id` with `Identifier` (in the en-us file) because abbreviations are not the best for localization. The only advantage that the series of letters `i`+`d` have is that they match the series of letters used in the flag `--id`. But we have localized descriptions that should contain the word `identifier`, hopefully matching the word that would be used in translating `Identifier` as a single noun.

Id has also the property of being a combination used in other languages. Identifier would need to be translated again. Maybe it would be best to change the en-us file to Id.

PR #5661, opened 2025-08-13
Improve issue forms & add corresponding label triggers

mdanish-kh:
## Change

PR improves the issues forms by giving the user the option to select the relevant area, as well as the relevant CLI command if applicable. Updated the policy bot triggers to automatically apply certain labels based on the user selection from the issue body.

cc @Trenly @denelon
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5661)

### DIFF
--- a/.github/policies/labelManagement.issueOpened.yml
+++ b/.github/policies/labelManagement.issueOpened.yml
@@ -23,6 +23,7 @@ configuration:
         then:
           - addLabel:
               label: Needs-Triage
+      ### Area Labels
       - description: Add area label for PowerShell Module
         if:
           - payloadType: Issues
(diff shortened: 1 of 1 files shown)


mdanish-kh:
@denelon if you could create these labels:

- Command-Repair
- Command-Font
- Command-DSCv3
- Command-MCP

.github/policies/labelManagement.issueOpened.yml
```diff
@@ -23,5 +23,260 @@ configuration:
...
           - addLabel:
               label: Needs-Triage
+      ### Area Labels
+      - description: Add area label for PowerShell Module
+        if:
+          - payloadType: Issues
+          - isAction:
+              action: Opened
+          - bodyContains:
+              pattern: 'Relevant area\(s\)\s*\n.*PowerShell Module'
```
Trenly, line 33: These patterns should be bounded. I'd suggest adding `^\s*` to check that it is at the beginning of a line, and `\s*$` to ensure that the relevant area ends in a new line. This will prevent matching against something like `I don't think the relevant area should be the PowerShell Module at all`

mdanish-kh: GitHub puts them in a single line separated by commas as:

```markdown
### Relevant area(s)

WinGet CLI, PowerShell Module, COM API

### Relevant command(s)

winget configure, winget download, winget import
```

I can add the `^\s*` to assert `Relevant area(s)` appears at the start (although not strictly needed too), but the ending one would not be needed as it would force the pattern to appear at the end, whereas a pattern like `winget download` can exist in the middle of the line as well

mdanish-kh: actually, the actual body would be something like

```
### Relevant command(s)
```

& `^\s*Relevant command` would not match the `###` at the start & not worth matching that? I think the current matching is suitable?

mdanish-kh:
Deciding to leave the matching is for now. @Trenly suggested this stricter version should the need ever arise

```
Relevant\scommand\(s\)$\s*^(winget\s*\S*(,\s)?)*(download)((,\s)?winget\s*\S*)*$
```

Notes: @Trenly notes that \S may not be supported by the GitOps so we can change it to `a-zA-Z` if needed

denelon (approved):

PR #5687, opened 2025-08-25
Update schema to 1.12 with Font InstallerType

### LINKED ISSUE #5686: Schema Update for Fonts
### Relevant area(s)

WinGet CLI

Update the WinGet Schema for Fonts to be installer types with appropriate validation. This is so tooling can pick up the updates and begin to validate manifests with fonts.

* Schema update to allow Fonts to be an InstallerType and NestedInstallerType.
* Adding Font to the Manifest common as an installer type.
* Tests associated with a font Schema update
* Any changes necessary to allow the NestedInstallerType for Fonts and their usage to be valid.

dkbennett:
* Updates schema to 1.12 and adds type "font" as a valid InstallerType and NestedInstallerType in the schema.
* Adds InstallerType Font to the manifest and associated Enum conversions.
* Updated some of the NestedInstaller validation that had assumed Portable type as the only type that could have multiple installer files (fonts can also have many fonts in a package).
* Added validation for 1.12 fonts in the manifest and associated tests for the schema version.

This PR does not add any functional support for Fonts, only the validation and schema update so the new schema can be used and is valid in the related WinGet tooling.

Tested:
* Passed the AppInstallerCLITests with [ManifestValidation] and ensured a clean result.
* Added new tests for the v 1.12 manifest updated with the new installer types in the test installer manifest to ensure Fonts are correctly passing the validation.

Closes #5686 

-----


src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
```diff
@@ -252,6 +253,13 @@ namespace AppInstaller::Manifest
     };
 
     static const std::unordered_set<std::filesystem::path> s_AllowedPortableFiletypes = { L".exe" };
+    static const std::unordered_set<std::filesystem::path> s_AllowedFontFiletypes = {
```
JohnMcPMS: I probably didn't see this during the portable review, but it would be much better if we didn't use a static that required dynamic allocation/destruction.  I would think that a `std::array<std::wstring_view>` should be able to be read-only memory (after loader fixups).

Also, we really shouldn't be defining this in the header.

JohnMcPMS (commented):
Would it be possible to get two more commits added:
1. Move your current 1.12 schema files out and copy the 1.11 files into the 1.12 directory, otherwise unchanged.  Commit this.
2. Move your 1.12 files back in over the 1.11 files.  Commit this.

Without at least a baseline of the 1.11 schema files, it is extremely annoying to review the changes between 1.11 and 1.12.

Another option that might set things up for success in the future would be to call the 1.12 directory `latest` and set the precedent of the pattern being:
1. copy the `latest` directory to the appropriate previous version name [so next time it would be `v1.12.0`]
2. update `latest` with the new changes

If you do this, make sure to create a commit in `latest` with 1.11 before putting the 1.12 files there.

schemas/JSON/manifests/v1.12.0/manifest.defaultLocale.1.12.0.json
```diff
@@ -0,0 +1,280 @@
...
+    },
+    "ManifestType": {
+      "type": "string",
+      "default": "defaultLocale",
+      "const": "defaultLocale",
+      "description": "The manifest type"
+    },
+    "ManifestVersion": {
+      "type": "string",
+      "default": "1.10.0",
```
yao-msft:
```suggestion
      "default": "1.12.0",
```

schemas/JSON/manifests/v1.12.0/manifest.locale.1.12.0.json
```diff
@@ -0,0 +1,271 @@
...
+    },
+    "ManifestType": {
+      "type": "string",
+      "default": "locale",
+      "const": "locale",
+      "description": "The manifest type"
+    },
+    "ManifestVersion": {
+      "type": "string",
+      "default": "1.10.0",
```
yao-msft:
```suggestion
      "default": "1.12.0",
```

schemas/JSON/manifests/v1.12.0/manifest.version.1.12.0.json
```diff
@@ -0,0 +1,46 @@
...
+    },
+    "ManifestType": {
+      "type": "string",
+      "default": "version",
+      "const": "version",
+      "description": "The manifest type"
+    },
+    "ManifestVersion": {
+      "type": "string",
+      "default": "1.10.0",
```
yao-msft:
```suggestion
      "default": "1.12.0",
```

schemas/JSON/manifests/v1.12.0/manifest.locale.1.12.0.json
dkbennett: Thanks! I fixed this when I did the v1.10.0 -> latest and then updated latest to v1.12.0, which should be in the most recent commit with a much easier to see diff.

schemas/JSON/manifests/v1.12.0/manifest.version.1.12.0.json
dkbennett: Fixed

schemas/JSON/manifests/v1.12.0/manifest.defaultLocale.1.12.0.json
dkbennett: Fixed

src/AppInstallerCommonCore/Public/winget/ManifestValidation.h
dkbennett: Fixed by changing to constexpr std::array<std::wstring_view, n> and putting into the anonymous namespace in ManifestValidation.cpp, for both font and portable filetypes, for read only memory with no more allocation/deallocation at runtime.

dkbennett:
> Would it be possible to get two more commits added:
> 
> 1. Move your current 1.12 schema files out and copy the 1.11 files into the 1.12 directory, otherwise unchanged.  Commit this.
> 2. Move your 1.12 files back in over the 1.11 files.  Commit this.
> 
> Without at least a baseline of the 1.11 schema files, it is extremely annoying to review the changes between 1.11 and 1.12.
> 
> Another option that might set things up for success in the future would be to call the 1.12 directory `latest` and set the precedent of the pattern being:
> 
> 1. copy the `latest` directory to the appropriate previous version name [so next time it would be `v1.12.0`]
> 2. update `latest` with the new changes
> 
> If you do this, make sure to create a commit in `latest` with 1.11 before putting the 1.12 files there.
> 
> [Note, 1.11 is actually the stand in for N-1, which I now think was 1.10?]

Fixed this using your future-proofing recommendation:
- Created 'latest' folder for schema.
- Copied 1.10 into latest, renamed version to 'latest' as well in the filename.
- Committed the above.
- Moved 1.12 into latest, renamed file version to be 'latest' for exact file diffing.
- Committed that. The clean diff for easy viewing is now here: https://github.com/microsoft/winget-cli/pull/5687/commits/e0a1a012227fbb63a84336604563bc2beb500d45
- Updated the resource for 1.12 to point to the latest.

PR #5689, opened 2025-08-26
Download MS Store package for target OS

### LINKED ISSUE #4996: Ability for winget to download a Microsoft Store application for specific Windows build

I would like to have the possibility to download Microsoft Store application for a specific version. Since by default, winget always downloads the latest version even if incompatible. This disallows to download LPX since one need to download the matching version. Also, for older version of Windows (LTSC) you can stumble upon incompatible application version.

This solves #4995.

# The problem
For example, for Microsoft.LanguageExperiencePackfr-FR (9NHMG4BJKMDG): each build has its version with the [build].a.b.c format. The last for 24H2 is 26100.22.41.0 and for Windows 10 21H2/22H2 the last is 19041.79.263.0. However, Winget only download 27754.1.2.0, for the future 25H2. Its manifest confirm it is not compatible with our Windows versions.
`<TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.27547.0" MaxVersionTested="10.0.27754.1000" />
`
Another example with Microsoft.Windows.Photos (9WZDNCRFJBH4) : for the Windows 10 1809 (LTSC), build 17763, the latest compatible is 2023.10030.27002.0. The last version compatible with this build in Winget downloaded version is 2019.19061.18920.0.

# Example solution
I would see for an example of solution the use of the "-v" flag in winget. For now, it is not used for Store downloads.
`winget download 9NHMG4BJKMDG -v 22000`
This would download the Windows 11 22H2 French LPX.

Another possibility is to make a filter with the platform name `-v "Windows.Desktop=22000"`.

## Technical implementation details
This would be easily technically implementable without any infrastructure change, only code. Winget calls two API:
- https://displaycatalog.mp.microsoft.com/v7.0/products/${ProductId}?fieldsTemplate=Details&market=FR&languages=fr-fr&catalogIds=4 : to get the app information, WuCategoryId to download it and all versions available and their min build.
- https://storeapps.api.cdp.microsoft.com/api/v2/contents/storeapps/namespaces/default/names/$WuCategoryId/versions/$version/files?action=GenerateDownloadInf : to get the download urls given the WuCategoryId.

The first call is done here [MSStoreDownload.cpp](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCommonCore/MSStoreDownload.cpp) and the other here [SFSClientImpl.cpp](https://github.com/microsoft/sfs-client/blob/main/client/src/details/SFSClientImpl.cpp).

Given these APIs winget already have all information at hands. It has all versions but Winget only choose the latest version and ignore the others. It just has to implement a best version for given build algorithm. This algorithm would have to be implemented in these two files.

Thanks, and have a nice day

JohnMcPMS:
Fixes #4996 

## Change
Adds the ability to specify the target OS version for downloads from the Microsoft Store.  This is achieved by filtering the set of applicable platforms in `GetSfsPackageFileSupportedPlatforms` further than the existing platform targeting.  Now they must also have a minimum specified OS version >= the target OS version.

`winget.exe download` has a new parameter `--os-version`.  This is a UINT64 version (4 part version with 2^16-1 maximum value for each part) that should be given as the target OS version.

COM `DownloadOptions` has a new property `TargetOSVersion`, which is a string of the same format as above.

PowerShell `Export-WinGetPackage` has a new parameter `-TargetOSVersion`, which is also a string of the same format.

Also added the previously implemented options to skip the license download (`--skip-license`) and the target platform (`--platform`) to COM (`SkipMicrosoftStoreLicense` and `Platform`) and PowerShell (`-SkipMicrosoftStoreLicense` and `-Platform`).

## Validation
Added a test to the existing ones for specifying a lower target version.

Manually verified that without the parameter provided, for `9NHMG4BJKMDG` (as exemplified in the issue) we download version `27920.1.1.0`.  Specifying version `10.0.22000.0` as the target OS, we download package version `22000.52.238.0` which has in it's manifest:
```XML
<TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.21251.0" MaxVersionTested="10.0.22000.2652" />
```
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5689)


src/AppInstallerCommonCore/MSStoreDownload.cpp
```diff
@@ -670,17 +698,44 @@ namespace AppInstaller::MSStore
             return Utility::Architecture::Unknown;
         }
 
-        std::vector<Manifest::PlatformEnum> GetSfsPackageFileSupportedPlatforms(const SFS::AppFile& appFile, Manifest::PlatformEnum requiredPlatform)
+        std::vector<Manifest::PlatformEnum> GetSfsPackageFileSupportedPlatforms(
+            const SFS::AppFile& appFile,
+            Manifest::PlatformEnum requiredPlatform,
+            const std::optional<Utility::UInt64Version>& targetOSVersion)
         {
+            UNREFERENCED_PARAMETER(targetOSVersion);
```
yao-msft: nit: remove

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1203,6 +1221,18 @@ namespace Microsoft.Management.Deployment
             /// Authentication arguments used when downloading the package installer if authentication is required.
             AuthenticationArguments AuthenticationArguments;
         }
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 13)]
```
yao-msft, line 1225: nit: the contract version comment in line 5 is still 1.12

yao-msft (approved):

src/AppInstallerCommonCore/Public/winget/MSStoreDownload.h
```diff
@@ -41,6 +41,8 @@ namespace AppInstaller::MSStore
             std::string locale,
             AppInstaller::Authentication::AuthenticationArguments authArgs);
 
+        void TargetOSVersion(std::optional<Utility::UInt64Version> targetOSVersion);
```
yao-msft, line 44: nit: should we move it to constructor? I don't see it needs to be changed after it's created.

src/AppInstallerCommonCore/MSStoreDownload.cpp
```diff
@@ -941,13 +1000,13 @@ namespace AppInstaller::MSStore
...
             // Populate main packages
-            result.MainPackages = PopulateSfsAppFileToMSStoreDownloadFileVector(appContent.GetFiles(), requiredArchitecture, requiredPlatform);
+            result.MainPackages = PopulateSfsAppFileToMSStoreDownloadFileVector(appContent.GetFiles(), requiredArchitecture, requiredPlatform, targetOSVersion);
 
             // Populate dependency packages
             for (auto const& dependencyEntry : appContent.GetPrerequisites())
             {
                 // Not passing in required platform for dependencies. Dependencies are mostly Windows.Universal.
-                auto dependencyPackages = PopulateSfsAppFileToMSStoreDownloadFileVector(dependencyEntry.GetFiles(), requiredArchitecture);
+                auto dependencyPackages = PopulateSfsAppFileToMSStoreDownloadFileVector(dependencyEntry.GetFiles(), requiredArchitecture, Manifest::PlatformEnum::Unknown, targetOSVersion);
```
yao-msft, line 1008: Is it possible for dependency packages to have higher min OS version than main package? Maybe we can pass nullopt for targetOSVersion?

src/Microsoft.Management.Deployment/PackageManager.idl
JohnMcPMS: Correctly so; that comment means that it has already been incremented for 1.12.

src/AppInstallerCommonCore/Public/winget/MSStoreDownload.h
JohnMcPMS: I'm not that concerned with invariants in an internal type and properties are much more optionally-set friendly.  I do dislike large numbers of arguments to functions though.

src/AppInstallerCommonCore/MSStoreDownload.cpp
JohnMcPMS: If we are talking about MSIX packages and scenarios where no versions have been removed (security, etc.), then I would say no.  There must have been a dependency that was also installable on the target OS version at some time.

yao-msft (approved):

PR #5701, opened 2025-08-28
Improve slow searches involving installed items

### LINKED ISSUE #5629: Winget command --word --commandline --parameter freezes

Using winget autocomplete in PowerShell (see https://github.com/microsoft/winget-cli/blob/master/doc/Completion.md) can cause a hang/freeze or at worst run extremely slowly. After some investigation, it's the `winget complete` feature itself that hangs. 

Specifically, here's a command line that freezes/hangs every time:
```
winget complete --word="Micro" --commandline="winget upgrade Micro" --position=2
```
I've verified this on both the 7.x PowerShell, legacy PowerShell (all with --noprofile) and even CMD.EXE for grins. 

Position doesn't seem to matter (it's not clear in the docs if position is 0 or 1 index based, I assumed 0, but it didn't matter here what value I used for position).

The value in `--word` and the the parameter to `winget upgrade` must match.

I have noticed that sometimes it does eventually respond, but this can upwards of a minute depending on the search criteria.

Winget version  is 1.11.400.

### Steps to reproduce

Execute the following command line:
```
winnget complete --word="Micro" --commandline="winget upgrade Micro" --position=2
```

### Expected behavior

Expected a list of apps that start with "Micro" to appear quickly.

### Actual behavior

The command either hangs or freezes or takes upwards of a minute to respond.

### Environment

```shell
❯ winget --info
Windows Package Manager v1.11.400

Windows: Windows.Desktop v10.0.22631.5624
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.26.400.0
```

JohnMcPMS:
Fixes #5629 
(and probably quite a few more list/upgrade some-query-provided performance related issues)

While things could still be made better, the good case (where files were already cached locally) of the previous behavior trying to complete `winget list Micro` was taking 51 seconds.  With this change, it takes 6 seconds on the debug build.  Given that there is intentionally little to no I/O and much of the CPU time is likely spent in our code, that should speed up more with the release.

## Issue
The indexed `IPackage` implementation for v2 needed to acquire the intermediate manifest to provide the response for `GetVersionKeys`.  Since this is used extensively to perform correlation between available packages and local items during a search, any available packages found by the query would then result in a download/disk read of the intermediate manifest.

## Change
Introduces `IPackage::GetMultiProperty`, allowing code that just wanted to get all of these values across all versions to do so more efficiently.  For the v2 index, this is how the data is already stored in the database.  For the other `IPackage` implementations, the data was also present and is just aggregated from each version's output.

## Validation
Added test for calling new interface function.
Added validation across `CompositeSource` tests to ensure that the functions known to need non-indexed data are not called.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5701)

### DIFF
--- a/src/AppInstallerRepositoryCore/RepositorySearch.cpp
+++ b/src/AppInstallerRepositoryCore/RepositorySearch.cpp
@@ -117,7 +117,7 @@ namespace AppInstaller::Repository
         case PackageMultiProperty::Tag: return PackageVersionMultiProperty::Tag;
         case PackageMultiProperty::Command: return PackageVersionMultiProperty::Command;
         default:
-            THROW_HR(E_UNEXPECTED);
+            THROW_HR_MSG(E_UNEXPECTED, "PackageMultiProperty must map to a PackageVersionMultiProperty");
         }
     }
 
--- a/src/AppInstallerRepositoryCore/RepositorySearch.cpp
+++ b/src/AppInstallerRepositoryCore/RepositorySearch.cpp
@@ -105,6 +105,22 @@ namespace AppInstaller::Repository
         return Version.empty() && Channel.empty();
     }
 
+    PackageVersionMultiProperty PackageMultiPropertyToPackageVersionMultiProperty(PackageMultiProperty property)
+    {
+        switch (property)
+        {
+        case PackageMultiProperty::PackageFamilyName: return PackageVersionMultiProperty::PackageFamilyName;
+        case PackageMultiProperty::ProductCode: return PackageVersionMultiProperty::ProductCode;
+        case PackageMultiProperty::UpgradeCode: return PackageVersionMultiProperty::UpgradeCode;
+        case PackageMultiProperty::NormalizedName: return PackageVersionMultiProperty::Name;
+        case PackageMultiProperty::NormalizedPublisher: return PackageVersionMultiProperty::Publisher;
+        case PackageMultiProperty::Tag: return PackageVersionMultiProperty::Tag;
+        case PackageMultiProperty::Command: return PackageVersionMultiProperty::Command;
+        default:
+            THROW_HR(E_UNEXPECTED);
+        }
+    }
+
     const char* UnsupportedRequestException::what() const noexcept
     {
         if (m_whatMessage.empty())
--- a/src/AppInstallerRepositoryCore/Rest/RestSource.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/RestSource.cpp
@@ -207,19 +207,19 @@ namespace AppInstaller::Repository::Rest
             switch (property)
             {
             case PackageVersionMultiProperty::PackageFamilyName:
-                for (std::string pfn : versionInfo.PackageFamilyNames)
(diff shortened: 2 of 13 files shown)


src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -195,6 +197,99 @@ namespace AppInstaller::Repository::Rest
...
+            PackageVersionMultiProperty property,
+            std::vector<Utility::LocIndString>& result,
+            void (*Action)(std::vector<Utility::LocIndString>&, Utility::LocIndString&&))
+        {
+            switch (property)
+            {
+            case PackageVersionMultiProperty::PackageFamilyName:
+                for (std::string pfn : versionInfo.PackageFamilyNames)
+                {
+                    Action(result, Utility::LocIndString{ pfn });
```
yao-msft, line 212: nit: std::move(pfn) and same for several below?

src/AppInstallerRepositoryCore/RepositorySearch.cpp
```diff
@@ -105,6 +105,22 @@ namespace AppInstaller::Repository
...
+        {
+        case PackageMultiProperty::PackageFamilyName: return PackageVersionMultiProperty::PackageFamilyName;
+        case PackageMultiProperty::ProductCode: return PackageVersionMultiProperty::ProductCode;
+        case PackageMultiProperty::UpgradeCode: return PackageVersionMultiProperty::UpgradeCode;
+        case PackageMultiProperty::NormalizedName: return PackageVersionMultiProperty::Name;
+        case PackageMultiProperty::NormalizedPublisher: return PackageVersionMultiProperty::Publisher;
+        case PackageMultiProperty::Tag: return PackageVersionMultiProperty::Tag;
+        case PackageMultiProperty::Command: return PackageVersionMultiProperty::Command;
+        default:
+            THROW_HR(E_UNEXPECTED);
```
dkbennett: I personally get quite frustrated whenever I see a program give E_UNEXPECTED. Perhaps E_INVALIDARG?

yao-msft (dismissed):

dkbennett (dismissed):

src/AppInstallerRepositoryCore/RepositorySearch.cpp
JohnMcPMS: My expectation is that this error can only be produced by an error on the part of the developer, and that said error should be evident to the developer before it ever makes it to PR.  Invalid argument makes no more sense than unexpected if the exception makes it to the end user.  In fact, it can be more confusing because they might think that it was their fault.

I've added a message at least.

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
JohnMcPMS: I chose to go with properly using `const&` for the enumeration instead.

yao-msft (approved):

Issue #5704, opened 2025-08-29
Support wildcard in `-Version` for `Repair-WinGetPackageManager


Trenly:
### Relevant area(s)

PowerShell Module

### Description of the new feature / enhancement

I'd like to be able to use `Repair-WinGetPackageManager -Version 1.11.*` to get the latest version of WinGet 1.11 that is not a pre-release version.

I'd like to be able to use `Repair-WinGetPackageManager -Version 1.11.* -IncludePrerelease` to get the latest version of WinGet 1.11 that is either a stable release or a pre-release

I'd like to be able to use `Repair-WinGetPackageManager -Version 1.11.3* -IncludePrerelease` to get the latest version of WinGet 1.11 that matches that pattern (`1.11.370-preview`)

Issue #4947, opened 2024-11-06
Improvements to make Repair a more complete bootstrapping solution


JohnMcPMS:
### Description of the new feature / enhancement

Suggestions from one of our very dedicated users:

1. Ensure that Repair also installs the winget-cli source package.
  a. This is actually a workaround for an OS issue that prevents winget.exe from installing it in certain scenarios
  b. An alternative solution would be to move away from installing it at all and have all code paths use the unpackaged path (this does have a performance cost though)
2. Fix `Win32Exception An error occurred trying to start process 'winget.exe'` error the first time Repair -AllUsers is run
  a. Possibly only repros in the headless scenario
3. Repair -Force should bypass checks and attempt to install everything

They also suggest effectively #4390 , but including all PowerShell modules.

florelis:
It should also install the unpackaged vcredist for that weird admin DLL load path bug. We have several issues open for "does not run as system" where the solution is to manually install it (e.g. #4798); might as well automate that.

PR #5711, opened 2025-09-08
PowerShell Repair enhancements

AmelBawa-msft:
- > Register for the user (when appropriate) after Repair -AllUsers

  Register AppInstaller when using `AddProvisionPackageAsync` which was not registering the package prior to this PR change.

- > Ensure that Repair also installs the winget-cli source package

  Added a "best effort" approach to programmatically attempt to downloading the source2.msix and install it.

- > Fix Win32Exception An error occurred trying to start process 'winget.exe' error the first time Repair -AllUsers is run

  Keeping this as-is since the code includes a descriptive message. In my repro, the issue occurred when App Installer was not installed, and the code attempted to invoke winget.exe to check if the application was present.

- > Repair -Force should bypass checks and attempt to install everything

  This is not addressed in this PR. It's unclear which constraints should be relaxed, as the current implementation already attempts to ensure the requested parameters are satisfied. The Force flag is currently used to trigger `ForceTargetApplicationShutdown` during package addition, which appears appropriate. I recommend opening a separate issue to propose and discuss the desired behavior.

- > I'd like to be able to use Repair-WinGetPackageManager -Version 1.11.* to get the latest version of WinGet 1.11 that is not a pre-release version.

  Added support for leading and trailing wildcard: `1.2.*`, `1.2*`, `1.*2.3`

### Related issues:
- #4947
- #5704

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5711)

### DIFF
--- a/src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/GitHubClient.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/GitHubClient.cs
@@ -95,7 +95,7 @@ namespace Microsoft.WinGet.Client.Engine.Helpers
             var allWinGetReleases = allReleases.Select(r => new WinGetVersion(r.TagName));
             if (TryGetLatestMatchingVersion(allWinGetReleases, version, includePrerelease, out var latestVersion))
             {
-                return latestVersion.TagVersion;
+                return latestVersion!.TagVersion;
             }
 
             return null;
--- a/src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/WinGetVersion.cs
+++ b/src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/WinGetVersion.cs
@@ -56,16 +56,6 @@ namespace Microsoft.WinGet.Client.Engine.Helpers
         }
 
         /// <summary>
-        /// Checks if the version string has a wildcard.
-        /// </summary>
-        /// <param name="version">The version string.</param>
-        /// <returns>True if it has a wildcard, false otherwise.</returns>
-        public static bool VersionHasWildcard(string version)
-        {
-            return version.Contains("*");
-        }
-
-        /// <summary>
         /// Gets the version as it appears as a tag.
         /// </summary>
         public string TagVersion { get; }
@@ -113,6 +103,16 @@ namespace Microsoft.WinGet.Client.Engine.Helpers
             }
 
             return new WinGetVersion(versionResult.StdOut.Replace(Environment.NewLine, string.Empty));
+        }
+
+        /// <summary>
+        /// Checks if the version string has a wildcard.
+        /// </summary>
+        /// <param name="version">The version string.</param>
+        /// <returns>True if it has a wildcard, false otherwise.</returns>
+        public static bool VersionHasWildcard(string version)
+        {
+            return version.Contains("*");
         }
 
         /// <summary>


src/PowerShell/Help/Microsoft.WinGet.Client/Repair-WinGetPackageManager.md
```diff
@@ -17,7 +17,7 @@ Repairs the installation of the WinGet client on your computer.
 ### IntegrityVersionSet (Default)
 
 ```
-Repair-WinGetPackageManager [-AllUsers] [-Force] [-Version <String>] [<CommonParameters>]
+Repair-WinGetPackageManager [-AllUsers] [-Force] [-Version <String>] [<CommonParameters>] [-IncludePreRelease]
```
Trenly: nit: Common Parameters is generally kept as the last item in the usage string

src/PowerShell/Help/Microsoft.WinGet.Client/Repair-WinGetPackageManager.md
```diff
@@ -54,6 +54,16 @@ This example shows how to repair they WinGet client by installing the latest ver
 it functions properly. The **Force** parameter shuts down the version that is currently running so
 that it can update the application files.
 
+### Example 3: Install a version with wildcards
+
+```powershell
+Repair-WinGetPackageManager -Version "1.*.1*" -Force
```
Trenly: Although this may be supported in code, for user facing documentation I'd suggest keeping it simple - e.g `1.12.*`

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
```diff
@@ -88,18 +89,113 @@ public void RepairUsingLatest(bool preRelease, bool allUsers, bool force)
...
+        /// <param name="includePrerelease">Include prerelease versions.</param>
+        /// <param name="result">The resulting version.</param>
+        /// <returns>True if a matching version was found.</returns>
+        private static bool TryGetLatestMatchingVersion(IEnumerable<WinGetVersion> versions, string pattern, bool includePrerelease, out WinGetVersion result)
+        {
+            pattern = string.IsNullOrWhiteSpace(pattern) ? "*" : pattern;
+
+            var parts = pattern.Split('.');
+            string major = parts[0];
+            string minor = parts.Length > 1 ? parts[1] : string.Empty;
```
Trenly: Since you’re including linq anyways, is there a reason not to use [ElementAtOrDefault()](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.elementatordefault?view=net-9.0) here?

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
```diff
@@ -88,18 +89,113 @@ public void RepairUsingLatest(bool preRelease, bool allUsers, bool force)
...
+                versions = versions.Where(v => !v.IsPrerelease);
+            }
+
+            versions = versions
+                .Where(v =>
+                    VersionPartMatch(major, v.Version.Major) &&
+                    VersionPartMatch(minor, v.Version.Minor) &&
+                    VersionPartMatch(build, v.Version.Build) &&
+                    VersionPartMatch(revision, v.Version.Revision))
+                .OrderBy(f => f.Version);
```
Trenly: `.MaxBy(f => f.Version);` would avoid having to sort the enumerable and should be safe even if the enumerable is empty

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
```diff
@@ -88,18 +89,113 @@ public void RepairUsingLatest(bool preRelease, bool allUsers, bool force)
...
+        }
+
+        private static bool VersionPartMatch(string partPattern, int partValue)
+        {
+            if (string.IsNullOrWhiteSpace(partPattern))
+            {
+                return true;
+            }
+
+            if (partPattern.EndsWith("*"))
```
Trenly: I know you aren’t claiming to support it in this PR, but it seems like it would be trivial to add the case for 
```
if (partPattern.StartsWith("*"))
{
    return partValue.ToString().EndsWith(partPattern.TrimStart('*'));
}
```

?

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/AppxModuleHelper.cs
```diff
@@ -94,6 +94,11 @@ internal class AppxModuleHelper
         private const string XamlPackage27 = "Microsoft.UI.Xaml.2.7";
         private const string XamlReleaseTag273 = "v2.7.3";
 
+        // WinGet Source
+        private const string WinGetSourceName = "Microsoft.Winget.Source";
+        private const string WinGetSourceMsixName = "source2.msix";
+        private const string WinGetSourceUrl = "https://cdn.winget.microsoft.com/cache/source2.msix";
```
Trenly: Is there a COM API for fetching source information? Since the winget source is a well-known source (can be disabled, but not removed), would it be a possible consideration to dynamically fetch the source URL through the API? That way, in the event a source3.msix ever becomes the default, this portion of the cmdlet would just work?

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
AmelBawa-msft: Thank you for the suggestion. Unfortunately, `MaxBy` is not supported in net48. And `Max` returns the property (similar to `Select`) instead of the item itself.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
AmelBawa-msft: I wasn’t sure initially if that would be a relevant use case, so I hesitated to go down that path and start supporting more edge cases like `*1*, 1*1, 1*1*1`, and so on. That said, if there's demand for those patterns, I’ll likely add support for them too 😉

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/AppxModuleHelper.cs
AmelBawa-msft: I was able to find the CDN URL `https://cdn.winget.microsoft.com/cache` but not the `source2.msix` file as part of the COM API. @JohnMcPMS, is there a recommended/forward compatible method to obtain the `source2.msix` URL instead of hardcoding it in the powershell module? Or specifying it here is okay?

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
Trenly: net48 compatibilty strikes again :(

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
```diff
@@ -88,18 +89,118 @@ public void RepairUsingLatest(bool preRelease, bool allUsers, bool force)
...
+
+            if (partPattern!.EndsWith("*"))
+            {
+                return partValue.ToString().StartsWith(partPattern.TrimEnd('*'));
+            }
+
+            return partPattern == partValue.ToString();
+        }
+
+        private async Task<string> ResolveVersionAsync(string version, bool includePrerelease)
```
JohnMcPMS: This function and its helpers should be members of `Microsoft.WinGet.Client.Engine.Helpers.GitHubClient` like `GetLatestReleaseTagNameAsync` is.

Also, it should have a heuristic looking for wildcards *before* invoking a bunch of web calls and shortcut return the input version if no wildcard is present.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
```diff
@@ -88,18 +89,118 @@ public void RepairUsingLatest(bool preRelease, bool allUsers, bool force)
...
+                versions = versions.Where(v => !v.IsPrerelease);
+            }
+
+            versions = versions
+                .Where(v =>
+                    VersionPartMatch(major, v.Version.Major) &&
+                    VersionPartMatch(minor, v.Version.Minor) &&
+                    VersionPartMatch(build, v.Version.Build) &&
+                    VersionPartMatch(revision, v.Version.Revision))
+                .OrderBy(f => f.Version);
```
JohnMcPMS: In the current `GetLatestReleaseAsync`, we assume that the releases are returned in a sorted order (latest firstest).  If that is true and maintained by all of the LINQ queries, then a `First` should be sufficient and optimal.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
```diff
@@ -88,18 +89,118 @@ public void RepairUsingLatest(bool preRelease, bool allUsers, bool force)
...
+        }
+
+        private static bool VersionPartMatch(string? partPattern, int partValue)
+        {
+            if (string.IsNullOrWhiteSpace(partPattern))
+            {
+                return true;
+            }
+
+            if (partPattern!.StartsWith("*"))
```
JohnMcPMS: The `!.` shouldn't be necessary...

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/AppxModuleHelper.cs
JohnMcPMS: It is an internal implementation detail, so there isn't a way to get the file name.  I don't see why we need to specify it here twice though.

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/AppxModuleHelper.cs
```diff
@@ -112,21 +117,33 @@ public AppxModuleHelper(PowerShellCmdlet pwshCmdlet)
...
+        /// <summary>
+        /// Calls Get-AppxPackage Microsoft.Winget.Source.
+        /// </summary>
+        /// <param name="allUsers">Whether to get for all users.</param>
         /// <returns>Result of Get-AppxPackage.</returns>
-        public PSObject? GetAppInstallerObject()
+        public PSObject? GetWinGetSourceObject(bool allUsers = false)
         {
-            return this.GetAppxObject(AppInstallerName);
+            return this.GetAppxObject(WinGetSourceName, allUsers);
```
JohnMcPMS: I don't think that we want/expect/should have the source installed for all users.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
AmelBawa-msft: I don’t think “latest release” necessarily means “latest version,” since you could technically release version 1.0.3 before 1.2.9. Releases are probably sorted by their creation date, which makes sense for identifying the latest release. But for this algorithm, it’s key to make sure the version list is sorted semantically, not just by release date.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
AmelBawa-msft: Even though I'm checking for null in the line above, the compiler still throws a CS8602 warning about a possible null reference. I guess it doesn't recognize that the method I used guarantees the variable isn't null.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
AmelBawa-msft: Good call, I’ve moved that logic into the helper class.
We’re only making a single `GetAll()` web request to resolve the version, which I think is sufficient.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
JohnMcPMS: Then we should update `GetLatestReleaseAsync` to match this understanding of semantics.  It is definitely not intended to be latest in terms of calendar, but latest in terms of the highest version.  Fortunately, these are the same sort order for release only listing.  When pre-release is involved, they could become interleaved, as the latest calendar date might be a release with a lower version number than a pre-release from before.

[The documentation for listing releases](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases) does not mention *ANY* ordering, so even assuming calendar is faulty.

I would suggest making `GetAllReleasesAsync` return a collection of some type containing both a `WinGetVersion` and the `Release`, sorted so that the highest version is first.  This makes the existing "latest" function more correct and your new function more efficient as it can simply get the `First` object matching the predicate rather than examining all of them.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
JohnMcPMS: I fully disagree, don't make any web request when we don't need one.

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/AppxModuleHelper.cs
JohnMcPMS: Can we not have the file name twice?  Use string concatenation or formatting or whatever to only have `source2.msix` the one time.

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/GitHubClient.cs
```diff
@@ -72,5 +77,104 @@ public async Task<Release> GetLatestReleaseAsync(bool includePrerelease)
...
+        /// Pattern only supports leading and trailing wildcards.
+        /// - For example, the pattern can be: 1.11.*, 1.11.3*, 1.11.*3
+        /// - But it cannot be: 1.*1*.1 or 1.1*1.1.
+        /// </remarks>
+        /// <param name="versions">List of versions to match against.</param>
+        /// <param name="pattern">Pattern to match.</param>
+        /// <param name="includePrerelease">Include prerelease versions.</param>
+        /// <param name="result">The resulting version.</param>
+        /// <returns>True if a matching version was found.</returns>
+        private static bool TryGetLatestMatchingVersion(IEnumerable<WinGetVersion> versions, string pattern, bool includePrerelease, out WinGetVersion result)
```
JohnMcPMS:
```suggestion
        private static bool TryGetLatestMatchingVersion(IEnumerable<WinGetVersion> versions, string pattern, bool includePrerelease, out WinGetVersion? result)
```
Because you very clearly return a null in some cases.  You can annotate with the appropriate attribute to mark it as not null when returning true.

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
JohnMcPMS: I think if you moved functionality for the matching to `WinGetVersion`, you could easily have something like:
```
if (!WinGetVersion.VersionHasWildcard(version))
{
    return version;
}
```

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/WinGetPackageManagerCommand.cs
Trenly: This apparently is a bug right now, actually, since the latest Pre-Release is `1.12.240-preview`, but using `Repair-WinGetPackageManager -IncludePrerelease -Latest` selects `1.11.510`

Would be really nice to have this fixed

Infinitay: I just faced the same issue tonight. I believe my winget was previously on 1.11.x but when running `Repair-WinGetPackageManager -Force -IncludePrerelease -Verbose` it wasn't getting the pre-release version. After stumbling upon what you wrote, I manually specified the latest pre-release version and passed it in alongside `-Version` and it finally worked.

AmelBawa-msft: Updated😊

src/PowerShell/Microsoft.WinGet.Client.Engine/Helpers/GitHubClient.cs
JohnMcPMS: My intention was that you would use the [`NotNullWhen`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=net-10.0) attribute and not need the `!` inside the conditional.

JohnMcPMS: You could also remove the `!` from the `null!` in the function now.

AmelBawa-msft: > My intention was that you would use the [NotNullWhen](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=net-10.0) attribute and not need the ! inside the conditional.

Unfortunately, the attribute is not supported in net48 desktop framework. It gives compiler errors when attempting.

Reference:
- https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=netstandard-2.1

PR #5719, opened 2025-09-13
Fix two unrelated version issues

### LINKED ISSUE #4928: Index V2 does not properly handle versions of `<part>.0` or `<part>.###0`
When moving from IndexV1 to IndexV2 the code which builds up the index was heavily modified to optimize the size of the index. This has resulted in a parsing error, where versions such as `128.0` or `132.0` do not have the trailing `.0` mapped into their version. When the version is more than two parts, or does not end in `0`, the version is parsed properly. This only affects IndexV2.

This seems to affect `latest_version`, `arp_min_version`, and `arp_max_version`

For example -
* `128.0` -> `128`
* `11.10` -> `11.1`

Image
Image

My best guess is that the version information isn't being handled explicitly as a string. When there is only a single decimal point, it then gets parsed as a number

JohnMcPMS:
Fixes #4928
Fixes providing "1" for the version when the version is actually "1.0" and the like (there is probably an issue filed but I didn't find it after a quick search).

## Issues
Due to SQLite helpfully not being picky about data types, the packages table in the v2 index was using INT64 for every column, when in reality they should mostly by strings (TEXT).  When the version was parsable as a number, it would drop unnecessary portions (trailing 0s).

New code in the v2 index and the composite package changes were using string comparison instead of `Version`, requiring an exact string match rather than `Version` equality.

## Changes
Use the data type that we already had set up rather than always INT64.  This will require a service update to actually fix anything (but doesn't require a client update).

Use `Version` comparisons rather than string in the v2 index and composite package version selection.

## Validation
Added new tests and confirmed via manual usage.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5719)

### DIFF
--- a/src/AppInstallerCLITests/SQLiteIndex.cpp
+++ b/src/AppInstallerCLITests/SQLiteIndex.cpp
@@ -3935,3 +3935,31 @@ TEST_CASE("SQLiteIndex_AddOrUpdateManifest", "[sqliteindex]")
         REQUIRE(!index.AddOrUpdateManifest(manifest, manifestPath));
     }
 }
+
+TEST_CASE("SQLiteIndex_VersionStringPreserved", "[sqliteindex]")
+{
+    TempFile tempFile{ "repolibtest_tempdb"s, ".db"s };
+    INFO("Using temporary file named: " << tempFile.GetPath());
+
+    Manifest manifest;
+    SQLiteIndex index = CreateTestIndex(tempFile);
+
+    string_t publisher = "Test";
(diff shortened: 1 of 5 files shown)


Trenly:
> Due to SQLite helpfully not being picky about data types

Should there be an issue filed for the SQLiteBuilder to support optionally adding `STRICT` onto the table creation SQLstatement ?

florelis (approved):

PR #5724, opened 2025-09-15
Heal tracking database if it can't open

JohnMcPMS:
## Change
When we fail to open the tracking database, rename it and create a new one.  The initial `Open` requires two things to succeed:
1. SQLite considers this a valid database file
2. The metadata table exists and contains a schema version that we can read

If an exception is thrown, we move the file (and the auxiliary journals if they exist) to new names (`installed.db` -> `installed-corrupted.db`) and create a new database.

## Validation
Added a unit test that corrupts the database and is able to open the new one.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5724)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -582,6 +582,7 @@ vns
 vsconfig
 vstest
 waitable
+wal
 wcex
 WDAG
 webpages


src/AppInstallerRepositoryCore/PackageTrackingCatalog.cpp
```diff
@@ -31,19 +32,39 @@ namespace AppInstaller::Repository
...
+                }
+                catch(...)
+                {
+                    LOG_CAUGHT_EXCEPTION_MSG("Exception opening tracking catalog");
+                }
+
+                // Move existing database and create a new one
+                std::filesystem::path destination{ trackingDB };
+                destination.replace_filename(c_PackageTrackingCorruptedFileName);
+                SQLite::SQLiteStorageBase::RenameSQLiteDatabase(trackingDB, destination, true);
```
dkbennett, line 65: Is there a goal to eventually try to fix or recover some data from corrupted databases?

dkbennett (approved):

src/AppInstallerRepositoryCore/PackageTrackingCatalog.cpp
JohnMcPMS: Certainly not programmatically, as the issue here is that SQLite could not open the file as a database (making it quite difficult to operate on) or that we are missing the table that contains the schema version.  The goal of the rename is to allow for human investigation if needed rather than removing any trace of the error.

PR #5726, opened 2025-09-17
Cache information responses from REST sources

JohnMcPMS:
## Change
Store and use the `/information` response from REST sources per their `cache-control` header.

At the base of this change is a new type of setting stream: `Encrypted`.  These settings are encrypted to the user and machine that the data is stored on.

Based on the log file from running with and without a cached value, the time between close log lines dropped from 189 ms to 8 ms for the msstore source.  This time saving should be in effect for the large majority of uses of WinGet.

## Validation
Added tests across various levels of the change.
Manually verified functionality of CLI.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5726)

### DIFF
--- a/src/AppInstallerCommonCore/Runtime.cpp
+++ b/src/AppInstallerCommonCore/Runtime.cpp
@@ -288,6 +288,11 @@ namespace AppInstaller::Runtime
             result.Path.assign(appStorage.LocalFolder().Path().c_str());
             mayBeInProfilePath = true;
             break;
+        case PathName::PackagedTemp:
+            result.Path.assign(appStorage.TemporaryFolder().Path().c_str());
+            result.Path /= s_DefaultTempDirectory;
+            mayBeInProfilePath = true;
+            break;
         case PathName::DefaultLogLocation:
             // To enable UIF collection through Feedback hub, we must put our logs here.
             result.Path.assign(appStorage.LocalFolder().Path().c_str());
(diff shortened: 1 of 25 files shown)


src/AppInstallerSharedLib/JsonUtil.cpp
```diff
@@ -263,4 +314,14 @@ namespace AppInstaller::JSON
 
         return true;
     }
+
+    bool IsValidNonEmptyStringValue(std::optional<std::wstring>& value)
+    {
+        if (Utility::IsEmptyOrWhitespace(value.value_or(L"")))
```
yao-msft, line 320: nit: return !Utility::IsEmptyOrWhitespace...

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -22,6 +22,7 @@ namespace AppInstaller::Runtime
     {
         using namespace std::string_view_literals;
         constexpr std::string_view s_DefaultTempDirectory = "WinGet"sv;
+        constexpr std::string_view s_SettingsFile_Relative = "Settings"sv;
```
yao-msft, line 25: nit: The s_SecureSettings_UserRelative below uses "settings", will that potentially cause conflicts in the future?

yao-msft (approved):

Issue #1885, opened 2022-01-31
Manifest fields for custom uninstall arguments and return codes


jedieaston:
### Description of the new feature / enhancement

We currently have a bunch of manifest fields for installing packages:
```
InstallerSwitches:
    Custom: /allthethings
    Silent:  /VERYSILENT
    SilentWithProgress: /somewhatsilent
InstallerSuccessCodes:
    - 12
    - 57
    - 99
 ```

But we don't have any for uninstalling packages. Many packages return funny error codes that still mean success after an uninstall, or require special arguments to make uninstall 100% silent. Currently, there isn't a way to tell winget to use these.

(I believe this will help with the Edge WebView2 issues, as it requires user interaction to uninstall and then it exits with a non-zero code, which halts the upgrade leaving the user with no WebView2).

### Proposed technical implementation details

Some matching manifest keys that work similarly to the existing installer fields, per installer (I realize that it may be difficult to make a best guess at `InstallerType` if winget didn't install the package to begin with, but that will need to be implemented anyway at some point). Ideally, these switches should be ran on the local uninstall executable, presuming it still exists. 

If a manifest isn't available for an app that a user wants to uninstall, then the current behavior should be preserved. 

```
UninstallerSwitches:
    Custom: /uninstallallthethings
    Silent:  /VERYSILENT
    SilentWithProgress: /somewhatsilent
UninstallerSuccessCodes:
    - 12
    - 57
    - 99
```

denelon:
Additional context from a thread on [Twitter](https://twitter.com/JeroenEbus/status/1620356651760091136).

I asked if we had samples of uninstall experiences that could be improved with arguments for silent:
* `winget uninstall citrix.workspace` is showing the prompt, CitrixWorkspaceApp22.12.0.48.exe /uninstall /silent silently removes the Citrix Workspace App without a prompt.
* Another one: "C:\Program Files\IrfanView\iv_uninstall.exe" /silent

koitsu:
Another one to add to the list as an example of where uninstaller exit codes are important: `winget uninstall mirc`.  This one is amusing in itself, as it uses the NSIS (Nullsoft installer).  The following occurs:

* Uninstaller is run
  * Note that the uninstaller binary, `uninstall.exe`, is located in the same directory as the mIRC software itself, i.e. `C:\Program Files\mIRC`
  * Also note that the uninstaller is not run silently (i.e. is interactive), despite the installer being run silently
[...]

PR #5734, opened 2025-09-22
Add Uninstaller Switches and Success Codes to Manifest Schema

dkbennett:
Motivated by issue #1885 but does not complete it.

This adds the UninstallerSwitches and UninstallerSuccessCodes to the manifest schema for 1.12 and all related tests and validation.

* UninstallerSwitches are a subset of the InstallerSwitches
* UninstallerSuccessCodes are essentially identical to InstallerSuccessCodes

Neither of these fields are consumed yet, this is updating the schema to support their addition.

Tested:
* RestSource tests and ManifestValidation tests pass

-----

### DIFF
--- a/schemas/JSON/manifests/latest/manifest.installer.latest.json
+++ b/schemas/JSON/manifests/latest/manifest.installer.latest.json
@@ -2,778 +2,851 @@
   "$id": "https://aka.ms/winget-manifest.installer.1.12.0.schema.json",
   "$schema": "http://json-schema.org/draft-07/schema#",
   "description": "A representation of a single-file manifest representing an app installers in the OWC. v1.12.0",
-  "definitions": {
-    "PackageIdentifier": {
-      "type": "string",
-      "pattern": "^[^\\.\\s\\\\/:\\*\\?\"<>\\|\\x01-\\x1f]{1,32}(\\.[^\\.\\s\\\\/:\\*\\?\"<>\\|\\x01-\\x1f]{1,32}){1,7}$",
-      "maxLength": 128,
-      "description": "The package unique identifier"
-    },
-    "PackageVersion": {
-      "type": "string",
-      "pattern": "^[^\\\\/:\\*\\?\"<>\\|\\x01-\\x1f]+$",
-      "maxLength": 128,
-      "description": "The package version"
-    },
-    "Locale": {
-      "type": [ "string", "null" ],
-      "pattern": "^([a-zA-Z]{2,3}|[iI]-[a-zA-Z]+|[xX]-[a-zA-Z]{1,8})(-[a-zA-Z]{1,8})*$",
-      "maxLength": 20,
-      "description": "The installer meta-data locale"
-    },
-    "Channel": {
-      "type": [ "string", "null" ],
-      "minLength": 1,
-      "maxLength": 16,
-      "description": "The distribution channel"
-    },
-    "Platform": {
-      "type": [ "array", "null" ],
-      "items": {
-        "title": "Platform",
-        "type": "string",
-        "enum": [
-          "Windows.Desktop",
-          "Windows.Universal"
-        ]
-      },
-      "maxItems": 2,
-      "uniqueItems": true,
-      "description": "The installer supported operating system"
-    },
-    "MinimumOSVersion": {
-      "type": [ "string", "null" ],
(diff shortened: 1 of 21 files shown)


schemas/JSON/manifests/latest/manifest.installer.latest.json
```diff
@@ -2,778 +2,838 @@
...
-    },
-    "NestedInstallerFiles": {
-      "type": [ "array", "null" ],
-      "items": {
-        "type": "object",
-        "title": "NestedInstallerFile",
-        "properties": {
-          "RelativeFilePath": {
+    "definitions": {
+        "PackageIdentifier": {
```
JohnMcPMS: Was this churn intentional?  The PR comment did not mention a refactoring of the schema like this, and I don't understand why we are doing it.

schemas/JSON/manifests/latest/manifest.singleton.latest.json
```diff
@@ -2,879 +2,939 @@
...
-    },
-    "Icon": {
-      "type": "object",
-      "properties": {
-        "IconUrl": {
-          "type": "string",
-          "pattern": "^([Hh][Tt][Tt][Pp][Ss]?)://.+$",
-          "maxLength": 2048,
-          "description": "The url of the hosted icon file"
+    "definitions": {
```
JohnMcPMS: Same.

src/AppInstallerRepositoryCore/Rest/Schema/1_12/Json/ManifestDeserializer_1_12.cpp
```diff
@@ -9,6 +9,81 @@ using namespace AppInstaller::Manifest;
 
 namespace AppInstaller::Repository::Rest::Schema::V1_12::Json
 {
+    namespace
+    {
+        // Uninstaller switches
+        constexpr std::string_view UninstallerSwitches = "UninstallerSwitches"sv;
```
JohnMcPMS, line 15: Can we not reuse the installer parsing?

JohnMcPMS (changes requested):
I feel like adding all of these to the schema without an implementation and/or plan on how to consume them will result in values that are unusable.  Silent already has a meaning since there is silent ARP entry data.  What do we do if a silent switch is present with that ARP data?  Interactive is the default assumption for uninstall ARP command line; when would we use the interactive switch?

If they exist in 1.12 schema but aren't implemented, people can add them without effect until they suddenly light up at some point in the future.  This is really the blocking issue to me.  I can imagine a path forward, even if it isn't clearly defined today, for all of the switches.  But allowing for these to be defined in manifests in winget-pkgs without any validation beyond "the manifest parses" seems very bad.

Trenly:
> Interactive is the default assumption for uninstall ARP command line;

I thought the default was:
* If the ARP contains `SilentUninstallString` use that
* Otherwise, fall back to `UninstallString`

schemas/JSON/manifests/latest/manifest.installer.latest.json
Trenly: This big of a refactoring of the schema could also break tools (such as YamlCreate) that expect the schema definitions to be mostly the same between versions, with only content being added to the existing fields, because some of these tools pull the validation requirements for pattern and length directly from the schema

JohnMcPMS:
> > Interactive is the default assumption for uninstall ARP command line;
> 
> I thought the default was:
> 
> * If the ARP contains `SilentUninstallString` use that
> * Otherwise, fall back to `UninstallString`

I wasn't complete enough in my statement: Interactive is the default assumption for the UninstallString value behavior.

schemas/JSON/manifests/latest/manifest.installer.latest.json
dkbennett: Not intentional, will fix

schemas/JSON/manifests/latest/manifest.singleton.latest.json
dkbennett: Yeah it was the VS editor having different spacing and I didn't catch it on the checkin, fixing. Not an intentional change, just VS really wanting a 4 space indentation. Something really unexpected happened here.

src/AppInstallerRepositoryCore/Rest/Schema/1_12/Json/ManifestDeserializer_1_12.cpp
dkbennett: They don't have the same set of switches - upgrade, update, and install location doesn't make sense for uninstall switches,

PR #5749, opened 2025-09-26
Shared build props

JohnMcPMS:
## Change
Due to an issue with the internal build, it was realized that we could be violating ODR due to a mismatch on the build configurations.  This change moves to using the MSBuild `Directory.Build.props` file to have all projects use the same settings.  It also adds a header using `#pragma detect_mismatch` to attempt to detect the issue at compilation.  This header is then included by the errors header, which is used by at least one TU in every project, if not all of them.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -29,15 +29,27 @@
     </ClCompile>
   </ItemDefinitionGroup>
 
+  <PropertyGroup Condition="'$(WingetDisableExperimentalFeatures)'=='true'">
+    <DefineConstants>$(DefineConstants);WINGET_DISABLE_EXPERIMENTAL_FEATURES</DefineConstants>
+  </PropertyGroup>
+
   <ItemDefinitionGroup Condition="'$(UseProdWingetServer)'=='true'">
     <ClCompile>
       <PreprocessorDefinitions>USE_PROD_WINGET_SERVER;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
   </ItemDefinitionGroup>
+
+  <PropertyGroup Condition="'$(UseProdWingetServer)'=='true'">
+    <DefineConstants>$(DefineConstants);USE_PROD_WINGET_SERVER</DefineConstants>
+  </PropertyGroup>
   
   <ItemDefinitionGroup Condition="'$(WingetEnableReleaseBuild)'=='true'">
     <ClCompile>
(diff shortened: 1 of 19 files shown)


src/Directory.Build.props
```diff
@@ -0,0 +1,55 @@
...
+  
+  <ItemDefinitionGroup Condition="'$(WingetEnableReleaseBuild)'=='true'">
+    <ClCompile>
+      <PreprocessorDefinitions>WINGET_ENABLE_RELEASE_BUILD;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+
+  <PropertyGroup Condition="'$(WingetEnableReleaseBuild)'=='true'">
+    <DefineConstants>$(DefineConstants);WINGET_ENABLE_RELEASE_BUILD</DefineConstants>
+  </PropertyGroup>
```
florelis, line 54: I think we could make this easier to read. Maybe something like

```suggestion
  <PropertyGroup>
    <WinGetMacros Condition=" '$(UseProdCLSIDs)' == 'true' ">USE_PROD_CLSIDS;$(WinGetMacros)</WinGetMacros>
    <WinGetMacros Condition=" '$(WingetDisableTestHooks)' == 'true' ">AICLI_DISABLE_TEST_HOOKS;$(WinGetMacros)</WinGetMacros>
    <WinGetMacros Condition=" '$(WingetDisableExperimentalFeatures)' == 'true' ">WINGET_DISABLE_EXPERIMENTAL_FEATURES;$(WinGetMacros)</WinGetMacros>
    <WinGetMacros Condition=" '$(UseProdWingetServer)' == 'true' ">USE_PROD_WINGET_SERVER;$(WinGetMacros)</WinGetMacros>
    <WinGetMacros Condition=" '$(WingetEnableReleaseBuild)' == 'true' ">WINGET_ENABLE_RELEASE_BUILD;$(WinGetMacros)</WinGetMacros>
  </PropertyGroup>

  <!-- For C++ -->
  <ItemDefinitionGroup>
    <ClCompile>
      <PreprocessorDefinitions Condition="'$(WinGetMacros)' != ''">$(WinGetMacros)%(PreprocessorDefinitions)</PreprocessorDefinitions>
    </ClCompile>
  </ItemDefinitionGroup>

  <!-- For C# -->
  <PropertyGroup>
    <DefineConstants Condition=" '$(WinGetMacros)' != '' ">$(WinGetMacros)$(DefineConstants)</DefineConstants>
  </PropertyGroup>
```

JohnMcPMS: Because it was a non-functionality comment, I merged this so it would for sure be in for next week and made your change here: #5751

PR #5775, opened 2025-10-07
Remove openssl from sfsclient cgmanifest

florelis:
Component Governance is reporting issues from openssl. The alert is caused by it being in the `cgmanifest.json` of `sfs-client` even though [they don't use it on Windows](https://github.com/microsoft/sfs-client/blob/63aab71b46d92977a60b231849bebcf6cbd37715/vcpkg.json#L21). Since it is not actually used, we can't do something like overriding the version (and even if we did, CG would also pick the version in the json).

Removing it from their cgmanifest should get rid of the alert, although it will cause issues the next time we update the subtree.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5775)

### DIFF
--- a/src/SfsClient/sfs-client/cgmanifest.json
+++ b/src/SfsClient/sfs-client/cgmanifest.json
@@ -65,25 +65,6 @@
       "component": {
         "type": "git",
         "git": {
-          "repositoryUrl": "https://github.com/openssl/openssl",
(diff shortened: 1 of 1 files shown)


florelis:
@microsoft-github-policy-service rerun

PR #5781, opened 2025-10-09
Improve window thread termination

JohnMcPMS:
## Change
Shut down the window thread as part of the `TerminationSignalHandler` destruction.
Properly unregister the window class after destroying the window on the thread.
Signal the shutdown via an `atomic_bool`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5781)

### DIFF
--- a/src/AppInstallerCLICore/ShutdownMonitoring.cpp
+++ b/src/AppInstallerCLICore/ShutdownMonitoring.cpp
@@ -104,8 +104,11 @@ namespace AppInstaller::ShutdownMonitoring
         // std::thread requires that any managed thread (joinable) be joined or detached before destructing
         if (m_windowThread.joinable())
         {
-            // Inform the thread that it should stop.
-            PostMessageW(m_windowHandle.get(), WM_DESTROY, 0, 0);
+            if (m_windowHandle)
+            {
+                // Inform the thread that it should stop.
+                PostMessageW(m_windowHandle.get(), WM_DESTROY, 0, 0);
+            }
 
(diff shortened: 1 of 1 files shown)


src/AppInstallerCLICore/ShutdownMonitoring.cpp
```diff
@@ -227,30 +238,36 @@ namespace AppInstaller::ShutdownMonitoring
...
         // Force message queue to be created.
         MSG msg;
         PeekMessage(&msg, NULL, WM_USER, WM_USER, PM_NOREMOVE);
         m_messageQueueReady.SetEvent();
 
-        // Message loop
-        BOOL getMessageResult;
-        while ((getMessageResult = GetMessage(&msg, m_windowHandle.get(), 0, 0)) != 0)
+        // Message loop, exits when the window handle has been destroyed
+        while (m_windowThreadShouldRun.load())
```
yao-msft: nit: do we need the load()

src/AppInstallerCLICore/ShutdownMonitoring.cpp
```diff
@@ -7,6 +7,8 @@
 #include <AppInstallerRuntime.h>
 #include <winget/COMStaticStorage.h>
 
+using namespace std::chrono_literals;
```
yao-msft, line 10: nit: not needed anymore?

PR #5787, opened 2025-10-09
Bump version to 1.28

florelis:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5787)


JohnMcPMS:
I feel like this should also fix at least the one known location of "we know these versions are 15 apart".

PR #5790, opened 2025-10-09
Allow unsigned source for development testing

dkbennett:
This formalizes a workaround I have been using for testing an unsigned Pre-Indexed Package Source. It would be convenient to have this in the code so I don't have to constantly stash/unstash the changes when moving around branches.

This works only when test hooks are enabled (i.e. a debug build) AND you set the preprocessor directive "AICLI_ALLOW_UNSIGNED_SOURCE"

The easy way to set this up is to add CL=/DAICLI_ALLOW_UNSIGNED_SOURCE to your Environment variables and restart visual studio.

I've noticed Visual Studio will not recognize the directive is set, but the compiler will set it when it builds and it will work. I did not make these const due to a web of compiler warnings and errors when a const value is specified first in an if statement.

- [] This pull request is related to an issue.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5790)

### DIFF
--- a/src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
@@ -20,7 +20,7 @@ namespace AppInstaller::Repository::Microsoft
 {
     namespace
     {
-        // To use an unsigned source, set AICLI_ALLOW_UNSTIGNED_SOURCE and use a debug build.
+        // To use an unsigned source, set AICLI_ALLOW_UNSIGNED_SOURCE and use a debug build.
         // Ex: set CL=/DAICLI_ALLOW_UNSIGNED_SOURCE
 #if ! defined( AICLI_DISABLE_TEST_HOOKS ) && defined( AICLI_ALLOW_UNSIGNED_SOURCE )
         static bool s_AllowUnsignedSource = true;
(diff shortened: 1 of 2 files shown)


src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -20,6 +20,14 @@ namespace AppInstaller::Repository::Microsoft
 {
     namespace
     {
+        // To use an unsigned source, set AICLI_ALLOW_UNSIGNED_SOURCE and use a debug build.
+        // Ex: set CL=/DAICLI_ALLOW_UNSIGNED_SOURCE
+#if ! defined( AICLI_DISABLE_TEST_HOOKS ) && defined( AICLI_ALLOW_UNSIGNED_SOURCE )
```
JohnMcPMS, line 25: I think I would rather see this hardened more to prevent what feels like potential for accidental disabling of the security.

I would do that by creating a function on the base class for use in calling `ValidateTrustInfo`.  Then create a derived type from the Desktop version that overrides that virtual with the one that doesn't check (and can do that easily from preprocessor instead of a global bool).

I would also create a separate semantic for using the Desktop version all the time, mostly because we might consider allowing that as choice.

Issue #5093, opened 2024-12-21
Cannot export pkg list to hidden file


mabequinho:
### Brief description of your issue

Cannot export to hidden file

### Steps to reproduce

winget export -o .\wingetpkgs.json 

where `wingetpkgs.json` has 'hidden attribute'

### Expected behavior

export the package list.

### Actual behavior

file is not modified but it does not return any error.

### Environment

```shell
❯ winget --info
Windows Package Manager v1.9.25200
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.26100.2605
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.24.25200.0
```

Trenly:
* Resolved by #5795

<!--
Close with reason: Resolved;

PR #5795, opened 2025-10-10
Allow exporting to hidden files

Trenly:
  - #5093
-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5795)


src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -157,8 +157,25 @@ namespace AppInstaller::CLI::Workflow
         auto packages = PackagesJson::CreateJson(context.Get<Execution::Data::PackageCollection>());
 
         std::filesystem::path outputFilePath{ context.Args.GetArg(Execution::Args::Type::OutputFile) };
+
+        // Check if the file exists and is hidden
+        DWORD attrs = std::filesystem::exists(outputFilePath) ? GetFileAttributesW(outputFilePath.c_str()) : INVALID_FILE_ATTRIBUTES;
```
JohnMcPMS: The `exists` check shouldn't be necessary; the return value from `GetFileAttributesW` should be `INVALID_FILE_ATTRIBUTES` for a non-existent file as well.

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -157,8 +157,25 @@ namespace AppInstaller::CLI::Workflow
...
         std::filesystem::path outputFilePath{ context.Args.GetArg(Execution::Args::Type::OutputFile) };
+
+        // Check if the file exists and is hidden
+        DWORD attrs = std::filesystem::exists(outputFilePath) ? GetFileAttributesW(outputFilePath.c_str()) : INVALID_FILE_ATTRIBUTES;
+        bool isHidden = (attrs != INVALID_FILE_ATTRIBUTES && (attrs & FILE_ATTRIBUTE_HIDDEN));
+
+        if (isHidden)
+        {
+            // Remove hidden attribute so we can write to it
+            SetFileAttributesW(outputFilePath.c_str(), attrs & ~FILE_ATTRIBUTE_HIDDEN);
```
JohnMcPMS: I would really like it if there were a way to simply open the hidden file rather than changing the attribute.  There might be a set of mode options that change the `CreateFile` call to avoid the documented error state (I would think specifically anything that moves to "open only existing files" might do it.  So more of a `if (exists) open_existing_with_truncate() else create()`.

If that cannot be done, then at a minimum we should use a scope exit to reset the attributes.

Trenly: @johnmcpms - Thank you for the feedback, sorry it took me so long to get to. I was able to get this working without changing the attributes by using WriteFile instead of an output file stream

src/AppInstallerCLICore/Workflows/ImportExportFlow.cpp
```diff
@@ -157,8 +157,22 @@ namespace AppInstaller::CLI::Workflow
...
+        // - For an existing hidden file, use TRUNCATE_EXISTING to clear its content while preserving its attributes.
+        // - Otherwise, use CREATE_ALWAYS to create a new file or overwrite an existing one.
+        DWORD creationDisposition = isHidden ? TRUNCATE_EXISTING : CREATE_ALWAYS;
+        wil::unique_hfile fileHandle{ CreateFileW(outputFilePath.c_str(), GENERIC_WRITE, 0, nullptr, creationDisposition, FILE_ATTRIBUTE_NORMAL, nullptr) };
+        THROW_LAST_ERROR_IF(!fileHandle);
+
+        Json::StreamWriterBuilder writerBuilder;
+        std::string jsonContent = Json::writeString(writerBuilder, packages);
+        DWORD bytesWritten = 0;
+        THROW_LAST_ERROR_IF(!WriteFile(fileHandle.get(), jsonContent.c_str(), static_cast<DWORD>(jsonContent.size()), &bytesWritten, nullptr));
```
JohnMcPMS: Potential partial write is not handled.  Would be better to have a string helper `WriteStringToFile` that does it.

JohnMcPMS (approved):

Issue #2036, opened 2022-03-22
Running into Windows path length limits for installer log files


mthalman:
### Brief description of your issue

This is somewhat related to https://github.com/microsoft/winget-cli/issues/1808 but for installer log files. Because winget sets the log location to an already deep path, the log files that get created by the installer can more easily run into the 260 path length limit of Windows.

In the repro below, the installer attempts to create a log with the path of `C:\Users\WDAGUtilityAccount\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir\WinGet-Microsoft.dotnet.6.1.322.12413-2022-03-22-14-16-07.487_013_Microsoft.NET.Sdk.Android.Manifest_6.0.100.31.0.101_preview.9.16_x64.msi.log` which is over the path length limit.

### Steps to reproduce

1. `git clone https://github.com/microsoft/winget-pkgs.git`
2. `cd winget-pkgs`
3. `git checkout 5e9640f`
5. `.\Tools\SandboxTest.ps1 manifests\m\Microsoft\DotNet\6.1.322.12413`

### Expected behavior

.NET SDK should be successfully installed in Windows Sandbox environment.

### Actual behavior

Install fails.

Output:

```
--> Installing WinGet
Tip: you can type 'Update-EnvironmentVariables' to update your environment variables, such as after installing a new software.

--> Configuring Winget
Admin setting enabled. 

--> Installing the Manifest 6.1.322.12413

Found Microsoft .NET SDK [Microsoft.dotnet] Version 6.1.322.12413
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://dotnetcli.azureedge.net/dotnet/Sdk/6.0.103/dotnet-sdk-6.0.103-win-x64.exe
  ██████████████████████████████   187 MB /  187 MB
Successfully verified installer hash
Starting package install...
Installer failed with exit code: 87
Installer log is available at: C:\Users\WDAGUtilityAccount\AppData\Local\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir\WinGet-Microsoft.dotnet.6.1.322.12413-2022-03-22-14-16-07.487.log

--> Refreshing environment variables

--> Comparing ARP Entries

DisplayName                                       DisplayVersion Publisher             ProductCode
-----------                                       -------------- ---------             -----------
[...]

PR #5802, opened 2025-10-14
Add setting for installer log file names

Trenly:
  - Waiting for bump of version to 1.28
  - Related to #2036

Sometimes the Manifest name is just too long. While I would love to be able to give users the ability to control the format directly using token replacement and all sorts of other cool stuff, that's a bit much for right now.

To help alleviate the issue, I've added a setting that controls how installer file names are generated. With this change, instead of being forced into the Package ID and version with timestamp, it is now possible to have the default installer log name be shorter. While this doesn't actually resolve the issue, it does make it possible to at least mitigate

* Manifest - Default, matches current behavior
* Timestamp - Includes just the timestamp as the installer log name
* GUID - Creates a new GUID and uses that as the installer log name
* ShortGuid - Creates a new GUID and uses the first 8 characters as the installer log name

Note that this only affects the *installer* log name, not the WinGet log name, and only for installers which support passing of the log name. This PR does not affect the behavior of `--log <location>`

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5802)


Trenly:
Marking as draft to fix merge conflicts

Trenly:
Finally got around to fixing the merge conflicts. 

Interesting note, if this gets merged, it will push the repo over 1000 unit tests :D
```
===============================================================================
All tests passed (28169 assertions in 1000 test cases)
```

JohnMcPMS (approved):

PR #5804, opened 2025-10-14
Enable MultiProcessorCompilation

Trenly:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5804)

### DIFF
--- a/src/Xlang/UndockedRegFreeWinRT/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/UndockedRegFreeWinRT.vcxproj
+++ b/src/Xlang/UndockedRegFreeWinRT/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/UndockedRegFreeWinRT.vcxproj
@@ -317,7 +317,7 @@
     <ClCompile Include="catalog.cpp" />
     <ClCompile Include="dllmain.cpp" />
     <ClCompile Include="typeresolution.cpp" />
-    <ClCompile Include="typeresolution.h" />
+    <ClInclude Include="typeresolution.h" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="catalog.h" />
--- a/src/Xlang/UndockedRegFreeWinRT/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/UndockedRegFreeWinRT.vcxproj.filters
+++ b/src/Xlang/UndockedRegFreeWinRT/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/UndockedRegFreeWinRT.vcxproj.filters
@@ -28,9 +28,6 @@
     <ClCompile Include="dllmain.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="typeresolution.h">
-      <Filter>Header Files</Filter>
-    </ClCompile>
     <ClCompile Include="typeresolution.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
@@ -52,6 +49,9 @@
       <Filter>Header Files</Filter>
     </ClInclude>
     <ClInclude Include="..\..\..\..\..\WinGetServer\Utils.h">
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="typeresolution.h">
       <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>


src/AppInstallerCLI/AppInstallerCLI.vcxproj
```diff
@@ -136,6 +136,7 @@
       <PreprocessorDefinitions>_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <WarningLevel>Level4</WarningLevel>
       <AdditionalOptions>%(AdditionalOptions) /permissive- /bigobj</AdditionalOptions>
+      <MultiProcessorCompilation>true</MultiProcessorCompilation>
```
florelis: This could also go in `/src/Directory.Build.props` so that you only need to add it in a single place. (I.e., what I wanted to do back in #3275 but I never finished...)

Trenly: I'm not quite skilled enough to finish that work - though it does look like that refactor would still be a big improvement

JohnMcPMS: I suspect that @florelis was suggesting that you put this property into the Directory.Build.props file that I did wire in to the existing branch already (without all of the other stuff that her change includes).

Trenly: I see. . .

I wasn’t aware that had already been hooked up

florelis (dismissed):

Trenly:
@florelis - Can I request a re-run? I'm not seeing what would have caused the PowerShell error

JohnMcPMS:
@Trenly these seem to have started when we release 1.11.510 and are spurious.  #5807 to start trying to get to the bottom of it.

Trenly:
@johnmcpms - I noticed on [Build #196599](https://dev.azure.com/shine-oss/winget-cli/_build/results?buildId=196599&view=results) there was an ENOSPC - Given that the test seemed to fail on installing PSTools, I'm curious if that step is also occasionally hitting ENOSPC thereby causing the install to fail and thus the test stage to fail, especially since ENOSPC doesn't affect all PRs

See also -
* https://dev.azure.com/shine-oss/winget-cli/_build/results?buildId=194765&view=results
* https://dev.azure.com/shine-oss/winget-cli/_build/results?buildId=191581&view=results

Edit: Created https://github.com/microsoft/winget-cli/issues/5808 for tracking

JohnMcPMS:
This happens so early in the job that I would be surprised if space were an issue.  But regardless it would be good to add collecting the release winget logs in case they also have details on the reason for the error.

src/Directory.Build.props
```diff
@@ -15,6 +15,7 @@
   <ItemDefinitionGroup Condition="'$(WinGetMacros)' != ''">
     <ClCompile>
       <PreprocessorDefinitions>$(WinGetMacros)%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <MultiProcessorCompilation>true</MultiProcessorCompilation>
```
JohnMcPMS, line 22: This should be in an unconditional `ItemDefinitionGroup`; this probably does nothing in the public repo since we don't set any of these values.

Trenly: Thanks; Updated

Trenly:
That failure doesn't make sense to me. . . 

I understand it's a linker error, but I don't see how that error would be appearing from this change - unless the dependencies aren't being built in the right order? But then why would it work locally and not here?

JohnMcPMS:
> That failure doesn't make sense to me. . .
> 
> I understand it's a linker error, but I don't see how that error would be appearing from this change - unless the dependencies aren't being built in the right order? But then why would it work locally and not here?

Were you clean building locally?  I would expect that is the only way to test for sure.

This warning is suspicious, because all of the unresolved methods are in that file:
```
x64\Release\typeresolution.obj : warning LNK4042: object specified more than once; extras ignored
```

It makes me think that the multiprocessor build is processing this file twice and one of them is not so good.

JohnMcPMS:
binlog for the (probable) reason, if not why it doesn't fail without:

typeresolution.h is being added to the ClCompile group because it is "Item Type" `C/C++ compiler`.

In `External - Do Not Modify` > `UndockedRegFreeWinRT` > `Header Files` > `typeresolution.h` > `Properties` > `General` > `Item Type` > should be `C/C++ header`.

Trenly:
> > That failure doesn't make sense to me. . .
> > I understand it's a linker error, but I don't see how that error would be appearing from this change - unless the dependencies aren't being built in the right order? But then why would it work locally and not here?
> 
> Were you clean building locally? I would expect that is the only way to test for sure.
> 
> This warning is suspicious, because all of the unresolved methods are in that file:
> 
> ```
> x64\Release\typeresolution.obj : warning LNK4042: object specified more than once; extras ignored
> ```
> 
> It makes me think that the multiprocessor build is processing this file twice and one of them is not so good.

Yes, I made sure to clean the solution first.

Trenly:
> binlog for the (probable) reason, if not why it doesn't fail without:
> 
> typeresolution.h is being added to the ClCompile group because it is "Item Type" `C/C++ compiler`.
> 
> In `External - Do Not Modify` > `UndockedRegFreeWinRT` > `Header Files` > `typeresolution.h` > `Properties` > `General` > `Item Type` > should be `C/C++ header`.

I could add a condition to specifically exclude the `UndockedRegFreeWinRT` project, perhaps?

Trenly:
> /azp run

Thank you

JohnMcPMS (approved):

florelis (approved):

PR #5807, opened 2025-10-15
Extract event log for potential crash info

JohnMcPMS:
## Change
We are seeing occasional errors during test setup that might be a crash.  These started happening when 1.11.510 became the latest version.  To better understand this and any future crashes that might occur, grab the Application event log from the pipeline.

Also grabs the GA winget logs since that is what is being used during this potential crash.

Also enables the unit tests to write a minidump if they crash, since that happened randomly during one of the builds for this PR.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5807)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -333,7 +333,6 @@ jobs:
       Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force
       Install-Module Microsoft.WinGet.Client -Repository PSGallery -Force
       try { Repair-WingetPackageManager -AllUsers -Latest -Verbose } catch { $_.Exception }
-      '{ "settings": { "logging": { "level": "verbose" } } }' | winget dscv3 user-settings-file --set 2>&1
       Install-WinGetPackage -Id Microsoft.Sysinternals.PsTools -Source winget
     displayName: Install Sysinternals PsTools Using Winget
     condition: succeededOrFailed()


florelis (approved):

azure-pipelines.yml
```diff
@@ -646,6 +646,20 @@ jobs:
       Contents: '**'
       TargetFolder: '$(artifactsDir)'
 
+  - task: PowerShell@2
+    displayName: 'Copy Application Event Logs to Artifacts'
```
florelis: Just realized that this is in the fuzzing job, instead of in the test job

src/AppInstallerCLITests/main.cpp
```diff
@@ -107,6 +108,15 @@ int main(int argc, char** argv)
...
             keepSQLLogging = true;
         }
+        else if ("-mdmp"s == argv[i])
+        {
+            Debugging::EnableSelfInitiatedMinidump();
+        }
+        else if ("-mdmpto"s == argv[i])
+        {
+            ++i;
+            Debugging::EnableSelfInitiatedMinidump(std::filesystem::path{ argv[i] });
```
Trenly: Is there ever a case where this could index past the end of the array, since you're manually incrementing the index?

JohnMcPMS: Definitely not safe, but this is also the unit tests so I don't mind if there are mild traps (until I do and someone points to this comment).  I can still add the appropriate check since I have to update spelling anyway.

florelis (dismissed):

JohnMcPMS:
Bouncing for CLA check...

PR #5859, opened 2025-11-05
Support associating export units with packages in subdirectories of install location

JohnMcPMS:
## Change
The primary motivation is to support directories below the install location to contain configuration units that we will associate with the package.  This is achieved by refactoring the association logic from a Package x Unit loop into a tree structure that is colored by package install locations.  This also has the benefit of making a O(N^2) algorithm into an O(N).

Units are first inserted into the tree based on their file path.  Then the install location of each package is recorded onto that tree as well.  Finally, during the export of each package, all resources at the install location and any that are descended from it but not under another package are included.

## Validation
Added unit tests for path tree and E2E test update for subdirectory.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5859)

### DIFF
--- a/src/AppInstallerCLIE2ETests/ConfigureCommand.cs
+++ b/src/AppInstallerCLIE2ETests/ConfigureCommand.cs
@@ -364,7 +364,7 @@ namespace AppInstallerCLIE2ETests
         public void ConfigureFindUnitProcessors()
         {
             // Find all unit processors.
-            var result = TestCommon.RunAICLICommand("test config-find-unit-processors", string.Empty, timeOut: 120000);
+            var result = TestCommon.RunAICLICommand("test config-find-unit-processors", string.Empty, timeOut: 300000);
             Assert.AreEqual(0, result.ExitCode);
             Assert.True(result.StdOut.Contains("Microsoft/OSInfo"));
 
--- a/src/AppInstallerCLITests/Downloader.cpp
+++ b/src/AppInstallerCLITests/Downloader.cpp
@@ -82,7 +82,7 @@ TEST_CASE("HttpStream_ReadLastFullPage", "[HttpStream]")
 
     for (size_t i = 0; i < 10; ++i)
     {
-        stream = GetReadOnlyStreamFromURI("https://cdn.winget.microsoft.com/cache/source2.msix");
+        stream = GetReadOnlyStreamFromURI("https://aka.ms/win32-x64-user-stable");
 
         stat = { 0 };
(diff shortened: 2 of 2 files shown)


src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1730,34 +1816,28 @@ namespace AppInstaller::CLI::Workflow
...
                 {
+                    AICLI_LOG(Config, Verbose, << "Exporting package `" << package.Id << "` at: " << package.InstalledLocation);
+
                     auto packageUnit = anon::CreateWinGetPackageUnit(package, source, context.Args.Contains(Args::Type::IncludeVersions), sourceUnit, packageUnitType);
                     configContext.Set().Units().Append(packageUnit);
 
                     // Try package settings export.
-                    for (auto itr = unitProcessors.begin(); itr != unitProcessors.end(); /* itr incremented in the logic */)
+                    auto unitsForPackage = unitProcessorTree.GetResourcesForPackage(package);
+                    for (auto itr = unitsForPackage.begin(); itr != unitsForPackage.end(); ++itr)
```
dkbennett: Since iterator is no longer being altered in the logic, could be converted to range-based for loop for readability, with GetResourcesForPackage() moved into the for expression. Also const reference since it does not appear to be altered in the following code.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -1730,34 +1816,28 @@ namespace AppInstaller::CLI::Workflow
...
-                                configContext.Set().Units().Append(exportedUnit);
-                            }
+                        auto exportedUnits = anon::ExportUnit(context, configUnit);
+                        anon::AddDependentUnit(exportedUnits, packageUnit);
 
-                            // Remove the unit processor from the list after export.
-                            itr = unitProcessors.erase(itr);
-                        }
-                        else
+                        for (auto exportedUnit : exportedUnits)
```
dkbennett: const reference? This isn't being changed here, unless the Append is altering it, which would be strange.

edit: May want the copy here, I didn't look at Append to see if it's moving or copying the input.

dkbennett (approved):
Nice tree utility, solution, and performance improvement!  Some minor comments, but LGTM.

dkbennett (dismissed):

PR #5888, opened 2025-11-18
Additional logging limitations and control

JohnMcPMS:
## Change
Expands the limitations on log files and allows them to be user configurable.

The existing limit was hardcoded to remove any files over 7 days old.  This was enforced by a background thread spun up during process start.  This same thread enforces the new limits on the aggregate logging directory, while additional code in the file logger enforces the individual file size limitation.

When new log lines would exceed the maximum file size for an individual file, the stream is repositioned after the initialization portion of the log, a log line is output to indicate the wrap, and the logs continue to overwrite the older data.

--- Copied from the update to `Settings.md` ---

### file

The `file` settings control the log files generated by winget during operation.  These settings apply to the automatic cleanup that happens whenever a Windows Package Manager process is run. They only apply to the default log location, which contains winget logs, AppInstaller logs (the MSIX install UI), and is the default location where installer logs are placed. The automatic cleanup happens at the beginning of the process, so the log file(s) generated by the current process will not be considered in the limits.

|Setting|Description|Default|Note|
|---|---|---|---|
|`ageLimitInDays`|The maximum age, in days, of files in the log directory; older files are deleted.|7 (days)|Set to 0 to disable this limit.|
|`totalSizeLimitInMB`|The maximum size, in megabytes, of all files in the log directory; the oldest files are deleted first.|128 (MB)|Set to 0 to disable this limit.|
|`countLimit`|The maximum number of files in the log directory; the oldest files are deleted first.|0|Set to 0 (the default) to disable this limit.|

These settings apply to the log files that winget writes, only as they are being written.  They do not apply to files written by installers or the AppInstaller UI.

|Setting|Description|Default|Note|
|---|---|---|---|
|`individualSizeLimitInMB`|The maximum size, in megabytes, of an individual log file. If a file would exceed this limit, the logs will wrap. Note that this limit is approximate and the actual files may exceed it by a few bytes.|16 (MB)|Set to 0 to disable this limit.|

--- End copy from `Settinds.md` ---

## Validation
Added new unit tests for various internal functions and file logger.
Manually verified log wrapping.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5888)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -161,6 +161,7 @@ ERRORONEXIT
 errstr
 ESRB
 etl
+evtx
 ewgp
 ewgs
 execustom
@@ -323,6 +324,7 @@ MBH
 MBs
 mday
 mdmp
+mdmpto
 MDs
 megamorf
 microsoftentraid
@@ -616,6 +618,7 @@ wincodec
 windir
 windowsdeveloper
 winerror
+winevt
 wingdi
 wingetconfigroot
 wingetcreate
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -340,7 +340,7 @@ jobs:
 
   - pwsh: |
       $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User")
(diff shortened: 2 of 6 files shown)


src/AppInstallerCommonCore/FileLogger.cpp
```diff
@@ -148,4 +186,49 @@ namespace AppInstaller::Logging
             throw std::system_error(errno, std::generic_category());
         }
     }
+
+    void FileLogger::GetDefaultMaximumFileSize()
```
florelis: Nit: "get" makes me think it will return something. Maybe `SetMaximumFileSizeToDefault()`?

florelis (approved):

florelis (dismissed):

PR #5895, opened 2025-11-25
Improve translation of 'explicit' in Japanese

nxckl:

-----

SourceListExplicit was previously translated as「成人指定」, which incorrectly implies NSFW content. In this context, explicit means "explicitly specified" or "directly designated," referring to sources that the user has explicitly specified.

I have updated the translation to 「明示的ソース」, which accurately reflects the intended meaning and avoids confusion.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5895)

### DIFF
--- a/Localization/Resources/ja-JP/winget.resw
+++ b/Localization/Resources/ja-JP/winget.resw
@@ -2891,7 +2891,7 @@
     <value>指定しない限り、ソースを検出から除外します</value>
   </data>
   <data name="SourceListExplicit" xml:space="preserve">
-    <value>成人指定</value>
+    <value>明示的ソース</value>
   </data>
   <data name="SourceTrustLevelArgumentDescription" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


JohnMcPMS:
It will be better to update the input to include a comment about the usage of this string so that the translators can fix all languages.
https://github.com/microsoft/winget-cli/blob/9a65ba4d8ae0cb2cc91e8e3df5abeb1bce764d1c/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw#L2893-L2895

A comment element like (using your PR description):
```XML
<comment>In this context, explicit means explicitly specified or directly designated, referring to sources that the user has explicitly specified.</comment>
```

JohnMcPMS:
Closing with change to en-us values in #6033

PR #5904, opened 2025-12-01
Enable Explicit toggling for sources (i.e. Enable/Disable)

### LINKED ISSUE #5797: Disable Source without Removing it
### Relevant area(s)

WinGet CLI

I don't typically use the `msstore` source. However, I do *occasionally* use it. Additionally, I have a private source that I only want to use every once in a while - I don't want to use it all the time, but I do like searching it or using it for infrequent updates.

It would be nice to have the ability to enable and disable sources without having to add / remove them so that I don't have to have all the source details all the time. 

This also slightly ties in with GPO where a source may be present but "disabled" by GPO - this would be a logical extension of that, formalized. 

`winget source enable`
`winget source disable`
`winget source limit` (?)

dkbennett:

-----
Closes #5797
This issue was related to enabling/disabling a source without removing it. Functionally in WinGet this is already captured with the "Explicit" property, where a source is not included unless it is explicitly referenced. This is effectively equivalent to a "disabled" source. In examining this problem we determined the best way to resolve this is to allow editing of a source's properties, specifically the "Explicit" property, without having to remove and re-add a source, which is problematic with default sources. This PR implements that solution to allow editing of a Source's properties, starting with the Explicit property. This works on User defined sources and Default sources.  In terms of policy, Edit follows Remove; if you can remove a source you can edit that source. If you cannot remove the source, then you cannot edit the source either.

This PR Introduces a new experimental feature, "**sourceEdit**" which has this new editing behavior. The user experience is not final so we are keeping it behind sourceEdit for now for testing and refinement.

**New functionality: source edit subcommand**
This allows a source to be edited without removing/readding. This currently only supports the "Explicit" property of a source, and includes User sources and Default sources (such as the winget-font source).

Aliases for this command: `set` and `config`

Example1:  `winget source edit -n winget --explicit true` 
This will make the winget source explicit.

Example2:  `winget source edit -n winget-font -e false`
This will make the winget-font source no longer explicit.

This interface is probably going to change later, which is one of the reasons why we are using the experimental feature.

Key implementation detail is the addition of an "IsOverride" property to the SourceDetailsInternal which functions similar to IsTombstone, only instead of deleting the source it overrides some of the source's properties, currently the "Explicit" property.  This is needed for the Default sources. For user sources the update is straightforward.

image

**Tested**
* Added CLI tests for the default source override validation
* Added E2E tests for the command validation
* Manual verification of the command functionality.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5904)

### DIFF
--- a/src/AppInstallerSharedLib/AppInstallerStrings.cpp
+++ b/src/AppInstallerSharedLib/AppInstallerStrings.cpp
@@ -943,12 +943,12 @@ namespace AppInstaller::Utility
     {
         try
         {
-            if (ICUCaseInsensitiveEquals(input, "false"sv))
+            if (CaseInsensitiveEquals(input, "false"sv))
             {
                 return { false };
             }
 
-            if (ICUCaseInsensitiveEquals(input, "true"sv))
+            if (CaseInsensitiveEquals(input, "true"sv))
             {
                 return { true };
             }


src/AppInstallerCLICore/Commands/SourceCommand.cpp
```diff
@@ -312,4 +313,38 @@ namespace AppInstaller::CLI
...
             Workflow::ExportSourceList;
     }
+
+    // Source Edit Command
+
+    std::vector<Argument> SourceEditCommand::GetArguments() const
+    {
+        return {
+            Argument::ForType(Args::Type::SourceName).SetRequired(true),
+            Argument::ForType(Args::Type::SourceExplicit),
```
JohnMcPMS: Thinking into the future of edit, I don't want to have to specify any parameters that I do not intend to change.  That makes a flag problematic, as it only has two states and we need at least 3 for the "nullopt".  Despite it being more clunky, I would prefer `--explicit true` and `--explicit false` here.  I think it also makes it clear when I'm setting it to false, which the example of `winget source edit font` does not.

src/AppInstallerCLICore/Workflows/SourceFlow.cpp
```diff
@@ -272,6 +272,39 @@ namespace AppInstaller::CLI::Workflow
...
+            Repository::Source source{ sd.Name, isExplicit };
+
+            // Is anything being changed??
+            if (!source.IsEditOfSource(currentSource))
+            {
+                context.Reporter.Info() << Resource::String::SourceEditNoChanges(Utility::LocIndView{ sd.Name }) << std::endl;
+                continue;
+            }
+
+            context.Reporter.Info() << Resource::String::SourceEditOne(Utility::LocIndView{ sd.Name }, Utility::LocIndView{ Utility::ConvertBoolToString(isExplicit) }) << std::endl;
```
JohnMcPMS: This is likely to become unusable if edit expands to other properties.  My first thought is to use a table with old and new properties, similar to `winget source list winget`:
```
Field       Value                                  New Value
Name        winget
Type        Microsoft.PreIndexed.Package
Argument    https://cdn.winget.microsoft.com/cache
Data        Microsoft.Winget.Source_8wekyb3d8bbwe
Identifier  Microsoft.Winget.Source_8wekyb3d8bbwe
Trust Level Trusted|StoreOrigin
Explicit    false                                  true
Updated     2025-12-01 16:15:11.000
```
You could of course remove some uneditable fields like the update time (or only have the currently editable fields listed at all).

src/AppInstallerCLICore/Workflows/SourceFlow.cpp
```diff
@@ -272,6 +272,39 @@ namespace AppInstaller::CLI::Workflow
...
+            auto editFunction = [&](IProgressCallback& progress)->bool { return source.Edit(progress); };
```
JohnMcPMS: I would think that anything that currently necessitates the progress reporting would not be editable.  So other than competition over the write to the file this should be fast enough to not need progress.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -545,6 +545,20 @@ They can be configured through the settings file 'winget settings'.</value>
...
+  </data>
+  <data name="SourceEditCommandShortDescription" xml:space="preserve">
+    <value>Edit properties of a source</value>
+  </data>
+  <data name="SourceEditOne" xml:space="preserve">
+    <value>Changing source: {0} to Explicit={1}...</value>
+    <comment>{Locked="{0}","{1}"} Message displayed to inform the user about a registered repository source that is currently being edited. {0} is a placeholder replaced by the repository source name. {1} is a placeholder replaced by the changed value.</comment>
+  </data>
+    <data name="SourceEditNoChanges" xml:space="preserve">
+    <value>The source named '{0}' is already in that desired state.</value>
```
JohnMcPMS:
```suggestion
    <value>The source named '{0}' is already in the desired state.</value>
```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -545,6 +545,20 @@ They can be configured through the settings file 'winget settings'.</value>
   <data name="SourceCommandShortDescription" xml:space="preserve">
     <value>Manage sources of packages</value>
   </data>
+  <data name="SourceEditCommandLongDescription" xml:space="preserve">
+    <value>Edit properties of an existing source. A source provides the data for you to discover and install packages. Only add a new source if you trust it as a secure location.</value>
```
JohnMcPMS:
```suggestion
    <value>Edit properties of an existing source. A source provides the data for you to discover and install packages.</value>
```

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -224,6 +224,9 @@ namespace AppInstaller::Repository
         // Constructor for a source to be added.
         Source(std::string_view name, std::string_view arg, std::string_view type, SourceTrustLevel trustLevel, bool isExplicit);
 
+        // Constructor for a source to be edited.
+        Source(std::string_view name, bool isExplicit);
```
JohnMcPMS: Similar to the CLI, but much easier to update later, maybe this should take in a `std::optional<bool>`.

src/AppInstallerRepositoryCore/SourceFactory.h
```diff
@@ -41,6 +41,10 @@ namespace AppInstaller::Repository
         // Return value indicates whether the action completed.
         virtual bool Remove(const SourceDetails& details, IProgressCallback& progress) = 0;
 
+        // Edit the source from the given details..
+        // Return value indicates whether the action completed.
+        virtual bool Edit(const SourceDetails& details, IProgressCallback& progress) = 0;
```
JohnMcPMS: Similar to the progress, I'm not sure that we would allow edits to make the kinds of changes that require the actual source to be involved.  This should be removed unless you have a strong case for a future edit scenario.

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -989,6 +1014,54 @@ namespace AppInstaller::Repository
...
+        if (result)
+        {
+            SourceList sourceList;
+            sourceList.EditSource(details);
+        }
+
+        return result;
+    }
+
+    bool Source::IsEditOfSource(const Source& otherSource)
```
JohnMcPMS: This feels clunky to me.  This pseudo-code seems more natural:
```
struct SourceEdit { std::optional<bool> explicit; };
...
Source existing{ name };
SourceEdit edits{ ... };
if (!existing.RequiresChanges(edits)) { message and return; }
existing.Edit(edits);
```

JohnMcPMS (changes requested):
All updates to the interface should include COM and PowerShell consideration.

src/AppInstallerCLICore/Workflows/SourceFlow.cpp
```diff
@@ -276,32 +276,30 @@ namespace AppInstaller::CLI::Workflow
...
+            auto isExplicit = sd.Explicit;
+            if (context.Args.Contains(Execution::Args::Type::SourceEditExplicit))
+            {
+                auto explicitArg = context.Args.GetArg(Execution::Args::Type::SourceEditExplicit);
+                isExplicit = Utility::ConvertStringToBool(explicitArg);
+            }
 
-            // Is anything being changed??
-            if (!source.IsEditOfSource(currentSource))
+            Repository::SourceEdit edits{ std::optional<bool>{ isExplicit } };
```
JohnMcPMS: Using an `optional` to store the edits means that we should only send a value when the argument is supplied.

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -225,7 +234,7 @@ namespace AppInstaller::Repository
         Source(std::string_view name, std::string_view arg, std::string_view type, SourceTrustLevel trustLevel, bool isExplicit);
 
         // Constructor for a source to be edited.
-        Source(std::string_view name, bool isExplicit);
+        Source(std::string_view name, std::optional<bool> isExplicit);
```
JohnMcPMS: Can this be removed now?

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -254,12 +254,14 @@ namespace AppInstaller::Repository::Microsoft
                 return UpdateBase(details, true, progress);
             }
 
-            bool Edit(const SourceDetails& details, IProgressCallback& progress) override final
+            // Applies the edits to the source.
+            bool Edit(SourceDetails& details, const SourceEdit& edits) override final
```
JohnMcPMS: I maintain that this virtual function currently serves no purpose and that I don't see a future purpose for it either.  Currently it is only used top copy a value from one parameter to another, which can certainly be done more generically.

src/AppInstallerSharedLib/Public/AppInstallerStrings.h
```diff
@@ -290,6 +290,9 @@ namespace AppInstaller::Utility
     // Converts the given boolean value to a string.
     std::string_view ConvertBoolToString(bool value);
 
+    // Converts the given string view into a bool.
+    bool ConvertStringToBool(const std::string_view& value);
```
JohnMcPMS: If this were `sts::optional<bool> TryConvertStringToBool` you could reuse it during the argument validation.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1438,5 +1439,53 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            // Check if running as admin/system.
+            // [NOTE:] For OutOfProc calls, the Windows Package Manager Service executes in the context initiated by the caller process,
+            // so the same admin/system validation check is applicable for both InProc and OutOfProc calls.
+            THROW_HR_IF(APPINSTALLER_CLI_ERROR_COMMAND_REQUIRES_ADMIN, !AppInstaller::Runtime::IsRunningAsAdminOrSystem());
+
+            auto matchingSource = GetMatchingSource(winrt::to_string(options.Name()));
+            THROW_HR_IF(APPINSTALLER_CLI_ERROR_SOURCE_NAME_DOES_NOT_EXIST, !matchingSource.has_value());
+            ::AppInstaller::Repository::Source sourceToEdit = ::AppInstaller::Repository::Source{ matchingSource.value().Name };
+
+            auto isExplicit = matchingSource.value().Explicit;
```
JohnMcPMS: Same comment as the CLI; although here you have a converter function that you can just directly call with the `OptionalBoolean` to make things even easier.

src/AppInstallerCLICore/Workflows/SourceFlow.cpp
dkbennett: Updated to list the changed values in this manner.

src/AppInstallerSharedLib/Public/AppInstallerStrings.h
dkbennett: I changed this to TryConvertStringToBool returning std::optional , where failures result in a nullopt, and used it for argument validation as well as suggested.

This required moving away from the istringstream conversion method, since std::boolalpha conversion treated anything not exactly "true" as false, which is invalid for validation, and also exposed a bug with the previous implementation, since "TRUE" would pass argument validation but result in false conversion once it went through std::boolalpha in the old version.  Instead I replaced it with a direct case insensitive check and anything not matching "true" or "false" is nullopt.

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
dkbennett: I gutted all of this plumbing for the virtual function and have simply "EditSource" and "RequiresChanges" now.

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
dkbennett: It was still used, but it was indeed unnecessary. Removed.

src/AppInstallerCLICore/Workflows/SourceFlow.cpp
dkbennett: Fixed this to use TryConvert and pass in the result directly to the SourceEdit with no default setting.

src/AppInstallerSharedLib/AppInstallerStrings.cpp
```diff
@@ -939,11 +939,26 @@ namespace AppInstaller::Utility
...
 
-    bool ConvertStringToBool(const std::string_view& input)
+    std::optional<bool> TryConvertStringToBool(const std::string_view& input)
     {
-        bool value;
-        std::istringstream(std::string(input)) >> std::boolalpha >> value;
-        return value;
+        try
+        {
+            if (ICUCaseInsensitiveEquals(input, "false"sv))
```
JohnMcPMS: perf: You only need to use the ICU functions when both strings are user provided.

src/Microsoft.Management.Deployment.InProc/Microsoft.Management.Deployment.InProc.dll.manifest
```diff
@@ -11,53 +11,105 @@
...
         description="AddPackageCatalogOptions"/>
+    <activatableClass
+        name="Microsoft.Management.Deployment.AddPackageCatalogOptions"
+        threadingModel="both"
+        xmlns="urn:schemas-microsoft-com:winrt.v1"/>
      <comClass
         clsid="{1125D3A6-E2CE-479A-91D5-71A3F6F8B00B}"
         threadingModel="Both"
         description="RemovePackageCatalogOptions"/>
+    <activatableClass
```
JohnMcPMS, line 102: Thanks :)

PR #5950, opened 2025-12-30
Add check to ensure vcpkg triplets match across projects

florelis:
If multiple projects use different vcpkg triplets, it can lead to build errors because only one triplet can be installed at a time, which means that parallel builds have race conditions. This has hit us multiple times when the `ReleaseStatic` configuration builds a project using the `Release` configuration.

This PR adds a build-time check to ensure that all projects use the same triplet, and show a descriptive error if not.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5950)

### DIFF
--- a/src/vcpkg.props
+++ b/src/vcpkg.props
@@ -42,9 +42,10 @@
     <WriteLinesToFile File="$(_TripletFile)" Lines="$(VcpkgTriplet)" Overwrite="true" />
   </Target>
 
-  <!-- Note: Visual Studio does not build the .sln in the same way that MSBuild.exe does, so the above target is not executed.
-        In that case we try to write the file on the first project we build.
-        There is a race condition for parallel builds in doing this, but it won't happen on pipeline builds... -->
+  <!-- Note: Visual Studio does not build the .sln in the same way that MSBuild.exe does.
+       VS builds each project directly, instead of calling MSBuild on the .sln.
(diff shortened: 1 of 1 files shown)


src/vcpkg.props
```diff
@@ -2,50 +2,68 @@
...
+      <_TripletFile>$(VcpkgInstalledDir).solution-triplet</_TripletFile>
   </PropertyGroup>
+
+  <Target Name="WriteSolutionTripletToFile"
+          BeforeTargets="Build"
+          Condition="'$(_IsBuildingSln)' == 'true'">
+    <WriteLinesToFile File="$(_TripletFile)" Lines="$(VcpkgTriplet)" Overwrite="true" />
+  </Target>
+
+  <!-- Note: Visual Studio does not build the .sln in the same way that MSBuild.exe does, so the above target is not executed.
```
JohnMcPMS: The comment does not make it clear to me which build does not execute the target above.

Hopefully it is VS (aka dev inner workflow) that uses the above so that the file is updated based on changes to the chosen target, but the error text below makes me less hopeful about that...

Maybe we can have another target that runs afterward that deletes the file so that manual cleanup is not required except in extreme cases?

florelis: Updated to delete the file in case of error

PR #5975, opened 2026-01-13
Diagnostics and fix for pipeline test failures

JohnMcPMS:
## Change
Add more data to try and get to the bottom of the test failures in the pipeline.

Most importantly adds an extensive tracing that wraps the failing task to hopefully get more details.

Also adds the PID to the startup logging and a message to indicate that the server is exiting due to its event being signaled.  These won't actually be used in the pipeline right now, as the failure is happening the in GA PS/COM interaction.  But they are helpful for diagnosing issues anyway.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5975)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -331,10 +331,9 @@ jobs:
     inputs:
       packageFeedSelector: 'nugetOrg'
 
-  # Temporarily enable this tracing collection until the issue with the following Install is fixed. Then move to debug only as the COM tracing in the e2e-test.template does.
   - task: CmdLine@2
     displayName: Start HAM trace
-    condition: succeededOrFailed()
+    condition: and(succeededOrFailed(), eq(variables['System.debug'], true))
     inputs:
       script: 'wpr -start $(Build.SourcesDirectory)\tools\HAMTrace\WER.HostActivityManager.wprp -filemode'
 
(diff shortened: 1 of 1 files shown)


src/AppInstallerCLICore/ShutdownMonitoring.cpp
```diff
@@ -352,15 +352,6 @@ namespace AppInstaller::ShutdownMonitoring
 
     void ServerShutdownSynchronization::Signal(CancelReason reason)
     {
-        {
-            // Check for registered components before creating a thread to do nothing
```
florelis, line 356: I'm guessing we're removing this so that we always invoke the callback?

florelis (dismissed):

JohnMcPMS:
Examining the trace shows that OS deployment is requesting that the package be terminated after we have started trying to use it.  This ends up killing the COM server out from under the PowerShell usage.  Added the provider for deployment to understand why it is requesting that since our Repair should no longer be running at that point.

JohnMcPMS:
The tracing shows that the issue is coming from AppReadiness registering the package due to our provisioning call.  Moving to install only for the current user removes that asynchronous behavior (there does not appear to be a good reason for the provisioning in the original PR).  I'm not currently aware of any method for us to wait for the AppReadiness registration to complete so that we can prevent the caller from proceeding to use the package.

Issue #1488, opened 2021-09-20
Configure Behaviors for multiple sources


denelon:
# Configure Behaviors for multiple sources

Sometimes the same package is in multiple sources.

Now the Windows Package Manager has access to the Microsoft Store, the Community Repository, and private sources. We need better behaviors.

# Proposed technical implementation details (optional)

One possibility is to have users continue to disambiguate when a package exists in multiple sources.
Another possibility is to have a "priority" on sources so the top priority source will be used (unless specified with --source).

Trenly:
My personal preference as a user would be to have the priority of sources, and potentially to have this be configurable through settings. I would imagine that Microsoft Store should have the top priority by default, then community, then private sources. However, I can see a use case where perhaps a user would like a private source or the community repository to have higher priority by default.

This could easily be configurable by having a list of sources in the settings file such as:
```json
"SourcePriority": {
    "winget" :  1,
    "microsoft":  2,
    "UserDefined": 0,
    "AnotherUDS": 1,
    "LowerPriority":  -1
}
```
The higher the number, the higher the priority. If two are tied for priority, then whichever comes first. Any sources that aren't defined in the priority list would be treated as priority 0, with the exception of the community repository and Microsoft store, which could default to 1 and 2 respectively. Any sources which have a defined priority but don't exist in the sources list could be ignored safely.

voltagex:
Currently the user experience isn't great, and it gets worse due to poor searching of the MS Store

![image](https://user-images.githubusercontent.com/83080728/135700854-b87e7be8-281a-4041-918e-c23e28a54bf5.png)
![image](https://user-images.githubusercontent.com/83080728/135700867-6b3f413c-bd71-47c6-b9df-1618fae49a75.png)

spragginsdesigns:
> My personal preference as a user would be to have the priority of sources, and potentially to have this be configurable through settings. I would imagine that Microsoft Store should have the top priority by default, then community, then private sources. However, I can see a use case where perhaps a user would like a private source or the community repository to have higher priority.
> 
> This could be configurable by having a list of sources in the settings file such as:
> 
> ```json
> "SourcePriority": {
>     "winget" :  1,
>     "microsoft":  2,
>     "UserDefined": 0,
>     "AnotherUDS": 1,
>     "LowerPriority":  -1
> }
> ```
> 
[...]

PR #6029, opened 2026-02-04
Source priority

JohnMcPMS:
Implements #1488

## Change
Adds the concept/value of priority to sources. This is an integer value (default 0) where greater numbers are sorted first. This is an editable value of a source. Priority has been added to the following interfaces:
1. CLI commands (`source` `add`/`edit`/`list`/`export`)
2. DSCv3 `source` resource
3. DSCv2 source resource
4. Configuration export
5. COM source Add and Edit functionality
6. PowerShell source objects
7. Group policy JSON

Sources are sorted by priority in the list of sources, which has minor effects such as changing the order in `source list` and the order of available sources when constructing the default composite source.  That default composite ordering has always had the effect of making equal match type/field results dependent on the order, but now the priority can affect the third order sorting.

The larger effect is when multiple sources return results for an operation targeting available packages (ex. `install`). In this case, if there is only one result amongst the sources at the highest returned priority, then that will be used instead of producing an error about multiple results. For example, if all sources return one result and the priorities are { Source A = 2, Source B = 1, Source C = 0 }, then the result from Source A would be installed without any additional user input (although a warning is output to indicate this has happened).  But if the priorities are { Source A = 2, Source B = 2, Source C = 0 }, then an error similar to the existing one will be presented with only the A and B source results.

### REST match criteria
A TODO has also been resolved in the REST source code that evaluates the match criteria of REST results so that they may be sorted properly amongst the winget-pkgs results.  Previously they were always using default values (0s) which would cause them to be sorted first.

## Validation
Added tests across the spectrum of use.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6029)

### DIFF
--- a/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
+++ b/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
@@ -351,7 +351,7 @@ They can be configured through the settings file 'winget settings'.</value>
     <value>Log location (if supported)</value>
   </data>
   <data name="MainCopyrightNotice" xml:space="preserve">
-    <value>Copyright (c) Microsoft Corporation. All rights reserved.</value>
+    <value>© 2026 Microsoft. All rights reserved.</value>
   </data>
   <data name="MainHomepage" xml:space="preserve">
     <value>Homepage</value>
--- a/doc/windows/package-manager/index.md
+++ b/doc/windows/package-manager/index.md
@@ -6,7 +6,7 @@ ms.topic: overview
 ms.localizationpriority: medium
 ---
 
-# Windows Package Manager (preview)
+# Windows Package Manager
 
 Windows Package Manager is a comprehensive [package manager solution](#understanding-package-managers) that consists of a command line tool and set of services for installing applications on Windows 10 and Windows 11.
 
--- a/tools/PowerShell/Microsoft.WinGet.Client/crescendo/WindowsPackageManager.psm1
+++ b/tools/PowerShell/Microsoft.WinGet.Client/crescendo/WindowsPackageManager.psm1
@@ -66,7 +66,7 @@ PROCESS {
 <#
 .SYNOPSIS
 Windows Package Manager v1.1.12702
-Copyright (c) Microsoft Corporation. All rights reserved.
+© Microsoft Corporation. All rights reserved.
 
 The winget command line utility enables installing applications and other packages from the command line.
 


src/AppInstallerCLIE2ETests/DSCv3SourceResourceCommand.cs
```diff
@@ -261,6 +274,7 @@ public void Source_Test_AllMatch()
                 Type = DefaultSourceType,
                 TrustLevel = TrustedTrustLevel,
                 Explicit = true,
+                Priority = 42,
```
mdanish-kh, line 277: I see what you did there

mdanish-kh:
> But if the priorities are { Source A = 2, Source B = 2, Source C = 0 }, then an error similar to the existing one will be presented with only the A and B source results.

In such a scenario (for a single result as you mentioned), it may be worth checking if one of the sources returned version "Unknown" for the package (common with `msstore`), and if such that one implicitly gets the lower priority

src/AppInstallerCLIE2ETests/DSCv3SourceResourceCommand.cs
JohnMcPMS: On obviously arbitrary number for testing purposes only...

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3535,4 +3535,19 @@ An unlocalized JSON fragment will follow on another line.</comment>
     <value>New Value</value>
     <comment>Column title for listing the new value.</comment>
   </data>
+  <data name="SourcePriorityArgumentDescription" xml:space="preserve">
+    <value>Priority with higher numbers first</value>
```
florelis: I don't love this string, but I can't think of anything much better. Maybe `Priority. Higher numbers go first`?

I think my issue is that I'm reading it as implying that there are other types of priority (where higher numbers aren't the first.)

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -1310,11 +1310,54 @@ namespace AppInstaller::CLI::Workflow
...
+                        // Current priority is lower, ignore the match.
+                        continue;
+                    }
+
+                    highestPriorityMatches.emplace_back(match);
+                }
+
+                if (highestPriorityMatches.size() < searchResult.Matches.size())
+                {
+                    AICLI_LOG(CLI, Info, << "Replacing search results with only those from the highest priority [" << (highestPriority ? std::to_string(highestPriority.value()) : "none"s) << "].");
```
florelis, line 1350: Do we need the case for `"none"`? AFAICT, `highestPriority` will always have a value here.

florelis (dismissed):

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: Think of ADO priority numbers; they use lower is more important.  I changed it slightly.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
JohnMcPMS: I think it probably will due to the `operationTargetsInstalled` that I added (after implementing this line).  But I have definitely thought that "this will always have a value if we get here" before and been wrong.  No reason to crash (exception really) just because we wanted to save a few characters and a branch.

PR #5962, opened 2026-01-08
Encapsulate argument with quotes in German translation for clarity

ravage84:
While the actual example command "winget pin" is already quoted, the "--include-pinned" argument was not. Due to German use of hyphens, this was less than optimal for readability.

- [X] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs).
- [ ] ~~I have updated the [Release Notes](../doc/ReleaseNotes.md).~~ => change too little to be relevant
- [ ] ~~This pull request is related to an issue.~~ => If a trivial change like this really needs to be related to an issue, someone please create one yourself. 🤦‍♂️ 

-----

PR #5943, opened 2025-12-22
Improve translation of 'FontFamily' and 'FontFace' in Chinese(zh-CN and zh-TW)

Na-moe:
- [x] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs).
- [ ] I have updated the [Release Notes](../doc/ReleaseNotes.md).
- [ ] This pull request is related to an issue.

-----

## Description
This PR improves the Chinese translation of technical terms 'FontFamily' and 'FontFace' to be more accurate and context-appropriate.

## Changes Made
* FontFamily: Updated translation from "家族" to "字体族"(zh-CN) and "字型族"(zh-TW). Related comments are appended.

  * "家族" literally means "family" in the sense of parents and children, which is not appropriate for typography context.

* FontFace: Updated translation from "脸" to "字体样式"(zh-CN) and "字型樣式"(zh-TW).

  * "脸" literally means the physical face of people, which is completely unrelated.
  * "字体样式" is the usual translation of font typeface.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5943)

PR #6033, opened 2026-02-06
Coalesce comments from loc suggestions

JohnMcPMS:
## Change
Put the localization suggestions from a few PRs into comments on the en-us version that goes through the localization process.

#5895
#5943
#5962

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6033)

### DIFF
--- a/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
+++ b/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
@@ -2909,7 +2909,7 @@ Please specify one of them using the --source option to proceed.</value>
   </data>
   <data name="SourceListExplicit" xml:space="preserve">
     <value>Explicit</value>
-    <comment>In this context, explicit means explicitly specified or directly designated, referring to sources that the user has explicitly specified.</comment>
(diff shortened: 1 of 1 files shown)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -2909,6 +2909,7 @@ Please specify one of them using the --source option to proceed.</value>
   </data>
   <data name="SourceListExplicit" xml:space="preserve">
     <value>Explicit</value>
+    <comment>In this context, explicit means explicitly specified or directly designated, referring to sources that the user has explicitly specified.</comment>
```
florelis: Nit: I think that saying "it's the opposite of implicit" could help, and also saying how this is used.

```suggestion
    <comment>Used as a header in a table that shows the properties of a "source". In this context, "explicit" means that the source needs to be explicitly specified or directly designated to be used, as opposed to being implicitly included.</comment>
```

florelis (approved):

PR #6067, opened 2026-02-24
Added more info for "Installer Types" values in Settings.md.

DandelionSprout:

-----
Strongly inspired by:
* https://github.com/microsoft/winget-pkgs/issues/342123
* And secondarily by how I've wondered how I could get packages to default to MSIX on my end by default.

The current documentation on the priority orders of installer types when using `winget install` without having specified `--installer-type`, was unfortunately pretty scarce. So I've filled in most of the gaps now.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6067)

### DIFF
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -182,7 +182,7 @@ The `architectures` behavior affects what architectures will be selected when in
 
 ### Installer Types
 
-The `installerTypes` behavior affects what installer types will be selected when installing a package. It can also determine which type to install by default if a manifest has multiple types: The first (leftmost) ones are looked for first, which can be convenient for users who for instance prefer portable packages or MSIX/AppX installations. The matching parameter is `--installer-type`.
+The `installerTypes` behavior affects what installer types will be selected when installing a package. It can also determine which type to install by default if a manifest has multiple types: The list is in priority order, with the first listed type being preferred over the others, and so on.  This is convenient for users who for instance prefer portable packages or MSIX/AppX installations. The matching parameter is `--installer-type`, which will override the settings.
 
 Allowed values as of version 1.12.470 include: `appx`, `burn`, `exe`, `font`, `inno`, `msi`, `msix`, `msstore`, `nullsoft`, `portable`, `wix`, `zip`
 


doc/Settings.md
```diff
@@ -180,7 +182,11 @@ The `architectures` behavior affects what architectures will be selected when in
 
 ### Installer Types
 
-The `installerTypes` behavior affects what installer types will be selected when installing a package. The matching parameter is `--installer-type`.
+The `installerTypes` behavior affects what installer types will be selected when installing a package. It can also determine which type to install by default if a manifest has multiple types: The first (leftmost) ones are looked for first, which can be convenient for users who for instance prefer portable packages or MSIX/AppX installations. The matching parameter is `--installer-type`.
```
JohnMcPMS:
```suggestion
The `installerTypes` behavior affects what installer types will be selected when installing a package. It can also determine which type to install by default if a manifest has multiple types: The list is in priority order, with the first listed type being preferred over the others, and so on.  This is convenient for users who for instance prefer portable packages or MSIX/AppX installations. The matching parameter is `--installer-type`, which will override the settings.
```

doc/Settings.md
```diff
@@ -180,7 +182,11 @@ The `architectures` behavior affects what architectures will be selected when in
...
+
+Allowed parameters as of version 1.12.470 include: `appx`, `burn`, `exe`, `font`, `inno`, `msi`, `msix`, `msstore`, `nullsoft`, `portable`, `wix`, `zip`
```
JohnMcPMS:
```suggestion
Allowed values as of version 1.12.470 include: `appx`, `burn`, `exe`, `font`, `inno`, `msi`, `msix`, `msstore`, `nullsoft`, `portable`, `wix`, `zip`
```

doc/Settings.md
```diff
@@ -180,7 +182,11 @@ The `architectures` behavior affects what architectures will be selected when in
...
+
+If this setting is not set, Winget defaults to the installer type that is listed highest up in the manifest's installer YAML (if the app has not been installed yet), or to the installer type that the existing installation of the package uses.
```
JohnMcPMS:
```suggestion
By default, and with all other properties being equal, WinGet defaults to the installer type that is listed first in the manifest's installer YAML if the package has not been installed yet.  If it is already installed, the same installer type will be required to ensure a proper upgrade.
```

DandelionSprout: I'm unsure if the double spacing before "If it is (…)" is supposed to be there, but sure thing.

JohnMcPMS: > I'm unsure if the double spacing before "If it is (…)" is supposed to be there, but sure thing.

I was taught to double space between lines. The markdown renderer/browser ignores it anyway. < Single
I was taught to double space between lines.  The markdown renderer/browser ignores it anyway. < Double

Feel free to make it a single space, it is just in my muscle memory while typing.

DandelionSprout:
Any time estimate for when the changes will get reflected on the https://learn.microsoft.com/en-us/windows/package-manager/winget/settings page, by the way?

denelon:
@DandelionSprout, you could submit a PR there as well, and I can help foster it through :)

Otherwise, it may take a bit longer for that content to get updated internally.

DandelionSprout:
I noticed around 15min after posting my comment that I indeed could submit there. 😅 And it has succeeded.

Issue #4550, opened 2024-06-13
The update process fails although the package was successfully updated.


ne-tlw:
### Brief description of your issue

Winget Log file shows that a package has been successfully installed. In the next line, the installation is marked as failed and a new installation of the package (instead of the update) is attempted. Here are three examples. The issue occurs with packages of all kind but not with all packages. Three examples attached. Actually, the update has been finished successfully.

### Steps to reproduce

Winget upgrade --id KeePassXCTeam.KeePassXC --accept-package-agreements --accept-source-agreements -h
Winget upgrade --id TeamViewer.TeamViewer --accept-package-agreements --accept-source-agreements -h
Winget upgrade --id Microsoft.EdgeWebView2Runtime --accept-package-agreements --accept-source-agreements -h

### Expected behavior

Upgrade finshed successfully.

### Actual behavior

Upgrade fails and Winget tries to upgrade packages at every run. 

08:12:23 - Updating KeePassXC from 2.7.7 to 2.7.8...
08:12:23 - ##########   WINGET UPGRADE PROCESS STARTS FOR APPLICATION ID 'KeePassXCTeam.KeePassXC'   ##########
08:12:23 - -> Running: Winget upgrade --id KeePassXCTeam.KeePassXC --accept-package-agreements --accept-source-agreements -h
Successfully installed
08:13:18 - -> An upgrade for KeePassXC failed, now trying an install instead...
08:13:18 - -> Running: Winget install --id KeePassXCTeam.KeePassXC --accept-package-agreements --accept-source-agreements --force -h
Successfully installed
08:13:34 - ##########   WINGET UPGRADE PROCESS FINISHED FOR APPLICATION ID 'KeePassXCTeam.KeePassXC'   ##########
08:13:34 - KeePassXC update failed

08:08:15 - Updating TeamViewer from 15.53.6 to 15.53.7...
08:08:15 - ##########   WINGET UPGRADE PROCESS STARTS FOR APPLICATION ID 'TeamViewer.TeamViewer'   ##########
08:08:15 - -> Running: Winget upgrade --id TeamViewer.TeamViewer --accept-package-agreements --accept-source-agreements -h
Successfully installed
08:09:29 - -> An upgrade for TeamViewer failed, now trying an install instead...
08:09:29 - -> Running: Winget install --id TeamViewer.TeamViewer --accept-package-agreements --accept-source-agreements --force -h
Successfully installed
[...]

PR #6097, opened 2026-03-21
Verify installed version matches upgrade target after installer success

AMDphreak:
## Summary

Follow-up to misleading **success** output when an **upgrade** exits **0** but Windows still reports an **older** installed version (see [#4550](https://github.com/microsoft/winget-cli/issues/4550) and related reports).

## Behavior

After a successful installer return code, **if** the operation is an **upgrade** (`InstallerExecutionUseUpdate`):

- **ARP path:** After ARP correlation finds the installed package, compare `Utility::Version` of the correlated **installed** version to the **manifest** package version. If the installed version sorts **less than** the expected (manifest) version, fail with `APPINSTALLER_CLI_ERROR_UPGRADE_INSTALLED_VERSION_MISMATCH` (`0x8A15008F`).
- **MSIX path:** When the manifest provides a package family name and the package is registered, compare using `Deployment::GetInstalledVersionStringForFamilyName` where applicable.
- **Skip** verification when reboot is required to finish, when versions are unknown, or when correlation does not yield a reliable installed version string.

## Notes

Portable/zip flows that do not correlate to ARP/MSIX are unchanged.

## Testing

CI / AppInstallerCLITests as usual.

Made with [Cursor](https://cursor.com)


JohnMcPMS:
Two things:
1. Building another PR on top of the changes in the first makes reviewing it independently difficult.  Please create a new branch for each change off of the mainline.  Yes, if you are adding errors you are very likely to create a conflict with yourself.
2. I don't know that we can take this change with it producing an error. Some installers are not well behaved and will not update the installed version number in an update.  Those usually get tagged in the manifest to prevent repeated upgrades, but I don't want to start treating updates to them as failures.  I would be more accepting of a warning message that indicates that the upgrade did this, but if the process produces success we have to treat it as such.

Trenly:
> I don't know that we can take this change with it producing an error. Some installers are not well behaved and will not update the installed version number in an update. Those usually get tagged in the manifest to prevent repeated upgrades, but I don't want to start treating updates to them as failures. I would be more accepting of a warning message that indicates that the upgrade did this, but if the process produces success we have to treat it as such.

I appreciate you callimg this out - The community repository just surpassed 12,000 packages that are available which means that any changes to reporting an error here would be breaking in my opinion; especially when you consider the automations and integrations enterprises and independent developers have built around the existing behavior of the CLI.

PR #6098, opened 2026-03-21
fix(install): defer success message until post-install checks

AMDphreak:
## Summary

Delays the **Successfully installed** / repair success line until after post-install work runs (ARP snapshot correlation, upgrade version verification, record install, installation notes).

## Details

- New context flags: `DeferInstallSuccessMessage`, `PendingDeferredInstallSuccess`, `SuppressDeferredInstallSuccess`.
- `InstallPackageInstaller` sets defer, runs the existing pipeline, then `ReportDeferredInstallSuccess`.
- `ReportInstallerResult`, `MsixInstall`, and Microsoft Store install/update/repair paths respect defer; MSIX registration-deferred suppresses generic success (warning already shown).
- `PendingDeferredInstallSuccess` avoids printing success when the installer never completed (for example install lock cancelled before a success path).

## Related

Complements accurate reporting after install (for example ARP / version checks); avoids claiming success before those steps finish.

## Testing

CI build; manual install scenarios recommended.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/AppInstallerCLICore/ExecutionContext.h
+++ b/src/AppInstallerCLICore/ExecutionContext.h
@@ -76,6 +76,12 @@ namespace AppInstaller::CLI::Execution
         RebootRequired = 0x400,
         RegisterResume = 0x800,
         InstallerExecutionUseRepair = 0x1000,
+        // InstallPackageInstaller: suppress immediate success until post-install checks complete.
+        DeferInstallSuccessMessage = 0x2000,
+        // MSIX registration deferred: do not emit generic success after checks (warning already shown).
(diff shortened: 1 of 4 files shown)


JohnMcPMS:
I don't agree with this change; the install was successful so we message that out.  If we fail to do our own bookkeeping that can result in a failure of the overall operation, but the package install has already concluded.  There can certainly be downstream problems created by that post-install failure, but any remediation would not involve re-running the package install.

If there is a real issue that arises here I would prefer to see a comprehensive solution to handling post-operation failures.  This change will end up making it look like the user request was not satisfied, when it is instead much closer to not getting a receipt with your purchase.

Issue #5844, opened 2025-10-30
WingetSourceCreator cannot sign package when spaces are in cert path


JohnnyElvis:
### Relevant area(s)

WinGet CLI

### Brief description of your issue

WingetSourceCreator cannot handle spaces in certificate path. Root cause is within the method SignFile in WingetSourceCreator\Helpers.cs

You need to change this:

`string command = $"sign /a /fd sha256 /f {signature.CertFile} ";`

to this:

`string command = $"sign /a /fd sha256 /f \"{signature.CertFile}\" ";`

You can implement #4948 while you are on it.

### Steps to reproduce

Build solution. You might want to apply fix from #4181 so you can actually work with the IndexCreationTool.

Run the IndexCreationTool with Signature.CertFile path having an empty space somewhere in between.

When using path e.g "D:/Jenkins/WrkDir/workspace/Winget Builder/Build Winget Package/signing_cert.pfx" you will be greeted with "SignTool Error: File not found: D:/Jenkins/WrkDir/workspace/Winget"

### Expected behavior

Being able to deal with empty spaces in a path

### Actual behavior

When using path e.g "D:/Jenkins/WrkDir/workspace/Winget Builder/Build Winget Package/signing_cert.pfx" you will be greeted with "SignTool Error: File not found: D:/Jenkins/WrkDir/workspace/Winget"

### Environment

```shell
N/A
```

denelon:
@JohnnyElvis I don't fully understand your scenario, but if you wanted to make a PR and reference this issue, I can have the engineering team review it.

JohnnyElvis:
Hi @denelon, the issue here are the missing quotes when calling signtool.exe. It is not related to any specific scenario.

Current implementation:
signtool.exe /a /fd sha256 /f C:\path with spaces\cert.pfm results in signtool not being able to find "C:\path" instead of the full path  "C:\path with spaces\cert.pfm"

Working implementation:
signtool.exe /a /fd sha256 /f "C:\path with spaces\cert.pfm" results in in signtoola ctually finding the full path.

If possible I'd like to submit the fixes since it is unpleasant maintaining a fork. Please share with me instructions on how to this properly.

To explain referenced issues:
Index Creation Tool (and likely LocalhostWebServer) do not work without applying #4181. Current implementation fails to populate working directory with manifest files.

#4948 is a nice to have when building sources.msix which need to remain valid even after signing cert itself expired

Issue #4948, opened 2024-11-07
Microsoft.WinGetSourceCreator.Helpres not using time server for signing


JohnnyElvis:
### Brief description of your issue

Microsoft.WinGetSourceCreator.Helpres SignFile is not using a time server for signing.

Once a code signing certificate expires signed packages will no longer be usable without a time stamp.

### Steps to reproduce

Look into src\WinGetSourceCreator\Helpres.cs

Line 39

### Expected behavior

Time stamp is added to signature

### Actual behavior

No time stamp is added to signature

![Image](https://github.com/user-attachments/assets/1ad04d3c-b146-4fee-8d4e-ecaa84144aad)

### Environment

Windows 11 Enterprise

denelon:
The PreIndexed package source is updated every time the publishing pipeline runs over at winget-pkgs. In general, the index gets refreshed mutiple times per day. I'm not sure we actually need or want to have this package to be usable if it's expired. The `winget source reset --force` command will reset the sources and the `winget source update` command would also help a user get the latest version of this package.

JohnnyElvis:
@denelon, sry for the late reply. Yes, in my use case it needs to be usable even if the signing cert did expire. The pre-indexed package source is being bult on demand. Additionally it's distribution to other environments is chained to a change process.

It's literally less than 5 minutes of work to add this

JohnnyElvis:
@denelon - to do this just alter code in "WingetSourceCreator\Helpers.cs" in the "SignFile" method to this:

`            string pathToSDK = SDKDetector.Instance.LatestSDKBinPath;
            string signtoolExecutable = Path.Combine(pathToSDK, "signtool.exe");
            string command = $"sign /a /fd sha256 /f {signature.CertFile} ";
            if (!string.IsNullOrEmpty(signature.Password))
            {
                command += $"/p {signature.Password} ";
            }
            if (!string.IsNullOrEmpty(signature.TimeServer))
            {
                command += $"/tr {signature.TimeServer} /td SHA256 ";
            }
            command += fileToSign;
            RunCommand(signtoolExecutable, command);`


The just add this line to "WingetSourceCreator\Model\Signature.cs" to the "Signature" class.

`public string? TimeServer { get; set; }`


It's done then!

denelon:
@JohnnyElvis I don't fully understand your scenario, but if you wanted to make a PR and reference this issue, I can have the engineering team review it.

PR #6113, opened 2026-03-27
Add Timeserver Support for SourceCreator and support spaces in paths and file names

Trenly:
  - #5844 
  - #4948 

Implemented as described in the issues, validated by copilot

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6113)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -138,6 +138,7 @@ devhome
 DFX
 DHAVE
 dic
+digicert
 diskfull
 DISPLAYCATALOG
 DMC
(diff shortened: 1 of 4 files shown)


florelis:
Out of curiosity, do you have a specific scenario where you need to have a timestamp on the source?
I understand that timestamping is good in general, just wondering where this came from because I can't imagine a source version being used long enough for the signing cert to have expired

florelis (approved):

JohnMcPMS:
> Out of curiosity, do you have a specific scenario where you need to have a timestamp on the source? I understand that timestamping is good in general, just wondering where this came from because I can't imagine a source version being used long enough for the signing cert to have expired

I don't think @Trenly has that scenario, but from the issue I think there are people using the tool to create private source data.  Maybe they have an offline scenario with relatively fixed package data?

Issue #1512, opened 2021-09-25
Enhance import/export schema to include override values


MisinformedDNA:
Add a `Override` property to the schema for each package.

Beyond using the Import feature to specify _what_ I want to install, I would like to specify _how_ I want to install. For instance, when I install `Git`, I prefer to exclude shell integration. For VS, I probably want to specify certain workloads.

At it's simplest form, a user can simply edit the schema file to include their desired overrides. A more complex use case would be for winget to remember the arguments used during install so that it can export those values as well.

We may also want to add a command line option for `winget input` to ignore the overrides.

# Proposed technical implementation details (optional)

Desired schema:

```json
"Packages": {
  "items": {
    "properties": {
      "PackageIdentifier": { },
      "Version": { },
      "Channel": { },
      "Scope": { },
      "Override": {
        "description": "Arguments passed through to --override",
        "type": "string"
      },
    }
  }
}
```

Desired example
```json
{
  "Sources": 
  [
    {
      "Packages": 
      [
        {
          "PackageIdentifier": "Microsoft.VisualStudio.2019.Enterprise",
          "Override": "--add Microsoft.VisualStudio.Workload.Universal"
        },
        {
          "PackageIdentifier": "Git.Git",
          "Override": "/NoShellIntegration /NoGitLfs"
        },
      ],
    }
  ],
}
```

jedieaston:
I really want this too, because for things like Visual Studio there's a very small chance you're going to use the arguments we pass in since you want your own workloads. Saving the custom arguments somewhere (although I know winget doesn't have a lot of state on the system right now, this might be something where that should change) so that they can be put into the file you get on `winget export` would be awesome!

alananderson-Allocate:
I was trying to use this for adding workloads to my install of Visual Studio.

Trouffman:
What can we do to help move on this?
Would be super useful especially for pre-defined install scenarios where the export / import feature would be helpful.

Note that the Import support for such feature should come before the support for populating this in the "export" function (as some parameters might not be good to re-export - relevant to only one machine, etc..)

Eonasdan:
Doesn't the [v2 schema](https://raw.githubusercontent.com/microsoft/winget-cli/master/schemas/JSON/packages/packages.schema.2.0.json) support doing this now?

Trouffman:
@Eonasdan it does not look like it.

There is a section 
              "Argument": {
                "description": "Argument used to install the source",
                "type": "string"
              },
[...]

PR #6118, opened 2026-03-30
Preserve overrides with export and import

Trenly:
  - #1512 

## ﻿Summary

When a package is installed with --override or --custom arguments, those values are now preserved through the winget export / winget import roundtrip, allowing packages to be reinstalled with the same
customizations automatically.

### Export (winget export)

 - The installed package metadata is queried for previously stored UserOverrideArguments and UserCustomSwitches values and written into the export file under the new OverrideArguments and CustomSwitches 
fields per package entry.
 - The packages schema v2.0 has been updated to define both new optional fields.

### Import (winget import)

 - When processing packages from an import file, OverrideArguments and CustomSwitches are now forwarded as --override and --custom args respectively into each package's install sub-context.
 - Both fields are fully independent and optional — packages without them are unaffected.

### Install metadata persistence

 - --override and --custom values supplied at install time are now stored in the package's installed metadata (UserOverrideArguments / UserCustomSwitches), making them available for future export 
operations and upgrade flows.

### Tests

 - PackageCollection unit tests covering serialization and deserialization of OverrideArguments and CustomSwitches in both the v2.0 schema parser and JSON writer, including cases where either, both, or 
neither field is present.
 - ExportFlow unit tests verifying that installed packages with stored override/custom metadata produce the correct fields in the export file.
 - CompositeSource unit tests covering propagation of UserOverrideArguments and UserCustomSwitches metadata.
 - ImportFlow unit tests (ImportFlow_WithOverrideArgs, ImportFlow_WithCustomSwitches) verifying that each field is independently applied during install — override args replace all installer arguments while
 custom switches are appended alongside the defaults.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6118)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/InstallFlow.cpp
@@ -1021,5 +1021,31 @@ namespace AppInstaller::CLI::Workflow
                 version.SetMetadata(Repository::PackageVersionMetadata::UserIntentLocale, itr->second);
             }
         }
+
+        if (context.Args.Contains(Execution::Args::Type::Override))
+        {
+            version.SetMetadata(Repository::PackageVersionMetadata::UserOverrideArguments, context.Args.GetArg(Execution::Args::Type::Override));
+        }
+        else
+        {
+            auto itr = installedMetadata.find(Repository::PackageVersionMetadata::UserOverrideArguments);
+            if (itr != installedMetadata.end())
+            {
+                version.SetMetadata(Repository::PackageVersionMetadata::UserOverrideArguments, itr->second);
+            }
+        }
+
+        if (context.Args.Contains(Execution::Args::Type::CustomSwitches))
+        {
+            version.SetMetadata(Repository::PackageVersionMetadata::UserCustomSwitches, context.Args.GetArg(Execution::Args::Type::CustomSwitches));
+        }
+        else
+        {
+            auto itr = installedMetadata.find(Repository::PackageVersionMetadata::UserCustomSwitches);
+            if (itr != installedMetadata.end())
+            {
+                version.SetMetadata(Repository::PackageVersionMetadata::UserCustomSwitches, itr->second);
+            }
+        }
     }
 }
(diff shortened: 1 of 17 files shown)


src/AppInstallerCLICore/Workflows/InstallFlow.cpp
```diff
@@ -1021,5 +1021,31 @@ namespace AppInstaller::CLI::Workflow
...
             }
         }
+
+        if (context.Args.Contains(Execution::Args::Type::Override))
+        {
+            version.SetMetadata(Repository::PackageVersionMetadata::UserOverrideArguments, context.Args.GetArg(Execution::Args::Type::Override));
+        }
+        else
+        {
+            auto itr = installedMetadata.find(Repository::PackageVersionMetadata::UserOverrideArguments);
```
JohnMcPMS: How should this behave in the event of upgrades requiring an override or custom switch?  It seems likely that you would only ever want to keep export the initial arguments, not when I had to pass `/doOneTimeWorkaroundToUpgrade` to get things to work.

I think it would be better if these metadata items were explicitly named "initial" or similar and were only ever carried forward.  Or were named "latest" and "for export" and work the way you have now.

JohnMcPMS (commented):
Everything is great except how we want to handle these values with the new version installs.  Makes me think we should have package level metadata, but that is probably an equally large plumbing change and an even larger index management change.

The other values are simpler things like locale or architecture intent, which we expect to remain consistent between upgrades without any user input.  But custom install arguments are definitely not going to be consistent inputs between initial install and upgrades.

src/AppInstallerCLICore/Workflows/InstallFlow.cpp
Trenly: My thought here was that an upgrade would generally either not need the arguments, or would be effectively a re-install with different arguments, such as if modifying the workloads an app is configured for.

However, I do see that there could be a one-time switch necessary for the upgrade. While in most cases this could be handled by the manifest's Upgrade installer switch, that isn’t a guarantee.

I'll think on this more, but will likely make it so that they are kept as the initial values. If a user wishes to change them as part of an upgrade, they could use `install --force`

JohnMcPMS (approved):

PR #6119, opened 2026-03-31
Admin setting and group policy for configuration processor path argument

JohnMcPMS:
## Change
Adds an admin setting (`ConfigurationProcessorPath`) and group policy (`EnableWindowsPackageManagerConfigurationProcessorPath`) to gate access to the `--processor-path` argument for configuration commands.

Yes, it sadly does take touching 23 files to implement that.

## Validation
New unit test for gp/setting interaction and e2e test for enforcement.  Manual confirmation of admin setting functionality as well.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6119)

### DIFF
--- a/doc/admx/en-US/DesktopAppInstaller.adml
+++ b/doc/admx/en-US/DesktopAppInstaller.adml
@@ -133,6 +133,13 @@ If you disable this setting, users will not be able to use the Windows Package M
         If you enable or do not configure this setting, users will be able to use the Windows Package Manager's MCP server.
 
         If you disable this setting, users will not be able to to use the Windows Package Manager's MCP server.</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration processor path</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPathExplanation">
+        This policy controls whether users can specify a custom DSC processor path via the --processor-path argument in Windows Package Manager configuration commands. This option is intended for testing purposes.
+
+        If you enable or do not configure this setting, users will be able to specify a custom DSC processor path in configuration commands.
+
+        If you disable this setting, users will not be able to specify a custom DSC processor path in configuration commands.</string>
       <string id="WindowsPackageManagerDefaultProxy">Set Windows Package Manager Default Proxy</string>
       <string id="WindowsPackageManagerDefaultProxyExplanation">This policy controls the default proxy used by the Windows Package Manager.
 
--- a/doc/admx/en-US/DesktopAppInstaller.adml
+++ b/doc/admx/en-US/DesktopAppInstaller.adml
@@ -133,7 +133,7 @@ If you disable this setting, users will not be able to use the Windows Package M
         If you enable or do not configure this setting, users will be able to use the Windows Package Manager's MCP server.
 
         If you disable this setting, users will not be able to to use the Windows Package Manager's MCP server.</string>
-      <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration processor path</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration Processor Path</string>
       <string id="EnableWindowsPackageManagerConfigurationProcessorPathExplanation">
         This policy controls whether users can specify a custom DSC processor path via the --processor-path argument in Windows Package Manager configuration commands.
 
--- a/doc/admx/en-US/DesktopAppInstaller.adml
+++ b/doc/admx/en-US/DesktopAppInstaller.adml
@@ -135,9 +135,11 @@ If you disable this setting, users will not be able to use the Windows Package M
         If you disable this setting, users will not be able to to use the Windows Package Manager's MCP server.</string>
       <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration processor path</string>
       <string id="EnableWindowsPackageManagerConfigurationProcessorPathExplanation">
-        This policy controls whether users can specify a custom DSC processor path via the --processor-path argument in Windows Package Manager configuration commands. This option is intended for testing purposes.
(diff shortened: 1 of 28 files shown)


doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -133,6 +133,13 @@ If you disable this setting, users will not be able to use the Windows Package M
         If you enable or do not configure this setting, users will be able to use the Windows Package Manager's MCP server.
 
         If you disable this setting, users will not be able to to use the Windows Package Manager's MCP server.</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration processor path</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPathExplanation">
+        This policy controls whether users can specify a custom DSC processor path via the --processor-path argument in Windows Package Manager configuration commands. This option is intended for testing purposes.
```
florelis: If it is for testing, do we need to expose it as a GP?

doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -133,6 +133,13 @@ If you disable this setting, users will not be able to use the Windows Package M
         If you enable or do not configure this setting, users will be able to use the Windows Package Manager's MCP server.
 
         If you disable this setting, users will not be able to to use the Windows Package Manager's MCP server.</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration processor path</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPathExplanation">
+        This policy controls whether users can specify a custom DSC processor path via the --processor-path argument in Windows Package Manager configuration commands. This option is intended for testing purposes.
+
+        If you enable or do not configure this setting, users will be able to specify a custom DSC processor path in configuration commands.
+
+        If you disable this setting, users will not be able to specify a custom DSC processor path in configuration commands.</string>
```
florelis, line 144: I assume that not being able to set the custom processor path is the more secure option? Shouldn't that be the default if not configured?

doc/admx/en-US/DesktopAppInstaller.adml
JohnMcPMS: I removed that bit of the text.  I think if it exists it is worth having the GP for.

doc/admx/en-US/DesktopAppInstaller.adml
JohnMcPMS: Yes, but maybe it should read "then the admin setting is considered, which is default disabled".  This is similar to local manifest, as is the text here.

Not Configured -> Admin setting is considered, defaults to disabled.
Enabled/Disabled -> Overrides admin setting.

doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -133,6 +133,13 @@ If you disable this setting, users will not be able to use the Windows Package M
         If you enable or do not configure this setting, users will be able to use the Windows Package Manager's MCP server.
 
         If you disable this setting, users will not be able to to use the Windows Package Manager's MCP server.</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration processor path</string>
```
Trenly: Nit: Seems most other GPO names are Title Cased

Trenly (commented):
Are there future plans for a `DefaultConfigurationProcessor` setting so that admins can force a custom processor, or users can set their own default (if allowed by policy) ?

JohnMcPMS:
> Are there future plans for a `DefaultConfigurationProcessor` setting so that admins can force a custom processor, or users can set their own default (if allowed by policy) ?

Not currently.  If that is strongly desired, I would suggest building a custom wrapper in the interim.

doc/admx/en-US/DesktopAppInstaller.adml
```diff
@@ -133,6 +133,15 @@ If you disable this setting, users will not be able to use the Windows Package M
         If you enable or do not configure this setting, users will be able to use the Windows Package Manager's MCP server.
 
         If you disable this setting, users will not be able to to use the Windows Package Manager's MCP server.</string>
+      <string id="EnableWindowsPackageManagerConfigurationProcessorPath">Enable Windows Package Manager Configuration Processor Path</string>
```
florelis: I'm not a fan of this string. Maybe "Enable custom Configuration Processor Path for Windows Package Manager"?

florelis (dismissed):

doc/admx/en-US/DesktopAppInstaller.adml
JohnMcPMS: @denelon

Two questions:
- Do you think we should have this policy?  We could just leave it as an admin setting only, but thus far I every admin setting has a policy.
- Have a naming opinion?

JohnMcPMS: Enable Windows Package Manager Configuration Processor Path

Enable Windows Package Manager Configuration Processor Override

Enable Windows Package Manager Custom Configuration Processor Path

Enable Custom Configuration Processor Path for Windows Package Manager

Enable Configuration Processor Override for Windows Package Manager

denelon: We have a bunch of policies under Desktop App Installer
image

The ones with "App Installer" in the name were from a time when the naming seemed to best align that way.
The newer ones have "Windows Package Manager" in their name.

I believe I would go with:

Enable Windows Package Manager Configuration Processor Override

When it is not configured, the elevated user can enable the setting.
When it is enabled, the user can use the setting (it's enabled).
When it is disabled, the user cannot use the setting (it's disabled, and the elevated user cannot enable the setting).

We know the elevated user can modify the GPO setting, so it's a best effort if the setting is disabled and the user can elevate.

florelis (approved):

Issue #3866, opened 2023-11-08
Ability to specify a default installer type in the manifest


mdanish-kh:
### Description of the new feature / enhancement

If two or more installers are equally suitable in terms of architecture, scope and other parameters etc, there should be a way to specify a default installer type for fresh install scenarios in the manifest.

### Example

As reported in https://github.com/microsoft/winget-pkgs/issues/125309, the manifest contains both portable and nullsoft installers that are equal in terms of arch, scope etc. In fresh install flow, winget selects the installer that is defined first in the manifest. Re-ordering the manifest would make it choose the other installer.

Steps to reproduce to verify that behavior would be:

1. Locally clone winget-pkgs and install Streetwriters.Notesnook
    `winget install -m manifests/s/Streetwriters/Notesnook/2.4.0`
	
	Notice it selects the portable installer (defined earlier in the manifest)    

2. Re-order the installer nodes
3. Install the manifest locally again. Notice it will select the nullsoft installer this time.

mdanish-kh:
[Policy] Area-Manifest

Trenly:
I'm wondering if there shouldn't be some sort of default preference built into the client. I know it wouldn't be everyone's first choice, but I would imagine that 90% of scenarios would be covered by an order of 
* MSIX / APPX
* MSI / Wix / Burn
* EXE / Nullsoft / Inno
* Portable

denelon:
I agree with @Trenly in terms of precedence/preference (assuming no preference/requirement exists in settings).

We have the "--installer-type" argument so a user can still get the "clean" install to be the installer type preferred, but I believe the default selection should match what was suggested.

PR #6123, opened 2026-04-02
Add default installer precedence if not defined by user

Trenly:
  - #3866 (addresses comment, does not resolve core issue)

This PR adds a default order of precedence for installer types if the user has not specified one. This ensures that ordering in a manifest is less important. While this has the potential to cause some upset with publishers who would prefer a different default order, the linked issue can remain open for tracking those requests.

Future work could involve merging the default preferences with user preferences so that any installer types the user has not specified as their preference still adhere to this default ordering, but I presumed that to be out of scope for this feature

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6123)


Trenly:
@jsoref - I'm not sure I understand why this error is coming up, when it didn't come up on PRs from a few days ago and nothing has changed in the workflow file. Any advice?

jsoref:
- This is tracked as https://github.com/check-spelling/check-spelling/issues/103 -- I've developed a fix based on this report and will try to release a version shortly.

Unfortunately the only choices are updating to the prerelease version (pinned by sha) or adding a guard that skips check-spelling for PRs where the base repo doesn't match the head repo (the check only applies in that case).

src/AppInstallerCommonCore/Manifest/ManifestComparator.cpp
```diff
@@ -228,6 +228,22 @@ namespace AppInstaller::Manifest
                 {
                     preference = Settings::User().Get<Settings::Setting::InstallerTypePreference>();
                     requirement = Settings::User().Get<Settings::Setting::InstallerTypeRequirement>();
+
+                    // Apply default precedence order when the user has not configured any installer type preferences or requirements.
+                    if (preference.empty() && requirement.empty())
```
JohnMcPMS, line 233: Feels like this should be set any time preferences are not provided.

Of if requirements are provided but preferences are not, then we set preferences to requirements by default.

Trenly: I thought it could be confusing if the requirements were copied into the preferences, unless that would be useful for dependency selection?

JohnMcPMS: Alternatively, we implicitly "copy" the preferences by using the requirements in `IsFirstBetter` when the preferences are empty.  That is probably cleaner than copying things in memory.

If preferences are empty, then you get manifest ordering (when other fields are equal of course) among the types in the requirements.  I don't think we want a case where the manifest has two types and you get one or the other based on whether you have requirements set to include both (unless it is using requirement ordering).

I feel like implicitly using the requirement ordering is the better usability option, and it feels natural to me to list my requirements in the order that I would prefer.

JohnMcPMS: @denelon for an opinion

Trenly: AH, I see now - Currently the requirements are just an applicability filter that work like "Any of these are valid" but the ordering is non-consequential.  By copying the requirements to the preferences (or using them in IsFirstBetter) it would allow for the ordering that the preferences give. 

That makes sense to me!!

denelon: I think the order I put my requirements or my preferences in should be treated as an ordered list.

Trenly: > I feel like implicitly using the requirement ordering is the better usability option, and it feels natural to me to list my requirements in the order that I would prefer.

Updated so that if requirements are present, they are used for the ordering and the preferences are the fallback

doc/Settings.md
```diff
@@ -137,6 +137,8 @@ The `archiveExtractionMethod` behavior affects how installer archives are extrac
 
 Some of the settings are duplicated under `preferences` and `requirements`. `preferences` affect how the various available options are sorted when choosing the one to act on.  For instance, the default scope of package installs is for the current user, but if that is not an option then a machine level installer will be chosen. `requirements` filter the options, potentially resulting in an empty list and a failure to install. In the previous example, a user scope requirement would result in no applicable installers and an error.
 
+When multiple values are listed under `requirements`, they are treated as an ordered preference in addition to a filter — the first listed value is preferred over subsequent ones when multiple valid options exist. If both `requirements` and `preferences` are set for the same field, the ordering from `requirements` takes precedence.
```
JohnMcPMS:
```suggestion
When multiple values are listed under `requirements`, they are treated as an ordered preference in addition to a filter — the first listed value is preferred over subsequent ones when multiple valid options exist. If both `requirements` and `preferences` are set for the same field, the ordering from `preferences` takes precedence.
```
Or I'm going to have another comment later...

doc/ReleaseNotes.md
```diff
@@ -47,6 +47,16 @@ The WinGet MCP server's existing tools have been extended with new parameters to
...
 
+### Default priority of installer types
+
+Installer type selection no longer depends on the order defined on the manifest. Instead, preference is given in this order:
+- MSIX
+- MSI / Wix / Burn
+- Nullsoft / Inno / EXE
+- Portable
+
+When a user configures installer type requirements or preferences, the order in which they are listed is now respected during installer selection. Requirements take precedence over preferences when both are configured.
```
JohnMcPMS:
```suggestion
When a user configures installer type requirements or preferences, the order in which they are listed is now respected during installer selection.
```
That other comment is seeming more likely...

src/AppInstallerCommonCore/Manifest/ManifestComparator.cpp
```diff
@@ -270,12 +286,15 @@ namespace AppInstaller::Manifest
 
             details::ComparisonResult IsFirstBetter(const ManifestInstaller& first, const ManifestInstaller& second) override
             {
-                if (m_preference.empty())
+                // Requirements take precedence over preferences for ordering; fall back to preferences when no requirements are set.
+                const auto& effectiveOrder = !m_requirement.empty() ? m_requirement : m_preference;
```
JohnMcPMS:
```suggestion
                // If no preferences are set, use requirement ordering instead.
                const auto& effectiveOrder = m_preference.empty() ? m_requirement : m_preference;
```
Since we have both, this preserves the previous behavior when both were set.  I also think it is the appropriate behavior when both are set, although it is obviously less work on the settings author to simply set the requirements in preference order.

JohnMcPMS (changes requested):

src/AppInstallerCommonCore/Manifest/ManifestComparator.cpp
Trenly: @JohnMcPMS - Wouldn't that conflict with the default precedence change though? Since the client is implicitly injecting preference order if preferences is empty?

JohnMcPMS: The default preferences are only set when both are empty per-the current change, which aligns with my request here.  I previously offered two options:
1. Always set default preferences if none provided
2. Use requirements as preferences if requirements provided without preferences

I think 2 is the better option and what my comments are intended to be pushing toward.

Trenly: Ah, thanks; Forgot it was only set if both were empty; Not enough ☕

JohnMcPMS (approved):

doc/ReleaseNotes.md
```diff
@@ -47,6 +47,16 @@ The WinGet MCP server's existing tools have been extended with new parameters to
 
 The PowerShell module now automatically uses `GH_TOKEN` or `GITHUB_TOKEN` environment variables to authenticate GitHub API requests. This significantly increases the GitHub API rate limit, preventing failures in CI/CD pipelines. Use `-Verbose` to see which token is being used.
 
+### Default priority of installer types
+
+Installer type selection no longer depends on the order defined on the manifest. Instead, preference is given in this order:
+- MSIX
```
beedell-roke, line 53: Does this include [`AppInstaller`](https://github.com/microsoft/winget-pkgs/issues/166364#issuecomment-2273966067), etcetera?

PR #6129, opened 2026-04-06
Dynamically select drive for test

Trenly:

Not all users have the same drive setup. For users with the `F:\` drive mounted, the `VerifyIsSameVolume` test fails when run locally, but will succeed on the pipelines as they don't have that specific drive mounted. This PR updates the test implementation to always select an unused drive letter. If all the drives happen to be mapped, the specific tests for casing will be skipped. This should ensure that the tests pass locally regardless of users drive mappings while still retaining the correct testing behavior

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6129)

### DIFF
--- a/src/AppInstallerCLITests/Filesystem.cpp
+++ b/src/AppInstallerCLITests/Filesystem.cpp
@@ -67,9 +67,9 @@ TEST_CASE("VerifyIsSameVolume", "[filesystem]")
     std::filesystem::path path3 = L"localPath\\test\\folder";
     std::filesystem::path path4 = L"test\\folder";
     std::filesystem::path path5 = L"D:\\test\\folder";
-	// path 6 dynamically generated below to be a non-existent drive letter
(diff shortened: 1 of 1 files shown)


src/AppInstallerCLITests/Filesystem.cpp
```diff
@@ -67,12 +67,33 @@ TEST_CASE("VerifyIsSameVolume", "[filesystem]")
...
+    {
+        if (!(driveMask & (1 << (c - L'A'))))
+        {
+            nonExistentDriveLetter = c;
+            break;
+        }
+    }
+
+    std::filesystem::path path6, path8;
+    if (nonExistentDriveLetter != L'\0')
```
JohnMcPMS: The test should just fail in the unlikely case that we didn't find a drive letter rather than not running these portions.

JohnMcPMS (approved):

Issue #6148, opened 2026-04-16
The explicit flag for source is translated like in "explicit content"


Fulgen301:
### Relevant area(s)

WinGet CLI

### Relevant command(s)

winget source

### Brief description of your issue

The german translation of `winget source list` translates "Explicit", aka whether a source should be implicitely used on only when the source is explicitely specified, as "Anstößig", which is the meaning of explicit like in explicit content.

```
Name           Argument                                                            Anstößig
-------------------------------------------------------------------------------------------
msstore        https://storeedgefd.dsx.mp.microsoft.com/v9.0                       false
winget         https://cdn.winget.microsoft.com/cache                              false
winget-font    https://cdn.winget.microsoft.com/fonts                              true
```

(I do hope the font repository doesn't have explicit content...)

### Steps to reproduce

Run `winget source list` on a german system.

### Expected behavior

"explicit" should be translated correctly.

### Actual behavior

"explicit" is not translated correctly.

### Environment

```shell
Windows-Paket-Manager v1.28.220
Copyright (c) Microsoft Corporation. Alle Rechte vorbehalten.

Windows: Windows.Desktop v10.0.26200.8246
Systemarchitektur: X64
Paket: Microsoft.DesktopAppInstaller v1.28.220.0
```

PR #6150, opened 2026-04-16
Add comments to loc strings

florelis:
Related to #6148. I asked Copilot to go through our strings to add comments to anything that was ambiguous or translated incorrectly in some language. Updating the comments does not trigger a re-translation, so we still need to submit a bug to the loc team.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6150)

### DIFF
--- a/doc/Developing.md
+++ b/doc/Developing.md
@@ -46,3 +46,31 @@ The unit tests are located inside the `AppInstallerCLITests` project. When the s
 
 > [!TIP]
 > If you just want to run a particular test, you can specify the test name as an argument to the executable. For example, `AppInstallerCLITests.exe EnsureSortedErrorList`.
+
+## Localization
+
+The English resource strings are the source of truth and live in:
+
+- `src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw`
+- `src/AppInstallerCLIPackage/Shared/Strings/en-us/Resources.resw`
+
+The localized strings under `Localization/Resources/<locale>/` are owned by Microsoft's internal localization team and are fetched automatically. **Do not edit files in `Localization/Resources/`**—any changes will be overwritten.
+
+### Adding or modifying resource strings
+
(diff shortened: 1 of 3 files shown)


doc/Developing.md
Trenly, line 1: The changes here feel like they would be good to have in the copilot instructions too

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -214,15 +214,19 @@
...
     <value>Disabled</value>
+    <comment>Status value in the 'winget features' table. Indicates an experimental feature is turned off. Paired with 'Enabled'.</comment>
   </data>
   <data name="FeaturesEnabled" xml:space="preserve">
     <value>Enabled</value>
+    <comment>Status value in the 'winget features' table. Indicates an experimental feature is active/turned on. Paired with 'Disabled'.</comment>
   </data>
   <data name="FeaturesFeature" xml:space="preserve">
     <value>Feature</value>
+    <comment>Column header in the 'winget features' output table. A 'feature' here is an experimental software capability. This is a noun, not a verb.</comment>
```
Trenly: Should this match the common phrasing below as `Do NOT translate as a verb` ?

JohnMcPMS:
> Updating the comments does not trigger a re-translation, so we still need to submit a bug to the loc team.

That is annoying; do you know if we change the symbol name and comment it would trigger?  As in, we update the comment and change it to `LocalizedStringResourceName2`.

florelis:
> > Updating the comments does not trigger a re-translation, so we still need to submit a bug to the loc team.
> 
> That is annoying; do you know if we change the symbol name and comment it would trigger? As in, we update the comment and change it to `LocalizedStringResourceName2`.

I'm not sure. I know in some cases if they see that a string is the same as another they may just use the same translation, but I don't know if they consider comments.

JohnMcPMS (approved):

PR #6156, opened 2026-04-18
Add policy for notifying authors of Localization Restriction

Trenly:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6156)

### DIFF
--- a/.github/policies/labelManagement.issueOpened.yml
+++ b/.github/policies/labelManagement.issueOpened.yml
@@ -24,7 +24,7 @@ configuration:
               pattern: Localization/Resources/*
           - not:
               activitySenderHasPermission: # Only trigger if the sender doesn't have write permissions,
-                permission: write          # to avoid adding the label for PRs from MSFT intentionally updating these files
+                permission: Write          # to avoid adding the label for PRs from MSFT intentionally updating these files
         then:
           - addLabel:
               label: Needs-Author-Feedback
--- a/.github/policies/labelManagement.issueOpened.yml
+++ b/.github/policies/labelManagement.issueOpened.yml
@@ -15,6 +15,36 @@ configuration:
               action: Opened
         then:
           - addCodeFlowLink
+      - description: Check for modifications of Localization files in new PRs
+        if:
+          - payloadType: Pull_Request
+          - isAction:
+              action: Opened
+          - filesMatchPattern:
+              pattern: Localization/Resources/*
+          - not:
+              activitySenderHasPermission: # Only trigger if the sender doesn't have write permissions,
+                permission: write          # to avoid adding the label for PRs from MSFT intentionally updating these files
+        then:
+          - addLabel:
+              label: Needs-Author-Feedback
+          - addLabel:
+              label: Blocking-Issue
+          - addReply:
+              reply: >-
+                Hello ${issueAuthor},
+
+
+                Thank you for your contribution! It looks like your PR includes changes to localization files.
+
+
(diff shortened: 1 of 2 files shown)


.github/policies/labelManagement.issueOpened.yml
```diff
@@ -15,6 +15,36 @@ configuration:
...
+      - description: Check for modifications of Localization files in new PRs
+        if:
+          - payloadType: Pull_Request
+          - isAction:
+              action: Opened
+          - filesMatchPattern:
+              pattern: Localization/Resources/*
+          - not:
+              activitySenderHasPermission: # Only trigger if the sender doesn't have write permissions,
+                permission: write          # to avoid adding the label for PRs from MSFT intentionally updating these files
```
mdanish-kh:
```suggestion
                permission: Write          # to avoid adding the label for PRs from MSFT intentionally updating these files
```

I'm not sure if this is a case insensitive check, but on winget-pkgs the value is capitalized

https://github.com/microsoft/winget-pkgs/blob/5f99c5de9a346d2f78336a35f705d1fe853a60a7/.github/policies/labelManagement.issueUpdated.yml#L400-L401

.github/policies/labelManagement.issueOpened.yml
```diff
@@ -15,6 +15,36 @@ configuration:
...
+                Hello ${issueAuthor},
+
+
+                Thank you for your contribution! It looks like your PR includes changes to localization files.
+
+
+                The English resource file [**/en-us/winget.resw](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw) is the only file contributors should edit for string changes. It feeds the Microsoft localization pipeline. The files under [Localization\Resources\<locale>\](https://github.com/microsoft/winget-cli/tree/master/Localization/Resources) are **automatically synced from Microsoft's internal localization system and must not be edited**. Any manual edits will be overwritten on the next sync.
+
+
+                Instead of this PR, it would be helpful if you could [open an issue](https://github.com/microsoft/winget-cli/issues/new?template=Documentation_Issue.yml) so the internal localization team can review it. For more information, please see the [Development Guide](https://github.com/microsoft/winget-cli/blob/master/doc/Developing.md#localization).
```
florelis: Maybe we should add an issue template for localization issues. It would be helpful to have things in the format we need to pass on to the localization team. Also, I'd change the wording from "so the internal localization team can review it" to something like "and we will forward it to the internal localization team for review". The localization team does not look at our GitHub issues, and I wouldn't want to create an expectation that there would be interaction from them here.

florelis: @denelon would you be okay with adding a template for translation issues?

Inputs we'd want for an issue template (basically copying the internal loc bug template):
* Language(s)
* Issue type: Incorrect Translation / Consistency / Gender / Grammar / Terminology / Style / Fluency / Other
* Source string (optional)
* Current translated string (optional)
* Suggested translated string (optional)
* Description / additional details
* How to find the bug
* Screenshots (optional)

Trenly: I added a basic template for now, happy to update it with additional information

denelon: Given we have to run through our localization process, and they may reject certain requests it might be good to let end users know that as well. @Trenly do you want me to approve this as is and do a separate PR to extend the template or do you want to add those other fields in this PR?

Trenly: @denelon - I'd prefer a separate PR just so we can at least get the policy working and tested, then introduce the friction of the additional template fields

denelon (approved):

Issue #2405, opened 2022-08-02
Manifests without installer info


lechacon:
### Description of the new feature / enhancement

Sometimes we can't or don't want to have a manifest for installing a package. For example if the publisher does not want winget to be a way to install, or the installer is behind auth.

But winget can do other things besides installing, like uninstalling or listing all installed apps by matching with ARP. What if we allowed manifests that did not have installer info to make these scenarios work?

Even if we don't do anything with a package, having it's ARP info for matching would be useful to prevent false matches with other manifests.

denelon:
Related to:
* #1900

Issue #1900, opened 2022-02-03
Package Tombstones


denelon:
### Description of the new feature / enhancement

Sometimes a package is temporarily not available, or the publisher has chosen not to be included in the Windows Package Manager Community App Repository.

A new Tombstone.YAML file would be used to support this behavior.

If the file is in the package directory, it could be used to hold identifiers to help prevent future submission rather than what would likely happen if there was no manifest for the package. It may contain SHA256 values for known versions to help inform users when they try to submit a manifest that the package has been tombstoned and cannot be submitted.

If the file is in a version directory, that version would no longer be available for search, install, or upgrade, but it would still allow users who have the package (as a dependency) so they could still install software that depended on that no longer available version of a package.

If a user were to try to install a package and the specific required version of the dependency had been tombstoned, the user would be informed that the dependency is no longer available.

Trenly:
Here is a list of some use cases for Tombstones, as well as their associated issues/pr's

| Issue # | Use Case |
| --- | --- |
| microsoft/winget-pkgs#42432 | This package has a Dynamic CDN which means that the Sha256 of the installer is different depending on who is downloading it and when. Tombstoning this package with no version available would allow for anyone opening a PR for this package to be informed as to why the package will not be added, and to save resource time by not running the full set of pipeline scans on the manifest. |
| microsoft/winget-pkgs#41777 | This package has versions existing in the repository. However, newer versions of the package are unable to pass the AntiVirus scans. Tombstoning this package for specific versions would allow anyone submitting a version that is known to fail the AVScan to be informed, and to save resource time. This would still allow for new versions, which are not tombstoned, to be submitted and scanned, at which point a tombstone can be added for the new versions if necessary. |
| microsoft/winget-pkgs#16896 | Several different versions of this package have been submitted to the community repository, and all fail due to being blocked by dependencies. This may be due to a number of factors such as users not searching for closed PR's, thinking newer versions have no dependency requirement, or some other line of reasoning. Having the ability to tombstone packages would allow for the ability to avoid duplicate PR's and would create a better experience for the user by keeping them informed. |
| microsoft/winget-pkgs#45359 | When packages are no longer available, their manifests are removed. However, it is possible in the case that a lapsed domain or other error may be transient. Having a tombstone would allow for the package metadata to be retained while removing the package from search results. This would allow for a package version to be restored more easily if a transient error is resolved. |
[...]

Issue #823, opened 2021-03-26
Support for filtering packages on installer method


denelon:
Several packages do not support or provide a silent install experience.

Assuming package manifests identify support for interactive installers, the client should be able to filter these packages so they aren't inadvertently executed for certain scenarios. The client should be able to support a preference for silent install experiences if specified. Additionally the client should be able to filter search results so that packages requiring interactive install can be omitted.

This could be implemented in settings, as well as via command-line arguments.

denelon:
This feature is required prior to accepting "interactive only" installers into the community repository. Many users are looking for headless/unattended import capabilities.

Trenly:
[Policy] Area-Settings

upintheairsheep:
The fact dependency support was added before this simple issue is kinda weird.

luzpaz:
Any traction here ? (asking due to #107357 being closed)

SkylerWallace:
Would love to see this get looked at again. Requiring silent installation is a blocker for many programs. Notable ones include HP Support Assistant and Samsung Magician.

Add a boolean indicator in the manifest like **InteractiveInstallOnly**.
Add a search filter to **winget search** that only lists installers that support silent install like **--supports-silent**.

Dragon1573:
Currently we have `InstallModes`property in both schema `v1.6.0` and `v1.7.0`. Changing this to a required property would help. Filter those who only contains `interactive`.

Stensel8:
Would be nice to get this implemented.

I made a PR to get a specific game mod into Winget, however if that game is not detected on the system, the mod installer would ask where to install it.

And there seems to be no way I can do that in Winget. 

See: https://github.com/microsoft/winget-pkgs/pull/357446

PR #6157, opened 2026-04-18
Add support for Manifests with No Installers

Trenly:
  - #2405
  - #1900

## Summary
 
 This PR introduces support for the `noinstaller` installer type (manifest version 1.30.0), which allows publishers to declare packages that have no downloadable installer. This enables WinGet to correlate against already-installed software even when no installation artifact is available — useful for system components, pre-installed software, and packages distributed through channels outside of WinGet. This also acts as a way to notify clients that a version has been removed for a specific reason, a package has been removed for a specific reason (CVEs, publisher request, moved to a new identifier, etc.), while still providing tools over at winget-pkgs like the duplicates checker a way to enforce removed packages by SHA.
 
 ## What's New
 
 ### NoInstaller Installer Type
 Publishers can now declare a manifest with `InstallerType: noinstaller` to indicate that a package exists and is recognized by WinGet, but has no installer to download or run. WinGet will correctly handle these packages across the `install`, `show`, and `upgrade` workflows.
 
 ### Installer Availability Message
 A new optional manifest field, `InstallerAvailabilityMessage`, allows publishers to provide a human-readable explanation when an installer is unavailable. This message is surfaced to the user in place of the generic fallback message, giving them actionable context (e.g., a link or explanation of where to obtain the software).
 
## This change may require changes to the validation framework, which is why the PR is being created as draft

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6157)


Trenly:
Tagging @hackean-msft as I'm not sure if the install validation would handle this cleanly currently or not; I believe it may be likely that the validation pipeline would need to skip installer validation for `noinstaller`

JohnMcPMS:
> manifest type

***installer*** type.  At first I thought you were introducing a whole new manifest file type, which seemed overkill.

This is definitely a design level thing before I'm going to look too deeply at 45 files 😄 

I could imagine @denelon wanting to consider this a lot more.  Thinking things like "We shouldn't show uninstallable packages in search results", which would require index work (he already wants a bitrack that I will have to defend every bit on to save bandwidth).

Trenly:
> > manifest type
> 
> _**installer**_ type. At first I thought you were introducing a whole new manifest file type, which seemed overkill.
> 
> This is definitely a design level thing before I'm going to look too deeply at 45 files 😄
> 
> I could imagine @denelon wanting to consider this a lot more. Thinking things like "We shouldn't show uninstallable packages in search results", which would require index work (he already wants a bitrack that I will have to defend every bit on to save bandwidth).

Yes, Installer type :)

Definitely need to consider design first which is why I created this as draft to discuss before I ask anyone to review this beast

I know there is a separate isse out there (#823) for filtering on installer method / installer type, so thought the work for hiding from search would fit in well with that work if / when we get to it - and I purposely made it so a custom message could be specified so that if a user does get an un-installable package (say for example, it was moved to a different package identifier, or removed by publisher request) then there would be a nice clean message. Looking forward to denelon's feedback :)

denelon:
This is an approach I hadn't considered before. 

On the surface I like it because it's not a new "filetype" requiring some new schema.

It might be strange to require an InstallerURL for something that shouldn't be installed so we might need to think about that field, but preserving things like the SHA256, product code, upgrade code is useful for correlation. I'm just not sure about reasoning over a PR to modify something with a known installer type. I'm also not sure we would accept PRs with this installer type unilaterally. There are lots of implications to reason over.

Trenly:
@denelon - this is buit in a way that wouldn’t require a URL; in fact, it prohibits a URL.

Trenly:
> I'm also not sure we would accept PRs with this installer type unilaterally. 

Thinking through this, validation could add Validation-No-Executables and be waivered, forcing MSFT review for each instance

sredna:
> I'm also not sure we would accept PRs with this installer type unilaterally.

This is pretty important, this feature should not be used by ISVs to block inclusion in WinGet "for corporate reasons" IMHO.

Other than in-box Windows components and maybe hardware specific OEM drivers/software, I don't see many other uses. WinGet already knows how to deal with `SystemComponent=1` in the ARP key.

denelon:
@JohnMcPMS is correct 🤣

This kind of a mechanism does seem like a reasonably interesting way to implement tombstones, but there certainly are implications around not wanting these to show up as "installable" or "upgradeable" versions of packages. I think there may also be several other considerations to be made around toggling between whether a version can or cannot be installed, and what happens when multiple variations (like architecture) with independent versioning show up.
* version 1.1 x64 noinstaller
* version 1.1 arm64 <supported installer>

I expect we're going to need a specification... I'll gladly defer to @JohnMcPMS on implementation detail preferences, but I think there are lots of edge cases to reason over. I do however feel like this will unblock a few things I'd like to see supported in WinGet.

Trenly:
> @JohnMcPMS is correct 🤣
> 
> This kind of a mechanism does seem like a reasonably interesting way to implement tombstones, but there certainly are implications around not wanting these to show up as "installable" or "upgradeable" versions of packages. I think there may also be several other considerations to be made around toggling between whether a version can or cannot be installed, and what happens when multiple variations (like architecture) with independent versioning show up.
> 
> * version 1.1 x64 noinstaller
> * version 1.1 arm64
> 
> I expect we're going to need a specification... I'll gladly defer to @JohnMcPMS on implementation detail preferences, but I think there are lots of edge cases to reason over. I do however feel like this will unblock a few things I'd like to see supported in WinGet.

Happy to write an accompanying specification - some of these questions are likely already covered by this implementation, and some I'd imagine to be deferred to other implementations.

PR #6165, opened 2026-04-21
Don't require --source when partial source failure leaves a single result

### LINKED ISSUE #6164: winget install requires --source even when only one result exists from working sources
### Relevant area(s)

WinGet CLI

### Relevant command(s)

winget install

When you have multiple sources configured and one of them is unreachable (e.g. a corporate feed that's offline, or a source that's temporarily down), `winget install <package>` will fail with something like:

```
The following packages were found in the working sources.
Please specify one of them using the --source option to continue.
```
...even if only a single package was found. There's nothing ambiguous about the result: one package, one source, but winget still refuses to proceed and forces you to re-run the command with an explicit `--source` flag.

This is pretty annoying in practice. If I have a private source that's down and I want to install something from the public winget source, I shouldn't need to know upfront that my other source is broken and manually work around it.

The right behavior should be: if exactly one result comes back from the sources that are working, just use it. The `--source` prompt makes sense when there are actually multiple ambiguous results to choose from, not when the answer is obvious.

### Steps to reproduce

Add a second source that is unreachable (or take one offline temporarily)
Run winget install <package> where the package exists on one of the working sources
Observe the error requiring --source despite there being a single unambiguous result

### Expected behavior

When there is only one source available, winget shouldn't ask me to specify the --source flag, this is redundant.

### Actual behavior

Winget returns this kind of error:
```
The following packages were found in the working sources.
Please specify one of them using the --source option to continue.
```

### Environment

```shell
Gestionnaire de package Windows v1.9.25200
Copyright (c) Microsoft Corporation. Tous droits réservés.

Windows: Windows.Desktop v10.0.26200.8246
Architecture du système : X64
Package : Microsoft.DesktopAppInstaller v1.24.25200.0

Répertoires Winget
Journaux                                     %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\Local…
Paramètres utilisateurs                      %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\Local…
Répertoire des liens portables (utilisateur) %LOCALAPPDATA%\Microsoft\WinGet\Links
Répertoire des liens portables (ordinateur)  C:\Program Files\WinGet\Links
Racine de package portable (utilisateur)     %LOCALAPPDATA%\Microsoft\WinGet\Packages
Racine du package portable                   C:\Program Files\WinGet\Packages
Racine de package portable (x86)             C:\Program Files (x86)\WinGet\Packages
Téléchargements du programme d’installation  %USERPROFILE%\Downloads

Liens
---------------------------------------------------------------------------------------------
Déclaration de confidentialité        https://aka.ms/winget-privacy
Contrat de licence                    https://aka.ms/winget-license
Avis de tiers                         https://aka.ms/winget-3rdPartyNotice
Page d’accueil                        https://aka.ms/winget
Conditions générales du Windows Store https://www.microsoft.com/en-us/storedocs/terms-of-sale

Paramètre administrateur                  État
---------------------------------------------------
LocalManifestFiles                        Désactivé
BypassCertificatePinningForMicrosoftStore Désactivé
InstallerHashOverride                     Désactivé
LocalArchiveMalwareScanOverride           Désactivé
ProxyCommandLineOptions                   Désactivé
DefaultProxy                              Désactivé
```

Sinan-Karakaya:
Fixes #6164

Right now, when one or more sources fail during an install search, winget shows the results from the sources that did respond and asks you to pick one via `--source` even if there's only one result. This happens in `HandleSearchResultFailures` via the `ShowSearchResultsOnPartialFailure` flag, which is set exclusively by the install command.

The fix is small: before routing into the "error and stop" path, check whether we're in install mode with exactly one match. If so, fall into the same "warn and continue" path that `TreatSourceFailuresAsWarning` already uses, the failed sources still get surfaced as warnings, but the install proceeds.

Existing behavior is preserved for:
- Zero results from working sources -> still a hard error
- Two or more results -> still shows the ambiguous list and requires `--source`
- All other commands (search, upgrade, etc.) -> unaffected, they don't set `ShowSearchResultsOnPartialFailure`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6165)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/PromptFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/PromptFlow.cpp
@@ -11,35 +11,36 @@ using namespace AppInstaller::Utility::literals;
 
 namespace AppInstaller::CLI::Workflow
 {
+    bool IsInteractivityAllowed(Execution::Context& context)
+    {
+        // Interactivity can be disabled for several reasons:
+        //   * We are running in a non-interactive context (e.g., COM call)
+        //   * It is disabled in the settings
+        //   * It was disabled from the command line
+
+        if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::DisableInteractivity))
+        {
+            AICLI_LOG(CLI, Verbose, << "Skipping prompt. Interactivity is disabled due to non-interactive context.");
+            return false;
+        }
+
+        if (context.Args.Contains(Execution::Args::Type::DisableInteractivity))
+        {
+            AICLI_LOG(CLI, Verbose, << "Skipping prompt. Interactivity is disabled by command line argument.");
+            return false;
+        }
+
+        if (Settings::User().Get<Settings::Setting::InteractivityDisable>())
+        {
+            AICLI_LOG(CLI, Verbose, << "Skipping prompt. Interactivity is disabled in settings.");
+            return false;
+        }
+
+        return true;
+    }
+
     namespace
     {
-        bool IsInteractivityAllowed(Execution::Context& context)
-        {
-            // Interactivity can be disabled for several reasons:
-            //   * We are running in a non-interactive context (e.g., COM call)
-            //   * It is disabled in the settings
-            //   * It was disabled from the command line
-
-            if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::DisableInteractivity))
-            {
-                AICLI_LOG(CLI, Verbose, << "Skipping prompt. Interactivity is disabled due to non-interactive context.");
-                return false;
-            }
-
-            if (context.Args.Contains(Execution::Args::Type::DisableInteractivity))
-            {
-                AICLI_LOG(CLI, Verbose, << "Skipping prompt. Interactivity is disabled by command line argument.");
-                return false;
-            }
-
-            if (Settings::User().Get<Settings::Setting::InteractivityDisable>())
-            {
-                AICLI_LOG(CLI, Verbose, << "Skipping prompt. Interactivity is disabled in settings.");
-                return false;
-            }
-
-            return true;
-        }
 
         bool HandleSourceAgreementsForOneSource(Execution::Context& context, const Repository::Source& source)
         {
--- a/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
+++ b/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
@@ -973,19 +973,45 @@ namespace AppInstaller::CLI::Workflow
 
         if (!searchResult.Failures.empty())
         {
-            // When the install command finds exactly one result from working sources, auto-proceed
-            // instead of requiring the user to specify --source.
-            bool singleResultOnPartialFailure =
-                WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::ShowSearchResultsOnPartialFailure) &&
-                searchResult.Matches.size() == 1;
-
-            if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::TreatSourceFailuresAsWarning) || singleResultOnPartialFailure)
+            if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::TreatSourceFailuresAsWarning))
             {
                 auto warn = context.Reporter.Warn();
                 for (const auto& failure : searchResult.Failures)
                 {
                     warn << Resource::String::SearchFailureWarning(Utility::LocIndView{ failure.SourceName }) << std::endl;
                 }
+            }
+            // When ShowSearchResultsOnPartialFailure is set (e.g. install) and exactly one match
+            // was found, prompt the user interactively rather than failing outright.
+            // Falls back to the hard error path if interactivity is disabled.
+            else if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::ShowSearchResultsOnPartialFailure) &&
+                     searchResult.Matches.size() == 1 &&
+                     IsInteractivityAllowed(context))
+            {
+                {
+                    auto warn = context.Reporter.Warn();
+                    for (const auto& failure : searchResult.Failures)
+                    {
+                        warn << Resource::String::SearchFailureWarning(Utility::LocIndView{ failure.SourceName }) << std::endl;
+                    }
+                }
+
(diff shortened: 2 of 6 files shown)


denelon:
I'm not sure that's the behavior we want.

Given multiple sources configured in WinGet and a "search sequence" if two sources would both return "one best match" the user should disambiguate, or source priority should "break the tie".

I'd be more in favor of an interactive prompt with the one best match and the source(s) with failures.

That also begs the question about what the default behavior should be when interactivity is disabled.

JohnMcPMS:
If there is an error searching one or more sources, we don't know if there is really only one result.  That is why there is a hard error and the user must explicitly choose the target source.

Sinan-Karakaya:
Thanks for the feedback, all fair points. I've reworked the approach based on the discussion:

Instead of silently auto-proceeding, the updated behavior now shows a `[Y] Yes / [N] No` prompt when `ShowSearchResultsOnPartialFailure` is set and exactly one match was found, so the user stays in control rather than having winget make the decision for them. The failures are still surfaced as warnings before the prompt, so the user knows what failed before deciding.

For the non-interactive case (COM callers, `--disable-interactivity`, the user setting), it falls straight through to the original hard error path. This way there are no prompt, no silent proceed, and it fails as expected.

On the "we don't know if there's really only one result" point: agreed, the prompt doesn't change that. But it shifts responsibility to the user: they're explicitly acknowledging they want to proceed despite the uncertainty, rather than winget guessing on their behalf. If they want certainty, they can still `--source` manually.

The comment was also updated to reference `ShowSearchResultsOnPartialFailure` directly instead of saying "install command", since that was both inaccurate and brittle. And I've added unit tests covering the three paths: interactive accept, interactive decline, and non-interactive (hard error).

The source priority / hunt sequence tie-breaking is a separate and more involved problem. I'd be happy to track that as a follow-up if that's useful, but I don't think it blocks this.

Sinan-Karakaya:

Trenly:
Maybe we need to re-think the user settings for interactivity a bit @denelon ?  I could see a case where a user would want to specify the default behavior for each interaction

For example, disable interactivity and don't accept any source agreements, but still proceed if there were source failures. Proceed through prompt for an install path with a user-defined default but hard stop on configuration review / continue prompt

Probably a bit outside the scope of this PR, but just something it brought to mind

denelon:
I created:
* https://github.com/microsoft/winget-cli/issues/6168

to address the tie-breaking scenario so this PR can be reviewed.

src/AppInstallerCLICore/Workflows/PromptFlow.cpp
```diff
@@ -11,35 +11,36 @@ using namespace AppInstaller::Utility::literals;
 
 namespace AppInstaller::CLI::Workflow
 {
-    namespace
+    bool IsInteractivityAllowed(Execution::Context& context)
```
JohnMcPMS, line 14: I don't want this to be publicly visible; callers should be required to go through our explicitly defined prompt flows rather than making arbitrary choices about handling interactivity.  That also makes the call sites less cluttered, as they are forced to simply invoke a prompt function with a default rather than have control flow.

I think that the correct approach is going to be refactoring the `Reporter` prompt methods out into `PromptFlow.cpp` and have them invoke this function in addition to other checks they do to determine whether to actually prompt or not.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -981,6 +981,38 @@ namespace AppInstaller::CLI::Workflow
                     warn << Resource::String::SearchFailureWarning(Utility::LocIndView{ failure.SourceName }) << std::endl;
                 }
             }
+            // When ShowSearchResultsOnPartialFailure is set (e.g. install) and exactly one match
+            // was found, prompt the user interactively rather than failing outright.
+            // Falls back to the hard error path if interactivity is disabled.
+            else if (WI_IsFlagSet(context.GetFlags(), Execution::ContextFlag::ShowSearchResultsOnPartialFailure) &&
```
JohnMcPMS, line 987: This block duplicates a lot of code from the other blocks.  I would rather we just accept the error output and move the prompt invocation into the correct location in the existing `else`.

If a warning about each source is a critical change to you, the entire function could probably be refactored to better make sematic decisions up front and then execute on those using a more common code path:

1. Are we using warnings or errors? Save that choice, select output stream and lambda body to match.
2. Enumerate failures, invoke output lambda, extract HRESULT (even if warning).
3. If we decided this is a warning, return.
4. Continue on with the second half of the existing `else` with the new prompt.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1332,6 +1332,10 @@ Do you agree to the terms?</value>
 Please specify one of them using the --source option to proceed.</value>
     <comment>{Locked="--source"} "working sources" as in "sources that are working correctly"</comment>
   </data>
+  <data name="SearchFailureSingleResultPrompt" xml:space="preserve">
+    <value>A package was found among the working sources. Would you like to proceed?</value>
+    <comment>"working sources" as in "sources that are working correctly"</comment>
```
JohnMcPMS, line 1337:
```suggestion
    <comment>"working sources" as in "package sources that are functioning properly"</comment>
```

src/AppInstallerCLITests/WorkFlow.cpp
```diff
@@ -187,3 +188,64 @@ TEST_CASE("Export_Settings", "[Settings][workflow]")
         REQUIRE(userSettingsFileValue.find("settings.json") != std::string::npos);
     }
 }
+
+TEST_CASE("HandleSearchResultFailures_SingleMatchWithPartialFailure", "[HandleSearchResultFailures][workflow]")
+{
+    auto makeSearchResult = []()
```
JohnMcPMS, line 194: nit: `catch2` runs the entire test multiple times, dynamically enabling sections each time to get the complete set.  You can just build the result in a local in the test rather than using a lambda.  Mostly saves a few lines of code.

Karl-WE:
another aspect is that sources can have different installers. e.g. PowerShell MSI vs MSIX - at least it was like this. There could be more recent examples. This would not be in favor of the idea for this request, would it? @sinan-karakaya

PR #6169, opened 2026-04-22
Improved manifest validations for MSI and Windows Feature names

JohnMcPMS:
## Change
Adds some additional validation for MSI switches and Windows Feature names:

1. MSI switches are checked during full manifest validation in the same way that they are checked when we attempt to use the MSI APIs rather than msiexec.  The API is used for fully all silent installs of MSIs, so we now ensure that those will work (6 total manifests in winget-pkgs were found that needed fixes).
2. Windows Feature names in dependencies are validated to be { alphanumeric, `-`, `_` }.  This is done both during full manifest validation and at runtime.

## Validation
A PowerShell script (tools/ManifestValidation/Invoke-ManifestValidation.ps1) is added that will run `wingetdev validate` against a directory recursively and generate a report.  There is support for resuming in the middle of the run.  This would probably perform much better if it were updated to use the WinGetUtil API, but that can be a future project.

The script was run against the entire winget-pkgs repository to ensure no significant impact, and the few manifests that were found will be fixed.

Unit tests are added.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6169)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -7,6 +7,7 @@ ACCESSDENIED
 ACCESSTOKEN
 acl
 adjacents
+adminproperties
 adml
 admx
 AFAIK
@@ -98,6 +99,7 @@ COMGLB
 commandline
 compressapi
 concurrencysal
+Consolas
 constexpr
 contactsupport
 contentfiles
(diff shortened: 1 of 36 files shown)


Trenly:
@johnmcpms - I thought most of that spelling was already fixed; is your branch behind master?

florelis (approved):

PR #6171, opened 2026-04-22
Optimize packaged temp ACL handling without weakening ACL repairs

### LINKED ISSUE #6162: Packaged temp path ACL reapplication on %TEMP%\WinGet causes recursive security traversal and severe slowdown
When running in packaged context, temp path initialization appears to reapply a protected inheritable ACL on `%TEMP%\WinGet` every time the temp path is resolved. On a large existing cache tree, that single ACL application causes Windows to recursively traverse and reprocess security on a very large subtree, which shows up in ProcMon as a large burst of metadata and security events and adds significant latency.

This does not appear to be caused by a semantic difference in the final ACL between packaged and unpackaged paths. The resulting ACL shape is effectively the same. The cost appears to come from applying it at different roots:

- packaged temp root: `%TEMP%\WinGet`
- unpackaged temp root: `%TEMP%\WinGet\defaultState`

## Observed Behavior

In ProcMon, the fast and slow traces look qualitatively similar. The slow case does not look like a different logical workflow. It looks like the same security-heavy workflow repeated over a much larger subtree.

Common event patterns in both traces:

- repeated `QueryDirectory`
- repeated `QueryBasicInformationFile`
- repeated `QueryStandardInformationFile`
- repeated `QueryNameInformationFile`
- repeated `Owner, DACL, DACL Protected`
- repeated `NO MORE FILES`
- handles opened with security-write style access consistent with ACL application

What stands out is scale, not sequence.

## Relevant Code Paths

Temp path retrieval goes through `src/AppInstallerSharedLib/Public/winget/Filesystem.h`, which always calls `InitializeAndGetPathTo(...)`.

`InitializeAndGetPathTo(...)` in `src/AppInstallerSharedLib/Filesystem.cpp` creates the directory and then unconditionally calls `ApplyACL()` when ACL metadata exists.

`ApplyACL()` builds explicit ACEs with inheritable flags and applies a protected DACL via `SetNamedSecurityInfoW`.

Packaged vs unpackaged temp roots are selected in `src/AppInstallerCommonCore/Runtime.cpp`:

- packaged temp root `%TEMP%\WinGet`
- unpackaged temp root `%TEMP%\WinGet\defaultState`

Cache paths resolve through temp in `src/AppInstallerCommonCore/FileCache.cpp`, so even cache path resolution can trigger temp-root ACL application before any actual file read/write occurs.

## Why This Looks Like ACL Propagation and Not Cache Enumeration

The cache code itself is path-based and direct. It does not intentionally walk the whole manifest/package tree just to service a lookup.

The broad ProcMon activity is therefore more consistent with Windows recursively reprocessing inherited security after one root `SetNamedSecurityInfoW` call than with winget logically enumerating all cached manifests/packages.

## Measured Tree Sizes

Current tree sizes on the test machine:

```text
PATH=C:\Users\mamoreau\AppData\Local\Temp\WinGet
EXISTS=True
DIRS=56171
FILES=673

PATH=C:\Users\mamoreau\AppData\Local\Temp\WinGet\defaultState
EXISTS=True
DIRS=704
FILES=310
```

The directory count ratio is about `79.8x`.

Additional subtree counts gathered during investigation:

```text
C:\Users\mamoreau\AppData\Local\Temp\WinGet\cache\V2_M\...\manifests
dirs=31293 files=170

C:\Users\mamoreau\AppData\Local\Temp\WinGet\cache\V2_PVD\...\packages
dirs=24122 files=181

C:\Users\mamoreau\AppData\Local\Temp\WinGet\defaultState\cache\V2_M\...\manifests
dirs=415 files=141

C:\Users\mamoreau\AppData\Local\Temp\WinGet\defaultState\cache\V2_PVD\...\packages
dirs=282 files=141
```

## Independent Reproduction Outside winget

To isolate the effect from winget logic, I copied `%TEMP%\WinGet` to `%TEMP%\WinGet2` and used a PowerShell P/Invoke repro that mirrors the packaged temp ACL call shape:

- owner = current user
- explicit ACEs for current user, Builtin Administrators, and Local System
- `SET_ACCESS`
- inheritance = `OBJECT_INHERIT_ACE | CONTAINER_INHERIT_ACE`
- security info = `OWNER_SECURITY_INFORMATION | DACL_SECURITY_INFORMATION | PROTECTED_DACL_SECURITY_INFORMATION`
- retry without owner if owner already matches current user

Measured results:

```text
TARGET=C:\Users\mamoreau\AppData\Local\Temp\WinGet2
EXISTS=True
EXIT=SUCCESS
ELAPSED_MS=23788.601
OWNER=DEVOLUTIONS\mamoreau
SDDL=O:S-1-5-21-...G:DUD:PAI(A;OICIIO;GA;;;SY)(A;;FA;;;SY)(A;;FA;;;BA)(A;OICIIO;GA;;;BA)(A;OICIIO;GA;;;S-1-5-21-...)(A;;FA;;;S-1-5-21-...)

TARGET=C:\Users\mamoreau\AppData\Local\Temp\WinGet2\defaultState
EXISTS=True
EXIT=SUCCESS
ELAPSED_MS=298.577
OWNER=DEVOLUTIONS\mamoreau
SDDL=O:S-1-5-21-...G:DUD:PAI(A;OICIIO;GA;;;SY)(A;;FA;;;SY)(A;;FA;;;BA)(A;OICIIO;GA;;;BA)(A;OICIIO;GA;;;S-1-5-21-...)(A;;FA;;;S-1-5-21-...)
```

Comparison:

```text
RootMs:23788.601
DefaultStateMs:298.577
Ratio:79.673
```

The resulting owner and SDDL were effectively the same. The time delta tracks the subtree size delta very closely.

This strongly suggests the slowdown is caused by applying the same protected inheritable ACL over a much larger subtree, not by a difference in final ACL semantics.

## Key Conclusion

A single packaged-style `SetNamedSecurityInfoW` call on `%TEMP%\WinGet` is sufficient to reproduce the same class of slowdown outside of winget itself. That makes the issue primarily about ACL application scope and unconditional reapplication, not source search or cache layout logic.

## Likely Root Cause

`InitializeAndGetPathTo()` currently re-applies ACLs whenever a path has ACL metadata. There does not appear to be:

- a short-circuit that checks whether owner/DACL/protection already match the desired state
- a process-level memoization that avoids reapplying the ACL repeatedly for the same temp root

As a result, resolving `PathName::Temp` in packaged mode can repeatedly trigger expensive recursive security propagation over `%TEMP%\WinGet`.

## Possible Fix Directions

- Avoid applying ACLs unconditionally on every temp path resolution.
- Add a fast short-circuit to skip `SetNamedSecurityInfoW` when owner/DACL/protection already match the desired state.
- Memoize successful ACL application per process for the temp root, if correctness allows it.
- Narrow ACL application scope so packaged mode does not target the large shared `%TEMP%\WinGet` root when a smaller subtree would suffice.

## Why This Matters

The issue scales with cache size. On machines with large `%TEMP%\WinGet` trees, a normal packaged operation that merely resolves a temp path can incur large, repeated ACL traversal cost and generate very large ProcMon traces, even when the logical file operation itself is small.

mamoreau-devolutions:

-----

Fixes #6162

## Summary

This change reduces the severe slowdown caused by recursive ACL propagation on packaged temp paths while preserving the security behavior that repairs improperly secured directories before use.

The fix keeps the existing fail-closed model, but avoids calling `ApplyACL()` when the target directory is already in the exact secure state winget expects.

## Changes

- add a fail-closed ACL inspection path before `ApplyACL()`
- only skip ACL reapplication when the current owner, DACL protection state, and effective permissions match the expected secure state
- continue to reapply ACLs on any mismatch, unreadable security descriptor, unknown principal, or unsupported ACL shape
- scope packaged temp paths under the runtime state name so any required repair stays bounded to the active winget-managed subtree instead of the full shared `%TEMP%\WinGet` tree
- add runtime coverage for the steady-state secure case and for mismatch cases that must still force ACL repair

## Why this is safe

The security intent of the original code is preserved: winget still repairs attacker-created or incorrectly ACL'd directories before using them.

This PR only removes unnecessary recursive ACL work in the steady state where the directory is already secure. If the ACL state cannot be verified confidently, winget keeps the existing repair behavior.

## Validation

- added runtime tests covering ACL skip/reapply behavior
- added coverage for packaged temp path behavior
- locally validated that steady-state temp-path resolution no longer pays the recursive ACL propagation cost when the directory is already secure
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6171)

### DIFF
--- a/src/AppInstallerCLITests/Runtime.cpp
+++ b/src/AppInstallerCLITests/Runtime.cpp
@@ -390,18 +390,6 @@ TEST_CASE("ShouldApplyACL_TrueWhenUnexpectedAceExists", "[runtime]")
     REQUIRE(details.ShouldApplyACL());
 }
 
-TEST_CASE("SetRuntimePathStateName_AffectsTempPath", "[runtime]")
-{
-    constexpr std::string_view stateName = "Issue6162State";
-    auto restoreStateName = wil::scope_exit([]() { Runtime::SetRuntimePathStateName("defaultState"); });
-
-    Runtime::SetRuntimePathStateName(std::string{ stateName });
-
-    PathDetails details = Runtime::GetPathDetailsFor(Runtime::PathName::Temp, true);
-
-    REQUIRE(details.Path.filename() == Utility::ConvertToUTF16(stateName));
-}
-
 TEST_CASE("VerifyDevModeEnabledCheck", "[runtime]")
 {
     if (!Runtime::IsRunningAsAdmin())
--- a/src/AppInstallerCLITests/Runtime.cpp
+++ b/src/AppInstallerCLITests/Runtime.cpp
@@ -28,6 +28,204 @@ void RequireAdminOwner(const std::filesystem::path& directory)
     REQUIRE(EqualSid(adminSID.get(), ownerSID));
 }
 
+DWORD AccessMaskFrom(ACEPermissions permissions)
+{
+    DWORD result = 0;
+
+    if (permissions == ACEPermissions::All)
+    {
+        result |= GENERIC_ALL;
+    }
+    else
+    {
+        if (WI_IsFlagSet(permissions, ACEPermissions::Read))
+        {
+            result |= GENERIC_READ;
+        }
+
+        if (WI_IsFlagSet(permissions, ACEPermissions::Write))
+        {
+            result |= GENERIC_WRITE | FILE_DELETE_CHILD;
+        }
+
+        if (WI_IsFlagSet(permissions, ACEPermissions::Execute))
+        {
+            result |= GENERIC_EXECUTE;
+        }
+    }
+
+    return result;
+}
+
+DWORD NormalizedAccessMaskFrom(ACEPermissions permissions)
+{
+    DWORD result = AccessMaskFrom(permissions);
+    GENERIC_MAPPING genericMapping
+    {
+        FILE_GENERIC_READ,
+        FILE_GENERIC_WRITE,
+        FILE_GENERIC_EXECUTE,
+        FILE_ALL_ACCESS,
+    };
+
+    MapGenericMask(&result, &genericMapping);
+    return result;
+}
+
+void ApplyAclForTest(const std::filesystem::path& directory, const std::optional<ACEPrincipal>& owner, const std::map<ACEPrincipal, ACEPermissions>& acl, bool protectDacl = true)
+{
+    auto userToken = wil::get_token_information<TOKEN_USER>();
+    auto adminSID = wil::make_static_sid(SECURITY_NT_AUTHORITY, SECURITY_BUILTIN_DOMAIN_RID, DOMAIN_ALIAS_RID_ADMINS);
+    auto systemSID = wil::make_static_sid(SECURITY_NT_AUTHORITY, SECURITY_LOCAL_SYSTEM_RID);
+    PSID ownerSID = nullptr;
+
+    struct ACEDetails
+    {
+        ACEPrincipal Principal;
+        PSID SID;
+        TRUSTEE_TYPE TrusteeType;
+    };
+
+    ACEDetails aceDetails[] =
+    {
+        { ACEPrincipal::CurrentUser, userToken->User.Sid, TRUSTEE_IS_USER },
+        { ACEPrincipal::Admins, adminSID.get(), TRUSTEE_IS_WELL_KNOWN_GROUP },
+        { ACEPrincipal::System, systemSID.get(), TRUSTEE_IS_USER },
+    };
+
+    ULONG entriesCount = 0;
+    std::array<EXPLICIT_ACCESS_W, ARRAYSIZE(aceDetails)> explicitAccess;
+
+    for (const auto& ace : aceDetails)
+    {
+        if (owner && owner.value() == ace.Principal)
+        {
+            ownerSID = ace.SID;
+        }
+
+        auto itr = acl.find(ace.Principal);
+        if (itr != acl.end())
+        {
+            EXPLICIT_ACCESS_W& entry = explicitAccess[entriesCount++];
+            entry = {};
+            entry.grfAccessPermissions = AccessMaskFrom(itr->second);
+            entry.grfAccessMode = SET_ACCESS;
+            entry.grfInheritance = CONTAINER_INHERIT_ACE | OBJECT_INHERIT_ACE;
+            entry.Trustee.TrusteeForm = TRUSTEE_IS_SID;
+            entry.Trustee.TrusteeType = ace.TrusteeType;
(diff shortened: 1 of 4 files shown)


mamoreau-devolutions:
@microsoft-github-policy-service agree company="Devolutions"

src/AppInstallerCommonCore/Runtime.cpp
```diff
@@ -280,6 +280,7 @@ namespace AppInstaller::Runtime
         {
         case PathName::Temp:
             result.Path = GetPathToUserTemp(forDisplay) / s_DefaultTempDirectory;
+            result.Path /= GetRuntimePathStateName();
```
JohnMcPMS: Packaged doesn't technically have a state name, but I get that it also happens to share a root with the unpackaged code.  There isn't a way to set it for the packaged code either.  I would prefer a different function or a parameter to indicated packaged/unpackaged and to change the default state name if we are going to move it into a subdirectory.

I'm also not sure how much pain this will end up causing downstream as we move the cheese on people.  If the performance problem is remediated 99% of the time by checking the ACL, I would prefer to simply not change the structure.

src/AppInstallerSharedLib/Filesystem.cpp
```diff
@@ -56,6 +72,223 @@ namespace AppInstaller::Filesystem
...
+                    });
+                }
+            }
+
+            return result;
+        }
+
+        std::optional<std::map<ACEPrincipal, PrincipalPermissions>> GetActualPermissions(
+            PACL acl,
+            const std::map<std::wstring, ACEPrincipal>& sidToPrincipal)
```
JohnMcPMS: Given the small size of the container here, it would likely be better to use a vector, a linear search, and `EqualSid` rather than the cost of allocating SID strings and doing a binary search over at most 3 items.

src/AppInstallerCLITests/Runtime.cpp
```diff
@@ -104,6 +302,106 @@ TEST_CASE("ApplyACL_CurrentUserOwner_SystemAll", "[runtime]")
...
+    constexpr std::string_view stateName = "Issue6162State";
+    auto restoreStateName = wil::scope_exit([]() { Runtime::SetRuntimePathStateName("defaultState"); });
+
+    Runtime::SetRuntimePathStateName(std::string{ stateName });
+
+    PathDetails details = Runtime::GetPathDetailsFor(Runtime::PathName::Temp, true);
+
+    REQUIRE(details.Path.filename() == Utility::ConvertToUTF16(stateName));
+}
+
```
JohnMcPMS:
```suggestion
```

Test is no longer needed.

JohnMcPMS (approved):

PR #6172, opened 2026-04-22
Mitigate packaged preindexed source open lock convoy

### LINKED ISSUE #6167: Packaged preindexed source read opens are serialized behind an exclusive mutex under concurrent winget usage

Packaged `winget` appears to serialize read-only opens of the preindexed `winget` source behind a single exclusive cross-process mutex. Under concurrent `winget` usage, commands that should be quick source reads can stall for tens of seconds or longer before doing meaningful work.

### Steps to reproduce

Run several concurrent `show` commands and print elapsed time for each job:

```powershell
$w = (Get-Command winget.exe).Source

1..6 | ForEach-Object {
    $i = $_
    Start-Job -ArgumentList $i, $w {
        param($i, $w)
        $s = [Diagnostics.Stopwatch]::StartNew()
        & $w show --id Devolutions.RemoteDesktopManager --exact --source winget --disable-interactivity --accept-source-agreements *> $null
        [pscustomobject]@{
            Job      = $i
            Seconds  = [math]::Round($s.Elapsed.TotalSeconds, 3)
            ExitCode = $LASTEXITCODE
        }
    }
} | Wait-Job | Receive-Job | Sort-Object Job | Format-Table -AutoSize
```

### Sample run

```text
Job Seconds ExitCode
--- ------- --------
  1   44.84        0
  2  163.93        0
  3  122.67        0
  4  204.52        0
  5   86.86        0
  6  241.91        0
```

All six commands succeeded, but some waited several minutes.

### Expected behavior

Read-only commands such as `winget show` and `winget search` should not block behind unrelated concurrent `winget` activity when they only need to open and read the packaged source index.

### Actual behavior

The command often appears idle for a long time, then completes quickly once source open proceeds.

### Analysis

The relevant path appears to be in:

- `src\AppInstallerRepositoryCore\Microsoft\PreIndexedPackageSourceFactory.cpp`
- `src\AppInstallerCommonCore\Synchronization.cpp`
- `src\AppInstallerRepositoryCore\RepositorySource.cpp`

Observed behavior suggests that read-only packaged source open uses the same named cross-process mutex as source update/add/remove operations:

```text
PreIndexedSourceCPL_Microsoft.Winget.Source_8wekyb3d8bbwe
```

In instrumented runs, the expensive part was the wait to acquire that mutex, not the actual source read work. Representative timings from local investigation:

- SQLite search: ~3-5 ms
- warm package version data load: ~6-10 ms
- warm manifest load: ~12-18 ms
- packaged source open once it proceeds: ~80-130 ms
- mutex wait in contended runs: ~10 s, ~26 s, and higher in multi-process runs

This creates a convoy effect:

1. one process acquires the preindexed source mutex
2. other processes queue behind it, including read-only commands
3. the waiting time dominates wall-clock duration even though their own source read work is small

One detail that stood out while reading the code is that comments refer to a reader-writer lock, while the current implementation appears to use a plain exclusive mutex.

### Impact

This likely affects more than `winget show`. Any command that opens the packaged `winget` source may be impacted, including:

- `winget search`
- `winget show`
- `winget install` during package lookup / manifest resolution
- `winget upgrade` during lookup / manifest resolution
- `winget download`

For scripts or tools that invoke `winget` concurrently like UniGetUI, or just when developing winget locally with multiple AI agents running winget commands concurrently, this can cause highly variable latency and very large slowdowns in otherwise read-mostly workflows.

mamoreau-devolutions:

-----

## Related issue

Closes #6167

## Summary

This change narrows the lock scope for packaged preindexed source opens.

Previously, opening a packaged preindexed source always acquired the named cross-process lock before doing the read-only open work. Under concurrent `winget` usage, that serialized source reads behind unrelated activity and caused large stalls before commands like `show` or `search` could do meaningful work.

This PR changes the packaged-context open path to:

- attempt the packaged source open optimistically first
- reuse a helper for extension lookup, content integrity verification, and immutable SQLite open
- fall back to the existing cross-process lock only if the optimistic open fails
- preserve cancellation behavior
- add timing/logging around the open path to make the mode and cost visible

The change is limited to PreIndexedPackageSourceFactory.cpp.

## Testing

Manual validation against the issue repro:
- compared behavior against concurrent packaged source reads
- verified the fallback path still retries under the existing lock when optimistic open fails

Static validation:
- no diagnostics reported for the touched file
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6172)

### DIFF
--- a/src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
+++ b/src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
@@ -412,6 +412,35 @@ namespace AppInstaller::Repository::Microsoft
             return IsAfterUpdateCheckTime(details.Name, timeToCheck, requestedUpdateInterval);
         }
 
+        SQLiteIndex OpenPackagedContextIndex(const SourceDetails& details, IProgressCallback& progress, long long& extensionLookupMs, long long& verifyContentIntegrityMs, long long& sqliteOpenMs)
+        {
+            const auto extensionLookupStart = std::chrono::steady_clock::now();
+            auto extension = GetExtensionFromDetails(details);
+            extensionLookupMs += std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - extensionLookupStart).count();
+            if (!extension)
+            {
+                AICLI_LOG(Repo, Info, << "Package not found " << details.Data);
+                THROW_HR(APPINSTALLER_CLI_ERROR_SOURCE_DATA_MISSING);
+            }
+
+            const auto verifyStart = std::chrono::steady_clock::now();
+            THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_NEEDS_REMEDIATION), !extension->VerifyContentIntegrity(progress));
+            verifyContentIntegrityMs += std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - verifyStart).count();
+
+            // To work around an issue with accessing the public folder, we are temporarily
+            // constructing the location ourself.  This was already the case for the non-packaged
+            // runtime, and we can fix both in the future.  The only problem with this is that
+            // the directory in the extension *must* be Public, rather than one set by the creator.
+            std::filesystem::path indexLocation = extension->GetPackagePath();
+            indexLocation /= s_PreIndexedPackageSourceFactory_IndexFilePath;
+
+            const auto sqliteOpenStart = std::chrono::steady_clock::now();
+            auto index = SQLiteIndex::Open(indexLocation.u8string(), SQLiteIndex::OpenDisposition::Immutable);
+            sqliteOpenMs += std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - sqliteOpenStart).count();
+
+            return index;
+        }
+
         struct PackagedContextSourceReference : public ISourceReference
         {
             PackagedContextSourceReference(const SourceDetails& details) : m_details(details)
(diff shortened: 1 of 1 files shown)


src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -433,34 +462,55 @@ namespace AppInstaller::Repository::Microsoft
...
-                {
-                    return {};
-                }
+                const auto openStart = std::chrono::steady_clock::now();
+                bool succeeded = false;
+                bool usedLockFallback = false;
+                long long extensionLookupMs = 0;
+                long long verifyContentIntegrityMs = 0;
+                long long sqliteOpenMs = 0;
+                auto logOpenTiming = wil::scope_exit([&]()
```
JohnMcPMS: Putting all of this in its own object, along with a helper type for each timer, would make things much more readable.

```C++
struct OpenTimer
{
~OpenTimer() { AICLI_LOG(...); }

struct SingleTimer { void Start(){ m_start = now(); } void End(){ Duration = now() - m_start; } ... };

SingleTimer ExtensionLookup;
...
};
```

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -433,34 +462,55 @@ namespace AppInstaller::Repository::Microsoft
...
-                indexLocation /= s_PreIndexedPackageSourceFactory_IndexFilePath;
+                    Synchronization::CrossProcessLock lock(CreateNameForCPL(m_details));
+                    usedLockFallback = true;
+                    if (!lock.Acquire(progress))
+                    {
+                        return {};
+                    }
 
-                SQLiteIndex index = SQLiteIndex::Open(indexLocation.u8string(), SQLiteIndex::OpenDisposition::Immutable);
+                    auto index = OpenPackagedContextIndex(m_details, progress, extensionLookupMs, verifyContentIntegrityMs, sqliteOpenMs);
```
JohnMcPMS: The lock could be scoped to only be held during `OpenPackagedContextIndex`, which would also remove the issue of holding it during the related ACL issue code path inside the construction of `SQLiteIndexSource`.

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -433,34 +462,55 @@ namespace AppInstaller::Repository::Microsoft
...
                 }
+                catch (...)
+                {
+                    if (progress.IsCancelledBy(CancelReason::Any))
+                    {
+                        throw;
+                    }
 
-                THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_NEEDS_REMEDIATION), !extension->VerifyContentIntegrity(progress));
+                    LOG_CAUGHT_EXCEPTION_MSG("Optimistic packaged source open failed, retrying under lock for source: %hs", m_details.Name.c_str());
```
JohnMcPMS, line 544: This and below can be moved out of the `catch` to reduce line shifting.

src/AppInstallerRepositoryCore/Microsoft/PreIndexedPackageSourceFactory.cpp
```diff
@@ -433,34 +462,55 @@ namespace AppInstaller::Repository::Microsoft
...
 
-                // We didn't use to store the source identifier, so we compute it here in case it's
-                // missing from the details.
-                m_details.Identifier = GetPackageFamilyNameFromDetails(m_details);
-                return std::make_shared<SQLiteIndexSource>(m_details, std::move(index), false, true);
+                    // We didn't use to store the source identifier, so we compute it here in case it's
+                    // missing from the details.
+                    m_details.Identifier = GetPackageFamilyNameFromDetails(m_details);
+                    succeeded = true;
+                    return std::make_shared<SQLiteIndexSource>(m_details, std::move(index), false, true);
```
JohnMcPMS: I would prefer if these lines weren't duplicated.

JohnMcPMS (approved):

Issue #4238, opened 2024-03-06
Enable sorting the output of WinGet List


denelon:
### Description of the new feature / enhancement

I'd like to be able to sort the output of the list command.

The client should also have settings to specify the default sort behavior in addition to the arguments used to specify the sorting behavior.

Trenly:
[Policy] Area-Output
[Policy] Area-Sorting

HarjitSi:
While this capability is implemented, you can do: 

winget ls | sort

The output isn't pretty but better than not having it.

nbro:
Why isn't the list sorted alphabetically by default? There must be a very good reason. Right? Why would we ever want an apparently random list of installed packages, where we need to scroll up and down multiple times to find the needle in the haystack? Who thought of this? Who was the clever person that thought "Well, users get too bored when things are easy to do, let's just annoy them, that's good user experience!".

denelon:
It was not a requirement when "list" was introduced as a WinGet feature. At that point in time, the priority was to provide capabilities to unblock critical use cases. Given WinGet is open source, anyone could have submitted a PR to sort the output. I don't believe I know any product managers who think it's a good idea to annoy users. 🤔

I'm excited we're finally getting it done! 😊

PR #6176, opened 2026-04-25
[Settings] Add sort field and direction types for output ordering

Madhusudhan-MSFT:
## Summary
Add settings infrastructure for configurable output sorting in `winget list`. This is
part 1 of a 3-part stack that lays the type system and settings foundation without
changing any runtime behavior.

Part of https://github.com/microsoft/winget-cli/issues/4238

## Changes
- **SortField and SortDirection enums**: Add `SortField` (Relevance, Name, Id, Version,
  Source, Available) and `SortDirection` (Ascending, Descending) to the settings type
  system. Relevance is internal-only and not exposed as a valid settings value.
- **Settings entries**: Register `OutputSortOrder` and `OutputSortDirection` with typed
  validation that maps JSON string values to enums using case-insensitive comparison.
  Default sort order is `[Name]` ascending. Invalid field names reject the entire
  setting and fall back to default.
- **JSON schema**: Extend `settings.schema.0.2.json` with an `output` section containing
  `sortOrder` (string array) and `sortDirection` (string enum with ascending default).
- **Static constexpr pattern**: String constants for field name and direction comparisons
  follow the established `static constexpr std::string_view` pattern used by
  `NetworkDownloader`, `LoggingLevelPreference`, and other existing validators.

## Impact
- No behavioral change — settings are defined but not yet consumed by any command or
  workflow. Existing `winget list` output remains unchanged.
- Users who edit `settings.json` to add `output.sortOrder` or `output.sortDirection`
  will have their values validated and stored, but the values will not take effect
  until the sort logic is wired in a follow-up PR.

## How Validated
- Unit tests added: 16 test sections across `SettingOutputSortOrder` and
  `SettingOutputSortDirection` test cases covering default values, single and multi-field
  configurations, case insensitivity, empty array, invalid fields, mixed valid/invalid,
  and wrong JSON types (62 assertions total).
- Compilation verified for AppInstallerCommonCore and AppInstallerCLITests.

## Follow-up
This PR is part 1 of a 3-PR stack:
- **PR1** (this): Settings types and schema
- **PR2**: CLI arguments (`--sort`, `--ascending`, `--descending`)
- **PR3**: Sort logic wiring, `ConvertToSortField` shared parser, and comprehensive tests

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6176)

### DIFF
--- a/src/AppInstallerCLITests/UserSettings.cpp
+++ b/src/AppInstallerCLITests/UserSettings.cpp
@@ -665,13 +665,12 @@ TEST_CASE("SettingOutputSortOrder", "[settings]")
 {
     auto again = DeleteUserSettingsFiles();
 
-    SECTION("Default value")
+    SECTION("Default value - empty vector sentinel for context-aware defaults")
     {
         UserSettingsTest userSettingTest;
 
         auto sortOrder = userSettingTest.Get<Setting::OutputSortOrder>();
-        REQUIRE(sortOrder.size() == 1);
-        REQUIRE(sortOrder[0] == SortField::Name);
+        REQUIRE(sortOrder.empty());
         REQUIRE(userSettingTest.GetWarnings().size() == 0);
     }
     SECTION("Single field")
@@ -683,6 +682,17 @@ TEST_CASE("SettingOutputSortOrder", "[settings]")
         auto sortOrder = userSettingTest.Get<Setting::OutputSortOrder>();
         REQUIRE(sortOrder.size() == 1);
         REQUIRE(sortOrder[0] == SortField::Id);
+        REQUIRE(userSettingTest.GetWarnings().size() == 0);
+    }
+    SECTION("Relevance field")
+    {
+        std::string_view json = R"({ "output": { "sortOrder": ["relevance"] } })";
+        SetSetting(Stream::PrimaryUserSettings, json);
+        UserSettingsTest userSettingTest;
+
+        auto sortOrder = userSettingTest.Get<Setting::OutputSortOrder>();
+        REQUIRE(sortOrder.size() == 1);
+        REQUIRE(sortOrder[0] == SortField::Relevance);
         REQUIRE(userSettingTest.GetWarnings().size() == 0);
     }
     SECTION("Multiple fields")
@@ -699,17 +709,18 @@ TEST_CASE("SettingOutputSortOrder", "[settings]")
     }
     SECTION("All valid fields")
     {
-        std::string_view json = R"({ "output": { "sortOrder": ["name", "id", "version", "source", "available"] } })";
+        std::string_view json = R"({ "output": { "sortOrder": ["relevance", "name", "id", "version", "source", "available"] } })";
         SetSetting(Stream::PrimaryUserSettings, json);
         UserSettingsTest userSettingTest;
 
         auto sortOrder = userSettingTest.Get<Setting::OutputSortOrder>();
-        REQUIRE(sortOrder.size() == 5);
-        REQUIRE(sortOrder[0] == SortField::Name);
-        REQUIRE(sortOrder[1] == SortField::Id);
-        REQUIRE(sortOrder[2] == SortField::Version);
-        REQUIRE(sortOrder[3] == SortField::Source);
-        REQUIRE(sortOrder[4] == SortField::Available);
+        REQUIRE(sortOrder.size() == 6);
+        REQUIRE(sortOrder[0] == SortField::Relevance);
+        REQUIRE(sortOrder[1] == SortField::Name);
+        REQUIRE(sortOrder[2] == SortField::Id);
+        REQUIRE(sortOrder[3] == SortField::Version);
+        REQUIRE(sortOrder[4] == SortField::Source);
+        REQUIRE(sortOrder[5] == SortField::Available);
         REQUIRE(userSettingTest.GetWarnings().size() == 0);
     }
     SECTION("Case insensitive")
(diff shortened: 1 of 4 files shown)


Trenly:
> Relevance is internal-only and not exposed as a valid settings value.

Is there a reason not to expose this? It feels like some users may specifically want this like if they want to sort by Source then by Relevance

src/AppInstallerCLITests/UserSettings.cpp
```diff
@@ -660,3 +660,181 @@ TEST_CASE("LoggingChannels", "[settings]")
...
+        std::string_view json = R"({ "output": { "sortOrder": "name" } })";
+        SetSetting(Stream::PrimaryUserSettings, json);
+        UserSettingsTest userSettingTest;
+
+        auto sortOrder = userSettingTest.Get<Setting::OutputSortOrder>();
+        REQUIRE(sortOrder.size() == 1);
+        REQUIRE(sortOrder[0] == SortField::Name);
+        REQUIRE(userSettingTest.GetWarnings().size() == 1);
+    }
+    SECTION("Wrong type - number in array")
```
Trenly, line 792: I'd also appreciate a test case for duplicate values, something like `["name", "id",  "Name"]`

JohnMcPMS:
> > Relevance is internal-only and not exposed as a valid settings value.
> 
> Is there a reason not to expose this? It feels like some users may specifically want this like if they want to sort by Source then by Relevance

JohnMcPMS:
Sorry, misclick...

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -208,6 +229,9 @@ namespace AppInstaller::Settings
         SETTINGMAPPING_SPECIALIZATION(Setting::LoggingFileCountLimit, uint32_t, uint32_t, 0, ".logging.file.countLimit"sv);
         // Interactivity
         SETTINGMAPPING_SPECIALIZATION(Setting::InteractivityDisable, bool, bool, false, ".interactivity.disable"sv);
+        // Output behavior
+        SETTINGMAPPING_SPECIALIZATION(Setting::OutputSortOrder, std::vector<std::string>, std::vector<SortField>, std::vector<SortField>{ SortField::Name }, ".output.sortOrder"sv);
```
JohnMcPMS: I thought from our discussion the default behavior would be:
1. No change if query provided
2. Name if no query

In order to do that while having only one setting, you would need this to be a sentinel value that indicates "default".  An empty vector could probably work for that, as could an optional (but the empty vector maps well to an empty array in JSON).

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -47,6 +47,24 @@ namespace AppInstaller::Settings
         Disabled,
     };
 
+    // Sort field for output ordering.
+    enum class SortField
+    {
+        Relevance,  // Internal: preserves current natural order (not valid in settings)
```
JohnMcPMS: This should be publicly visible.  The expectation is that sources return their results in relevance order already.  `[relevance]` would then mean "keep the pre-change behavior".  If relevance was ever at the end, it would be a no-op as we should already use order preserving sorts.  The issue comes when relevance is not last, at which point we would need a way of knowing when relevance was equal to sort within those bands.  That isn't a value that you have to work with, although you could infer it from the match criteria.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -482,6 +482,62 @@ namespace AppInstaller::Settings
...
+            static constexpr std::string_view s_sortField_name = "name";
+            static constexpr std::string_view s_sortField_id = "id";
+            static constexpr std::string_view s_sortField_version = "version";
+            static constexpr std::string_view s_sortField_source = "source";
+            static constexpr std::string_view s_sortField_available = "available";
+
+            std::vector<SortField> fields;
+            for (auto const& entry : value)
+            {
+                if (Utility::CaseInsensitiveEquals(entry, s_sortField_name))
```
JohnMcPMS: If performing multiple case-insensitive comparisons, it is much more efficient to manually convert the `entry` to lowercase once and check equality.

Madhusudhan-MSFT:
Good question @Trenly — `relevance` means "preserve the current natural ordering" (i.e., don't re-sort). It's internal-only today for two reasons:

1. `sortOrder: ["relevance"]` would be a no-op — identical to having no sort setting at all
2. Combining it with other fields (e.g., `["source", "relevance"]`) isn't meaningful — once a primary sort reshuffles results, the original ordering is lost, so relevance as a tiebreaker would be non-deterministic and potentially confusing

If we were to expose it, it would only make sense as the sole value (e.g., `sortOrder: ["relevance"]`) to explicitly opt out of sorting — not combined with other fields.

@denelon — would you like us to expose `relevance` as a valid settings value (restricted to standalone use), or keep it internal-only?

denelon:
I like the idea of exposing "relevance" since someone might want to have that documented and be able to put it in their settings.

Trenly:
> If we were to expose it, it would only make sense as the sole value (e.g., sortOrder: ["relevance"]) to explicitly opt out of sorting — not combined with other fields.

I suppose it depends on the use case? The 'Source -> Relevance' one makes sense to me, since you may want that as opposed to `Source -> Name`, but there is also a case where the match strength of several items is the same. 

Take, for example, the case where I do `winget search <query>`.  Relevance ordering places monikers first, then IDs, followed by product codes / commands / tags. I could see where a user may want to keep the relevance order there, but in each section have the results sorted by Name so that it would be something like this (Assuming `Relevance -> Name`)

| Name | ID | Match |
| --- | --- | --- |
| This Package | Example.One | moniker: query |
| Query Viewer | Query.Viewer | |
| Four | Example.Four | tag: query |
| Three | Example.Three | tag: query |
| Two | Example.Two | tag: query |

Madhusudhan-MSFT:
Closing in favor of #6177 which now includes all changes from this PR plus the CLI argument registration. Reviewer feedback to consolidate into a single PR.

Madhusudhan-MSFT:
This PR has been closed. All changes have been consolidated into PR #6177.

PR #6179, opened 2026-04-25
feat: add upgrade delay setting for upgrade --all

### LINKED ISSUE #6178: Upgrade delay setting for 'winget upgrade --all'
### Relevant area(s)

WinGet CLI

Add an optional user setting `installBehavior.upgradeDelayInDays` that delays upgrades when running `winget upgrade --all` by skipping updates whose `ReleaseDate` is newer than the configured age, reducing exposure to newly published (and so, supply-chain/bug risky) releases.
So, for exemple, if `installBehavior.upgradeDelayInDays` is set to 5 days, and a new version of `Mozilla.Thunderbird` was released just yesterday, when running `winget upgrade --all`, the installation of this upgrade will be delayed until in 4 days.

It is the same idea as this config param from dependabot: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

Read `installBehavior.upgradeDelayInDays` from settings (default: 0).
During `winget upgrade --all`, for each available upgrade, if `installBehavior.upgradeDelayInDays` > 0, compare "today" to the package's ReleaseDate (prefer the selected applicable installer’s `ReleaseDate`, fallback to `manifest.DefaultInstallerInfo.ReleaseDate`).
If the age is less than the configured delay, skip that upgrade, continue processing the rest, and show a summary indicating how many upgrades were skipped due to the delay (with guidance to upgrade individually to override).
If `ReleaseDate` is missing/invalid, skip that upgrade in the same way as above.
If the age is more than the configured delay, process that upgrade in a classical way.

Those verifications are located in `UpdateAllApplicable`, so upgrading a single package is'nt affected by them.

mdevolde:

-----

This pull request contains the full implementation of the feature, but is under "draft" status until a maintainer agreed with my related issue (and potentially with a spec to write).

# feat: add upgrade delay setting for upgrade --all

### Summary
Add a new user setting `installBehavior.upgradeDelayInDays` (default: 0) to delay `winget upgrade --all` until an update is at least **N** days old, and permit to reduce risks to download a very recent release that could be part of a supply chain attack or contains bugs.

### Changes
- During `winget upgrade --all`, winget reads `ReleaseDate` (preferred: selected applicable installer, fallback: `manifest.DefaultInstallerInfo.ReleaseDate`).
- If the update is newer than the configured delay, it is skipped and a summary message is shown indicating how many upgrades were skipped due to the upgrade delay (upgrades with missing `ReleaseDate` are also skipped if `installBehavior.upgradeDelayInDays` > 0).
- Users can still upgrade a specific package individually to bypass the “--all” delay behavior.

### Testing
- Added/updated `AppInstallerCLITests` coverage for eligible vs too recent upgrades and verified `UpdateFlow_UpdateAllApplicable_UpgradeDelay` passes.

-----

Closes #6178
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6179)

### DIFF
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -218,6 +218,18 @@ The `maxResumes` setting determines the maximum number of times that a command m
 
 > Note: [The resume behavior is an experimental feature.](#resume)
 
+### Upgrade delay
+
+The `upgradeDelayInDays` setting delays upgrades when using `winget upgrade --all` by skipping upgrades whose manifest `ReleaseDate` is more recent than the configured age. The default value is 0 (disabled).
+
+```json
+    "installBehavior": {
+        "upgradeDelayInDays": 14
+    },
+```
+
+> Note: If the package manifest does not include a valid `ReleaseDate`, the upgrade will be skipped when this setting is enabled. To upgrade anyway, use `winget upgrade <package>` which does not consider this setting, or use `winget upgrade --all --force` which will ignore the age of the upgrade.
+
 ## Uninstall Behavior
 
 The `uninstallBehavior` settings affect the default behavior of uninstalling (where applicable) packages.
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -28,6 +28,10 @@ match criteria that factor into the result ordering. This will prevent them from
 
 ## Minor Features
 
+### Upgrade delay for `winget upgrade --all`
+
+Added the `installBehavior.upgradeDelayInDays` user setting to delay `winget upgrade --all` until an upgrade's `ReleaseDate` is at least N days old. This helps reduce exposure to newly published updates that may be part of a supply chain attack.
+
 ### Preserve installer arguments across export and import
 
 `winget export` now captures the `--override` and `--custom` arguments that were used when a package was originally installed and saves them into the export file. When subsequently running `winget import`, those values are automatically re-applied during installation — `--override` replaces all installer arguments and `--custom` appends extra switches — so packages can be reinstalled with the same customizations without any manual intervention. Both fields are optional and independent of each other; packages without stored installer arguments are unaffected.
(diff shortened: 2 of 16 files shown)


mdevolde:

doc/Settings.md
```diff
@@ -218,6 +218,18 @@ The `maxResumes` setting determines the maximum number of times that a command m
...
+
+The `upgradeDelayInDays` setting delays upgrades when using `winget upgrade --all` by skipping upgrades whose manifest `ReleaseDate` is more recent than the configured age. The default value is 0 (disabled).
+
+```json
+    "installBehavior": {
+        "upgradeDelayInDays": 14
+    },
+```
+
+> Note: If the package manifest does not include a valid `ReleaseDate`, the upgrade will be skipped when this setting is enabled. To upgrade anyway, use `winget upgrade <package>` which does not consider this setting.
```
Trenly: I think this would likely prevent the upgrade of most applications with `--all`, as a majority of manifests don’t specify a release date and it isn’t a mandatory field. Although you've made this as an opt-in feature, I'm not sure this makes sense for most users. I'd suggest having `--force` bypass the check for installer dates so users don’t have to specify every single package if there are several without release dates

mdevolde: Thanks for your review, I'm editing in that way.

petergoodey:
@mdevolde Just found this. I'll try and get a look through in a few days time

mdevolde:
@petergoodey I invite you to carefully read the entire discussion in the linked issue, where a few ideas are presented. The challenge lies in obtaining the publication date from a source that is independent of a potential attacker (this condition is not met in my current contribution, which retrieves the information from the manifest, which can be manipulated by an attacker, and where the publication date is optional).

PR #6183, opened 2026-04-28
Add JSON validation output and interop presentation model

JohnMcPMS:
## Change
Adds a new flag to request JSON from manifest creation/validation pass.  Doing so enables the interop layer to provide the validation warnings/errors as an object model.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6183)

### DIFF
--- a/src/WinGetUtil/WinGetUtil.h
+++ b/src/WinGetUtil/WinGetUtil.h
@@ -49,6 +49,10 @@ extern "C"
 
         // Return error if a network address is present in installer switches.
         ReturnErrorOnNetworkAddressInSwitches = 0x2000,
+
+        // Return the failure or warning message as a JSON string containing both the full message
+        // and a structured list of individual errors/warnings.
+        ReturnResponseAsJSON = 0x4000,
     };
 
     DEFINE_ENUM_FLAG_OPERATORS(WinGetCreateManifestOption);
--- a/src/WinGetUtil/WinGetUtil.h
+++ b/src/WinGetUtil/WinGetUtil.h
@@ -52,7 +52,7 @@ extern "C"
 
         // Return the failure or warning message as a JSON string containing both the full message
         // and a structured list of individual errors/warnings.
-        ReturnResponseAsJSON = 0x4000,
+        ReturnResponseAsJson = 0x4000,
     };
 
     DEFINE_ENUM_FLAG_OPERATORS(WinGetCreateManifestOption);
--- a/src/WinGetUtilInterop/Api/WinGetFactory.cs
+++ b/src/WinGetUtilInterop/Api/WinGetFactory.cs
@@ -1,4 +1,4 @@
-﻿// -----------------------------------------------------------------------------
+// -----------------------------------------------------------------------------
 // <copyright file="WinGetFactory.cs" company="Microsoft Corporation">
 //     Copyright (c) Microsoft Corporation. Licensed under the MIT License.
 // </copyright>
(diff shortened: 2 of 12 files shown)


src/WinGetUtil/WinGetUtil.h
```diff
@@ -49,6 +49,10 @@ extern "C"
 
         // Return error if a network address is present in installer switches.
         ReturnErrorOnNetworkAddressInSwitches = 0x2000,
+
+        // Return the failure or warning message as a JSON string containing both the full message
+        // and a structured list of individual errors/warnings.
+        ReturnResponseAsJSON = 0x4000,
```
yao-msft:
```suggestion
        ReturnResponseAsJson = 0x4000,
```
nit: Json was used across the project. 

Same for the corresponding one in UtilInterop

yao-msft (dismissed):

alialobidm:
``

src/WinGetUtilInterop/Api/WinGetFactory.cs
```diff
@@ -140,6 +153,50 @@ public IWinGetInstallerMetadata BeginInstallerMetadataCollection(
             }
         }
 
+        private static CreateManifestResult ParseJsonManifestResult(bool succeeded, string json, IWinGetManifest manifestHandle)
+        {
+            try
+            {
+                using var doc = JsonDocument.Parse(json);
```
msftrubengu, line 160: Is there a reason why not using JsonSerializer.Deserialize with some private class or ManifestDiagnostic?

JohnMcPMS: Tokens already paid, so here is the 🤖 reasoning:

## Manual `TryGetProperty` vs. `JsonSerializer.Deserialize<T>` — Trade-offs
   
   ### In favor of the current approach (manual `TryGetProperty`)
   
   **Explicit field-level defaults.**
   Each `TryGetProperty` line specifies its own fallback (`string.Empty`, `0`, `"Error"`).
   With a deserializer DTO, missing fields silently get CLR defaults — `null` for `string`, not `string.Empty`.
   Matching the same semantics would require nullable annotations and post-construction fixup.
   
   **Decoupled domain type.**
   `ManifestDiagnostic` is the public API type.
   A deserializer approach would require either a separate private DTO (plus a mapping step) or
   decorating `ManifestDiagnostic` itself with `[JsonPropertyName]` attributes, leaking the wire format into the 
  public model.
   
   **`ManifestErrorId` and `ManifestDiagnosticLevel` require custom handling regardless.**
   The JSON carries strings (`"RequiredFieldMissing"`, `"Warning"`); the deserializer cannot map these automatically.
   You would need a `JsonConverter<ManifestErrorId>` and a `JsonConverter<ManifestDiagnosticLevel>` —
   roughly the same logic that is inlined today, just more indirectly.
   
   **Resilience to format changes.**
   `TryGetProperty` degrades gracefully field-by-field.
   A deserializer that hits an unexpected property or type mismatch can throw (or silently drop) the whole object.
   Because the JSON producer is native C++, subtle type differences (e.g., integer vs. string for `line`) are more 
  likely.
   
   ---
   
   ### In favor of `JsonSerializer.Deserialize<T>`
   
   **Less boilerplate as the schema grows.**
   If the number of fields doubles, the deserializer approach scales better than one `TryGetProperty` call per field.
   
   **Correctness guarantees from the type system.**
   `[JsonRequired]` can enforce mandatory fields and produce a clear exception rather than a silently zero-filled 
  `ManifestDiagnostic`.
   
   **Easier isolated testing.**
   DTO deserialization can be tested independently of the rest of the parsing logic.
   
   ---
   
   ### Verdict
   
   The manual approach is the right call here.
   The schema is stable and small (8 fields), both enum conversions need custom logic regardless,
   `ManifestDiagnostic` is a public type that should not be decorated with JSON attributes,
   and `TryGetProperty`-with-fallback gives controlled degradation across the C++/C# boundary.
   A deserializer would require a private DTO, custom converters, and a mapping step —
   more total code for no meaningful gain at this scale.

msftrubengu (approved):

PR #6187, opened 2026-04-29
Reset default sources without dropping them

### LINKED ISSUE #1575: winget source reset <name> removes default sources
`winget source reset <name>` removes the named source when it is a default rather than resetting it.  This is because `DropSource` just removes the source from the list without invoking its `Remove` as intended, but for a default source the removal from the list places a tombstone.

For a default source we should actually drop any existing item or tombstone in the list (save the platform special case).  The result should be that resetting a default source by name gets you the exact same state as a clean install would for that source (source is configured but has empty metadata).

# Steps to reproduce

`winget source reset msstore`

# Expected behavior

The default source is still configured but without metadata.  Agreements are reset for instance.

# Actual behavior

The source is removed (dropped really).

# Environment

```none
> winget --info
Windows Package Manager v1.1.12702

Windows: Windows.Desktop v10.0.22474.1000
Package: Microsoft.DesktopAppInstaller v1.16.12702.0
```

Trenly:
  - Resolves https://github.com/microsoft/winget-cli/issues/1575

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6187)

### DIFF
--- a/src/AppInstallerRepositoryCore/RepositorySource.cpp
+++ b/src/AppInstallerRepositoryCore/RepositorySource.cpp
@@ -1105,7 +1105,18 @@ namespace AppInstaller::Repository
                 AICLI_LOG(Repo, Info, << "Named source to be dropped, found: " << source->Name);
 
                 EnsureSourceIsRemovable(*source);
-                sourceList.RemoveSource(*source);
+
+                // For default sources and overrides of default sources, reset to clean state
+                // (remove user customizations and clear metadata) instead of tombstoning.
+                if (source->Origin == SourceOrigin::Default ||
+                    (source->Origin == SourceOrigin::User && source->IsOverride))
+                {
+                    sourceList.ResetSource(*source);
+                }
+                else
+                {
+                    sourceList.RemoveSource(*source);
+                }
 
                 return true;
             }
--- a/src/AppInstallerRepositoryCore/RepositorySource.cpp
+++ b/src/AppInstallerRepositoryCore/RepositorySource.cpp
@@ -1094,7 +1094,7 @@ namespace AppInstaller::Repository
         {
             SourceList sourceList;
 
-            auto source = sourceList.GetCurrentSource(name);
+            auto source = sourceList.GetSource(name);
             if (!source)
             {
                 AICLI_LOG(Repo, Info, << "Named source to be dropped, but not found: " << name);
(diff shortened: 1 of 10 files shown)


src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -661,6 +661,63 @@ namespace AppInstaller::Repository
...
+                OverwriteMetadata();
+            }
+        }
+
+        THROW_HR_IF_MSG(E_UNEXPECTED, !sourcesSet, "Too many attempts at SetSourcesByOrigin");
+
+        // Reload to bring back the Default source (now that the user entry is removed)
+        // and to get fresh metadata for all other sources.
+        OverwriteSourceList();
+        OverwriteMetadata();
```
JohnMcPMS: Shouldn't be needed as this builds `m_sourceList` from stored contents and you just set it above.

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -661,6 +661,63 @@ namespace AppInstaller::Repository
...
+        bool metadataSet = false;
+        for (size_t i = 0; !metadataSet && i < 10; ++i)
+        {
+            auto target = FindSource(details.Name, true);
+            if (target != m_sourceList.end())
+            {
+                target->LastUpdateTime = {};
+                target->DoNotUpdateBefore = {};
+                target->AcceptedAgreementFields = 0;
+                target->AcceptedAgreementsIdentifier = {};
```
JohnMcPMS: Make a `ResetMetadataFields` function to hold this semantic, implemented near the other functions so that any new metadata doesn't get forgotten.

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -661,6 +661,63 @@ namespace AppInstaller::Repository
...
+        THROW_HR_IF_MSG(E_UNEXPECTED, !sourcesSet, "Too many attempts at SetSourcesByOrigin");
+
+        // Reload to bring back the Default source (now that the user entry is removed)
+        // and to get fresh metadata for all other sources.
+        OverwriteSourceList();
+        OverwriteMetadata();
+
+        // Clear the metadata for the reset source and save.
+        bool metadataSet = false;
+        for (size_t i = 0; !metadataSet && i < 10; ++i)
```
JohnMcPMS: I see the issue with sharing `SaveMetadataInternal`, but I would prefer that there be a `ResetMetadataInternal` that contains this loop rather than it be implemented here.

src/AppInstallerRepositoryCore/RepositorySource.cpp
JohnMcPMS: While we are at it, should we also get tombstones here so you can reset `winget` if you have removed it?

src/AppInstallerRepositoryCore/SourceList.cpp
```diff
@@ -1006,4 +1047,30 @@ namespace AppInstaller::Repository
...
+            {
+                target->ResetMetadataFields();
+            }
+
+            metadataSet = SetMetadata(m_sourceList);
+
+            if (!metadataSet)
+            {
+                OverwriteSourceList();
+            }
```
JohnMcPMS:
```suggestion
```
We don't need to reload the source list after failing to update the metadata.

src/AppInstallerRepositoryCore/RepositorySource.cpp
Trenly: Yes, that makes sense

JohnMcPMS (approved):

Issue #3234, opened 2023-05-09
Add "Limit-WinGetPackage"


denelon:
### Description of the new feature / enhancement

The Microsoft.WinGet.Client module needs to have parity with the capabilities for `winget pin`.

### Proposed technical implementation details

Refactor code as needed into COM so the WinGet client and the PowerShell modules are using the same interface.

denelon:
Additional discussion regarding the cmdlet names at:
* https://github.com/microsoft/winget-cli/discussions/2980#discussioncomment-4999565

Issue #3214, opened 2023-05-04
Record date pin was added


Trenly:
### Description of the new feature / enhancement

As a user, I'd like to be able to periodically audit my pins. Project needs change over time. It may be possible that at one point a project required a specific version of a package, but then later added support for newer versions of the same package. This typically doesn't happen overnight which means that I probably won't be able to remember when I first added the pin. So, every time I want to audit the packages I have pinned, I have to go through every single one rather than ones that are older than say . . . 6 months. 

It would be extremely useful to be able to see the date a pin was added so when I audit my pins, I can skip ones which are only a month or two old. 

* Related to #3212

### Proposed technical implementation details

Also show the date in `winget pin list`

Issue #3213, opened 2023-05-04
Add note to pins


Trenly:
### Description of the new feature / enhancement

As a user, I may not always remember why I have a package pinned. It would be nice if a note could be added when creating the pin and if the note could be shown when listing the pins

```
winget pin add NodeJs.NodeJs -v 17.* --note 'ProjectA requires version 17'
winget pin add Discord.Discord --note 'Squirrel installers go nuts with winget'
winget pin add PackageABC --blocking --note 'Upgrading wipes config - be sure to export config before update'
```

* Related to #3212

PR #6190, opened 2026-04-29
Bring Pinning to PowerShell Cmdlets

Trenly:
Bring Pinning to PowerShell Cmdlets

  - #3214
  - #3213
  - #3234

## Summary

This PR extends the WinGet pinning system with new metadata, a new CLI subcommand, expanded COM API surface, and PowerShell cmdlets for managing pins.

### CLI Changes

- **`winget pin add`**: Added `--note` argument to attach an optional freeform note to a pin. The timestamp when a pin is created is now recorded automatically.
- **`winget pin list --details`**: New option that displays detailed information about pins for all packages or a specific package (query by ID, name, or keyword).

### Pinning Index Schema (v1.1)

- New schema version adds `DateAdded` and `Note` columns to the pin table to persist the new metadata.

### COM API (`Microsoft.Management.Deployment`, contract v30)

- `PackageManager.GetAllPins()` — retrieve all pins across all sources.
- `PackageManager.GetPins(CatalogPackage)` — retrieve pins for a specific package.
- `PackageManager.PinPackage(CatalogPackage, PinPackageOptions)` — add or update a pin.
- `PackageManager.UnpinPackage(CatalogPackage)` — remove all pins for a package.
- `PackageManager.ResetAllPins(String sourceName)` — reset all pins, optionally scoped to a source.
- New `PackagePin` runtime class exposing `PackageId`, `SourceId`, `Type`, `GatedVersion`, `DateAdded`, `Note`, and `IsForInstalledPackage`.
- New `PinPackageOptions` and `PinPackageResult` runtime classes.
- New `PackagePinType` enum (`PinnedByManifest`, `Pinning`, `Gating`, `Blocking`).

### PowerShell (`Microsoft.WinGet.Client`)

- `Add-WinGetPin` — pin a package with a specified pin type, optional gated version, note, and force flag.
- `Get-WinGetPin` — list pins, filterable by package/source.
- `Remove-WinGetPin` — remove pins for a package.
- `Reset-WinGetPin` — reset all pins, optionally scoped to a source.
- `Get-WinGetPackage` — new `IsPinned` property on returned objects.

cc @denelon for Naming

-----


denelon:
Were you able to figure out the PlatyPS MAML stuff for the PowerShell help?

Trenly:
> Were you able to figure out the PlatyPS MAML stuff for the PowerShell help?

`Get-Help` seems to work the same way for the new cmdlets as it does for the others. I might look at improving the PowerShell help separately in the future. This PR is already bigger than I'm sure John might like it to be, don't want to make anyone too frazzled

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -197,9 +199,21 @@ namespace AppInstaller::CLI::Workflow
 
         if (!pinsToAddOrUpdate.empty())
         {
-            for (const auto& pin : pinsToAddOrUpdate)
+            std::string dateAdded = Utility::TimePointToString(
```
JohnMcPMS: We should store the date in a form that can be more easily reasoned about later.  Typically, this is a Unix timestamp.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -197,9 +199,21 @@ namespace AppInstaller::CLI::Workflow
...
 
+            std::optional<std::string> note;
+            if (context.Args.Contains(Execution::Args::Type::PinNote))
             {
+                note = std::string{ context.Args.GetArg(Execution::Args::Type::PinNote) };
+            }
+
+            for (auto& pin : pinsToAddOrUpdate)
+            {
+                pin.SetDateAdded(dateAdded);
```
JohnMcPMS: The object model should use a `std::chrono::<CLOCK>::timepoint` (or `time_point`?) rather than the encoded version for storage.  Which is to say, take in a std time and convert it only at the point of storage.

src/AppInstallerCLICore/Commands/PinCommand.cpp
```diff
@@ -23,6 +23,7 @@ namespace AppInstaller::CLI
             std::make_unique<PinRemoveCommand>(FullName()),
             std::make_unique<PinListCommand>(FullName()),
             std::make_unique<PinResetCommand>(FullName()),
+            std::make_unique<PinShowCommand>(FullName()),
```
JohnMcPMS: I would prefer that we follow the model of package list and have `--details` on the existing `list` command.  That also removes the need to create a whole new search mechanism and just makes the change in `ReportPins`.

Followup: Well it looks like it would need a whole new search mechanism to actually implement the arguments that it claims to support...

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1913,6 +1917,48 @@ Please specify one of them using the --source option to proceed.</value>
     <value>Pinned version</value>
     <comment>Table header for the version to which a package is pinned; meaning it should not update from that version.</comment>
   </data>
+  <data name="PinDateAdded" xml:space="preserve">
+    <value>Date added</value>
+    <comment>Label shown in the pin show output for when the pin was added or last updated.</comment>
```
JohnMcPMS:
```suggestion
    <value>Time Pinned</value>
    <comment>Label shown in the pin show output for when the pin was added or last updated.</comment>
```
The comment suggests it will be the last write time, not the first write time.  The string should actually reflect that.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1913,6 +1917,48 @@ Please specify one of them using the --source option to proceed.</value>
...
+  <data name="PinDateAdded" xml:space="preserve">
+    <value>Date added</value>
+    <comment>Label shown in the pin show output for when the pin was added or last updated.</comment>
+  </data>
+  <data name="PinNote" xml:space="preserve">
+    <value>Note</value>
+    <comment>Label shown in the pin show output for the user-provided note stored with the pin.</comment>
+  </data>
+  <data name="PinShowCommandShortDescription" xml:space="preserve">
+    <value>Show details about a pin</value>
```
JohnMcPMS: All strings should have comments.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1913,6 +1917,48 @@ Please specify one of them using the --source option to proceed.</value>
...
+  </data>
+  <data name="PinShowCommandLongDescription" xml:space="preserve">
+    <value>Show detailed information about a specific pin, including the package ID, version, type, date added, and any note stored with the pin.</value>
```
JohnMcPMS:
```suggestion
    <value>Show detailed information about a specific pin, including the package identifier, version, type, date added, and any note stored with the pin.</value>
```

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -1913,6 +1917,48 @@ Please specify one of them using the --source option to proceed.</value>
...
+  </data>
+  <data name="PinShowLabelDateAdded" xml:space="preserve">
+    <value>Date added:</value>
```
JohnMcPMS: Keep in line with other.

src/AppInstallerCLITests/PinFlow.cpp
```diff
@@ -198,4 +198,221 @@ TEST_CASE("PinFlow_ResetEmpty", "[PinFlow][workflow]")
...
+    INFO(pinAddOutput.str());
+
+    auto index = PinningIndex::Open(indexFile.GetPath().u8string(), SQLiteStorageBase::OpenDisposition::Read);
+    auto pins = index.GetAllPins();
+    REQUIRE(pins.size() == 1);
+    REQUIRE(pins[0].GetNote().has_value());
+    REQUIRE(pins[0].GetNote().value() == "my test note");
+}
+
+TEST_CASE("PinFlow_Add_WithoutNote", "[PinFlow][workflow]")
```
JohnMcPMS: Feels like this and the date "defaults check" could just be additional validations into an existing "only supplies required inputs" test.

src/AppInstallerCommonCore/Public/winget/Pin.h
```diff
@@ -97,6 +99,11 @@ namespace AppInstaller::Pinning
         PinType GetType() const { return m_type; }
         const PinKey& GetKey() const { return m_key; }
         const Utility::GatedVersion& GetGatedVersion() const { return m_gatedVersion; }
+        const std::string& GetDateAdded() const { return m_dateAdded; }
```
JohnMcPMS: As mentioned above, use chrono in the object model.  Should probably also be an optional rather than using a sentinel value.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinningIndexInterface.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+#include "Microsoft/Schema/Pinning_1_0/PinningIndexInterface.h"
+
+namespace AppInstaller::Repository::Microsoft::Schema::Pinning_V1_1
+{
+    struct PinningIndexInterface : public Pinning_V1_0::PinningIndexInterface
+    {
+        // Version 1.1
```
JohnMcPMS: I typically use these comments to separate the various functions into groups based on when they were introduced to the interface.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinningIndexInterface.h
```diff
@@ -0,0 +1,21 @@
...
+        SQLite::Version GetVersion() const override;
+        void CreateTables(SQLite::Connection& connection) override;
+        bool MigrateFrom(SQLite::Connection& connection, const IPinningIndex* current) override;
+
+    private:
+        SQLite::rowid_t AddPin(SQLite::Connection& connection, const Pinning::Pin& pin) override;
```
JohnMcPMS, line 13: Why are these private?  I honestly don't know what C++ does in that case, I would assume that they are private if accessed through a `PinningIndexInterface` but public if accessed through a `IPinningIndex`.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinningIndexInterface_1_1.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // Version 1.1
+    SQLite::Version PinningIndexInterface::GetVersion() const
+    {
+        return { 1, 1 };
+    }
+
+    void PinningIndexInterface::CreateTables(SQLite::Connection& connection)
+    {
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "createpintable_v1_1");
+        Pinning_V1_1::PinTable::Create(connection);
```
JohnMcPMS: I would prefer this to be:
1. Create 1.0 tables
2. Migrate to 1.1

This makes the uncommon case common at a small performance penalty to a one time operation.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+    static constexpr std::string_view s_PinTable_DateAdded_Column = "date_added"sv;
+    static constexpr std::string_view s_PinTable_Note_Column = "note"sv;
+    static constexpr std::string_view s_PinTable_Index = "pin_index"sv;
+
+    std::string_view PinTable::TableName()
+    {
+        return s_PinTable_Table_Name;
+    }
+
+    void PinTable::Create(SQLite::Connection& connection)
```
JohnMcPMS: [File level comment really]

I don't like that this appears to be a complete copy of the 1.0 code rather than reuse of the underlying code.  I do think that it would require a refactor to make that underlying code more sharable given that this is a single table database with a simple object design.  Maybe I will just have to be sad while also being pragmatic. 🐼 

Specifically for this function though, I think it should be removed per-my earlier comment.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+            .Columns({ s_PinTable_PackageId_Column, s_PinTable_SourceId_Column });
+        createIndexBuilder.Execute(connection);
+
+        savepoint.Commit();
+    }
+
+    void PinTable::MigrateFrom1_0(SQLite::Connection& connection)
+    {
+        SQLite::Statement addDateAdded = SQLite::Statement::Create(connection,
+            "ALTER TABLE pin ADD COLUMN date_added TEXT NOT NULL DEFAULT ''");
```
JohnMcPMS: I've noticed that the 🤖 doesn't like using my builder pattern and just says "I know how to SQL".  But I don't always know how to SQL and I ***do*** want to use symbolic names for the tables and columns.  So please builder it.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+        addDateAdded.Execute();
+
+        SQLite::Statement addNote = SQLite::Statement::Create(connection,
+            "ALTER TABLE pin ADD COLUMN note TEXT");
+        addNote.Execute();
+    }
+
+    std::optional<SQLite::rowid_t> PinTable::GetIdByPinKey(SQLite::Connection& connection, const Pinning::PinKey& pinKey)
```
JohnMcPMS: This looks like it does not need to be copied into 1.1

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+        builder.InsertInto(s_PinTable_Table_Name)
+            .Columns({
+                s_PinTable_PackageId_Column,
+                s_PinTable_SourceId_Column,
+                s_PinTable_Type_Column,
+                s_PinTable_Version_Column,
+                s_PinTable_DateAdded_Column,
+                s_PinTable_Note_Column })
+            .Values(
+                (std::string_view)pinKey.PackageId,
```
JohnMcPMS: No C-style casts please.  Plus `string_view` is inefficient if you have a `string` already.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+        StatementBuilder createIndexBuilder;
+        createIndexBuilder.CreateUniqueIndex(s_PinTable_Index).On(s_PinTable_Table_Name)
+            .Columns({ s_PinTable_PackageId_Column, s_PinTable_SourceId_Column });
+        createIndexBuilder.Execute(connection);
+
+        savepoint.Commit();
+    }
+
+    void PinTable::MigrateFrom1_0(SQLite::Connection& connection)
+    {
```
JohnMcPMS: Any multi-step write should use a Savepoint to create a transaction.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+        SQLite::Builder::StatementBuilder builder;
+        const auto& pinKey = pin.GetKey();
+        builder.Update(s_PinTable_Table_Name).Set()
+            .Column(s_PinTable_PackageId_Column).Equals((std::string_view)pinKey.PackageId)
+            .Column(s_PinTable_SourceId_Column).Equals(pinKey.SourceId)
+            .Column(s_PinTable_Type_Column).Equals(pin.GetType())
+            .Column(s_PinTable_Version_Column).Equals(pin.GetGatedVersion().ToString())
+            .Column(s_PinTable_DateAdded_Column).Equals((std::string_view)pin.GetDateAdded());
+
+        // Use Unbound (= ?) for null note so SQLite stores NULL via = NULL, not the invalid SET syntax IS NULL.
```
JohnMcPMS: I've seen the 🤖 complain about this before... if it is wrong, why won't you fix it 🤖 ?  Why do just avoid it?

Anyway, we should fix it rather than workarounds.  It is referencing `StatementBuilder& Equals(const std::optional<ValueType>& value)`, and the issue is coming from the double semantics it is supporting.  The first is in a where clause, when `IS NULL` is the correct thing to do.  The second, and applicable here, is the update assignment.  Feels like we just need an alternate function or another parameter to specify the semantics to use when the optional is empty.

If the builder was more sophisticated, it would store state to prevent the need for it to be specified.  But that is a whole lot more complicated.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+        {
+            builder.Column(s_PinTable_Note_Column).Equals(SQLite::Builder::Unbound);
+        }
+
+        builder.Where(SQLite::RowIDName).Equals(pinId);
+        builder.Execute(connection);
+        return connection.GetChanges() != 0;
+    }
+
+    void PinTable::RemovePinById(SQLite::Connection& connection, SQLite::rowid_t pinId)
```
JohnMcPMS: Another one that doesn't seem to be new for 1.1

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+        SQLite::Statement select = builder.Prepare(connection);
+
+        if (!select.Step())
+        {
+            return {};
+        }
+
+        auto [packageId, sourceId, pinType, gatedVersion, dateAdded, note] =
+            select.GetRow<std::string, std::string, Pinning::PinType, std::string, std::string, std::optional<std::string>>();
+        return GetPinFromRow(packageId, sourceId, pinType, gatedVersion, dateAdded, std::move(note));
```
JohnMcPMS: Why only move `note`?  Why not move all?  Or take in a `tuple&&` in `GetPinFromRow`.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include "pch.h"
+#include "PinTable.h"
+#include <winget/SQLiteStatementBuilder.h>
+#include "Microsoft/Schema/IPinningIndex.h"
+
+namespace AppInstaller::Repository::Microsoft::Schema::Pinning_V1_1
+{
+    namespace
+    {
+        std::optional<Pinning::Pin> GetPinFromRow(
```
JohnMcPMS: Maybe this should be a `protected` on 1.0 since it looks like the 1.1 updates are just "set more values".

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
```diff
@@ -0,0 +1,242 @@
...
+            if (pin)
+            {
+                pins.push_back(std::move(pin.value()));
+            }
+        }
+
+        return pins;
+    }
+
+    bool PinTable::ResetAllPins(SQLite::Connection& connection, std::string_view sourceId)
```
JohnMcPMS: Another unchanged

src/AppInstallerRepositoryCore/AppInstallerRepositoryCore.vcxproj
```diff
@@ -449,6 +451,10 @@
     <ClCompile Include="Microsoft\Schema\ISQLiteIndex.cpp" />
     <ClCompile Include="Microsoft\Schema\Pinning_1_0\PinningIndexInterface_1_0.cpp" />
     <ClCompile Include="Microsoft\Schema\Pinning_1_0\PinTable.cpp" />
+    <ClCompile Include="Microsoft\Schema\Pinning_1_1\PinningIndexInterface_1_1.cpp" />
+    <ClCompile Include="Microsoft\Schema\Pinning_1_1\PinTable.cpp">
+      <ObjectFileName>$(IntDir)Schema\Pinning_1_1\</ObjectFileName>
```
JohnMcPMS: ?

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -186,13 +187,24 @@ namespace AppInstaller::Repository::Microsoft
 
     std::unique_ptr<Schema::IPinningIndex> PinningIndex::CreateIPinningIndex() const
     {
-        if (m_version == SQLite::Version{ 1, 0 } ||
-            m_version.MajorVersion == 1 ||
-            m_version.IsLatest())
+        // Always return the latest interface; migration will handle version mismatches.
+        return std::make_unique<Schema::Pinning_V1_1::PinningIndexInterface>();
```
JohnMcPMS: I don't mind moving the logic to a static, but we shouldn't just always create a 1.1 (if we always wanted latest, there is a better way to do that).  I would remove this function.

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -201,12 +213,46 @@ namespace AppInstaller::Repository::Microsoft
     {
         AICLI_LOG(Repo, Info, << "Opened Pinning Index with version [" << m_version << "], last write [" << GetLastWriteTime() << "]");
         m_interface = CreateIPinningIndex();
-        THROW_HR_IF(APPINSTALLER_CLI_ERROR_CANNOT_WRITE_TO_UPLEVEL_INDEX, disposition == SQLiteStorageBase::OpenDisposition::ReadWrite && m_version != m_interface->GetVersion());
+
+        if (m_version != m_interface->GetVersion())
```
JohnMcPMS: I much prefer that we create the correct interface, then if opening writable we create a latest interface, check the version, attempt to migrate and replace `m_interface` if it succeeds.

src/Microsoft.Management.Deployment/PackageManager.h
```diff
@@ -54,6 +54,12 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         winrt::hstring Version() const;
         // Contract 28.0
         winrt::Microsoft::Management::Deployment::EditPackageCatalogResult EditPackageCatalog(winrt::Microsoft::Management::Deployment::EditPackageCatalogOptions options);
+        // Contract 30.0
```
JohnMcPMS: 29 for now

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1679,8 +1679,130 @@ namespace Microsoft.Management.Deployment
...
+            /// Requires Force = true to overwrite an existing pin of a different type.
+            PinPackageResult PinPackage(CatalogPackage package, PinPackageOptions options);
+
+            /// Remove all pins associated with the specified package.
+            /// Returns PackagePinNotFound if no pin exists for the package.
+            PinPackageResult UnpinPackage(CatalogPackage package);
+
+            /// Reset (remove) all pins. If sourceName is non-empty, only pins from that
+            /// catalog source are removed; otherwise all pins are removed.
+            PinPackageResult ResetAllPins(String sourceName);
```
JohnMcPMS:
```suggestion
            PinPackageResult ResetAllPins(PackageCatalogReference catalog);
```

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1679,8 +1679,130 @@ namespace Microsoft.Management.Deployment
...
+        String SourceId { get; };
+
+        /// The type of the pin.
+        PackagePinType Type { get; };
+
+        /// The gated version range. Only meaningful when Type is Gating; empty otherwise.
+        String GatedVersion { get; };
+
+        /// The UTC date/time when the pin was added (ISO 8601 format).
+        String DateAdded { get; };
```
JohnMcPMS: The appropriate WinRT time type instead of string.

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -1679,8 +1679,130 @@ namespace Microsoft.Management.Deployment
...
+    /// IMPLEMENTATION NOTE: Pinning::Pin from AppInstaller::Pinning
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 29)]
+    runtimeclass PackagePin
+    {
+        /// The package ID that the pin applies to (for available-package pins) or the
+        /// ProductCode/PackageFamilyName (for installed-package pins).
+        String PackageId { get; };
+
+        /// The source identifier the pin applies to. Empty for installed-package pins.
+        String SourceId { get; };
```
JohnMcPMS: Feels like it should be a `PackageCatalogReference`.  Should not use the word "source".

src/Microsoft.Management.Deployment/PackagePin.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "pch.h"
+#include "PackagePin.h"
+#include "PackagePin.g.cpp"
+#include <wil\cppwinrt_wrl.h>
+
+namespace winrt::Microsoft::Management::Deployment::implementation
+{
+    namespace
+    {
+        winrt::Microsoft::Management::Deployment::PackagePinType ConvertPinType(::AppInstaller::Pinning::PinType type)
```
JohnMcPMS: Move to Converters

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1488,5 +1495,268 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        {
+            std::vector<::AppInstaller::Pinning::PinKey> pinKeys;
+
+            // Keys for each available source version
+            for (auto const& versionId : package.AvailableVersions())
+            {
+                auto versionInfo = package.GetPackageVersionInfo(versionId);
+                if (versionInfo)
+                {
+                    std::string packageId = winrt::to_string(package.Id());
```
JohnMcPMS: Must use the `versionInfo` package identifier.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1488,5 +1495,268 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        winrt::Microsoft::Management::Deployment::CatalogPackage package,
+        winrt::Microsoft::Management::Deployment::PinPackageOptions options)
+    {
+        LogStartupIfApplicable();
+
+        HRESULT terminationHR = S_OK;
+        try
+        {
+            THROW_HR_IF_NULL(E_POINTER, package);
+            THROW_HR_IF_NULL(E_POINTER, options);
```
JohnMcPMS: Invariants based on parameters or security should be function errors, not wrapped in a result.  You call it wrong, you don't get a nice response.

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1488,5 +1495,268 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            {
+                note = winrt::to_string(options.Note());
+            }
+
+            for (const auto& pinKey : pinKeys)
+            {
+                auto newPin = CreatePinFromOptions(pinKey, options);
+
+                auto existingPin = pinningData.GetPin(pinKey);
+                if (existingPin && !(*existingPin == newPin))
```
JohnMcPMS: Shouldn't this only be comparing the pin type rather than the entire pin?

src/Microsoft.Management.Deployment/PackageManager.cpp
```diff
@@ -1488,5 +1495,268 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+            auto pinKeys = GetPinKeysForCatalogPackage(package, /* includeInstalled */ true);
+            auto pinningData = ::AppInstaller::Pinning::PinningData{ ::AppInstaller::Pinning::PinningData::Disposition::ReadWrite };
+
+            bool anyRemoved = false;
+            for (const auto& pinKey : pinKeys)
+            {
+                auto existingPin = pinningData.GetPin(pinKey);
+                if (existingPin)
+                {
+                    pinningData.RemovePin(pinKey);
```
JohnMcPMS: Can we not just `RemovePin`?  If it throws on not existing, can we add `TryRemovePin` that returns a bool instead of throwing based on existence?

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/PSObjects/PSPackagePinType.cs
```diff
@@ -0,0 +1,29 @@
...
+//     Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+// </copyright>
+// -----------------------------------------------------------------------------
+
+namespace Microsoft.WinGet.Client.PSObjects
+{
+    /// <summary>
+    /// Must match the user-settable values of Microsoft.Management.Deployment.PackagePinType.
+    /// </summary>
+    public enum PSPackagePinType
```
JohnMcPMS, line 12: Do we need the "pinned by the manifest" type?

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddPinCmdlet.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// <summary>
+    /// Adds a pin for a package. Mirrors the behavior of <c>winget pin add</c>.
+    /// </summary>
+    [Cmdlet(
+        VerbsCommon.Add,
+        Constants.WinGetNouns.Pin,
+        DefaultParameterSetName = Constants.FoundSet,
+        SupportsShouldProcess = true)]
+    [OutputType(typeof(PSPinResult))]
+    public sealed class AddPinCmdlet : PackageCmdlet
```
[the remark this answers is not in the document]
JohnMcPMS: Yes, PS tests please.

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddPinCmdlet.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        private PinPackageCommand command = null;
+
+        /// <summary>
+        /// Gets or sets the pin type. Defaults to <see cref="PSPackagePinType.Pinning"/>.
+        /// </summary>
+        [Parameter(ValueFromPipelineByPropertyName = true)]
+        public PSPackagePinType PinType { get; set; } = PSPackagePinType.Pinning;
```
JohnMcPMS: Can you create parameter sets based on specific values for a parameter?  Could associate Gated with GatedVersion.

src/AppInstallerRepositoryCore/AppInstallerRepositoryCore.vcxproj
Trenly: Both schema versions contain a file named PinTable.cpp. MSBuild derives object file names from the source filename only — without the override, both would compile to $(IntDir)PinTable.obj, causing a link-time collision. The override routes the 1.1 object to a distinct subdirectory resolving the conflict

src/Microsoft.Management.Deployment/PackageManager.cpp
Trenly: I don't think so. The CLI compares the version as well for gating pins so that updating a gating pin by overwriting it requires `--force` .  e.g `winget pin add Microsoft.WingetCreate --version 1.9` and then running `winget pin add Microsoft.WingetCreate --version 1.10` would give an error that there is already a pin and it needs to be overriden.

This matches the behavior of the CLI

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/PSObjects/PSPackagePinType.cs
Trenly: I don't believe we do, since it isn't user-settable and the CLI doesn't expose it when listing pins as far as I can tell

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddPinCmdlet.cs
Trenly: Not that I know of (and seems 🤖 doesn't think so either). However, that did bring to mind a cleaner way to have parity with the CLI using `-Blocking` and `-GatedVersion`

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddPinCmdlet.cs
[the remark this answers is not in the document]
Trenly: Will do before requesting re-review; Changes will likely be pushed ahead of this

src/AppInstallerCLICore/Commands/PinCommand.cpp
Trenly: Yes, we would need a whole new search mechanism anyways. I do think that there is some overlap with `winget list <query> --details`, but to me it felt more akin to `winget show <query>`.  The semantics of `list` vs `show` as a base command don't really apply to pins. Where `show` in the base command is for a remote package / manifest information, `list` is for installed applications.

I will think on this and probably end up refactoring to extend the behavior of list instead of adding show

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
Trenly: Done, and gave the 🤖  some additional guidance in the instructions file

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable.cpp
Trenly: I decided to go for an alternate assignment function to help the 🤖

src/AppInstallerRepositoryCore/AppInstallerRepositoryCore.vcxproj
Trenly: Will change the file name for clarity

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/PSObjects/PSPackagePinType.cs
JohnMcPMS: It feels like a thing that we probably should expose everywhere, but that can be a separate change.

.github/copilot-instructions.md
```diff
@@ -133,6 +133,105 @@ void WorkflowTask(Execution::Context& context)
 - Parsing in `AppInstallerCommonCore/Manifest/`
 - Multi-file manifests: installer, locale, version, defaultLocale
 
+## SQLite Statement Builder
```
JohnMcPMS: This should go into a path specific instructions file (with the appropriate target paths).  It is a lot of lines that are only needed in a few cases.

.github/copilot-instructions.md
```diff
@@ -133,6 +133,105 @@ void WorkflowTask(Execution::Context& context)
...
+### Nullable values: `Equals()` vs `AssignValue()`
+
+Both accept `std::optional<T>`, but they behave differently when the optional is empty and must be used in the right context:
+
+| Method | Empty optional emits | Use in |
+|---|---|---|
+| `Equals(optional<T>)` | `IS NULL` | **WHERE** / filter clauses |
+| `AssignValue(optional<T>)` | `= ?` (binds NULL) | **UPDATE SET** assignments |
+
+Using `Equals(optional)` in an UPDATE SET clause is a bug — SQLite does not accept `col = IS NULL`.
```
JohnMcPMS:
```suggestion
Using `Equals(optional)` in an UPDATE SET clause is a bug — SQLite does not accept `col IS NULL`.
```

src/AppInstallerCLITests/PinningIndex.cpp
```diff
@@ -2,10 +2,12 @@
 // Licensed under the MIT License.
 #include "pch.h"
 #include "TestCommon.h"
+#include "PinTestCommon.h"
+#include <AppInstallerDateTime.h>
 #include <Microsoft/PinningIndex.h>
 #include <Microsoft/Schema/IPinningIndex.h>
 #include <Microsoft/Schema/Pinning_1_0/PinTable.h>
-#include <Microsoft/Schema/Pinning_1_1/PinTable.h>
+#include <Microsoft/Schema/Pinning_1_1/PinTable_1_1.h>
```
JohnMcPMS: We can definitely leave the header as `PinTable.h`.

src/AppInstallerCLITests/SQLiteWrapper.cpp
```diff
@@ -54,7 +54,7 @@ void InsertIntoSimpleTestTable(Connection& connection, int firstVal, const std::
 void UpdateSimpleTestTable(Connection& connection, int firstVal, const std::string& secondVal)
 {
     Builder::StatementBuilder update;
-    update.Update(s_tableName).Set().Column(s_firstColumn).Equals(firstVal).Column(s_secondColumn).Equals(secondVal);
+    update.Update(s_tableName).Set().Column(s_firstColumn).AssignValue(firstVal).Column(s_secondColumn).AssignValue(secondVal);
```
JohnMcPMS: New test that provides an empty optional?

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_0/PinningIndexInterface.h
```diff
@@ -19,5 +19,11 @@ namespace AppInstaller::Repository::Microsoft::Schema::Pinning_V1_0
         std::optional<Pinning::Pin> GetPin(SQLite::Connection& connection, const Pinning::PinKey& pinKey) override;
         std::vector<Pinning::Pin> GetAllPins(SQLite::Connection& connection) override;
         bool ResetAllPins(SQLite::Connection& connection, std::string_view sourceId) override;
+
+    protected:
+        virtual SQLite::rowid_t IAddPin(SQLite::Connection& connection, const Pinning::Pin& pin);
```
JohnMcPMS: The proper mechanism would be:
```
V1_1::AddPin(pin)
{
savepoint_1_1;
pin_row = V1_0::AddPin(pin);
modify_new_pin_with_new_fields(pin_row, pin);
savepoint_1_1.commit();
}
```

This is certainly inefficient, but it ensures the proper separation of schema concerns.  So a bug fix in 1_0 doesn't have to be duplicated across the range.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_0/PinTable.h
```diff
@@ -39,5 +36,11 @@ namespace AppInstaller::Repository::Microsoft::Schema::Pinning_V1_0
         // Resets all pins from a given source, or from all sources if none is specified.
         // Returns a value indicating whether there were any changes.
         static bool ResetAllPins(SQLite::Connection& connection, std::string_view sourceId = {});
+
+    protected:
+        // Creates the table with named indices.
+        static void Create(SQLite::Connection& connection);
```
JohnMcPMS: Why is this protected?

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable_1_1.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+            result->SetDateAdded(std::move(dateAdded));
+            result->SetNote(std::move(note));
+
+            return result;
+        }
+    }
+
+    using namespace std::string_view_literals;
+    static constexpr std::string_view s_PinTable_Table_Name = "pin"sv;
```
JohnMcPMS: These here and 1.0 should be moved to under the `PinTable` type so that they are not defined multiple times.

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -241,18 +236,14 @@ namespace AppInstaller::Repository::Microsoft
...
                 }
             }
-            else
-            {
-                // Read-only open: use an interface matching the stored schema version to avoid querying missing columns.
-                AICLI_LOG(Repo, Info, << "Read-only open with older schema [" << m_version << "]; using compatible interface");
-                m_interface = CreateIPinningIndexForVersion(m_version);
-            }
+
+            m_interface = std::move(latestInterface);
```
JohnMcPMS: Should be inside the above `if` block (is a functional no-op when they are already the same except for a tiny bit of move overhead).

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddPinCmdlet.cs
```diff
@@ -61,6 +73,30 @@ public sealed class AddPinCmdlet : PackageCmdlet
         /// </summary>
         protected override void ProcessRecord()
         {
+            if (this.Blocking && !string.IsNullOrEmpty(this.GatedVersion))
```
JohnMcPMS, line 76: Can we do this through parameter sets?  Not sure if it is more complicated or by how much, but it would better integrate with PS (at least I assume that once it has decided which set you are using on the shell it won't offer tab-suggestions from the other sets).

src/PowerShell/Microsoft.WinGet.Client.Engine/Commands/FinderPackageCommand.cs
```diff
@@ -82,9 +84,21 @@ public void Get(string psPackageFieldMatchOption)
                 () => this.FindPackages(
                     CompositeSearchBehavior.LocalCatalogs,
                     PSEnumHelpers.ToPackageFieldMatchOption(psPackageFieldMatchOption)));
+
+            if (results.Count == 0)
+            {
+                return;
+            }
+
+            // Fetch all pins in a single COM call and build a lookup set to avoid
```
JohnMcPMS: This feels like it is shifting the cost up front; a cost that doesn't need to be paid when the caller doesn't examine the `IsPinned` property.  This isn't a single COM call either, ever `PackageId` property retrieval is also one.  I think it would be better to just cache the `IsPinned` state from the previous change lazily rather than forcibly populating it.

JohnMcPMS (commented):
I would like to see the PinTable stuff be thought through more to re-use code in a way that allows the 1.0 code to stay largely untouched and 1.0 to leverage it without anything over.

src/PowerShell/Microsoft.WinGet.Client.Cmdlets/Cmdlets/AddPinCmdlet.cs
Trenly: I took a look and didn't see an easy way to do that using parameter sets. I'll keep digging though

Trenly:
Will address the comments in a few days - been super busy; 

@JohnMcPMS - Assuming this will go in vNext, does that mean I should then be bumping to the next contract version?

JohnMcPMS:
> Will address the comments in a few days - been super busy;
> 
> @JohnMcPMS - Assuming this will go in vNext, does that mean I should then be bumping to the next contract version?

Yes, it should move to 30 as we are branching the 1.29 RC imminently.

.github/copilot-instructions.md
Trenly: The current wording is correct. The bug in `Update().Set().Equals()` is that when it is null the resulting SQL is `col = IS NULL` when used for assignment. When used for simply checking equality (not in an UPDATE SET) it results in `col IS NULL` which is correct

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_0/PinTable.h
Trenly: It didn't explicitly need to be public and my 🤖 thought it would be better protected, but I agree that it's an unnecessary change - will revert

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -95,6 +108,37 @@ namespace AppInstaller::CLI::Workflow
 
             return searchRequest;
         }
+
+        bool IsMatchForListQuery(const Pinning::Pin& pin, Execution::Context& context)
```
JohnMcPMS, line 112: Feels like this should only cover ID filtering as it is something we actually have in `Pinning::Pin`.  That would leave name and query filtering to be done in the loop that creates the row data.  While that would be far less efficient, I think it would be preferable to the fake name search that happens here.

src/AppInstallerCLICore/Workflows/PinFlow.cpp
```diff
@@ -292,18 +360,76 @@ namespace AppInstaller::CLI::Workflow
...
+            for (const auto& row : rowData)
+            {
+                table.OutputLine({ row.PackageName, row.PackageId, row.Version, row.SourceName, row.PinType, row.PinnedVersion });
+            }
+
+            table.Complete();
+        }
+        else
+        {
+            Execution::TableOutput<8> table(context.Reporter,
```
JohnMcPMS: I don't think that `--details` should produce a table with more columns.  One of the reasons for details view is to see the whole values, and the table system will truncate as needed to keep the one line per row.  It would be better to go to the layout with each value on its own line(s).

src/AppInstallerCLITests/PinFlow.cpp
```diff
@@ -198,4 +203,278 @@ TEST_CASE("PinFlow_ResetEmpty", "[PinFlow][workflow]")
...
+    OverrideForCompositeInstalledSource(listContext, CreateListTestSource({ "NoNote.Package" }));
+
+    PinListCommand pinList({});
+    pinList.Execute(listContext);
+    INFO(listOutput.str());
+
+    REQUIRE_FALSE(listContext.IsTerminated());
+    REQUIRE(listOutput.str().find("NoNote.Package") != std::string::npos);
+    REQUIRE(listOutput.str().find(Resource::LocString(Resource::String::PinDateAdded)) != std::string::npos);
+    REQUIRE(listOutput.str().find(Resource::LocString(Resource::String::PinNote)) != std::string::npos);
```
JohnMcPMS: I thought that the table system will remove columns with no values...

src/AppInstallerRepositoryCore/PinningData.cpp
```diff
@@ -93,6 +93,17 @@ namespace AppInstaller::Pinning
         m_database->RemovePin(pinKey);
     }
 
+    bool PinningData::TryRemovePin(const PinKey& pinKey)
```
JohnMcPMS, line 96: This is still subject to a race, but I imagine you wanted it as a convenience method, not as a "never throws".  The problem is that the next person to use it may think that it can never fail (if preconditions are met).  It would be better if the database itself had a `TryRemovePin` that used a transaction to ensure atomicity of the get/remove pair.

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -184,29 +185,64 @@ namespace AppInstaller::Repository::Microsoft
...
+                bool migrated = latestInterface->MigrateFrom(m_dbconn, m_interface.get());
+
+                if (migrated)
+                {
+                    latestInterface->GetVersion().SetSchemaVersion(m_dbconn);
+                    SetLastWriteTime();
+                    savepoint.Commit();
+                    m_version = latestInterface->GetVersion();
+                    m_interface = std::move(latestInterface);
+                    AICLI_LOG(Repo, Info, << "Migration successful");
```
JohnMcPMS:
```suggestion
                    AICLI_LOG(Repo, Info, << "... migration successful");
```
For these types of followup logs, I prefer something that suggests it is related to a previous line. Apply to the failed as well.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable_1_1.cpp
```diff
@@ -0,0 +1,170 @@
...
+        using namespace SQLite::Builder;
+
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "migratepintable_v1_0_to_v1_1_pintable");
+
+        StatementBuilder addDateAdded;
+        addDateAdded.AlterTable(s_PinTable_Table_Name).Add(s_PinTable_DateAdded_Column, Type::Integer);
+        addDateAdded.Execute(connection);
+
+        StatementBuilder addNote;
+        addNote.AlterTable(s_PinTable_Table_Name).Add(s_PinTable_Note_Column, Type::Text);
```
JohnMcPMS, line 70: Can't you add multiple columns at once?

src/AppInstallerRepositoryCore/Microsoft/PinningIndex.cpp
```diff
@@ -184,29 +185,64 @@ namespace AppInstaller::Repository::Microsoft
...
+                }
+                else
+                {
+                    AICLI_LOG(Repo, Error, << "Migration failed");
+                    THROW_HR(APPINSTALLER_CLI_ERROR_CANNOT_WRITE_TO_UPLEVEL_INDEX);
```
JohnMcPMS: I'm not sure that a failed migration is/should be fatal here.  At a minimum, it seems like there should be metadata that we can use to determine that rather than assuming it in all cases.  Or maybe even a more minimum, if they have the same major version then we didn't think the new data was critical previously so why would it be now.

For this specific change, a migration failure would mean a loss of pin write times and notes.  Then presumably some potential confusion/investigation when the notes were silently lost (could easily be logged, more difficult but possible to tell the user).

If we fail on the migration, then the user can no longer write pins until we resolve the issue.  It is more plainly put in our face of course, but also much more impactful on them.

So it is a balancing act of likelihood of failure and impact of the choice in all directions.  My gut says to simply log it and let this move forward.  I would rather discover and fix the problem slower than force some user to be completely blocked.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_0/PinningIndexInterface.h
```diff
@@ -10,13 +10,21 @@ namespace AppInstaller::Repository::Microsoft::Schema::Pinning_V1_0
         // Version 1.0
         SQLite::Version GetVersion() const override;
         void CreateTables(SQLite::Connection& connection) override;
+        bool MigrateFrom(SQLite::Connection& connection, const IPinningIndex* current) override;
 
-    private:
         SQLite::rowid_t AddPin(SQLite::Connection& connection, const Pinning::Pin& pin) override;
+
+    private:
```
JohnMcPMS: Aren't these all the actual public interface?  Not sure why they were ever private.

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinningIndexInterface_1_1.cpp
```diff
@@ -0,0 +1,71 @@
...
+        return true;
+    }
+
+    SQLite::rowid_t PinningIndexInterface::AddPin(SQLite::Connection& connection, const Pinning::Pin& pin)
+    {
+        SQLite::Savepoint savepoint = SQLite::Savepoint::Create(connection, "addpin_v1_1");
+
+        Pinning_V1_0::PinningIndexInterface base;
+        SQLite::rowid_t pinId = base.AddPin(connection, pin);
+        THROW_HR_IF(E_UNEXPECTED, !Pinning_V1_1::PinTable::UpdatePinById(connection, pinId, pin));
```
JohnMcPMS: This feels like doing things twice, and I don't like the secondary protected interface design here.

This code invokes the base add, which in turn invokes its protected interface add, which in turn invokes our 1.1 add, which should just do the right thing.  Then it updates the pin, seemingly with no changes since our 1.1 add should have done the right thing.

The protected interface portions that we use in other schema design is to allow smaller portions of the overall action to be changed.  This change is using that to replace the entire action portion, leaving only some bookkeeping happening in the primary function.

I would revert the protected interface portions in favor of:
1. 1.1 invokes the 1.0 function
2. 1.0 function returns value(s) needed by the 1.1 function to make further updates
3. 1.1 function applies its updates

For step 2 to work, you may have to move the actual body of the interface implementation into a protected, non-virtual method that returns a non-interface result.

For the read-only methods, you can probably get away with deferring to the table without a need to invoke the base.  The table might invoke its base/shared static that takes in a column list so that at least some amount of code can be reused.

JohnMcPMS (commented):
Just a note for the future, please build things in layers 🥲 :

1. CLI
2. COM
3. PowerShell

That will help with faster reviews and iteration when we can focus on each part.

Trenly:
> 1. CLI
> 2. COM
> 3. PowerShell
> 
> That will help with faster reviews and iteration when we can focus on each part.

Apprecate the advice and will do for the future (And thank you for your patience with me and my new friend 🤖 )

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinTable_1_1.cpp
Trenly: I'm not seeing any examples in the codebase where that's done today - perhaps I'm just missing it?

src/AppInstallerCLITests/PinFlow.cpp
Trenly: This appears to be accurate - will update

src/AppInstallerCLICore/Workflows/PinFlow.cpp
Trenly: This is why I originally went with `winget pin show` similar to `winget show` - so that details about an individual pin could be shown line by line, similar to `winget list` vs `winget show` where `list` shows the packages installed and `show` gives the details line by line.

I'm going to change the behavior here to always show all columns, rendering `--details` moot. Since this PR is large enough already, if we decide that we want an implmentation for `--details` or `pin show` that can be raised separately

src/AppInstallerRepositoryCore/Microsoft/Schema/Pinning_1_1/PinningIndexInterface_1_1.cpp
Trenly: I'm not fully understanding - Are you saying to defer all the add/get/update to the tables? So then something like `UpdatePin` in `1_1` would call `1_0` implementation of ` UpdatePin()` to update the base pin information, the `1)0` implementation would return the row key that `1_1` needs, and then `1_1` would update just the new fields?

JohnMcPMS: Yes, and similarly Add_1_1 would call Add_1_0, then update the 1_1 fields for the rowid it returned.

JohnMcPMS:
Looking good (sorry about the slow feedback).  I'm wondering if we should find some other method to record changes than a single file that commonly produces conflicts.

Trenly:
> Looking good (sorry about the slow feedback). I'm wondering if we should find some other method to record changes than a single file that commonly produces conflicts.

Thats what I've been wondering. . .

PR #6177, opened 2026-04-25
[Settings, ListCommand] Add sort types, settings infrastructure, and CLI arguments for output ordering

Madhusudhan-MSFT:
## Summary

Add settings infrastructure and CLI arguments for configurable output sorting in
`winget list`. This lays the full type system, settings plumbing, and argument
registration without changing any runtime behavior — sort logic is wired in a
follow-up PR.

Part of https://github.com/microsoft/winget-cli/issues/4238

### Settings types and schema
- **SortField and SortDirection enums**: Add `SortField` (Relevance, Name, Id, Version,
  Source, Available) and `SortDirection` (Ascending, Descending) to the settings type
  system. Relevance preserves the current natural ordering and is publicly configurable
  (users can set `["relevance"]` to explicitly opt out of sorting).
- **Settings entries**: Register `OutputSortOrder` and `OutputSortDirection` with typed
  validation using lowercase-once comparison (`Utility::ToLower` + `==`). Default sort
  order is empty (`{}`) — an empty vector acts as a sentinel meaning "use context-aware
  defaults" (the follow-up PR's sort logic decides: relevance for queries, Name for
  no-query).
- **JSON schema**: Extend `settings.schema.0.2.json` with an `output` section containing
  `sortOrder` (string array with relevance, name, id, version, source, available) and
  `sortDirection` (string enum with ascending default).
- **Static constexpr pattern**: String constants for field name and direction comparisons
  follow the established `static constexpr std::string_view` pattern used by existing
  validators.

### CLI arguments
- **Execution args**: Add `Sort`, `SortAscending`, and `SortDescending` arg types to
  `ExecutionArgs.h`.
- **Argument registration**: Define `--sort` (multi-use, max 7 values), `--ascending`/`--asc`,
  and `--descending`/`--desc` in `Argument.cpp` with resource string descriptions.
  `--ascending` and `--descending` are mutually exclusive via `ExclusiveOf`.
- **ListCommand**: Add the three new arguments to `ListCommand::GetArguments()`.
- **Resource strings**: Add localized descriptions for all three arguments in
  `winget.resw` and resource IDs in `Resources.h`.

## Impact

- No behavioral change — settings and arguments are defined but not yet consumed by
  any workflow. Existing `winget list` output remains unchanged.
- `winget list --sort name` is accepted without error but output remains unsorted until
  the sort logic PR lands.
- `--ascending` and `--descending` are mutually exclusive — passing both produces an error.

## How Validated

[...]

Issue #1769, opened 2021-12-02
[New Feature]: Customizable sorting for search output


denelon:
### Description of the new feature / enhancement

I would like to be able to customize how the output from search is displayed.

Today, packages are grouped based on the order the source is configured.

The results from a REST source are returned in the order they are received.
The results from the Microsoft Windows Package Manager App Repository are returned based on where the match was found.

### Proposed technical implementation details

Allow a user to configure whether results are grouped by source or coalesced together.
> This assumes the default sort criteria from Windows Package Manager for coalesced results. The quality of the manifest from a REST source will impact these results.

Allow a user to order results alphabetically.
> This may be performed on package name or package identifier.

If the output is coalesced, and there are duplicates, the order of two equally ranked results will be ordered by source priority. This should be driven by source priority from:

* #1488

mstevens83:
Please also implement sorting (by name, id, source, ...) for the `list` command.

dimension-zero:
Mega up-vote

Trenly:
[Policy] Area-Output

AnLAdmin:
Here's a workaround:
`$nameVar = winget list ; $namevar[4..($namevar.length)] | Sort-Object | Format-Table`

But a integrated sort ability would be nice.

denelon:
Just an update since I was checking related issues. We've got the implementation for sorting the output of list in progress.
* https://github.com/microsoft/winget-cli/pull/6191

denelon:
**Issue summary**

- Requests customizable sorting for `winget search` output (grouped-by-source vs. coalesced, alphabetical by name/id, tie-break by source priority).

Closing as a duplicate of #142, which covers the same customizable-sort request. I've consolidated the specifics from this issue (grouped-vs-coalesced display, alphabetical ordering by name/id, and duplicate tie-breaking by source priority via #1488) into #142, which is now scoped to the remaining `winget search` work. Please follow along there.

**Related issues**

- #142
- #1488

---
*Drafted with the help of GitHub Copilot.*

Issue #1352, opened 2021-08-06
Output of the UPGRADE command should be sorted alphabetically


danstis:
# Description of the new feature/enhancement

The output seems to be in no meaningful order when running `winget upgrade` to list all packages with updates. 
I believe it would be easier to read the output if this was sorted alphabetically by Name or Id.

I understand the similarity to this issue #142 however I believe the upgrade command does not have to sort by relevancy, as the packages are already installed.

# Proposed technical implementation details (optional)

The output of the `winget upgrade` command to be sorted alphabetically by Name,

Karl-WE:
Please consider the same for Winget install.
- eventually include more sorting options like in PowerShell Sort-Table
new perferred default for install / upgrade:

sort by name ascending (alphabetically not alphanumerically


add a new option for sorting

**Example 1:**
`winget install --sort [ColumnName] `

[ColumnName]: Name, ID, Version, InstallDate (respect windows regional settings)

as a sub command of --sort
`--ascending (default if not specified) / --descending `


as a sub command of --sort
`--numeric / --alphabetical (default if not specified)`

**Example 2:**

`winget install --sort Version --descending --numeric`
**result:** lists all installed packages found in winget repo sorted by column name descending (numerically)

**Example 3:**
`winget  upgrade --sort Name --ascending --alphabetically `

aka 

`winget upgrade --sort `

(Name --acending --alphabetically could be ommited as they are default)

**result:** lists all upgradeable packages found in winget repo sorted by column name ascending alphabetically


@denelon I feel like this there might be thoughts about what will be the short command for --sort as it cannot be -s for certain conflicts (foremost logical -s should mean silent)

However as we see --tags has no short command I do not think that --sort needs a short command.

denelon:
Did you mean `winget list` for Example 1?

Upgrade makes sense.

Yes, I've got a few different suggestions for the single character shortcuts. I think this would be breaking, so renaming things might be a 2.0 milestone. For the sake of clarity, I wouldn't necessarily wait for a specific date to have a breaking [semver](https://semver.org) version. We're heads down on 1.1 right now. 

There are some other defaults people have asked for like "silent" install vs. "silent with progress". I think the upcoming PowerShell nouns and verbs may also influence the future direction in some cases.

Karl-WE:
Oh it was late. Certainly meant list, not install. Thank you for spotting this @denelon
Corrected the examples.

Karl-WE:
[...]

PR #6191, opened 2026-04-29
Implement sort logic for winget list output

### LINKED ISSUE #4238: Enable sorting the output of WinGet List

Madhusudhan-MSFT:
## Summary

Implements the sort logic for `winget list` output, completing the feature requested in #4238. Builds on PR #6177 (settings parsing + CLI handling).

### Changes

**Sort orchestration** (`WorkflowBase.cpp`):
- `SortInstalledPackagesTableLines` — resolves sort fields from CLI `--sort` > settings `output.sortOrder` > query-aware default, determines direction, sorts via `std::stable_sort`, applies permutation back.
- When the free-text query (`-q`) is used and no sort preference is configured, relevance ordering is preserved. If `output.sortOrder` is configured, it is respected even with queries.
- `FAIL_FAST_MSG` for unrecognized sort field values — ensures invalid enum values crash loudly in debug builds.

**Sort helper** (`PackageTableSortHelper.h/.cpp`):
- `SortablePackageEntry` — holds pre-computed case-folded sortable fields, computed only for fields needed via `ComputeSortFieldMask` bitmask optimization.
- `CompareByField` — case-insensitive ordinal comparison using pre-computed `FoldCase` values.
- `SortBy` — template sort executor used by the workflow; standalone `SortEntries` wrapper available for unit tests.
- `SortField` enum uses flag-bit values (`uint32_t`) with `DEFINE_ENUM_FLAG_OPERATORS` for efficient bitmask field tracking.

**CLI integration** (`ListCommand.cpp`):
- `--sort` (repeatable, limit 6), `--ascending`/`--asc`, `--descending`/`--desc` arguments.

**Argument validation** (`Command.cpp`):
- Validates `--sort` values in `ValidateArguments`, following the `--scope`/`--authentication-mode` pattern.

**Documentation** (`list.md`):
- `--sort`, `--ascending`, `--descending` in options table. New "Sorting output" section.

### ⚠️ Behavior change: default sort order

Previously, `winget list` output preserved the source's natural ordering (source-determined relevance ranking). **This PR changes the default to sort alphabetically by name (ascending)** when no `--sort` argument, no `-q` query, and no `output.sortOrder` setting is configured.

| Scenario | Behavior |
|----------|----------|
| No query, no settings, no `--sort` | Sorts by name (ascending) — **new default** |
| `-q` query with no sort preference | Preserves source relevance ordering (unchanged) |
| `--sort relevance` or `"sortOrder": ["relevance"]` | Preserves source ordering explicitly |
| `"sortOrder": []` (empty array) | Same as no setting — applies default name sort |
| `output.sortOrder` configured + `-q` query | Settings override relevance (explicit user preference wins) |

**To restore previous (source-determined) ordering:**
- CLI: `winget list --sort relevance`
- Settings: `{ "output": { "sortOrder": ["relevance"] } }`

> **Note:** An empty `sortOrder` array is equivalent to the setting not being present — it does not disable sorting. Use `"relevance"` explicitly to preserve source ordering.

### Design decisions

| Decision | Rationale |
|----------|-----------|
| **Default sort by name** | When no settings and no query, sort alphabetically by name — most natural default for browsing installed packages. |
| **Relevance preservation with queries** | Free-text query (`-q`) without configured sort preserves source relevance ranking. Settings override this if configured — explicit user preference takes priority. |
| **`stable_sort`** | Preserves relative order for multi-field sorting (e.g., `--sort source --sort name` groups by source, alphabetical within). |
| **Flag-bit `SortField` enum** | Enables `ComputeSortFieldMask` to skip unused field extraction — measurable win for large package lists. |
| **Pre-computed `FoldCase`** | Expensive case-folding done once per entry at construction, not on every comparison call. |
| **`FAIL_FAST_MSG` on invalid sort** | Catches enum drift at development time rather than silent misbehavior. |

### How validated

**Unit tests** — 20 test cases:
- 18 sort tests in `PackageTableSortHelper.cpp`: single/multi-field sorting, both directions, edge cases, case-insensitive comparison, relevance no-op
- 1 drift detection test in `Command.cpp`: reads `ListCommand::GetArguments()` limit and validates against known `ConvertToSortField` strings — breaks if enum changes without updating limit
- 1 enforcement test in `Command.cpp`: verifies `SetCountLimit(6)` rejects 7 `--sort` values with `TooManyArgError`

**Manual testing** — 21 scenarios on `wingetdev` (Debug x64): defaults, single/multi-field sorts, direction flags, query + sort interaction, settings precedence, edge cases, invalid input. All passing.

Fixes #4238


Trenly:
Is this only applicable to `winget list` or to any command that presents the user with a table of packages (upgrade, search, pin, etc.)

Trenly:
> Key behavior to note: Settings sortOrder is intentionally NOT applied when query/filter arguments are present. Only explicit --sort CLI arguments can override relevance ordering. The --descending flag applies to whichever sort fields are active (CLI or settings).

This seems like it would be counter to what a user would want. If I did `winget list Microsoft` and my default sort order in settings is `Name` I'd expect both the filter and the sorting to work

Madhusudhan-MSFT:
> Is this only applicable to `winget list` or to any command that presents the user with a table of packages (upgrade, search, pin, etc.)

The current implementation is scoped to `winget list` per the feature request in #4238. The sort infrastructure (schema, settings, argument definitions) was designed to be reusable, so extending to other table-outputting commands like `upgrade` or `search` would be straightforward if there's demand for it.

Madhusudhan-MSFT:
> This seems like it would be counter to what a user would want. If I did `winget list Microsoft` and my default sort order in settings is `Name` I'd expect both the filter and the sorting to work

This was discussed and agreed upon during the initial design. The intent is to preserve source relevance ordering for queries since a settings-based sort could push the best match down without the user realizing why. Explicit `--sort` always works with queries (e.g., `winget list Microsoft --sort name`).

@denelon — I believe this aligns with what we discussed during the design phase, would you mind confirming?

Trenly:
> if there's demand for it.

I think there is, but agree with the approach of limiting the scope for now and letting Demitrius figure out when to bring the rest in
* https://github.com/microsoft/winget-cli/issues/1769
* https://github.com/microsoft/winget-cli/issues/1352

JohnMcPMS:
> > This seems like it would be counter to what a user would want. If I did `winget list Microsoft` and my default sort order in settings is `Name` I'd expect both the filter and the sorting to work
> 
> This was discussed and agreed upon during the initial design. The intent is to preserve source relevance ordering for queries since a settings-based sort could push the best match down without the user realizing why. Explicit `--sort` always works with queries (e.g., `winget list Microsoft --sort name`).
> 
> @denelon — I believe this aligns with what we discussed during the design phase, would you mind confirming?

From the design portion, my feedback was that the ***default*** behavior should be to remain unchanged with a query but could have a new ordering without a query.  If the user does supply a setting value, my expectation would be that it would be used with or without a query unless overridden with a command line argument.

What constitutes a query?  Probably only the actual `query` argument, not `id` or other fields.  Relevance would largely by the same for those already.

Madhusudhan-MSFT:
> From the design portion, my feedback was that the ***default*** behavior should be to remain unchanged with a query but could have a new ordering without a query. If the user does supply a setting value, my expectation would be that it would be used with or without a query unless overridden with a command line argument.
>
> What constitutes a query? Probably only the actual `query` argument, not `id` or other fields. Relevance would largely by the same for those already.

Thanks for the clarification, John. Updated in `70bd157e`:

- **Default unchanged** — without `output.sortOrder` configured, positional `query` preserves relevance as before
- **Settings always honored** — if user configures `output.sortOrder`, it applies regardless of arguments (user explicitly opted in)
- **"Query" narrowed** — only the positional `query` argument (not `--id`/`--name`/`--tag`) is considered relevance-affecting in the no-settings case

Removed `HasRelevanceAffectingArgs` in favor of `sortFields.empty() && Args.Contains(Query)`.

doc/windows/package-manager/winget/list.md
```diff
@@ -70,6 +73,49 @@ The following example limits the output of list to 9 apps.
 
 ![list count command](images/list-count.png)
 
+## Sorting output
+
+By default, results are displayed in the order returned by the package source (no sorting applied). You can control sorting through command-line arguments or user settings.
```
JohnMcPMS: This depends on whether there is a query or not, correct?

src/AppInstallerCLICore/Commands/ListCommand.cpp
```diff
@@ -32,7 +32,9 @@ namespace AppInstaller::CLI
             Argument{ Execution::Args::Type::IncludeUnknown, Resource::String::IncludeUnknownInListArgumentDescription, ArgumentType::Flag },
             Argument{ Execution::Args::Type::IncludePinned, Resource::String::IncludePinnedInListArgumentDescription, ArgumentType::Flag},
             Argument::ForType(Execution::Args::Type::ListDetails),
-            Argument{ Execution::Args::Type::Sort, Resource::String::SortArgumentDescription, ArgumentType::Standard },
+            // 7 = one more than the number of user-facing SortField values (Relevance, Name, Id, Version, Source, Available),
```
JohnMcPMS: If there is going to be a limit, it should be the number of options.  A test that provides all of them can detect a failure to update that limit if more are added.

src/AppInstallerCLICore/Workflows/ListSortHelper.h
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+#include <winget/LocIndependent.h>
+#include <winget/UserSettings.h>
+
+#include <vector>
+
+namespace AppInstaller::CLI::Workflow
+{
+    // Lightweight sortable representation of a package line in list output.
+    // Decoupled from ICompositePackage/IPackageVersion to enable unit testing.
```
JohnMcPMS:
```suggestion
    // Decoupled from ICompositePackage/IPackageVersion to ease unit testing.
```
Claiming that interfaces make it impossible to unit test is a bit untrue.

src/AppInstallerCLICore/Workflows/ListSortHelper.cpp
```diff
@@ -0,0 +1,87 @@
...
+namespace AppInstaller::CLI::Workflow
+{
+    int CompareByField(const SortablePackageEntry& a, const SortablePackageEntry& b, Settings::SortField field)
+    {
+        using Settings::SortField;
+
+        switch (field)
+        {
+        case SortField::Name:
+            return Utility::FoldCase(std::string_view{ a.Name.get() }).compare(Utility::FoldCase(std::string_view{ b.Name.get() }));
```
JohnMcPMS: Operations like `FoldCase` are expensive, the `SortablePackageEntry` should contain precomputed comparable values for all fields rather than computing them every single time.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -459,6 +461,110 @@ namespace AppInstaller::CLI::Workflow
             }
         }
 
+        // Returns true if the execution context contains filter arguments that
```
JohnMcPMS: Doesn't appear to return a value.

src/AppInstallerCLICore/Workflows/ListSortHelper.cpp
```diff
@@ -0,0 +1,87 @@
...
+                if (vb < va) return 1;
+            }
+            return 0;
+        }
+        default:
+            return 0;
+        }
+    }
+
+    // Currently used in tests; production code uses CompareByField with an index-based
```
JohnMcPMS: Tests shouldn't be testing a test-only function.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -459,6 +461,110 @@ namespace AppInstaller::CLI::Workflow
...
+            if (hasExplicitSort)
+            {
+                for (const auto& arg : *context.Args.GetArgs(Execution::Args::Type::Sort))
+                {
+                    auto field = ConvertToSortField(arg);
+                    if (field)
+                    {
+                        sortFields.emplace_back(field.value());
+                    }
+                    // Invalid values should not reach here; ValidateArgumentsInternal
```
JohnMcPMS: Put an assert to validate the comment.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -459,6 +461,110 @@ namespace AppInstaller::CLI::Workflow
...
+
+                // When the free-text query argument is present and the user has NOT
+                // configured a sort preference in settings, preserve relevance ordering.
+                // Only the positional query argument populates searchRequest.Query and
+                // produces meaningful relevance ranking; filter arguments like --id,
+                // --name, --tag etc. use exact/substring matching where all results
+                // have equivalent relevance.
+                // If the user explicitly configured output.sortOrder, respect it even
+                // with queries — that is an explicit user preference.
+                if (sortFields.empty() && context.Args.Contains(Execution::Args::Type::Query))
```
JohnMcPMS: When `sortFields` is empty but no query is provided, shouldn't we be setting the actual default sort fields here?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -459,6 +461,110 @@ namespace AppInstaller::CLI::Workflow
...
+            {
+                direction = User().Get<Setting::OutputSortDirection>();
+            }
+
+            // 3. Project into SortablePackageEntry, sort, then apply the permutation back
+            std::vector<SortablePackageEntry> sortable;
+            sortable.reserve(lines.size());
+            for (const auto& line : lines)
+            {
+                sortable.push_back({ line.Name, line.Id, line.InstalledVersion, line.AvailableVersion, line.Source });
```
JohnMcPMS: From another comment, we actually want to construct all of the comparable properties, not just copy.  Have a constructor that does so.

Additionally, we could optimize by only constructing the properties that we would actually compare based on the `sortFields`.  If the enum values were converted to flag bits, it could be efficiently checked.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -459,6 +461,110 @@ namespace AppInstaller::CLI::Workflow
...
+            }
+
+            // Build index array to track the permutation
+            std::vector<size_t> indices(lines.size());
+            std::iota(indices.begin(), indices.end(), 0);
```
JohnMcPMS: Why not have the entries just contain their related index?  Then the separated sort function makes more sense (and can easily be called from this one).

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -3,11 +3,13 @@
 #include "pch.h"
 #include "WorkflowBase.h"
 #include "ExecutionContext.h"
-#include <winget/ManifestComparator.h>
+#include "ListSortHelper.h"
```
JohnMcPMS:
```suggestion
#include "PackageTableSortHelper.h"
```
Eventually not used exclusively for `list`.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -459,6 +461,110 @@ namespace AppInstaller::CLI::Workflow
             }
         }
 
+        // Returns true if the execution context contains filter arguments that
+        // produce relevance-ordered results (query, name, id, moniker, tag, command).
+        // Sorts a vector of InstalledPackagesTableLine according to the user's sort preferences.
+        // Resolution order: CLI args (--sort) > settings (output.sortOrder) > query-aware default.
+        void SortInstalledPackagesTableLines(Execution::Context& context, std::vector<InstalledPackagesTableLine>& lines)
```
JohnMcPMS, line 466: This function should absolutely be in the sort helper, templated.  And then it should be refactored to extract the sort fields and order determination out from the only part that needs to be templated: conversion from input vector to sortables, then sorting of the input vector via the indices.

doc/windows/package-manager/winget/list.md
florelis, line 1: Please add a mention of this feature to the [release notes](https://github.com/microsoft/winget-cli/blob/master/doc/ReleaseNotes.md)

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.h
```diff
@@ -0,0 +1,103 @@
...
+    int CompareByField(const SortablePackageEntry& a, const SortablePackageEntry& b, Settings::SortField field);
+
+    // Sorts a vector of sortable entries by the given fields and direction.
+    void SortEntries(
+        std::vector<SortablePackageEntry>& entries,
+        const std::vector<Settings::SortField>& sortFields,
+        Settings::SortDirection direction);
+
+    // Computes a bitmask of all sort fields so the constructor can skip unused fields.
+    inline Settings::SortField ComputeSortFieldMask(const std::vector<Settings::SortField>& sortFields)
```
florelis: nit: why does this need to be inline?

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.cpp
```diff
@@ -0,0 +1,111 @@
...
+        size_t originalIndex,
+        std::string_view name,
+        std::string_view id,
+        std::string_view installedVersion,
+        std::string_view availableVersion,
+        std::string_view source,
+        Settings::SortField fieldMask)
+        : OriginalIndex(originalIndex)
+    {
+        if (WI_IsAnyFlagSet(fieldMask, Settings::SortField::Name))
```
florelis: nit: if it's only one, use `WI_IsFlagSet`

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.h
```diff
@@ -0,0 +1,103 @@
...
+
+        SortEntries(entries, sortFields, direction);
+
+        std::vector<T> sorted;
+        sorted.reserve(items.size());
+        for (const auto& entry : entries)
+        {
+            sorted.push_back(std::move(items[entry.OriginalIndex]));
+        }
+        items = std::move(sorted);
```
florelis, line 106: Why not sort `items` directly? I imagine the idea is that we can avoid doing the case folding multiple times for each string if we do it all upfront, but I'm not sure if the performance improvement is worth the added complexity. The bottleneck is most likely going to be the search, and sorting at most a few hundred items here isn't going to take that long regardless of how slow each comparison is.

I would also think that having a case-insensitive comparison function could be way better in many cases. For example, if the first letters of each string are all different, we would end up only folding those, instead of the full strings.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -1226,6 +1320,7 @@ namespace AppInstaller::CLI::Workflow
             }
         }
 
+        SortInstalledPackagesTableLines(context, lines);
         OutputInstalledPackages(context, lines);
```
florelis: Would it make sense to do the sorting inside `OutputInstalledPackages` to avoid the repetition?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -459,6 +461,98 @@ namespace AppInstaller::CLI::Workflow
...
+                    }
+                }
+            }
+            else
+            {
+                sortFields = User().Get<Setting::OutputSortOrder>();
+
+                if (sortFields.empty())
+                {
+                    if (context.Args.Contains(Execution::Args::Type::Query))
```
florelis: Should this also check for `Type::MultiQuery`? `list` doesn't take it, but it may help us future-proof the sorting if we ever use it in more commands.

src/AppInstallerCLITests/Command.cpp
```diff
@@ -665,3 +667,55 @@ TEST_CASE("ParseArguments_PositionalWithTooManyValues", "[command]")
...
+        }
+    }
+    REQUIRE(sortLimit > 0);
+
+    // Every valid sort field string must convert successfully.
+    std::vector<std::string_view> allFieldNames = {
+        "relevance", "name", "id", "version", "source", "available"
+    };
+
+    REQUIRE(allFieldNames.size() == sortLimit);
```
florelis: This only ensures that the count limit matches the vector declared here. We still could easily add another sort field and update neither this list nor the count limit.

I think we have helper functions to get the list of all possible values of an enum that we could use here.

src/AppInstallerCLITests/PackageTableSortHelper.cpp
```diff
@@ -0,0 +1,328 @@
...
+    }
+    SECTION("Equal")
+    {
+        REQUIRE(CompareByField(a, a, SortField::Name) == 0);
+    }
+    SECTION("Case-insensitive")
+    {
+        auto upper = MakeEntry("ALPHA", "a.id", "1.0");
+        REQUIRE(CompareByField(a, upper, SortField::Name) == 0);
+    }
```
florelis, line 75: Would it be worth it to have all of these tests for each field?

src/AppInstallerCLITests/PackageTableSortHelper.cpp
```diff
@@ -0,0 +1,328 @@
...
+    std::vector<SortablePackageEntry> entries = {
+        MakeEntry("charlie", "c", "1.0"),
+        MakeEntry("ALPHA", "a", "1.0"),
+        MakeEntry("Beta", "b", "1.0"),
+    };
+
+    SortEntries(entries, { SortField::Name }, SortDirection::Ascending);
+
+    auto ids = GetIds(entries);
+    REQUIRE(ids == std::vector<std::string>{ "a", "b", "c" });
```
florelis: nit: why verify the ids instead of the names?

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.h
Madhusudhan-MSFT: Pre-fold was requested by JohnMcPMS in an earlier review round. At this scale (~100-200 items) the difference is negligible. Happy to revisit if profiling shows otherwise.

src/AppInstallerCLITests/PackageTableSortHelper.cpp
Madhusudhan-MSFT: Name covers the full Less/Greater/Equal/Case-insensitive pattern. Id and Source share the same code path so one case test suffices. Version and Available have distinct semantics with targeted cases. ValidateSortResult also asserts all fields per entry across every sort test.

src/AppInstallerCLITests/Command.cpp
Madhusudhan-MSFT: GetAllExponentialEnumValues requires an E::Max sentinel which SortField currently lacks. The current test provides the necessary baseline coverage; we can extend it to use the enum reflection helper in a follow-up once the sentinel is added.

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.h
JohnMcPMS: At 200 items, a merge sort would expect 2125 comparisons (`1.39 * n * log2(n)`).  Each comparison is two folds (4250), so the equal number of characters folds would occur at an average comparison length of 200 / 4250, or 4.7% of the average length.  If you have to compare more characters than that on average, pre-folding will be better.  At 100 items, this goes to 10.8%.  And this doesn't take into account the relative cost of `N` function calls vs `2 * 1.39 * log2(N)` function calls, which might be on the low end due to needing to potentially use the ICU break iterators to actually find the UTF-8 character sequences to individually fold.

I don't expect that we will have this many items regularly, but at the small numbers that are likely with a query I don't think any choice matters much.  I would rather optimize for the worst case even if the best case gets a little worse.

florelis (dismissed):

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.h
```diff
@@ -0,0 +1,95 @@
...
+
+        // Precomputed case-folded sort keys
+        std::string FoldedName;
+        std::string FoldedId;
+        std::string FoldedSource;
+
+        // Precomputed parsed versions
+        Utility::Version ParsedInstalledVersion;
+        Utility::Version ParsedAvailableVersion;
+        bool HasAvailableVersion = false;
```
JohnMcPMS:
```suggestion
        std::optional<Utility::Version> ParsedAvailableVersion;
```

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.cpp
```diff
@@ -0,0 +1,121 @@
...
+            if (HasAvailableVersion)
+            {
+                ParsedAvailableVersion = Utility::Version{ std::string{ availableVersion } };
+            }
+        }
+    }
+
+    Settings::SortField ComputeSortFieldMask(const std::vector<Settings::SortField>& sortFields)
+    {
+        Settings::SortField mask = Settings::SortField::Relevance;
```
JohnMcPMS: Why is `Relevance` always included?  There should be a `None` or similar for the 0 value.

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.cpp
```diff
@@ -0,0 +1,121 @@
...
+        {
+            if (a.ParsedInstalledVersion < b.ParsedInstalledVersion) return -1;
+            if (b.ParsedInstalledVersion < a.ParsedInstalledVersion) return 1;
+            return 0;
+        }
+        case SortField::Source:
+            return a.FoldedSource.compare(b.FoldedSource);
+        case SortField::Available:
+        {
+            if (a.HasAvailableVersion != b.HasAvailableVersion)
```
JohnMcPMS: Converting to an optional gets automatic empty vs not-empty comparison along with the value comparison.  Look it up to ensure it does what you want though, as it might put the empties at the wrong end from where you want.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
```diff
@@ -485,31 +483,42 @@ namespace AppInstaller::CLI::Workflow
                     {
                         sortFields.emplace_back(field.value());
                     }
-                    // Invalid values should not reach here; ValidateArgumentsInternal
-                    // rejects them with an error before workflow execution begins.
+                    else
+                    {
+                        // Invalid values should not reach here; ValidateArguments
+                        // rejects them with a CommandException before workflow execution begins.
+                        FAIL_FAST_MSG("Unexpected sort field value reached workflow; validation should have caught this.");
```
JohnMcPMS: nit: I was thinking an assert, not a process exit.  Is this the kind of thing that warrants that?

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
JohnMcPMS: nit: Eventually all of this "determine the sort fields and order" will end up in the helper file as well.

src/AppInstallerCLITests/Command.cpp
JohnMcPMS: I agree that this test does not cover anything useful.  You have ensured that a list defined in the test is as long as the argument limit.

src/AppInstallerCLITests/Command.cpp
```diff
@@ -665,3 +667,55 @@ TEST_CASE("ParseArguments_PositionalWithTooManyValues", "[command]")
...
+        auto field = Settings::ConvertToSortField(name);
+        REQUIRE(field.has_value());
+    }
+}
+
+TEST_CASE("ValidateArguments_StandardArgExceedsCountLimit", "[command]")
+{
+    Args args;
+    TestCommand command({
+            Argument{ "sort", 's', Args::Type::Sort, DefaultDesc, ArgumentType::Standard }.SetCountLimit(6),
```
JohnMcPMS: Similarly, this test seems to be checking that the count limit on arguments is implemented properly in the general sense.  I honestly don't care if the argument takes in additional values over the total number of fields, the user is just wasting their own compute time recomparing.  The only test that I care about existing is that you can provide all of the fields.  Removing the limit would ensure that was possible, but if we are going to have the limit, the test needs to generate the set dynamically and ensure that they can all be provided.

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -47,17 +47,20 @@ namespace AppInstaller::Settings
...
+    // via ComputeSortFieldMask, so the constructor can skip unused field computation.
+    enum class SortField : uint32_t
     {
-        Relevance,  // Preserves current natural order (source-defined relevance ranking)
-        Name,
-        Id,
-        Version,
-        Source,
-        Available,
+        Relevance = 0x0,   // Preserves current natural order (source-defined relevance ranking)
```
JohnMcPMS: As said in another comment, we shouldn't treat relevance as "none".

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.cpp
```diff
@@ -5,56 +5,121 @@
...
-    int CompareByField(const SortablePackageEntry& a, const SortablePackageEntry& b, Settings::SortField field)
+    SortParameters ResolveSortParameters(
+        const std::vector<std::string_view>& explicitSortArgs,
+        bool hasQuery,
+        bool hasExplicitAscending,
+        bool hasExplicitDescending)
     {
-        using Settings::SortField;
+        SortParameters result;
+        std::vector<SortField> sortFields;
```
JohnMcPMS: nit: why not just use `result.Fields` directly?

If the answer is allowing `[relevance]` to be returned, I don't see a problem.  It is true, and `ShouldSort` is what you expect anyone to look at first.

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.cpp
```diff
@@ -5,56 +5,121 @@
...
+        }
+
+        // Relevance-only means preserve source ordering.
+        if (sortFields.size() == 1 && sortFields[0] == SortField::Relevance)
+        {
+            return result; // ShouldSort = false
+        }
 
+        // Resolve direction
+        SortDirection direction = SortDirection::Ascending;
```
JohnMcPMS: nit: same here

For both of these it is more a style/preference thing as performance should be equivalent.  I just personally like removing the 4 lines of code but I'm open to hearing about reasons not to.  I can imagine a reason that is about first evaluating and then collecting the data independently for clarity.  I can also imagine a NRVO type argument (although maybe this should be the constructor of `SortParameters` if that is a concern).

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.cpp
```diff
@@ -71,35 +136,41 @@ namespace AppInstaller::CLI::Workflow
...
             return 0;
         }
     }
 
     void SortEntries(
         std::vector<SortablePackageEntry>& entries,
-        const std::vector<Settings::SortField>& sortFields,
-        Settings::SortDirection direction)
+        const std::vector<SortField>& sortFields,
+        SortDirection direction)
```
JohnMcPMS: `const SortParameters&`?

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.cpp
```diff
@@ -71,35 +136,41 @@ namespace AppInstaller::CLI::Workflow
...
     {
         if (entries.size() <= 1 || sortFields.empty())
         {
             return;
         }
 
         // Relevance-only means no sorting
-        if (sortFields.size() == 1 && sortFields[0] == Settings::SortField::Relevance)
+        if (sortFields.size() == 1 && sortFields[0] == SortField::Relevance)
```
JohnMcPMS: `sortParams.ShouldSort`?

src/AppInstallerCLICore/Workflows/PackageTableSortHelper.h
```diff
@@ -51,6 +51,28 @@ namespace AppInstaller::CLI::Workflow
...
+    };
+
+    // Resolves the effective sort parameters from explicit CLI args, user settings,
+    // and query context. Resolution order: explicit args > settings > query-aware default.
+    // Parameters:
+    //   explicitSortArgs - raw string values from --sort (empty if not provided)
+    //   hasQuery - whether a free-text query argument is present
+    //   hasExplicitAscending - whether --ascending was passed
+    //   hasExplicitDescending - whether --descending was passed
+    SortParameters ResolveSortParameters(
```
JohnMcPMS: Why not be a constructor of `SortParameters`?  You can leave the default constructor included as well if needed.

src/AppInstallerCLICore/Commands/ListCommand.cpp
```diff
@@ -32,7 +33,7 @@ namespace AppInstaller::CLI
             Argument{ Execution::Args::Type::IncludeUnknown, Resource::String::IncludeUnknownInListArgumentDescription, ArgumentType::Flag },
             Argument{ Execution::Args::Type::IncludePinned, Resource::String::IncludePinnedInListArgumentDescription, ArgumentType::Flag},
             Argument::ForType(Execution::Args::Type::ListDetails),
-            Argument{ Execution::Args::Type::Sort, Resource::String::SortArgumentDescription, ArgumentType::Standard }.SetCountLimit(6),
+            Argument{ Execution::Args::Type::Sort, Resource::String::SortArgumentDescription, ArgumentType::Standard }.SetCountLimit(AppInstaller::GetAllExponentialEnumValues<Settings::SortField>(Settings::SortField::None).size()),
```
JohnMcPMS: nit: There is probably a more efficient way that we could do this at compile time, like `consteval size_t GetExponentialEnumValuesCount<Enum>` that does the same algorithm just counting the number of items until we hit `Max`.

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
JohnMcPMS: I meant all of it.  I don't see a reason to pessimistically split out the context/argument extraction from the resolution portion.  If we expand to other commands, the arguments are likely to stay the same.  Maybe certain fields are not valid and can be checked by those commands, but this shared portion should be able to stay unchanged.

src/AppInstallerCLITests/PackageTableSortHelper.cpp
```diff
@@ -32,9 +32,12 @@ namespace
             REQUIRE(actual[i].FoldedName == expected[i].FoldedName);
             REQUIRE(actual[i].FoldedId == expected[i].FoldedId);
             REQUIRE(actual[i].FoldedSource == expected[i].FoldedSource);
-            REQUIRE(actual[i].HasAvailableVersion == expected[i].HasAvailableVersion);
+            REQUIRE(actual[i].ParsedAvailableVersion.has_value() == expected[i].ParsedAvailableVersion.has_value());
             REQUIRE(actual[i].ParsedInstalledVersion == expected[i].ParsedInstalledVersion);
-            REQUIRE(actual[i].ParsedAvailableVersion == expected[i].ParsedAvailableVersion);
+            if (actual[i].ParsedAvailableVersion.has_value() && expected[i].ParsedAvailableVersion.has_value())
```
JohnMcPMS: `optional` gives you all of these `ParsedAvailableVersion` checks at once through its equality operator.

src/AppInstallerCLICore/Commands/ListCommand.cpp
Madhusudhan-MSFT: `consteval` is a C++20 feature — the project targets C++17. Using `static constexpr` at the call site achieves the same compile-time evaluation.

JohnMcPMS (approved):

Madhusudhan-MSFT:
**Note:** The test failures observed in the build pipeline are confirmed to be unrelated to this change.

HarjitSi:
When will be able to download a release build with this?

JohnMcPMS:
> When will be able to download a release build with this?

Normal development build timeframe is 2-3 weeks after the merge.

PR #6193, opened 2026-05-01
Configuration processor auditing improvements

JohnMcPMS:
## Change
When a processor path is provided, inform the user about that fact in the output.  Ensure that the server agrees on the hash of the processor that is provided in its CLI arguments.

When the target processor is an app execution alias, the hash is of the contents of the reparse point data.  Otherwise, it is a hash of the contents of the file itself.

```
Custom processor path:
  Path: C:\Program Files\WindowsApps\Microsoft.DesiredStateConfiguration_3.2.0.0_x64__8wekyb3d8bbwe\dsc.exe
  Hash: 6ee88bd4c93c4a94539a0af0667ace8ffba48f5b8732930e1421721621ca19de
  Signed By: CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US
```

## Validation
Tests are added for the scenario and individual operations.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6193)


src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3279,6 +3252,34 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="ConfigurationProcessorPath" xml:space="preserve">
     <value>Specify the path to the configuration processor</value>
   </data>
+  <data name="ConfigurationProcessorPathAudit" xml:space="preserve">
+    <value>Custom DSC processor path in use:</value>
+    <comment>Displayed as a warning header when a custom --processor-path argument is provided.</comment>
```
florelis, line 3296: Does DSC need to be locked?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3279,6 +3252,34 @@ Please specify one of them using the --source option to proceed.</value>
   <data name="ConfigurationProcessorPath" xml:space="preserve">
     <value>Specify the path to the configuration processor</value>
   </data>
+  <data name="ConfigurationProcessorPathAudit" xml:space="preserve">
+    <value>Custom DSC processor path in use:</value>
+    <comment>Displayed as a warning header when a custom --processor-path argument is provided.</comment>
+  </data>
+  <data name="ConfigurationProcessorPathAuditHash" xml:space="preserve">
+    <value>  Hash: {0}</value>
```
florelis: Don't we usually add the indentation in code?

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
```diff
@@ -3279,6 +3252,34 @@ Please specify one of them using the --source option to proceed.</value>
...
+    <comment>{Locked="{0}"} {0} is the SHA256 hash hex string of the processor executable or its reparse data.</comment>
+  </data>
+  <data name="ConfigurationProcessorPathAuditIsAlias" xml:space="preserve">
+    <value>  Type: App execution alias</value>
+    <comment>Indicates the processor path is an app execution alias (MSIX package link).</comment>
+  </data>
+  <data name="ConfigurationProcessorPathAuditPath" xml:space="preserve">
+    <value>  Path: {0}</value>
+    <comment>{Locked="{0}"} {0} is the file system path to the custom DSC processor executable.</comment>
```
florelis: I don't think mentioning "DSC processor" adds any meaningful context for the translator, and it may even be noise. (e.g., "processor" could be interpreted as a CPU)

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
florelis, line 3296: We should make it clear that this means "the following is the path to something that processes DSCs, which is not the default processor, and is being used:".

"DSC processor" could be read as "processor that is DSC" (take DSC as an adjective)

"Custom DSC processor path" could be read as "path to a processor of custom DSCs"

"path in use" could be read as "the path is in use"

src/AppInstallerCLI.sln
```diff
@@ -242,6 +241,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "copilot", "copilot", "{B978
 		..\.github\copilot-instructions.md = ..\.github\copilot-instructions.md
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "doc", "doc", "{3FF6C881-2548-486E-8D70-7555A90030F5}"
+	ProjectSection(SolutionItems) = preProject
+		..\doc\windows\package-manager\winget\returnCodes.md = ..\doc\windows\package-manager\winget\returnCodes.md
+		..\doc\Settings.md = ..\doc\Settings.md
```
florelis, line 247: Should we move all the `.md`s here?

src/AppInstallerSharedLib/Certificates.cpp
```diff
@@ -736,4 +789,56 @@ namespace AppInstaller::Certificates
...
+            return {};
+        }
+
+        PCCERT_CONTEXT certContext = signer->pasCertChain[0].pCert;
+        if (!certContext)
+        {
+            return {};
+        }
+
+        return GetNameString(certContext, CERT_NAME_SIMPLE_DISPLAY_TYPE, false);
```
florelis: I think we should use the full subject, not just one part of it. We would probably want to be able to distinguish between something signed by `CN=John, C=US` and something signed by `CN=John, C=MX`

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
```diff
@@ -129,6 +131,90 @@ namespace AppInstaller::CLI::Workflow
...
+                    &bytesReturned,
+                    nullptr));
+
+                // Confirm it is specifically an app execution alias, not another reparse type.
+                THROW_HR_IF(E_INVALIDARG, bytesReturned < sizeof(DWORD));
+                DWORD reparseTag = *reinterpret_cast<DWORD*>(reparseBuffer.data());
+                THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_REPARSE_TAG_MISMATCH), reparseTag != IO_REPARSE_TAG_APPEXECLINK);
+
+                result.IsAlias = true;
+                result.HashString = Utility::SHA256::ConvertToString(
```
florelis, line 209: Could we extract the signer/publisher from the package that owns the alias?

florelis (dismissed):
I'm not familiar with the configuration code, so my review of that part was superficial and it may be wise to get somebody else to look at it.

src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp
JohnMcPMS: We don't know that with OS APIs.  We could infer things about the path and likely find the family name, then the full name, then extract the signing information.  But all of that is a guess, and a bad actor could easily replace the alias contents while we claimed signature information from the other package.

doc/windows/package-manager/winget/returnCodes.md
```diff
@@ -233,3 +233,4 @@ ms.localizationpriority: medium
 | 0x8A15C110 | -1978285814 | WINGET_CONFIG_ERROR_UNIT_SETTING_CONFIG_ROOT | A unit contains a setting that requires the config root. |
 | 0x8A15C111 | -1978285813 | WINGET_CONFIG_ERROR_UNIT_IMPORT_MODULE_ADMIN | Loading the module for the configuration unit failed because it requires administrator privileges to run. |
 | 0x8A15C112 | -1978285812 | WINGET_CONFIG_ERROR_NOT_SUPPORTED_BY_PROCESSOR | Operation is not supported by the configuration processor. |
+| 0x8A15C113 | -1978285811 | WINGET_CONFIG_ERROR_PROCESSOR_HASH_MISMATCH | The DSC processor hash provided does not match hash of the target file. |
```
yao-msft: Just realized the decimals for error codes starting from 0x8A15C110 were incorrect , we did not use ABCD..

src/AppInstallerCLICore/ConfigurationDynamicRuntimeFactory.cpp
```diff
@@ -355,6 +357,37 @@ namespace AppInstaller::CLI::ConfigurationRemoting
...
+                        winrt::hstring pathHash = m_dynamicFactory->Lookup(ToHString(PropertyName::FoundDscExecutablePathHash));
+                        if (!pathHash.empty())
+                        {
+                            json["processorPathHash"] = Utility::ConvertToUTF8(pathHash);
+                        }
+
+                        winrt::hstring pathIsAlias = m_dynamicFactory->Lookup(ToHString(PropertyName::FoundDscExecutablePathIsAlias));
+                        if (!pathIsAlias.empty())
+                        {
+                            json["processorPathIsAlias"] = (pathIsAlias == L"True");
```
yao-msft: Should the value be comparing to "true"? It seems inconsistent compared with line 388

src/Microsoft.Management.Configuration.UnitTests/Helpers/Errors.cs
```diff
@@ -46,6 +46,7 @@ internal static class Errors
         public static readonly int WINGET_CONFIG_ERROR_UNIT_SETTING_CONFIG_ROOT             = unchecked((int)0x8A15C110);
         public static readonly int WINGET_CONFIG_ERROR_UNIT_IMPORT_MODULE_ADMIN             = unchecked((int)0x8A15C111);
         public static readonly int WINGET_CONFIG_ERROR_NOT_SUPPORTED_BY_PROCESSOR           = unchecked((int)0x8A15C112);
+        public static readonly int WINGET_CONFIG_ERROR_PROCESSOR_HASH_MISMATCH              = unchecked((int)0x8A15C113);
         public static readonly int WINGET_CONFIG_ERROR_PARAMETER_INTEGRITY_BOUNDARY         = unchecked((int)0x8A15C013);
```
yao-msft: nit: This one should be moved to above section

yao-msft (dismissed):

doc/windows/package-manager/winget/returnCodes.md
JohnMcPMS: Fixed and added markdown generation via `winget error -o file.md` which fixed a few more things that were off.

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
yao-msft, line 2150: Should we update the command long description?

yao-msft (approved):

src/AppInstallerCLIPackage/Shared/Strings/en-us/winget.resw
JohnMcPMS: I don't think so.  It isn't the primary purpose of the command and I don't expect that many/any people will run it outside of us.  The one fun thing you can do is run it to get a localized markdown file.

PR #6194, opened 2026-05-01
Store provisioning mitigation

JohnMcPMS:
## Change
On older versions of Windows, when a package was already present but not provisioned, attempting to do a store install of an MSIX package for machine scope would end up as a no-op.  Newer versions handle this properly, but to ensure consistency this change checks the provisioning state after a Store install to an attempts to provision if not as expected.

## Validation
A test that forces the provisioning check is added.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6194)


src/AppInstallerCommonCore/MSStore.cpp
```diff
@@ -378,7 +435,24 @@ namespace AppInstaller::MSStore
...
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        if (TestHooks::s_ProvisionAfterInstall)
+        {
+            shouldProvision = *TestHooks::s_ProvisionAfterInstall;
+        }
+#endif
+
+        if (SUCCEEDED(hr) && shouldProvision)
+        {
+            // Mitigation: on older OS versions the Store install service may not provision the package
```
florelis: Should we check for the specific OS builds with this issue?
Could you add a comment of what versions have this issue so that we can remove it in the future once we don't support them anymore?

florelis (dismissed):

src/AppInstallerCommonCore/MSStore.cpp
JohnMcPMS: I don't think we need to actually check, as the mitigation is best effort.

Windows 11 (26100), and I will add that to the comments so that we can remove it in (checks current non-official support timeline) ~2035.

florelis (approved):

Issue #4231, opened 2024-03-05
--silent is not actually silent


kaifu-ck:
### Brief description of your issue

This has previously implicitly been reported in #1927 by @jedieaston and explicitly in #3891 by @shalinikhare27, but unwilling to fix the problem, the developers first declared that #1927 is a feature request (since redirecting the output from stdout to log would be a new feature) and then pretended that #3891 is a duplicate, giving themselves an excuse to ignore it.

So **this** is explicitly **not** a feature request and therefore neither directly a duplicate of #1927, nor transitively by way of #3891.

I am reporting that the existing feature --silent is broken, under the expectation that the faulty implementation will be corrected to be in line with both its implicit use and user expectation.

The fact that it's broken is actually acknowledged in the documentation, which says

> Runs the installer in silent mode. This suppresses all UI. The default experience shows installer progress.

A CLI is a UI. Not a GUI, usually not a TUI, but very much a UI - always and indisputably.  
"This suppresses all UI" and "The default experience shows installer progress" could not both be true if this feature worked correctly.
So clearly somebody already realized that --silent isn't actually silent, and just pretended that that was a reasonable way for --silent to work.

The fact that there is no hint how to change "the default experience" is a related documentation flaw, but independent of the fact that --silent clearly and documentedly is not silent.

So. --silent is supposed to be silent by name and documented to suppress all UI.
It's not and it doesn't.

That is a flaw in the existing implementation that I am reporting, not a feature request for new functionality.

### Steps to reproduce

Use --silent on any operation, look at your screen.

### Expected behavior

When looking at your screen, you shouldn't see any output.
The application is silent.

### Actual behavior

When looking at your screen, you will see basically all output.
The application is not silent.

### Environment

```shell
Doesn't matter - universal issue for months, documented in the official documentation.
```

methanoid:
This is a bug... it does NOT do what it should...  I tried adding every switch I could find to silence the darn thing but couldn't...  I'm confused with the number of times this IS reported but flagged as a duplicate or ignored. So not sure WHICH one I need to +1 on... but it IS annoying as it IS basic functionality

kaifu-ck:
Having been notified due to the previous comment, I just realized that I probably misunderstood what the phrasing "The default experience shows installer progress." was meant to convey.  
Presumably, "the default experience" does not refer to the default experience _of_ `--silent`, but to the default experience _without_ `--silent`. The default _winget_ experience.

[...]

PR #6201, opened 2026-05-03
Specification: UserMessages manifest field

### LINKED ISSUE #3483: Make InstallationNotes flow-dependent

denelon:
## 📖 Description
Adds the full specification for the `UserMessages` manifest field, covering how messages are shown during install, upgrade, uninstall, and download flows. The spec includes the flow behavior matrix, CLI arguments, settings, PowerShell cmdlet integration, and an update to `doc/specs/spec-template.md`.

Created with GitHub Copilot assistance.

## 🔗 References
Resolves #3483

## 🔍 Validation
- Specification content reviewed by Trenly.

## ✅ Checklist
- [x] Linked to an issue

## 📋 Issue Type
- [ ] Bug fix
- [x] Feature
- [ ] Task

### DIFF
--- a/doc/specs/#3483 - UserMessages.md
+++ b/doc/specs/#3483 - UserMessages.md
@@ -1,0 +1,562 @@
+---
+author: Demitrius Nelon denelon, GitHub Copilot <Copilot>
+created on: 2026-05-03
+last updated: 2026-05-03
+issue id: 3483
+---
+
+# UserMessages
+
+For [#3483](https://github.com/microsoft/winget-cli/issues/3483)
+
+## Abstract
+
+This specification proposes adding an optional `UserMessages` object to the WinGet package manifest schema. The new field provides structured, flow-dependent messaging to users before and after install, upgrade, and uninstall operations. It is designed as a non-breaking, additive change that coexists with the existing `InstallationNotes` string field during a transition period and eventually replaces it.
+
+## Inspiration
+
+The existing `InstallationNotes` field is a single string displayed after a successful install. Community feedback in [#3483](https://github.com/microsoft/winget-cli/issues/3483) identified several limitations:
+
+1. **No flow specificity.** The same message appears regardless of whether the user is installing, upgrading, or uninstalling. Upgrade and uninstall operations have no mechanism for user-facing notes at all.
+2. **No pre-action messaging.** Some packages need to inform users of prerequisites, data-loss risks, or configuration steps before an operation begins. Today there is no way to surface this information through the manifest.
+3. **No structured extensibility.** The single string cannot be extended to support new scenarios without breaking the schema contract.
+
+Changing `InstallationNotes` from a string to an object (as originally proposed in #3483) would break existing manifests and clients. This specification introduces a new field alongside the existing one, providing a clean migration path.
+
+## Solution Design
+
+### Schema Changes
+
+A new optional `UserMessages` object is added to the **defaultLocale** and **locale** manifest types. The field is localizable — locale-specific manifests may override any or all messages defined in the defaultLocale manifest, following the existing locale fallback behavior.
+
+```json
+"UserMessages": {
+    "type": ["object", "null"],
+    "description": "Optional messages displayed to the user before or after specific package operations.",
+    "properties": {
+        "PreInstall": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 2048,
+            "description": "Confirmation request displayed to the user before a package installation begins."
+        },
+        "PostInstall": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 10000,
+            "description": "Message displayed to the user upon completion of a package installation. This is the designated successor to InstallationNotes."
+        },
+        "PreUpgrade": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 2048,
+            "description": "Confirmation request displayed to the user before a package upgrade begins."
+        },
+        "PostUpgrade": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 10000,
+            "description": "Message displayed to the user upon completion of a package upgrade."
+        },
+        "PreUninstall": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 2048,
+            "description": "Confirmation request displayed to the user before a package uninstall begins."
+        },
+        "PostUninstall": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 10000,
+            "description": "Message displayed to the user upon completion of a package uninstall."
+        }
+    },
+    "additionalProperties": false
+}
+```
+
+> [!NOTE]
+> **Field limits:** Pre-action messages use a lower `maxLength` of 2048 characters because they gate the operation with a user prompt in interactive mode. Shorter messages improve readability and reduce the risk of users dismissing important information. Post-action messages retain the 10000-character limit consistent with the existing `InstallationNotes` field.
+
+### Schema Version
+
+This change requires a new manifest schema version (the next minor version after the current release, e.g., 1.13.0). The `UserMessages` field is added to:
+
+- `defaultLocale.schema.json`
+- `locale.schema.json`
+
+The `installer.schema.json` is not affected. User-facing messages are locale-sensitive content and belong in the locale manifests.
+
+### Existing `InstallationNotes` Field
+
+The existing `InstallationNotes` field is **unchanged** in this schema version. It continues to function as it does today. During the transition period, manifests may include both fields for backward compatibility.
+
+### Client Behavior
+
+#### Precedence Rules
+
+When both `InstallationNotes` and `UserMessages.PostInstall` are present in a manifest:
+
+- **Clients that support `UserMessages`:** Display `UserMessages.PostInstall` and ignore `InstallationNotes`.
+- **Clients that do not support `UserMessages`:** Display `InstallationNotes` (current behavior, unaffected by the unknown field).
+
+#### Pre-Action Messages (PreInstall, PreUpgrade, PreUninstall)
+
+Pre-action messages are displayed **before** the operation begins and represent information the user should be aware of before proceeding.
+
+**Interactive mode (default):**
+
+```
+The following information has been provided by the manifest author:
+
+  This package requires 2 GB of disk space and will modify your PATH.
+
+Do you wish to continue? [Y/n]:
+```
+
+The user is prompted to confirm. Entering `N` cancels the operation. Entering `Y` or pressing Enter proceeds.
+
+**Non-interactive mode (`--disable-interactivity`):**
+
+The pre-action message is displayed (rendered to the output stream) but no prompt is shown and the operation proceeds automatically. This ensures automation pipelines see the message in logs without being blocked.
+
+```
+The following information has been provided by the manifest author:
+
+  This package requires 2 GB of disk space and will modify your PATH.
+
+Proceeding automatically (non-interactive mode)...
+```
+
+> [!NOTE]
+> **Design rationale:** Pre-action messages are informational, not contractual agreements. Unlike license terms (which require explicit acceptance via `--accept-package-agreements`), user messages are manifest-author-provided guidance. The `--disable-interactivity` flag is the appropriate suppressor because it already governs all interactive prompts in WinGet. A new flag is not warranted.
+
+#### Post-Action Messages (PostInstall, PostUpgrade, PostUninstall)
+
+Post-action messages are displayed **after** the operation completes successfully. They are informational only — no prompt is shown.
+
+```
+Successfully installed ExampleApp [Publisher.ExampleApp] v2.0.0
+
+  Restart your terminal to use the CLI.
+```
+
+Post-action messages are displayed in both interactive and non-interactive modes. They are suppressed when the `userMessages.disablePostActionMessages` setting is enabled.
+
+#### Flow Behavior Matrix
+
+| Scenario | Pre-message | Post-message | Notes |
+|----------|-------------|--------------|-------|
+| `winget install` | `PreInstall` shown, prompt Y/n | `PostInstall` shown | Standard single-package install |
+| `winget upgrade <package>` | `PreUpgrade` shown, prompt Y/n | `PostUpgrade` shown | Single-package upgrade |
+| `winget upgrade --all` | Each package's `PreUpgrade` shown, prompt Y/n per package | Each package's `PostUpgrade` shown | See "Bulk Operations" below |
+| `winget uninstall` | `PreUninstall` shown, prompt Y/n | `PostUninstall` shown | Standard uninstall |
+| `winget install --disable-interactivity` | `PreInstall` displayed, no prompt | `PostInstall` shown | Non-interactive install |
+| `winget install --silent` | `PreInstall` displayed, no prompt | `PostInstall` shown | Silent implies non-interactive |
+| Dependency packages | Not shown | Not shown | Messages apply to the root package only, not transitive dependencies |
+| `winget import` | Each package's `Pre*` shown, prompt Y/n per package | Each package's `Post*` shown | Follows the same per-package model |
+| WinGet Configuration (DSC) | Displayed in log output, no prompt | Displayed in log output | Configuration is inherently non-interactive |
+| `winget show` | N/A | N/A | `show` displays `UserMessages` as package metadata (see below) |
+| `winget repair` | `PreInstall` shown, prompt Y/n | `PostInstall` shown | Repair reuses install messages |
+| `winget export` | N/A | N/A | Export captures package list, not messages |
+| Operation failure | `Pre*` shown before attempt | `Post*` NOT shown | Post-messages only appear on success |
+| `disableInstallNotes` setting enabled | `Pre*` still shown (not affected) | `Post*` still shown (controlled by `userMessages` settings, not `disableInstallNotes`) | Legacy setting does not affect new fields |
+| `userMessages.disablePostActionMessages` enabled | `Pre*` still shown | `Post*` suppressed | New setting for new fields |
+| `userMessages.disablePreActionMessages` enabled | `Pre*` suppressed, no prompt | `Post*` still shown | Suppresses both display and prompt |
+| `--ignore-user-messages` | `Pre*` suppressed, no prompt | `Post*` suppressed | CLI argument overrides settings for this invocation |
+| `--ignore-pre-action-messages` | `Pre*` suppressed, no prompt | `Post*` still shown | CLI argument overrides `disablePreActionMessages` setting |
+| `--ignore-post-action-messages` | `Pre*` still shown | `Post*` suppressed | CLI argument overrides `disablePostActionMessages` setting |
+| `winget validate` | Not shown, not prompted | Not shown | Validation checks schema correctness only; messages are not rendered |
+
+#### Bulk Operations (`upgrade --all`, `import`)
+
+For bulk operations, pre-action messages are shown per-package. In interactive mode, each package with a pre-action message prompts individually. This is consistent with how `--accept-package-agreements` works today — per-package, not a single blanket prompt.
+
+A future enhancement could aggregate pre-messages and present them as a summary before beginning the batch, but that is out of scope for this specification.
+
+### Settings
+
+A new `userMessages` settings object is introduced under the root of the WinGet settings file. The existing `installBehavior.disableInstallNotes` setting is **not modified** and continues to control only the legacy `InstallationNotes` field.
+
+```json
+{
+    "userMessages": {
+        "disablePreActionMessages": false,
+        "disablePostActionMessages": false
+    }
+}
+```
+
+| Setting | Default | Description |
+|---------|---------|-------------|
+| `userMessages.disablePreActionMessages` | `false` | When `true`, suppresses all pre-action messages (`PreInstall`, `PreUpgrade`, `PreUninstall`) and their associated prompts. The operation proceeds as if the user confirmed. |
+| `userMessages.disablePostActionMessages` | `false` | When `true`, suppresses all post-action messages (`PostInstall`, `PostUpgrade`, `PostUninstall`). |
+
+**Interaction with existing settings:**
+
+The following table shows what is displayed on **clients that support `UserMessages`** during an install operation when both `InstallationNotes` and `UserMessages.PostInstall` are present. Per the precedence rules above, supporting clients display `UserMessages.PostInstall` and ignore `InstallationNotes`.
+
+| `disableInstallNotes` | `userMessages.disablePostActionMessages` | `InstallationNotes` | `UserMessages.Post*` |
+|----------------------|------------------------------------------|---------------------|----------------------|
+| `false` | `false` | Ignored (superseded) | Shown |
+| `true` | `false` | Ignored (superseded) | Shown |
+| `false` | `true` | Ignored (superseded) | Suppressed |
+| `true` | `true` | Ignored (superseded) | Suppressed |
+
+On **older clients** that do not support `UserMessages`, only `InstallationNotes` is relevant and `disableInstallNotes` controls it as it does today.
+
+### CLI Arguments
+
+Corresponding CLI arguments are provided so users can override the settings on a per-invocation basis.
+
+| Argument | Applies to | Description |
+|----------|-----------|-------------|
+| `--ignore-user-messages` | `install`, `upgrade`, `uninstall`, `repair`, `import` | Suppresses all `UserMessages` (both pre- and post-action) for this invocation. Pre-action prompts are skipped and the operation proceeds. |
+| `--ignore-pre-action-messages` | `install`, `upgrade`, `uninstall`, `repair`, `import` | Suppresses pre-action messages and their prompts for this invocation. Post-action messages are still shown. |
+| `--ignore-post-action-messages` | `install`, `upgrade`, `uninstall`, `repair`, `import` | Suppresses post-action messages for this invocation. Pre-action messages and prompts are still shown. |
+
+**Precedence:** CLI arguments override settings. If a user has `userMessages.disablePreActionMessages` set to `false` in settings but passes `--ignore-pre-action-messages`, the pre-action messages are suppressed for that invocation.
+
+**Relationship to `--disable-interactivity`:** The `--disable-interactivity` flag suppresses the **prompt** on pre-action messages but still displays the message text. The `--ignore-pre-action-messages` flag suppresses both the message and the prompt entirely.
+
+Example usage:
+
+```
+winget upgrade --all --ignore-pre-action-messages
+winget install Publisher.ExampleApp --ignore-user-messages
+```
+
+### COM API
+
+The COM API must surface `UserMessages` in a structured manner for programmatic consumers.
+
+A new `IPackageUserMessages` interface is introduced, and a new versioned interface (e.g., `IPackageCatalogInfo2` or `IPackageVersionInfo2`) exposes a `UserMessages` property that returns this object. This follows the established WinGet pattern for additive COM API changes — new interface versions are introduced rather than modifying existing interfaces, ensuring existing consumers remain compatible.
+
+```idl
+interface IPackageUserMessages : IInspectable
+{
+    String PreInstall { get; };
+    String PostInstall { get; };
+    String PreUpgrade { get; };
+    String PostUpgrade { get; };
+    String PreUninstall { get; };
+    String PostUninstall { get; };
+}
+```
+
+Consumers query for the new interface version to access `UserMessages`. Existing consumers that use the original interface are unaffected.
+
+COM consumers are responsible for their own interactivity model. The WinGet COM API does not prompt — it returns the data and the consumer decides how to handle it.
+
+### PowerShell Cmdlets
+
+The WinGet PowerShell module cmdlets consume the COM API and must surface `UserMessages` appropriately.
+
+#### Affected Cmdlets
+
+| Cmdlet | Pre-message field | Post-message field |
+|--------|-------------------|-------------------|
+| `Install-WinGetPackage` | `PreInstall` | `PostInstall` |
+| `Update-WinGetPackage` | `PreUpgrade` | `PostUpgrade` |
+| `Uninstall-WinGetPackage` | `PreUninstall` | `PostUninstall` |
+| `Repair-WinGetPackage` | `PreInstall` (reuses install messages) | `PostInstall` (reuses install messages) |
+
+#### Interactive Behavior
+
+When running interactively in a PowerShell session, cmdlets follow the same pattern as the CLI:
+
+```powershell
+PS> Install-WinGetPackage -Id Publisher.ExampleApp
+
+The following information has been provided by the manifest author:
+
+  This package requires 2 GB of disk space and will modify your PATH.
+
+Do you wish to continue?
+[Y] Yes  [N] No  (default is "Y"):
+```
+
+The prompt uses `$Host.UI.PromptForChoice` (or equivalent) to integrate with PowerShell's native prompting, supporting `-Confirm` and `-WhatIf` patterns where applicable.
+
+#### Non-Interactive / Pipeline Behavior
+
+When running non-interactively (e.g., in a script with no host, or when the user passes `-Force`), pre-action messages are written to the verbose stream (`Write-Verbose`) and the operation proceeds without prompting:
+
+```powershell
+PS> Install-WinGetPackage -Id Publisher.ExampleApp -Force -Verbose
+VERBOSE: Manifest author message (PreInstall): This package requires 2 GB of disk space and will modify your PATH.
+```
+
+Post-action messages are written to the information stream (`Write-Information`) so they appear by default but can be suppressed with `-InformationAction SilentlyContinue`.
+
+#### Output Object
+
+The result objects returned by cmdlets should include UserMessages in their output when present:
+
+```powershell
+PS> $result = Install-WinGetPackage -Id Publisher.ExampleApp
+PS> $result.UserMessages
+
+PreInstall  : This package requires 2 GB of disk space and will modify your PATH.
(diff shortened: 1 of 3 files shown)


doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,371 @@
...
+```json
+"UserMessages": {
+    "type": ["object", "null"],
+    "description": "Optional messages displayed to the user before or after specific package operations.",
+    "properties": {
+        "PreInstall": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 2048,
+            "description": "Message displayed to the user before a package installation begins."
```
Trenly: `Confirmation request displayed` would make it more clear that setting this will invoke a user prompt - same for below

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,464 @@
...
+
+Post-action messages are displayed **after** the operation completes successfully. They are informational only — no prompt is shown.
+
+```
+Successfully installed ExampleApp [Publisher.ExampleApp] v2.0.0
+
+  Restart your terminal to use the CLI.
+```
+
+Post-action messages are displayed in both interactive and non-interactive modes. They are suppressed when the `disableInstallNotes` setting is enabled (extended to cover all post-action messages, not just `InstallationNotes`).
```
Trenly: This is inconsistent with the settings section below

.github/instructions/specs.instructions.md
```diff
@@ -0,0 +1,117 @@
+# Writing WinGet Specifications
+
+These instructions apply when creating or editing files in `doc/specs/`.
```
Trenly: Add the `applyTo` spec at the very top line:

```
---
applyTo: "doc/specs/*"
---
```

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+
+## Abstract
+
+This specification proposes adding an optional `UserMessages` object to the WinGet package manifest schema. The new field provides structured, flow-dependent messaging to users before and after install, upgrade, and uninstall operations. It is designed as a non-breaking, additive change that coexists with the existing `InstallationNotes` string field during a transition period and eventually replaces it.
+
+## Inspiration
+
+The existing `InstallationNotes` field is a single string displayed after a successful install. Community feedback in [#3483](https://github.com/microsoft/winget-cli/issues/3483) identified several limitations:
+
+1. **No flow specificity.** The same message appears regardless of whether the user is installing, upgrading, or uninstalling. Upgrade and uninstall operations have no mechanism for user-facing notes at all.
```
JohnMcPMS, line 20: Does the same message appear when upgrading/uninstalling, or does no message appear?

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+            "maxLength": 2048,
+            "description": "Confirmation request displayed to the user before a package upgrade begins."
+        },
+        "PostUpgrade": {
+            "type": ["string", "null"],
+            "minLength": 1,
+            "maxLength": 10000,
+            "description": "Message displayed to the user upon completion of a package upgrade."
+        },
+        "PreUninstall": {
```
JohnMcPMS, line 61: Putting anything in the manifest and making it an expectation that this will affect uninstall is complicating to the current architecture.  What version of this message affects uninstall?  Are we expected to retrieve this messaging even if the package was not installed through winget?  Hopefully this spec answers those questions later, but the answer may make the overall cost of implementing higher to set up the infrastructure.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+    "additionalProperties": false
+}
+```
+
+> [!NOTE]
+> **Field limits:** Pre-action messages use a lower `maxLength` of 2048 characters because they gate the operation with a user prompt in interactive mode. Shorter messages improve readability and reduce the risk of users dismissing important information. Post-action messages retain the 10000-character limit consistent with the existing `InstallationNotes` field.
+
+### Schema Version
+
+This change requires a new manifest schema version (the next minor version after the current release, e.g., 1.13.0). The `UserMessages` field is added to:
```
JohnMcPMS: I would assume we actually want to jump the schema versions to match the client versions.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+
+- `defaultLocale.schema.json`
+- `locale.schema.json`
+
+The `installer.schema.json` is not affected. User-facing messages are locale-sensitive content and belong in the locale manifests.
+
+### Existing `InstallationNotes` Field
+
+The existing `InstallationNotes` field is **unchanged** in this schema version. It continues to function as it does today. During the transition period, manifests may include both fields for backward compatibility.
```
JohnMcPMS:
```suggestion
The existing `InstallationNotes` field is **unchanged** in this schema version. It continues to function as it does today. During the transition period, manifests should include both fields for backward compatibility.
```
Otherwise you get no notes in older clients.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+
+```
+The following information has been provided by the manifest author:
+
+  This package requires 2 GB of disk space and will modify your PATH.
+
+Do you wish to continue? [Y/n]:
+```
+
+The user is prompted to confirm. Entering `N` cancels the operation. Entering `Y` or pressing Enter proceeds.
```
JohnMcPMS, line 140: This will for sure break some people's automation, but they of course have the tools today to ensure that it wouldn't (disabling interactivity or just using PowerShell).

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+Post-action messages are displayed in both interactive and non-interactive modes. They are suppressed when the `userMessages.disablePostActionMessages` setting is enabled.
+
+#### Flow Behavior Matrix
+
+| Scenario | Pre-message | Post-message | Notes |
+|----------|-------------|--------------|-------|
+| `winget install` | `PreInstall` shown, prompt Y/n | `PostInstall` shown | Standard single-package install |
+| `winget upgrade <package>` | `PreUpgrade` shown, prompt Y/n | `PostUpgrade` shown | Single-package upgrade |
+| `winget upgrade --all` | Each package's `PreUpgrade` shown, prompt Y/n per package | Each package's `PostUpgrade` shown | See "Bulk Operations" below |
+| `winget uninstall` | `PreUninstall` shown, prompt Y/n | `PostUninstall` shown | Standard uninstall |
```
JohnMcPMS: But which messages?  For install and upgrade, we have to have the manifest in question.  Uninstall could be the message at the time of install, the message at the time of uninstall (from the target version or the closest version or the latest version), or no message if it didn't come from winget even if it is correlated.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+| `winget install --disable-interactivity` | `PreInstall` displayed, no prompt | `PostInstall` shown | Non-interactive install |
+| `winget install --silent` | `PreInstall` displayed, no prompt | `PostInstall` shown | Silent implies non-interactive |
```
JohnMcPMS: Our default is non-interactive *installers*, silent is about the visuals.  I don't know that we should be muddying the use here.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+| `winget import` | Each package's `Pre*` shown, prompt Y/n per package | Each package's `Post*` shown | Follows the same per-package model |
+| WinGet Configuration (DSC) | Displayed in log output, no prompt | Displayed in log output | Configuration is inherently non-interactive |
+| `winget show` | N/A | N/A | `show` displays `UserMessages` as package metadata (see below) |
+| `winget repair` | `PreInstall` shown, prompt Y/n | `PostInstall` shown | Repair reuses install messages |
+| `winget export` | N/A | N/A | Export captures package list, not messages |
+| Operation failure | `Pre*` shown before attempt | `Post*` NOT shown | Post-messages only appear on success |
+| `disableInstallNotes` setting enabled | `Pre*` still shown (not affected) | `Post*` still shown (controlled by `userMessages` settings, not `disableInstallNotes`) | Legacy setting does not affect new fields |
+| `userMessages.disablePostActionMessages` enabled | `Pre*` still shown | `Post*` suppressed | New setting for new fields |
+| `userMessages.disablePreActionMessages` enabled | `Pre*` suppressed, no prompt | `Post*` still shown | Suppresses both display and prompt |
```
JohnMcPMS: Should we also have a `disablePreActionPrompts` that shows the message, but disables the prompt?

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+| `--ignore-user-messages` | `Pre*` suppressed, no prompt | `Post*` suppressed | CLI argument overrides settings for this invocation |
+| `--ignore-pre-action-messages` | `Pre*` suppressed, no prompt | `Post*` still shown | CLI argument overrides `disablePreActionMessages` setting |
+| `--ignore-post-action-messages` | `Pre*` still shown | `Post*` suppressed | CLI argument overrides `disablePostActionMessages` setting |
+| `winget validate` | Not shown, not prompted | Not shown | Validation checks schema correctness only; messages are not rendered |
+
+#### Bulk Operations (`upgrade --all`, `import`)
+
+For bulk operations, pre-action messages are shown per-package. In interactive mode, each package with a pre-action message prompts individually. This is consistent with how `--accept-package-agreements` works today — per-package, not a single blanket prompt.
+
+A future enhancement could aggregate pre-messages and present them as a summary before beginning the batch, but that is out of scope for this specification.
```
JohnMcPMS: We already handle all of the preconditions before beginning the batch and I would expect that this would be handled similarly.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+
+### Settings
+
+A new `userMessages` settings object is introduced under the root of the WinGet settings file. The existing `installBehavior.disableInstallNotes` setting is **not modified** and continues to control only the legacy `InstallationNotes` field.
+
+```json
+{
+    "userMessages": {
```
JohnMcPMS: This doesn't seem like it should be top-level, but maybe that is because we haven't had a strong design around the settings schema and so it doesn't fit anywhere else.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+```
+
+| Setting | Default | Description |
+|---------|---------|-------------|
+| `userMessages.disablePreActionMessages` | `false` | When `true`, suppresses all pre-action messages (`PreInstall`, `PreUpgrade`, `PreUninstall`) and their associated prompts. The operation proceeds as if the user confirmed. |
+| `userMessages.disablePostActionMessages` | `false` | When `true`, suppresses all post-action messages (`PostInstall`, `PostUpgrade`, `PostUninstall`). |
+
+**Interaction with existing settings:**
+
+The following table shows what is displayed on **clients that support `UserMessages`** during an install operation when both `InstallationNotes` and `UserMessages.PostInstall` are present. Per the precedence rules above, supporting clients display `UserMessages.PostInstall` and ignore `InstallationNotes`.
```
JohnMcPMS: This table is confusing.  I would allow the section about the manifest proper precedence to cover that, and then given that we have some already decided on post message to show, how do the settings affect it.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+
+Corresponding CLI arguments are provided so users can override the settings on a per-invocation basis.
+
+| Argument | Applies to | Description |
+|----------|-----------|-------------|
+| `--ignore-user-messages` | `install`, `upgrade`, `uninstall`, `repair`, `import` | Suppresses all `UserMessages` (both pre- and post-action) for this invocation. Pre-action prompts are skipped and the operation proceeds. |
+| `--ignore-pre-action-messages` | `install`, `upgrade`, `uninstall`, `repair`, `import` | Suppresses pre-action messages and their prompts for this invocation. Post-action messages are still shown. |
+| `--ignore-post-action-messages` | `install`, `upgrade`, `uninstall`, `repair`, `import` | Suppresses post-action messages for this invocation. Pre-action messages and prompts are still shown. |
+
+**Precedence:** CLI arguments override settings. If a user has `userMessages.disablePreActionMessages` set to `false` in settings but passes `--ignore-pre-action-messages`, the pre-action messages are suppressed for that invocation.
```
JohnMcPMS: There is no ability to have the setting set to disable messages but then request messages for a specific invocation.  I don't know that this is a problem, just calling it out.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+```
+winget upgrade --all --ignore-pre-action-messages
+winget install Publisher.ExampleApp --ignore-user-messages
+```
+
+### COM API
+
+The COM API must surface `UserMessages` in a structured manner for programmatic consumers.
+
+A new `IPackageUserMessages` interface is introduced, and a new versioned interface (e.g., `IPackageCatalogInfo2` or `IPackageVersionInfo2`) exposes a `UserMessages` property that returns this object. This follows the established WinGet pattern for additive COM API changes — new interface versions are introduced rather than modifying existing interfaces, ensuring existing consumers remain compatible.
```
JohnMcPMS: This should either be vague about the names or extremely specific and accurate about both names and the objects that implement them.

I think it is reasonable to simply state, "The data will be available in the COM API as six strings" without prescribing exactly how.  It is also reasonable to prescribe the exact shape and how it is linked.  But this half describes something that is definitely not how even this part will look.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+}
+```
+
+Consumers query for the new interface version to access `UserMessages`. Existing consumers that use the original interface are unaffected.
+
+COM consumers are responsible for their own interactivity model. The WinGet COM API does not prompt — it returns the data and the consumer decides how to handle it.
+
+### PowerShell Cmdlets
+
+The WinGet PowerShell module cmdlets consume the COM API and must surface `UserMessages` appropriately.
```
JohnMcPMS, line 267: It feels like we might want to consider exposing the settings via COM so that well behaved callers (like our own PS module) could leverage them to make decisions about the use of these optional data points.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+| `winget upgrade --all` | Each package's `PreUpgrade` shown, prompt Y/n per package | Each package's `PostUpgrade` shown | See "Bulk Operations" below |
+| `winget uninstall` | `PreUninstall` shown, prompt Y/n | `PostUninstall` shown | Standard uninstall |
+| `winget install --disable-interactivity` | `PreInstall` displayed, no prompt | `PostInstall` shown | Non-interactive install |
+| `winget install --silent` | `PreInstall` displayed, no prompt | `PostInstall` shown | Silent implies non-interactive |
+| Dependency packages | Not shown | Not shown | Messages apply to the root package only, not transitive dependencies |
+| `winget import` | Each package's `Pre*` shown, prompt Y/n per package | Each package's `Post*` shown | Follows the same per-package model |
+| WinGet Configuration (DSC) | Displayed in log output, no prompt | Displayed in log output | Configuration is inherently non-interactive |
+| `winget show` | N/A | N/A | `show` displays `UserMessages` as package metadata (see below) |
+| `winget repair` | `PreInstall` shown, prompt Y/n | `PostInstall` shown | Repair reuses install messages |
```
JohnMcPMS: Is that appropriate?  I suppose it depends on the actual repair action to some degree.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+```powershell
+$result = Install-WinGetPackage -Id Publisher.ExampleApp -Force
+if ($result.UserMessages.PostInstall) {
+    Write-Host "Note from manifest author: $($result.UserMessages.PostInstall)" -ForegroundColor Yellow
+}
+```
+
+#### Find-WinGetPackage / Show-WinGetPackage
+
+The `Show-WinGetPackage` (or `Get-WinGetPackage`) cmdlet should include `UserMessages` in its output when displaying package metadata, allowing users to review messages before deciding to install:
```
JohnMcPMS: 🤖 hallucinating?  I know that a `Show` type command has been requested, but we don't have it.  And `Get` doesn't make sense as that is for already installed things.

doc/specs/#3483 - UserMessages.md
Trenly: My preference would be target version if it is correlated, latest version if no target version exists, and no message if not correlated at all, but leaving to Demitrius for his decision

doc/specs/#3483 - UserMessages.md
Trenly: Reference - there is already complaints that silent displays output from the CLI
* #4231

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+The existing `InstallationNotes` field is **unchanged** in this schema version. It continues to function as it does today. During the transition period, manifests may include both fields for backward compatibility.
+
+### Client Behavior
+
+#### Precedence Rules
+
+When both `InstallationNotes` and `UserMessages.PostInstall` are present in a manifest:
+
+- **Clients that support `UserMessages`:** Display `UserMessages.PostInstall` and ignore `InstallationNotes`.
+- **Clients that do not support `UserMessages`:** Display `InstallationNotes` (current behavior, unaffected by the unknown field).
```
florelis, line 113: Should we require that they be the same text in winget-pkgs?
I can't think of an example, but it seems like having users get different messages may cause issues.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+| `winget install --silent` | `PreInstall` displayed, no prompt | `PostInstall` shown | Silent implies non-interactive |
+| Dependency packages | Not shown | Not shown | Messages apply to the root package only, not transitive dependencies |
+| `winget import` | Each package's `Pre*` shown, prompt Y/n per package | Each package's `Post*` shown | Follows the same per-package model |
+| WinGet Configuration (DSC) | Displayed in log output, no prompt | Displayed in log output | Configuration is inherently non-interactive |
+| `winget show` | N/A | N/A | `show` displays `UserMessages` as package metadata (see below) |
+| `winget repair` | `PreInstall` shown, prompt Y/n | `PostInstall` shown | Repair reuses install messages |
+| `winget export` | N/A | N/A | Export captures package list, not messages |
+| Operation failure | `Pre*` shown before attempt | `Post*` NOT shown | Post-messages only appear on success |
+| `disableInstallNotes` setting enabled | `Pre*` still shown (not affected) | `Post*` still shown (controlled by `userMessages` settings, not `disableInstallNotes`) | Legacy setting does not affect new fields |
```
florelis: I would expect this to at least disable post install messages. If I set this at some point, I wouldn't want the messages to start showing up again just because the developers decided to change the name of the field.

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+  Restart your terminal to use the CLI.
+```
+
+Post-action messages are displayed in both interactive and non-interactive modes. They are suppressed when the `userMessages.disablePostActionMessages` setting is enabled.
+
+#### Flow Behavior Matrix
+
+| Scenario | Pre-message | Post-message | Notes |
+|----------|-------------|--------------|-------|
+| `winget install` | `PreInstall` shown, prompt Y/n | `PostInstall` shown | Standard single-package install |
```
florelis: Nit: Could you standardize the language for the message columns? It's a bit confusing having to reason about it while considering that `shown == displayed == still shown` and `suppressed == not shown`

doc/specs/#3483 - UserMessages.md
```diff
@@ -0,0 +1,562 @@
...
+
+The `winget validate` command and the automated validation pipeline in the winget-pkgs repository must handle `UserMessages` correctly:
+
+1. **Schema validation:** Validate `UserMessages` against the JSON schema (field types, lengths, required/optional).
+2. **Non-interactive execution:** The validation pipeline runs with `--disable-interactivity`. Pre-messages must not block validation.
+3. **Content validation:** The pipeline should validate that:
+   - Messages do not contain ANSI escape sequences or control characters.
+   - Messages do not contain URLs that appear to be phishing attempts (leveraging existing content moderation if applicable).
+4. **Deprecation warnings:** During the transition period, if a manifest contains `InstallationNotes` but no `UserMessages.PostInstall`, a warning (not an error) may be emitted suggesting the manifest author add the `UserMessages` equivalent.
+5. **Rendering validation:** Validate that pre-messages render cleanly when displayed (no truncation at the 2048-character limit that would break mid-word or mid-sentence).
```
florelis: Wouldn't this fall under schema validation?

doc/specs/#3483 - UserMessages.md
florelis: That feels like a different conversation. The current behavior is that `--silent` only affects the installer and not winget, so making it affect this would be inconsistent and maybe unexpected.

If we want to make `--silent` silence the output from winget and not just the installer, it would for sure affect the user messages, but it would also affect many other things outside the scope of this spec that may be better discussed separately.

doc/specs/#3483 - UserMessages.md
```diff
@@ -100,7 +112,18 @@ When both `InstallationNotes` and `UserMessages.PostInstall` are present in a ma
...
-#### Pre-Action Messages (PreInstall, PreUpgrade, PreUninstall)
+> [!NOTE]
+> Each operation displays only its corresponding `UserMessages` field. There is no cross-field fallback between operations — for example, if a manifest defines `PreInstall` but not `PreUpgrade`, no pre-action message is displayed during an upgrade.
+
+#### Uninstall Message Sourcing
+
+For uninstall operations, `PreUninstall` and `PostUninstall` are sourced as follows:
+
+- If the package is correlated to a specific installed version, use that version's manifest messages.
+- If the package is correlated but no target version manifest exists, use the latest available version's messages.
```
florelis, line 123: Could we try to find what the "closest" version is instead of using the latest?
For example, if the user has v1.2 and we have manifests for v1.1 and v2.0, the notes from v1.1 are probably more relevant.

doc/specs/#3483 - UserMessages.md
```diff
@@ -141,67 +167,67 @@ Successfully installed ExampleApp [Publisher.ExampleApp] v2.0.0
...
-| `winget uninstall` | `PreUninstall` shown, prompt Y/n | `PostUninstall` shown | Standard uninstall |
-| `winget install --disable-interactivity` | `PreInstall` displayed, no prompt | `PostInstall` shown | Non-interactive install |
-| `winget install --silent` | `PreInstall` displayed, no prompt | `PostInstall` shown | Silent implies non-interactive |
-| Dependency packages | Not shown | Not shown | Messages apply to the root package only, not transitive dependencies |
-| `winget import` | Each package's `Pre*` shown, prompt Y/n per package | Each package's `Post*` shown | Follows the same per-package model |
-| WinGet Configuration (DSC) | Displayed in log output, no prompt | Displayed in log output | Configuration is inherently non-interactive |
+| `winget install` | `PreInstall` Displayed; Prompted (Y/n) | `PostInstall` Displayed | Standard single-package install |
+| `winget upgrade <package>` | `PreUpgrade` Displayed; Prompted (Y/n) | `PostUpgrade` Displayed | Single-package upgrade |
+| `winget upgrade --all` | All `PreUpgrade` messages Displayed together; Prompted (Y/n) once | Each `PostUpgrade` Displayed | See "Bulk Operations" below |
```
florelis, line 178: I assume "All messages displayed together, prompted once" also applies to other commands when passing multiple packages, like `winget install app1 app2`?

doc/specs/#3483 - UserMessages.md
```diff
@@ -141,67 +167,67 @@ Successfully installed ExampleApp [Publisher.ExampleApp] v2.0.0
...
+| `disableInstallNotes` setting enabled | `Pre*` Displayed | `PostInstall` Suppressed; other `Post*` Displayed | Backward compatibility behavior for migrated install notes |
+| `userExperience.userMessages.disablePostActionMessages` enabled | `Pre*` Displayed | `Post*` Suppressed | New setting for new fields |
+| `userExperience.userMessages.disablePreActionMessages` enabled | Suppressed; Prompt Suppressed | `Post*` Displayed | Suppresses both display and prompt |
+| `--show-user-messages` | `Pre*` Displayed; Prompted (Y/n) in interactive mode | `Post*` Displayed | Overrides settings that suppress messages for this invocation |
+| `--ignore-user-messages` | Suppressed; Prompt Suppressed | `Post*` Suppressed | CLI argument overrides settings for this invocation |
+| `--ignore-pre-action-messages` | Suppressed; Prompt Suppressed | `Post*` Displayed | CLI argument overrides `userExperience.userMessages.disablePreActionMessages` |
+| `--ignore-post-action-messages` | `Pre*` Displayed | `Post*` Suppressed | CLI argument overrides `userExperience.userMessages.disablePostActionMessages` |
```
florelis, line 194: We have global show/ignore options, and ignore options for pre/post messages. Why not show option for pre/post?

doc/specs/#3483 - UserMessages.md
```diff
@@ -141,67 +167,67 @@ Successfully installed ExampleApp [Publisher.ExampleApp] v2.0.0
...
+| `winget validate` | Suppressed | Suppressed | Validation checks schema correctness only; messages are not rendered |
 
 #### Bulk Operations (`upgrade --all`, `import`)
```
florelis, line 197: `install`, `upgrade`, etc also have bulk operations

doc/specs/#3483 - UserMessages.md
florelis, line 194: I assume we have some rules about mutual exclusivity for these?

doc/specs/#3483 - UserMessages.md
```diff
@@ -353,13 +366,15 @@ User Messages:
...
+  Pre-Repair: Close the application before starting repair.
+  Post-Repair: Restart the application to verify the repair completed successfully.
 ```
 
 No prompt is displayed — `show` is read-only.
 
 #### `winget repair`
 
-The `repair` command reuses `PreInstall` and `PostInstall` messages. The behavior is identical to `install` — pre-message with prompt in interactive mode, post-message after success.
+The `repair` command uses the dedicated `PreRepair` and `PostRepair` fields. Its behavior mirrors the other operation-specific flows — pre-message with prompt in interactive mode, post-message after success.
```
florelis, line 377: This could be removed since it is covered already

doc/specs/#3483 - UserMessages.md
```diff
@@ -506,23 +525,23 @@ Messages in locale manifests override those in the defaultLocale manifest on a p
...
 
-2. **Bulk operation noise.** `upgrade --all` with many packages containing pre-messages could be verbose and slow. A future enhancement to aggregate messages would address this, but is out of scope here.
+2. **Bulk operation noise.** `upgrade --all` with many packages containing pre-messages could still be verbose even when those messages are presented together before execution.
 
 3. **Transition-period duplication.** During the transition, manifest authors must duplicate their install note in both `InstallationNotes` and `UserMessages.PostInstall`. This is a known trade-off for backward compatibility and is temporary.
 
-4. **Content moderation burden.** The winget-pkgs community reviewers now have six new fields to review per manifest. Automated tooling to flag suspicious content in these fields would reduce this burden.
+4. **Content moderation burden.** The winget-pkgs community reviewers now have eight new fields to review per manifest. Automated tooling to flag suspicious content in these fields would reduce this burden.
```
florelis, line 544: Maybe we could have a special value that means "reuse the install message". For example, if the message is to alert the user that the installer will reboot the device, it would be the same for install/upgrade/repair and manifest authors shouldn't need to repeat it.

doc/specs/#3483 - UserMessages.md
```diff
@@ -141,67 +167,67 @@ Successfully installed ExampleApp [Publisher.ExampleApp] v2.0.0
...
-| `winget validate` | Not shown, not prompted | Not shown | Validation checks schema correctness only; messages are not rendered |
+| Operation failure | `Pre*` Displayed before attempt | Suppressed | Post-messages only appear on success |
+| `disableInstallNotes` setting enabled | `Pre*` Displayed | `PostInstall` Suppressed; other `Post*` Displayed | Backward compatibility behavior for migrated install notes |
+| `userExperience.userMessages.disablePostActionMessages` enabled | `Pre*` Displayed | `Post*` Suppressed | New setting for new fields |
+| `userExperience.userMessages.disablePreActionMessages` enabled | Suppressed; Prompt Suppressed | `Post*` Displayed | Suppresses both display and prompt |
+| `--show-user-messages` | `Pre*` Displayed; Prompted (Y/n) in interactive mode | `Post*` Displayed | Overrides settings that suppress messages for this invocation |
+| `--ignore-user-messages` | Suppressed; Prompt Suppressed | `Post*` Suppressed | CLI argument overrides settings for this invocation |
+| `--ignore-pre-action-messages` | Suppressed; Prompt Suppressed | `Post*` Displayed | CLI argument overrides `userExperience.userMessages.disablePreActionMessages` |
```
florelis, line 193: Nit: It feels like we are giving the messages two names - "action messages" and "user messages" and that may be confusing.

florelis, line 193: Nit: Maybe use "hide", "suppress" or "disable" instead of "ignore"? To me "ignore" doesn't necessarily mean that it won't be shown, just that it won't affect anything. Same as if we disable interactivity for pre* messages

doc/specs/#3483 - UserMessages.md
```diff
@@ -141,67 +167,67 @@ Successfully installed ExampleApp [Publisher.ExampleApp] v2.0.0
...
-| `--ignore-post-action-messages` | `Pre*` still shown | `Post*` suppressed | CLI argument overrides `disablePostActionMessages` setting |
-| `winget validate` | Not shown, not prompted | Not shown | Validation checks schema correctness only; messages are not rendered |
+| Operation failure | `Pre*` Displayed before attempt | Suppressed | Post-messages only appear on success |
+| `disableInstallNotes` setting enabled | `Pre*` Displayed | `PostInstall` Suppressed; other `Post*` Displayed | Backward compatibility behavior for migrated install notes |
+| `userExperience.userMessages.disablePostActionMessages` enabled | `Pre*` Displayed | `Post*` Suppressed | New setting for new fields |
+| `userExperience.userMessages.disablePreActionMessages` enabled | Suppressed; Prompt Suppressed | `Post*` Displayed | Suppresses both display and prompt |
+| `--show-user-messages` | `Pre*` Displayed; Prompted (Y/n) in interactive mode | `Post*` Displayed | Overrides settings that suppress messages for this invocation |
+| `--ignore-user-messages` | Suppressed; Prompt Suppressed | `Post*` Suppressed | CLI argument overrides settings for this invocation |
```
florelis, line 192: This sounds like it will suppress all output from winget. It may be good to have that since people have asked for it, but it may be confusing if ignoring "user messages" only includes "pre/post action messages"

doc/specs/#3483 - UserMessages.md
JohnMcPMS: Closest in your example is tough to say, maybe the thing worth mentioning was introduced in 1.2 and continues in 2.0.  But similarly, maybe it existed in 1.*, but 2.0 fixes it.  Tough call.

doc/specs/#3483 - UserMessages.md
```diff
@@ -100,7 +112,18 @@ When both `InstallationNotes` and `UserMessages.PostInstall` are present in a ma
...
+> [!NOTE]
+> Each operation displays only its corresponding `UserMessages` field. There is no cross-field fallback between operations — for example, if a manifest defines `PreInstall` but not `PreUpgrade`, no pre-action message is displayed during an upgrade.
+
+#### Uninstall Message Sourcing
+
+For uninstall operations, `PreUninstall` and `PostUninstall` are sourced as follows:
+
+- If the package is correlated to a specific installed version, use that version's manifest messages.
+- If the package is correlated but no target version manifest exists, use the latest available version's messages.
+- If the package is not correlated (for example, it was not installed through WinGet), no uninstall message is displayed.
```
JohnMcPMS, line 124: So no request to store the manifest/uninstall messages and display them if the version remains unchanged.

doc/specs/#3483 - UserMessages.md
JohnMcPMS: I would also assume that all bulk operations behave the same.  We probably want an additional message in bulk operations (maybe operation dependent) that indicates how the user should move forward if they want to opt-out of a single of the package operations (install -> remove from list, upgrade -> pin the package, etc.).

doc/specs/#3483 - UserMessages.md
JohnMcPMS: Or change the design from object to array with usage:
```
UserMessages:
  - Use: [PreInstall, PreUpgrade, PreRepair]
    Message: PreMessage
  - Use: [PostInstall, PostUpgrade, PostRepair]
    Message: PostMessage
```
Obviously this has an easier time of creating use overlap, but validation can easily catch that and local use can ignore the conflicts if present.

PR #6203, opened 2026-05-03
Add memory instructions for agent improvements across 🤖 users

Trenly:

@denelon
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6203)


.github/instructions/copilot-memory.instructions.md
```diff
@@ -0,0 +1,30 @@
...
+2. **Assess the scope.** Ask yourself: *Would other contributors to this repository benefit from knowing this? Would a future agent session make the same mistake without this guidance?* If yes, it is a candidate for shared repository memory (this file). If the correction is personal to this user (e.g., a workflow preference or output style), it is a candidate for user memory instead.
+
+3. **Ask the user before storing.** Do not silently write to either location. Instead, ask:
+
+   > "Should I remember this correction for the future? If so, should I add it to the **repository memory** (`copilot-memory.md`) so all contributors' agents benefit, or as a **user memory** (via the memory skill) that applies only to your sessions?"
+
+   Offer both options clearly and let the user decide.
+
+4. **Act on the answer.**
+   - If **repository memory**: add a concise, unambiguous entry under the appropriate section of this file (`copilot-memory.instructions.md`) and commit it.
```
florelis, line 24: Does it need to be added here? Depending on the correction, I think it may make more sense to add it to the main instructions file or some other instructions file we have.

Trenly: It doesn't have to be, no, but since all instructions work together, I figured it made sense to have it be it's own corrections that were made. A user can ask the agent to store it in the main instructions file too.

My main worry with throwing everything inside the main instructions file was maintainability. The main  instructions file could get very large if every update went into it.  By having the self-corrections go into a separate file it differentiates user intent from automated learnings. This way the main instructions file remains easier to review and keep accurate over time, and if any of these corrections / learnings need to be updated, they can be.  We could also add an authority statement that any directions from the primary instructions file override the memories here

florelis (approved):

.github/instructions/copilot-memory.instructions.md
JohnMcPMS, line 24: For repository memories, we should also consider if they should be scoped to prevent the root instructions from becoming too large.  For instance, your addition of the SQL builder should be scoped to the locations that use it (shared, repository, tests).

PR #6220, opened 2026-05-10
Omit prerelease versions from update selection.

4kbyte:
## 📖 Description

Prevent a prerelease version from being offered as an upgrade.

## 🔗 References

microsoft/winget-pkgs#368913

## 🔍 Validation

TDD: unit test.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6220)

### DIFF
--- a/src/AppInstallerCLITests/CompositeSource.cpp
+++ b/src/AppInstallerCLITests/CompositeSource.cpp
@@ -761,6 +761,46 @@ TEST_CASE("CompositePackage_AvailableVersions_NoChannelFilteredOut", "[Composite
     REQUIRE(latestVersion->GetProperty(PackageVersionProperty::Channel).get() == channel);
 }
 
+TEST_CASE("CompositePackage_PrereleaseAvailable_StableInstalled_NoUpdate", "[CompositeSource]")
+{
+    // microsoft/winget-pkgs#368913: a higher-precedence pre-release must not be
+    // offered as an upgrade to a stable installation.
+    auto availableVersion = GENERATE("3.14.5"sv, "3.14.5rc1"sv);
+    const bool isPrerelease = (availableVersion == "3.14.5rc1"sv);
+
+    CompositeTestSetup setup;
+    setup.Installed->Everything.Matches.emplace_back(setup.MakeInstalled().WithVersion("3.14.4"sv), Criteria());
+    setup.Available->SearchFunction = [&](const SearchRequest&)
+    {
+        auto manifest = MakeDefaultManifest(availableVersion);
+        // ManifestComparator::GetPreferredInstaller drops Architecture::Unknown on the host.
+        manifest.Installers[0].BaseInstallerType = Manifest::InstallerTypeEnum::Exe;
+        manifest.Installers[0].Arch = Utility::Architecture::Neutral;
+
+        SearchResult result;
+        result.Matches.emplace_back(
+            TestCompositePackage::Make(std::vector<Manifest::Manifest>{ manifest }, setup.Available),
+            Criteria());
+        return result;
+    };
+
+    SearchResult result = setup.Search();
+    REQUIRE(result.Matches.size() == 1);
+
+    auto latest = GetLatestApplicableVersion(result.Matches[0].Package);
+    if (isPrerelease)
+    {
+        REQUIRE_FALSE(latest.UpdateAvailable);
+        REQUIRE_FALSE(latest.LatestApplicableVersion);
+    }
+    else
+    {
+        REQUIRE(latest.UpdateAvailable);
+        REQUIRE(latest.LatestApplicableVersion);
+        REQUIRE(latest.LatestApplicableVersion->GetProperty(PackageVersionProperty::Version).get() == availableVersion);
+    }
+}
+
 TEST_CASE("CompositeSource_MultipleAvailableSources_MatchAll", "[CompositeSource]")
 {
     TestCommon::TestUserSettings testSettings;
--- a/src/AppInstallerRepositoryCore/PackageVersionSelection.cpp
+++ b/src/AppInstallerRepositoryCore/PackageVersionSelection.cpp
@@ -10,6 +10,20 @@ namespace AppInstaller::Repository
 {
     namespace
     {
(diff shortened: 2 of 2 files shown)


4kbyte:

Trenly:
This has the potential to be a breaking change for publishers who intentionally ship releases like this. Especially because there isn’t a switch that you've provided to bypass this behavior and haven’t updated any of the documentation to go with this change.

Your test case also doesn’t actually guarantee the sorting because `3.14.5` is always sorted as a higher version than `3.14.5rc1` regardless.

I understand the frustration that a release candidate slipping into stable packages causes, but I'm not sure that this check achieves the goal of preventing that in a meaningful way. Sure, it will stop some cases, but it breaks others and doesn’t capture the full spectum of behavior that WinGet has to account for. 

Something like #147 would seem more along the lines of what is needed here, or an actual flag in the manifest indicating that a release is a pre-release.
* #147 


Ultimately, though, this is a failure of the moderation in the source repository (winget-pkgs) and as a user of WinGet I'd rather see a proper fix there than a quick fix that has the potential to cause additional confusion when using the client.

4kbyte:
Intent: Make stable releases the default upgrade target.

Correct that SemVer sorts 3.14.5 higher than 3.14.5rc1 (SemVer 2.0.0 section 11). That's why the patch is focused on `GetLatestApplicableVersion` and not SemVer handling. The test compares installed 3.14.4 against a single available manifest per case, not 3.14.5 against 3.14.5rc1; no sort is asserted.
 
This is larger than user experience; it's chain of trust. The winget project has become load-bearing for unattended upgrades. The original issue microsoft/winget-pkgs#368913 observes that Microsoft Store invokes `winget.exe upgrade --scope user --accept-source-agreements --disable-interactivity` daily.
 
A switch and docs are straightforward to add upon consensus. Complementary to #147, not a substitute.

JohnMcPMS:
We can't make this change as the new default behavior without the potential for breaking something as @Trenly pointed out.  It is not a security feature, it is a hurdle for a mistake to get tripped up on.  Unfortunately it is susceptible to false positives in blocking.

If it were a setting that enabled this behavior, then `--force` could override it, that would be much easier to take without additional investigation.

florelis:
Maybe we could run this on all the packages we currently have to know exactly what packages would be affected. If we know that it would not cause many issues, it may be easier to take the change. Although there would still be risk with other sources.

JohnMcPMS:
> Maybe we could run this on all the packages we currently have to know exactly what packages would be affected. If we know that it would not cause many issues, it may be easier to take the change. Although there would still be risk with other sources.

I had written that previously and replaced it with a "no" because there could be private sources with arbitrary version information out there that we do not know about.  We cannot start blocking upgrades based on new rules without a lot more work than the code.  Even if we all agree this is a reasonable thing and we can show it doesn't affect winget-pkgs at all (unless mistakes show up and it does its job), we would have to broadly communicate this breaking behavior change well in advance and include a policy setting that could disable it.

Trenly:
I was also thinking that we could run it against all packages - but with over 12k packages and many thousands more individual package versions, it's a big lift to do that. I believe John recently had to do that for a change and it was not exactly fun. I've done similar scans when the repo was smaller and it still took several hours to complete.

Unless a repo scan is the only option for fixing something, it should be avoided; Not only because it may take significant time, or is logistically challenging (because new package versions are being added as the scan is performed, lag time between getting client updates and pkgs updates, etc.) but the implication of such a scan being required is that the change is a breaking change in some way

Karl-WE:
How about the approach not changing the default behaviour to avoid breaking changes, rather adding a new option similar to `--include-unknown`, which could be `--exclude-prereleases` or `--stable-releases`, inverting the default logic, excluding releases based on specifics such as these strings / regex?

ID contains 'preview'
Version contains '-rc*
Version contains '-preview*
Version contains '-pre'
Version contains '-alpha*

source for examples: `winget search preview`

denelon:
If we had the feature for release channels, that might make more sense. This is a challenging area in the world of Windows software and the various ways different distribution channels are handled.

Karl-WE:
I agree Demitrius. The approach defining channels could also be a worth a version 2.0 consideration. 

From the top of my mind:

- adding a new mandatory field in the YAML to declare if the software is a stable or any kind of preview channel.

- the current system of version description could be harmonized. This would benefit the new sorting feature. 
Currently there are a lot of ways to express this in the version description. E. g. -pre1 or -pre.1 etc. I am not sure where this is coming from. File properties from installer? Can Copilot help here to update the YAMLs at scale? 

- adding a new column in winget output after moniker to indicate the channel.

- adding new command switches as proposed above for winget and PowerShell.

- ensure that stable channel is the default and error handling if there is no applicable stable version available, but just one in preview so user get informed using the new switch.

denelon:
Our desire is to add support for release channels without a breaking change. That will mean that initially, it's an optional feature for the manifest schema. Once that feature has fully rolled out, and we believe we've covered the matrix of possibilities for how different packages behave, we will consider a breaking change to make some of those fields mandatory. 

The breaking change would likely be more of a policy decision for the community repository as opposed to a breaking change in the client. There is a large set of edge cases, and we don't want to break user experience until the last responsible moment.

Adding a mandatory field to the schema would require touching tens of thousands of manifests here in this repository, or something else equally complex and or creative.

ranm-msft (commented):
Not weighing in on the heuristic-vs-release-channels direction, which has already been covered above. One narrow testing gap on the code as written:

`installedIsPrerelease` is the escape hatch that preserves prerelease-to-prerelease upgrades, and it is the one branch the new test never exercises. Every generated case starts from a stable installed version, so the filter is always active and the bypass path is never executed. A user already on a prerelease build is exactly the population this change must not strand, so if the bypass ever regressed, the suite would stay green.

Could you add a case with a prerelease installed version and a newer prerelease available version, asserting the newer prerelease is still offered? That pins the behavior that makes the filter safe to enable.

PR #6222, opened 2026-05-12
Update `configure export --all` for recent DSC changes

JohnMcPMS:
## 📖 Description
Updates our `ExportAll` test to use `_` instead of `.` in the name of the subdirectory resource as DSC 3.2 is now strict about this.

Improves our DSC resource filtering mechanism to target some well-known resources and then exclude all resources that reside in that same location.  This will prevent new DSC provided resources from being included by default.

## 🔗 References
None

## 🔍 Validation
Manual runs of `export --all` to ensure filtering.
Regression tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6222)


yao-msft (approved):
:shipit:

PR #6224, opened 2026-05-12
Mitigate stack overflow issue

JohnMcPMS:
## 📖 Description
We see some very rare crashes where the stack is too small to support the WIL message formatting of `THROW_HR_MSG`.  This change attempts to mitigate that through a few targeted reductions in stack usage (doesn't get a lot but these were the only high ROI locations found) and by probing the remaining stack to decide whether to include the exception message or not on the throw.

## 🔗 References
None.

## 🔍 Validation
Manual run with forced error still includes the exception message.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6224)

### DIFF
--- a/src/AppInstallerSharedLib/Runtime.cpp
+++ b/src/AppInstallerSharedLib/Runtime.cpp
@@ -218,4 +218,19 @@ namespace AppInstaller::Runtime
     {
         return wil::get_token_information<TOKEN_ELEVATION_TYPE>() == TokenElevationTypeLimited;
     }
+
+    DECLSPEC_NOINLINE bool IsStackAvailable(size_t bytes)
+    {
+        // https://devblogs.microsoft.com/oldnewthing/20200610-00/?p=103855
+        ULONG_PTR low, high;
+        GetCurrentThreadStackLimits(&low, &high);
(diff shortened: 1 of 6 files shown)


yao-msft (dismissed):

src/AppInstallerSharedLib/Runtime.cpp
```diff
@@ -218,4 +218,19 @@ namespace AppInstaller::Runtime
...
         return wil::get_token_information<TOKEN_ELEVATION_TYPE>() == TokenElevationTypeLimited;
     }
+
+    DECLSPEC_NOINLINE bool IsStackAvailable(size_t bytes)
+    {
+        // https://devblogs.microsoft.com/oldnewthing/20200610-00/?p=103855
+        ULONG_PTR low, high;
+        GetCurrentThreadStackLimits(&low, &high);
+        auto remaining = reinterpret_cast<ULONG_PTR>(&low) - low;
+        if (remaining > high - low) {
```
yao-msft:
```suggestion
        if (remaining > high - low)
        {
```
nit

yao-msft (approved):

PR #6233, opened 2026-05-19
In-proc certificate pinning validation override

JohnMcPMS:
## 📖 Description
Adds the ability to override the certificate pinning validation with their own handler for in-proc COM callers.  In-proc callers can provide a handler delegate on their `PackageCatalogReference` object(s) before calling `Connect` and will receive a callback when winget would execute a certificate pinning validation.  If they accept the server connection, the certificate will be cached as it is with the internal check and automatically approved for any future connections.

Note that `Connect` may not actually trigger the callback if the `/information` for the catalog is already cached, but you must set the callback before the `Connect` call for it to be attached to the connected catalog for future use.

The callback can only be set for the MS Store catalog when the group policy `BypassCertificatePinningForMicrosoftStore` is not configured.

## 🔍 Validation
Added a sample CLI caller that provides the user with a prompt to accept or reject the connection.  Manual validation of logs and caching behavior.
Added E2E tests for the callback.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6233)

### DIFF
--- a/samples/ConnectionValidationSample/ConnectionValidationSample.csproj
+++ b/samples/ConnectionValidationSample/ConnectionValidationSample.csproj
@@ -38,10 +38,6 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Link>Microsoft.Management.Deployment.dll</Link>
     </Content>
-    <Content Include="$(WinGetSrcBinRoot)\Microsoft.Management.Deployment.InProc\Microsoft.Management.Deployment.InProc.pdb">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>Microsoft.Management.Deployment.InProc.pdb</Link>
-    </Content>
     <Content Include="$(WinGetSrcBinRoot)\WindowsPackageManager\WindowsPackageManager.dll">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Link>WindowsPackageManager.dll</Link>
--- a/samples/ConnectionValidationSample/ConnectionValidationSample.csproj
+++ b/samples/ConnectionValidationSample/ConnectionValidationSample.csproj
@@ -1,0 +1,55 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0-windows10.0.26100.0</TargetFramework>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Platforms>x64;x86;arm64</Platforms>
+    <!-- Override these to match the platform/configuration you built the solution with. -->
+    <WinGetBuildPlatform Condition="'$(WinGetBuildPlatform)' == ''">x64</WinGetBuildPlatform>
+    <WinGetBuildConfiguration Condition="'$(WinGetBuildConfiguration)' == ''">Debug</WinGetBuildConfiguration>
+    <!-- Platform-specific build output root. -->
+    <WinGetSrcBinRoot>$([MSBuild]::NormalizePath('$(MSBuildThisFileDirectory)..\..\src\$(WinGetBuildPlatform)\$(WinGetBuildConfiguration)'))</WinGetSrcBinRoot>
+  </PropertyGroup>
+
+  <!--
+    Generates a CsWinRT projection from the built winmd so this sample needs no published NuGet package.
+    Build the solution first (src\AppInstallerCLI.sln), then pass a catalog name as the argument.
+    Override the defaults if needed: -p:WinGetBuildPlatform=x64 -p:WinGetBuildConfiguration=Release
+  -->
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.2.0" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <CsWinRTWindowsMetadata>10.0.26100.0</CsWinRTWindowsMetadata>
+    <CsWinRTIncludes>Microsoft.Management.Deployment</CsWinRTIncludes>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <CsWinRTInputs Include="$(WinGetSrcBinRoot)\Microsoft.Management.Deployment\Microsoft.Management.Deployment.winmd" />
+  </ItemGroup>
+
+  <!-- Copy the native in-proc COM DLL next to the sample so CsWinRT can activate it. -->
+  <ItemGroup>
+    <Content Include="$(WinGetSrcBinRoot)\Microsoft.Management.Deployment.InProc\Microsoft.Management.Deployment.dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>Microsoft.Management.Deployment.dll</Link>
+    </Content>
+    <Content Include="$(WinGetSrcBinRoot)\Microsoft.Management.Deployment.InProc\Microsoft.Management.Deployment.InProc.pdb">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>Microsoft.Management.Deployment.InProc.pdb</Link>
+    </Content>
+    <Content Include="$(WinGetSrcBinRoot)\WindowsPackageManager\WindowsPackageManager.dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>WindowsPackageManager.dll</Link>
+    </Content>
+    <Content Include="$(WinGetSrcBinRoot)\WindowsPackageManager\WindowsPackageManager.pdb">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>WindowsPackageManager.pdb</Link>
+    </Content>
+  </ItemGroup>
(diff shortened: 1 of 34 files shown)


src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -859,6 +859,29 @@ namespace Microsoft.Management.Deployment
...
     }
 
+    /// Result of a connection validation callback.
+    [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 29)]
+    enum PackageCatalogConnectionValidationResult
+    {
+        /// The connection was accepted.
+        Ok,
+        /// The connection was rejected because the certificate was not accepted.
+        CertificateRejected,
```
florelis, line 869: Should we add a generic "rejected" result? It feels weird to have this setup that looks like you could do any validation you want with it, but you're only allowed to do one specific validation.

src/Microsoft.Management.Deployment/PackageCatalogReference.h
```diff
@@ -32,9 +32,12 @@ namespace winrt::Microsoft::Management::Deployment::implementation
         void InstalledPackageInformationOnly(bool value);
         winrt::Microsoft::Management::Deployment::AuthenticationArguments AuthenticationArguments();
         void AuthenticationArguments(winrt::Microsoft::Management::Deployment::AuthenticationArguments const& value);
-        winrt::Microsoft::Management::Deployment::AuthenticationInfo AuthenticationInfo();
-        // Contract 12.0
+        winrt::Microsoft::Management::Deployment::AuthenticationInfo AuthenticationInfo();
+        // Contract 12.0
         winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Management::Deployment::RefreshPackageCatalogResult, double> RefreshPackageCatalogAsync();
```
florelis, line 37: nit: not your change, but this spacing looks weird

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
florelis, line 215: nit: it bothers me that this isn't in the same order as the other one

src/Microsoft.Management.Deployment.Projection/Microsoft.Management.Deployment.Projection.csproj
```diff
@@ -16,8 +16,30 @@
...
+      Windows.Networking.DomainNameType;
+      Windows.Networking.IEndpointPair;
+      Windows.Networking.EndpointPair;
+      Windows.Networking.IHostName;
+      Windows.Networking.HostName;
+      Windows.Security.Cryptography.Certificates;
+      Windows.Storage;
+      Windows.Storage.Provider.FileUpdateStatus;
+      Windows.System.ProcessorArchitecture;
+      Windows.System.IUser;
```
florelis, line 36: Why does this need to include so many more types now?

src/AppInstallerSharedLib/Public/winget/Certificates.h
```diff
@@ -163,22 +174,38 @@ namespace AppInstaller::Certificates
...
+        double GetRemainingLifetimePercentage() const override;
 
     private:
         std::vector<PinningDetails> m_chain;
         bool m_partial = false;
     };
 
+    // A chain validation that delegates to a caller-supplied callback.
+    // Construction enforces that the BypassCertificatePinningForMicrosoftStore group policy is not configured;
+    // throws APPINSTALLER_CLI_ERROR_BLOCKED_BY_POLICY if it is.
```
florelis: Not true?

src/Microsoft.Management.Deployment/PackageManager.idl
```diff
@@ -967,6 +990,13 @@ namespace Microsoft.Management.Deployment
             /// The progress range is from 0 to 100.
             Windows.Foundation.IAsyncOperationWithProgress<RefreshPackageCatalogResult, Double> RefreshPackageCatalogAsync();
         }
+
+        [contract(Microsoft.Management.Deployment.WindowsPackageManagerContract, 29)]
+        {
+            /// A callback invoked to validate the server certificate during Connect or ConnectAsync.
+            /// Only available to in-process callers; out-of-process callers will receive E_ACCESSDENIED on set.
+            PackageCatalogConnectionValidationHandler ConnectionValidationHandler;
```
florelis, line 1000: Should we mention the group policy here?
Do callers have a way to check for the group policy, or do we expect them to always wrap this in a try-catch? If they need to always use a try-catch, should we change this to a `boolean TrySetConnectionValidationHandler()` method to prevent people from not handling that case?

src/Microsoft.Management.Deployment/PackageManager.idl
JohnMcPMS: My expectation was that there might be additional reasons in the future, possibly with more values coming into the callback.  The reason is mostly to allow for our own logging and debugging.

src/Microsoft.Management.Deployment/PackageCatalogReference.h
JohnMcPMS: I'm not following.  Not having a newline to start the new Contract comment?

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
JohnMcPMS: And it is actually an issue.

src/Microsoft.Management.Deployment.Projection/Microsoft.Management.Deployment.Projection.csproj
JohnMcPMS: `Windows.Security.Cryptography.Certificates.Certificate` and its transitive needs.

src/Microsoft.Management.Deployment/PackageManager.idl
JohnMcPMS: Yes to comment, no to try.

samples/ConnectionValidationSample/ConnectionValidationSample.csproj
```diff
@@ -0,0 +1,55 @@
...
+
+  <!-- Copy the native in-proc COM DLL next to the sample so CsWinRT can activate it. -->
+  <ItemGroup>
+    <Content Include="$(WinGetSrcBinRoot)\Microsoft.Management.Deployment.InProc\Microsoft.Management.Deployment.dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>Microsoft.Management.Deployment.dll</Link>
+    </Content>
+    <Content Include="$(WinGetSrcBinRoot)\Microsoft.Management.Deployment.InProc\Microsoft.Management.Deployment.InProc.pdb">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>Microsoft.Management.Deployment.InProc.pdb</Link>
```
yao-msft: nit: Does it need to be renamed to remove the .inproc part for the debugger to automatically pick up the pdb?

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -330,4 +361,27 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+    void PackageCatalogReference::ConnectionValidationHandler(winrt::Microsoft::Management::Deployment::PackageCatalogConnectionValidationHandler const& value)
+    {
+        using namespace AppInstaller::Settings;
+
+        auto [hr, callerProcessId] = GetCallerProcessId();
+        THROW_IF_FAILED(hr);
+        THROW_HR_IF(E_ACCESSDENIED, callerProcessId != GetCurrentProcessId());
+
+        // For the well-known Microsoft Store source, the callback is not allowed when the
+        // certificate pinning bypass policy is configured (in either direction).
```
yao-msft: I think Intune customers will likely configure the policy to be false, and this will block Intune's custom handler in that case.

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
```diff
@@ -330,4 +361,27 @@ namespace winrt::Microsoft::Management::Deployment::implementation
...
+        return m_connectionValidationHandler;
+    }
+
+    void PackageCatalogReference::ConnectionValidationHandler(winrt::Microsoft::Management::Deployment::PackageCatalogConnectionValidationHandler const& value)
+    {
+        using namespace AppInstaller::Settings;
+
+        auto [hr, callerProcessId] = GetCallerProcessId();
+        THROW_IF_FAILED(hr);
+        THROW_HR_IF(E_ACCESSDENIED, callerProcessId != GetCurrentProcessId());
```
yao-msft: nit: this could be in a helper function like IsInProcCaller()

src/Microsoft.Management.Deployment/PackageCatalogReference.cpp
JohnMcPMS: I will discuss with them more about this.

samples/ConnectionValidationSample/ConnectionValidationSample.csproj
JohnMcPMS: It probably does, but we could also just remove that pdb as well since it is just a thunk into the implementation dll.

src/Microsoft.Management.Deployment/PackageCatalogReference.h
florelis: Nevermind. It looked like there was an empty line after the `// Contract 12.0`, but it is just the wrapping 

Image

florelis (approved):

PR #6236, opened 2026-05-20
Move pre-check errors to post-check

JohnMcPMS:
## 📖 Description
We had some pre-check errors for various issues in the OS.  Move those to attempt the operation and convert any failures in the scenario on the older OS to the well-known HRESULT.

## 🔍 Validation
Existing regression tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6236)

### DIFF
--- a/src/AppInstallerCommonCore/MSStore.cpp
+++ b/src/AppInstallerCommonCore/MSStore.cpp
@@ -401,15 +401,6 @@ namespace AppInstaller::MSStore
 
         if (m_scope == Manifest::ScopeEnum::Machine)
         {
-            // TODO: There was a bug in InstallService where admin user is incorrectly identified as not admin,
-            // causing false access denied on many OS versions.
-            // Remove this check when the OS bug is fixed and back ported.
-            if (!Runtime::IsRunningAsSystem())
-            {
-                AICLI_LOG(Core, Error, << "Device wide install for msstore type is not supported under admin context.");
-                return APPINSTALLER_CLI_ERROR_INSTALL_SYSTEM_NOT_SUPPORTED;
-            }
-
             installOptions.InstallForAllUsers(true);
         }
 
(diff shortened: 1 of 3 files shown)


src/AppInstallerCommonCore/MSStore.cpp
```diff
@@ -438,6 +429,18 @@ namespace AppInstaller::MSStore
 
         HRESULT hr = WaitForOperation(m_productId, m_isSilentMode, installItems, progress, monitor);
 
+        // There was a bug in InstallService where admin users were incorrectly identified as non-admin,
+        // causing false access denied errors; fixed in 10.0.26100.0. On older OS versions, convert any
+        // failure under these conditions to the error that was previously always returned.
+        if (FAILED(hr) &&
```
florelis, line 435: Should we check for the specific HR?

florelis (approved):

src/AppInstallerCommonCore/MSStore.cpp
```diff
@@ -438,6 +429,18 @@ namespace AppInstaller::MSStore
...
+        // There was a bug in InstallService where admin users were incorrectly identified as non-admin,
+        // causing false access denied errors; fixed in 10.0.26100.0. On older OS versions, convert any
+        // failure under these conditions to the error that was previously always returned.
+        if (FAILED(hr) &&
+            m_scope == Manifest::ScopeEnum::Machine &&
+            !Runtime::IsRunningAsSystem() &&
+            !Runtime::IsCurrentOSVersionGreaterThanOrEqual(Utility::Version{ "10.0.26100.0" }))
+        {
+            AICLI_LOG(Core, Error, << "Device wide install for msstore type is not supported under admin context on this OS version. Error: " << hr);
+            return APPINSTALLER_CLI_ERROR_INSTALL_SYSTEM_NOT_SUPPORTED;
```
yao-msft, line 441: I know it's a copy paste but nit: RETURN_HR()

src/AppInstallerCommonCore/MSStore.cpp
JohnMcPMS: I don't know that we know it nor do we know that it is specific to the failure.  That could be a future improvement but I don't see it causing any impact with this change.

PR #6243, opened 2026-05-22
Make sfs-client a vcpkg port

florelis:
## 📖 Description

Removes the git-subtree for sfs-client and replaces it with a local vcpkg port. This port is based on the [template provided by sfs-client](https://github.com/microsoft/sfs-client/tree/main/sfs-client-vcpkg-port/sfs-client)

* Updated the script that creates a local overlay to fetch this port.
* Added a small patch to fix a build error in the port.
* Created a new folder for the custom port patches. This means duplicating the files, but prevents losing them when re-creating the ports.
* Added a job to the build pipeline to ensure that the ports still match what the creation script produces, to ensure they're kept in sync.

This fixes a CG alert for c-ares triggered by having sfs-client's cgmanifest checked in, which caused CG to consider that as the version used despite us updating the dependency.

## 🔍 Validation

It builds :D

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6243)


src/VcpkgPortOverlay/patches/sfs-client/remove-unconditional-toolchain-override.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
JohnMcPMS, line 1: Why are there 2 copies of this patch?  And also the cpprestsdk patch.

azure-pipelines.yml
```diff
@@ -720,3 +720,28 @@ jobs:
...
       SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+
+- job: 'VerifyPortOverlay'
+  displayName: 'Verify Port Overlay'
+  variables:
+    runCodesignValidationInjection: ${{ false }}
+    skipComponentGovernanceDetection: ${{ true }}
+
+  steps:
+  - task: PowerShell@2
```
JohnMcPMS:
```
Consider migrating to PowerShell 7+, where Invoke-WebRequest uses secure parsing by default and does not depend on Internet Explorer components.
```
The pipeline failure appears to be caused by Windows PowerShell.

src/VcpkgPortOverlay/patches/sfs-client/remove-unconditional-toolchain-override.patch
florelis: `VcpkgPortOverlay/patches` is there as the source of truth for the ports. The idea is that you could nuke everything else under `VcpkgPortOverlay/` and recreate it with the script.

The other copies, and everything else in the ports, are there as a convenience so that you don't need to run the script locally as the first step of building. I added the pipeline check to make sure that these don't get outdated.

A better solution would be keep only the `patches/` folder and have MSBuild generate everything else (and ignore it for git). I was just too lazy to do that. I'll see if I can get it to work.

JohnMcPMS (approved):

PR #6251, opened 2026-05-26
Ensure portable command alias does not escape directory

florelis:
## 📖 Description

This adds validation to ensure that a `PortableCommandAlias` does not use relative paths to point outside of its directory. It uses the same validation we already have for `RelativeFilePath`

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6251)

### DIFF
--- a/src/AppInstallerSharedLib/Filesystem.cpp
+++ b/src/AppInstallerSharedLib/Filesystem.cpp
@@ -370,19 +370,11 @@ namespace AppInstaller::Filesystem
         return (GetVolumeInformationFlags(path) & FILE_SUPPORTS_REPARSE_POINTS) != 0;
     }
 
-    bool PathEscapesBaseDirectory(const std::filesystem::path& target, const std::filesystem::path& base)
-    {
-        const auto& targetPath = std::filesystem::weakly_canonical(target);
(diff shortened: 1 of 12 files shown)


src/AppInstallerSharedLib/Filesystem.cpp
```diff
@@ -378,6 +378,13 @@ namespace AppInstaller::Filesystem
         return (b != basePath.end());
     }
 
+    bool PathEscapesBaseDirectory(std::string_view relativePath)
+    {
+        const std::filesystem::path& basePath = std::filesystem::current_path();
```
JohnMcPMS: If we don't already have one, I would like a test that checks how `PathEscapesBaseDirectory` works when the current directory is `C:\` and generally what happens when the `..`s outnumber the actual directory count.

florelis: Updated the implementation to not use the current directory.

JohnMcPMS (approved):

PR #6270, opened 2026-06-06
Fall back to hard link when symlink creation lacks privilege

### LINKED ISSUE #6058: WinGet install with PortableCommandAlias gives wrong message when installed with non elevated privileges
### Relevant area(s)

WinGet CLI

### Relevant command(s)

winget install

On a non elevated shell, when installing a package with a PortableCommandAlias in its winget-pkgs manifest, the installer wrongly outputs "Command line alias added: <x>", which is not true. See: https://github.com/microsoft/winget-cli/issues/6043#issuecomment-3917408050

Image

Image

As this did not happen, it should not output this message.

### Steps to reproduce

`winget install int128.kubelogin`

### Expected behavior

Found kubelogin [int128.kubelogin] Version 1.35.2
Extracting archive...
Successfully extracted archive
Starting package install...
Successfully installed

### Actual behavior

Found kubelogin [int128.kubelogin] Version 1.35.2
Extracting archive...
Successfully extracted archive
Starting package install...
Command line alias added: "kubectl-oidc_login" #< wrong message
Successfully installed

### Environment

```shell
[winget --info]
Windows Package Manager v1.28.190

Windows: Windows.Desktop v10.0.22631.6491
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.28.190.0

Winget Directories
-------------------------------------------------------------------------------------------------------------------------------
Logs                               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir
User Settings                      %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\settings.json
Portable Package Root (User)       %LOCALAPPDATA%\Microsoft\WinGet\Packages
Portable Package Root              C:\Program Files\WinGet\Packages
Installer Downloads                %USERPROFILE%\Downloads
Configuration Modules              %LOCALAPPDATA%\Microsoft\WinGet\Configuration\Modules

Links

ProxyCommandLineOptions                   Disabled
DefaultProxy                              Disabled
```

rich-purnell:
### 👤 User Story

When WinGet runs without administrator privileges or Developer Mode enabled, `std::filesystem::create_symlink()` fails with the error:

```
Error code: 1314 - create_symlink: A required privilege is not held by the client.
```

This happens because creating symbolic links requires the `SeCreateSymbolicLinkPrivilege`, which is only granted to administrators or available when Developer Mode is turned on. While WinGet typically runs with those privileges, there are scenarios where they are not held — for example, when running as a standard user or in a restricted environment.

I wrote a local test program that confirmed this behavior. Under those conditions, `std::filesystem::create_symlink(target, link)` fails with the above error, but `std::filesystem::create_hard_link(target, link)` succeeds. Since a hard link provides identical file resolution behavior for this use case (pointing to the same file on disk), it is a viable fallback.

This change adds a fallback: when symlink creation fails specifically with `ERROR_PRIVILEGE_NOT_HELD`, the code attempts a hard link instead before returning `false`.

Such a change may not solve all problems (for example, in some special file system structures), but it can reduce the impact scope.

### ⚙️ Core Modification

The symlink management scope is narrowed from **Portable packages** to **individual executable files**. The per-executable symlink creation flow is illustrated below:

```mermaid
flowchart TD
    A(Start)
    B[Create symlink]
    C{Succeeded?}
    D[Exit successfully]
    E{Permission denied?}
    F[Create Hard Link]
    G{Succeeded?}
    H[Add install directory to PATH]
    I[Error]
    Z(End)
    
    A --> B --> C
    C -->|Yes| D
    C -->|No| E
    E -->|Yes| F --> G
    E -->|No| I --> Z
    G -->|Yes| D
    G -->|No| H --> D
    D --> Z
```

Since a single Portable package may contain **multiple executables requiring symlinks** distributed across **different paths**, the outcome of one symlink operation **cannot be assumed** for others. Each executable must undergo independent validation.

Impact:

1. The `InstallDirectoryAddedToPath` variable is no longer used to control symlink creation. A full attempt sequence is executed **for each individual executable** during every creation attempt.
2. The `m_hasCreatedSymlink` and `m_hasRemovedSymlink` variables determine whether to add/remove the **Links path** from PATH. Since the user's environment may differ during each install/update/uninstall operation, **these two values must not be persisted**. Instead, the user's environment must be determined by **function return values** during each install/update/uninstall process.

## 🔗 References

Resolves #6058

## 🔍 Validation

Build succeeds with no errors or warnings.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6270)

### DIFF
--- a/src/AppInstallerCLICore/PortableInstaller.cpp
+++ b/src/AppInstallerCLICore/PortableInstaller.cpp
@@ -126,7 +126,7 @@ namespace AppInstaller::CLI::Portable
         {
             std::filesystem::path symlinkTargetPath{ Utility::ConvertToUTF16(entry.SymlinkTarget) };
 
-            if (BinariesDependOnPath && !InstallDirectoryAddedToPath)
+            if (BinariesDependOnPath)
             {
                 // Scenario indicated by 'ArchiveBinariesDependOnPath' manifest entry.
                 // Skip symlink creation for portables dependent on binaries that require the install directory to be added to PATH.
@@ -135,7 +135,7 @@ namespace AppInstaller::CLI::Portable
                 AICLI_LOG(Core, Info, << "Install directory added to PATH: " << installDirectory);
                 CommitToARPEntry(PortableValueName::InstallDirectoryAddedToPath, InstallDirectoryAddedToPath = true);
             }
-            else if (!InstallDirectoryAddedToPath)
+            else
             {
                 std::filesystem::file_status status = std::filesystem::status(filePath);
                 if (std::filesystem::is_directory(status))
@@ -191,7 +191,7 @@ namespace AppInstaller::CLI::Portable
             else if (InstallDirectoryAddedToPath)
             {
                 // If symlink doesn't exist, check if install directory was added to PATH directly and remove.
-                RemoveFromPathVariable(std::filesystem::path(Utility::ConvertToUTF16(entry.SymlinkTarget)).parent_path());
+                RemoveFromPathVariable(std::filesystem::path(Utility::ConvertToUTF16(entry.SymlinkTarget)).parent_path(), false);
             }
         }
         else if (fileType == PortableFileType::Directory && std::filesystem::exists(filePath))
@@ -370,9 +370,9 @@ namespace AppInstaller::CLI::Portable
         }
     }
 
-    void PortableInstaller::RemoveFromPathVariable(std::filesystem::path value)
+	void PortableInstaller::RemoveFromPathVariable(std::filesystem::path value, bool checkIfEmpty /*= true*/)
     {
-        if (std::filesystem::exists(value) && !std::filesystem::is_empty(value))
+		if (checkIfEmpty && std::filesystem::exists(value) && !std::filesystem::is_empty(value))
         {
             AICLI_LOG(Core, Info, << "Install directory is not empty: " << value);
         }
@@ -382,7 +382,6 @@ namespace AppInstaller::CLI::Portable
             // Necessary for handling old path values associated with winget-cli#5033
             if (PathVariable(GetScope()).Remove(value) || PathVariable(GetScope()).Remove(value.make_preferred()))
             {
-                InstallDirectoryAddedToPath = false;
                 AICLI_LOG(CLI, Info, << "Removed target directory from PATH registry: " << value);
             }
             else
--- a/src/AppInstallerCLICore/PortableInstaller.cpp
+++ b/src/AppInstallerCLICore/PortableInstaller.cpp
@@ -157,7 +157,13 @@ namespace AppInstaller::CLI::Portable
 
                 if (Filesystem::CreateSymlink(symlinkTargetPath, filePath))
                 {
+                    m_hasCreatedSymlink = true;
                     AICLI_LOG(Core, Info, << "Symlink created at: " << filePath << " with target path: " << symlinkTargetPath);
+                }
+                else if (Filesystem::CreateFileHardLink(symlinkTargetPath, filePath))
+                {
+                    m_hasCreatedSymlink = true;
+                    AICLI_LOG(Core, Info, << "Hardlink created at: " << filePath << " with target path: " << symlinkTargetPath);
                 }
                 else
                 {
@@ -187,8 +193,15 @@ namespace AppInstaller::CLI::Portable
             {
                 AICLI_LOG(CLI, Info, << "Deleting portable symlink at: " << filePath);
                 std::filesystem::remove(filePath);
+				m_hasRemovedSymlink = true;
             }
-            else if (InstallDirectoryAddedToPath)
+            else if (std::filesystem::exists(filePath) && std::filesystem::is_regular_file(std::filesystem::status(filePath)))
+            {
+				AICLI_LOG(CLI, Info, << "Deleting portable hard link at: " << filePath);
+				std::filesystem::remove(filePath);
+                m_hasRemovedSymlink = true;
+            }
+            else
             {
                 // If symlink doesn't exist, check if install directory was added to PATH directly and remove.
                 RemoveFromPathVariable(std::filesystem::path(Utility::ConvertToUTF16(entry.SymlinkTarget)).parent_path(), false);
@@ -289,7 +302,7 @@ namespace AppInstaller::CLI::Portable
 
         ApplyDesiredState();
 
-        if (!InstallDirectoryAddedToPath)
+        if (m_hasCreatedSymlink)
         {
             AddToPathVariable(GetPortableLinksLocation(GetScope()));
         }
@@ -302,20 +315,20 @@ namespace AppInstaller::CLI::Portable
         }
     }
 
-    void PortableInstaller::Uninstall()
-    {
-        ApplyDesiredState();
+	void PortableInstaller::Uninstall()
+	{
+		ApplyDesiredState();
 
-        RemoveInstallDirectory();
+		RemoveInstallDirectory();
 
-        if (!InstallDirectoryAddedToPath)
-        {
-            RemoveFromPathVariable(GetPortableLinksLocation(GetScope()));
-        }
+		if (m_hasRemovedSymlink)
+		{
+			RemoveFromPathVariable(GetPortableLinksLocation(GetScope()));
+		}
 
-        m_portableARPEntry.Delete();
-        AICLI_LOG(CLI, Info, << "PortableARPEntry deleted.");
-    }
+		m_portableARPEntry.Delete();
+		AICLI_LOG(CLI, Info, << "PortableARPEntry deleted.");
+	}
 
     void PortableInstaller::CreateTargetInstallDirectory()
     {
@@ -371,25 +384,25 @@ namespace AppInstaller::CLI::Portable
     }
 
 	void PortableInstaller::RemoveFromPathVariable(std::filesystem::path value, bool checkIfEmpty /*= true*/)
-    {
+	{
 		if (checkIfEmpty && std::filesystem::exists(value) && !std::filesystem::is_empty(value))
-        {
-            AICLI_LOG(Core, Info, << "Install directory is not empty: " << value);
-        }
-        else
-        {
+		{
+			AICLI_LOG(Core, Info, << "Install directory is not empty: " << value);
+		}
+		else
+		{
 			// Attempt to remove both the original and the preferred format to ensure removal
-            // Necessary for handling old path values associated with winget-cli#5033
-            if (PathVariable(GetScope()).Remove(value) || PathVariable(GetScope()).Remove(value.make_preferred()))
-            {
-                AICLI_LOG(CLI, Info, << "Removed target directory from PATH registry: " << value);
-            }
-            else
-            {
-                AICLI_LOG(CLI, Info, << "Target directory not removed from PATH registry: " << value);
-            }
-        }
-    }
+			// Necessary for handling old path values associated with winget-cli#5033
+			if (PathVariable(GetScope()).Remove(value) || PathVariable(GetScope()).Remove(value.make_preferred()))
+			{
+				AICLI_LOG(CLI, Info, << "Removed target directory from PATH registry: " << value);
+			}
+			else
+			{
+				AICLI_LOG(CLI, Info, << "Target directory not removed from PATH registry: " << value);
+			}
+		}
+	}
 
     void PortableInstaller::SetAppsAndFeaturesMetadata(const Manifest::Manifest& manifest, const std::vector<AppInstaller::Manifest::AppsAndFeaturesEntry>& entries)
     {
--- a/src/AppInstallerCLICore/PortableInstaller.cpp
+++ b/src/AppInstallerCLICore/PortableInstaller.cpp
@@ -194,11 +194,6 @@ namespace AppInstaller::CLI::Portable
                 RemoveFromPathVariable(std::filesystem::path(Utility::ConvertToUTF16(entry.SymlinkTarget)).parent_path());
             }
         }
-        else if (fileType == PortableFileType::Symlink && Filesystem::SymlinkExists(filePath))
-        {
-            AICLI_LOG(CLI, Info, << "Deleting portable symlink at: " << filePath);
-            std::filesystem::remove(filePath);
-        }
         else if (fileType == PortableFileType::Directory && std::filesystem::exists(filePath))
         {
             AICLI_LOG(CLI, Info, << "Removing directory at " << filePath);
--- a/src/AppInstallerCLICore/PortableInstaller.h
+++ b/src/AppInstallerCLICore/PortableInstaller.h
@@ -114,6 +114,6 @@ namespace AppInstaller::CLI::Portable
         void RemoveInstallDirectory();
 
         void AddToPathVariable(std::filesystem::path value);
-        void RemoveFromPathVariable(std::filesystem::path value);
+        void RemoveFromPathVariable(std::filesystem::path value, bool checkIfEmpty = true);
     };
 }
--- a/src/AppInstallerCLICore/PortableInstaller.h
+++ b/src/AppInstallerCLICore/PortableInstaller.h
@@ -35,6 +35,7 @@ namespace AppInstaller::CLI::Portable
         bool InstallDirectoryCreated = false;
         bool BinariesDependOnPath = false;
         // If we fail to create a symlink, add install directory to PATH variable
+        // TODO: Variable is redundant, remove it.
         bool InstallDirectoryAddedToPath = false;
 
         bool IsUpdate = false;
@@ -94,6 +95,9 @@ namespace AppInstaller::CLI::Portable
         AppInstaller::Manifest::AppsAndFeaturesEntry GetAppsAndFeaturesEntry();
 
     private:
+        // True if any symlink creation succeeds. No persistence required.
+        bool m_hasCreatedSymlink = false;
+        bool m_hasRemovedSymlink = false;
         PortableARPEntry m_portableARPEntry;
         std::vector<AppInstaller::Portable::PortableFileEntry> m_desiredEntries;
         std::vector<AppInstaller::Portable::PortableFileEntry> m_expectedEntries;
--- a/src/AppInstallerCLITests/InstallFlow.cpp
+++ b/src/AppInstallerCLITests/InstallFlow.cpp
@@ -731,7 +731,7 @@ TEST_CASE("InstallFlow_Portable_SymlinkCreationFail", "[InstallFlow][workflow]")
     TestCommon::TempDirectory tempDirectory("TestPortableInstallRoot", false);
     std::ostringstream installOutput;
     TestContext installContext{ installOutput, std::cin };
-    auto PreviousThreadGlobals = installContext.SetForCurrentThread();
+    installContext.SetForCurrentThread();
     OverridePortableInstaller(installContext);
     TestHook::SetCreateSymlinkResult_Override createSymlinkResultOverride(false);
     const auto& targetDirectory = tempDirectory.GetPath();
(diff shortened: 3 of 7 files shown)


rich-purnell:

Trenly (changes requested):
The uninstall flow may need to account for hard links with this change. Tests should be added to validate hardlink creation, path resolution, and removal.

What happens in the upgrade scenario where a hardlink exists but the user now has symlink permissions - does it stay a hardlink, or should it be replaced with a symlink?

rich-purnell:
> The uninstall flow may need to account for hard links with this change. Tests should be added to validate hardlink creation, path resolution, and removal.
> 
> What happens in the upgrade scenario where a hardlink exists but the user now has symlink permissions - does it stay a hardlink, or should it be replaced with a symlink?

I see that in the current installation flow, before creating a symlink, `std::filesystem::remove(filePath)` is called to remove the existing symlink. My testing shows that it also correctly removes hard links. So when a user gains permissions, reinstallation will remove the hard link and then create a symlink. I did overlook the uninstall part – I will check it later. Regarding adding new tests, since this is my first time working on this project, I need some time to understand which scenarios are already covered by the existing tests.

Kobi-Blade:
Hard links only work when the link and target are on the same NTFS volume, and they cannot link directories at all.

WinGet needs cross volume and directory level linking for portable packages, and it must track and remove links cleanly during upgrades.

Symlinks support all of that, hard links don’t.

sysopmd:
> Hard links only work when the link and target are on the same NTFS volume, and they cannot link directories at all.
> 
> WinGet needs cross volume and directory level linking for portable packages, and it must track and remove links cleanly during upgrades.
> 
> Symlinks support all of that, hard links don’t.

I was about to post a similar comment. However, I expect that a winget user is a more-than-average expert user, so they just need the proper guidance. I think the following  should work:
- if symlink works, just use that
- if not, try and fallback on hard link, possibly advising the user with a warning, explaining that winget works best in developer mode, where it can use symlinks, and now we are using a hardlink fallback
- if hard link creation fails, fail the installation or at least show an error, explain to the user the reason and advice to either use developer more or high privileges. One could also suggest to retry installation with elevated privileges as follows `winget install <package-id> --scope user`. Did I get it right that this should install as user without the permission problem? Of course it is not viable if the user does not have admin access, but at least the user knows her/his options

Kobi-Blade:
> > Hard links only work when the link and target are on the same NTFS volume, and they cannot link directories at all.
> > WinGet needs cross volume and directory level linking for portable packages, and it must track and remove links cleanly during upgrades.
> > Symlinks support all of that, hard links don’t.
> 
> I was about to post a similar comment. However, I expect that a winget user is a more-than-average expert user, so they just need the proper guidance. I think the following should work:
> 
> * if symlink works, just use that
> * if not, try and fallback on hard link, possibly advising the user with a warning, explaining that winget works best in developer mode, where it can use symlinks, and now we are using a hardlink fallback
> * if hard link creation fails, fail the installation or at least show an error, explain to the user the reason and advice to either use developer more or high privileges. One could also suggest to retry installation with elevated privileges as follows `winget install <package-id> --scope user`. Did I get it right that this should install as user without the permission problem? Of course it is not viable if the user does not have admin access, but at least the user knows her/his options

On that note, this PR was created because the average WinGet user is often not aware they should be running in developer mode or with elevation.

This PR is more prone to generating issue reports, and doesn't solve the root problem.

Honestly, we should be solving the symlink problem directly with more robust error checking and handling. For instance, we can trigger an elevated command for the symlink (which will prompt UAC), if the initial link creation fails due to permissions issues.

sysopmd:
ok, sorry, I thought that UAC was not a possibility for some reason. Then I understand it is preferrable to my suggestion about elevated '--scope' command.
However my understanding is that winget is designed to work for regular, non-elevated users without any special developer settings enabled.
So in this case, I think the original PR was a way to ease the user experience. However, we can also state that if a package require symlinks, then elevation or developer mode is required. I wonder which is the best way to communicate to the user that what they are seeing depends on the characteristic of the package they are installing and why.
Likely, elevation with UAC prompt provides a smoother experience. However, the user should be informed that they can also opt for developer mode.

> > > Hard links only work when the link and target are on the same NTFS volume, and they cannot link directories at all.
> > > WinGet needs cross volume and directory level linking for portable packages, and it must track and remove links cleanly during upgrades.
> > > Symlinks support all of that, hard links don’t.
> > 
> > 
> > I was about to post a similar comment. However, I expect that a winget user is a more-than-average expert user, so they just need the proper guidance. I think the following should work:
> > 
> > * if symlink works, just use that
> > * if not, try and fallback on hard link, possibly advising the user with a warning, explaining that winget works best in developer mode, where it can use symlinks, and now we are using a hardlink fallback
> > * if hard link creation fails, fail the installation or at least show an error, explain to the user the reason and advice to either use developer more or high privileges. One could also suggest to retry installation with elevated privileges as follows `winget install <package-id> --scope user`. Did I get it right that this should install as user without the permission problem? Of course it is not viable if the user does not have admin access, but at least the user knows her/his options
> 
> On that note, this PR was created because the average WinGet user is often not aware they should be running in developer mode or with elevation.
> 
> This PR is more prone to generating issue reports, and doesn't solve the root problem.
> 
> Honestly, we should be solving the symlink problem directly with more robust error checking and handling. For instance, we can trigger an elevated command for the symlink (which will prompt UAC), if the initial link creation fails due to permissions issues.

Kobi-Blade:
> ok, sorry, I thought that UAC was not a possibility for some reason. Then I understand it is preferrable to my suggestion about elevated '--scope' command. However my understanding is that winget is designed to work for regular, non-elevated users without any special developer settings enabled. So in this case, I think the original PR was a way to ease the user experience. However, we can also state that if a package require symlinks, then elevation or developer mode is required. I wonder which is the best way to communicate to the user that what they are seeing depends on the characteristic of the package they are installing and why. Likely, elevation with UAC prompt provides a smoother experience. However, the user should be informed that they can also opt for developer mode.

Keep in mind that the elevated command would only create the symlink, the package would still be installed in the user scope.

WinGet already informs the user that it lacks permissions in its current implementation, https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerSharedLib/Filesystem.cpp#L466

I do not disagree with anything you've said. I just don't believe this PR is the solution we should be looking for, and I mean no offense to the original poster.

sysopmd:
I see. I am not familiar with winget development. I hope I am not creating useless noise, please tell me if this is the case.
My intent here is to provide user-level feedback.

Regarding your last comment 
> WinGet already informs the user that it lacks permissions in its current implementation, https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerSharedLib/Filesystem.cpp#L466

I did not have such experience (see [my comment on 6058](https://github.com/microsoft/winget-cli/issues/6058#issuecomment-4046281300)): *no user-visible feedback* was provided in case the symlink fails.
But then the tools does not work. In [PortableInstaller.cpp](https://github.com/microsoft/winget-cli/blob/2f4f2cf16806debffc6c9d7e724b5c900f455a4f/src/AppInstallerCLICore/PortableInstaller.cpp#L164) I see that messages are logged, but they do not reach the user (at least they did not in my attempt - should I try a more recent version?). The path addition does not always help because sometimes link names are different than exe name.

Summarizing I think that requesting elevation is a good fix, but I would add a message explaining why elevation is required.
And in case the user cancels and the link cannot be created, I think there should be proper messaging indicating an installation failure or at least that the package might misbehave. Does it make any sense?

rich-purnell:
1. Thank you for your attention on this matter.

2. As a user, after installing Godot using `winget install -e --id GodotEngine.GodotEngine`, I received the following output:

```
Added command line alias: "godot"
Added command line alias: "godot_console"
Successfully installed
Note: Note when running WinGet to Install This Package *Without Admin Privileges*:
WinGet *Can Not* Create Command Line Alias(es) unless Admin Privileges are provided, as without Admin Privileges Winget *Can Not* Create Symbolic Links for those Command Lines Aliases.
```

As you can see, it reports both success and failure. Prior to this installation, I had always assumed that the aliases were successfully created, because your "if alias creation fails, add the app's installation directory to PATH" strategy effectively deceived me into thinking that. It was only when I noticed that Godot's program names did not match the aliases that I began to investigate the issue. At that point, I discovered that "symlink creation" had never worked on my system.

3. When I understood the root cause, my first thought was "why not pop up a UAC elevation prompt?" However, after some consideration, I realized the reason. Windows inherently treats symlink creation as a potentially dangerous operation, and since WinGet's portable package manifests are community-contributed, arbitrarily providing administrator privileges would carry significant security risks.

4. Without privilege elevation, the only viable option for implementing aliases is hard links. As I noted in my PR description, hard links do not solve all problems, but they reduce the impact surface. Hard links are not suitable in two scenarios: a. `%LOCALAPPDATA%\Microsoft\WinGet\Links` and `%LOCALAPPDATA%\Microsoft\WinGet\Packages` reside on different NTFS volumes; b. an alias is needed for a directory. I have not encountered any use cases that require these scenarios. I understand that the installation location can be configured via options, but it would also be possible to create the corresponding `Links` directory on the relevant volume and add multiple `Links` directories to PATH. I would even argue that WinGet should completely replace symlinks with hard links, as they are more secure.

5. The worst-case outcome of attempting to fall back to a hard link when symlink creation fails is that nothing changes compared to the current strategy. However, if it succeeds, it resolves part of the alias creation failure problem.

6. The current PR is crude and unacceptable; for now, just treat it as a proof of concept. I am working on refining the code.

7. Until I saw that UAC is considered acceptable. Can you confirm this? Because that would render much of the discussion around this issue moot.

Trenly:
> Until I saw that UAC is considered acceptable. Can you confirm this? Because that would render much of the discussion around this issue moot.

One core tenet of WinGet is that it must be able to install applications unattended - if there were to be a UAC prompt, it should come right at the beginning while a user is still at the machine; However, with installer selection being potentially deferred, especially in the case of `winget import` I don’t know that it would be possible to make the determination of if elevation will be needed until too late in the install process

florelis:
What are all the limitations with hard links? Just being on the same NTFS volume?
We could detect those conditions or the failure to create the hard link and just continue with the current behavior in that case.

So:
Dev mode or Admin -> Symlink
On same NTFS volume -> Hardlink
Not dev, not admin, on different NTFS volume -> Add install location to path

I think the last case would be extremely rare. The default locations for the executables and the links are both in `%LocalAppData%\Microsoft\WinGet\`. I don't see much reason for why someone would put that in a separate volumen.
If installing machine-wide to `%ProgramFiles%\`, then they have admin permissions and can make symlinks.
For hardlinks to not work, the user would need to have a custom install directory that lives in a separate volume. I don't think that would happen much for non devs.

JohnMcPMS:
Symlinks have an additional property; the actual executable is located in its directory that may contain dependencies like DLLs.  So to the running code, it appears to be at its resolved location, not the symlink.  A hardlink would appear to be running at the hardlink location (it is just another "normal" file that happens to have the same data stream as another file).

One potential option for the issue of creating aliases would be to still add the location to PATH while creating the hardlinks in the target location (rather than the Links directory).

sredna:
> Symlinks have an additional property; the actual executable is located in its directory that may contain dependencies like DLLs.

This abstraction is not perfect, there are still cases where being a symlink is not the same as the real .exe.

This issue just further proves that using small forwarder/shim exe files is the best solution to solve the %path% question (IMHO).

JohnMcPMS:
> This issue just further proves that using small forwarder/shim exe files is the best solution to solve the %path% question (IMHO).

Those are problematic from a signing perspective, not to mention all of the ways in which perfectly forwarding everything from one process to another is difficult/fragile.

We should investigate the breadth of support for [App Paths](https://learn.microsoft.com/en-us/windows/win32/shell/app-registration) amongst users who are having issues with the current setup flow.  If they are mostly using shells that integrate with this registration mechanism, it would probably be a better solution than continuing down the current path.

Overall, it would probably be good to have many of these options (and CLI params), choose the best defaults, and allow for `winget repair` to target portables while updating the mechanism by which they are linked into discovery.

sredna:
> Those are problematic from a signing perspective

If you don't care about the icon, it would just be the same WinGet provided shim that is signed (and hardlinked for each new app).

> not to mention all of the ways in which perfectly forwarding everything from one process to another is difficult/fragile. 

The only thing you have to forward is the command line arguments, everything else (environment and current directory) is automatic.

florelis:
> If you don't care about the icon, it would just be the same WinGet provided shim that is signed (and hardlinked for each new app).

That's problematic at least for cases where you want to run as admin.

If there is an UAC it should be "do you want to run Cool App, published by Contoso (verified by signature)?"

If we create a shim that is specific to this app, we could not sign it, so the UAC would be "do you want to run Cool App, published by Contoso (unverified)?" Not having the publisher verified should be a red flag for people running this.

If we use a common shim that we sign beforehand, the UAC would always say "do you want to run WinGet Shim, published by Microsoft (verified by signature)?" This could mislead people into thinking the app is trustworthy and from MS, not just a random app from who-knows-who.

sredna:
> That's problematic at least for cases where you want to run as admin.

The shim would never request UAC elevation in its manifest. If someone enumerates %path% searching for a specific filename and then on purpose uses "runas" with ShellExecute to elevate then yes, you would get a elevation request on the shim itself. But yes, this is a case that is problematic.

rich-purnell:
> The uninstall flow may need to account for hard links with this change. Tests should be added to validate hardlink creation, path resolution, and removal.
> 
> What happens in the upgrade scenario where a hardlink exists but the user now has symlink permissions - does it stay a hardlink, or should it be replaced with a symlink?

Tested with and without admin privileges:

- Installing, updating, and uninstalling Portable packages via `wingetdev`.
- Installing an older version of a Portable package via `winget`, then updating it via `wingetdev`.

rich-purnell:
> Symlinks have an additional property; the actual executable is located in its directory that may contain dependencies like DLLs. So to the running code, it appears to be at its resolved location, not the symlink. A hardlink would appear to be running at the hardlink location (it is just another "normal" file that happens to have the same data stream as another file).
> 
> One potential option for the issue of creating aliases would be to still add the location to PATH while creating the hardlinks in the target location (rather than the Links directory).

Your suggestion is well-considered, but I plan to proceed with the original implementation for the following reasons:  

The ArchiveBinariesDependOnPath option already exists to allow users to choose whether to add the installation directory to PATH.  
Consolidating aliases under the Links directory is preferable to adding multiple paths to PATH.  

However, your proposal could improve the "Add install directory to PATH" process: if an alias differs from the original executable name, create a hard link in the program's installation directory and add that directory to PATH. This refinement can be considered for future implementation.

rich-purnell:
Additionally, I've noticed that many test cases lack actual "testing" procedures. For instance, the `InstallFlow_Portable` case does not execute the Install process at all—it merely creates a text file and exits. Numerous `OverrideXXX` functions in the `WorkflowCommon.cpp` file also appear to be unimplemented. I am uncertain whether this is intentional design or if there is an issue with my execution.

Trenly:
I have raised a separate PR based on my interpretation of all these comments. I do want to give @rich-purnell thanks and credit for the ideas, and clarify that I mean no offense by raising the changes as a separate PR.

> One potential option for the issue of creating aliases would be to still add the location to PATH while creating the hardlinks in the target location (rather than the Links directory).

This is the approach I took, though I didn't couple it to the symlink flow.

> So:
> Dev mode or Admin -> Symlink
> On same NTFS volume -> Hardlink
> Not dev, not admin, on different NTFS volume -> Add install location to path

While this made sense logically to me, it felt too conditional. I've suggested
* Dev mode or Admin -> Symlink + Hardlink (if aliased)
* On same NTFS volume -> Hardlink (if aliased) & Install location on path (preserve current pathing behavior)
* Not dev, not admin, on different NTFS volume -> Hardlink (if aliased) & Install location on path (preserve current pathing behavior)

> The ArchiveBinariesDependOnPath option already exists to allow users to choose whether to add the installation directory to PATH.
> Consolidating aliases under the Links directory is preferable to adding multiple paths to PATH.
> 
> However, your proposal could improve the "Add install directory to PATH" process: if an alias differs from the original executable name, create a hard link in the program's installation directory and add that directory to PATH. This refinement can be considered for future implementation.

WinGet will add the install directory to path in a at least these two cases:
* Explicitly required by `ArchiveBinariesDependOnPath`
* Non-Elevated user and Developer Mode disabled

So, by always creating the hardlink, even if the symlink creation fails and WinGet adds the install directory to path (which is the second case), the alias would always exist

rich-purnell:
@Trenly I’m not sure that "both" is the most concise approach. If you prefer hardlinks so much, why not use them exclusively and drop symlinks entirely?

Also, it appears that your PR may not have been sufficiently tested. I’ll comment on the issues I’ve found under your PR.

ranm-msft (commented):
Leaving the design debate aside; these are three things in the uninstall/PATH paths as written.

**1. `RemoveFile` deletes an unverified regular file.** The new `else if (exists(filePath) && is_regular_file(...))` branch removes whatever regular file sits at the recorded alias path, with no check that it is actually a hard link to this package's target. Symlinks get `VerifySymlink` before removal, and `VerifyExpectedState` treats a plain regular file at that path as valid rather than as a replacement. If symlink creation previously failed and WinGet fell back to PATH, no link was ever created there, yet uninstall will now delete anything occupying that name. Comparing the file index against the target (`GetFileInformationByHandle`) would restore parity with the symlink path.

**2. Uninstall PATH cleanup moved from persisted to in-run state.** `Uninstall()` now removes the Links directory from PATH only when `m_hasRemovedSymlink` was set during this invocation. If an alias created at install time was later deleted manually, that flag stays false and the now-unused Links directory is left on PATH; the previous persisted-state check cleaned it up.

**3. `InstallDirectoryAddedToPath` is no longer cleared.** The assignment in `RemoveFromPathVariable` was dropped, but the field is still committed as `true` elsewhere, so the ARP entry can keep claiming the install directory is on PATH after it has been replaced by a link. Either keep it in sync or remove and migrate it here.

Minor: `CreateFileHardLink` swallows the exception with no log, so failures are undiagnosable from the logs.

rich-purnell:
> Leaving the design debate aside; these are three things in the uninstall/PATH paths as written.
> 
> **1. `RemoveFile` deletes an unverified regular file.** The new `else if (exists(filePath) && is_regular_file(...))` branch removes whatever regular file sits at the recorded alias path, with no check that it is actually a hard link to this package's target. Symlinks get `VerifySymlink` before removal, and `VerifyExpectedState` treats a plain regular file at that path as valid rather than as a replacement. If symlink creation previously failed and WinGet fell back to PATH, no link was ever created there, yet uninstall will now delete anything occupying that name. Comparing the file index against the target (`GetFileInformationByHandle`) would restore parity with the symlink path.
> 
> **2. Uninstall PATH cleanup moved from persisted to in-run state.** `Uninstall()` now removes the Links directory from PATH only when `m_hasRemovedSymlink` was set during this invocation. If an alias created at install time was later deleted manually, that flag stays false and the now-unused Links directory is left on PATH; the previous persisted-state check cleaned it up.
> 
> **3. `InstallDirectoryAddedToPath` is no longer cleared.** The assignment in `RemoveFromPathVariable` was dropped, but the field is still committed as `true` elsewhere, so the ARP entry can keep claiming the install directory is on PATH after it has been replaced by a link. Either keep it in sync or remove and migrate it here.
> 
> Minor: `CreateFileHardLink` swallows the exception with no log, so failures are undiagnosable from the logs.

@ranm-msft Are you pasting the conclusions from an AI scan of the PR?

Issue #6221, opened 2026-05-12
Clarify manifest entries usage/purpose


sredna:
### Brief description of your issue

Some entries are a bit unclear (IMHO) and some [example value annotations](https://json-schema.org/understanding-json-schema/reference/annotations) in the schema would be nice as well.

## ProductCode

> ProductCode could be used for correlation of packages across sources
 
 is simply not descriptive enough.

I tried to add some feedback in an [existing issue](https://github.com/microsoft/winget-cli/issues/921) about this but got no response.
- A MSI product code? Then the documentation should say so and the validator should ensure that this is a GUID.
- The name of the ARP uninstall subkey? If this key is used to determine if the package is already is installed then that should be called out in the value description. Calling this a ProductCode is perhaps a misnomer because that is MSI specific terminology.

installer.md currently links to a MSDN page that starts with "The product code is a GUID ..." but the overall usage in the community repository contains free form text entries as well (and some GUIDs are missing the {} brackets).

If this is always a GUID, the schema should have a `pattern:` for GUID validation and include `"examples": ["{MSI-GUID}"]`, otherwise it should include `"examples": ["{MSI-GUID}", "ContosoApp", "Fabrikam.FantastikApp"]`

## UpgradeCode

UpgradeCode is definitely a GUID but the schema just uses `"$ref": "#/definitions/ProductCode"` which has the same poor description and no GUID validation.

The schema should be updated to include `"examples": ["{MSI-GUID}"]` and `"$comment": "https://learn.microsoft.com/en-us/windows/win32/msi/upgradecode"`

## InstallerSwitches\Silent

The description is problematic.

 > Silent is the value that should be passed to the installer when user chooses a **silent or quiet** install 

The word "quiet" is not used anywhere else in the entire schema, is this just a synonym for silent or is it a different type of install mode? If it's just a synonym, it should be removed or replaced with "silent/quiet" to remove the word "or".

## Dependencies\WindowsFeatures

Should include `"examples": ["ntvdm", "DirectPlay", "Microsoft-Windows-Subsystem-Linux"]`

## Capabilities

- `Capabilities` should include `"examples": ["contacts", "internetClient", "globalMediaControl"]`
- `RestrictedCapabilities` should include `"examples": ["deviceUnlock", "runFullTrust", "broadFileSystemAccess"]`

## RequireExplicitUpgrade

The schema and installer.md descriptions vary too much, the one in installer.md is much better.

## ElevationRequirement

[...]

PR #5997, opened 2026-01-21
Add DSC resource list to manifest

JohnMcPMS:
Implements #3523 

## Change
Creates manifest version 1.28.0, adding the DSC resource list to the installer, similar to described in the issue.  Is a little deeper and more extensible than the suggested version.  As an example, winget's own full list would look like:
```YAML
DesiredStateConfiguration:
  PowerShell:
    - RepositoryUrl: https://www.powershellgallery.com/api/v2
      ModuleName: Microsoft.WinGet.DSC
      Resources:
        - Name: WinGetUserSettings
        - Name: WinGetAdminSettings
        - Name: WinGetSource
        - Name: WinGetPackageManager
        - Name: WinGetPackage
    - RepositoryUrl: https://mcr.microsoft.com/
      ModuleName: Microsoft.WinGet.DSC
      Resources:
        - Name: WinGetUserSettings
        - Name: WinGetAdminSettings
        - Name: WinGetSource
        - Name: WinGetPackageManager
        - Name: WinGetPackage
  DSCv3:
    Resources:
      - Type: Microsoft.WinGet/AdminSettings
      - Type: Microsoft.WinGet/Package
      - Type: Microsoft.WinGet/Source
      - Type: Microsoft.WinGet/UserSettingsFile
```
But `PowerShell` will be blocked from winget-pkgs due to the inherently independent nature of the modules and packages.  With a future ownership model we could likely allow them to be listed.

## Validation
Standard unit tests added covering new manifest version features.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/5997)

PR #6277, opened 2026-06-10
Clarify some manifest values and add examples

sredna:
## 📖 Description
 - Tries to clarify the description of some of the manifest values.
 - Adds value examples.
 - Adds the UpgradeCode definition.

#### Notes:
- `UpgradeCode` is restricted only to GUIDs because this is a MSI feature. I have verified that all existing manifests in the community repository only uses GUIDs here.
 - `DefaultInstallLocation` is restricted to strings starting with `%` or `\`. The thinking is that the only sane values are `%ProgramFiles%\MyApp`, `%LOCALAPPDATA%\Programs\MyApp` and `\AppInRootOfDrive`.

## 🔗 References
Fixes some of the issues reported in #6221 

## 🔍 Validation

None/YOLO

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6277)

### DIFF
--- a/schemas/JSON/manifests/latest/manifest.installer.latest.json
+++ b/schemas/JSON/manifests/latest/manifest.installer.latest.json
@@ -546,9 +546,8 @@
       "properties": {
         "DefaultInstallLocation": {
           "type": [ "string", "null" ],
-          "minLength": 2,
+          "minLength": 1,
           "maxLength": 2048,
-          "pattern": "^[%\\]",
           "examples": ["%ProgramFiles%\\ContosoApp"],
           "description": "Represents the default installed package location. Used for deeper installation detection."
         },
--- a/schemas/JSON/manifests/latest/manifest.installer.latest.json
+++ b/schemas/JSON/manifests/latest/manifest.installer.latest.json
@@ -313,7 +313,7 @@
           },
           "maxItems": 16,
           "uniqueItems": true,
-          "examples": ["ntvdm", "DirectPlay", "Microsoft-Windows-Subsystem-Linux"],
+          "examples": ["DirectPlay", "Microsoft-Windows-Subsystem-Linux"],
           "description": "List of Windows feature dependencies"
         },
         "WindowsLibraries": {
@@ -368,7 +368,7 @@
       "type": [ "string", "null" ],
       "minLength": 1,
       "maxLength": 255,
-      "examples": ["{MSI-GUID}", "ContosoApp", "Fabrikam.FantastikApp"],
+      "examples": ["{MSI-GUID}", "ContosoApp"],
       "description": "The name of the ARP registry key, could be used for correlation of packages across sources"
     },
     "UpgradeCode": {
--- a/schemas/JSON/manifests/latest/manifest.installer.latest.json
+++ b/schemas/JSON/manifests/latest/manifest.installer.latest.json
@@ -152,7 +152,7 @@
           "type": [ "string", "null" ],
           "minLength": 1,
           "maxLength": 512,
-          "description": "Silent is the value that should be passed to the installer when user chooses a silent or quiet install"
+          "description": "Silent is the value that should be passed to the installer when user chooses a silent install"
         },
         "SilentWithProgress": {
           "type": [ "string", "null" ],
@@ -313,6 +313,7 @@
           },
           "maxItems": 16,
           "uniqueItems": true,
+          "examples": ["ntvdm", "DirectPlay", "Microsoft-Windows-Subsystem-Linux"],
           "description": "List of Windows feature dependencies"
         },
         "WindowsLibraries": {
@@ -367,7 +368,17 @@
       "type": [ "string", "null" ],
       "minLength": 1,
       "maxLength": 255,
-      "description": "ProductCode could be used for correlation of packages across sources"
+      "examples": ["{MSI-GUID}", "ContosoApp", "Fabrikam.FantastikApp"],
+      "description": "The name of the ARP registry key, could be used for correlation of packages across sources"
+    },
+    "UpgradeCode": {
+      "type": [ "string", "null" ],
+      "pattern": "^{[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}}$",
+      "minLength": 38,
+      "maxLength": 38,
+      "examples": ["{MSI-GUID}"],
+      "description": "Represents a related set of products, could be used for correlation of packages across sources",
+      "$comment": "https://learn.microsoft.com/en-us/windows/win32/msi/using-an-upgradecode"
     },
     "Capabilities": {
       "type": [ "array", "null" ],
@@ -378,6 +389,7 @@
       },
       "maxItems": 1000,
       "uniqueItems": true,
+      "examples": ["contacts", "internetClient", "globalMediaControl"],
       "description": "List of appx or msix installer capabilities"
     },
     "RestrictedCapabilities": {
@@ -389,6 +401,7 @@
       },
       "maxItems": 1000,
       "uniqueItems": true,
+      "examples": ["deviceUnlock", "runFullTrust", "broadFileSystemAccess"],
       "description": "List of appx or msix installer restricted capabilities"
     },
     "Market": {
@@ -501,7 +514,7 @@
           "$ref": "#/definitions/ProductCode"
         },
         "UpgradeCode": {
-          "$ref": "#/definitions/ProductCode"
+          "$ref": "#/definitions/UpgradeCode"
         },
         "InstallerType": {
           "$ref": "#/definitions/InstallerType"
@@ -533,8 +546,10 @@
       "properties": {
         "DefaultInstallLocation": {
           "type": [ "string", "null" ],
-          "minLength": 1,
+          "minLength": 2,
           "maxLength": 2048,
+          "pattern": "^[%\\]",
+          "examples": ["%ProgramFiles%\\ContosoApp"],
           "description": "Represents the default installed package location. Used for deeper installation detection."
         },
         "Files": {
--- a/schemas/JSON/manifests/latest/manifest.singleton.latest.json
+++ b/schemas/JSON/manifests/latest/manifest.singleton.latest.json
@@ -647,9 +647,8 @@
       "properties": {
         "DefaultInstallLocation": {
           "type": [ "string", "null" ],
-          "minLength": 2,
+          "minLength": 1,
           "maxLength": 2048,
-          "pattern": "^[%\\]",
           "examples": ["%ProgramFiles%\\ContosoApp"],
           "description": "Represents the default installed package location. Used for deeper installation detection."
         },
--- a/schemas/JSON/manifests/latest/manifest.singleton.latest.json
+++ b/schemas/JSON/manifests/latest/manifest.singleton.latest.json
@@ -415,7 +415,7 @@
           },
           "maxItems": 16,
           "uniqueItems": true,
-          "examples": ["ntvdm", "DirectPlay", "Microsoft-Windows-Subsystem-Linux"],
+          "examples": ["DirectPlay", "Microsoft-Windows-Subsystem-Linux"],
           "description": "List of Windows feature dependencies"
         },
         "WindowsLibraries": {
@@ -469,7 +469,7 @@
       "type": [ "string", "null" ],
       "minLength": 1,
       "maxLength": 255,
-      "examples": ["{MSI-GUID}", "ContosoApp", "Fabrikam.FantastikApp"],
+      "examples": ["{MSI-GUID}", "ContosoApp"],
       "description": "The name of the ARP registry key, could be used for correlation of packages across sources"
     },
     "UpgradeCode": {
--- a/schemas/JSON/manifests/latest/manifest.singleton.latest.json
+++ b/schemas/JSON/manifests/latest/manifest.singleton.latest.json
@@ -254,7 +254,7 @@
           "type": [ "string", "null" ],
           "minLength": 1,
           "maxLength": 512,
-          "description": "Silent is the value that should be passed to the installer when user chooses a silent or quiet install"
+          "description": "Silent is the value that should be passed to the installer when user chooses a silent install"
         },
         "SilentWithProgress": {
           "type": [ "string", "null" ],
@@ -415,6 +415,7 @@
           },
           "maxItems": 16,
           "uniqueItems": true,
+          "examples": ["ntvdm", "DirectPlay", "Microsoft-Windows-Subsystem-Linux"],
           "description": "List of Windows feature dependencies"
         },
         "WindowsLibraries": {
@@ -468,7 +469,17 @@
       "type": [ "string", "null" ],
       "minLength": 1,
       "maxLength": 255,
-      "description": "ProductCode could be used for correlation of packages across sources"
+      "examples": ["{MSI-GUID}", "ContosoApp", "Fabrikam.FantastikApp"],
+      "description": "The name of the ARP registry key, could be used for correlation of packages across sources"
+    },
+    "UpgradeCode": {
+      "type": [ "string", "null" ],
+      "pattern": "^{[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}}$",
+      "minLength": 38,
+      "maxLength": 38,
+      "examples": ["{MSI-GUID}"],
+      "description": "Represents a related set of products, could be used for correlation of packages across sources",
+      "$comment": "https://learn.microsoft.com/en-us/windows/win32/msi/using-an-upgradecode"
     },
     "Capabilities": {
       "type": [ "array", "null" ],
@@ -479,6 +490,7 @@
       },
       "maxItems": 1000,
       "uniqueItems": true,
+      "examples": ["contacts", "internetClient", "globalMediaControl"],
       "description": "List of appx or msix installer capabilities"
     },
     "RestrictedCapabilities": {
@@ -490,6 +502,7 @@
       },
       "maxItems": 1000,
       "uniqueItems": true,
+      "examples": ["deviceUnlock", "runFullTrust", "broadFileSystemAccess"],
       "description": "List of appx or msix installer restricted capabilities"
     },
     "Market": {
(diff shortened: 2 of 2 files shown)


JohnMcPMS (commented):
The changes seem reasonable to me, but I don' actually touch the schema much so I would like someone else to take a look too.

The flow for semantic changes to the schema as part of a new minor version also requires quite a bit more setup.  See #5997 and/or #5011.

I think I will also create some additional documentation to prevent the need for you to parse through those if you want to wait for that.

JohnMcPMS:
#6280 ; you could either grab the script from that PR or wait for it to be merged.  But #6279 is updating the minor version that the script picks up so probably want that too.

schemas/JSON/manifests/latest/manifest.installer.latest.json
```diff
@@ -367,7 +368,17 @@
       "type": [ "string", "null" ],
       "minLength": 1,
       "maxLength": 255,
-      "description": "ProductCode could be used for correlation of packages across sources"
+      "examples": ["{MSI-GUID}", "ContosoApp"],
+      "description": "The name of the ARP registry key, could be used for correlation of packages across sources"
+    },
+    "UpgradeCode": {
+      "type": [ "string", "null" ],
+      "pattern": "^{[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}}$",
```
denelon, line 376: I could be mistaken, but I believe I've seen instances of a product code or an upgrade code value not being a GUID. In general, MSI product codes and upgrade codes do follow that convention, but I'm concerned this might break some things.

sredna: I searched the community repository and found none (for upgrade code, clearly not true for product code but that is not what we are talking about here). Upgrade codes are strictly a MSI concept and will always be a GUID.

Or to put it another way, how would a non-GUID work? Where in the registry would it be stored and which Win32/MSI API provides enumeration support?

schemas/JSON/manifests/latest/manifest.installer.latest.json
```diff
@@ -533,8 +546,10 @@
       "properties": {
         "DefaultInstallLocation": {
           "type": [ "string", "null" ],
-          "minLength": 1,
+          "minLength": 2,
           "maxLength": 2048,
+          "pattern": "^[%\\]",
```
pl4nty: A lot of apps [hardcode a `C:\` prefix](https://github.com/search?q=repo%3Amicrosoft%2Fwinget-pkgs+%22DefaultInstallLocation%3A+C%3A%5C%5C%22&type=code). Most of these are parsed directly from the installer by komac

sredna: But it does not make sense to allow that. I don't think this value is actually used yet but when it is, there are pretty much only 3 places that make sense:

- `%ProgramFiles%\MyApp` (Or in `%ProgramFiles(x86)%`)
- `%LocalAppData%\Programs\MyApp`
- `\MyApp` (In the root of any drive? The same drive as `%windir%`?)

pl4nty: Unfortunately a lot of apps do not install to paths that make sense. I don't think we can block apps from WinGet just because they install to an unusual path

sredna: It's not about blocking anyone but manifests should match the spec. Some manifests might need to change `c:\MyApp` to `\MyApp` etc. to be compliant if they want to update to the latest manifest version.

pl4nty: I think that could be misleading though. These apps don't install to `\MyApp` or `%SystemDrive%\MyApp`, they specifically hardcode `C:\MyApp`

Trenly: The fact that "some manifests may need to change" would indicate to me this is a breaking change as manifest versions should be forward compatible as mich as possible.

sredna: > I think that could be misleading though. These apps don't install to `\MyApp` or `%SystemDrive%\MyApp`, they specifically hardcode `C:\MyApp`

C:\ might not even exist on all systems.

Somebody from Microsoft should chime in and say the exact allowed format of this value (which is the point of this PR in the first place).

schemas/JSON/manifests/latest/manifest.installer.latest.json
denelon: I must have been mistaken. 👍

schemas/JSON/manifests/latest/manifest.installer.latest.json
denelon: Do we have any concrete examples of packages in the repository that exhibit this behavior. I suspect we'll need to use a few real examples to make the right decision here.

sredna: > Do we have any concrete examples of packages in the repository that exhibit this behavior.

Examples of drive c: root folders (I think this is all the packages that do this):

- [KOG.Elsworld.es.installer.yaml](https://github.com/microsoft/winget-pkgs/blob/6d1991a4555d37eb397d28517444054324651358/manifests/k/KOG/Elsworld/es/BR.L.221103.0.1/KOG.Elsworld.es.installer.yaml): `C:\KOGGames\Elsword64`
- [Coq.CoqPlatform.installer.yaml](https://github.com/microsoft/winget-pkgs/blob/6d1991a4555d37eb397d28517444054324651358/manifests/c/Coq/CoqPlatform/2025.08.3/Coq.CoqPlatform.installer.yaml): `C:\Rocq-Platform~9.0~2025.08`
- [CharruaSoft.PrintSCP.installer.yaml](https://github.com/microsoft/winget-pkgs/blob/6d1991a4555d37eb397d28517444054324651358/manifests/c/CharruaSoft/PrintSCP/12.0/CharruaSoft.PrintSCP.installer.yaml#L23): `C:\CharruaSoft\PrintSCP`
- [martinrotter.RSSGuard5.installer.yaml](https://github.com/microsoft/winget-pkgs/blob/6d1991a4555d37eb397d28517444054324651358/manifests/m/martinrotter/RSSGuard5/5.2.1/martinrotter.RSSGuard5.installer.yaml): `C:\RSS Guard 5`

Examples of broken entries:
- [Ethereum.geth.installer.yaml](https://github.com/microsoft/winget-pkgs/blob/6d1991a4555d37eb397d28517444054324651358/manifests/e/Ethereum/geth/1.16.3/Ethereum.geth.installer.yaml): `InstallDir` (This looks like a bug in some sort of automated script. In NSIS installers, the default install directory is specified like `InstallDir "$ProgramFiles\MyApp"`)
- [Humanity.DJV2.installer.yaml](https://github.com/microsoft/winget-pkgs/blob/6d1991a4555d37eb397d28517444054324651358/manifests/h/Humanity/DJV2/3.3.4/Humanity.DJV2.installer.yaml): `Documents\DJV 3.3.4` (I assume they mean My Documents but that should be specified as `%userprofile%\Documents`)
- GAM-Team.gam.installer.yaml used to specify `./GAM7` (relative to what?) in older versions but has been corrected to `%SystemDrive%\GAM7` in newer versions so this is no longer a problem nor a breaking change.

denelon: @JohnMcPMS what are your thoughts here?

florelis: I don't think we should add a new restriction on the possible values for a field, even if they may be "wrong", because that would break backwards compatibility. We could clean up the community repo, but we can't be sure it's not used like that in some private third party source.

Also, my understanding is that winget doesn't use this to decide where to install a package to, and instead it just tells us where a package installs by default. I think in those cases there is a meaningful difference between having `C:\Program Files\MyApp\` and `%ProgramFiles%\MyApp\`: If `%ProgramFiles%` is configured in the machine as something different, but the installer uses `C:\Program Files\` instead of a variable, then it will try to install there regardless of what your `%ProgramFiles%` is or whether you even have a `C:\` drive.

Same for having a relative path. If an installer defaults to `.\MyApp\` relative to where it is run from, then we can't say it's going to be `%SystemDrive%\MyApp\` or anything like that. If the app is run from winget then we know it will be relative to winget's cwd, but I think that could change, for example if we're running inproc.

sredna: > ... because that would break backwards compatibility. We could clean up the community repo, but we can't be sure it's not used like that in some private third party source.

How would it break backwards compatibility? This only applies to new manifest versions, not anything existing in any repo.

That being said, I can remove the restriction on the format of this field if Microsoft decides that it is just free-form for display to the user and not something WinGet is ever going to parse.

> Same for having a relative path. If an installer defaults to `.\MyApp\` relative to where it is run from, then we can't say it's going to be `%SystemDrive%\MyApp\` or anything like that.

Which is why my stance is that while a root relative path (`\MyApp`) is at least acceptable, a CWD relative path makes no sense. I can do `cd %temp% & winget show MyApp` which would make it relative to %temp% which is not sane.

Trenly: > How would it break backwards compatibility? This only applies to new manifest versions, not anything existing in any repo.

Because of semantic versioning. It's expected that a minor version bump does not contain any breaking changes. Meaning, a manifest version 1.x should continue to work and be valid even if the only thing that gets bumped is the version number and schema header.  Because the schemas determine how WinGet processes the values, if a 1.12 manifest specified `C:\Package-Name` updating it to v1.30 would cause it to become invalid, which goes against "no breaking changes in minor versions".

sredna: I removed that restriction. 

I'll try to remember to nag about better specification for that when 2.0 is around the corner.

florelis: > I'll try to remember to nag about better specification for that when 2.0 is around the corner.

You can mention it in #2128

schemas/JSON/manifests/latest/manifest.installer.latest.json
florelis: I just searched the community repo and found a few that don't do this.

I saw:
* GUID without braces
  * [crloewen.lazywslc](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/c/crloewen/lazywslc/0.1.4/crloewen.lazywslc.installer.yaml#L19)
  * [Datahunter.Datahunter](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/d/Datahunter/Datahunter/6.0.3/Datahunter.Datahunter.installer.yaml#L11)
  * [Datahunter.DatahunterTNT](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/d/Datahunter/DatahunterTNT/2.8.3/Datahunter.DatahunterTNT.installer.yaml#L13)
  * [Mickem.NSClient](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/m/Mickem/NSClient/0.14.1/Mickem.NSClient.installer.yaml#L13)
  * [OO-Software.Defrag.Professional](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/Defrag/Professional/29.1.11203/OO-Software.Defrag.Professional.installer.yaml#L30)
  * [OO-Software.Defrag.Server](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/Defrag/Server/29.1.11203/OO-Software.Defrag.Server.installer.yaml#L30)
  * [OO-Software.SafeErase.Professional](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/SafeErase/Professional/19.3.1052/OO-Software.SafeErase.Professional.installer.yaml#L30)
  * [OO-Software.SafeErase.Server](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/SafeErase/Server/19.3.1052/OO-Software.SafeErase.Server.installer.yaml#L37)
* GUID, followed by `_is1`
  * [v2RayTun.Windows](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/v/v2RayTun/Windows/3.8.12/v2RayTun.Windows.installer.yaml#L19)
* `{*}`
  * [Suwayomi.Suwayomi-Server](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/s/Suwayomi/Suwayomi-Server/1.0.0/Suwayomi.Suwayomi-Server.installer.yaml#L19)

This change would break those packages. I would like to manually check if they actually use that UpgradeCode, or if it is an authoring error. If it is an error, maybe we could update the manifests and take this change.

sredna: > GUID without braces 

I did not look at all of these but the first one is newer than my PR so no wonder I did not see it. It does say "created using wingetcreate 1.12.8.0" so, does wingetcreate need to enforce braces?

> * GUID, followed by `_is1`

That one was meant to be ProductCode because Inno Setup installers append that to the name of their registry key. Inno Setup is not MSI based so having a UpgradeCode makes no sense.

> {*}

A bug in [Komac](https://github.com/russellbanks/Komac)? This is also an older version of Suwayomi.Suwayomi-Server (v1, they are at v2.3 now)

florelis: Looking at the code in `ARPHelper.cpp`, these need to be GUIDs with braces. When we read them from the registry, we parse the value as a GUID and create the string with the specific format. So, if the manifest has anything but a GUID with braces, it won't do anything.

So, this restriction is good and we should clean up these manifests. But I wonder if there may be somebody with a private source that uses bad UpgradeCodes...

(Edit: I had to look at the code to figure it out, but apparently it was me who wrote it lol)

florelis (approved):
Looks good to me, but I would like @denelon or @johnmcpms to weigh in on whether to take the new pattern for UpgradeCode. It does limit to valid codes, but there may be someone out there with invalid codes in their manifests.

denelon:
If we enable this, are there any existing manifests that would fail if something like wingetcreate were to be used to generate a new manifest and essentially just copy metadata forward for the "new" version?

florelis:
> If we enable this, are there any existing manifests that would fail if something like wingetcreate were to be used to generate a new manifest and essentially just copy metadata forward for the "new" version?

Yes:
[crloewen.lazywslc](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/c/crloewen/lazywslc/0.1.4/crloewen.lazywslc.installer.yaml#L19)
[Datahunter.Datahunter](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/d/Datahunter/Datahunter/6.0.3/Datahunter.Datahunter.installer.yaml#L11)
[Datahunter.DatahunterTNT](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/d/Datahunter/DatahunterTNT/2.8.3/Datahunter.DatahunterTNT.installer.yaml#L13)
[Mickem.NSClient](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/m/Mickem/NSClient/0.14.1/Mickem.NSClient.installer.yaml#L13)
[OO-Software.Defrag.Professional](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/Defrag/Professional/29.1.11203/OO-Software.Defrag.Professional.installer.yaml#L30)
[OO-Software.Defrag.Server](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/Defrag/Server/29.1.11203/OO-Software.Defrag.Server.installer.yaml#L30)
[OO-Software.SafeErase.Professional](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/SafeErase/Professional/19.3.1052/OO-Software.SafeErase.Professional.installer.yaml#L30)
[OO-Software.SafeErase.Server](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/o/OO-Software/SafeErase/Server/19.3.1052/OO-Software.SafeErase.Server.installer.yaml#L37)
[v2RayTun.Windows](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/v/v2RayTun/Windows/3.8.12/v2RayTun.Windows.installer.yaml#L19)
[Suwayomi.Suwayomi-Server](https://github.com/microsoft/winget-pkgs/blob/a2e87c1afdb6951b2996cf71a04a90614fab3da8/manifests/s/Suwayomi/Suwayomi-Server/1.0.0/Suwayomi.Suwayomi-Server.installer.yaml#L19)

denelon:
We should fix anything that would fail before accepting a change like this.

sredna:
> We should fix anything that would fail before accepting a change like this.

@denelon I have submitted PRs for all of these and most have been fixed already.

Two remaining that need attention:
- https://github.com/microsoft/winget-pkgs/pull/414601?notification_referrer_id=NT_kwHOAAJPl9oAJVJlcG9zaXRvcnk7MTk3Mjc1NTUxO0lzc3VlOzUxMDQ3Mjk1ODU#event-29193803948
- https://github.com/microsoft/winget-pkgs/pull/413875?notification_referrer_id=NT_kwHOAAJPl9oAJVJlcG9zaXRvcnk7MTk3Mjc1NTUxO0lzc3VlOzUwOTM1Mzg5NTE#issuecomment-5258625174

Issue #2899, opened 2023-02-01
Support installing other archive formats


rollingmoai:
### Description of the new feature / enhancement

Continuation of #140. See https://github.com/microsoft/winget-cli/issues/140#issuecomment-706523554:
> I agree that supporting other compression methods (zstd, lzma, etc.) and archive formats (7z, tar, etc.) is good and should eventually be added. Zips are the most common, so supporting just them is a good first step and I think it's better to have it ASAP and only after that think about other formats.

Other than .zip, most other software binaries are only available by either `.7z` and `.rar`. Examples:
- https://github.com/hashcat/hashcat/releases/download/v6.2.6/hashcat-6.2.6.7z
- https://sourceforge.net/projects/mpv-player-windows/files/release/mpv-0.35.1-x86_64.7z/download
- https://github.com/RussellJerome/UnrealModLoader/releases/download/v2.2.1/UnrealModLoader_V2.2.1.rar

Hopefully `.7z` and `.rar` gets added first, they're quite common. Other archive formats such as `.tar`, `.gz`, `.zstd` etc. can be added soon.

aescanes:
I support this.
Would like to be able to install mpv using winget or git-versions of gyan's ffmpeg

mloskot:
The lack of `.tar.gz` support is a blocker for packaging Kubernetes `kustomize` tool, see #101705

denelon:
@superusercode yes, we want to see if it's possible to take advantage of the new features in libarchive to support more formats. As soon as we have time to dig into the new Insider version of Windows and figure out the right way to handle this in earlier versions of Windows (back compatibility is tricky) we want this capability.

kbdharun:
+1 for adding `7z` support as it is the blocker for adding beta builds of [Dolphin emulator](https://dolphin-emu.org/download/) to the manifest under `DolphinEmulator.Beta.Dolphin`.

lucasfcnunes:
Adding `.tar.bz2` or [`.conda`](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/packages.html#conda-file-format) opens up the possibility of using the entire `conda-forge` [building system](https://conda-forge.org/status), which has more than [20k feedstocks](https://github.com/conda-forge/feedstocks/tree/main/feedstocks).

e.g.
https://anaconda.org/conda-forge/micromamba/files
https://anaconda.org/conda-forge/anaconda-project/files
https://anaconda.org/conda-forge/postgresql/files
[...]

PR #6278, opened 2026-06-10
Add `tar` archive installer type

tianon-sso:
## 📖 Description

Introduces `InstallerTypeEnum::Tar` as a new archive type alongside `InstallerTypeEnum::Zip`, allowing manifests to specify `InstallerType: tar`.  Aliases `tgz`, `tbz2`, `tbz`, `txz`, `tzst`, `tar.gz`, `tar.bz2`, `tar.xz`, `tar.zst`, and `tar.lz4` all map to the same enum value -- defined once in `IsTarInstallerTypeAlias` in `ManifestCommon.h` and shared by both the YAML and REST deserialization paths.

Extraction always routes through `tar.exe` (libarchive, ships with Windows 10 1803+) rather than the Shell API, which only understands zip.  For local-manifest installs, malware scanning is skipped with a visible warning -- `ScanZipFile` is zip-specific and a tar scanner is out of scope.  The remaining archive workflow layers (nested installer validation, path verification, nested installer resolution, repair) reuse the existing `IsArchiveType`-driven code paths, which are extended here to include `Tar`.

Updates `PackageInstallerType` in the COM/WinRT IDL (contract 14), the PowerShell enum, the REST deserializer, and the latest manifest JSON schema.

Assisted-By: "claude my eyes right out"

## 🔗 References

Implements part of https://github.com/microsoft/winget-cli/issues/2899 (not "resolves" because this is only tar and wrapped-tar formats, not 7z or rar as noted in the issue)

## 🔍 Validation

TBD

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6278)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
+++ b/src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
@@ -21,6 +21,13 @@ namespace AppInstaller::CLI::Workflow
     {
         if (context.Args.Contains(Execution::Args::Type::Manifest))
         {
+            if (context.Get<Execution::Data::Installer>().value().BaseInstallerType == InstallerTypeEnum::Tar)
+            {
+                AICLI_LOG(CLI, Warning, << "Archive malware scan is not supported for tar archives");
+                context.Reporter.Warn() << Resource::String::ArchiveScanNotSupportedForTar << std::endl;
+                return;
+            }
+
             bool scanResult = Archive::ScanZipFile(context.Get<Execution::Data::InstallerPath>());
 
             if (scanResult)
@@ -54,7 +61,10 @@ namespace AppInstaller::CLI::Workflow
         AICLI_LOG(CLI, Info, << "Extracting archive to: " << destinationFolder);
         context.Reporter.Info() << Resource::String::ExtractingArchive << std::endl;
 
-        if (Settings::User().Get<Settings::Setting::ArchiveExtractionMethod>() == Archive::ExtractionMethod::Tar)
+        const bool useTar = context.Get<Execution::Data::Installer>().value().BaseInstallerType == InstallerTypeEnum::Tar
+            || Settings::User().Get<Settings::Setting::ArchiveExtractionMethod>() == Archive::ExtractionMethod::Tar;
+
+        if (useTar)
         {
             context << ShellExecuteExtractArchive(installerPath, destinationFolder);
         }
--- a/src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
+++ b/src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
@@ -3,6 +3,7 @@
 #pragma once
 #include <AppInstallerStrings.h>
 #include <AppInstallerVersions.h>
+#include <algorithm>
 #include <functional>
 #include <map>
 #include <set>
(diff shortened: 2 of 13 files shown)


tianon-sso:
I'm opening this as a draft because I'm not sure what the level of interest in it is or whether it's worth doing more work to validate / test / document it further. :bow: :heart:

(I'm also happy if someone else wants to take it from here -- no pride in authorship here, just scratching an itch and I'm perfectly fine with this being closed if there's no appetite for it 💖)

tianon-sso:
(what I thought was most interesting was that the "difficult" bits of this change already exist in the codebase -- actually using `tar.exe` :heart:)

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
```diff
@@ -21,6 +21,13 @@ namespace AppInstaller::CLI::Workflow
     {
         if (context.Args.Contains(Execution::Args::Type::Manifest))
         {
+            if (context.Get<Execution::Data::Installer>().value().BaseInstallerType == InstallerTypeEnum::Tar)
+            {
+                AICLI_LOG(CLI, Warning, << "Archive malware scan is not supported for tar archives");
```
JohnMcPMS, line 26: I think we might need to figure out the answer to this gap before we could accept it.  I doubt tar is immune to the same kinds malicious use as zip.

src/AppInstallerCommonCore/Public/winget/ManifestCommon.h
```diff
@@ -117,6 +119,16 @@ namespace AppInstaller::Manifest
         Font,
     };
 
+    inline bool IsTarInstallerTypeAlias(std::string_view s)
+    {
+        using namespace std::string_view_literals;
+        static constexpr std::string_view aliases[] = {
+            "tar"sv, "tgz"sv, "tbz2"sv, "tbz"sv, "txz"sv, "tzst"sv,
```
JohnMcPMS, line 126: Not a big fan of a huge number of aliases here; I would rather just have the one name for it.  `Tar` might not even be the best name depending on how tar.exe determines the format.  As it is, this type is more like "compatible with `tar.exe -xf` based on the file name".  I think it would actually be good long term if we can set the format of tar to match what we expect, but I couldn't find that option.

So I guess what I'm saying is, I just want `Tar` but I also want to prevent people from lying and saying `Tar` when it is actually some `Foo` that tar.exe supports.  If we shipped with this and then eventually had some version of purelib that could scan tars, those "foos" would suddenly be broken unless the scanner just supports everything that tar.exe supports.

JohnMcPMS (commented):
Would need tests, although not as many as if you had needed to add invoking tar.exe.

src/AppInstallerCLICore/Workflows/ArchiveFlow.cpp
pl4nty: I've been looking into this in #4910, unfortunately AMSI doesn't look like a suitable alternative

tianon-sso:
Unfortunately I don't have as much bandwidth to see this through as I thought I might, so I'm going to close, but I'm totally fine with someone else taking any/all of this and carrying it. :bow: :heart:

pl4nty:
Thanks for your work on this! I'd love to pick it up once I've finished my other PRs, and see if there's a path forward for the malware scanning

PR #6281, opened 2026-06-11
feat: Add version diff highlighting to upgrade and list output

keanehatescoding:
Colorize the changed suffix of version strings when displaying pending upgrades via `winget upgrade` and `winget list`. The common prefix of the installed and available version is rendered in the default color, while the diverging suffix is highlighted: bright red for the installed version and bright green for the available version.

Highlighting is suppressed when VT is disabled (redirected output, --nowrap, or --no-vt), preserving clean machine-readable output.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6281)

### DIFF
--- a/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
+++ b/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
-#include "pch.h"
 #include "WorkflowBase.h"
 #include "ExecutionContext.h"
 #include "PackageTableSortHelper.h"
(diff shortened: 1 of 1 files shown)


Trenly (changes requested):
Test cases need to be added for this. It also should be opt-in as a user setting under the visual options as this had stong potential to break existing scripts relying on the raw output

src/AppInstallerCLICore/Workflows/WorkflowBase.cpp
florelis, line 1: Please keep the existing formatting for the code.

florelis:
> this had stong potential to break existing scripts relying on the raw output

Does it? I thought that we were able to detect when we were not running in a terminal and would produce the output as if we had --no-vt.

Trenly:
> > this had stong potential to break existing scripts relying on the raw output
> 
> Does it? I thought that we were able to detect when we were not running in a terminal and would produce the output as if we had --no-vt.

Hmmm. That is true (granted it is a recent addition) - but is there not the possibility that someone takes un-ridrected output from the attached terminal?

Either way, it seems like a strong case for a user setting - though that point may be moot since the head branch has been deleted, though I'm not sure why as I could imagine tests (and even a setting) would have been relatively straightforward to add compared to the other changes in the pr

PR #6258, opened 2026-05-29
Update NuGet Packages

Trenly:
## 📖 Description

Updates NuGet package versions in `src\Directory.Packages.props` to newer supported releases. These updates move the repository onto current package baselines that include accumulated servicing fixes, improve compatibility with newer .NET/runtime toolchains, and reduce long-term maintenance risk from version drift.

### 📦 Packages Updated

18 NuGet package versions were updated:

| Package | Old Version | New Version |
|---------|------------|------------|
| `coverlet.collector` | 3.1.2 | 10.0.1 |
| `Microsoft.ApplicationInsights` | 2.21.0 | 3.1.2 |
| `Microsoft.AspNet.WebApi.Client` | 5.2.9 | 6.0.0 |
| `Microsoft.Extensions.Hosting` | 9.0.6 | 10.0.8 |
| `Microsoft.NET.Test.Sdk` | 17.3.2 | 18.6.0 |
| `Microsoft.PowerShell.SDK` | 7.4.12 | 7.4.16 |
| `Microsoft.Windows.SDK.Contracts` | 10.0.26100.1742 | 10.0.28000.1839 |
| `Newtonsoft.Json` | 13.0.3 | 13.0.4 |
| `nunit` | 3.12.0 | 3.14.0 |
| `NUnit3TestAdapter` | 3.15.1 | 6.2.0 |
| `Octokit` | 4.0.3 | 14.0.0 |
| `Semver` | 2.3.0 | 3.0.0 |
| `System.Data.SqlClient` | 4.9.0 | 4.9.1 |
| `System.Text.Json` | 9.0.6 | 10.0.8 |
| `xunit` | 2.4.2 | 2.9.3 |
| `xunit.runner.visualstudio` | 2.4.5 | 3.1.5 |
| `YamlDotNet` | 16.3.0 | 18.0.0 |

### Why This Helps

- **Aligns key dependencies** with actively serviced versions used by current SDK/tooling
- **Brings upstream bug fixes** from package lines, including reliability fixes in test discovery/execution stacks and runtime/library servicing fixes delivered through newer .NET package trains
- **Improves test framework coverage** by upgrading NUnit (3.12.0 → 4.6.1), NUnit3TestAdapter (3.15.1 → 6.2.0), and xunit (2.4.2 → 2.9.3) to modern versions
- **Maintains healthier security posture** by updating transitive dependencies

### 🔒 Security Improvements

Updated packages include security fixes from:
[...]

PR #6287, opened 2026-06-16
Update NUnit to v4

Trenly:
## 📖 Description

Upgrade the AppInstallerCLIE2ETests suite to NUnit 4 and modernize the supporting test code to the NUnit 4 constraint model. This lands on the latest available NUnit 4.x release in NuGet (`4.6.1`)

The update converts the E2E suite off NUnit 3 classic assertions, fixes the delegate-based assertion overloads required by NUnit 4, and carries the package version change through the affected test projects that use centralized package management.

I opted for full migration instead of the simpler `Directory.Build.props` change as I expect this will be a one-time migration and I generally believe that using the newer format will be more performant in the future (plus I had the extra tokens to burn, so why not)

## 🔗 References

* Split from #6258
* NUnit 4 migration guide: https://docs.nunit.org/articles/nunit/release-notes/Nunit4.0-MigrationGuide.html

## 🔍 Validation

- Search-based verification confirmed the old NUnit 3 classic assertions were removed from `src/AppInstallerCLIE2ETests`.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [x] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6287)


src/AppInstallerCLIE2ETests/Helpers/TestCommon.cs
```diff
@@ -426,15 +426,15 @@ public static void VerifyFontPackage(
...
+                        Assert.That(fileList.Count, Is.EqualTo(valueNames.Length));
                     }
                 }
             }
 
             // Verify each package file we expect to exist actually exists.
             foreach (var file in fileList)
             {
-                Assert.IsTrue(File.Exists(file));
+                Assert.That(File.Exists(file), Is.True);
```
florelis: I think you can do `Assert.That(file, Does.Exist)`, or even go up one level and do `Assert.That(fileList, Has.All.Exist)`

src/AppInstallerCLIE2ETests/Helpers/TestCommon.cs
```diff
@@ -675,9 +675,9 @@ public static void AssertInstallerDownload(
             string installerDownloadPath = Path.Combine(downloadDir, expectedFileName + installerExtension);
             string manifestDownloadPath = Path.Combine(downloadDir, expectedFileName + ".yaml");
 
-            Assert.IsTrue(Directory.Exists(downloadDir), $"Download directory does not exist: {downloadDir}");
-            Assert.IsTrue(File.Exists(installerDownloadPath), $"Installer file does not exist: {installerDownloadPath}");
-            Assert.IsTrue(File.Exists(manifestDownloadPath), $"Manifest file does not exist: {manifestDownloadPath}");
+            Assert.That(Directory.Exists(downloadDir), Is.True, $"Download directory does not exist: {downloadDir}");
+            Assert.That(File.Exists(installerDownloadPath), Is.True, $"Installer file does not exist: {installerDownloadPath}");
+            Assert.That(File.Exists(manifestDownloadPath), Is.True, $"Manifest file does not exist: {manifestDownloadPath}");
```
florelis: Can use `Does.Exist`

src/AppInstallerCLIE2ETests/Helpers/TestCommon.cs
```diff
@@ -1111,7 +1111,7 @@ public static string GetExpectedModulePath(TestModuleLocation location)
         public static string GetConfigurationInstanceIdentifierFor(string name)
         {
             var result = TestCommon.RunAICLICommand("configure list", string.Empty);
-            Assert.AreEqual(0, result.ExitCode);
+            Assert.That(result.ExitCode, Is.EqualTo(0));
```
florelis:
```suggestion
            Assert.That(result.ExitCode, Is.Zero);
```

src/AppInstallerCLIE2ETests/Interop/GroupPolicyForInterop.cs
```diff
@@ -70,56 +70,80 @@ public void DisableWinGetPolicy()
...
-
-            groupPolicyException = Assert.Catch<GroupPolicyException>(() => { RemovePackageCatalogOptions packageManagerSettings = this.TestFactory.CreateRemovePackageCatalogOptions(); });
-            Assert.AreEqual(Constants.BlockByWinGetPolicyErrorMessage, groupPolicyException.Message);
-            Assert.AreEqual(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY, groupPolicyException.HResult);
-
-            groupPolicyException = Assert.Catch<GroupPolicyException>(() => { EditPackageCatalogOptions packageManagerSettings = this.TestFactory.CreateEditPackageCatalogOptions(); });
-            Assert.AreEqual(Constants.BlockByWinGetPolicyErrorMessage, groupPolicyException.Message);
-            Assert.AreEqual(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY, groupPolicyException.HResult);
+            Assert.That(
+                (Action)(() => { PackageManager packageManager = this.TestFactory.CreatePackageManager(); }),
```
florelis: Is the cast to ` Action` needed?

src/AppInstallerCLIE2ETests/Interop/GroupPolicyForInterop.cs
```diff
@@ -70,56 +70,80 @@ public void DisableWinGetPolicy()
...
+                Throws.TypeOf<GroupPolicyException>()
+                    .With.Message.EqualTo(Constants.BlockByWinGetPolicyErrorMessage)
+                    .And.Property("HResult").EqualTo(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY));
```
florelis: I wonder if it would make sense to add something like an `HResultConstraint` so we could do `And.HResult.EqualTo(...)`

src/AppInstallerCLIE2ETests/Interop/InstallInterop.cs
```diff
@@ -669,8 +669,8 @@ public async Task InstallExe_VerifyInstalledCatalog()
 
             // Ensure package is not installed
             var installedResult = this.FindAllPackages(installedCatalogReference, PackageMatchField.ProductCode, PackageFieldMatchOption.Equals, Constants.ExeInstalledDefaultProductCode);
-            Assert.IsNotNull(installedResult);
-            Assert.AreEqual(0, installedResult.Count);
+            Assert.That(installedResult, Is.Not.Null);
+            Assert.That(installedResult.Count, Is.EqualTo(0));
```
florelis:
```suggestion
            Assert.That(installedResult.Count, Is.Zero);
```

src/AppInstallerCLIE2ETests/Interop/GroupPolicyForInterop.cs
florelis: Or maybe keep the old pattern
```csharp
groupPolicyException = Assert.Throws<GroupPolicyException>(() => { ... });
Assert.That(groupPolicyException.message, Is.EqualTo(...));
```

src/AppInstallerCLIE2ETests/Interop/RepairInterop.cs
```diff
@@ -283,7 +283,7 @@ public async Task RepairInnoInstallerWithInstallerRepairBehavior()
             await this.RepairPackageAndValidateStatus(searchResult.CatalogPackage, this.TestFactory.CreateRepairOptions(), RepairResultStatus.Ok);
 
             // Cleanup
-            Assert.True(TestCommon.VerifyTestExeRepairCompletedAndCleanup(this.installDir, "Installer Repair operation"));
+            Assert.That(TestCommon.VerifyTestExeRepairCompletedAndCleanup(this.installDir, "Installer Repair operation"), Is.True);
```
florelis, line 286: Not for this PR, but maybe at some point we could rewrite these helpers to be constraints. Like `Assert.That(this.installDir, Repair.Exe.Completed)` (but better)

src/AppInstallerCLIE2ETests/Interop/Shutdown.cs
```diff
@@ -38,17 +38,17 @@ public void NoActiveOperations()
             var packageManager = this.TestFactory.CreatePackageManager();
 
             var servers = WinGetServerInstance.GetInstances();
-            Assert.AreEqual(1, servers.Count);
+            Assert.That(servers.Count, Is.EqualTo(1));
 
             var server = servers[0];
-            Assert.IsTrue(server.HasWindow);
+            Assert.That(server.HasWindow, Is.True);
```
florelis:
```suggestion
            Assert.That(server.HasWindow);
```

src/AppInstallerCLIE2ETests/Interop/UninstallInterop.cs
```diff
@@ -261,20 +261,20 @@ public async Task UninstallPortableModifiedSymlink()
...
             searchResult = this.FindOnePackage(this.compositeSource, PackageMatchField.Id, PackageFieldMatchOption.Equals, packageId);
-            Assert.NotNull(searchResult.CatalogPackage.InstalledVersion);
+            Assert.That(searchResult.CatalogPackage.InstalledVersion, Is.Not.Null);
 
             // Uninstall
             var uninstallResult = await this.packageManager.UninstallPackageAsync(searchResult.CatalogPackage, this.TestFactory.CreateUninstallOptions());
-            Assert.AreEqual(UninstallResultStatus.UninstallError, uninstallResult.Status);
-            Assert.True(modifiedSymlinkInfo.Exists, "Modified symlink should still exist");
+            Assert.That(uninstallResult.Status, Is.EqualTo(UninstallResultStatus.UninstallError));
+            Assert.That(modifiedSymlinkInfo.Exists, Is.True, "Modified symlink should still exist");
```
florelis, line 277:
```suggestion
            Assert.That(modifiedSymlinkInfo, Does.Exist, "Modified symlink should still exist");
```

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilDownload.cs
```diff
@@ -30,8 +30,8 @@ public void WinGetUtil_Download()
             // Download
             WinGetUtilWrapper.WinGetDownload(installerUrl, filePath, sha256Hash, hashSize);
 
-            Assert.True(File.Exists(filePath));
-            Assert.False(sha256Hash.All(byteVal => byteVal == 0));
+            Assert.That(File.Exists(filePath), Is.True);
+            Assert.That(sha256Hash.All(byteVal => byteVal == 0), Is.False);
```
florelis:
```suggestion
            Assert.That(sha256Hash, Is.Not.All.Zero);
```

src/AppInstallerCLIE2ETests/WinGetUtil/WinGetUtilManifest.cs
```diff
@@ -155,10 +155,10 @@ private static void ValidateSchemaHeaderFailure(string manifestsFilePath, WinGet
...
                 createManifestOption);
 
-            Assert.False(succeeded);
-            Assert.AreEqual(IntPtr.Zero, manifestHandle);
-            Assert.IsNotNull(createFailureMessage);
-            Assert.IsTrue(createFailureMessage.Contains(expectedError));
+            Assert.That(succeeded, Is.False);
+            Assert.That(manifestHandle, Is.EqualTo(IntPtr.Zero));
+            Assert.That(createFailureMessage, Is.Not.Null);
+            Assert.That(createFailureMessage.Contains(expectedError), Is.True);
```
florelis: Here and in every other `Contains()`
```suggestion
            Assert.That(createFailureMessage, Contains.Substring(expectedError));
```

src/AppInstallerCLIE2ETests/ConfigureProcessorPathCommand.cs
```diff
@@ -114,10 +114,8 @@ public void ProcessorPath_AuditOutput_HashIsValidSHA256()
                 ? result.StdOut.Substring(hashStart, hashEnd - hashStart).Trim()
                 : result.StdOut.Substring(hashStart).Trim();
 
-            Assert.AreEqual(64, hashValue.Length, $"Expected 64-character SHA256 hash, got: '{hashValue}'");
-            Assert.True(
-                Regex.IsMatch(hashValue, "^[0-9a-f]{64}$"),
-                $"Expected lowercase hex hash, got: '{hashValue}'");
+            Assert.That(hashValue.Length, Is.EqualTo(64), $"Expected 64-character SHA256 hash, got: '{hashValue}'");
+            Assert.That(Regex.IsMatch(hashValue, "^[0-9a-f]{64}$"), Is.True, $"Expected lowercase hex hash, got: '{hashValue}'");
```
florelis:
```suggestion
            Assert.That(hashValue, DOes.Match("^[0-9a-f]{64}$"), $"Expected lowercase hex hash, got: '{hashValue}'");
```

src/AppInstallerCLIE2ETests/DSCv3AdminSettingsResourceCommand.cs
```diff
@@ -77,11 +77,11 @@ public void AdminSettings_Get(string function)
...
 
             AdminSettingsResourceData output = GetSingleOutputLineAs<AdminSettingsResourceData>(result.StdOut);
-            Assert.IsNotNull(output);
-            Assert.IsNotNull(output.Settings);
-            Assert.IsTrue(output.Settings.ContainsKey(LocalManifestFiles));
-            Assert.IsFalse(output.Settings.ContainsKey(DefaultProxy));
-            Assert.IsFalse(output.Settings.ContainsKey(NotAnAdminSettingName));
+            Assert.That(output, Is.Not.Null);
+            Assert.That(output.Settings, Is.Not.Null);
+            Assert.That(output.Settings.ContainsKey(LocalManifestFiles), Is.True);
```
florelis, line 82: Depending on the type of `Settings`, maybe this could be
```csharp
            Assert.That(output.Settings, Contains.Key(LocalManifestFiles));
```

src/AppInstallerCLIE2ETests/DSCv3SourceResourceCommand.cs
```diff
@@ -244,7 +244,7 @@ public void Source_Test_PropertyMismatch(bool useDefaultArgument, string trustLe
                     resourceData.Priority = (int)testValue;
                     break;
                 default:
-                    Assert.Fail($"{targetProperty} is not a handled case.");
+                    Assert.That(false, Is.True, $"{targetProperty} is not a handled case.");
```
florelis: Can't we keep `Assert.Fail()`?

florelis (commented):
I think it would be good to use some of the native constraints in more places. Like
`Assert.That(x, Is.EqualTo(0)) -> Assert.That(x, Is.Zero)`
`Assert.That(s.Contains("substr"), Is.True) -> Assert.That(s, Contains.Substring("substr"))`
`Assert.That(File.Exists(f), Is.True) -> Assert.That(f, Does.Exist)`

src/AppInstallerCLIE2ETests/Interop/GroupPolicyForInterop.cs
Trenly: Investigated; The casts are necessary for disambiguation:
```
2>E:\winget-cli\src\AppInstallerCLIE2ETests\Interop\GroupPolicyForInterop.cs(109,20,109,24): error CS0121: The call is ambiguous between the following methods or properties: 'NUnit.Framework.Assert.That(NUnit.Framework.TestDelegate, NUnit.Framework.Constraints.IResolveConstraint, NUnit.Framework.NUnitString, string, string)' and 'NUnit.Framework.Assert.That(System.Action, NUnit.Framework.Constraints.IResolveConstraint, NUnit.Framework.NUnitString, string, string)'
```

src/AppInstallerCLIE2ETests/Interop/GroupPolicyForInterop.cs
Trenly: Th old pattern seems a little cleaner, though the cast is still necessary. Updated to be as clear as possible

src/AppInstallerCLIE2ETests/DSCv3AdminSettingsResourceCommand.cs
Trenly: It could be, but I wanted to keep all the tests consistent with how they handle JSON. I'm happy to update across all the files if that is preferred, but don't see a lot of benefit in that. Of course, it may also be easier to take that as a follow on to minimize the amount of time needed for a re-review on this PR

florelis (dismissed):

src/Directory.Packages.props
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="coverlet.collector" Version="3.1.2" />
-    <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.21.0" />
+    <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.23.0" />
```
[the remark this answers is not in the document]
Trenly: Dependency update is required to avoid a downgrade warning

src/Directory.Packages.props
```diff
@@ -22,8 +22,8 @@
     <PackageVersion Include="ModelContextProtocol" Version="0.3.0-preview.3" />
     <PackageVersion Include="Moq" Version="4.20.72" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageVersion Include="nunit" Version="3.12.0" />
-    <PackageVersion Include="NUnit3TestAdapter" Version="3.15.1" />
+    <PackageVersion Include="nunit" Version="4.6.1" />
+    <PackageVersion Include="NUnit3TestAdapter" Version="6.2.0" />
```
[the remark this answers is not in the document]
Trenly: Updated description

florelis:
I don't think any of the Copilot comments need to be resolved. The BOM issue I would want to have fixed, but I think that would be a repo-level thing, like the normalization of line endings we did.

src/AppInstallerCLIE2ETests/Interop/GroupPolicyForInterop.cs
```diff
@@ -70,56 +70,56 @@ public void DisableWinGetPolicy()
...
-            Assert.AreEqual(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY, groupPolicyException.HResult);
+            GroupPolicyException packageMatchFilterException = Assert.Throws<GroupPolicyException>((Action)(() => { PackageMatchFilter packageMatchFilter = this.TestFactory.CreatePackageMatchFilter(); }));
+            Assert.That(packageMatchFilterException.Message, Is.EqualTo(Constants.BlockByWinGetPolicyErrorMessage));
+            Assert.That(packageMatchFilterException.HResult, Is.EqualTo(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY));
 
-            groupPolicyException = Assert.Catch<GroupPolicyException>(() => { RepairOptions repairOptions = this.TestFactory.CreateRepairOptions(); });
-            Assert.AreEqual(Constants.BlockByWinGetPolicyErrorMessage, groupPolicyException.Message);
-            Assert.AreEqual(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY, groupPolicyException.HResult);
+            GroupPolicyException repairOptionsException = Assert.Throws<GroupPolicyException>((Action)(() => { RepairOptions repairOptions = this.TestFactory.CreateRepairOptions(); }));
```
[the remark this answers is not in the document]
Trenly: The action cast is necessary to prevent a compilation failure

src/AppInstallerCLIE2ETests/Interop/GroupPolicyForInterop.cs
```diff
@@ -70,56 +70,56 @@ public void DisableWinGetPolicy()
...
-            Assert.AreEqual(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY, groupPolicyException.HResult);
+            GroupPolicyException packageManagerException = Assert.Throws<GroupPolicyException>((Action)(() => { PackageManager packageManager = this.TestFactory.CreatePackageManager(); }));
+            Assert.That(packageManagerException.Message, Is.EqualTo(Constants.BlockByWinGetPolicyErrorMessage));
+            Assert.That(packageManagerException.HResult, Is.EqualTo(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY));
 
-            groupPolicyException = Assert.Catch<GroupPolicyException>(() => { FindPackagesOptions findPackagesOptions = this.TestFactory.CreateFindPackagesOptions(); });
-            Assert.AreEqual(Constants.BlockByWinGetPolicyErrorMessage, groupPolicyException.Message);
-            Assert.AreEqual(Constants.ErrorCode.ERROR_BLOCKED_BY_POLICY, groupPolicyException.HResult);
+            GroupPolicyException findPackagesOptionsException = Assert.Throws<GroupPolicyException>((Action)(() => { FindPackagesOptions findPackagesOptions = this.TestFactory.CreateFindPackagesOptions(); }));
```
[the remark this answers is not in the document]
Trenly: The action cast is necessary to prevent a compilation failure

Trenly:
> I don't think any of the Copilot comments need to be resolved. The BOM issue I would want to have fixed, but I think that would be a repo-level thing, like the normalization of line endings we did.

Reading through them, I agree. If the pipelines fail, I'll address the two that I've left open. I've replied to the others with my justification for resolving them

Trenly:
This looks fun . . . 
```
##[error]TearDown failed for test fixture AppInstallerCLIE2ETests.SetUpFixture
##[error]One or more child tests were ignored
##[error]TearDown : System.IO.IOException : The file 'C:\Users\VssAdministrator\AppData\Local\Temp\E2ETestLogs\Unpackaged\WinGetCOM-2026-07-14-21-20-31.670.log' already exists.
##[error]StackTrace: --TearDown
##[error]   at System.IO.FileSystem.CopyFile(String sourceFullPath, String destFullPath, Boolean overwrite)
##[error]   at AppInstallerCLIE2ETests.Helpers.TestCommon.CopyDirectory(String sourceDirName, String destDirName) in D:\a\1\s\src\AppInstallerCLIE2ETests\Helpers\TestCommon.cs:line 1070
##[error]   at AppInstallerCLIE2ETests.Helpers.TestCommon.PublishE2ETestLogs() in D:\a\1\s\src\AppInstallerCLIE2ETests\Helpers\TestCommon.cs:line 545
##[error]   at AppInstallerCLIE2ETests.SetUpFixture.TearDown() in D:\a\1\s\src\AppInstallerCLIE2ETests\SetUpFixture.cs:line 84
##[error]   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
##[error]   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

Will trace down the root cause when I get home.  However . . . looks like the tests themselves are passing, so that's promising at least
```
Total tests: 100
     Passed: 94
    Skipped: 6
```

Trenly:
@florelis - This is an already existing defect which happens in Azure because the log files currently have the `overwrite` flag set to `false`. When a PR is run multiple times in Azure the files already exist so this copy fails. You can see this bug in the [logs from #6365](https://dev.azure.com/shine-oss/winget-cli/_build/results?buildId=366399&view=logs&j=6be4421a-07cc-59f5-6f5a-a15018768f23&t=c5b9722e-3ece-57b9-37e4-b94e8b6771f9&l=168).

It appears that this has always been a bug in the test adapter and it was fixed on the newer version of the NuGet package, which is why it is causing a fault now.

I would like some advice on how to approach the fix. It feels like this may be a case where just setting overwrite to `true` is acceptable, but didn't want to do that without confirming first

florelis:
I don't think it has to do with the PR running multiple times. I believe it happens because we copy the winget log folder to the output directory at multiple times within the same run, so we end up trying to copy the same file again. And I'd guess it started happening because the previous adapter would ignore errors during teardown.

Overwriting should be fine since it should be the same file, even the timestamp has to match. We could add the parameter to `CopyDirectory()` so we can limit it to only the places where we are copying the logs. Worst case, it can't be worse than what we have today - we would still be missing one of the logs files, it's just the older one that is missing now.

A better solution could be to prevent the double copy from happening to begin with. I believe this could be prevented if we cleared the winget logs during the `OneTimeSetUp`. So, each time we call the E2E tests with different options, we would only need to copy what that test run generated.

Trenly:
> I don't think it has to do with the PR running multiple times. I believe it happens because we copy the winget log folder to the output directory at multiple times within the same run, so we end up trying to copy the same file again. And I'd guess it started happening because the previous adapter would ignore errors during teardown.
> 

I see . . . .That makes sense

> Overwriting should be fine since it should be the same file, even the timestamp has to match. We could add the parameter to `CopyDirectory()` so we can limit it to only the places where we are copying the logs. Worst case, it can't be worse than what we have today - we would still be missing one of the logs files, it's just the older one that is missing now.
> 

100% agree. I'll take this path for simplicity

> A better solution could be to prevent the double copy from happening to begin with. I believe this could be prevented if we cleared the winget logs during the `OneTimeSetUp`. So, each time we call the E2E tests with different options, we would only need to copy what that test run generated.

Sounds like this will be another great follow on issue which I can file and look at

florelis (approved):

Trenly:
Ah, finally, a green checkmark

Issue #6298, opened 2026-06-18
Day 0 Update: Self-update on first command when client is outdated


denelon:
### Relevant area(s)

WinGet CLI

### Description of the new feature / enhancement

**Day 0 Update** — When the first command requiring the source index is executed (e.g., `winget install`, `winget upgrade`, `winget search`, or `winget configure`), WinGet should check whether a newer stable version of the client is available and prompt the user to update before proceeding.

This addresses several chicken-and-egg problems:
- A user installs WinGet (or gets it via Windows) but the version is already outdated by the time they first use it.
- New features (e.g., `winget configure`) require a recent client, but users don't know they need to update first.
- The `winget configure enable` step should be unnecessary if the first command executed is `winget configure` on a client that supports it natively — but today users hit "unrecognized command" errors because their client is too old.

**Proposed behavior:**
1. The preIndexed package source includes metadata indicating the "latest stable" client version.
2. On first command execution requiring the index, WinGet compares its own version against the advertised latest.
3. If outdated, WinGet informs the user in the current terminal window that an update is needed.
4. WinGet shuts down, updates itself, then relaunches in a new window with the original command (using `--wait` so the new window doesn't close immediately).
5. The user sees their original command execute on the updated client.

This eliminates the "install WinGet, then immediately need to update WinGet" friction and removes barriers to new functionality adoption.

### Proposed technical implementation details

- Add a `latestStableVersion` field to the preIndexed package source metadata.
- Client version comparison on first index access per session.
- Self-update flow: notify user → close current process → trigger update (via Microsoft Store or direct package) → relaunch with original arguments + `--wait`.
- Group Policy controls: `EnableDay0Update` (toggle), `Day0UpdateSource` (Store vs. direct).
- Settings: `day0Update.enabled`, `day0Update.dismissCount` (allow N dismissals before suppressing).

Issue #5841, opened 2025-10-29
Augmented cache control for PreIndexed sources.


denelon:
In addition to better behavior when WinGet cannot refresh the cache, some users would prefer greater control of what to do in the event the cache is stale.

WinGet currently does have a TTL setting for the default SLA of the cache, but maybe there is a need for even more explicit control.

Users can run `winget source update` to force a cache update (assuming the network is functioning correctly), but there is no notion if running a command and explicitly avoiding the cache refresh.

From:
* #5834 

> [@lkrms](https://github.com/lkrms) we have been discussing the logic used on a retry and the best way to proceed if we can't fetch an updated cache. In one case if we're not able to get the "initial" cache, it would be a complete failure. In the event of a stale cache we could proceed with that stale cache and some kind of informational message about how old the cache is.
> 
> It makes sense for `winget` to keep its cache in sync transparently; I'm coming at it from a Linux package management angle (I've already mentioned `pacman`, but `apt` and others are similar), where in the absence of optional tooling, the cache isn't refreshed from upstream unless you explicitly request it.
> 
> I'm not suggesting `winget` should do likewise--it's a different beast with a different audience--but the case for allowing users to "sync" and "install" as distinct operations if they choose to is arguably stronger for `winget` than for a typical Linux package manager, simply because its cache is, by definition, decoupled from the hosting platform of basically every package it services.
> 
> If I can't sync my `pacman` cache, I can't install packages either, because they all come from the same place; but `winget` doesn't suffer from this constraint 🙂. Its default behaviour when the cache is unreachable is important to get right, especially when you factor in bootstrapping scenarios, but perhaps it could be left as-is, albeit with improvements to how the failure is communicated to end-users. If a `--no-sync` option (and/or a `no-implicit-cache-sync` setting) were available, it could be suggested as a way to mitigate the issue temporarily, for example. And for users who would always enable `--no-sync`, I guess you'd need a standalone `sync` command. Best of both worlds? 

 _Originally posted by @lkrms in [#5834](https://github.com/microsoft/winget-cli/issues/5834#issuecomment-3461320674)_

lkrms:
I hadn't picked up on the existing `winget source update` command; the only missing piece here seems to be a `--no-sync` option/setting pair (probably with a more consistent name, e.g. `--no-update`). 🙏🏻

Issue #4868, opened 2024-10-11
Cannot enable WinGet Configuration in WinGet 1.9.2507


denelon:
### Brief description of your issue

WinGet Configuration fails to be enabled.

### Steps to reproduce

Run `winget configure --enable`

### Expected behavior

WinGet Configuration is enabled.

### Actual behavior

```
Enabling configuration components. Requires store access.
An unexpected error occurred while executing the command:
0x8a15002b : No applicable update found
```

### Environment

Windows Package Manager v1.9.2507
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.26100.2023
System Architecture: Arm64
Package: Microsoft.DesktopAppInstaller v1.24.2507.0

PR #6299, opened 2026-06-18
Specification: Day 0 Update

denelon:
## 📖 Description

Specification for the Day 0 Update feature — enabling WinGet to detect that it is outdated on first meaningful command execution and self-update before proceeding. This eliminates the chicken-and-egg problem where WinGet must already be current to use new functionality (e.g., `winget configure`), and removes the need for `winget configure enable` on fresh installs.

Key aspects:
- A `latestStableClientVersion` field in preIndexed source metadata signals the latest stable client version
- On first index access, the client compares its version and prompts to update if outdated
- Self-update flow: notify user → exit → update via existing infrastructure → relaunch original command using `winget resume` via `ShellExecute`
- Group Policy controls for enterprise management (`EnableDay0Update`, `ForceMinimumClientVersion`)
- COM API exposes status only (never prompts or auto-updates)
- Respects `--nowarn` (suppresses update warning; command fails with error code if client is too old)
- Targets `Microsoft.AppInstaller` from the winget default source explicitly (no implicit agreement acceptance)

**Changes addressing review feedback (June 22):**
- Clarified how client infers new functionality from version signal and index/schema version
- Added `--nowarn` support: suppresses warning entirely, fails with error code if command requires newer client
- Replaced `wt.exe`/`cmd.exe` relaunch with `ShellExecute` on app execution alias — no terminal dependency
- Leveraged `winget resume` (experimental) for restoring original command context after self-update
- Removed `--accept-package-agreements`/`--accept-source-agreements` from self-update command; explicit `--source winget` to target the default source without asserting acceptance on behalf of the user

Authored with GitHub Copilot assistance.

## 🔗 References

Related Issues:
- https://github.com/microsoft/winget-cli/issues/6298
- https://github.com/microsoft/winget-cli/issues/5841
- https://github.com/microsoft/winget-cli/issues/4868

## 🔍 Validation

Spec document — no code changes to validate.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6299)

### DIFF
--- a/doc/specs/#6298 - Day 0 Update.md
+++ b/doc/specs/#6298 - Day 0 Update.md
@@ -95,43 +95,46 @@ When an update is available:
    - Record the dismissal (increment counter in local state).
    - Proceed with the original command on the current version.
    - If the command fails because the client is too old (e.g., unrecognized subcommand), show a targeted message: "This feature requires WinGet 1.29+. Run `winget upgrade Microsoft.AppInstaller` to update."
+   - The client determines available functionality in the newer version by examining the version signal in the source metadata. When the source index or schema version exceeds what the current client understands, the client infers that new functionality is available in the advertised `latestStableClientVersion`. This heuristic may not capture all edge cases, but provides a reasonable signal for the most common scenario: a user attempting a command that requires a newer client.
 
 4. **If non-interactive** (`--disable-interactivity`, COM API, Configuration):
    - Do not prompt. Proceed with current version.
    - Emit a warning to stderr: "WinGet update available (1.30.1234.0). Run `winget upgrade Microsoft.AppInstaller` to update."
+   - If `--nowarn` is specified, suppress the warning entirely. The command proceeds silently on the current version. If the current version cannot execute the command (e.g., unrecognized subcommand), the command fails with an appropriate error code — the update does not occur implicitly.
 
 ### Self-Update Mechanism
 
-The update uses the existing `winget upgrade` infrastructure targeting `Microsoft.AppInstaller`:
+The update targets `Microsoft.AppInstaller` from the winget default source:
 
 ```
-winget upgrade --id Microsoft.AppInstaller --silent --accept-package-agreements --accept-source-agreements
+winget upgrade --id Microsoft.AppInstaller --source winget --silent
 ```
 
 Because WinGet cannot update its own running process:
 
-1. WinGet spawns a detached update orchestrator process (a small helper EXE bundled with the client).
-2. The orchestrator:
+1. WinGet records the original command line and context using the resume infrastructure (`winget resume`).
+2. WinGet spawns a detached update orchestrator process (a small helper EXE bundled with the client).
+3. The orchestrator:
    - Waits for the original WinGet process to exit.
    - Triggers the Microsoft Store update or direct MSIX install.
-   - On success: launches a new `conhost`/`wt` window with the saved command + `--wait`.
-   - On failure: launches a new window with an error message and the original command for manual retry.
+   - On success: uses `ShellExecute` to invoke the app execution alias (`winget`) with `resume` to continue the original operation. This avoids a hard dependency on Windows Terminal or `cmd.exe` and works regardless of the user's default terminal application.
+   - On failure: uses `ShellExecute` to launch `winget` with an error message displayed and the original command for manual retry.
 
 ### Relaunch Behavior
 
-The relaunch uses `--wait` to keep the new terminal window open after the command completes:
+The relaunch leverages `winget resume` (currently experimental) to restore the original command context:
 
 ```
-wt.exe --wait -- winget configure myconfig.yaml
+winget resume --resume-id <saved-id>
 ```
 
-If Windows Terminal is not available, fall back to:
+The resume infrastructure records the full command line, working directory, and any relevant state. After a successful self-update, the orchestrator invokes `winget resume` via `ShellExecute` on the app execution alias. This delegates terminal window creation to the operating system's default handler, avoiding any dependency on a specific terminal application.
+
+If `winget resume` is unavailable (older client somehow reached this path), fall back to `ShellExecute` with the original command line:
 
 ```
-cmd.exe /k winget configure myconfig.yaml
+ShellExecute(NULL, "open", "winget", "configure myconfig.yaml --wait", workingDir, SW_SHOWNORMAL)
 ```
-
-This ensures the user sees the full output of their original command and can inspect results.
 
 ### `winget configure` Special Case
 
--- a/doc/specs/#6298 - Day 0 Update.md
+++ b/doc/specs/#6298 - Day 0 Update.md
@@ -11,12 +11,12 @@ For [#6298](https://github.com/microsoft/winget-cli/issues/6298).
 
 ## Abstract
 
-Enable WinGet to detect that it is outdated and self-update on first meaningful command execution. When the first command requiring the source index runs, the client checks a version signal in the preIndexed package source, determines whether a newer stable release exists, and — with user consent — updates itself and relaunches the original command. This eliminates the "chicken and egg" problem where WinGet must already be current to use new functionality (e.g., `winget configure`).
+Enable WinGet to detect that it is outdated and self-update on first meaningful command execution. When the first command requiring the source index runs, the client checks a version signal in the winget default source's preIndexed package metadata, determines whether a newer stable release exists, and — with user consent — updates itself and relaunches the original command. This eliminates the "chicken and egg" problem where WinGet must already be current to use new functionality (e.g., `winget configure`).
 
 ## Inspiration
 
 - Users who receive WinGet via Windows or the Microsoft Store often have an outdated client by the time they first use it. New features (Configuration, DSC v3, parallel install) are unreachable until the user independently discovers they need to update.
-- `winget configure enable` is a friction point: users attempting `winget configure` for the first time hit "unrecognized command" because their client predates the feature. Day 0 Update removes this barrier by updating the client before the command fails.
+- `winget configure enable` is a friction point: many devices ship with the "Stub App" version of App Installer (containing only WinGet's package management code). Users attempting `winget configure` for the first time must first run `winget configure enable` to download the configuration management components. Day 0 Update removes this barrier by updating the client to a full version before the command fails.
 - Other package managers solve this with auto-update (Homebrew, rustup, pip self-check). WinGet currently has no equivalent.
 - Enterprise provisioning scenarios (OOBE, Configuration-as-Code) benefit from guaranteed-current clients without requiring a separate pre-flight update step.
 
@@ -46,30 +46,30 @@ This metadata is updated as part of the source publication pipeline whenever a n
 User runs: winget configure myconfig.yaml
          │
          ▼
-┌─────────────────────────────┐
-│ Source index access required │
-│ (first command this session)│
-└──────────────┬──────────────┘
-               │
-               ▼
-┌─────────────────────────────┐
-│ Read latestStableClientVersion │
-│ from source metadata          │
-└──────────────┬──────────────┘
-               │
-               ▼
-┌─────────────────────────────┐
-│ Compare against own version │
-│ (Package.Current.Id.Version)│
-└──────────────┬──────────────┘
-               │
-        ┌──────┴──────┐
-        │ Up to date? │
-        └──────┬──────┘
-          Yes  │  No
-          │    │
-          ▼    ▼
-       Proceed  ──► Day 0 Update Flow
+┌───────────────────────────────────┐
+│ Source index access required       │
+│ (first command this session)       │
+└────────────────┬──────────────────┘
+                 │
+                 ▼
+┌───────────────────────────────────┐
+│ Read latestStableClientVersion     │
+│ from source metadata               │
+└────────────────┬──────────────────┘
+                 │
+                 ▼
+┌───────────────────────────────────┐
+│ Compare against own version        │
+│ (Package.Current.Id.Version)       │
+└────────────────┬──────────────────┘
+                 │
+          ┌──────┴──────┐
+          │ Up to date? │
+          └──────┬──────┘
+            Yes  │  No
+            │    │
+            ▼    ▼
+         Proceed  ──► Day 0 Update Flow
 ```
 
 ### Update Flow
@@ -85,11 +85,10 @@ When an update is available:
    ```
 
 2. **If accepted:**
-   - Record the original command line arguments.
-   - Initiate the self-update via `winget upgrade Microsoft.AppInstaller --silent`.
+   - Record the original command line arguments via `winget resume` infrastructure.
+   - Initiate the self-update via `winget upgrade Microsoft.AppInstaller --source winget --silent`.
    - The current process exits (WinGet cannot update itself while running).
-   - A watchdog process (or scheduled task) monitors the update completion.
-   - On completion, launch a **new terminal window** with the original command, using `--wait` to prevent the window from closing on completion.
+   - The update orchestrator (see [Self-Update Mechanism](#self-update-mechanism)) handles completion and relaunch.
 
 3. **If declined:**
    - Record the dismissal (increment counter in local state).
@@ -251,7 +250,8 @@ No manifest schema changes required. The version signal lives in source metadata
 │  Updating is recommended to access all current features.        │
 │  WinGet will restart and run your command after updating.       │
 │                                                                 │
-│  [U]pdate now    [S]kip this time    [D]on't ask again          │
+│  Update now    Skip this time    Don't ask again             │
+│  [U]            [S]               [D]                         │
 └─────────────────────────────────────────────────────────────────┘
 ```
 
--- a/doc/specs/#6298 - Day 0 Update.md
+++ b/doc/specs/#6298 - Day 0 Update.md
@@ -1,0 +1,353 @@
+---
+author: Demitrius Nelon denelon, GitHub Copilot Copilot
+created on: 2026-06-18
+last updated: 2026-06-18
+issue id: 6298
+---
+
+# Day 0 Update
+
+For [#6298](https://github.com/microsoft/winget-cli/issues/6298).
+
+## Abstract
+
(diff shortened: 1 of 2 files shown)


doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,353 @@
...
+   - Record the original command line arguments.
+   - Initiate the self-update via `winget upgrade Microsoft.AppInstaller --silent`.
+   - The current process exits (WinGet cannot update itself while running).
+   - A watchdog process (or scheduled task) monitors the update completion.
+   - On completion, launch a **new terminal window** with the original command, using `--wait` to prevent the window from closing on completion.
+
+3. **If declined:**
+   - Record the dismissal (increment counter in local state).
+   - Proceed with the original command on the current version.
+   - If the command fails because the client is too old (e.g., unrecognized subcommand), show a targeted message: "This feature requires WinGet 1.29+. Run `winget upgrade Microsoft.AppInstaller` to update."
```
Trenly, line 96: How would the cli know which commands are supported in newer features if it's old?

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,353 @@
...
+
+4. **If non-interactive** (`--disable-interactivity`, COM API, Configuration):
+   - Do not prompt. Proceed with current version.
+   - Emit a warning to stderr: "WinGet update available (1.30.1234.0). Run `winget upgrade Microsoft.AppInstaller` to update."
```
Trenly, line 101: Must respect `--nowarn`

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,353 @@
...
+   - Triggers the Microsoft Store update or direct MSIX install.
+   - On success: launches a new `conhost`/`wt` window with the saved command + `--wait`.
+   - On failure: launches a new window with an error message and the original command for manual retry.
+
+### Relaunch Behavior
+
+The relaunch uses `--wait` to keep the new terminal window open after the command completes:
+
+```
+wt.exe --wait -- winget configure myconfig.yaml
```
Trenly: Why not `winget resume` (currently experimental) ?

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+If Windows Terminal is not available, fall back to:
```
Trenly: why route through wt.exe or cmd.exe instead of using ShellExecute to call the app execution alias, as if the user were calling `winget configure myconfig.yaml --wait` from the run dialog?

doc/specs/#6298 - Day 0 Update.md
denelon: My current thinking is that when we check for the newer version, we'll have that data. Then when we're dealing with a newer index or schema version, we can make some assumptions about stuff that the client doesn't understand, and it can be associated with the perceived to be available latest version. It may not get "all edge cases", but it's my current best assumption.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,353 @@
...
+4. **If non-interactive** (`--disable-interactivity`, COM API, Configuration):
+   - Do not prompt. Proceed with current version.
+   - Emit a warning to stderr: "WinGet update available (1.30.1234.0). Run `winget upgrade Microsoft.AppInstaller` to update."
+
+### Self-Update Mechanism
+
+The update uses the existing `winget upgrade` infrastructure targeting `Microsoft.AppInstaller`:
+
+```
+winget upgrade --id Microsoft.AppInstaller --silent --accept-package-agreements --accept-source-agreements
```
mdanish-kh: Any concerns around passing `--accept-package-agreements`, `--accept-source-agreements` with regards to explicitly accepting the agreements for the user? Should we remove them?

- https://github.com/microsoft/winget-pkgs/issues/131328

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+issue id: 6298
+---
+
+# Day 0 Update
+
+For [#6298](https://github.com/microsoft/winget-cli/issues/6298).
+
+## Abstract
+
+Enable WinGet to detect that it is outdated and self-update on first meaningful command execution. When the first command requiring the source index runs, the client checks a version signal in the preIndexed package source, determines whether a newer stable release exists, and — with user consent — updates itself and relaunches the original command. This eliminates the "chicken and egg" problem where WinGet must already be current to use new functionality (e.g., `winget configure`).
```
Trenly: Another thought - what if the preIndexed package source is disallowed by GPO, or removed by the user before the first command that would use it?  Should this perhaps be a different endpoint that gives the information?

denelon: I'd expect enterprise customers to push out new versions to their devices. If they are blocking the default source, there might be another way to reason about updating ourself, but I haven't gone through all the thinking yet.

Trenly: Yeah, it's another 🐔 / 🥚 - especially if the auto udpate is intended to use the default source to update itself.  There's also more thought needed to ensure that the version referenced in the index actually exists as a package in the winget source

florelis:
Some thoughts about the timeline...

Assuming that Windows' release schedule stays as it has been for the past couple of years, I think the cutoff date for being in an 2_H2 build would be around March-April, and for a 2_H1 would be around September-October the year before. If we assume further that we keep winget's release schedule the same, out options would be..
* 1.29 (Jun '26) -> 27H1 (already too late)
* 1.30 (Oct '26) -> 27H2
* 1.31 (Feb '27) -> 27H2
* 1.32 (Jun '27) -> 28H1
* 1.33 (Oct '27) -> 28H2

And who knows if they're even going to keep that pattern...

I think ideally we would want to have this complete in the client for 1.30, so we can update the service and start getting the version metadata in the source by 1.31, and then we'd have some time to validate it works before baking it into the OS. And we'd be looking at it being included in 27H2. Anything later than that, we'd be looking at 28H1 or 28H2 (since they seem to skip H1s).

This all also depends on the OS _not_ implementing a mechanism for updating preinstalled Store apps while installing/updating Windows. If they do that before the version we'd include this on, then it all becomes moot.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+```json
+{
+  "latestStableClientVersion": "1.30.1234.0",
+  "minimumRecommendedClientVersion": "1.28.0.0",
+  "day0UpdatePackageId": "Microsoft.AppInstaller"
+}
+```
+
+- `latestStableClientVersion` — The newest stable release of the WinGet client.
+- `minimumRecommendedClientVersion` — The minimum version that supports all current source features (e.g., new schema fields, API endpoints). Clients below this threshold get a stronger update prompt.
```
florelis, line 38: I think we need another field for the minimum working version. It is a different message if we say "you have an outdated version, you can update for a better experience" than if we say "you have an outdated version that doesn't work anymore, you need to update"

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+
+### Version Signal in PreIndexed Source
+
+The preIndexed package source already carries metadata about available packages. A new metadata field advertises the latest stable client version:
+
+```json
+{
+  "latestStableClientVersion": "1.30.1234.0",
+  "minimumRecommendedClientVersion": "1.28.0.0",
+  "day0UpdatePackageId": "Microsoft.AppInstaller"
```
florelis, line 33: Maybe this is just me, but "day 0" makes me think of zero-day exploits. I would prefer if we used another term.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+This metadata is updated as part of the source publication pipeline whenever a new stable client ships.
+
+### Detection Flow
+
+```
+User runs: winget configure myconfig.yaml
+         │
+         ▼
+┌───────────────────────────────────┐
+│ Source index access required       │
```
florelis, line 50: `winget configure` now requires installing the source?

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+1. **Inform the user** in the current terminal:
+   ```
+   A newer version of WinGet is available (1.30.1234.0 → current: 1.28.567.0).
+   Updating is recommended to access all features.
+   
+   Update now? [Y/n]:
+   ```
+
+2. **If accepted:**
+   - Record the original command line arguments via `winget resume` infrastructure.
```
florelis, line 88: I have no idea how `winget resume` is implemented, but it is intended for reboots, right? If so, I don't think that would be the right tool. I would expect a resume across a reboot to require something like a RunOnce key, and that doesn't seem appropriate here.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+   - Initiate the self-update via `winget upgrade Microsoft.AppInstaller --source winget --silent`.
```
florelis, line 89: nit: maybe that's an equivalent of what we'd do, but I doubt we would want to invoke the command line like this.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+The update targets `Microsoft.AppInstaller` from the winget default source:
+
+```
+winget upgrade --id Microsoft.AppInstaller --source winget --silent
+```
+
+Because WinGet cannot update its own running process:
+
+1. WinGet records the original command line and context using the resume infrastructure (`winget resume`).
+2. WinGet spawns a detached update orchestrator process (a small helper EXE bundled with the client).
```
florelis, line 115: I think we would have to copy this small .exe to a separate location, since the helper itself would also block the installation if it is running.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+
+### Relaunch Behavior
+
+The relaunch leverages `winget resume` (currently experimental) to restore the original command context:
+
+```
+winget resume --resume-id <saved-id>
+```
+
+The resume infrastructure records the full command line, working directory, and any relevant state. After a successful self-update, the orchestrator invokes `winget resume` via `ShellExecute` on the app execution alias. This delegates terminal window creation to the operating system's default handler, avoiding any dependency on a specific terminal application.
```
florelis, line 130: > any relevant state

Should it record things like settings?

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+
+If `winget resume` is unavailable (older client somehow reached this path), fall back to `ShellExecute` with the original command line:
```
florelis, line 132: This isn't possible. An old client without resume would also not have any code we add for this auto-update.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+
+```
+ShellExecute(NULL, "open", "winget", "configure myconfig.yaml --wait", workingDir, SW_SHOWNORMAL)
+```
+
+### `winget configure` Special Case
+
+When the first command is `winget configure` and the current client doesn't recognize it (pre-Configuration builds):
```
florelis, line 140: Not possible for us to add code for this in old builds without configuration

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+- This eliminates the need for `winget configure enable` entirely on fresh installs — the user gets the latest client which has Configuration enabled by default.
+
+### Session Throttling
+
+To avoid pestering users:
+
+- Day 0 Update check runs **once per session** (once per process lifetime).
+- After dismissal, suppress for 24 hours (configurable).
+- After N dismissals (default: 3), suppress permanently until the setting is reset or a major version ships.
+- The check adds negligible latency (metadata is already being fetched as part of source index access).
```
florelis, line 153: Not if we're doing something that does not use the source index.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+
+- The Day 0 Update check happens *before* command parsing fails.
+- If `latestStableClientVersion` indicates the feature is available in the new version, the update prompt specifically mentions: "The `configure` command requires WinGet 1.29+. Update now to proceed."
+- This eliminates the need for `winget configure enable` entirely on fresh installs — the user gets the latest client which has Configuration enabled by default.
+
+### Session Throttling
+
+To avoid pestering users:
+
+- Day 0 Update check runs **once per session** (once per process lifetime).
```
florelis, line 150: What is a session? Just a  process?

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+- After N dismissals (default: 3), suppress permanently until the setting is reset or a major version ships.
+- The check adds negligible latency (metadata is already being fetched as part of source index access).
+
+### Group Policy Controls
+
+| Policy | Type | Default | Description |
+|--------|------|---------|-------------|
+| `EnableDay0Update` | Bool | Enabled | Master toggle for Day 0 Update |
+| `Day0UpdateBehavior` | Enum | Prompt | `Prompt` / `AutoUpdate` / `NotifyOnly` |
+| `Day0UpdateSource` | Enum | Store | `Store` / `Direct` (direct MSIX from CDN) |
```
florelis, line 161: Download MSIX from CDN sounds expensive.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+
+```idl
+interface IPackageManagerStatus
+{
+    Boolean IsClientUpdateAvailable { get; };
+    String LatestStableClientVersion { get; };
+    String CurrentClientVersion { get; };
+    String MinimumRecommendedClientVersion { get; };
+}
+```
```
florelis, line 206: Would we add a function to perform the update? Otherwise, we'd have to ask them to update "Microsoft.AppInstaller" from the API, and deal with the process blocking the update. Then there is the issue that this would kill the service, and one would have to reconnect.

doc/specs/#6298 - Day 0 Update.md
```diff
@@ -0,0 +1,356 @@
...
+- **Process handoff complexity**: The "exit → update → relaunch" flow has edge cases (terminal closed by user mid-update, update stuck, insufficient permissions).
+- **Store update latency**: Microsoft Store updates can be slow or queued. Direct MSIX may be more reliable for Day 0 scenarios.
+- **Admin vs. user context**: Self-update may require elevation depending on the install method. The orchestrator must handle both contexts gracefully.
+- **Circular dependency**: If the source index itself requires a newer client to parse, the version check can't happen. Mitigation: the `latestStableClientVersion` field uses a fixed, simple schema that all client versions can read.
+- **Configuration-as-Code**: Automated `winget configure` flows in CI/CD should never be interrupted. Non-interactive mode and `--no-update-check` address this, but documentation must be clear.
+
+## Future Considerations
+
+- **Staged rollout**: Day 0 Update could integrate with the existing feature toggle system for gradual rollout.
+- **Delta updates**: Rather than full package replacement, future iterations could support incremental/delta client updates for faster Day 0 flow.
```
florelis, line 344: If we update from the Store, we should be getting this for free

doc/specs/#6298 - Day 0 Update.md
Trenly: When the resume experimental feature is enabled, the client creates checkpoints. The resume command should just load that checkpoint https://github.com/microsoft/winget-cli/pull/3508. I also don't believe that in the restart flow registry keys are used, instead the application is registered for restart with Win32 APIs https://github.com/microsoft/winget-cli/pull/3858.  

There may be some minor work to ensure the checkpoint is there for when WinGet aborts it's own process and re-schedules itself, but I think it's absolutely the right use case for the resume - as it's effectively a reboot, just in this case it's a process reboot and not a machine reboot

doc/specs/#6298 - Day 0 Update.md
Trenly: Also, isn't winget able to update itself using deferred registration as an msix ?

doc/specs/#6298 - Day 0 Update.md
denelon: I think there is going to be a tough balance between what "works" in some form or another, and the intent of getting the device quickly to the latest stable version of the client. This "chicken and egg" problem has been there for quite some time. The expired certificates for the "msstore" source, and support for dependencies, and support for configuration have all three been factors leading into where we are today.

If for example, we were to decide to force a 2.x version of the schema and that meant the device "must" update I'm not sure what that would look like. The notion of 1.x is deprecated, however seems like a good idea, and we could keep that around as a logical "minimal" source without manifests in it could be a way to help a future migration succeed faster/better.

doc/specs/#6298 - Day 0 Update.md
denelon: I'm open to suggestions :)

doc/specs/#6298 - Day 0 Update.md
denelon: "Most" configuration files I've seen involve installing packages which would require having a "source".

doc/specs/#6298 - Day 0 Update.md
denelon: That's why the PR is still in Draft :) The robot is making all kinds of assumptions. I'm very appreciative of all the feedback to whip this into shape. Once we get "closer" to what's reasonable, I'll condense some of those learnings into copilot-instructions files to help future specs to become "better".

doc/specs/#6298 - Day 0 Update.md
denelon: Yeah, more hallucination here. We're going to have to get some minimum viable version of the self-update mechanism in place and out in the wild before we get the benefits of all this work, but we need to figure out how to get out beyond these chicken and egg problems where the client on the device is essentially non-functional for some reason(s).

PR #6310, opened 2026-06-21
Add persistent output locale override setting

### LINKED ISSUE #422: Allow the user to override winget's locale setting
*I moved this from a comment to its own issue because it seems to be unique enough to warrant a dedicated feature request.*

I'd like to override the winget locale to change its output language. My OS locale is not English so localized tools make it more difficult to post tutorials and posts for an international audience on my tech blog. This would also help you as developers to get output for bug reports in a language that you can understand.

I'd be really happy if you added an environment variable that would enable me to override the winget locale (like `LANG=en_US.UTF-8 <command>` or `LC_MESSAGES en_US.UTF-8 <command>` in Linux):

<https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html>

> For example, assume you are a Swedish user in Spain, and you want your programs to handle numbers and dates according to Spanish conventions, and only the messages should be in Swedish. [...] [S]et the LANG variable to es_ES.UTF-8 and the LC_MESSAGES variable to sv_SE.UTF-8 [...].
> 
> LC_ALL is an environment variable that overrides all of these. It is typically used in scripts that run particular programs. For example, configure scripts generated by GNU autoconf use LC_ALL to make sure that the configuration tests don’t operate in locale dependent ways.

Other options could include a winget command line parameter though that would need to be provided on every winget call.

A third option is to store it in the proposed JSON configuration (#416) to have a means of persisting the locale override.

Trenly:
## 📖 Description
Add a persistent `output.locale` user setting that allows overriding winget interface string resolution by BCP47 tag.

Current implementation details:
- Reads `.output.locale` at CLI startup and applies the override only when the setting is non-empty.
- Applies the override through `Resource::SetLanguageOverride`, which sets the MRT `Language` qualifier for winget resource loading.
- Treats override failures as non-fatal (logs warning and continues without the override).
- Uses `wil::srwlock` in the shared resource loader for synchronized read/write access when resolving strings and updating locale override.
- Does not apply this setting in server initialization.

This PR also updates settings schema/documentation and local package resource fallback handling for localized `winget.resw` files.

## 🔗 References
Resolves #422

## 🤖 Copilot Assistance
This PR includes changes authored with GitHub Copilot assistance.

## 🔍 Validation
- Added/updated unit tests for `OutputLocale` setting parsing/normalization behavior.
- Added E2E coverage verifying help output changes when `output.locale` changes.

## ✅ Checklist
- [x] Linked to an issue

## 📋 Issue Type
- [ ] Bug fix
- [x] Feature
- [ ] Task


src/AppInstallerCLIPackage/AppInstallerCLIPackage.wapproj
Trenly, line 1: These changes to the wapproj were required for me to be able to verify the functionality locally with `wingetdev -?`

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -561,6 +561,16 @@ namespace AppInstaller::Settings
 
             return {};
         }
+
+        WINGET_VALIDATE_SIGNATURE(OutputLocale)
+        {
+            if (Locale::IsWellFormedBcp47Tag(value))
```
florelis: Should we also be checking that it is one of our supported locales?

src/AppInstallerCLITests/UserSettings.cpp
```diff
@@ -925,6 +925,46 @@ TEST_CASE("SettingOutputSortDirection", "[settings]")
     }
 }
 
+TEST_CASE("SettingOutputLocale", "[settings]")
```
florelis, line 928: Add a case for "valid, unsupported locale"

src/AppInstallerCLICore/Core.cpp
```diff
@@ -89,6 +102,7 @@ namespace AppInstaller::CLI
         std::signal(SIGABRT, abort_signal_handler);
 
         init_apartment();
+        auto outputLocaleOverride = ApplyOutputLocaleOverride();
```
JohnMcPMS: This seems very early to be doing this, I would suggest doing it when you do the check later on.

src/AppInstallerCLICore/Core.cpp
```diff
@@ -212,6 +231,8 @@ namespace AppInstaller::CLI
 
     void ServerInitialize()
     {
+        auto outputLocaleOverride = ApplyOutputLocaleOverride();
```
JohnMcPMS: I don't think the server needs to concern itself with the output locale.

src/AppInstallerCLICore/Core.cpp
```diff
@@ -227,10 +248,16 @@ namespace AppInstaller::CLI
...
         AppInstaller::CLI::Execution::COMContext::SetLoggers();
+        if (outputLocaleOverride)
+        {
+            AICLI_LOG(CLI, Info, << "Applied output locale override from settings: " << outputLocaleOverride.value());
+        }
     }
 
     void InProcInitialize()
     {
+        auto outputLocaleOverride = ApplyOutputLocaleOverride();
```
JohnMcPMS: And we definitely don't want to change the locale on any process we are loaded in to.

doc/windows/package-manager/winget/settings.md
```diff
@@ -98,6 +98,23 @@ The `locale` behavior affects the choice of installer based on installer locale.
     },
 ```
 
+### Output
+
+The `output` settings affect winget interface output behavior.
+
+#### locale
```
florelis, line 105: It is a bit of a stretch, but would it make sense to put this in the `visual` settings?

src/AppInstallerCLICore/Core.cpp
```diff
@@ -88,8 +101,7 @@ namespace AppInstaller::CLI
 
         std::signal(SIGABRT, abort_signal_handler);
 
-        init_apartment();
-
+        init_apartment()
```
florelis: Missing ;

doc/windows/package-manager/winget/settings.md
Trenly: I considered that, but since it isn't strictly visual, I thought `output` would be better here, especially if there's ever any plans for structured data settings

src/AppInstallerCommonCore/UserSettings.cpp
Trenly: I was thinking probably not - if the value isn't one of the supported locales, then the CLI will simply fall back to what it does today, which is use the next best locale based on the available locales.

I could potentially see putting a warning in the logs, or even outputting the warning to the user - but it would be my preference to file that as a separate feature request rather than forcing the settings value to be a supported locale since there could exist a world where supported locales are added or removed in the future and it would be nice to have this not be a surface that could lead to a breaking change.

florelis: Adding locales would not be a breaking change. Removing them would be one, but I don't think it will ever happen, and if it does we can just return gracefully here with a warning, as if they had no preference.

I think it would be pretty confusing to be told "you can put any locale here", and then going to use the tool and now it's in a weird language that the system decided for some reason was close enough. Especially if they are using a display language that is supported, but we override it with a language that isn't supported and they end up getting a third unrelated language.

It would also make things easier because we could add the specific supported values to the settings schema for autocomplete.

Trenly: Updated

src/AppInstallerCommonCore/Public/winget/UserSettings.h
```diff
@@ -66,6 +66,26 @@ namespace AppInstaller::Settings
...
+        KoKR,
+        PtBR,
+        RuRU,
+        ZhCN,
+        ZhTW,
+    };
+
+    // Converts OutputLocale to its locale tag value. Returns empty for OutputLocale::Unset.
+    std::string_view OutputLocaleToString(OutputLocale locale);
+
```
florelis: This is overly complicated. Just make a function `bool IsSupportedLocale(string localeTag)` that checks if it is in a fixed list

src/AppInstallerCLICore/Core.cpp
```diff
@@ -78,6 +79,24 @@ namespace AppInstaller::CLI
...
             ShutdownMonitoring::ServerShutdownSynchronization::AddComponent(main);
         }
+
+        std::string ApplyOutputLocaleOverride()
+        {
+            std::string localeTag{ Settings::OutputLocaleToString(Settings::User().Get<Settings::Setting::OutputLocale>()) };
+
+            try
+            {
+                // Always apply the setting value, including empty string, to clear any prior override.
```
florelis: I don't think there is a way to have a prior override?

florelis (changes requested):

src/AppInstallerCLICore/Core.cpp
Trenly: During testing, the set locale persisted even when the setting was cleared unless actively unset. Even across restarts of the terminal

src/AppInstallerCommonCore/Public/winget/UserSettings.h
Trenly: Using an enum seemed more robust to me than just checking in a list, especially for wiring up a future `--output-locale` argument and COM API

I understand its added complexity, but the explicit handling seemed worth the tradeoff in my opinion.

src/AppInstallerCLICore/Core.cpp
florelis: Ah, I see. I also found it in the docs:
> The PrimaryLanguageOverride setting is persisted between sessions.

Please expand the comment to mention that the setting applies across sessions and is not only for the current process.

Also, does this override the language for other executables in the same app package too?

src/AppInstallerCommonCore/Public/winget/UserSettings.h
florelis: The only thing we do is: read the json -> convert the string to enum -> convert back to string -> pass string to the API. I don't think that using an enum adds any value in this. And, we have the precedent that functions dealing with locales use strings too; like the `PrimaryLanguageOverride'

src/AppInstallerCLICore/Core.cpp
Trenly: > Also, does this override the language for other executables in the same app package too?

According to what I could find, it is package-scoped

florelis: In that case, I think we'll need to pair it with an internal change to App Installer to always reset the locale on startup. It wouldn't make sense to have the setting from winget affect the UI for installing an MSIX. It would be nice if winget could undo the change on exit, but I think that wouldn't work in case of a crash or something like that. I'll have to try this change with App Installer to see what happens.

src/AppInstallerCommonCore/UserSettings.cpp
```diff
@@ -235,6 +237,37 @@ namespace AppInstaller::Settings
         return std::nullopt;
     }
 
+    namespace
+    {
```
florelis: I think we already have an anonymous namespace at the top

florelis (dismissed):
LGTM, but I'd want to hold off on merging until we have a plan for what to do with App Installer.

Trenly:
@florelis - I get the same E2E test failures from my last checkout on `master`

* Edit:  Related to #6354


https://github.com/user-attachments/assets/e9e2f1bf-23dd-4dfe-8e20-f0d60c7dd7fa

src/AppInstallerCLICore/Core.cpp
```diff
@@ -78,6 +79,25 @@ namespace AppInstaller::CLI
...
+        std::string ApplyOutputLocaleOverride()
+        {
+            std::string localeTag{ Settings::User().Get<Settings::Setting::OutputLocale>() };
+
+            try
+            {
+                // Always apply the setting value, including empty string, to clear a persisted override from
+                // previous sessions. PrimaryLanguageOverride is package-scoped and persists across sessions.
+                winrt::Windows::Globalization::ApplicationLanguages::PrimaryLanguageOverride(Utility::ConvertToUTF16(localeTag));
+            }
```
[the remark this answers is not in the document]
florelis: I don't think this applies, as we don't use qualifiers anywhere AFAIK

florelis (dismissed):

florelis:
Looks like we will need to reset the override in App Installer. Internal PR number for that: 16173490
I won't merge this until that other change is in place.

florelis:
Question from internal discussion:
Instead of overriding the primary language like this, would it be possible to just add the locale qualifier when creating the resource loader? That would avoid setting the locale package-wide, and we wouldn't need to make any changes to App Installer.

src/AppInstallerSharedLib/Resources.cpp
```diff
@@ -130,6 +180,11 @@ namespace AppInstaller
                 }
             }
         };
+
+        bool SetLanguageOverride(std::string_view localeTag)
+        {
+            return Loader::Instance().SetLanguageOverride(Utility::ConvertToUTF16(localeTag));
```
florelis, line 191: Nit: Assuming this is done before any localized strings are fetched, this is going to create a loader with default options, and then immediately overwrite it with a new one that uses the locale. We could make it so that we only needed to create it once, but that's probably too small a gain to care.

florelis (dismissed):

Trenly:
Grrr, pesky build failures

Trenly:
@florelis - I've merged the latest changes from `master` to see if that resolves the ADO build issue despite it building locally without them. I've verified it builds locally after the merge:

image

Trenly:
## Root Cause Analysis

> [!NOTE]
> This comment was authored by GitHub Copilot.

**Root cause:** `ApplyOutputLocaleOverride()` unconditionally accessed `Loader::Instance()` at startup for *every* command. The `WinGetGenerateDSCv3Manifests` build step runs `winget.exe dscv3 --manifest` as an **unpackaged process** (before MSIX packaging), and the refactored `Loader()` constructor now calls `CreateLoader({})` which calls `ResourceContext::ResetGlobalQualifierValues()` before the `ResourceLoader()` probe.

The original code relied on `ResourceLoader()` throwing a catchable `hresult_error` when `resource.pri` is not found (unpackaged context) — this prevented execution from ever reaching `ResourceLoader::GetForViewIndependentUse`, which **fast-fails with an uncatchable crash** (`0xC0000409`) when no debugger is attached. Calling `ResetGlobalQualifierValues` first changed the WinRT resource state and broke this safety guard.

## Fix

Two changes applied in commit 25c246a8:

1. **`Resources.cpp`**: Restored the original constructor initialization path — no `ResetGlobalQualifierValues` call, preserving the `ResourceLoader()` probe guard
2. **`Core.cpp`**: Guard `ApplyOutputLocaleOverride()` to return early when the locale setting is empty, avoiding `Loader::Instance()` initialization at startup for commands that don't need UI output (like `dscv3 --manifest`)

src/AppInstallerSharedLib/Resources.cpp
```diff
@@ -106,20 +113,69 @@ namespace AppInstaller
...
+                    AICLI_LOG(CLI, Error, << "Failure applying resource locale override (" << Utility::ConvertToUTF8(localeTag) << ") with error: " << hre.code());
+                }
+
+                return false;
+            }
+
         private:
             winrt::Windows::ApplicationModel::Resources::ResourceLoader m_wingetLoader;
+            winrt::hstring m_localeOverride;
+            mutable std::mutex m_lock;
```
JohnMcPMS: A `wil::srwlock` would probably be far more performant.

src/AppInstallerCLICore/Core.cpp
```diff
@@ -78,6 +79,24 @@ namespace AppInstaller::CLI
...
+        std::string ApplyOutputLocaleOverride()
+        {
+            std::string localeTag{ Settings::User().Get<Settings::Setting::OutputLocale>() };
+
+            if (localeTag.empty())
+            {
+                return {};
+            }
+
+            if (!AppInstaller::Resource::SetLanguageOverride(localeTag))
```
JohnMcPMS, line 92: This can throw; do we think it important enough to respect the target locale that we should terminate the winget.exe process if we cannot?

Trenly: Right now the throw should be caught internally to the function and result in the `false` return, walking through the warning path here. My thought was that this should be best-effort and not breaking. This would most closely match the behavior today where if the Locale coming from the OS doesn't map to an avaliable resource file it falls back to whatever the resource loader deems is the best fit. 

I'm not opposed to making it a hard failure, but without knowing how often it would or could throw, just makes me a bit uneasy. Is this a decision that needs to be made now? Could we do something in the future like adding a telemetry event sampling language overrides as a `what-if` to better understand the potential impact of making it a terminating failure?

JohnMcPMS (approved):

PR #6313, opened 2026-06-21
Refactor SourceType handling to enum

### LINKED ISSUE #4463: Move SourceType to an Enum

Currently source types are all either `std::string` or `std::string_view` in the code. To me, it seems like this isn't a very good way of ensuring the source types are well managed

Refactor the code to use a `SourceType` enum with a method for converting to `SourceType` from a string or string_view, and a method for converting `SourceType` to string view. This can then be used for stronger argument validation on the `source add` command

Trenly:
## 📖 Description
This refactor moves internal source-type handling to Repository::SourceType enum paths across repository, CLI, and COM layers, while preserving string conversion only at external boundaries (CLI/JSON/GP/COM input-output). It also removes accidental string-based source construction paths and updates source-related tests/fixtures to use canonical source type values.

Key changes include:
- SourceDetails.Type and internal source factory routing now use SourceType.
- ISourceFactory::GetForType now takes SourceType.
- Source constructor string-type overload removal and default-type string helper removal at call sites.
- Source parsing/serialization boundaries explicitly convert using TryConvertToSourceTypeEnum/SourceTypeEnumToString.
- Source-related tests updated for enum semantics, including legacy tombstone/override source-list parsing compatibility.

Copilot assisted with implementation.

## 🔗 References
Resolves #4463

## 🔍 Validation
- Executed source-focused test subset:
  - ./src/x64/Debug/AppInstallerCLITests/AppInstallerCLITests.exe -v high -d yes '*Source*'

## ✅ Checklist
- [x] Linked to an issue

## 📋 Issue Type
- [ ] Bug fix
- [ ] Feature
- [x] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6313)


doc/ReleaseNotes.md
```diff
@@ -4,4 +4,5 @@ Nothing yet.
 
 ## Bug Fixes
 
+* Refactored source type handling to use a `SourceType` enum with centralized string conversion, and added stronger `source add` type validation (#4463).
```
florelis: I'm not sure we need to mention this here. An internal refactor doesn't seem too relevant to someone asking "what's new in this release?"

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -61,6 +61,26 @@ namespace AppInstaller::Repository
...
 
+    // Defines known source types.
+    enum class SourceType
+    {
+        PreIndexedPackage,
+        Rest,
+        PredefinedInstalled,
+        PredefinedWriteable,
+        PackageTracking,
+        ConfigurableTest,
```
florelis, line 72: Should this be gated by the test hooks being enabled?

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -61,6 +61,26 @@ namespace AppInstaller::Repository
...
+    };
+
+    // Converts a source type string to the corresponding SourceType enum.
+    SourceType ConvertToSourceTypeEnum(std::string_view sourceType);
+
+    // Attempts to convert a source type string to the corresponding SourceType enum.
+    std::optional<SourceType> TryConvertToSourceTypeEnum(std::string_view sourceType);
+
+    // Converts a SourceType enum to the corresponding canonical string.
+    std::string_view SourceTypeEnumToString(SourceType sourceType);
```
florelis: Nit: Is this the naming scheme we usually follow for these enum<->string conversions? I see the one below is just `ToString`

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -120,7 +140,9 @@ namespace AppInstaller::Repository
         std::string Name;
 
         // The type of the source.
-        std::string Type;
+        // Defaults to PreIndexedPackage if not specified, moved here from call sites during migration from
+        // String type to Enum type to avoid conversion in multiple places.
```
florelis: Is there any reason we'd need this historical note?

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
```diff
@@ -61,6 +61,26 @@ namespace AppInstaller::Repository
...
+        PredefinedWriteable,
+        PackageTracking,
+        ConfigurableTest,
+    };
+
+    // Converts a source type string to the corresponding SourceType enum.
+    SourceType ConvertToSourceTypeEnum(std::string_view sourceType);
+
+    // Attempts to convert a source type string to the corresponding SourceType enum.
+    std::optional<SourceType> TryConvertToSourceTypeEnum(std::string_view sourceType);
```
florelis, line 79: Nit: Do we need both? Do we usually have both?

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -276,46 +276,106 @@ namespace AppInstaller::Repository
         };
     }
 
-    std::unique_ptr<ISourceFactory> ISourceFactory::GetForType(std::string_view type)
+    std::unique_ptr<ISourceFactory> ISourceFactory::GetForType(SourceType type)
     {
 #ifndef AICLI_DISABLE_TEST_HOOKS
         // Tests can ensure case matching
```
florelis: Remove outdated comment

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -276,46 +276,106 @@ namespace AppInstaller::Repository
...
+        case SourceType::PredefinedWriteable:
+            return Microsoft::PredefinedWriteableSourceFactory::Create();
+        case SourceType::PackageTracking:
+            return PackageTrackingCatalogSourceFactory::Create();
+#ifndef AICLI_DISABLE_TEST_HOOKS
+        case SourceType::ConfigurableTest:
             return Microsoft::ConfigurableTestSourceFactory::Create();
-        }
+#else
+        case SourceType::ConfigurableTest:
```
florelis: Duplicated

src/AppInstallerRepositoryCore/RepositorySource.cpp
```diff
@@ -907,14 +966,8 @@ namespace AppInstaller::Repository
...
-        // If the source type is empty, use a default.
-        // AddSourceForDetails will also check for empty, but we need the actual type before that for validation.
-        if (sourceDetails.Type.empty())
-        {
-            sourceDetails.Type = GetDefaultSourceType();
-        }
-
-        AICLI_LOG(Repo, Info, << "Adding source: Name[" << sourceDetails.Name << "], Type[" << sourceDetails.Type << "], Arg[" << sourceDetails.Arg << "]");
+        // If the source type was empty, the constructor of sourceDetails will have defaulted it
+        AICLI_LOG(Repo, Info, << "Adding source: Name[" << sourceDetails.Name << "], Type[" << SourceTypeEnumToString(sourceDetails.Type) << "], Arg[" << sourceDetails.Arg << "]");
```
florelis: I wonder if it would be worth it to keep the semantics of "type can be empty and that means default" to simplify the changes in logic for this PR. I'm not sure how much we win changing the meaning now, and whether we may be missing some place.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
```diff
@@ -3,19 +3,18 @@
...
 namespace winrt::Microsoft::Management::Deployment
 {
     namespace ProgressSinkFactory
     {
         std::shared_ptr<AppInstaller::IProgressSink> CreatePackageCatalogProgressSink(std::string sourceType, std::function<void(double)> progressReporter, bool removeOperation)
         {
-            if (sourceType.empty()
-                || Utility::CaseInsensitiveEquals( Repository::Microsoft::PredefinedInstalledSourceFactory::Type(), sourceType))
+            auto sourceTypeEnum = ::AppInstaller::Repository::TryConvertToSourceTypeEnum(sourceType).value_or(::AppInstaller::Repository::Source::GetDefaultSourceType());
+            if (sourceTypeEnum == ::AppInstaller::Repository::SourceType::PreIndexedPackage || sourceTypeEnum == ::AppInstaller::Repository::SourceType::PredefinedInstalled)
```
florelis: I'm not sure what this does, but the logic changed. If you get the 'preindexed package' as input, you will execute this conditional.

florelis (commented):
I haven't finished reviewing. My main comment so far is: Why move where in the code empty types are handled? We could have kept it the same, and then you wouldn't have as many logic changes to verify.

doc/ReleaseNotes.md
Trenly: Fair - 🤖  was just following default instructions, will remove

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
Trenly: I opted to leave the enum alone and have the test hooks handle the implementation / backend so that enum indexes don't change between debug / test hook builds and release builds

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
Trenly: Probably not - but I thought it might help a review

src/AppInstallerRepositoryCore/Public/winget/RepositorySource.h
Trenly: We usually don't have both, and instead have `<Enum>::Unknown` as the fallback. The reason for having both here was to avoid needing the `Unknown` type in the enum

src/AppInstallerRepositoryCore/RepositorySource.cpp
Trenly: That's actually why I added the comment - because I wanted it to be clear that by movng to a strongly typed enum here we would be losing some of that nuance of "Type can be empty"

I'm sure that I could find a way to keep that nuance, but semantically I'd rather always have an explicit type, that way there isn't any potential confusion about what empty actually means. If empty is passed into the constructor it gets defaulted, and IMO that makes all the downstream logic simpler because callsites don't have to worry whether it is empty or not, it becomes a case they don't need to worry about.

src/Microsoft.Management.Deployment/PackageCatalogProgress.cpp
Trenly: This is an artifact of not allowing an empty source type. If `sourceType` were empty in this new flow, `TryConvertToSourceTypeEnum` is an empty optional and, according to the old logic, means 'default'.  This is handled by the `.value_or` setting it to the default.  

Then, because the source type could be default or predefined installed, I check both.

In either case, I believe that the PreIndexed package is intended to follow into this conditional as it is setting the weights for the progress bar which depend on downloading and installing the index, which are both part of the pre-indexed package flow. Excluding it here would cause it to not have the same behavior

Trenly: After some additional investigation there is no reason this needs to be, or actually should be, handled as a string. The value is being roundtripped from call sites converting to a string before passing it here where it's converted right back into an enum.  I've updated this so the value is passed as an enum value directly into the sink which allows the logic to more closely mirror the original and be much more readable

JohnMcPMS:
This is more of a comment for the issue, but since we have a PR:

I'm not seeing the rationale for this change.  We haven't had any issues that I'm aware of, and externally it isn't visible.  Moving from what is effectively a very large number space (strings) to a smaller one doesn't seem like it is going to make things any better.  In fact, it makes off by one errors much more likely to cause a problem beyond the error they would otherwise lead to.

Trenly:
> This is more of a comment for the issue, but since we have a PR:
> 
> I'm not seeing the rationale for this change. We haven't had any issues that I'm aware of, and externally it isn't visible. Moving from what is effectively a very large number space (strings) to a smaller one doesn't seem like it is going to make things any better. In fact, it makes off by one errors much more likely to cause a problem beyond the error they would otherwise lead to.

It's true that we haven't had any issues that I'm aware of either. My thought was that it would be more difficult to make an error when there is a smaller number space than it would be otherwise. However, given the limited value here and the feedback, I'm okay moving on from this. 

Simply had assumed that because there was a specific set of values that could / should be used, a strictly typed set would be better

PR #6321, opened 2026-06-23
Format WinINet HRESULT messages

### LINKED ISSUE #6319: Network failures should decode WinINet HRESULTs and include actionable proxy context

When `winget` fails to refresh a source or download through WinINet, the user-facing error can be too low-level to act on:

```text
InternetOpenUrl() failed.
0x80072efd : unknown error
```

In the case below, the failure was caused by local proxy configuration, but the output did not mention the effective proxy, the source URL being refreshed, or any recovery path such as checking proxy settings or using `--no-proxy` when available.

### Steps to reproduce

Environment:

```text
winget: v1.28.240
PowerShell: 7.6.3
Installed package: Microsoft.PowerShell 7.6.3.0
Source: winget -> https://cdn.winget.microsoft.com/cache
```

Proxy state:

```text
WinHTTP proxy: 127.0.0.1:7090
WinINet/user proxy: 127.0.0.1:7091
Git proxy: http://127.0.0.1:7091
127.0.0.1:7090: not listening
127.0.0.1:7091: listening, but TLS through it fails after CONNECT
```

Run:

```powershell
winget upgrade --id Microsoft.PowerShell --source winget --accept-package-agreements --accept-source-agreements --disable-interactivity
```

### Expected behavior

The error should expose enough context to diagnose the failure. For example:

- Decode `0x80072efd` as a WinINet/Win32 connection failure instead of `unknown error`.
- Include the source name or URL involved, where safe.
- Include whether a winget proxy/default proxy is active, where safe and without leaking credentials.
- Suggest relevant recovery actions, such as checking proxy settings or trying `--no-proxy` if enabled.

### Actual behavior

```text
尝试更新源失败： winget
执行此命令时发生意外错误：
InternetOpenUrl() failed.
0x80072efd : unknown error
```

The same URL succeeds when bypassing the proxy:

```powershell
curl.exe -I --noproxy "*" https://cdn.winget.microsoft.com/cache/source.msix
```

```text
HTTP/1.1 200 OK
Content-Length: 17357706
```

The same URL through the local proxy fails:

```powershell
curl.exe -I --proxy http://127.0.0.1:7091 https://cdn.winget.microsoft.com/cache/source.msix
```

```text
HTTP/1.1 200 Connection established
curl: (35) schannel: failed to receive handshake, SSL/TLS connection failed
```

### Source-level triage

Local checkout: `microsoft/winget-cli` at `5eb96e8`.

Relevant paths:

- `src/AppInstallerCommonCore/Downloader.cpp`
  - logs `WinINet downloading from url`
  - reads `Network().GetProxyUri()`
  - calls `InternetOpenUrl(...)`
  - throws `THROW_LAST_ERROR_IF_NULL_MSG(urlFile, "InternetOpenUrl() failed.")`
- `src/AppInstallerSharedLib/Errors.cpp`
  - `GetUserPresentableMessageForHR` prints `0x... : ...`
  - unknown external HRESULTs fall through to `std::system_category().message(hr)`
- `src/AppInstallerCLICore/Workflows/WorkflowBase.cpp`
  - catches `wil::ResultException`
  - writes `UnexpectedErrorExecutingCommand` plus `GetUserPresentableMessage(re)`

`0x80072efd` is `HRESULT_FROM_WIN32(ERROR_INTERNET_CANNOT_CONNECT)`. The current rendering appears to pass the full HRESULT to the system category instead of decoding the Win32 code via `HRESULT_CODE(hr)`, which can produce `unknown error`.

### Suggested fix direction

Minimal:

- In `Errors.cpp`, when an HRESULT has `FACILITY_WIN32`, use `HRESULT_CODE(hr)` for the system message.

Better diagnostic:

- In `Downloader.cpp`, enrich `InternetOpenUrl()` failure context with URL/source and proxy mode.
- Redact credentials if a proxy URI is printed.

Tests:

- Extend `src/AppInstallerCLITests/Errors.cpp` for `HRESULT_FROM_WIN32(ERROR_INTERNET_CANNOT_CONNECT)` and assert the user message is no longer `unknown error`.
- Add a workflow-level test using a download hook that throws `HRESULT_FROM_WIN32(ERROR_INTERNET_CANNOT_CONNECT)` and asserts the output includes a useful diagnostic.

### Related issues / PRs

- #566: exact `InternetOpenUrl() failed. 0x80072efd` report, closed
- #5412: duplicate of #566
- #1546: friendly error for `0x80072ee7 : unknown error`, open
- #190 / #4203: proxy support
- #5030 / #5054: proxy path related behavior

This issue is intended as a narrow diagnostic enhancement, not another connectivity bug report.

starSumi:
## Summary

Fixes microsoft/winget-cli#6319.

`HRESULT_FROM_WIN32(ERROR_INTERNET_CANNOT_CONNECT)` is a WinINet error. On MSVC, `std::system_category().message(12029)` and `std::system_category().message(0x80072efd)` both return `unknown error`, so decoding the HRESULT alone cannot provide the useful WinINet diagnostic.

This change decodes `FACILITY_WIN32` HRESULTs and retrieves WinINet-range messages from `wininet.dll`, falling back to `std::system_category` for all other errors.

## Changes

- Centralize system error formatting for both user-facing HRESULT paths.
- Retrieve localized WinINet messages from the WinINet message table.
- Cover both `GetUserPresentableMessage` and `HResultInformation::Find(...)->GetDescription()`.

## Validation

- Evidence: an isolated MSVC probe showed that the system category returns unknown error for both 12029 and 0x80072efd, while the WinINet message table returns the localized connection-failure message.
- `git diff --check` passed.
- An isolated MSVC probe verified that the system category returns `unknown error` for both values while the WinINet message table returns the localized connection-failure message.
- I restored the NuGet packages locally, but could not run `AppInstallerCLITests` because the first-time vcpkg registry fetch was interrupted before the required historical port trees were available. The updated PR CI is queued for the x64 and x86 test validation.

### DIFF
--- a/src/AppInstallerSharedLib/Errors.cpp
+++ b/src/AppInstallerSharedLib/Errors.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 #include "pch.h"
+#include <wininet.h>
 #include "Public/AppInstallerErrors.h"
 #include "Public/AppInstallerLogging.h"
 #include "Public/AppInstallerStrings.h"
@@ -343,6 +344,55 @@ namespace AppInstaller
                 UnknownHResultInformation(hr).GetDescription();
         }
 
+        int GetSystemErrorCode(HRESULT hr)
+        {
+            return static_cast<int>(HRESULT_FACILITY(hr) == FACILITY_WIN32 ? HRESULT_CODE(hr) : hr);
+        }
+
+        // WinINet errors are not present in the system message table.
+        std::optional<std::string> GetWinInetErrorMessage(int errorCode)
+        {
+            if (errorCode < ERROR_INTERNET_OUT_OF_HANDLES || errorCode > INTERNET_ERROR_LAST)
+            {
+                return std::nullopt;
+            }
+
+            wil::unique_hmodule module{ LoadLibraryExW(L"wininet.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32) };
+            if (!module)
+            {
+                return std::nullopt;
+            }
+
+            LPWSTR buffer = nullptr;
+            const DWORD messageLength = FormatMessageW(
+                FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_HMODULE | FORMAT_MESSAGE_IGNORE_INSERTS,
+                module.get(), errorCode, 0, reinterpret_cast<LPWSTR>(&buffer), 0, nullptr);
+            if (!messageLength)
+            {
+                return std::nullopt;
+            }
+
+            auto freeBuffer = wil::scope_exit([&]() { LocalFree(buffer); });
+            std::string message = Utility::ConvertToUTF8(std::wstring_view{ buffer, messageLength });
+            Utility::Trim(message);
+            return message.empty() ? std::nullopt : std::optional<std::string>{ std::move(message) };
+        }
+
+        std::string GetSystemErrorMessage(HRESULT hr)
+        {
+            const int errorCode = GetSystemErrorCode(hr);
+            if (HRESULT_FACILITY(hr) == FACILITY_WIN32)
+            {
+                auto winInetMessage = GetWinInetErrorMessage(errorCode);
+                if (winInetMessage)
+                {
+                    return std::move(winInetMessage).value();
+                }
+            }
+
+            return std::system_category().message(errorCode);
+        }
+
         void GetUserPresentableMessageForHR(std::ostringstream& strstr, HRESULT hr)
         {
             strstr << "0x" << Logging::SetHRFormat << hr << " : ";
@@ -361,8 +411,7 @@ namespace AppInstaller
                 }
                 else
                 {
-                    strstr << std::system_category().message(
-                        HRESULT_FACILITY(hr) == FACILITY_WIN32 ? HRESULT_CODE(hr) : hr);
+                    strstr << GetSystemErrorMessage(hr);
                 }
             }
         }
@@ -438,7 +487,7 @@ namespace AppInstaller
             }
             else
             {
-                return Utility::LocIndString{ std::system_category().message(m_value) };
+                return Utility::LocIndString{ GetSystemErrorMessage(m_value) };
             }
         }
 
--- a/src/AppInstallerSharedLib/Errors.cpp
+++ b/src/AppInstallerSharedLib/Errors.cpp
@@ -361,7 +361,8 @@ namespace AppInstaller
                 }
                 else
                 {
-                    strstr << std::system_category().message(hr);
+                    strstr << std::system_category().message(
+                        HRESULT_FACILITY(hr) == FACILITY_WIN32 ? HRESULT_CODE(hr) : hr);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


starSumi:

florelis (dismissed):

starSumi:
Follow-up: I found the cause of the previous x64 and x86 test failures. On MSVC, `std::system_category().message(12029)` and `std::system_category().message(0x80072efd)` both return `unknown error`, so decoding the HRESULT alone does not produce a WinINet message.

This update retrieves the localized message from `wininet.dll` for WinINet-range codes, uses the same helper in `HResultInformation::GetDescription()`, and extends the regression test to cover both user-facing paths.

Local validation: `git diff --check` passed, and an isolated MSVC probe verified the message-table behavior. I could not run `AppInstallerCLITests` locally because the first-time vcpkg registry fetch was interrupted before its historical port trees were available; the refreshed PR CI is queued for the x64 and x86 tests.

starSumi:
Azure reports that I do not have permission to queue the x64/x86 pipeline for the updated head eefe8e7. @florelis, could you please rerun it when convenient?

Evidence: the prior test failure was caused by the system-category assumption. This head now reads the WinINet message table and covers both user-facing formatting paths.

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -343,6 +344,55 @@ namespace AppInstaller
...
+            if (HRESULT_FACILITY(hr) == FACILITY_WIN32)
+            {
+                auto winInetMessage = GetWinInetErrorMessage(errorCode);
+                if (winInetMessage)
+                {
+                    return std::move(winInetMessage).value();
+                }
+            }
+
+            return std::system_category().message(errorCode);
```
JohnMcPMS: Did you intentionally change the Win32 errors to look up via code only?  I don't know that it is required, and it would make the code above cleaner to simply extract the code when needed rather than unconditionally.

src/AppInstallerSharedLib/Errors.cpp
```diff
@@ -343,6 +344,55 @@ namespace AppInstaller
...
+
+        // WinINet errors are not present in the system message table.
+        std::optional<std::string> GetWinInetErrorMessage(int errorCode)
+        {
+            if (errorCode < ERROR_INTERNET_OUT_OF_HANDLES || errorCode > INTERNET_ERROR_LAST)
+            {
+                return std::nullopt;
+            }
+
+            wil::unique_hmodule module{ LoadLibraryExW(L"wininet.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32) };
```
JohnMcPMS: Should specify `LOAD_LIBRARY_AS_DATAFILE` since we are only extracting a message.

src/AppInstallerSharedLib/Errors.cpp
starSumi: Updated in `9c2d47b`. `HRESULT_CODE(hr)` is now evaluated only for the WinINet message-table lookup, while the fallback continues to pass the original HRESULT to `std::system_category()`.

src/AppInstallerSharedLib/Errors.cpp
starSumi: Updated in `9c2d47b`. The module is now loaded with `LOAD_LIBRARY_AS_DATAFILE | LOAD_LIBRARY_SEARCH_SYSTEM32`, so it is mapped for resource access while retaining the System32-only search boundary.

JohnMcPMS (approved):

PR #6301, opened 2026-06-19
Fix cpprest checked iterator build error

Trenly:
## 📖 Description
Adds a patch for `cpprestsdk` to allow it to build in VS2026 18.5.3

## 🔗 References
* Resolves #6300

## 🔍 Validation
Built locally

## ✅ Checklist

- [x] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [x] Linked to an issue
- [ ] Updated [Release Notes](../doc/ReleaseNotes.md) (if applicable)
- [ ] Updated documentation (if applicable)
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed)

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6301)

PR #6339, opened 2026-06-26
Clean vcpkg artifacts on solution clean

### LINKED ISSUE #6338: Clean Solution does not remove vcpkg artifacts
### Relevant area(s)
WinGet CLI

### Relevant command(s)
N/A (Visual Studio Build > Clean Solution)

When running **Clean Solution** in Visual Studio for `src\AppInstallerCLI.sln`, vcpkg artifacts are left behind. Specifically, `src\vcpkg_installed\` and `src\VcpkgPortOverlay\.created` are not removed.

### Steps to reproduce
1. Open `src\AppInstallerCLI.sln` in Visual Studio.
2. Build the solution once so vcpkg artifacts exist.
3. Run **Build > Clean Solution**.
4. Inspect `src\vcpkg_installed\` and `src\VcpkgPortOverlay\.created`.

### Expected behavior
Clean Solution should remove repository-local vcpkg artifacts so dependency state is reset.

### Actual behavior
`src\vcpkg_installed\` remains and overlay stamp file persists, so vcpkg state carries across cleans.

### Environment
Visual Studio 2022
Windows 11

Trenly:
## 📖 Description
Add a `CleanVcpkg` target in `src\vcpkg.props` so Visual Studio **Clean Solution** removes repository-local vcpkg artifacts.

The target:
- Deletes `$(VcpkgInstalledDir)` (`src\vcpkg_installed\`)
- Deletes `src\VcpkgPortOverlay\.created` so overlay ports regenerate on next build

Also adds an XML comment block explaining why this cleanup is necessary.

## 🔗 References
Resolves #6338

## 🔍 Validation
Manual local verification of MSBuild target wiring only.

## ✅ Checklist
- [x] Linked to an issue

## 📋 Issue Type
- [x] Bug fix
- [ ] Feature
- [ ] Task

---
Assisted-by: GitHub Copilot
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6339)

### DIFF
--- a/src/vcpkg.props
+++ b/src/vcpkg.props
@@ -61,6 +61,14 @@
     <Exec Command="pwsh.exe -ExecutionPolicy Bypass -File &quot;$(MSBuildThisFileDirectory)VcpkgPortOverlay\CreatePortOverlay.ps1&quot; -StampFile &quot;$(MSBuildThisFileDirectory)VcpkgPortOverlay\.created&quot;" />
   </Target>
 
+  <!-- Visual Studio "Clean Solution" only clears project build outputs. vcpkg artifacts
+       live under src\vcpkg_installed and are not removed unless we clean them explicitly.
+       Removing the overlay stamp forces regeneration of overlay ports on the next build. -->
+  <Target Name="CleanVcpkg" AfterTargets="Clean">
+    <RemoveDir Directories="$(VcpkgInstalledDir)" />
+    <Delete Files="$(MSBuildThisFileDirectory)VcpkgPortOverlay\.created" />
+  </Target>
+
   <Target Name="EnsureNoTripletMismatch"
           BeforeTargets="PrepareForBuild"
           Condition="'$(_IsBuildingSln)' != 'true'" >


florelis (approved):

florelis:
What's the scenario you're thinking about that may require cleaning the vcpkgs?
I can only think of the case where you are updating the ports from our custom overlay. If that is the only case, maybe we don't need it in most cases and could have it gated behind some option, or make the target smarter to only clean up when the ports changed.

Trenly:
> What's the scenario you're thinking about that may require cleaning the vcpkgs? I can only think of the case where you are updating the ports from our custom overlay. If that is the only case, maybe we don't need it in most cases and could have it gated behind some option, or make the target smarter to only clean up when the ports changed.

Take for example #6301 - That was only discovered because I removed the vcpkgs. In most workflows, I would expect "Clean" to remove all build artifacts and produce a 100% brand new build - which wasn’t the case. 

While a simple rebuild would skip updating the vcpkgs, if I am explicitly running a clean, I would have thought that the vcpkgs would have also been cleaned up too - regardless of if there were port changes or not. In my mind, Clean means to perform the next build as if it were the first build the system ever ran

PR #6344, opened 2026-07-02
Block admin uninstall of user package (1.29)

JohnMcPMS:
## 📖 Description
Block admin uninstall of user scope package.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6344)


florelis (approved):

src/AppInstallerCLICore/Workflows/UninstallFlow.cpp
```diff
@@ -214,16 +229,27 @@ namespace AppInstaller::CLI::Workflow
             return;
         }
 
-        const std::string installedTypeString = installedPackageVersion->GetMetadata()[PackageVersionMetadata::InstalledType];
-        switch (ConvertToInstallerTypeEnum(installedTypeString))
+        auto packageMetadata = installedPackageVersion->GetMetadata();
+        auto installedType = ConvertToInstallerTypeEnum(packageMetadata[PackageVersionMetadata::InstalledType]);
+
+        // When running as admin, block attempt to uninstall user scope package to prevent EOP paths.
```
henrivdr01, line 235: @JohnMcPMS do you have more context on this?
What is the EOP paths concern?

florelis: @henrivdr01 The uninstall command is read from the registry, and for packages installed as regular user, that is in HKCU, which does not require admin to write. So, we shouldn't just run that command as admin because somebody could have changed it.

Issue #1753, opened 2021-11-29
Provide Structured Data Format Options


jdhitsolutions:
### Description of the new feature / enhancement

Short of adding native PowerShell support, it would be very helpful if a command-line `Winget upgrade` or `winget list` had a format switch that would write output to a format like CSV, JSON, or even XML.  That way I could run a native Winget command in a PowerShell function and more easily work with the output.  Right now, parsing the native Winget output is a very, very tedious task.

### Proposed technical implementation details

I'd like to run a command like this in PowerShell

```powershell
winget upgrade --format json | Out-file upgrade.json
```

denelon:
@jdhitsolutions we are working on PowerShell cmdlets. We are also discussing adding a switch to provide JSON output via the COM interface to help with our own ability to support PowerShell pipelining. You can see some of the work in progress on [my fork](https://github.com/denelon/winget-cli/tree/PowerShellFHL/PowerShell/Microsoft.WinGet.Client/src).

rkeithhill-keysight:
Since my feature request was closed as a duplicate of this one, I wanted to add my example usages here:
```powershell
winget list -o csv | ConvertFrom-Csv | Where-Object Name -match Microsft | Sort-Object Name
```
You could also support a List format that would not truncate text e.g.:

```
winget list -o list

Name:      JetBrains ETW Host Service (x64) 
Id:        JetBrains.ReSharper
Version:   < 2022.2.2
Available: 2022.2.3
Source:    winget

Name: ...
...
```
I would imagine the default `--output-format` would be `Table` to match how winget displays by default.

treestryder:
Seems odd, after all these years of Powershell, that any group at Microsoft would roll their own command-line parser and not pipe .Net objects to a formater. <shrug>

rkeithhill-keysight:
Have you seen [System.CommandLine](https://learn.microsoft.com/en-us/dotnet/standard/commandline/)?  Or the [dotnet ](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet) CLI. 🤷‍♂️

treestryder:
Isn't the "PackageManagement" Powershell module a full-fledged package manager? Why didn't they just use that?

halr9000:
[...]

PR #6346, opened 2026-07-03
Add JSON format for list outputs

PrzemyslawKlys:
## 📖 Description

Adds `--format json` support for package listing output.

This enables machine-readable output for:

- `winget list --format json`
- `winget upgrade --format json` when used in list-only mode

The default table output is unchanged.

The JSON output includes installed package fields that match the existing list and upgrade table data, including package name, id, installed version, available version, and source. It also keeps JSON parseable for empty results and common error paths such as source-open failures, WinGet policy blocks, and generic execution failures.

`winget upgrade --format json` is intentionally limited to non-mutating upgrade listing. Commands such as `winget upgrade --all --format json` are rejected so package installation progress is not mixed with structured output.

## 🔗 References

Related to #1753.

## 🔍 Validation

Added focused test coverage for:

- JSON list output
- JSON upgrade-list output
- empty result output
- settings warnings before JSON output
- source-open failures
- WinGet policy blocks
- generic execution failures
- preserving help output for `--format json --help`
- rejecting JSON output for unsupported upgrade execution modes
- rejecting `--details` with JSON list output

Ran the focused `AppInstallerCLITests` scenarios for the JSON list and upgrade flows locally.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6346)

### DIFF
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -1,6 +1,6 @@
 ## New in v1.29
 
-Nothing yet.
+* Added `--format json` support for `winget list` and `winget upgrade` when listing available upgrades.
 
 ## Bug Fixes
 
--- a/src/AppInstallerCLICore/Argument.cpp
+++ b/src/AppInstallerCLICore/Argument.cpp
@@ -195,6 +195,8 @@ namespace AppInstaller::CLI
             return { type, "ascending"_liv, "asc"_liv, ArgTypeCategory::None, ArgTypeExclusiveSet::SortDirection };
         case Execution::Args::Type::SortDescending:
             return { type, "descending"_liv, "desc"_liv, ArgTypeCategory::None, ArgTypeExclusiveSet::SortDirection };
+        case Execution::Args::Type::OutputFormat:
+            return { type, "format"_liv };
 
         // Pin command
         case Execution::Args::Type::GatedVersion:
@@ -508,6 +510,8 @@ namespace AppInstaller::CLI
             return Argument{ type, Resource::String::CorrelationArgumentDescription, ArgumentType::Standard, Argument::Visibility::Hidden };
         case Args::Type::ListDetails:
             return Argument{ type, Resource::String::ListDetailsArgumentDescription, ArgumentType::Flag, Argument::Visibility::Help };
+        case Args::Type::OutputFormat:
+            return Argument{ type, Resource::String::OutputFormatArgumentDescription, ArgumentType::Standard, Argument::Visibility::Help };
         default:
             THROW_HR(E_UNEXPECTED);
         }
--- a/src/AppInstallerCLICore/Command.cpp
+++ b/src/AppInstallerCLICore/Command.cpp
@@ -819,6 +819,17 @@ namespace AppInstaller::CLI
             }
         }
 
+        if (execArgs.Contains(Execution::Args::Type::OutputFormat))
+        {
+            auto format = execArgs.GetArg(Execution::Args::Type::OutputFormat);
+            if (!Utility::CaseInsensitiveEquals(format, "table"sv) &&
+                !Utility::CaseInsensitiveEquals(format, "json"sv))
+            {
+                auto validOptions = Utility::Join(", "_liv, std::vector<Utility::LocIndString>{ "table"_lis, "json"_lis });
+                throw CommandException(Resource::String::InvalidArgumentValueError(ArgumentCommon::ForType(Execution::Args::Type::OutputFormat).Name, validOptions));
+            }
+        }
+
         Argument::ValidateExclusiveArguments(execArgs);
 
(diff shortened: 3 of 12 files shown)


florelis:
I don't think this implementation would scale well for extending it to the rest of the flows. As it is, any time we did any output, we would have to add a check for if we want JSON and do something different. I would prefer this to be handled in the `Reporter` as much as possible. For the tables in particular, the `TableOutput` could be extended to handle the output to JSON.

Maybe we should write a spec for this. Just to think out loud about what I think we could do for the implementation...
* We need to create schemas for all the objects we want to output, because people will definitely take a dependency on it and we can't break them later.
* We can add a flag to the reporter to tell it whether the output has to be text or JSON (or any other thing in the future)
* We would generally move away from passing strings directly to the reporter, and instead pass it objects that it knows how to format for each case. We would still output strings directly, but those would be more decorative (for example, "Installation succeeded" at the end) and the reporter would just discard them if it is using JSON.
  For example, for errors, we would pass an "error" object with an HResult and a message. For JSON it would add that as an object. For normal output it would print just the message. And then in both cases throw the HResult
* This would be too big to do in a single PR. We would add `--format` as an experimental argument, and then convert the output over time. If we set a flag in the `Reporter` to discard all normal strings when using JSON, it would initially return nothing for all commands, which I think could be a good default until we implement each one.

denelon:
Is there a reason the Microsoft.WinGet.Client PowerShell module isn't suitable for JSON output?

I know several folks would like JSON output for WinGet, but generally that is because they want to build something that can use the JSON output like an API. We do have COM APIs, and we built the PowerShell module that is likely a better candidate for structured JSON output.

PrzemyslawKlys:
> Is there a reason the Microsoft.WinGet.Client PowerShell module isn't suitable for JSON output?
> 
> I know several folks would like JSON output for WinGet, but generally that is because they want to build something that can use the JSON output like an API. We do have COM APIs, and we built the PowerShell module that is likely a better candidate for structured JSON output.

This is answered in the linked issue by all the people asking for structured output. PowerShell is great but you are not delivering it as part of the system and having more outputs isn't going to bring too much of additional maintance.

denelon:
I would want the "shape" of the JSON to conform with the PowerShell objects in either case if we are going to go down this path. As @florelis suggested, we should probably have a specification so we can capture this consistently. I would also want some kind of telemetry event to inform us when folks are using the JSON output so we can understand the usage patterns better.

We've been discussing internally something that could allow us to ship PowerShell in the packaged application, and we've reached out to the PowerShell team about how this might work and be extended to other packaged applications that also offer a PowerShell module. @SteveL-MSFT for visibility.

ranm-msft (commented):
Not weighing into the architecture and schema discussion above; this is one concrete thing in the implementation as written.

**The JSON output inherits a table presentation rule and discards known provenance.** In `ReportListResult`, `sourceName` is populated for every correlated package ("Always show the source for correlated packages"), but it only reaches `InstalledPackagesTableLine` through `shouldShowSource ? sourceName : Utility::LocIndString()`, and `shouldShowSource` is `source.IsComposite() && source.GetAvailableSources().size() > 1`. That gate exists so the human table can hide a Source column that would be identical on every row.

The JSON serializer then reads `line.Source.get()`, so on the common single-source machine every element of `packages[]` serializes `"source": ""` even though the source was known and is already flowing through the same function. A human reading a table can infer the single source from context; a machine consumer parsing `--format json` cannot, and `sourceFailures[].source` in the same document is populated, so the empty value reads as "no source" rather than "column hidden".

Could the JSON path keep `sourceName` whenever `latestVersion` provides one, and leave the `shouldShowSource` suppression to the table renderer only? It would also be worth asserting `packages[].source` in the tests, since nothing currently covers it.

PrzemyslawKlys:
I've put the design proposal in #6534. It covers the JSON shape, PowerShell alignment, reporter ownership, and the first `list`/listing-only `upgrade` slice. I'd appreciate feedback there on the contract before I reshape this implementation around it.

The current PR also has a concrete source-field bug in the single-source case: the JSON path inherits the table's decision to hide the Source column. I'll keep the actual source value in the structured result when updating the code. I haven't changed this PR's head yet; the spec is a proposal, not an accepted design.

PrzemyslawKlys:
I've opened #6536 as the draft implementation of the proposal in #6534 rather than trying to reshape this conflicting branch in place. The new implementation moves the document lifecycle into the Reporter, adds versioned schemas and telemetry, and keeps source provenance in the typed result even for a single source. I'm leaving this PR unchanged while the design direction is reviewed.

PR #6357, opened 2026-07-07
Ensure relative paths are rooted

florelis:
## 📖 Description

The way we were checking that a relative path was inside a directory was wrong because it didn't check if the path was rooted. This PR fixes that.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6357)

### DIFF
--- a/src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
+++ b/src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
@@ -306,6 +306,14 @@ namespace AppInstaller::Manifest
                 {
                     resultErrors.emplace_back(ManifestError::ExceededCommandsLimit);
                 }
+
+                if (!installer.Commands.empty())
+                {
+                    if (AppInstaller::Filesystem::PathEscapesBaseDirectory(installer.Commands[0]))
+                    {
+                        resultErrors.emplace_back(ManifestError::PortableCommandAliasEscapesDirectory, "Commands");
+                    }
+                }
             }
 
             if (installer.EffectiveInstallerType() == InstallerTypeEnum::Portable)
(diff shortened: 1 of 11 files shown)


Trenly (approved):
I know this is a minor change, but should there be an entry in ReleaseNotes under the bug fix section?

```md
* Fixed relative path handling for rooted paths
```

florelis:
> I know this is a minor change, but should there be an entry in ReleaseNotes under the bug fix section?
> 
> ```
> * Fixed relative path handling for rooted paths
> ```

Sure. I forgot we had a section for bug fixes, and it didn't seem big enough to list it otherwise.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -306,6 +306,14 @@ namespace AppInstaller::Manifest
                 {
                     resultErrors.emplace_back(ManifestError::ExceededCommandsLimit);
                 }
+
+                if (!installer.Commands.empty())
+                {
+                    if (AppInstaller::Filesystem::PathEscapesBaseDirectory(installer.Commands[0]))
```
JohnMcPMS, line 312: Why only the first command?  Is there a reason to allow the others to escape?

JohnMcPMS: The same check should be applied in the actual workflow even if it is a duplicated effort.

florelis: > Why only the first command? Is there a reason to allow the others to escape?

Because there is only a single command in this case. We check `Commands.size()` right before

JohnMcPMS: If the default diff view had given me one more line...

We still need to apply the escape check when we are actually acting on the install.

JohnMcPMS (approved):

Issue #4944, opened 2024-11-05
COM API client FindPackages function fails in CreateInstance method when run under LOCAL SYSTEM Context. Unable to find WindowsPackageManagerServer.exe


dave-patchblox:
### Brief description of your issue

COM Client seems to have an issue when run under the SYSTEM context. I do not believe this is the same issue folks have described where they cannot launch winget that is located in %PROGRAMFILES%\windowsapps\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe. On my Windows 10 machine, the native wingetcli runs as expected as either a normal user, administrator equivalent OR SYSTEM. IOW, the failure does not repro on native winget client.

The issue I am seeing has to do with the calling of the 'FindPackages' API under the SYSTEM context. This API appears to use WindowsPackageManagerFactory under the hood. This API attempts to instantiate the WindowsPackageManagerServer COM server. The code for WindowsPackageManagerServer found here https://github.com/microsoft/winget-cli/tree/master/src/WinGetServer
/WinGetServerManualActivation_Client.cpp seems to lookup the path to the EXE by calling GetPackageLocation. This function calls FindPackagesByPackageFamily (for a PROD build the "Microsoft.DesktopAppInstaller_8wekyb3d8bbwe" directory and exe name as "WindowsPackageManagerServer.exe" are used). On my machine, the path it comes up with is...

%WINDIR%\System32\config\systemprofile\AppData\Local\Microsoft\windowsapps\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe

That directory doesnt exist by default and the COM invocation fails with a ERROR_FILE_NOT_FOUND. FWIW, if I copy the actual folder and it contents where the server actually exists (%PROGRAMFILES%\windowsapps\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe.) to that directory, then the COM server is happily created. Obviously that is just a test to triage the issue. So, the question I have is WHY is FindPackagesByPackageFamily returning %WINDIR%\System32\config\systemprofile\AppData\Local\Microsoft\windowsapps? 

FWIW, as a logged in user, I seem to have a symlink/alias directory under my user profile at C:\Users\<me>\AppData\Local\Microsoft\WindowsApps\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe that was created by something. I am wondering if the code that creates that symlink/alias hasnt been run for the SYSTEM profile. 

### Steps to reproduce

[...]

PR #2922, opened 2023-02-04
Fix GetFullNameFromFamilyName for non-elevated context

yao-msft:
PackageManager.FindPackages requires elevated context, the fix will be to revert to use previous FindPackagesByPackageFamily

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/winget-cli/pulls/2922)

PR #6496, opened 2026-08-31
Handle empty package enumeration during COM manual activation

RatinCN:
## 📖 Description

The initial sizing call to `FindPackagesByPackageFamily` returns `ERROR_INSUFFICIENT_BUFFER` when matching packages exist, but can validly return `ERROR_SUCCESS` with a count of zero when no package is registered for the current identity.

The current code passes that success result to a failure-only WIL macro, causing the calling process to CRASH through FAIL_FAST instead of receiving a normal HRESULT. This change accepts the valid empty result and lets the existing control flow return `ERROR_PACKAGE_NOT_REGISTERED_FOR_USER` instead.

The registered-package path and propagation of other Win32 errors remain unchanged. This is consistent with the handling established in #2922.

For the SYSTEM scenario in #4944, this prevents the direct CRASH caused by FAIL_FAST, but does not add OutOfProc COM support, enumerate packages registered to other users, or use staged packages.

## 🔗 References

- Related to #4944; addresses the FAIL_FAST CRASH only.
- Similar empty-result handling was introduced in #2922.
- [`FindPackagesByPackageFamily` documentation](https://learn.microsoft.com/windows/win32/api/appmodel/nf-appmodel-findpackagesbypackagefamily)

## 🔍 Validation

- Reproduced the existing SYSTEM CRASH with `0xC0000409` (`FAIL_FAST`).
- Confirmed that the API returns `ERROR_SUCCESS` with a package count of zero under SYSTEM.
- Built `Microsoft.Management.Deployment.OutOfProc` for x64 Debug with the production package family and CLSIDs.
- Verified the modified DLL as a normal user: exit code 0 and 237 packages enumerated.
- Verified the modified DLL under SYSTEM: caught `COMException` with HRESULT `0x80073D35`; exit code 0 with no process crash.

## ✅ Checklist

- [x] Linked to an issue
- [x] Updated [Release Notes](../doc/ReleaseNotes.md) (not applicable)
- [x] Updated documentation (not applicable)
- [x] Updated [Copilot instructions](.github/copilot-instructions.md) (not applicable)

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6496)

### DIFF
--- a/src/WinGetServer/WinGetServerManualActivation_Client.cpp
+++ b/src/WinGetServer/WinGetServerManualActivation_Client.cpp
@@ -113,7 +113,7 @@ private:
         std::unique_ptr<UINT32[]> properties;
 
         LONG result = FindPackagesByPackageFamily(pfn.c_str(), PACKAGE_FILTER_HEAD, &count, nullptr, &bufferLength, nullptr, nullptr);
-        THROW_WIN32_IF(result, result != ERROR_INSUFFICIENT_BUFFER);
+        THROW_WIN32_IF(result, result != ERROR_INSUFFICIENT_BUFFER && result != ERROR_SUCCESS);
 
         for (size_t i = 0; i < 10 && result == ERROR_INSUFFICIENT_BUFFER; ++i)
         {


JohnMcPMS:
This code is not relevant for SYSTEM use, but the change appears righteous for standard users.

RatinCN:
> This code is not relevant for SYSTEM use, but the change appears righteous for standard users.

Agreed. SYSTEM is not supported for this OutOfProc path and was only used to reproduce the issue. This change handles the valid empty result by returning `ERROR_PACKAGE_NOT_REGISTERED_FOR_USER` instead of crashing through `FAIL_FAST`.

JohnMcPMS (approved):

PR #6500, opened 2026-09-01
Delta Index Spec

JohnMcPMS:
## 📖 Description
Spec for delta index support; see spec for many more details.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6500)

### DIFF
--- a/doc/specs/Delta Index.md
+++ b/doc/specs/Delta Index.md
@@ -148,7 +148,7 @@ Delta creation takes three inputs beyond the working index:
 
 | Input | Purpose |
 |---|---|
-| **Baseline index file** | Read-only source for the "before" state. Every changed package is diffed against this. Must carry a baseline GUID; generation fails if it does not. The GUID is copied into the delta. |
+| **Baseline index file** | Read-only source for the "before" state. Every changed package is compared against this. Must carry a baseline GUID; generation fails if it does not. The GUID is copied into the delta. |
 | **Baseline relative path** | Where the baseline package will live, relative to the source root. Recorded in the delta so the client can locate the baseline. Deliberately **independent** of the version so that the service retains freedom in how it lays out storage. |
 | **Baseline MSIX package version** | The identity version of the baseline package. Recorded in the delta so the client can determine whether the baseline it already has is the one this delta needs, without a network round trip. |
 
@@ -529,7 +529,7 @@ optimization, never a correctness dependency.
               │                           │
               ▼                           ▼
    ┌────────────────────┐      ┌─────────────────────────┐
-   │  delta.db  (main)  │      │  baseline.db (ATTACHed) │
+   │  delta.db  (main)  │      │  baseline.db (ATTACH)   │
    │  delta_packages    │      │  packages               │
    │  delta_tags2 ...   │      │  tags2 ...              │
    └────────────────────┘      └─────────────────────────┘
--- a/doc/specs/Delta Index.md
+++ b/doc/specs/Delta Index.md
@@ -1,0 +1,955 @@
+---
+author: JohnMcPMS, GitHub Copilot <Copilot>
+created on: 2026-09-01
+last updated: 2026-09-01
+---
+
+# Delta Index
+
+## Abstract
+
+The pre-indexed package source ships a complete SQLite index inside an MSIX package on every
+update cycle, so every client downloads the entire index even though consecutive publishes
+differ by a tiny fraction of their content. This spec proposes a **delta index**: a small
+supplemental SQLite database holding only the rows that changed since a designated **baseline**
+index, which the client combines with the baseline at query time using `ATTACH` and TEMP VIEWs.
+The merge is transparent to the existing V2 read path, so search, correlation, and every other
+consumer of the index operate without modification.
+
+## Inspiration
+
+Source update is the most frequent network operation winget performs and the one with the worst
+ratio of bytes transferred to information gained. The index grows monotonically with the size of
+the community repository, so the cost of this operation increases for every user over time,
+while the amount of *new* data in each publish stays roughly constant.
+
+The drivers:
+
+- **Egress cost.** The full index is served to every client on every update cycle. The bytes are
+  dominated by data the client already has.
+- **User-visible time and bandwidth.** Users on metered or slow connections pay the full index
+  cost repeatedly.
+- **The trend is adverse.** As the community repository grows, so does the fixed per-update cost.
+  Nothing in the current design amortizes it.
+
+A proof of concept validated the core mechanism — generation of a delta from a baseline, and
+transparent merged reads via SQLite views — and a measurement tool walking the git history of the
+community repository was used to characterize how delta size grows relative to the full index
+over time. Those measurements inform the baseline refresh cadence discussed under
+[Baseline selection policy](#baseline-selection-policy).
+
+## Solution Design
+
+### Terminology
+
+| Term | Meaning |
+|---|---|
+| **Full index** | The complete V2 index as published in the pre-indexed package source (`index.db` inside `source2.msix`). |
+| **Baseline** | A full index that has been *explicitly designated* as the reference point for one or more deltas, which stamps it with a baseline GUID. Structurally identical to a full index; "baseline" is a role conferred by designation, not a distinct format. |
+| **Baseline GUID** | A unique identifier stamped into an index when it is designated a baseline. Deltas record the GUID of the baseline they were built against; the client uses GUID equality to guarantee it is pairing the right two files. |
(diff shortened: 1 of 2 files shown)


doc/specs/Delta Index.md
```diff
@@ -0,0 +1,955 @@
...
+  │      full    → fixed location:   <root>/source2.msix (unchanged)│
+  └─────────────────────────────────────────────────────────────────┘
+```
+
+#### Step 2 — baseline selection and designation
+
+Whether to roll the baseline is a **service policy decision**, not a client or index concern.
+The index code is told what the baseline is; it does not choose.
+
+A baseline roll is expensive: every delta client pays a full-index download at their next
```
florelis, line 116: What is a "delta client"? A client capable of consuming deltas, or a client that has a baseline and a delta? I'm assuming the first.

florelis, line 116: This is probably out of scope for now, but I wonder if we could find a way to avoid paying the full cost for baseline rolls. I'm thinking of MSIX deltas, but IIRC you've said that wouldn't help because we download the full .msix for source updates.

doc/specs/Delta Index.md
```diff
@@ -0,0 +1,955 @@
...
+[Baseline selection policy](#baseline-selection-policy).
+
+##### Designation is an explicit act
+
+Being a baseline is **not** an implicit property of any prepared index. The service calls a
+dedicated function — `MarkAsBaseline` — on a prepared full index to confer the role. That
+function:
+
+1. Validates the index is a legitimate baseline candidate: post-`PrepareForPackaging`, schema
+   V2.1 or higher, and not itself a delta.
```
florelis, line 128: > not itself a delta

That just means it has the identity for the full index, not the identity for the index delta, right?

doc/specs/Delta Index.md
```diff
@@ -0,0 +1,955 @@
...
+The index code is told what the baseline is; it does not choose.
+
+A baseline roll is expensive: every delta client pays a full-index download at their next
+update. A baseline that is too old is also expensive: the delta grows monotonically until it
+approaches the size of the full index. The optimum sits between those, and is discussed under
+[Baseline selection policy](#baseline-selection-policy).
+
+##### Designation is an explicit act
+
+Being a baseline is **not** an implicit property of any prepared index. The service calls a
```
florelis, line 123: Why would we not designate all indexes as potential baselines? Couldn't we just choose to not use them for future deltas even if they have a designated GUID?

doc/specs/Delta Index.md
```diff
@@ -0,0 +1,955 @@
...
+per V2 table role.
+
+**Metadata values:**
+
+| Key | Type | Description |
+|-----|------|-------------|
+| `BaselineGuid` | TEXT | GUID of the baseline this delta was built against |
+| `BaselineRelativePath` | TEXT | Publish location of the baseline, relative to the source root |
+| `BaselinePackageVersion` | TEXT | MSIX package version of the baseline package |
+| `MinimumBaselineSchemaVersion` | TEXT | Lowest baseline schema version this delta can be applied to |
```
florelis, line 230: Aren't deltas applied to a single baseline with a fixed schema version?

doc/specs/Delta Index.md
```diff
@@ -0,0 +1,955 @@
...
+The client verifies that the baseline it holds carries **the same baseline GUID the delta
+names**. This is an equality check on an identifier stamped at designation time, not a content
+comparison.
+
+A GUID rather than a hash of the baseline file, for three reasons:
+
+- **`PrepareForPackaging` is not byte-reproducible.** It ends in a VACUUM and stamps fresh
+  timestamps, so regenerating a logically identical baseline yields a different file. A hash
+  would force the service to archive the exact published bytes for as long as any delta
+  references them, and would break the moment a baseline was repackaged or re-vacuumed.
```
florelis, line 451: Why would we ever regenerate a baseline, or discard it while a delta is still available?

doc/specs/Delta Index.md
```diff
@@ -0,0 +1,955 @@
...
+- **Identity should survive packaging.** The client extracts the index from an MSIX. Hash
+  validation makes correctness depend on extraction being byte-faithful; GUID equality depends
+  only on the logical content the service designated.
+- **Designation is where identity belongs.** The GUID is created by the same explicit act that
+  confers the baseline role, so the two cannot drift apart.
+
+**Affinity is not integrity, and does not need to be.** The GUID proves the delta and baseline
+were built as a pair; it does not prove the baseline is uncorrupted. Corruption and tampering
```
florelis, line 459: One thing a GUID doesn't prove is that the assignment was unique, whereas a hash does. (Both are possible with low probability; I mean that with a hash there is no risk of a programming error assigning the same ID to the same baseline.

doc/specs/Delta Index.md
```diff
@@ -0,0 +1,955 @@
...
+> [!NOTE]
+> Because the console output is unchanged, there is no impact on scripts that parse source
+> update output.
+
+## Capabilities
+
+### Accessibility
+
+No impact. The feature adds no new console output, no new prompts, and no new interactive
+elements. Existing source update output — which is already screen-reader compatible — is
```
florelis, line 805: > which is already screen-reader compatible

Neat. I did not know this

doc/specs/Delta Index.md
JohnMcPMS: The former.

JohnMcPMS: MSIX deltas might work for this as the baseline can be immutable.  We would have to see how well deltas actually work for us though.  If the service doesn't remove old ones, we can survey them after a few baseline rolls.

We might also want to consider if things like not vacuuming the database would produce better binary diffs.  That could mean larger baselines but could be worth it if the binary diff was smaller.

doc/specs/Delta Index.md
JohnMcPMS: No, identity is not known here.  Deltas will be "tagged" with metadata and a different schema.

doc/specs/Delta Index.md
JohnMcPMS: We could, but it is wasted space (albeit a tiny amount).

ranm-msft (commented):
I read this alongside #6523 rather than on its own, and the design holds up well. The rowid-stability requirement in particular is the load-bearing piece, and I am glad it is stated explicitly rather than left implicit.

One thing worth reconciling before this merges: the concrete view definitions here have drifted from what #6523 implements, in two places.

1. **Supersession identity.** The spec supersedes packages with `p.id NOT IN (SELECT id FROM delta_packages)`; `CreatePackagesView` uses `NOT EXISTS (... d.rowid = b.rowid)`. The implementation's comments treat rowid as *the* supersession identity, which reads like the deliberate choice - but then it is the spec that should say rowid.

2. **Association suppression.** The spec's subquery carries `AND d.is_removed = 1`; `CreateAssociationView` has no `is_removed` predicate at all - it suppresses the baseline row whenever the delta names the pair. For deltas the described generator produces these are equivalent, since an `is_removed = 0` pair is only emitted when it is absent from the baseline. So this is not a correctness problem. But the implementation is deliberately defensive against a delta that violates that invariant, and the spec does not state the invariant anywhere.

Since the SQL here is not illustrative hand-waving - there is prose underneath explaining exactly why each predicate works - could the spec be updated to document the merge contract as implemented?

Trenly:
One thought I had - If the delta is what tells the client the baseline, is there a window where there is a new baseline but no new delta such that `Baseline A + Delta A` is older than `Baseline B` and a client which fetches `Delta A` is actually working on older data?

Perhaps this is already covered if the previous baseline becomes unavailable whenever a new baseline is published (triggering full index fallback), or a zero-diff delta being generated when a new baseline is published?

JohnMcPMS:
> One thought I had - If the delta is what tells the client the baseline, is there a window where there is a new baseline but no new delta such that `Baseline A + Delta A` is older than `Baseline B` and a client which fetches `Delta A` is actually working on older data?
> 
> Perhaps this is already covered if the previous baseline becomes unavailable whenever a new baseline is published (triggering full index fallback), or a zero-diff delta being generated when a new baseline is published?

I went with the zero-diff delta.  When the service decides to make a baseline, the code will "diff" the baseline against itself to produce an empty delta.  Special cases are to be avoided, and not worth saving the transfer cost for 1 out of the 100+ deltas we will create per baseline (and the smallest one at that).

Trenly:
Thanks! Just wanted to be sure I haven't gone insane yet

PR #6503, opened 2026-09-02
Disable incremental link on pipeline

JohnMcPMS:
## 📖 Description
Pipeline builds have been failing sporadically in the build step due to a space issue (yet again).  Disabling incremental linking there reduces the disk footprint.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6503)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -265,6 +265,8 @@ insufficientmemory
 Intelli
 INTRESOURCE
 invalidparameter
+iobj
+ipdb
 ishelp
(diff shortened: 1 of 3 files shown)


florelis (dismissed):
I think the real solution long term is to use custom build agents with more capacity. (Plus, maybe we'd get faster builds!)

We also probably should move to GitHub Actions for the public CI builds at some point.

florelis (approved):

Issue #5345, opened 2025-04-01
Deduplicating results from multiple sources


denelon:
### Description of the new feature / enhancement

When I run `winget install powertoys` I get duplicate results for the "same package".

WinGet should be able to determine if the package from more than one WinGet source is a duplicate (SHA256 for example). I should be informed there is a setting I could use to specify which source to default to when this happens.

![Image](https://github.com/user-attachments/assets/5b84fa35-1e57-4630-a274-25b1604263cb)

### Proposed technical implementation details

Add settings for the sequence to use when multiple sources return the "same" package (they may have different names, identifiers, and other metadata), but the installer is "unique" given the SHA256.

When WinGet detects duplicates and there are no sequences specified, the user should be informed about the feature.

When a user has settings for the sequence to be used when more than one source contains the same package, then WinGet should install the package from the preferred source (based on the sequence).

Trenly:
* Related to #1620 but specifically refined for sources

denelon:
* And: https://github.com/microsoft/winget-cli/issues/1769

denelon:
There may also be cases where the two sources aren't both offering the "latest" version of the package in question. In that case the SHA256 wouldn't match, so we will need to dig a bit deeper into a solution that works when this happens.

jamrobot:
> There may also be cases where the two sources aren't both offering the "latest" version of the package in question. In that case the SHA256 wouldn't match, so we will need to dig a bit deeper into a solution that works when this happens.

At this point, yes, MSStore latest version is v0.90, and WinGet is v0.89. And due to some issue in WinGet package submission process which cause the delay. 
But even if the issue fixed, version mismatch will still here (MSStore v0.90 vs WinGet v0.90.1_hotfix version).
I think the right fix should be done in winget to allow user to specify preferred source if duplication found.

From application developer side, it is possible that version mismatch in certain period of time could be by-design.

denelon:
For the "current" issue with PowerToys, I'd anticipate the "new" version published to WinGet would also be published to the Microsoft Store so both would have the same "latest" version available. It's almost certain there would be some delay before both sources become "eventually consistent" with having the same version available in both sources.

I've been thinking about a few different possibilities for dealing with these gaps in time where the same version isn't available in both sources. We may be able to create a record on the local device when we have a strong correlation between two package identifiers that have the "same" package so in the future, when the sources are offering different versions, WinGet can gracefully handle the situation.

[...]

PR #6509, opened 2026-09-08
Promote source priority to a feature

AmelBawa-msft:
## 📖 Description

Promotes source priority to a regular feature by removing experimental gating from CLI and DSC workflows and retiring the `experimentalFeatures.sourcePriority` setting.

Uses the existing commands and options. Default priorities, selection rules, and tie handling are unchanged. Updates documentation, release notes, and regression coverage.

REST filtering, interactive resolution, and heuristic deduplication remain out of scope for this PR.

## 🔗 References

- Related to #5345 — source-priority workstream only.
- Builds on #6029.

## 🔍 Validation

- Built the CLI, native tests, and E2E test project.
- Passed **43 native test cases with 792 assertions**, covering priority ordering, ties, negative values, argument boundaries, disabled experimental-feature policy, source output, and DSC source operations.
- Passed CLI smoke checks for `source add --help`, `source edit --help`, `features`, and `dscv3 source --schema`.

## ✅ Checklist

- [x] Linked to an issue
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed) — not applicable

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6509)

### DIFF
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -2,6 +2,10 @@
 
 ## New Features
 
+### Source priority
+
+Source priority is now available without enabling an experimental feature. Use `winget source add --priority <value>` or `winget source edit --name <source> --priority <value>` to configure it. Higher values take precedence; sources with equal priority still require disambiguation when multiple matches remain.
+
 ### Output locale override
 
 Added a persistent `output.locale` setting to override winget interface language using a BCP47 tag.
--- a/doc/Settings.md
+++ b/doc/Settings.md
@@ -450,15 +450,3 @@ This feature enables support for fonts via `winget settings`. The `winget font l
         "fonts": true
   },
 ```
-
-### sourcePriority
-
-This feature enables sources to have a priority value assigned. Sources with a higher priority will appear earlier in search results and will be selected for installing new packages when multiple sources have a matching package.
-
-Note that search result ordering is dependent on several factors, and source priority is the lowest field in that currently (match quality and field are more important).
-
-```json
-  "experimentalFeatures": {
-        "sourcePriority": true
-  },
-```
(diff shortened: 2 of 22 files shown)


JohnMcPMS (commented):
The biggest issue with this PR is that the feature branch name seems to be based solely on this change and not the overall feature.

JohnMcPMS (approved):

ranm-msft (commented):
Read through the de-gating at `ffc7ab8`. The mechanical removal looks complete and the new native tests are the good kind: `SourcePriority_SearchResult` covers the tied (`{0,2,2}`), all-default (`{0,0,0}`) and all-negative (`{-3,-1,-2}`) cases, not just the happy path, and running each with `ExperimentalFeatures` policy set to both `NotConfigured` and `Disabled` is exactly the right way to prove the feature no longer rides on that policy. Compatibility for anyone who had `experimentalFeatures.sourcePriority` set also looks clean — the `Experimental` definition in `settings.schema.0.2.json` has no `additionalProperties: false`, and `UserSettings` only validates over the known `Setting` enum, so the stale key is silently ignored rather than rejected.

One gap that this PR seems like the right moment to surface, since it's the "promote to a feature" change:

**Group Policy sources parse `Priority` but never apply it.** `SourceFromPolicy` carries `std::optional<int32_t> Priority`, `GroupPolicy.cpp` reads it out of the policy JSON (~L239-242) and round-trips it in `ToJsonString` (~L371-373). But in `SourceList.cpp`, the `SourceOrigin::GroupPolicy` branch builds `SourceDetailsInternal` by copying `Name`/`Type`/`Arg`/`Data`/`Identifier`/`Explicit`/`TrustLevel`/pinning and never copies `Priority`, so it stays default-`0`.

That was inert while the feature was gated. Now that the `stable_sort` and the highest-priority narrowing in `WorkflowBase.cpp` are unconditional, an admin can set `Priority` in `AdditionalSources`, have it accepted and echoed back, and still be outranked by any user-added source created with `--priority 1`.

Intentional for now, or worth a follow-up to copy it through?

PR #6514, opened 2026-09-10
Enforce explicit ID filters on REST search results

### LINKED ISSUE #2966: Multiple filters not applied

When using both `--name` and `--id` I would expect a specific package to be found. Multiple results are returned.

### Steps to reproduce

Run `winget show --name "Microsoft Teams" --id "Microsoft.Teams"`

### Expected behavior

```
Found Microsoft Teams [Microsoft.Teams]
Version: 1.6.00.376
Publisher: Microsoft Corporation
Publisher Url: https://www.microsoft.com/en-us
Publisher Support Url: https://support.microsoft.com/en-us/teams
Author: Microsoft Corporation
Moniker: teams
Description: Make amazing things happen together at home, work, and school.
Homepage: https://www.microsoft.com/en-us/microsoft-teams/group-chat-software
License: Proprietary
License Url: https://www.microsoft.com/en-us/legal/terms-of-use
Privacy Url: https://privacy.microsoft.com/en-us/privacystatement
Copyright: (c) Microsoft Corporation. All rights reserved.
Copyright Url: https://www.microsoft.com/en-us/legal/terms-of-use
Release Notes Url: https://support.microsoft.com/en-us/office/what-s-new-in-microsoft-teams-d7092a6d-c896-424c-b362-a472d5f105de
Tags:
  call
  calling
  chat
  collaborate
  collaboration
  conferencing
  meet
  meetings
  team
  teams
  video
  video-conferencing
  voice
  voip
  webinars
Installer:
  Installer Type: exe
  Installer Url: https://statics.teams.cdn.office.net/production-windows-x64/1.6.00.376/Teams_windows_x64.exe
  Installer SHA256: 09c4f7bb78a1ec44132214b6a7467dd7c3c9f176a9176cf9c370c9a67530d44b
```

### Actual behavior

```
Name            Id              Source
---------------------------------------
Microsoft Teams XP8BT8DW290MPQ  msstore
Microsoft Teams Microsoft.Teams winget
```

### Environment

```shell
PS C:\git\winget-pkgs> winget --info
Windows Package Manager (Preview) v1.5.441-preview

System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.441.0

Links

```

AmelBawa-msft:
## 📖 Description

Reject REST results that fail explicit package ID filters, preventing a matching name or query from admitting a package with a different ID.

`FilterSearchResult` reuses comparisons through `MatchesRequest` for Exact, CaseInsensitive, StartsWith, and Substring matching. Filtering runs before result limits and also applies to optimized ID lookups, preserving pagination and the substring-ID fallback.

Also fix `ICUCaseInsensitiveStartsWith` to fold complete strings before comparing prefixes, preventing valid Unicode matches from being discarded. Add regression coverage and release notes.

Local rejection is limited to explicit ID filters. Other fields and unsupported match types remain server-evaluated.

## 🎞️ Demo
winget-rest-id-filter-before-after

## 🔗 References

Resolves #2966.

## 🔍 Validation

- Built x64 Debug successfully. Passed 202 native test cases / 2,959 assertions covering strings, REST interfaces, matching, completion, source priority, and composite sources.
- Deployed `wingetdev` and passed 10 sequential, read-only live CLI checks covering combined filters, case sensitivity, invalid IDs, and Store lookups. Verbose logs confirmed that the original Teams scenario discards Store package `XP8BT8DW290MPQ` because its ID does not match `Microsoft.Teams`.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6514)

### DIFF
--- a/src/AppInstallerRepositoryCore/MatchCriteriaResolver.cpp
+++ b/src/AppInstallerRepositoryCore/MatchCriteriaResolver.cpp
@@ -9,11 +9,6 @@ namespace AppInstaller::Repository
     {
         using ValueMatchFunction = bool (*)(const Utility::NormalizedString&, const Utility::NormalizedString&);
 
-        bool ValueMatchFunction_AlwaysFalse(const Utility::NormalizedString&, const Utility::NormalizedString&)
-        {
-            return false;
-        }
-
         bool ValueMatchFunction_Exact(const Utility::NormalizedString& a, const Utility::NormalizedString& b)
         {
             return a == b;
@@ -50,7 +45,7 @@ namespace AppInstaller::Repository
             case MatchType::FuzzySubstring:
             case MatchType::Wildcard:
             default:
-                return ValueMatchFunction_AlwaysFalse;
+                return nullptr;
             }
         }
 
@@ -167,6 +162,16 @@ namespace AppInstaller::Repository
 
             return MatchType::Exact == result.Type;
         }
+    }
+
+    std::optional<bool> MatchesRequest(const RequestMatch& request, const Utility::NormalizedString& value)
+    {
+        if (auto matchFunction = GetMatchTypeFunction(request.Type))
+        {
+            return matchFunction(value, request.Value);
+        }
+
+        return std::nullopt;
     }
 
     PackageMatchFilter FindBestMatchCriteria(const SearchRequest& request, const IPackageVersion* packageVersion)
--- a/src/AppInstallerRepositoryCore/Rest/Schema/1_0/RestInterface_1_0.cpp
+++ b/src/AppInstallerRepositoryCore/Rest/Schema/1_0/RestInterface_1_0.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 #include "pch.h"
+#include "MatchCriteriaResolver.h"
 #include "Rest/Schema/1_0/Interface.h"
 #include "Rest/Schema/IRestClient.h"
 #include <winget/HttpClientHelper.h>
@@ -21,6 +22,32 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
         // Query params
         constexpr std::string_view VersionQueryParam = "Version"sv;
         constexpr std::string_view ChannelQueryParam = "Channel"sv;
+
+        void FilterSearchResult(const SearchRequest& request, IRestClient::SearchResult& result)
+        {
+            for (const auto& filter : request.Filters)
+            {
+                // Other fields may match metadata omitted from the search response, such as localized names.
+                if (filter.Field != PackageMatchField::Id)
+                {
+                    continue;
+                }
+
+                auto& matches = result.Matches;
+                matches.erase(std::remove_if(matches.begin(), matches.end(), [&](const IRestClient::Package& package)
+                {
+                    auto match = MatchesRequest(filter, package.PackageInformation.PackageIdentifier);
+                    if (match && !match.value())
+                    {
+                        AICLI_LOG(Repo, Verbose, << "Discarding REST package " << package.PackageInformation.PackageIdentifier <<
+                            ": does not match ID filter '" << filter.Value << "' [" << ToString(filter.Type) << "]");
+                        return true;
+                    }
+
+                    return false;
+                }), matches.end());
+            }
+        }
 
         utility::string_t GetSearchEndpoint(const std::string& restApiUri)
         {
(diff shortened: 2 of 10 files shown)


src/AppInstallerRepositoryCore/MatchCriteriaResolver.cpp
```diff
@@ -50,7 +45,7 @@ namespace AppInstaller::Repository
             case MatchType::FuzzySubstring:
             case MatchType::Wildcard:
             default:
-                return ValueMatchFunction_AlwaysFalse;
+                return nullptr;
```
JohnMcPMS, line 48: Fix the other caller to handle a possible null now. I get that it uses a fixed list and that list won't ever get a null from this implementation, but I don't like having that silent coupling.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/RestInterface_1_0.cpp
```diff
@@ -22,6 +23,32 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0
         constexpr std::string_view VersionQueryParam = "Version"sv;
         constexpr std::string_view ChannelQueryParam = "Channel"sv;
 
+        void FilterSearchResult(const SearchRequest& request, IRestClient::SearchResult& result)
+        {
+            for (const auto& filter : request.Filters)
```
JohnMcPMS, line 28: This ignores `Inclusions`.  The goal should be to ensure that the results reflect the request fully, not just that it works in the narrow case we are targeting.

JohnMcPMS (changes requested):
The scope is too limited and doesn't meet the criteria for the "hero" case. This function generates that:
https://github.com/microsoft/winget-cli/blob/c17eadbcacf2f5244b13de0d563d921c33ac4af6/src/AppInstallerCLICore/Workflows/WorkflowBase.cpp#L879-L906

We need a more generalized filtering pass, and it should probably live outside of the REST code since it isn't REST specific.

ranm-msft (commented):
Read the diff against head (64ae352) and walked `FilterSearchResult` through `SearchInternal`'s continuation loop. Not re-raising the `Inclusions` gap or the generalization/placement point - John has those covered.

One thing I'd like your read on, and the new `Search_ExplicitIdFilters_Continuation` test suggests you already thought about it: the loop's exit condition is `results.Matches.size() < request.MaximumResults`, and filtering now runs before that count advances. The test encodes exactly that - `allFiltered` with `MaximumResults == 1` still issues all three requests. I think that is the right semantics, since a valid match can live on page three and you cannot stop after N raw candidates. But it does mean `MaximumResults` no longer bounds round trips at all when a source returns candidates the filter rejects, which is the lax-source case this PR exists for.

Is it worth a separate bound on continuation requests - a page budget, or repeated-token detection - so "limit results" and "limit work" stay distinct concepts? Against a source that pages a broad name match, a bounded request can now walk the whole chain one POST at a time. Happy to be told the practical chain length makes this theoretical.

src/AppInstallerRepositoryCore/MatchCriteriaResolver.cpp
AmelBawa-msft: Added a null check at the other call site so it no longer relies on the fixed match-type list guaranteeing a valid comparator.

src/AppInstallerRepositoryCore/Rest/Schema/1_0/RestInterface_1_0.cpp
AmelBawa-msft: Expanded the pass beyond explicit ID filters to evaluate `(Query OR Inclusions...) AND Filters...`. The shared matching logic lives in `MatchCriteriaResolver`, with selective manifest retrieval handled by the REST adapter.

Added regression coverage for positional requests and both directions: a matching inclusion cannot bypass a failed filter, and matching filters cannot override a failed selector group. Source-defined queries and criteria we cannot verify remain unknown; only definite mismatches are discarded.

src/AppInstallerRepositoryCore/Rest/RestSource.cpp
```diff
@@ -291,6 +293,30 @@ namespace AppInstaller::Repository::Rest
...
+            {
+                if (version.Manifest)
+                {
+                    auto names = version.Manifest->GetPackageNames();
+                    auto publishers = version.Manifest->GetPublishers();
+                    for (const auto& name : names)
+                    {
+                        for (const auto& publisher : publishers)
+                        {
+                            result.emplace_back(Utility::LocIndString{ name }, Utility::LocIndString{ publisher });
```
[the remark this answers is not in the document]
AmelBawa-msft: Replaced the Cartesian products with a shared `Manifest::GetNameAndPublisherPairs()` helper that preserves each localization/ARP entry’s name and publisher, including the existing default fallback rules.

Both the cached REST path and `CompositeSource`’s manifest-download path now use it, and `ARPCorrelation` reuses the same helper. Added regression coverage confirming that genuine pairs correlate and mixed pairs do not, through both retrieval paths, plus coverage for fallback behavior.

PR #6519, opened 2026-09-11
Add a debug command to check Store cert pinning

JohnMcPMS:
## 📖 Description
Adds a debug command (`validate-store-pinning`) to check the current Store cert pinning against a target URL.

## 🔍 Validation
Works against the current URL and fails against an arbitrary non-Microsoft URL.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6519)

### DIFF
--- a/src/AppInstallerCLICore/Commands/DebugCommand.cpp
+++ b/src/AppInstallerCLICore/Commands/DebugCommand.cpp
@@ -14,8 +14,6 @@
 #include <winget/Certificates.h>
 #include <winget/HttpClientHelper.h>
 #include <winget/RepositorySource.h>
-#include <iomanip>
-#include <mutex>
 
 using namespace AppInstaller::CLI::Execution;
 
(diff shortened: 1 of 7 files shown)


ranm-msft (approved):
Read through this one carefully; it looks correct to me.

The part I spent the most time on was the lifetime aggregation, and I convinced myself it is right: a chain's remaining lifetime is the **minimum** across its certificates (a chain is only as healthy as its soonest-expiring cert), while the configuration takes the **maximum** across its chains (one validating chain is sufficient). That is the semantics you want for an early-warning signal, and it is an easy one to get backwards.

I also double-checked that `GetRemainingLifetimePercentage` returns *remaining* rather than elapsed lifetime - it does. The comment above the declaration reads ambiguously enough that I went and looked, so it might be worth a word if you are touching that file anyway, but it is pre-existing and not from this PR.

Approving.

Out of curiosity rather than as review feedback: is this meant to stay a manual debug command, or would you eventually want the lifetime check running as a health signal in the shipping client? The aggregation looks useful well outside `winget debug`.

PR #6526, opened 2026-09-16
Move pipeline jobs to templates

JohnMcPMS:
## 📖 Description
Move the pipeline jobs for building, testing and PowerShell to templates so that they can be more flexible.

## 🔍 Validation
The pipeline is the test.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6526)


templates/README.md
```diff
@@ -0,0 +1,321 @@
+# Pipeline templates
+
+These templates hold the Azure Pipelines jobs that build and test winget-cli. `azure-pipelines.yml`
+at the repository root consumes them, and so can a pipeline in another repository that vendors
```
florelis, line 4: > another repository that vendors winget-cli as a subtree

Maybe we should specify that we mean specifically the internal copy we use for official builds, so it doesn't sound like we intend to support other people doing this.

templates/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+Nested template references need no help. Azure Pipelines resolves a `- template:` reference inside a
+template file relative to *that file*, so the job templates find their siblings at any depth. Only
+the entry point has to know where the subtree lives.
+
+## `WINGET_SOURCE_ROOT`
+
+Not everything that needs the source root can be handed a task input. Some winget-cli content
+resolves the sources at *runtime*, from inside a process the pipeline merely launches, and that
+content used to read `BUILD_SOURCESDIRECTORY` — which is the root of the repository being built, and
```
florelis: Do we need the historical note?

templates/README.md
```diff
@@ -0,0 +1,321 @@
...
+jobs it follows the artifact copy. It is listed here because it is the one part of the contract that
+is neither a parameter nor a path in this directory.
+
+## What a consuming pipeline must provide
+
+**Pipeline variables.** None. `jobs-build.yml` sets what it needs at job scope — `solution`, so it
+can be rooted, and `EnableDetectorVcpkg`, so Component Governance detects the vcpkg dependencies.
+Both were previously pipeline-level variables in `azure-pipelines.yml`; a consumer that omitted
+`EnableDetectorVcpkg` would have gotten a passing scan that silently skipped the native
+dependencies, so the template owns it now.
```
florelis: Do we need the historical note?

templates/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+## The test jobs never check out
+
+`Test` and `BuildPowerShellModule` both declare `checkout: none` and read every source file they
+need from the build artifact.
+
+Two reasons. The jobs read scripts and test data that have to match the binaries they are testing,
+and a checkout is not guaranteed to match: it resolves to whatever the consuming repository's
+default branch or commit gives that job, which for a subtree consumer can drift from the sources the
+artifact was built from. Taking the files from the artifact removes the question. It also means the
```
florelis, line 140: I don't think this can drift in the same build

templates/README.md
```diff
@@ -0,0 +1,321 @@
...
+- template: subtree/templates/jobs-build.yml
+  parameters:
+    additionalMSBuildArgs: >-
+      /p:WingetDisableTestHooks=false
+      /p:UseProdCLSIDs=false
+```
+
+These must be command line properties. A consumer that assigns the property in a `.props` file
+assigns it unconditionally, and only a global property — which is what `/p:` creates — takes
+precedence over that.
```
florelis: I don't think we need this? Sounds like the AI documenting the problems it found along the way

templates/jobs-build.yml
```diff
@@ -0,0 +1,368 @@
...
+    displayName: Restore AppInstallerCLIPackage
+    inputs:
+      restoreSolution: '${{ parameters.sourceRoot }}\src\AppInstallerCLIPackage\AppInstallerCLIPackage.wapproj'
+      feedsToUse: 'config'
+      nugetConfigPath: '$(nugetConfig)'
+
+  - task: NuGetCommand@2
+    displayName: Restore AppInstallerTestMsixInstaller
+    inputs:
+      restoreSolution: '${{ parameters.sourceRoot }}\src\AppInstallerTestMsixInstaller\AppInstallerTestMsixInstaller.wapproj'
```
florelis: nit: Maybe we could do `${{ parameters.sourceRoot }}\\src\**\*.wapproj`

templates/jobs-build.yml
```diff
@@ -0,0 +1,368 @@
...
+      targetPath: '$(artifactsDir)'
+      artifact: 'Build.$(artifactIdentifier)'
+    condition: succeededOrFailed()
+
+  - task: ComponentGovernanceComponentDetection@0
+    displayName: Component Governance
+    inputs:
+      scanType: 'Register'
+      verbosity: 'Verbose'
+      alertWarningLevel: 'High'
```
florelis, line 361: Do we need this?

florelis (dismissed):

templates/README.md
JohnMcPMS: Maybe they have fixed it, but it definitely used to be able to.  If the merged view got updated between jobs (so if the main branch were updated), the pipeline could run jobs with different code because each job would just clone the current merged view when it ran.

templates/jobs-build.yml
JohnMcPMS: This guy certainly thought we did: #6

Madhusudhan-MSFT (approved):

PR #6530, opened 2026-09-17
Fix correlation tool

JohnMcPMS:
## 📖 Description
Fix some issues with the correlation tooling that have drifted.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6530)


tools/CorrelationTestbed/Test-CorrelationInSandbox.ps1
```diff
@@ -107,12 +107,12 @@ $ Enable-WindowsOptionalFeature -Online -FeatureName 'Containers-DisposableClien
...
 
 function Close-WindowsSandbox {
-    $sandbox = Get-Process 'WindowsSandboxClient' -ErrorAction SilentlyContinue
+    $sandbox = Get-Process 'WindowsSandboxRemoteSession' -ErrorAction SilentlyContinue
     if ($sandbox)
     {
       Write-Host '--> Closing Windows Sandbox'
 
-      $sandboxServer = Get-Process 'WindowsSandbox' -ErrorAction SilentlyContinue
+      $sandboxServer = Get-Process 'WindowsSandboxServer' -ErrorAction SilentlyContinue
```
ranm-msft, line 115: Could we probe both process-name generations here? The newer Sandbox uses `WindowsSandboxRemoteSession`/`WindowsSandboxServer`, but the older implementation can still be active when the Store upgrade is unavailable or fails - and with `-ErrorAction SilentlyContinue` plus the `if ($sandbox)` gate, cleanup then becomes a silent no-op and sandboxes leak between runs. Passing both names to each `Get-Process` (e.g. `Get-Process 'WindowsSandboxRemoteSession','WindowsSandboxClient' -ErrorAction SilentlyContinue`) would keep both generations working.

ranm-msft (commented):
Skimmed this one - single point below, on the sandbox process names.

florelis (approved):

PR #6531, opened 2026-09-18
Client enforce manifest fields used for paths

JohnMcPMS:
## 📖 Description
Enforce that fields used for paths cannot have invalid characters or sequences that would be used to traverse the filesystem.

## 🔍 Validation
Added tests for enforcement.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6531)


src/AppInstallerSharedLib/Filesystem.cpp
```diff
@@ -393,6 +394,15 @@ namespace AppInstaller::Filesystem
         return false;
     }
 
+    void ThrowIfPathEscapesBaseDirectory(std::string_view relativePath, HRESULT hr)
```
Trenly: Is there a reason for this new utility function and not just a macro call at each callsite? `THROW_HR_IF_MSG(APPINSTALLER_CLI_ERROR_INVALD_MANIFEST, PathEscapesBaseDirectory(relativePath), "Path part points to a location outside of its base directory: %s", relativePath)`

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -110,12 +113,87 @@ namespace AppInstaller::Manifest
...
+
+            std::string fieldNameString{ fieldName };
+            std::string valueString{ value };
+
+            if (value.length() > s_MaxPathFieldLength)
+            {
+                resultErrors.emplace_back(ManifestError::FieldExceedsMaxLength, fieldNameString, valueString);
+            }
+
+            for (char character : value)
```
Trenly, line 142: Instead of looping, isn't there a Shell API that will clean the path and tell you if there were any errors? [PathCleanupSpec](https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-pathcleanupspec#return-value). 

Or GetFileAttributesW and check the last error specificlly for `ERROR_INVALD_NAME` ?

JohnMcPMS: `PathCleanupSpec` requires a fully qualified path.  I think we could consider using it instead of outright failing in the use paths.

Trenly: Ah, I see. Thanks!

src/AppInstallerCLICore/Workflows/DownloadFlow.cpp
```diff
@@ -35,8 +35,11 @@ namespace AppInstaller::CLI::Workflow
         {
             const auto& manifest = context.Get<Execution::Data::Manifest>();
 
+            std::string pathPart = manifest.Id + '.' + manifest.Version;
+            THROW_HR_IF_MSG(APPINSTALLER_CLI_ERROR_INVALID_MANIFEST, Filesystem::PathEscapesBaseDirectory(pathPart), "Path part points to a location outside of its base directory: %hs", pathPart.c_str());
+
             std::filesystem::path tempInstallerPath = Runtime::GetPathTo(Runtime::PathName::Temp);
-            tempInstallerPath /= Utility::ConvertToUTF16(manifest.Id + '.' + manifest.Version);
+            tempInstallerPath /= Utility::ConvertToUTF16(pathPart);
```
Trenly, line 42: It might be a non-issue, but this introduces a change to the string between time of check and time of use. If the UTF8 representation passes the check, is it guaranteed that the UTF16 represenation will too?

ranm-msft (commented):
Looked at this from the "what can still reach the filesystem" angle rather than the diff shape. Enforcing the schema rules on Id/Version at runtime is the right call - REST sources never went through the schema, so that was a real hole - and the point-of-use `PathEscapesBaseDirectory` checks land where the values actually get concatenated (ContextOrchestrator, the temp/download dirs, the log name, Fonts).

One gap in `ValidatePathFieldValue`: the character set covers the Win32 invalid characters and relative specifiers, but not the other two things that make a name unusable as a path component.

- Reserved DOS device names are matched on the base name even when followed by an extension - `NUL.txt` and `NUL.tar.gz` are documented as equivalent to `NUL` - so a PackageIdentifier like `NUL.foo` still produces the component `NUL.foo.1.0`. The dotted-segment shape of an identifier doesn't save us here.
- PackageVersion terminates the constructed component, so a trailing dot or space ends up at the end of the directory/file name, where Win32 may normalize, alias or reject it depending on the API and the namespace form.

Separately, the control-character check starts at `0x01`, so an embedded NUL passes - a REST source can send `\u0000` and everything downstream is null-terminated.

Is the intent to cover those in the validator too, or to let them fail at the point of use?

ranm-msft (commented):
Re-reviewed at head against my earlier comment. Reserved DOS device names, embedded NUL/control characters (`<= 0x1f`), and the trailing-dot case are all addressed now - `ValidatePathFieldValue` routing through `MakeSuitablePathPart` and mapping to `ReservedPathName` covers those cleanly. Thanks for picking those up.

One case from that list looks like it is still open, and narrower than I originally framed it: **`PackageVersion` values ending in a space, arriving over REST**.

YAML manifests are safe - population trims `PackageVersion`, so a trailing space from the community repo does not survive. The REST path is different: `DeserializeVersionInfo` preserves the raw value, `ValidatePackageVersion` passes `disallowWhitespace: false` so spaces are legal, and `MakeSuitablePathPart` only treats a final `.` as illegal, not a final space. Because `ValidatePathFieldValue` only catches exceptions rather than comparing the sanitized output back to the input, a sanitization that is a silent no-op passes.

`GetPathPart` puts `Version` last, so the trailing space lands at the end of the component. On Windows `"v1 "` aliases to `"v1"` for both directories and files, which means two distinct REST versions can collide on one path.

Worth rejecting a trailing space alongside the trailing-dot case, or having the validator reject whenever `MakeSuitablePathPart` would alter the value? The latter closes this class of gap generally rather than one character at a time.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -94,6 +94,10 @@ namespace AppInstaller::Manifest
                 { AppInstaller::Manifest::ManifestError::BlockedMsiProperty, "Contains a blocked MSI property."sv },
                 { AppInstaller::Manifest::ManifestError::InvalidMsiSwitches, "Contains invalid MSI switches."sv },
                 { AppInstaller::Manifest::ManifestError::ContainsNetworkAddress, "Installer switch contains network address."sv },
+                { AppInstaller::Manifest::ManifestError::InvalidPathCharacters, "The field value contains characters that are not allowed because the value is used to construct a file system path."sv },
+                { AppInstaller::Manifest::ManifestError::FieldExceedsMaxLength, "The field value exceeds the maximum allowed length."sv },
+                { AppInstaller::Manifest::ManifestError::FieldEscapesDirectory, "The field value must not point to a location outside of its base directory."sv },
```
florelis: Do we still need the versions specific to the portable installers?

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -110,12 +114,101 @@ namespace AppInstaller::Manifest
...
+
+        // The characters excluded by the manifest schema from fields that are used to construct file system paths.
+        constexpr std::string_view s_InvalidPathFieldCharacters = "\\/:*?\"<>|"sv;
+
+        // The maximum length declared by the manifest schema for fields that are used to construct file system paths.
+        constexpr size_t s_MaxPathFieldLength = 128;
+
+        // Validates a manifest field value that is used to construct file system paths.
+        // disallowWhitespace: set for fields whose schema definition excludes whitespace (for example, PackageIdentifier).
+        std::vector<ValidationError> ValidatePathFieldValue(std::string_view fieldName, std::string_view value, bool disallowWhitespace)
```
florelis: nit: Why not "allowWhitespace"?

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -110,12 +114,101 @@ namespace AppInstaller::Manifest
...
+            {
+                auto rawCharacter = static_cast<unsigned char>(character);
+
+                // Nulls, control characters, characters that are not valid in a file system path, and (optionally) whitespace.
+                // Note that a null is rejected even though the schema pattern does not exclude it; it cannot appear in a
+                // YAML manifest, but a REST source can produce one and it would truncate any path that it is used in.
+                if (rawCharacter <= 0x1f ||
+                    rawCharacter == 0x7f ||
+                    s_InvalidPathFieldCharacters.find(character) != std::string_view::npos ||
+                    (disallowWhitespace && rawCharacter == ' '))
```
florelis: Is `' '` the only whitespace we care about?

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -110,12 +114,101 @@ namespace AppInstaller::Manifest
...
+                {
+                    resultErrors.emplace_back(ManifestError::InvalidPathCharacters, fieldNameString, valueString);
+                    break;
+                }
+            }
+
+            // The character restrictions above prevent traversal using path separators, but the value can still
```
florelis: Up to this point I thought the fields could have paths, not just path *parts* / filenames. I think we should make that clearer.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
```diff
@@ -110,12 +114,101 @@ namespace AppInstaller::Manifest
...
+            // consist solely of relative path specifiers (for instance, ".."). Reject those as well.
+            if (AppInstaller::Filesystem::PathEscapesBaseDirectory(value))
```
florelis: Maybe it is the best way to do it, but it feels to do this if we just want to check for `".."`

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/ManifestDeserializer_1_0.cpp
```diff
@@ -150,6 +151,10 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0::Json
             THROW_HR(APPINSTALLER_CLI_ERROR_RESTSOURCE_INVALID_DATA);
         }
 
+        // The schema allows surrounding whitespace in these values, but they are used to construct file system
```
florelis: nit: "these" makes me think there will be several, but it's hard to see that it also applies to the package version further down.

src/AppInstallerCommonCore/Public/winget/Manifest.h
```diff
@@ -72,4 +73,13 @@ namespace AppInstaller::Manifest
             std::function<const string_t& (const ManifestInstaller&)> extractStringFromInstaller = {},
             std::function<const string_t& (const AppsAndFeaturesEntry&)> extractStringFromAppsAndFeaturesEntry = {}) const;
     };
+
+    // Creates a file system path part from a value that originated in manifest data.
+    // Manifest validation rejects values that are not safe to use in a path, but the values can also come from
+    // sources that do not go through it (for instance, installed package data), so they are sanitized here as
+    // well. Throws if the result would point outside of the directory that it is used in.
+    std::filesystem::path GetPathPart(std::string_view value);
```
florelis: Why does this need to be public? I would think the other one would be enough. (And if it needs to be public, it feels weird to have it here)

src/AppInstallerRepositoryCore/Rest/Schema/1_0/Json/SearchResponseDeserializer_1_0.cpp
```diff
@@ -20,6 +22,24 @@ namespace AppInstaller::Repository::Rest::Schema::V1_0::Json
         constexpr std::string_view Versions = "Versions"sv;
         constexpr std::string_view PackageVersion = "PackageVersion"sv;
         constexpr std::string_view Channel = "Channel"sv;
+
+        // The package identifier and version flow into file system paths, so the manifest schema restrictions
+        // on them are enforced here as well; the schema itself is not applied to REST responses.
+        bool IsValidPathFieldValue(std::string_view fieldName, const std::vector<AppInstaller::Manifest::ValidationError>& validationErrors)
```
florelis: The naming feels weird. It says "IsValidPathFieldValue" but what it does is check if a list of errors from a previous check has hard errors.

src/AppInstallerCLITests/RestInterface_1_0.cpp
```diff
@@ -622,6 +650,45 @@ TEST_CASE("GetManifests_GoodResponse", "[RestSource][Interface_1_0]")
     sampleManifest.VerifyInstallers_AllFields(manifest);
 }
 
+TEST_CASE("GetManifests_GoodResponse_PathFieldWhitespaceTrimmed", "[RestSource][Interface_1_0]")
+{
+    // The schema permits surrounding whitespace in these values, so this is a valid response. Trimming at parse
+    // time keeps the stored value consistent with version comparison and with the file system path built from it.
```
florelis: Nit: Could you mention in the comment that only PackageIdentifier and PackageVersion have interesting data for this test?

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
JohnMcPMS: I suppose it depends on what the default is; it is preferrable for the default to be `false` in my opinion.  That doesn't actually affect the choice though since there are all of two fields.  So my biggest reason is inertia at the moment.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
JohnMcPMS: I feel like the more specific ones are useful for the more likely scenario.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
JohnMcPMS: Other whitespace are already excluded by the control character block at 150.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
JohnMcPMS: I'm not following; please elaborate.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
JohnMcPMS: I would rather call the centralized function that is more expensive than implement another version and miss something.

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
florelis: Aren't there other fancy Unicode space characters? Or is this ASCII only?

src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp
florelis: I was reading "Validate Path Fields" as "validate any field that can contain a path, including things like `foo/bar`".
It wasn't until I read this function that I realized it meant "fields that can be used as part of a file or folder name, a single part of the path"

PR #6532, opened 2026-09-18
Fix dangling view

JohnMcPMS:
## 📖 Description
Fonts added a dynamic extension without changing the return from view to string.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6532)


ranm-msft (approved):
Verified the lifetime problem rather than just the shape of the change: `fileName.extension()` hands back a temporary `path`, so the `wstring_view` built from its `c_str()` was already pointing at freed storage by the time the caller used it. Returning `std::wstring` and holding `fileName` by value closes both halves of that. The other branches were returning literals, so they were never at risk.

Nothing else from me.

Issue #184, opened 2020-05-19
JSON Formatted Output Support


RichiCoder1:
# Description of the new feature/enhancement

To enable automation and consumption via other tools, it'd be great if current and future commands supported a JSON output mode.

mikemaccana:
Modern-day Windows command line apps output structured objects. These can already be converted to JSON, Excel, CSV, or whatever other format is preferred. The issue is that winget doesn't output any structured data. See #196

RichiCoder1:
@mikemaccana As mentioned on the other two issues, PowerShell isn't available in all Windows SKUs which this is targeting. For native PowerShell support, you're looking for https://github.com/microsoft/winget-cli/issues/221.

megamorf:
@RichiCoder1 Could you list all target SKUs and/or a link to the requirements spec, please?

RichiCoder1:
It's been mentioned multiple times across multiple sources:
* https://github.com/microsoft/winget-cli/issues/221#issuecomment-632985508
* https://twitter.com/richturn_ms/status/1262973495559876608?s=20

denelon:
@RichiCoder1 I'm going through the backlog as a part of our planning exercise for the next few milestones. I'm going to close this as a Duplicate of #221.

RichiCoder1:
I'm cool with that as long as the ultimate functionality is still flexible enough to be consumed from other tools

denelon:
Yes, that is the intent. Rather than building some kind of `--output-JSON` flag, the native PowerShell implementation will provide objects that may be used, and will not have the truncated output string behavior currently exhibited in the winget.exe client.

mikemaccana:
Woo native powershell implementation! That’s exactly what I personally
wanted.

denelon:
I'm going to re-open this issue. It's not high-priority (unless we see a lot of 👍) since we have the option in PowerShell, but I can see value in what is being asked for.

We still have more work to completely move everything the CLI is calling into the COM APIs, and once that's done, providing some kind of output type/format "should" be a bit easier to implement.

Banthaf0dder:
> I'm going to re-open this issue. It's not high-priority (unless we see a lot of 👍) since we have the option in PowerShell, but I can see value in what is being asked for.
> 
> We still have more work to completely move everything the CLI is calling into the COM APIs, and once that's done, providing some kind of output type/format "should" be a bit easier to implement.

[...]

PR #6534, opened 2026-09-19
Spec: structured output for WinGet CLI

PrzemyslawKlys:
## Description

I put the JSON output proposal into a spec so we can agree on the contract before rebuilding the implementation in #6346.

The proposal starts with installed-package and available-upgrade listings. It keeps serialization in the reporter, uses PowerShell-aligned package fields where the concepts match, and defines the selected upgrade version separately from the full version list.

This revision also makes the incomplete and failure cases explicit: warnings and errors are typed, truncated results are visible to automation, partial failures return usable data with a nonzero exit code, and the first supported and unsupported command modes are listed directly in the spec.

This remains a draft for design review. It does not add `--format json` or close the tracking issue.

## Related Issues

- #184
- #6346

## Checklist

- [x] Linked to the tracking issue
- [ ] Design reviewed and agreed


PrzemyslawKlys:
I've now opened #6536 as the draft implementation of this proposal. It stays behind the experimental feature, and I'll keep it aligned with any changes we make here. The implementation branch currently includes the spec commits because GitHub cannot base a PR in this repository on a branch that exists only in a fork; once this spec lands, I'll rebase it to an implementation-only diff.

Issue #5298, opened 2025-03-13
[OPTIMIZATION] Use %LocalAppData% instead of hardcoded paths


o3wiz:
### Description of the new feature / enhancement

Winget currently installs packages to the absolute path:

```
C:\Users\<user_name>\AppData\Local\Microsoft\WinGet\Packages
```

Instead, using `%LocalAppData%` would improve flexibility and compatibility:

```
%LocalAppData%\Microsoft\WinGet\Packages
```

Trenly:
It already does use the indirect path to LOCALAPPDATA through the [KNOWNFOLDERID](https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid) Shell API

https://github.com/microsoft/winget-cli/blob/a3cea271bb84cf2a3358d97242effc348cce18a5/src/AppInstallerCommonCore/Runtime.cpp#L191-L200

https://github.com/microsoft/winget-cli/blob/a3cea271bb84cf2a3358d97242effc348cce18a5/src/AppInstallerSharedLib/Filesystem.cpp#L310-L315

o3wiz:
@Trenly 
It looks like the implementation should behave as you've described, but in practice, the absolute path still ends up being written directly into the `PATH` environment variable.

![Image](https://github.com/user-attachments/assets/dab1bf58-1cd6-43bc-801a-6680cc3dbc45)

Trenly:
@o3wiz - But isn't that the way it should be? Imagine the case where a package is installed with WinGet, then a user changes where `%LOCALAPPDATA%` points to without moving the already installed file

Trenly:
I see. . .

ArnieGA:
@Trenly 
Here's another use case example:  usernames that contain a space. It affects other applications that leverage Winget, like UnigetUI, where it causes a warning about winget having to be fixed (reset). In this case in particular, an exception is always thrown in Powershell related to the user's directory in Windows.

For example:
- **Username**: "Firstname Lastname"
- **%LocalAppData%**:  "C:\Windows\Firstname Lastname\AppData\Local"
- **Error thrown when resetting Winget**:
```
'C:\Users\Firstname' is not recognized as an internal or external command,
operable program or batch file.
ERROR: The process "winget.exe" not found.
```

It seems as if having a space in the username was not considered, or purposely left out.

whindsaks:
PathUnExpandEnvStrings

hb20007:
[...]

PR #6549, opened 2026-09-22
Store PATH entries in environment variable form

IsAvaible:
## 📖 Description

Portable installs append the fully expanded links path
(`C:\Users\<name>\AppData\Local\Microsoft\WinGet\Links`) to the user `PATH`.
It breaks silently on profile moves/renames, leaks the user name into the
registry, and has caused encoding bugs (#4317).

This PR stores entries in environment variable form when under a well-known
folder (`%LOCALAPPDATA%\Microsoft\WinGet\Links`). Values remain
`REG_EXPAND_SZ`, so Windows expands them at logon, no change for `PATH`
consumers. `Contains`/`Remove` were reworked from substring search to
per-entry normalized comparison so both old (expanded) and new
(variable-form) entries are handled safely.

### What changed

- **`AppInstallerSharedLib`:** new `Filesystem::GetUnexpandedPath()`.
  Unexpands to `%LOCALAPPDATA%`/`%APPDATA%`/`%USERPROFILE%` (User scope only)
  or `%ProgramData%`/`%ProgramFiles%`/`%ProgramFiles(x86)%`/`%SystemRoot%`,
  with slash, quote, trailing-slash (drive-root aware), and NFKC
  normalization plus separator-boundary matching.
- **`AppInstallerCommonCore` (`PathVariable`):** `Append` stores the
  unexpanded form (Machine scope never gets user vars; empty targets
  rejected); `Contains`/`Remove` compare normalized + expanded per-entry
  values (fixes prefix/subpath corruption, handles quotes, legacy entries,
  missing/empty `PATH`, overlong/malformed entries); new injectable
  constructor `(scope, key, readOnly, broadcastEnvironmentChange)`.
- **Tests + docs:** `PathVariable` tests moved to volatile registry keys (no
  admin, no broadcast); new `GetUnexpandedPath` case and 10 new
  `PathVariable` cases; Release Notes updated.

### Compatibility / Limitations

- Legacy expanded entries are detected, never duplicated, and cleanly removed;
  all other entries keep their stored form.
- Existing entries stay expanded until removed/re-added; the portable index
  still stores expanded paths (possible follow-up).

## 🔗 References

- Partially addresses #5298 (use `%LocalAppData%` instead of hardcoded paths)
- Related: #4317 (multibyte usernames, helped by variable form + NFKC)

## 🔍 Validation

**Automated** (from `src\<ARCH>\<Config>\AppInstallerCLITests`):

```
AppInstallerCLITests.exe "[pathVariable]"
AppInstallerCLITests.exe "GetUnexpandedPath"
AppInstallerCLITests.exe RefreshEnvironmentVariable_User
AppInstallerCLITests.exe VerifyPathRefreshExpandsValues
```

Covers: variable-form storage (User vs. Machine), legacy-entry dedup/removal,
subpath/prefix preservation, exact (non-substring) matching, quoted/empty/
NFKC/overlong inputs. All `PathVariable` tests use volatile keys, no admin,
no real-`PATH` modification.

**Manual:**

1. `wingetdev install <portable package>`
2. `reg query HKCU\Environment /v Path` shows `%LOCALAPPDATA%\...\Links`
3. New shell resolves the portable alias
4. `wingetdev uninstall` removes the entry, neighbors intact

## ✅ Checklist

- [x] Linked to an issue
- [x] Updated [Release Notes](../doc/ReleaseNotes.md)
- [ ] Updated documentation (not applicable)
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (not applicable)

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

> **AI Disclosure:**
> This feature was coded with Antigravitiy and Gemini Flash 3.8 Medium. I’ve tested it locally and verified the behaviour, but please feel free to suggest refinements.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-cli/pull/6549)


IsAvaible:

Trenly:
Could you add E2E coverage for the new PATH persistence behavior? The current unit tests validate normalization and registry-key behavior with injected keys, but they do not verify the complete portable install/uninstall flow against the real user or machine environment registry.

Suggested scenarios:

1. **User-scope install stores an environment-variable PATH entry**  
   Install a portable package that uses the shared Links directory, then inspect `HKCU\Environment\Path` with environment-variable expansion disabled. Verify that the raw value contains:

   ```text
   %LOCALAPPDATA%\Microsoft\WinGet\Links;
   ```

   and does not contain the expanded absolute profile path. Also verify that the normal expanded registry read resolves the entry to the current Links directory and that the portable command is available.

2. **Machine-scope install does not store user-scoped variables**  
   For an elevated machine-scope portable install, inspect `HKLM\System\CurrentControlSet\Control\Session Manager\Environment\Path`. Verify that the raw entry does not use `%LOCALAPPDATA%`, `%APPDATA%`, or `%USERPROFILE%`. Where the install path is under a system-known location, verify that an appropriate system variable such as `%ProgramFiles%` or `%ProgramData%` is used.

3. **Uninstall and cleanup**  
   Verify that uninstall removes the variable-form PATH entry and leaves unrelated PATH entries intact. The assertion should inspect both the raw and expanded registry values.

4. **Shared Links directory and deduplication**  
   Install two portable packages that use the same Links directory. Confirm that only one `%LOCALAPPDATA%\Microsoft\WinGet\Links;` entry exists. Uninstall the first package and verify that the entry remains; uninstall the second and verify that the entry is removed.

5. **Mixed user- and machine-scope installs**  
   Install one portable package at user scope and another at machine scope, with each package creating an entry in its respective Links directory. Inspect both registry locations and verify that:

   - `HKCU\Environment\Path` contains only the user Links directory, represented as `%LOCALAPPDATA%\Microsoft\WinGet\Links`.
   - `HKLM\System\CurrentControlSet\Control\Session Manager\Environment\Path` contains only the machine Links directory, represented using an appropriate machine-scoped path or variable such as `%ProgramFiles%\WinGet\Links`.
   - The user Links directory is not written to the machine PATH, and the machine Links directory is not written to the user PATH.
   - Uninstalling the user-scoped package removes only the user Links entry and leaves the machine entry intact.
   - Uninstalling the machine-scoped package then removes only the machine Links entry.

   This would ensure that the two Links directories do not collide and that variable-form storage does not cause scope leakage.

6. **Legacy expanded-entry compatibility**  
   Seed the user PATH with the old fully expanded Links path, then run the relevant install/repair and uninstall flow. Verify that WinGet recognizes and removes the legacy entry without removing adjacent PATH entries.

7. **PATH refresh expansion and ordering**  
   Exercise the install path that refreshes the current process PATH, using PATH values containing environment-variable references. Verify that the process environment contains expanded absolute paths rather than `%...%` references, with machine PATH entries preceding user PATH entries.

8. **Archive binaries that depend on PATH**  
   Install `AppInstallerTest.ArchivePortableWithBinariesDependentOnPath` and verify that the package’s install directory is added to the appropriate PATH registry value rather than the shared Links directory. Inspect the raw registry value to confirm that the entry uses the expected environment-variable form when the install directory is under a known folder, and verify that the archive’s dependent binaries can run successfully through the installed package. After uninstall, confirm that only the install-directory PATH entry is removed and that unrelated PATH entries remain intact. This should also cover the machine-scope variant to ensure cleanup uses the same scope in which the entry was created.

`TestCommon.VerifyPortablePackage` already reads both expanded and unexpanded registry values, so extending that helper with an expected raw PATH entry would likely avoid duplicating registry-inspection logic. These tests would validate the user-visible behavior and the actual registry boundary that the semantic unit tests cannot cover.

IsAvaible:
Hey @Trenly, thanks for the quick feedback! I've added the tests in [this commit](https://github.com/microsoft/winget-cli/pull/6549/commits/ac4b2b652f7ddb8120dbe9015f3ea28714a66df4). Let me know what you think :)

src/AppInstallerCLIE2ETests/Helpers/TestCommon.cs
```diff
@@ -474,16 +476,29 @@ public static void VerifyPortablePackage(
...
+            bool rawIsAddedToPath = false;
+            string rawPathValue = null;
             string pathDiagnostics;
             string pathSubKey = scope == Scope.User ? Constants.PathSubKey_User : Constants.PathSubKey_Machine;
             using (RegistryKey environmentRegistryKey = baseKey.OpenSubKey(pathSubKey, true))
             {
                 string pathName = "Path";
                 var currentPathValue = (string)environmentRegistryKey.GetValue(pathName);
-                var rawPathValue = (string)environmentRegistryKey.GetValue(pathName, null, RegistryValueOptions.DoNotExpandEnvironmentNames);
+                rawPathValue = (string)environmentRegistryKey.GetValue(pathName, null, RegistryValueOptions.DoNotExpandEnvironmentNames);
```
Trenly, line 487: Why is `var` removed here?

src/AppInstallerCLIE2ETests/Helpers/TestCommon.cs
```diff
@@ -532,12 +560,126 @@ public static bool PathContainsValue(string value, Scope scope = Scope.User)
...
 
+        /// <summary>
+        /// Gets the raw PATH registry value without environment variable expansion.
+        /// </summary>
+        /// <param name="scope">Scope.</param>
+        /// <returns>The raw PATH string.</returns>
+        public static string GetRawPathValue(Scope scope = Scope.User)
+        {
+            RegistryKey baseKey = scope == Scope.User ? Registry.CurrentUser : Registry.LocalMachine;
+            string pathSubKey = scope == Scope.User ? Constants.PathSubKey_User : Constants.PathSubKey_Machine;
```
Trenly, line 577: These ternaries duplicate the same conditional logic. An  if / else  block would evaluate the condition once and make the mutually exclusive branches clearer

Same comment for below; Only posting once to avoid multiple comments

src/AppInstallerCLIE2ETests/PortablePathVariableTests.cs
```diff
@@ -0,0 +1,640 @@
...
+                Assert.That(rawPath, Does.Contain(expectedRawEntry), $"Raw PATH value should contain {expectedRawEntry}");
+                Assert.That(rawPath, Does.Not.Contain(userLinksDirClean), "Raw PATH value should not contain the expanded absolute profile path.");
+
+                // Verify registry value kind is REG_EXPAND_SZ
+                RegistryValueKind valueKind = TestCommon.GetPathRegisterValueKind(TestCommon.Scope.User);
+                Assert.That(valueKind, Is.EqualTo(RegistryValueKind.ExpandString), "PATH registry value kind should be REG_EXPAND_SZ");
+
+                // Verify normal expanded registry read resolves to the current Links directory
+                string expandedPath = TestCommon.GetExpandedPathValue(TestCommon.Scope.User);
+                Assert.That(expandedPath, Does.Contain(userLinksDirClean), "Expanded PATH should contain the resolved Links directory.");
```
Trenly, line 60: Instead of adding `GetExpandedPathValue` and `GetRawPathValue` , it could make more sense to change the signature on `PathContainsValue` to be `PathContainsValue(string value, Scope scope = Scope.User, bool expanded = true)`  which would avoid some of the duplicate logic for fetching the registry keys. Then it's just a matter of setting the correct expansion option on the call inside `PathContainsValue`

src/AppInstallerCLIE2ETests/Helpers/TestCommon.cs
```diff
@@ -532,12 +560,126 @@ public static bool PathContainsValue(string value, Scope scope = Scope.User)
...
+            using RegistryKey environmentRegistryKey = baseKey.OpenSubKey(pathSubKey, false);
+            return (string)environmentRegistryKey?.GetValue("Path") ?? string.Empty;
+        }
+
+        /// <summary>
+        /// Gets the PATH registry value kind.
+        /// </summary>
+        /// <param name="scope">Scope.</param>
+        /// <returns>The registry value kind, or ExpandString if not found.</returns>
```
Trenly, line 599: Why `ExpandString` if not found and not `None` or `Unknown` ?

src/AppInstallerCLIE2ETests/PortablePathVariableTests.cs
```diff
@@ -0,0 +1,640 @@
...
+
+                // Verify portable command is available on disk
+                Assert.That(File.Exists(symlinkPath), Is.True, $"Portable command symlink should exist at: {symlinkPath}");
+
+                // Verify command is available and executable via PATH lookup
+                string refreshedPath = TestCommon.GetExpandedPathValue(TestCommon.Scope.Machine).TrimEnd(';') + ";" +
+                                       TestCommon.GetExpandedPathValue(TestCommon.Scope.User);
+                ProcessStartInfo startInfo = new ProcessStartInfo("cmd.exe", $"/c {Constants.AppInstallerTestExeInstallerExe} /NoOperation")
```
Trenly, line 68: Apologies for any misdirection in my original comment - I don't know that we need to actually run the command. It is probably sufficient to ensure that the path was updated like you expect. However, if you do want to verify the command is available, I'd probably use the `SearchPathW` windows API to check that the executable name resolves without needing to start the process

src/AppInstallerCommonCore/Public/winget/PathVariable.h
```diff
@@ -12,6 +12,9 @@ namespace AppInstaller::Registry::Environment
     {
         PathVariable(Manifest::ScopeEnum scope, bool readOnly = false);
 
+        // Constructor overload for dependency injection (e.g. unit testing with volatile keys).
```
Trenly, line 15: If this is for testing only, should it be wrapped by the test hooks conditional compile?

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -16,22 +16,87 @@ namespace AppInstaller::Registry::Environment
 
         void EnsurePathValueEndsWithSemicolon(std::string& value)
         {
-            if (value.back() != ';')
+            if (value.empty() || value.back() != ';')
```
Trenly, line 19: I think this has the potntial to introduce an empty PATH entry which could cause current-directory lookup. Should this be `!value.empty() && value.back() != ';'` ?

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -16,22 +16,87 @@ namespace AppInstaller::Registry::Environment
...
             {
                 value += ';';
             }
         }
 
+        // Preserves trailing slash for drive root (e.g. "C:\")
+        constexpr size_t s_DriveRootLength = 3;
+
+        // Cleans up a raw path entry by stripping leading/trailing whitespace, semicolons, and enclosing quotes.
+        void CleanPathEntry(std::wstring& entry)
```
Trenly, line 29: nit: Generally functions which may perform modification like this have `bool` return type in this codebase to indicate to the caller whether the underlying data was modified or not

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -16,22 +16,87 @@ namespace AppInstaller::Registry::Environment
...
+            {
+                return {};
+            }
+
+            std::wstring expanded;
+            try
+            {
+                expanded = Utility::ExpandEnvironmentVariables(trimmedEntry);
+            }
+            catch (...)
```
Trenly, line 65: Is this really a case we want to catch and fall back to the trimmed entry on? If the path can't be expanded, it feels like that's a case where terminating the context with an internal error would be appropriate instead of masking whatever caused the path to be un-expandable

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -16,22 +16,87 @@ namespace AppInstaller::Registry::Environment
...
+            {
+                expanded = trimmedEntry;
+            }
+
+            std::filesystem::path p{ std::move(expanded) };
```
Trenly, line 70: Please no single letter variable names.

Trenly, line 70: nit: Since the constructor builds the path from a reference based constructor, the move doesn't provide optimization here, and ownership of `expanded` isn't important since it's consumed immediately anyways.

src/AppInstallerCommonCore/PathVariable.cpp
```diff
@@ -16,22 +16,87 @@ namespace AppInstaller::Registry::Environment
...
             std::string result;
             std::vector<std::string> pathEntries = Split(value, ';');
-            for (std::string& pathEntry : pathEntries)
+            for (const std::string& pathEntry : pathEntries)
             {
                 if (!pathEntry.empty())
                 {
-                    result += AppInstaller::Filesystem::GetExpandedPath(pathEntry).u8string();
-                    result += ';';
+                    std::wstring expanded = NormalizeAndExpandPathEntry(pathEntry);
```
Trenly, line 94: `NormalizeAndExpandPathEntry` converts `pathEntry` to UTF-16, then below the result is converted back to UTF-8 if it isn't emtpy. Is there a way to avoid converting between the two encodings?

src/AppInstallerSharedLib/Filesystem.cpp
```diff
@@ -562,6 +562,115 @@ namespace AppInstaller::Filesystem
         }
     }
 
+    std::filesystem::path GetUnexpandedPath(const std::filesystem::path& path, bool allowUserVariables)
```
Trenly, line 565: There should already be helpers in `Runtime.cpp` that can be extended to do this. Specifically `ReplaceProfilePathsWithEnvironmentVariable` as an example of how we already do path collapsing, and `ReplaceCommonPathPrefix` as the method which performs the replacement. There's also `GetWellKnownFolderPath` instead of trying to expand the environment variables individually to get their path on disk.

